Issue #2324, opened 2021-06-18
the `LV_TXT_BREAK_CHARS` macro should be more flexible


guoweilkd:
in current, `LV_TXT_BREAK_CHARS   " ,.;:-_"` is friendly for english. but it's not good for chinese and other language. Because every word in Chinese can be used break char. so, We also need a range of judgment , like this:`letter >= 0x4E00 && letter <= 0x9FA5` . 

The final code like this,

```
#define LV_LETTER_IS_BREAK_CHARS(letter)  \
    ({\
        char* str = " ,.;:-_";\
        bool ret = false;\
        if (letter >= 0x4E00 && letter <= 0x9FA5) {\
            ret = true;\
        } else {\
            for(int i = 0; str[i] != '\0'; i++) {\
                if(letter == (uint32_t)str[i]) {\
                    ret = true;\
                    break;\
                }\
            }\
        } \
        ret;\
    })
```
But I don't think it's a good implementation, maybe you have a better idea.

kisvegabor:
It sounds reasonable. 

It can be easily added here https://github.com/lvgl/lvgl/blob/master/src/misc/lv_txt.c#L908

Could you send a Pull request? :slightly_smiling_face:

PR #2327, opened 2021-06-21
enhance the function of break_chars

guoweilkd:

PR for #2324

### Checkpoints

### DIFF
--- a/src/extra/widgets/span/lv_span.c
+++ b/src/extra/widgets/span/lv_span.c
@@ -50,7 +50,6 @@ static lv_opa_t lv_span_get_style_text_blend_mode(lv_obj_t * par, lv_span_t * sp
 static int32_t lv_span_get_style_text_decor(lv_obj_t * par, lv_span_t * span);
 
 static inline void span_text_check(const char ** text);
-static inline bool is_break_char(uint32_t letter);
 static void get_txt_coords(const lv_obj_t * span, lv_area_t * area);
 static void lv_draw_span(lv_obj_t * spans, const lv_area_t * coords, const lv_area_t * mask);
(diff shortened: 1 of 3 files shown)


embeddedt (commented):
Maybe it's better to remove `LV_TXT_BREAK_CHARS` entirely and put the value in `str` like you did in your issue example. I find it confusing as a reader when there are two macros which appear to do the same thing.

Adding an error message to `lv_txt.h` when `LV_TXT_BREAK_CHARS` is defined is good enough, in my opinion; users are expected to update their configuration file when moving between releases.

```c
#ifdef LV_TXT_BREAK_CHARS
#error LV_TXT_BREAK_CHARS has been replaced by LV_LETTER_IS_BREAK_CHARS. Update your config file.
#endif
```

It also simplifies `_lv_txt_is_break_char`.

kisvegabor:
I'd simply add `if ((letter >= 0x4E00 && letter <= 0x9FA5)) return true;` to `_lv_txt_is_break_char`. Altough I don't know the Chiease language but from your comment it seems it's common to break on any character. 

So is this customization really required?

The `(letter >= 0xFF00 && letter <= 0xFFFF)` range [contains arabic characters](https://www.ling.upenn.edu/courses/Spring_2003/ling538/UnicodeRanges.html) too so I think the range needs to be narrowed.
`

guoweilkd:
Yes, it's more concise. 
users can configure punctuation in `LV_TXT_BREAK_CHARS` macro.(eg:`#define LV_TXT_BREAK_CHARS " ,.;:-_\uFF01\uFF0C\uFF1A\uFF1B"`). so, I deleted the statements( `letter >= 0xFF00 && letter <= 0xFFFF`).

embeddedt (approved):

kisvegabor:
Merged, thank you very much!

PR #2346, opened 2021-07-02
Add missing lv_obj_class_t externs

amirgon:
The Micropython bindings assumes each LVGL object has a 'class' extern in the form of lv_objectname_class

Related: https://github.com/lvgl/lv_binding_micropython/issues/135

### DIFF
--- a/src/extra/widgets/animimg/lv_animimg.h
+++ b/src/extra/widgets/animimg/lv_animimg.h
@@ -29,6 +29,9 @@ extern "C" {
 /**********************
  *      TYPEDEFS
  **********************/
+
+extern const lv_obj_class_t lv_animimg_class;
+
 /*Data of image*/
 typedef struct {
     lv_img_t img;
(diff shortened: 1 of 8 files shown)


amirgon:
Actually I'm not sure about defining multiple classes to `lv_obj_class`.
That would cause ambiguity (more than one widget type has the same class) so we wouldn't be able to translate from class back to widget.  

Would it make sense to require **all** widgets to have proper class, including `lv_calendar_header_arrow`, `lv_calendar_header_dropdown` and `lv_spinner`?

embeddedt:
I think that is necessary for widget type-checking to work in C as well.

amirgon:
> Would it make sense to require **all** widgets to have proper class, including `lv_calendar_header_arrow`, `lv_calendar_header_dropdown` and `lv_spinner`?

I've added classes for all widgets where class was missing.  
In cases where there is no constructor I've left it `NULL`.  
A distinct class for each widget is required for https://github.com/lvgl/lv_binding_micropython/issues/135

After this is fixed and all widgets have classes I'll be able to integrate https://github.com/lvgl/lv_binding_micropython/pull/161 and open another PR on LVGL Python examples to remove `__cast__` which would no longer be required (on `event.get_target()` and `get_child()` for example).

kisvegabor:
In 3fb8baf I implemented the missing constructors and added the missing `extern`a you have shown in this PR.

amirgon:
> In [3fb8baf](https://github.com/lvgl/lvgl/commit/3fb8baf503411e006765020f60f295a4be16ba2d) I implemented the missing constructors and added the missing `extern`a you have shown in this PR.

Great!
Thanks.

PR #2352, opened 2021-07-06
Update win.md

wreyford:
In the paragraph **Title and buttons**
should be lv_add_btn and not lv_add_btn_right
And in the prargraph **Events**
should also be lv_add_btn


kisvegabor:
Thank you!

PR #2358, opened 2021-07-07
build_micropython.yml workflow fix

amirgon:
Micropython dev-8.0 was merged to main. Change workflow to checkout main branch instead of dev-8.0

### DIFF
--- a/.github/workflows/build_micropython.yml
+++ b/.github/workflows/build_micropython.yml
@@ -20,7 +20,7 @@ jobs:
     - name: Clone lv_micropython
       run: |
         git clone https://github.com/lvgl/lv_micropython.git .
-        git checkout dev-8.0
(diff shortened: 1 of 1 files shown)


embeddedt:
@amirgon It seems that the tests in `master` failed due to a core dump, although that did not happen on the PR itself.

amirgon:
> @amirgon It seems that the tests in `master` failed due to a core dump, although that did not happen on the PR itself.

That's strange. It's passing locally, I can't reproduce this.  
Also, when I tried to re-run this on GitHub actions it passed.

embeddedt:
The failure seems to be random. I pushed an unrelated commit to lvgl `master` and the MicroPython tests passed that time.

I guess we'll monitor it for the next little while and see what happens.

PR #2366, opened 2021-07-11
Update lv_example_calendar_1.py

amirgon:
Small improvements:

- Remove cast from get_pressed_date
- Check return value of get_pressed_date
- Call set_today_date on clicked date
- Compact highlighted_days
- Added a switch to show different header type


embeddedt:
Thanks!

PR #2374, opened 2021-07-15
perf(draw) reimplement circle drawing algorithms

kisvegabor:
I reimplemented the circle drawing. Now it's about 50% faster. 

The part of the algorithm is the calculated circle circumferences are cached and it introduced something new the masks API. Once a mask is initialized (e.g. with `lv_draw_mask_radius_init`) now it needs to be explicitly freed with `lv_draw_mask_free_param`. So it's not compatible with v8.0. Do you think it's acceptable for a so rarely used feature if we clearly communicate the change?


embeddedt:
>  now it needs to be explicitly freed with lv_draw_mask_free_param. So it's not compatible with v8.0. Do you think it's acceptable for a so rarely used feature if we clearly communicate the change?

I think it's acceptable in this specific case. Perhaps we can add a stub `lv_draw_mask_free_param` macro to 8.0.x which does nothing in order to keep source compatibility between 8.0 and 8.1.

kisvegabor:
Great! I updated the docs about the usage of masks but I am not sure where to describe these changes. Do you have any ideas?

embeddedt:
A note in the changelog should be enough. I hope users check it habitually when upgrading. Since there are less frequent minor releases, you could also announce the change on the forum.

kisvegabor:
I merge it. The docs is updated to mention the need for the extra extra `lv_draw_mask_free`.

PR #2382, opened 2021-07-18
fix(disp) Fix assert failure in lv_disp_remove

XuNeo:
### lv_disp_remove will assert fail

In lv_disp_remove, delete disp's  screens will assert failed because disp is already removed from _lv_disp_ll, thus makes object invalid.
Should firstly delete disp's objects before remove it from list.

### Checkpoints


kisvegabor:
Looks good, thanks!

PR #2384, opened 2021-07-19
Fix lv_example_snapshot_1.py

amirgon:
Allow the script to run standalone: append empty string to sys.path, initialize display, use relative path to image, raise exception on error.

With these fixes the script can run with: `micropython -i lv_example_snapshot_1.py`

### DIFF
--- a/examples/others/snapshot/lv_example_snapshot_1.py
+++ b/examples/others/snapshot/lv_example_snapshot_1.py
@@ -1,4 +1,11 @@
+import sys
+sys.path.append('') # See: https://github.com/micropython/micropython/issues/6419
+
+script_path = __file__[:__file__.rfind('/')] if __file__.find('/') >= 0 else '.'
+
 import lvgl as lv
+import display_driver
+import gc
 from imagetools import get_png_info, open_png
 
 # Register PNG image decoder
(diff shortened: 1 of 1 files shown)


kisvegabor:
Why should we make specifically this example runnable standalone?

amirgon:
> Why should we make specifically this example runnable standalone?

It's a good question.  
Today many examples under lvgl are not runnable standalone while all examples under lv_binding_micropython are.
Ideally I think all examples should be able to run standalone (for example import all the modules they need, look for resources relative to their path etc.).  

Specifically this example (and any example that imports `imagetools.py`) cannot run in the simulator today because the [JS port doesn't support `@micropython.native`](https://github.com/lvgl/lv_binding_micropython/issues/167) so the only way to run it is standalone.  
I made these changes locally just to see it running, then thought it could be useful for others.

Do you see a downside to allowing examples under lvgl to run standalone?

kisvegabor:
I see. Of course, it's better to make it easier to run the examples and it's fine to make experiments with the specific example. 

I wonder if there is a way to make the mandatory parts more user-friendly or preferably hide somehow. E.g. this one is quite cryptic:
```py
import sys
sys.path.append('') # See: https://github.com/micropython/micropython/issues/6419
script_path = __file__[:__file__.rfind('/')] if __file__.find('/') >= 0 else '.'
```

Isn't there a problem if `display_driver` is imported? Maybe the user manages the drivers in a different way with a custom display.

amirgon:
I agree, it's too cryptic this way.
Let's keep it as it is for readability.

Closing.

PR #2390, opened 2021-07-21
feat(obj) Backport keypad and encoder scrolling from v7 `lv_page` to v8 `lv_obj`

summivox:
Note to maintainers: This is my first time contributing code here. My design might not be totally in line with what you folks have in mind, but it does scratch my own itch in a reasonable way. Hope we could work something out together.

### Background

I am developing on a board with a non-touch screen, a clickable encoder, and two additional buttons.
Currently I am trying to migrate my UI to LVGL v8.

### Issue

In v8, `lv_page` is collapsed into `lv_obj`. However scrolling by keypad and encoder does not work on a v8 `lv_obj`:
- `lv_obj` does not accept arrow keys anyway
- the encoder will not go into edit mode (so it will not send arrow keys) because `lv_obj` is non-editable

### Solution

Both issues are addressed here:

- `lv_obj` will accept arrow keys if it is scrollable but not editable; this will trigger manual scrolling in the same way as v7 `lv_page`.
- When considering "editable" status in the context of encoder indev processing, scrollable is also considered "editable".

New behavior:

- Using a keypad, this simply backports the `lv_page` behavior (arrow keys => scroll).
- Using a encoder, this will make a scrollable but non-editable object appear to be editable to encoder indev handling. If the object is added to a group, the encoder will be able to enter the editing mode on it for scrolling.

Basically I want this behavior to be opt-in, which is consistent with the v7 pattern of using an explicit `lv_page` to enable scrolling of arbitrary content. This is made possible in this design:

- The scrollable flag can be easily modified on a per-object basis.
- No editable status has been touched for any widget --- this is important because the editable status is per-class.
- Scrollable containers will not be auto-added to the default group. To opt-in, the UI designer must manually add a tab stop.

Regarding the scrollable flag: It might even be desirable to disable this flag by default, and maybe even automatically re-enable when the scrollbars show up. Alternatively we can use `lv_obj_get_scroll_top` and friends to discover the actual scrollability. I'd be happy to change this behavior to whatever the maintainers feel appropriate. My gut feeling is that (akin to HTML/CSS) "declared scrollable" vs. "actually scrollable" need to be more explicit even in LVGL's internal handling.

### Example

https://user-images.githubusercontent.com/1770521/126459529-6512e550-e901-41f1-9790-bf26b55ec3c0.mp4

## Checkpoints

### DIFF
--- a/src/core/lv_obj.c
+++ b/src/core/lv_obj.c
@@ -702,6 +702,34 @@ static void lv_obj_event(const lv_obj_class_t * class_p, lv_event_t * e)
                 if(res != LV_RES_OK) return;
             }
         }
+        else if(lv_obj_has_flag(obj, LV_OBJ_FLAG_SCROLLABLE) && !lv_obj_is_editable(obj)) {
+            /*scroll by keypad or encoder*/
+            lv_anim_enable_t anim_enable = LV_ANIM_OFF;
+            lv_coord_t sl = lv_obj_get_scroll_left(obj);
+            lv_coord_t sr = lv_obj_get_scroll_right(obj);
+            char c = *((char *)lv_event_get_param(e));
+            if(c == LV_KEY_DOWN) {
+                /*use scroll_to_x/y functions to enforce scroll limits*/
+                lv_obj_scroll_to_y(obj, lv_obj_get_scroll_y(obj) + lv_obj_get_height(obj) / 4, anim_enable);
+            }
+            else if(c == LV_KEY_UP) {
+                lv_obj_scroll_to_y(obj, lv_obj_get_scroll_y(obj) - lv_obj_get_height(obj) / 4, anim_enable);
+            }
+            else if(c == LV_KEY_RIGHT) {
+                /*If the object can't be scrolled horizontally then scroll it vertically*/
+                if(!((lv_obj_get_scroll_dir(obj) & LV_DIR_HOR) && (sl > 0 || sr > 0)))
+                    lv_obj_scroll_to_y(obj, lv_obj_get_scroll_y(obj) + lv_obj_get_height(obj) / 4, anim_enable);
+                else
+                    lv_obj_scroll_to_x(obj, lv_obj_get_scroll_x(obj) + lv_obj_get_width(obj) / 4, anim_enable);
+            }
+            else if(c == LV_KEY_LEFT) {
+                /*If the object can't be scrolled horizontally then scroll it vertically*/
+                if(!((lv_obj_get_scroll_dir(obj) & LV_DIR_HOR) && (sl > 0 || sr > 0)))
+                    lv_obj_scroll_to_y(obj, lv_obj_get_scroll_y(obj) - lv_obj_get_height(obj) / 4, anim_enable);
+                else
+                    lv_obj_scroll_to_x(obj, lv_obj_get_scroll_x(obj) - lv_obj_get_width(obj) / 4, anim_enable);
+            }
+        }
     }
     else if(code == LV_EVENT_FOCUSED) {
         if(lv_obj_has_flag(obj, LV_OBJ_FLAG_SCROLL_ON_FOCUS)) {
--- a/docs/widgets/obj.md
+++ b/docs/widgets/obj.md
@@ -166,7 +166,9 @@ By default, the objects can be clicked only on their coordinates, however, this 
 Learn more about [Events](/overview/event).
 
 ## Keys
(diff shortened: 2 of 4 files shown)


src/core/lv_obj.c
```diff
@@ -702,6 +702,34 @@ static void lv_obj_event(const lv_obj_class_t * class_p, lv_event_t * e)
...
+            if(c == LV_KEY_DOWN) {
+                /*use scroll_to_x/y functions to enforce scroll limits*/
+                lv_obj_scroll_to_y(obj, lv_obj_get_scroll_y(obj) + lv_obj_get_height(obj) / 4, anim_enable);
+            }
+            else if(c == LV_KEY_UP) {
+                lv_obj_scroll_to_y(obj, lv_obj_get_scroll_y(obj) - lv_obj_get_height(obj) / 4, anim_enable);
+            }
+            else if(c == LV_KEY_RIGHT) {
+                /*If the object can't be scrolled horizontally then scroll it vertically*/
+                if(!((lv_obj_get_scroll_dir(obj) & LV_DIR_HOR) && (sl > 0 || sr > 0)))
```
summivox, line 720: This is keeping the same order as `lv_page` implementation.
My personal preference would be to avoid the condition negation here.

src/core/lv_obj.c
```diff
@@ -702,6 +702,34 @@ static void lv_obj_event(const lv_obj_class_t * class_p, lv_event_t * e)
...
+                    lv_obj_scroll_to_y(obj, lv_obj_get_scroll_y(obj) + lv_obj_get_height(obj) / 4, anim_enable);
+                else
+                    lv_obj_scroll_to_x(obj, lv_obj_get_scroll_x(obj) + lv_obj_get_width(obj) / 4, anim_enable);
+            }
+            else if(c == LV_KEY_LEFT) {
+                /*If the object can't be scrolled horizontally then scroll it vertically*/
+                if(!((lv_obj_get_scroll_dir(obj) & LV_DIR_HOR) && (sl > 0 || sr > 0)))
```
summivox, line 727: ditto

summivox, line 727: ditto

kisvegabor:
Hi,

I agree it'd be useful to add these features back. I was thinking if it can conflict if other widget features if the widget is scrollable but somehow it's managed by the widget. Anyway, I could imagine any possible issues. 

In general, I like your implementation!

Could you add a few lines about in the docs? E.g. [here](https://github.com/lvgl/lvgl/blob/master/docs/widgets/obj.md#keys)

summivox:
I wrote a brief line about this feature in the doc. Not sure how I can preview the rendered changes.

-----------------

> if it can conflict if other widget features if the widget is scrollable but somehow it's managed by the widget

> I could imagine any possible issues.

I agree even though I couldn't think of one yet, because normally an interactive widget (e.g. textarea, slider) should self-declare as `editable` --- otherwise they would have never worked with encoders to begin with.

An alternative would be to add another flag/field for "scroll-by-key"; IMHO this is redundant.
Regardless I agree that more could be done on making the behavior more bullet-proof and polished.

With such a limited board I could only help so much in testing.
Sounds like the core widget unit testing effort could be useful for changes like this ;)

As a sidenote: Do you folks have a plan for opening an officially curated "lvgl-widget-contrib" (name tentative) repo? IIRC many pluggable projects operate this way.

embeddedt:
> Not sure how I can preview the rendered changes.

You can't, outside of [GitHub's Markdown preview](https://github.com/lvgl/lvgl/blob/ad99f1726fb1b2260044db79a87fc0b2d9ca8d93/docs/widgets/obj.md#keys) (which doesn't work fully), however your change looks fine to me.

> As a sidenote: Do you folks have a plan for opening an officially curated "lvgl-widget-contrib" (name tentative) repo? IIRC many pluggable projects operate this way.

The [`src/extra`](https://github.com/lvgl/lvgl/tree/master/src/extra#extra-components folder is intended for widgets that are not considered part of the core library. Is that what you mean?

summivox:
I am aware of the `extra` folder. I had something else in mind but it's a sidenote anyway.

Please let me know if you have any concerns with this patch.
Meanwhile I'll try to exercise it in my project to see anything weird happens ;)

kisvegabor:
> I am aware of the extra folder. I had something else in mind but it's a sidenote anyway.

I'm very curious about your opinion on this. Please elaborate :slightly_smiling_face: 

> Please let me know if you have any concerns with this patch.

It seems fine, I merge it. Thanks very much!

PR #2401, opened 2021-07-23
feat(fs) add caching to lv_fs.c - WIP

PGNetHun:

WIP - do not merge yet, just review!

Add caching to file system read operations.
It can be configured during FS driver initialization (`lv_fs_drv_t`):
- `enable_cache` (bool): enable (`true`), or disable (`false`) caching
- `cache_size` (uint16): the size of cache buffer in bytes

Main functionality is in `lv_fs_read_cached` function

Please notice, that cache buffer is created for every single file, and not for the driver as a "shared" buffer!
So, for example if 2 files are opened parallel, then 2 cache buffers are used (with 512 bytes buffer size 2x512=1024 bytes will be allocated). Buffers are independent, so memory allocation and freeing up is independent, and will be done with LVGL functions (and also with MicroPython memory allocation functions).

Using a driver-level (bigger) "shared" buffer is more complex, and require locking of buffers, maybe thread sync also.
Notice also, that cache buffer is created only when there is a read for a smaller chunk of data (smaller than cache buffer size).
Cache buffer is freed up when file is closed.

"Virtual" file position is used (`lv_fs_file_cache_t` -> `file_position`) to tell the caller app the file position:
if an app reads only 16 bytes, and cache size is 512 bytes, then in reality 512 bytes are read (real file position will be 512), but for the app we need to tell, that file position is still 16 -> that's why it's called "virtual" file position.

**Important**
Caching is implemented only for read operation (and file position related functions): `lv_fs_read`, `lv_fs_seek`, `lv_fs_tell`
Write operation needs to be fixed also, because at the moment it does not handle "virtual" file position, it writes at the real file position.
So if a file is opened in RW mode (for example to overwrite some bytes), and cache buffer is used, then write could happen at wrong file position:
for example if cache buffer is 512 bytes, and there was a read of 16 bytes (-> virtual file position=16, real file position=512), then next write will happen at file position 512, and not at 16. (even when `lv_fs_seek` is called)

### Checkpoints


PGNetHun:
TEST with MicroPython

I've also created a test application in MicroPython, which can be run in MicroPython unix simulator or on real device.
[fs_perf_test.zip](https://github.com/lvgl/lvgl/files/6869367/fs_perf_test.zip)

1. Compile `lv_micropython` with FS cache modifications, build `unix` port
2. Extract files from `fs_perf_test.zip`
3. Run: `[path to lv_micropython]/ports/unix/micropython fs_perf_test.py`  (or `micropython-dev`)

Please notice, that `lvgl_fs_driver.py` is also refactored to store actual file in `LVGL_FS_File` object instead of in Python dictionary, AND `read_cb()` reads content directly into passed buffer (`file.readinto()`).

My test results (`unix` port, 14.5 kB font file):
```
Font loading speed WITHOUT cache
File close - Read count: 13321, Seek count: 199
Function "load_font" time = 108.881ms
--------------------------------
Font loading speed WITH 512 bytes cache
File close - Read count: 49, Seek count: 22
Function "load_font" time =  1.418ms
--------------------------------
Font loading speed WITH 1024 bytes cache
File close - Read count: 28, Seek count: 11
Function "load_font" time =  1.363ms
--------------------------------
Font loading speed WITH 4096 bytes cache
File close - Read count: 8, Seek count: 4
Function "load_font" time =  1.367ms
--------------------------------
Font loading speed WITH 8192 bytes cache
File close - Read count: 4, Seek count: 2
Function "load_font" time =  1.120ms
```

On real device (M5Core2) the speed increase is huge: from ~6 seconds down to ~120 ms (4096 bytes cache size).

src/misc/lv_fs.h
```diff
@@ -65,10 +65,12 @@ typedef enum {
     LV_FS_SEEK_SET = 0x00,      /**< Set the position from absolutely (from the start of file)*/
     LV_FS_SEEK_CUR = 0x01,      /**< Set the position from the current position*/
     LV_FS_SEEK_END = 0x02,      /**< Set the position from the end of the file*/
-}lv_fs_whence_t;
+} lv_fs_whence_t;
 
 typedef struct _lv_fs_drv_t {
     char letter;
+    bool enable_cache;
+    uint16_t cache_size;
```
embeddedt, line 73: Just a thought: perhaps it's better to merge these and use `cache_size == 0` to mean that the cache is disabled? Then `LV_FS_CACHE_DEFAULT_SIZE` wouldn't be needed.

embeddedt (commented):
Permissions on the files you changed went from 644 to 755. I'm guessing this is because of interactions between Windows and Linux, but can you please revert that when you get a chance? :slightly_smiling_face: It doesn't really matter from a practical perspective, but it adds extra noise to the Git history.

src/misc/lv_fs.c
```diff
@@ -114,9 +119,91 @@ lv_fs_res_t lv_fs_close(lv_fs_file_t * file_p)
...
+    }
 
     file_p->file_d = NULL;
     file_p->drv    = NULL;
+    file_p->cache  = NULL;
+
+    return res;
+}
+
+lv_fs_res_t lv_fs_read_cached(lv_fs_file_t * file_p, char * buf, uint32_t btr, uint32_t * br)
```
kisvegabor, line 138: It can be `static`, right?

src/misc/lv_fs.c
```diff
@@ -114,9 +119,91 @@ lv_fs_res_t lv_fs_close(lv_fs_file_t * file_p)
...
+    
+    if(start <= file_position && file_position < end) {
+        // Data can be read from cache buffer
+        
+        uint16_t buffer_offset = file_position - start;
+        uint16_t buffer_remaining_length = buffer_size - buffer_offset;
+        
+        if(btr <= buffer_remaining_length) {
+            // Data is in cache buffer, and buffer end not reached, no need to read from FS
+            memcpy(buf, buffer + buffer_offset, btr);
```
kisvegabor, line 155: Please use `lv_memcpy`. On some platforms the standard `memcpy` is very slow.

src/misc/lv_fs.c
```diff
@@ -114,9 +119,91 @@ lv_fs_res_t lv_fs_close(lv_fs_file_t * file_p)
...
+{
+    lv_fs_res_t res = LV_FS_RES_OK;
+    uint32_t file_position = file_p->cache->file_position;
+    uint32_t start = file_p->cache->start;
+    uint32_t end = file_p->cache->end;
+    char * buffer = file_p->cache->buffer;
+    uint16_t buffer_size = file_p->drv->cache_size;
+    
+    if(start <= file_position && file_position < end) {
+        // Data can be read from cache buffer
```
kisvegabor, line 148: Please use `/* comment */` for comments.

src/misc/lv_fs.h
kisvegabor: I agree.

kisvegabor (commented):
Thank you! I like your implementeation. I had only a few minor comemnts.

kisvegabor:
If it speeds up font loading so much and you have worked a lot with it it'd be sad to not finish it :slightly_smiling_face: 

Is there anything left with this feature?

PR #2402, opened 2021-07-23
feat(ob) add lv_is_initialized

amirgon:
Added lv_is_initialized to indicate whether LVGL was initialized

While it's possible to try initialize LVGL after it is already initialized, it warns 'lv_init: already inited' in such case. Therefore it is useful to be able to check on runtime whether LVGL was initialized or not before calling lv_init().

Also useful for making sure LVGL was initialized before trying to register a display driver


embeddedt:
Thanks! This is useful in a lot of different scenarios.

PR #2404, opened 2021-07-24
fix(table) avoid set cell value fmt corrupting table row height

summivox:

I was porting my v7 UI to v8.
There was a table that was designed to occupy almost the entire height of the screen (minus a few pixels). In v8 the not only is the table height wrong, it also broke the scrolling behavior of its parent (likely some internal state went inconsistent).

I realized that this can be repro'd in my code only when a certain (merged) cell is set using `lv_table_set_cell_value_fmt` first. If I call `lv_table_set_cell_value` first and let it finish redrawing, then the table will not become corrupted, even if I call `lv_table_set_cell_value` subsequently. 

This prompted me to look at the code. It is apparent that `lv_table_set_cell_value_fmt` had been rolling its own row height calculation, while in `lv_table_set_cell_value` this part has been refactored out to `refr_size(obj, row);`. It is likely that the inlined impl in `lv_table_set_cell_value_fmt` went out of sync in subtle ways. This patch reunifies the two codepaths.

------

--- rants below ---

While this gives the correct behavior, here we actually have an instance of [accidentally quadratic](https://accidentallyquadratic.tumblr.com/) problem: Suppose the table is R rows and C cols, then setting cell value (not necessarily formatting!) is `O(R^2*C)` instead of `O(R*C)`. It is true that you are not supposed to have a table that big (and constantly updated) in LVGL but still we could do better.

My suggestion is we can defer the update until refresh. It's certainly possible to push this burden to the user using a scope pair e.g. `lv_table_update_begin()` and `lv_table_update_end()`. Alternatively it might be possible to transparently do this internally.

I still find the v8 `lv_table` very limiting compared to v7. It used to be declarative with "cell classes"; now it's imperative and tightly coupled to the implementation details. I don't think a normal user should be touching low-level drawing data structures.

### Checkpoints
- [ ] Update the documentation --- N/A

### DIFF
--- a/src/widgets/lv_table.c
+++ b/src/widgets/lv_table.c
@@ -179,25 +179,7 @@ void lv_table_set_cell_value_fmt(lv_obj_t * obj, uint16_t row, uint16_t col, con
     va_end(ap2);
 
     table->cell_data[cell][0] = ctrl;
-
-    /*Refresh the row height*/
-    lv_coord_t cell_left = lv_obj_get_style_pad_left(obj, LV_PART_ITEMS);
-    lv_coord_t cell_right = lv_obj_get_style_pad_right(obj, LV_PART_ITEMS);
-    lv_coord_t cell_top = lv_obj_get_style_pad_top(obj, LV_PART_ITEMS);
-    lv_coord_t cell_bottom = lv_obj_get_style_pad_bottom(obj, LV_PART_ITEMS);
-
-    lv_coord_t letter_space = lv_obj_get_style_text_letter_space(obj, LV_PART_ITEMS);
-    lv_coord_t line_space = lv_obj_get_style_text_line_space(obj, LV_PART_ITEMS);
-    const lv_font_t * font = lv_obj_get_style_text_font(obj, LV_PART_ITEMS);
-
-    lv_coord_t h = get_row_height(obj, row, font, letter_space, line_space,
-                                       cell_left, cell_right, cell_top, cell_bottom);
-
-
-    lv_coord_t minh = lv_obj_get_style_min_height(obj, LV_PART_ITEMS);
-    lv_coord_t maxh = lv_obj_get_style_max_height(obj, LV_PART_ITEMS);
-
-    table->row_h[row] = LV_CLAMP(minh, h, maxh);
+    refr_size(obj, row);
 
     lv_obj_invalidate(obj);
 }
--- a/src/widgets/lv_table.c
+++ b/src/widgets/lv_table.c
@@ -36,7 +36,7 @@ static void draw_main(lv_event_t * e);
 static lv_coord_t get_row_height(lv_obj_t * obj, uint16_t row_id, const lv_font_t * font,
                                  lv_coord_t letter_space, lv_coord_t line_space,
                                  lv_coord_t cell_left, lv_coord_t cell_right, lv_coord_t cell_top, lv_coord_t cell_bottom);
-static void refr_size(lv_obj_t * obj, uint32_t strat_row);
+static void refr_size(lv_obj_t * obj, uint32_t start_row);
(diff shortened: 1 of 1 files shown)


kisvegabor:
Thanks for the detailed description.

> While this gives the correct behavior, here we actually have an instance of accidentally quadratic problem:

Probably we should replace `refr_size` with these lines to refresh only one row's height:
```c
table->row_h[i] = get_row_height(obj, i, font, letter_space, line_space,
                                       cell_left, cell_right, cell_top, cell_bottom);
table->row_h[i] = LV_CLAMP(minh, table->row_h[i], maxh);
```

> I still find the v8 lv_table very limiting compared to v7. It used to be declarative with "cell classes"; now it's imperative and tightly coupled to the implementation details. I don't think a normal user should be touching low-level drawing data structures.

I think the current method is quite flexible. We can reimplement the v7 approach on top of it. E.g. in `src/extra/widgets/table` folder create a `lv_table_cell_style.c` file 2 API functions that can be sued like this: 
```c
lv_style_t styles[4]; //Cell styles
lv_table_set_cell_styles(table, styles); //Add an event with `styles` `user_data` 
```
and
```c
lv_table_set_cell_style_type(table, row, col, 0..3); //Use lv_table_add_cell_ctrl(table, row, col, LV_TABLE_CELL_CTRL_CUSTOM_1..2); 
```

So basically have an event that mimics the v7 behaviour.


What do you think?

summivox:
Firstly, my otherwise normal table would corrupt without this patch. I didn't really bother to take a better look, but after looking, it seem that the only actual change is exactly `refr_size` refreshes this row and below, while the inlined version in `lv_table_set_cell_value_fmt` refreshes only this row. This means that at present the patched version is *necessary* unless other relevant code is refactored / updated to remove this need.

Regarding the styling: I like the idea of adding appropriate high-level APIs on top of a flexible architecture, but in this case I believe underlying implementation changes are needed. For example, `refr_size` currently only pulls style values from `LV_PART_ITEMS` to calculate the row height. I didn't look at v7 code but it used to be possible to have e.g. big font in one cell, small font in another, and still get the right row height (my UI on v7 did just that; that's how I found out...). There might be other surprises down the line as well.

kisvegabor:
> Firstly, my otherwise normal table would corrupt without this patch. I didn't really bother to take a better look, but after looking, it seem that the only actual change is exactly refr_size refreshes this row and below, while the inlined version in lv_table_set_cell_value_fmt refreshes only this row. This means that at present the patched version is necessary unless other relevant code is refactored / updated to remove this need.

I've tried it out it worked normally with the original code in a simple test using `lv_table_set_cell_value_fmt`. Please send a code snippet to reproduce the issue.

 > Regarding the styling: I like the idea of adding appropriate high-level APIs on top of a flexible architecture, but in this case I believe underlying implementation changes are needed. For example, refr_size currently only pulls style values from LV_PART_ITEMS to calculate the row height. I didn't look at v7 code but it used to be possible to have e.g. big font in one cell, small font in another, and still get the right row height (my UI on v7 did just that; that's how I found out...). There might be other surprises down the line as well.

That's right. Calculating the cells' height would be very complicated this way. What about always getting the padding from `LV_PART_ITEMS` styles but allowing to set the row height manually. E.g. it's `LV_SIZE_CONTENT` by default but it can be overwritten to any fixed value.

summivox:
After investigating my code it seems that another factor could be causing this:

I called `lv_scr_load(screen)` on the screen containing my table **from a button click event handler**. If I replace this with `lv_scr_load_anim(...)`, even with zero animation time, then the table would not be corrupted.

That kinda complicates things... I'll leave this patch here for your reference.

kisvegabor:
Thanks for the update. 

Can you put together a short cod snippet to demonstrate the issue?

PR #2406, opened 2021-07-26
fix(examples) remove symlinks

amirgon:
Remove symlinks from Python examples because the JS simulator doesn't follow them and sees the target filename instead

### DIFF
--- a/examples/others/snapshot/lv_example_snapshot_1.py
+++ b/examples/others/snapshot/lv_example_snapshot_1.py
@@ -18,10 +18,10 @@ label.set_text(" memory free:" + str(mem_free/1024) + " kB")
 
 # Create an image from the png file
 try:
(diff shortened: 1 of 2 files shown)


amirgon:
@embeddedt as a workaround I've removed symlinks from the examples.
Is it possible to support symlinks in the simulator? Currently when opening a synlink we get the target name instead of the target file itself.

embeddedt:
Apparently `raw.githubusercontent.com` URLs do not follow symlinks. :confused:

I checked the response headers and it doesn't appear that there is anything that can be used to differentiate a symlink from a text file, so I don't think I can easily implement support for symlinks unless GitHub changes the way this works in the future.

amirgon:
> so I don't think I can easily implement support for symlinks unless GitHub changes the way this works in the future.

Ok.  
So the workaround for now is to avoid symlinks in the examples.

PR #2419, opened 2021-07-30
fix(examples) Use lv.grid_fr

amirgon:
Instead of re-implementing LV_GRID_FR, use lv.grid_fr

Also use lv.GRID_TEMPLATE.LAST instead of lv.COORD.MAX where needed


amirgon:
CC @uraich - 
No need to implement C macros in Python. There should always be some backing inline function instead.

embeddedt:
Thanks.

PR #2420, opened 2021-07-30
Remove main component dependency

### LINKED ISSUE #2403: LVGL doesn't link when added to non-main ESP-IDF component
- [x] Check the related part of the [Documentation](https://docs.lvgl.io/)
- [x] Update lvgl to the latest version
- [ ] Reproduce the issue in a [Simulator](https://docs.lvgl.io/latest/en/html/get-started/pc-simulator.html)

### Describe the bug
Using lvgl as an ESP-IDF component from within another ESP-IDF component that is not the main component fails to link. Throws multiple undefined reference errors.

### To Reproduce
Instead of adding lvgl code to main.c or main.cpp try using the library in another ESP-IDF component and calling that component from main. It will fail to link.

### Expected behavior
Using lvgl from ESP-IDF components other than `main` should link properly.

### Suggested Fix
The [CMakeLists.txt file for lvgl `REQUIRES main`](https://github.com/lvgl/lvgl/blob/105bf24318a6d70d758223bb87ac44c5521cc341/CMakeLists.txt#L7) which it should not to be a true component. This seems to create a circular dependency issue. Removing the REQUIRES fixed the issue for me.

sidwarkd:
Resolves #2403

### Checkpoints

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -3,8 +3,7 @@ if(ESP_PLATFORM)
 file(GLOB_RECURSE SOURCES src/*.c)
 
 idf_component_register(SRCS ${SOURCES}
-                       INCLUDE_DIRS . src ../
-                       REQUIRES main)
+                       INCLUDE_DIRS . src ../)
 
 target_compile_definitions(${COMPONENT_LIB} PUBLIC "-DLV_CONF_INCLUDE_SIMPLE")
 


sidwarkd:
Not sure if this required a CHANGELOG entry. Just let me know if you would like it added there.

kisvegabor:
Thank you, merging.

PR #2442, opened 2021-08-07
feat(switch) add smooth animation when changing state

embeddedt:

The switch now animates smoothly when changing state as it did in 7.11 and earlier.

Based on @HX2003's work. I made a few minor changes to fix build errors and improve readability, but the implementation is 100% identical.

### Checkpoints

### DIFF
--- a/src/widgets/lv_switch.h
+++ b/src/widgets/lv_switch.h
@@ -17,11 +17,6 @@ extern "C" {
 
 #if LV_USE_SWITCH != 0
 
-/*Testing of dependencies*/
-#if LV_USE_SLIDER == 0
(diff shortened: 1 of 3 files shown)


src/widgets/lv_switch.h
```diff
@@ -32,8 +27,15 @@ extern "C" {
  *      TYPEDEFS
  **********************/
 
+typedef struct {
+    lv_obj_t * sw;
+    int32_t anim_state;
+} _lv_switch_anim_t;
```
kisvegabor: Do we really need this `struct`? The switch can have on one knob so it seems we can store `anim_state` in `lv_switch_t`.

kisvegabor:
Looks good, thank you!

PR #2444, opened 2021-08-09
fix(cmake) require 'main' for Micropython

amirgon:
Commit f2c2393b305cd71d2fc01ff8945965dccb8488b4 (#2420) breaks lv_micropython esp32 port. Apparently main dependency is require for LVGL when building Micropython, otherwise GC related code breaks

This commit adds 'main' dependency to LVGL conditionally only if building inside lv_micropython

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2,8 +2,16 @@ if(ESP_PLATFORM)
 
 file(GLOB_RECURSE SOURCES src/*.c)
 
-idf_component_register(SRCS ${SOURCES}
-                       INCLUDE_DIRS . src ../)
+idf_build_get_property(LV_MICROPYTHON LV_MICROPYTHON)
+
+if (LV_MICROPYTHON)
+    idf_component_register(SRCS ${SOURCES}
+                           INCLUDE_DIRS . src ../
+                           REQUIRES main)
+else()
+    idf_component_register(SRCS ${SOURCES}
+                           INCLUDE_DIRS . src ../)
+endif()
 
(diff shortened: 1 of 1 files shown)


embeddedt:
If ESP32 depends on our `CMakeLists.txt` file, should I add it to our Micropython CI workflow? We can only test the build, not the actual execution, but that would prevent this breakage from occurring in the future.

EDIT: I just re-read your comment and now I'm confused. Was this a build error or a runtime error?

embeddedt (approved):

amirgon:
> should I add it to our Micropython CI workflow?

It's possible, in fact lv_micropython [already runs a CI for building esp32 port](https://github.com/lvgl/lv_micropython/blob/master/.github/workflows/ports_esp32.yml) and it will fail if we update now the LVGL git submodule to LVGL head.

However, LVGL CI already builds and runs the micropython *unix port* which covers almost everything.  
The only thing I can think of where esp32 port build would fail while the unix port pass is the cmake script (the unix port uses Make) and possibly some differences in gcc version.

So, while we can add esp32 lv_micropython build CI to LVGL, its added value would be relatively small.

> Was this a build error or a runtime error?

It was a build error, specific to esp32 port.

kisvegabor (approved):

embeddedt:
>The only thing I can think of where esp32 port build would fail while the unix port pass is the cmake script (the unix port uses Make) and possibly some differences in gcc version.
> So, while we can add esp32 lv_micropython build CI to LVGL, its added value would be relatively small.

Let's leave it for now, then; we can add it easily later if it becomes more valuable to be tested.

PR #2460, opened 2021-08-17
feat(display) add direct_mode drawing mode

kisvegabor:

This PR implements a new drawing mode in which a screen sized buffer needs to be provided for LVGL.  LVGL draws directly into that buffer on absolute coordinates. It means the buffer always store the real image that should be on the screen.

With this feature the union of the changed areas can be sent to the display in the blank period. 

If 2 buffers are used they needs to be synchronized manually. E.g. copy the changed areas to the inactive buffer on the last refreshed area. 

Some related useful functions and data:
- `lv_disp_flush_is_last(disp_drv)` tells if it's the last refreshed area i nthe flush callback
- `_lv_refr_get_disp_refreshing()` gives the display being refreshed
-  `disp->inv_areas[LV_INV_BUF_SIZE]` contains the invalidated areas
- `disp->inv_area_joined[LV_INV_BUF_SIZE];` if `1` that area was joined into an other one and ignored
- `disp->inv_p` number of valid elements in `inv_areas`

cc @sebastienfagard 

### Checkpoints


Miguel0101:
How it works? Does LVGL draw directly to the buffer? Which blank period?

kisvegabor:
The framebuffer still should be in the MCU's memory. The PR doesn't add any synchronization features with VSYNC.

Miguel0101:
Does this speed up the drawing buffer?

kisvegabor:
In general not, but in some cases it avoids coping LVGL buffer to the frame buffer.

sebastienfagard:
Hi, I tested this PR with single buffer on i.MXRT595 and DSI Command Mode panel. It works fine but it does not seem to improve the performance (frame rate). The reason for that seems to be that DSI transfer to panel must be finished before the cpu/gpu can draw again into the buffer, this creates a latency in the refresh loop. The only way to remove this latency is to copy with GPU the damaged area into a separate display buffer, as the GPU is overall faster than DSI transfer it reduces the latency, but it is still there. 
Do you have a trick to get a better performance?

kisvegabor:
Hi,

Do you have an idea/measurement data how long does it take to send a whole frame via DSI and to copy it with GPU?

FASTSHIFT:
@kisvegabor 
Is this PR going to merge?

kisvegabor:
In my test it worked as expected but waiting for @sebastienfagard's feedback.

@FASTSHIFT do you also need this feature? If you can describe your setup to see how it fits and if there anything else to improve?

fraengers:
Is this supposed to be working with rotation? Because it's not working for me. I might be doing something wrong though.

kisvegabor:
@fraengers I haven't tested it but it should work with rotation too. What's the problem?

fraengers:
Setup: STM32H7 (no RTOS) with a 320x480px display (8bit parallel interface on FMC) which is mounted rotated cw by 90°, so my GUI is 480x320px.

Without direct mode it looks like this, which is how it's supposed to look:
screen_sw_rotation

But with direct mode it looks like this:
screen_directmode

Code:
```c
#ifndef DIRECTMODE
static lv_color_t buf1[320 * 480 / 10];
#endif

void GUI_Init(void) {
	lv_init();
#ifndef DIRECTMODE
	lv_disp_draw_buf_init(&draw_buf, buf1, NULL, 320 * 480 / 10);
#else
	lv_disp_draw_buf_init(&draw_buf, (uint32_t *)0x24035000, NULL, 320 * 480);  
// 0x24035000 is the start of my sram display buffer
#endif
	static lv_disp_drv_t disp_drv = {}; 
	lv_disp_drv_init(&disp_drv); 
	disp_drv.flush_cb = my_disp_flush; 
	disp_drv.draw_buf = &draw_buf;  
	disp_drv.hor_res = 320;
	disp_drv.ver_res = 480; 
	disp_drv.rotated = LV_DISP_ROT_90;
	disp_drv.sw_rotate = 1;
#ifdef DIRECTMODE
	disp_drv.direct_mode = 1;
#endif
	lv_disp_drv_register(&disp_drv);
//...
}

void my_disp_wait(lv_disp_drv_t * disp) {
	lv_disp_flush_ready(disp);
}

void my_disp_flush(lv_disp_drv_t * disp, const lv_area_t * area, lv_color_t * color_p) {
#ifndef DIRECTMODE
			LCD_DrawDot(x, y, (*color_p).full);
			color_p++;
		}
	}
#endif
	lv_disp_flush_ready(disp);
}
```

In normal mode, it renders the rotated data in the 1/10 size buffer and that gets flushed per software into my full sized buffer (320x480px). (The posted images are the content of this buffer, just rotated into landscape orientation.) The microcontroller continuously writes this buffer per DMA at 10 Hz to the displays GRAM.

Trying different settings just results in other stange renderings. Here i swapped `disp_drv.hor_res` and `disp_drv.ver_res` in direct mode.
screen_directmode_480x320

Ultimately, what I want to achieve is to skip the step where the software has to copy the smaller rendering buffer to an area in my full sized buffer. The idea is to get a little bit more fps (right now i get only about 5fps).

FASTSHIFT:
> In my test it worked as expected but waiting for @sebastienfagard's feedback.
> 
> @FASTSHIFT do you also need this feature? If you can describe your setup to see how it fits and if there anything else to improve?

Yes, I am trying to find suitable hardware for testing.

sebastienfagard:
> Hi,
> 
> Do you have an idea/measurement data how long does it take to send a whole frame via DSI and to copy it with GPU?

Hi @kisvegabor , I have measured the time spent transferring the draw-buffer over DSI for the circular Smart panel 400x400: it takes around 12ms for the transfer alone, and you also need to add a delay of [0;16]ms waiting for the TE interrupt, to avoid tearing. Which results in an average 20 ms spent waiting for the transfer to end before LVGL can draw again in the buffer.
In comparison copying the draw-buffer dirty areas into an intermediate display-buffer takes around 2ms using GPU. Once the copy is done LVGL can continue rendering in the draw-buffer.
The DSI configuration on RT595 board uses only 2 data lanes, but even if you connect the smart using 4 data lanes, the "Direct-mode" will not provide a better frame rate overall. The only advantage of "direct-mode" seems to be the low usage of graphical RAM.
Do you agree with the following?

kisvegabor:
@sebastienfagard 

Yes I agree. It seems "direct mode" is not better than the current buffering options of LVGL. 

Unless someone can show a use-case where "direct mode" has significant advantages I close this PR.

MouriNaruto:
@kisvegabor 

I guess the direct mode may also useful for win32drv to improve the performance. But I need some test.

Reference: https://github.com/lvgl/lv_drivers/issues/185

Kenji Mouri

FASTSHIFT:
@kisvegabor 
I tested it on STM32F429, using `direct_mode` can save one frame of buffer, which is very meaningful for MCUs with very limited RAM space. I think `direct_mode` can replace the old `full_refresh` mode because it is more efficient.

xiaoxiang781216:
@kisvegabor could you include this feature in 8.1 release? this patch is useful in some case.

kisvegabor:
> I tested it on STM32F429, using direct_mode can save one frame of buffer, which is very meaningful for MCUs with very limited RAM space. I think direct_mode can replace the old full_refresh mode because it is more efficient.

If you use 1 frame buffer, and LVGL renders directly there the process of rendering can be visible on the screen resulting in flickering. 


Anyway, the the display drivers can be very divers and it might be useful for some people so I merge it as you suggested.

FASTSHIFT:
> > I tested it on STM32F429, using direct_mode can save one frame of buffer, which is very meaningful for MCUs with very limited RAM space. I think direct_mode can replace the old full_refresh mode because it is more efficient.
> 
> If you use 1 frame buffer, and LVGL renders directly there the process of rendering can be visible on the screen resulting in flickering.
> 
> Anyway, the the display drivers can be very divers and it might be useful for some people so I merge it as you suggested.

I am using a 2-frame buffer. If you use the original rendering method, you need at least 3 frame buffers (one frame of LVGL is used for graphics rendering, and the other two frames are exchanged framebuffer) to ensure the integrity of the image

Issue #2459, opened 2021-08-17
Removing of an event callback using the _lv_event_dsc_t address does not work


glueckm:
### Describe the bug
When you add an event callback to an object the address for the event_dsc will be returend.
This can be used to remove the event callback at a later time.

But, if you add a second callback, the internal array ```obj->spec_attr->event_dsc``` get's ```remalloced``` which can lead to a new location of the WHOLE array.
Therefore when you try to remove the event callback by using the address returned by ```lv_obj_add_event_cb``` it is possible that this address is no longer valid and the removale of the callback fails.

kisvegabor:
That's true. We can either store only pointers to `event_dsc`s (has extra memory cost) or remove the option of removing event callback this way.

glueckm:
For my use case a need the option with the event descriptors because use the same callback multible times and therfore cannot use the callback variant. 

Or we add the option to add the user data for removing. The callback/userdata combination is unique. 

Martin

kisvegabor:
> Or we add the option to add the user data for removing. The callback/userdata combination is unique.


I like this idea. @embeddedt what do you think?

embeddedt:
I agree as well.

glueckm:
I have already implemented the feature in my project.
Should I provide PR?
if yes, how should I name the new function?

kisvegabor:
I've just checked how to implement it and stucked at naming too. :smile: 

So now we have
```c
//Remove an event with matching callback
bool lv_obj_remove_event_cb(struct _lv_obj_t * obj, lv_event_cb_t event_cb);

//Remove an event with matching dsc
bool lv_obj_remove_event_dsc(struct _lv_obj_t * obj, struct _lv_event_dsc_t * event_dsc);
```

And now we need
```c
//Remove an event with matching callback and user_data
bool lv_obj_remove_event_cb_user_data(struct _lv_obj_t * obj, lv_event_cb_t event_cb, const void * user_data);
```

But this name is ugly. What about simply `lv_obj_remove_event`?

glueckm:
> But this name is ugly. What about simply lv_obj_remove_event?

I have no problem with this name... But it feels like it does not fit well with the other names.
```lv_obj_remove_event_cb_user_data``` does fit better compare to the other names

embeddedt:
`lv_obj_remove_event_with_user_data`?

glueckm:
> ```lv_obj_remove_event_with_user_data```
Sounds better

[...]

PR #2492, opened 2021-08-25
Error with function “lv_obj_remove_event_cb”

reinhard-qian:
It may be missing a line of code:
obj->spec_attr->event_dsc[i].filter = obj->spec_attr->event_dsc[i+1].filter;
The “lv_obj_remove_event_dsc” function is the same


reinhard-qian:
Error with function “lv_obj_remove_event_cb”

kisvegabor:
Great catch, thank you.

To avoid similar issues it seems safer to use: 
```c
obj->spec_attr->event_dsc[i] = obj->spec_attr->event_dsc[i+1];
```

reinhard-qian:
Wow, you're right. It wasn't safe before

kisvegabor:
Could you modify the PR accordingly?

reinhard-qian:
All right, it's changed

embeddedt (approved):
Note that (I think) we are planning to deprecate/remove `lv_obj_remove_event_dsc` in the near future because of #2459.

kisvegabor:
Thank you!

reinhard-qian:
I'm glad to be involved！

PR #2495, opened 2021-08-25
ci(tests) convert Makefile to CMake

### LINKED ISSUE #2474: Converting `tests/Makefile` to CMake
### Introduce the problem
1. LVGL tests are too closely coupled to the OS (AKA don't run on Windows).
2. LVGL tests do not build and run in parallel. (we do `make -j`, but only one make at a time).

### Examples and cases
1. The dominant OS for LVGL development is probably Linux, having the tests be platform agnostic would be beneficial, as more developers could run them before uploading to GitHub.
2. The tests would run approx. 4x faster on GitHub (maybe locally too) if run in parallel. 

### Suggested solution
One solution is to convert `tests/Makefile` to a CMake project file. The LVGL project files are already CMake, so this would eliminate one project file format from the mix. It would collect all build artifacts into a common subdirectory, differently named for each test configuration. This would allow parallel build/execution, as well as simplifying the cleaning task.

cmumford:
This change switches the building of LVGL tests to use cmake
which is more portable than make. Additionally, whenever
cmake can be used, instead of the Python script (`main.py`),
the former is preferred.

The interface to `main.py` is unchanged, and tests are built
and executed the same as before.

This closes https://github.com/lvgl/lvgl/issues/2474.


cmumford:
Failed due to `No such file or directory: 'gcovr'`. I'll look into it.

C47D:
@cmumford Do you think it would make sense to use python virtual environment inside the test directory? Also we could provide a requirements.txt file, and install all the missing modules (like gcovr) while in CI based on this file.

cmumford:
> @cmumford Do you think it would make sense to use python virtual environment inside the test directory? Also we could provide a requirements.txt file, and install all the missing modules (like gcovr) while in CI based on this file.

Hi Carlos. I think the main reason for a virtual environment is so that module versions can be more easily controlled. `main.py` only uses build-in Python modules, so I don't see a benefit to using a venv.

```python
import glob
import shutil
import subprocess
import sys
import os
```

Today, `gcovr` is a Python script, but it was never imported or used as a Python module. I'm really not sure why it was found before, but was not in this change. My only guess is that subprocess.check_call() doesn't search the same path as os.system(). I solved it by doing a `apt-get install gcovr` as it is a Debian package and more easily installed that way instead of pip.

Back to Python - I would really love it if `main.py` could be eliminated entirely. All it does is determine the test case names and use that for auto-generating the test runner source, and for executing the tests themselves. Maybe a subject for a future PR.

LMK if you still think that a virtual environment is a good idea and maybe I can look into that.

C47D:
Hi @cmumford,

It seems to me that I arrived a little late to the party.

> Today, gcovr is a Python script, but it was never imported or used as a Python module.

I mentioned gcovr being a python module because of the requirements list found in the [tests/README](https://github.com/lvgl/lvgl/tree/master/tests#requirements-linux) file, I saw 05145cd and realized it's not being installed as a python module. So if gcovr isn't a python module we can totally forget about venv 😄 . I'm getting gcovr installed in my WSL based Ubuntu as you did in 05145cd.

> Back to Python - I would really love it if main.py could be eliminated entirely. All it does is determine the test case names and use that for auto-generating the test runner source, and for executing the tests themselves. Maybe a subject for a future PR.

I used Unity for a bit before and AFAIK you should be able to generate the test runners with Ruby, I'm using that in [here](https://github.com/C47D/WFlash/blob/master/test/Makefile#L47)

```
# Remove wflash_ll from the dependencies so it doesn't clash with the fake
test_wflash: test_wflash.c ../src/wflash.c # ../src/wflash_ll.c
	# Generate runner with ruby
	ruby Unity/auto/generate_test_runner.rb $@.c $@.runner.c
	# Compilamos
	$(CC) $(CFLAGS) $^ -o $@_Runner.out $@.runner.c
	# Ejecutamos el Runner generado
	./$@_Runner.out
```

What I don't like about the current `main.py` script is that I don't have any feedback when the build step starts, I killed the process because I didn't knew if it was running or not.

Regards

C47D:
@cmumford I noticed you've also updated the requirement list in this PR, can we add installing `libpng-dev` in the `tests/README.md`?

cmumford:
> @cmumford I noticed you've also updated the requirement list in this PR, can we add installing `libpng-dev` in the `tests/README.md`?

Done - thx for pointing that out. In a future PR I plan on moving all dependency package installations to a Bash script. Then the README.md and WorkFlows will just refer to or run something like `installdeps.sh`.

I thought about auto-generating the test runner code in the cmake file - similar to your make snippet above. I left it as-is in this PR mostly to keep it somewhat simple and "reviewable". But, I do like the of moving as much work as possible into the cmake file as it is a bit more portable.

Regarding your statement:

> What I don't like about the current `main.py` script is that I don't have any feedback when the build step starts, I killed the process because I didn't knew if it was running or not.

Do you mean the one in this PR? I get feedback right away, but maybe that's because I have a SSD. I'll look into where there may be a perceived hang, but LMK where you think that might be. In most places there are "doing XXX..." type messages, where the last print flushes to stdout.

C47D:
> Do you mean the one in this PR? I get feedback right away, but maybe that's because I have a SSD. I'll look into where there may be a perceived hang, but LMK where you think that might be. In most places there are "doing XXX..." type messages, where the last print flushes to stdout.

Yes, the one in here, I got a pretty old machine, so I guess it's because of that. Thanks for adding the missing dependency to the requirement list.

kisvegabor (commented):
@cmumford Looks like a great upgrade. Your version is much more professional than mine was :slightly_smiling_face: 

Looks good from my side. If @embeddedt also approves the PR we can merge it.

kisvegabor:
@C47D 
> What I don't like about the current main.py script is that I don't have any feedback when the build step starts, I killed the process because I didn't knew if it was running or not.

In the current version (not in this PR) make is called with `-s` (silence) because due to the large amount of build logs it was difficult to see and find which "build case"  caused error/warning.

embeddedt (approved):

cmumford:
> In the current version (not in this PR) make is called with `-s` (silence) because due to the large amount of build logs it was difficult to see and find which "build case" caused error/warning.

I think the only to build silently is to specify that cmake use make as the build system generator. I was hoping to leave that unspecified so that it could decide (make, ninja, etc.). What do you think about splitting out build and test in the workflow into two separate steps? I could even do each build in its own step.

kisvegabor:
> What do you think about splitting out build and test in the workflow into two separate steps? I could even do each build in its own step.

IMO CMake is flexible enough and acceptable by the vast majority of the developers. Before adding extra complexity (layers) I suggest waiting until there is a real need for it. 

It's only my 2 cents. I'm surely not the most experienced person here in the fields of CI and testing.

kisvegabor:
Merging, thank you very much! :slightly_smiling_face:

PR #2498, opened 2021-08-26
fix scroll example sqort types

shuai532209720:
line 32 of lv_example_scroll_6.c, if LV_USE_LARGE_COORD not configured,
x_sqr will overflow when r is greater than 256.

Signed-off-by: liushuai25 <liushuai25@xiaomi.com>


kisvegabor:
Thanks!

PR #2505, opened 2021-08-28
ci(tests): Build Unix port MicroPython submodules in parallel.

cmumford:
Added missing `-j <numcores>` when building MicroPython submodules.

### DIFF
--- a/.github/workflows/build_micropython.yml
+++ b/.github/workflows/build_micropython.yml
@@ -24,7 +24,7 @@ jobs:
     - name: Initialize lv_bindings submodule
       run: git submodule update --init --recursive lib/lv_bindings
     - name: Update Unix port submodules
-      run: make -C ports/unix VARIANT=dev DEBUG=1 submodules
+      run: make -C ports/unix -j $(nproc) VARIANT=dev DEBUG=1 submodules
     - name: Checkout LVGL submodule
       working-directory: ./lib/lv_bindings/lvgl
       run: |


cmumford:
Looks like the CI step in which this is included drops from 51 sec. to 41 sec. Nothing massive, but I'll take it.

embeddedt:
Have you seen a measurable performance improvement from this? I also thought of doing it in the past but [the target](https://github.com/lvgl/lv_micropython/blob/1fcd6e97f29f817a56237bf70d55edb159c8ccf6/py/mkrules.mk#L201-L205) doesn't look like it can be parallelized. I wonder if your 10 second improvement is caused by Actions being less loaded at this time of night.

cmumford:
> Have you seen a measurable performance improvement from this? I also thought of doing it in the past but [the target](https://github.com/lvgl/lv_micropython/blob/1fcd6e97f29f817a56237bf70d55edb159c8ccf6/py/mkrules.mk#L201-L205) doesn't look like it can be parallelized.

Good point. Maybe the gain of 10 sec. was just a measurement error. We could either leave the `-j` where it won't do anything unless the make target improves, or I can delete it, and add a workflow comment to explain it's absence. What do you prefer?

embeddedt:
I'd prefer removing it, only because upstream is not running it with `-j`, and I'd rather not deal with random CI failures that could result if they change the way it works behind the scenes.

cmumford:
SGTM. I'll abandon this and add add a comment in another PR to prevent other devs from trying the same.

PR #2508, opened 2021-08-28
[WIP] ci(tests) compile with -Wdiscarded-qualifiers

embeddedt:
See https://forum.lvgl.io/t/compiler-warnings-assignment-discards-const-qualifier-from-pointer-target-type/6502.


kisvegabor:
Hmmm, why don't we see errors?

embeddedt:
Not sure. I need to do more digging - ran out of time before. I couldn't see the errors locally either. I'm suspicious that it's related to optimization.

PR #2509, opened 2021-08-28
fix(color) Bad cast in lv_color_mix() caused UB with 16bpp or less

kevinpt:
Left shift overflows when lv_color_t is 16 bits or less because of cast outside parens.

Note that this showed up in a UBSan warning. Other devs should consider enabling it on platforms that support it.


embeddedt (approved):

kisvegabor:
Thanks! 
I always build with `-fsanitize=undefined` on gcc 7.5.0 but haven't get warnings yet. Maybe I should update it :)

PR #2510, opened 2021-08-29
Add dependency on GNU parallel.

cmumford:
lv_binding_micropython added a dependency on GNU parallel[1] to allow
the tests to be run in parallel.

[1] https://www.gnu.org/software/parallel/


cmumford:
Only needs to land if https://github.com/lvgl/lv_binding_micropython/pull/172 is merged.

PR #2512, opened 2021-08-31
fixes for rp2 port (format strings only)

eudoxos:
Fix compilation errors due to type width mismatch on Raspberry Pi Pico:
```
/home/eudoxos/build/lv_micropython/lib/lv_bindings/lvgl/src/extra/widgets/spinbox/lv_spinbox.c: In function 'lv_spinbox_updatevalue':
/home/eudoxos/build/lv_micropython/lib/lv_bindings/lvgl/src/extra/widgets/spinbox/lv_spinbox.c:429:43: error: format '%d' expects argument of type 'int', but argument 4 has type 'int32_t' {aka 'long int'} [-Werror=format=]
  429 |     lv_snprintf(digits, sizeof(digits), "%d", LV_ABS(spinbox->value));
      |                                          ~^
      |                                           |
      |                                           int
      |                                          %ld

```
Part of https://github.com/lvgl/lv_micropython/pull/42


eudoxos:
An alternative (less readable, but fully portable) aproach is to use the `PRId32` macro from `inttypes.h` (c99) for int32_t:

```c
#include<inttypes.h>
/* ... */
lv_snprintf(buf, sizeof(buf), "[%s]\t(%" PRId32 ".%03" PRId32 ", +%" PRId32 ")\t %s: %s \t(in %s line #%d)\n", lvl_prefix[level], t / 1000, t % 1000, t - last_log_time, func, msg, &file[p], line);
```
Would this be preferred?

embeddedt:
This change on its own is not enough, as the opposite error now appears on platforms where `int32_t` is defined to be `int`.

`inttypes.h` is the cleanest solution, but I don't know whether it's available everywhere, even though it's required by C99.

eudoxos:
Not sure whether `lv_printf.h` is the best place to `#include<inttypes.h>` but it gets included from `lvgl.h`.

eudoxos:
> inttypes.h is the cleanest solution, but I don't know whether it's available everywhere, even though it's required by C99.

Based on [this comment](https://github.com/libgit2/libgit2/pull/3477) even MSVC added `inttypes.h` back in 2013.

kisvegabor:
I've just tested it but Microchips XC16 doesn't have `#include <inttypes.h>`.

eudoxos:
> I've just tested it but Microchips XC16 doesn't have `#include <inttypes.h>`.

One could either check for `inttypes.h` and export `HAVE_INTTYPES_H`, but it seems LVGL does not have feature tests like this for other things. The easiest might be then just to forget about `inttypes.h` and define the respective macros by hand, like this?

```c
#if INT_MAX==32767
   #define PRId32 "d" 
#elif INT_MAX==2147483647
   #define PRId32 "ld"
#else
   #error Unhandled int width?
#endif
```

kisvegabor:
> The easiest might be then just to forget about inttypes.h and define the respective macros by hand, like this?

I like it but we should use prefixed define names to avoid collisions.

eudoxos:
> I like it but we should use prefixed define names to avoid collisions.

Like this?

kisvegabor:
I can't comment on the Cmake changes but the printf related changes looks good to me. If @embeddedt  or @amirgon say the CMake part is ok, we can merge it. 

Thank you!

kisvegabor:
@cmumford could you take a look at the failing CI? It shouldn't be related to this PR.

eudoxos:
> @cmumford could you take a look at the failing CI? It shouldn't be related to this PR.

It is, unfortunately. I will check it out later.

```
../../lib/lv_bindings/lvgl/src/extra/widgets/chart/lv_chart.c: In function ‘draw_y_ticks’:
../../lib/lv_bindings/lvgl/src/extra/widgets/chart/lv_chart.c:1409:43: error: format ‘%ld’ expects argument of type ‘long int’, but argument 4 has type ‘int32_t’ {aka ‘int’} [-Werror=format=]
 1409 |             lv_snprintf(buf, sizeof(buf), "%" LV_PRId32, tick_value);
      |                                           ^~~            ~~~~~~~~~~
      |                                                          |
      |                                                          int32_t {aka int}
```

CMakeLists.txt
```diff
@@ -1,77 +1,96 @@
...
-endif() # CONFIG_LVGL
+else(ESP_PLATFORM)
 
-else()
+    file(GLOB_RECURSE SOURCES src/*.c)
+    file(GLOB_RECURSE EXAMPLE_SOURCES ${CMAKE_CURRENT_LIST_DIR}/examples/*.c)
 
-file(GLOB_RECURSE SOURCES ${CMAKE_CURRENT_LIST_DIR}/src/*.c)
-add_library(lvgl STATIC ${SOURCES})
+    if(MICROPY_DIR)
```
amirgon: ESP32+Micropython already use LVGL as a library, and this change would break it.  

Your changes are specific to the RPi port, and [as discussed](https://github.com/lvgl/lv_micropython/pull/42#discussion_r700865385) I think it's best to look for a way to do something similar to the ESP32 here instead of adding LVGL sources directly to Micropython.

cmumford:
> @cmumford could you take a look at the failing CI? It shouldn't be related to this PR.

The LVGL tests were failing due to:

```
Error: /home/runner/work/lvgl/lvgl/tests/src/lv_test_indev.c:2:10: fatal error: ../lvgl.h: No such file or directory
    2 | #include "../lvgl.h"
      |          ^~~~~~~~~~~
```

But I think that @amirgon identified the root cause in the comment above.

eudoxos:
I reverted all `CMakeLists.txt` so that this PR is only about formats things and can get merged.

The Micropython module discussion will continue in https://github.com/lvgl/lv_micropython/pull/42 .

embeddedt (approved):

kisvegabor:
Thank you!

amirgon:
@eudoxos This PR breaks Micropython ESP32 port.
See https://github.com/lvgl/lv_micropython/pull/43/checks?check_run_id=3519395094

```
[ 93%] Building C object esp-idf/lvgl/CMakeFiles/__idf_lvgl.dir/src/extra/widgets/imgbtn/lv_imgbtn.c.obj
/home/runner/work/lv_micropython/lv_micropython/lib/lv_bindings/lvgl/src/extra/widgets/chart/lv_chart.c:1409:43: error: format '%ld' expects argument of type 'long int', but argument 4 has type 'int32_t' {aka 'int'} [-Werror=format=]
             lv_snprintf(buf, sizeof(buf), "%" LV_PRId32, tick_value);
                                           ^~~            ~~~~~~~~~~
```

@embeddedt You  [recently suggested](https://github.com/lvgl/lvgl/pull/2444#issuecomment-895562031) adding MP-ESP32 CI to LVGL. I'm starting to think this is a good idea...

eudoxos:
> This PR breaks Micropython ESP32 port.

I understand your pain of an extra nuisance. The cross-platform perspective is that it was broken already, only working accidentally.

The current logic is depends on `LONG_MAX`: if it is `2**31`, then `int_32t` should be formatted as `%ld`, and if not (meaning likely a 64-bit system), it is a `%d`. This covered those ports I (and the CI) could test. (Branching by `INT_MAX` did not work as I had thought.)

Unless someone can figure out the logic better (there is nothing at SO), `LV_PRId32` definition must be port-specific. The easiest might be to `#include<inttypes.h>` for all archs, except for explicitly handled cases where that header is missing (like XC16 @kisvegabor mentioned), which I suspect won't be that many:

```c
#if ARCH_XC16
   #define LV_PRI32d "l"
#elif ARCH_SOMETHING_ELSE
   #define LV_PRI32d "ld"
#else 
   #include<inttypes.h>
   #define LV_PRI32d PRI32d
#endif  
```
Would that be sufficient?

eudoxos:
Continuing in #2532 .

PR #2515, opened 2021-09-01
ci(tests): build and run tests in parallel.

cmumford:
Create one GitHub workflow job for each build option build
or test execution. This allows each of the five builds to
be executed in parallel as well as making it easier to
navigate to the test output as it now resides in its own
job.

This change **does** change the command-line arguments to
the test execution script. This is required to allow the
build options name to be passed in.

New `tests/main.py` command-line help:

```
usage: main.py [-h] [--build-options BUILD_OPTIONS] [--clean]
               [--report]
               [{build,test} [{build,test} ...]]

Build and/or run LVGL tests.

positional arguments:
  {build,test}          build: compile build tests, test:
                        compile/run executable tests.

optional arguments:
  -h, --help            show this help message and exit
  --build-options BUILD_OPTIONS
                        the build option name to build or run.
                        When omitted all build configurations
                        are used.
  --clean               clean existing build artifacts before
                        operation.
  --report              generate code coverage report for
                        tests.

This program builds and optionally runs the LVGL test programs.
There are two types of LVGL tests: "build", and "test". The
build-only tests, as their name suggests, only verify that the
program successfully compiles and links (with various build
options). There are also a set of tests that execute to verify
correct LVGL library behavior.
```


embeddedt (approved):

kisvegabor (approved):
Looks good, thanks you!

embeddedt:
The one downside to this change is that you now get over-notified in some cases about build errors:

screenshot

However, I don't think there is an easy way to fix this.

cmumford:
Yeah, I see your point. Maybe a feature to some, but a bug to others. Am I correct, that you still get a single email notification for a failed PR regardless of the number of failures?

embeddedt:
Yes; I believe that is correct.

PR #2529, opened 2021-09-05
Add lvgl interface target for micropython

eudoxos:
The CMake interface target for micropython is now working. Non-micropython builds are not affected.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -68,10 +68,25 @@ endif() # CONFIG_LVGL
 
 else()
 
-file(GLOB_RECURSE SOURCES ${CMAKE_CURRENT_LIST_DIR}/src/*.c)
-add_library(lvgl STATIC ${SOURCES})
+    file(GLOB_RECURSE SOURCES ${CMAKE_CURRENT_LIST_DIR}/src/*.c)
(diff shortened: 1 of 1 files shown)


amirgon:
So this assumes Micropython-ESP32 migrates to usermod right?
In current state this change would break the ESP32 port.

eudoxos:
> So this assumes Micropython-ESP32 migrates to usermod right?

Quite on the contrary, EPS32 and Zephyr should not be affected in the slightest.

I wanted to add indentation to `CMakeLists.txt` to make that clear but found it too intrusive. It is like this:

```cmake
if(ESP_PLATFORM)
# ... [unintended]
elseif(ZEPHYR_BASE)
# ... [unintended]
else()
    # [here is where my changes start]
    # ... 
    if(MICROPY_DIR)
        # ...
    else(MICROPY_DIR)
        # ...
    endif(MICROPY_DIR)
endif()
```

amirgon:
> Quite on the contrary, EPS32 and Zephyr should not be affected in the slightest.

Got it. Thanks for the clarification!

kisvegabor:
I can't judge if these changes are correct, so please let me know if it's ready to merge.

amirgon:
> I can't judge if these changes are correct, so please let me know if it's ready to merge.

Looks good to me, although I can't test this since it's RPi specific change.

kisvegabor:
Thank you, merging.

PR #2532, opened 2021-09-06
attempt at a better LV_PRId32

eudoxos:
This should fix ESP32 regression with #2512 (`LV_PRId32`) the proper way. Ports without `inttypes.h` (hopefully only a few, added `__XC16__` as an example) must be manually handled, for the rest, use platform's `inttypes.h`.

### DIFF
--- a/src/misc/lv_printf.h
+++ b/src/misc/lv_printf.h
@@ -34,13 +34,13 @@
 #ifndef _LV_PRINTF_H_
 #define _LV_PRINTF_H_
 
-/* ports without C99's inttypes.h */
-#ifdef __XC16__
-   #define LV_PRId32 "d"
-#else
-   /* if you get an error here, handle your port above */
+#if __has_include(<inttypes.h>)
    #include<inttypes.h>
+   /* platform-specific printf format for int32_t, usually "d" or "ld" */
    #define LV_PRId32 PRId32
+#else
+   /* hope this is correct for ports without inttypes.h */
+   #define LV_PRId32 "d"
 #endif
 
 #ifdef __cplusplus
--- a/src/misc/lv_printf.h
+++ b/src/misc/lv_printf.h
@@ -34,13 +34,13 @@
 #ifndef _LV_PRINTF_H_
 #define _LV_PRINTF_H_
 
-/* not all compilers provide inttypes.h which would define these */
-#include<limits.h>
-
-#if LONG_MAX==2147483647
-   #define LV_PRId32 "ld"
-#else
+/* ports without C99's inttypes.h */
+#ifdef __XC16__
    #define LV_PRId32 "d"
+#else
+   /* if you get an error here, handle your port above */
+   #include<inttypes.h>
+   #define LV_PRId32 PRId32
 #endif
 
 #ifdef __cplusplus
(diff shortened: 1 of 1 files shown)


amirgon:
What is `__XC16__` specifically? 
How do I identify the right macro for each platform?

kisvegabor:
> What is __XC16__ specifically?

It's for Microchip's compiler (XC16) that is used for 16 bit MCUs.  These MCUs are targeting graphics too so we should support them.

@eudoxos 
> Branching by INT_MAX did not work as I had thought.

What was the problem with it?

amirgon:
> It's for Microchip's compiler (XC16) that is used for 16 bit MCUs.

@eudoxos I don't get it. How does this PR solve the ESP32 problem? ESP32 uses Xtensa not Microchip.

eudoxos:
> What was the problem with it?

IIRC... it was the first attempt like this:

```c
#if INT_MAX==32767
   #define LV_PRId32 "d"
#elif INT_MAX==2147483647
   #define LV_PRId32 "ld"
#else
   #error Unrecognized width of int type (neither 16-bit nor 32-bit).
#endif
```
and it was failing in the CI for some platforms (don't remember the details, sorry); and then I realized `PRId32` does not specify format for `int` but it tries to guess whether `int32_t` is an `int` (`%d`) or `long int` (`%ld`) for that platform. So then I was checking LONG_MAX, whether it is 32-bit, then it should be ok to use `ld` for `int32_t`, otherwise use `d`, but it does not work for ESP32. I don't know where the problem in the logic is, the topic is quite arcane to me.

> @eudoxos I don't get it. How does this PR solve the ESP32 problem? ESP32 uses Xtensa not Microchip.

"Normal" (C99-compliant) platforms provide `inttypes.h`, that solves the ESP32 problem. See e.g. [esp-idf's test_attr.c](https://github.com/espressif/esp-idf/blob/606557b48bd195e400c2e3928d8d7c7d4ed9b050/components/esp_common/test/test_attr.c#L1) which uses that file. So you are fine.

> What is __XC16__ specifically?
> How do I identify the right macro for each platform?

Each platform should document that, I googled that one out.

Another possibility would be to add `CHECK_INCLUDE_FILE(<inttypes.h> LV_HAVE_INTTYPES_H)` to `CMakeLists.txt` but it would be the first feature-check there, and would not handle `Makefile`-based ports (if any, I don't know but `Makefile` is there).

kisvegabor:
I'd prefer avoiding platform specific code in lvgl but if there is no other option I'm ok with the `__XC16__` check.

embeddedt:
Would [`__has_include`](https://github.com/lvgl/lvgl/blob/d6ff681853fc8e28992f9213391fb67346681a82/src/lv_conf_internal.h#L22-L23) be more reliable than checking for `__XC16__`?

eudoxos:
> Would [`__has_include`](https://github.com/lvgl/lvgl/blob/d6ff681853fc8e28992f9213391fb67346681a82/src/lv_conf_internal.h#L22-L23) be more reliable than checking for `__XC16__`?

That would be completely awesome, of course. I did not think it was available (in c++, it only appeared in c++17), but if you use it already, then we are good to go.

amirgon:
This fixes ESP32 build.  
@kisvegabor / @embeddedt - could you merge this?

Related: https://github.com/lvgl/lv_micropython/pull/43#issuecomment-913232965

embeddedt:
Merged! Thank you.

PR #2536, opened 2021-09-06
fix(zoom) multiplication overflow with zoom on 16-bit platforms

kevinpt:

lv_coord_t is typically an int16_t. On 32-bit and larger platforms this will be promoted to int (32-bits wide)
before any operations are performed and then there is enough precision to prevent any numeric range
issues. However, on a 16-bit platform int16_t is (generally) the same rank as int and no promotion will
happen. This creates an overflow risk with multiplications. Zoom factors are encoded as a fixed point
number so the risk is even greater.

Consider the case where both a style transform zoom property is set and an image zoom is set. If both
are 2.0 they will be multiplied for a 4.0 overall zoom factor. In Q8.8 fixed point this is:
   (512 * 512) >> 8 =  (262144) >> 8 = 1024 == 4.0
On a 16-bit platform the intermediate multiply overflows before the final shift adjustment. Even if one
property is left at the default LV_IMG_ZOOM_NONE (256 == 1.0) an overflow will still happen. 

This affects both images and the lv_chart widget. I cleaned up all the muls I could find related to zoom
by casting to int32_t. For lv_chart the expressions are dropping the fractional part and a round up by
adding 0.5 would be more correct but probably won't make any visual improvement so I left it as is.

### Checkpoints

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -28,6 +28,7 @@
 - fix(color) Bad cast in lv_color_mix() caused UB with 16bpp or less
 - fix(examples) don't compile assets unless needed
 - docs(all) Proofread, fix typos and add clarifications in confusing areas
+- fix(zoom) multiplication overflow with zoom calculations on 16-bit platforms
 
 ## v8.0.2 (16.07.2021)
 - fix(theme) improve button focus of keyboard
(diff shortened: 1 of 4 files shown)


embeddedt (approved):
We really need a way of running CI tests on a 16-bit architecture.

kisvegabor:
> We really need a way of running CI tests on a 16-bit architecture.

It'd be really great. Maybe do you already have some ideas? :slightly_smiling_face:

PR #2537, opened 2021-09-07
feat(btnmatrix/keyboard): add option to show popovers on button press

Johennes:
  
This adds a new option that, when enabled, shows popovers when pressing buttons, similar to how the system keyboards on Android and iOS behave.

Here is a short video of me using the feature in a framebuffer application on my laptop:

https://user-images.githubusercontent.com/1137962/132401580-5b0bd0b2-80de-4a89-beb4-7950a4fe0f03.mp4

### Checkpoints

I haven't updated the documentation yet because I first wanted to get feedback on the change.

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -34,6 +34,7 @@
 - feat(msgbox): add function to get selected button index
 - fix(btnmatrix): make ORed values work correctly with lv_btnmatrix_has_btn_ctrl
 - fix(snapshot): snapshot is affected by parent's style because of wrong coordinates.
+- feat(btnmatrix/keyboard): add option to show popovers on button press
 
 ## v8.0.2 (16.07.2021)
 - fix(theme) improve button focus of keyboard
--- a/docs/widgets/extra/keyboard.md
+++ b/docs/widgets/extra/keyboard.md
@@ -32,6 +32,12 @@ To set the mode manually, use `lv_keyboard_set_mode(kb, mode)`. The default mode
 You can assign a [Text area](/widgets/core/textarea) to the Keyboard to automatically put the clicked characters there.
 To assign the text area, use `lv_keyboard_set_textarea(kb, ta)`.
 
+### Key Popovers
+To enable key popovers on press, like on common Android and iOS keyboards, use `lv_keyboard_set_popovers(kb, true)`. The default control maps are preconfigured to only show the popovers on keys that produce a symbol and not on e.g. space. If you use a custom keymap, set the `LV_BTNMATRIX_CTRL_POPOVER` flag for all keys that you want to show a popover.
+
+Note that popovers for keys in the top row will draw outside the widget boundaries. To account for this, reserve extra free space on top of the keyboard or ensure that the keyboard is added _after_ any widgets adjacent to its top boundary so that the popovers can draw over those.
+
+The popovers currently are merely a visual effect and don't allow selecting additional characters such as accents yet.
 
 ### New Keymap
 You can specify a new map (layout) for the keyboard with `lv_keyboard_set_map(kb, map)` and `lv_keyboard_set_ctrl_map(kb, ctrl_map)`.
--- a/src/widgets/lv_btnmatrix.c
+++ b/src/widgets/lv_btnmatrix.c
@@ -43,6 +43,7 @@ static bool button_is_checked(lv_btnmatrix_ctrl_t ctrl_bits);
 static bool button_is_repeat_disabled(lv_btnmatrix_ctrl_t ctrl_bits);
 static bool button_is_inactive(lv_btnmatrix_ctrl_t ctrl_bits);
 static bool button_is_click_trig(lv_btnmatrix_ctrl_t ctrl_bits);
+static bool button_is_popover(lv_btnmatrix_ctrl_t ctrl_bits);
 static bool button_is_checkable(lv_btnmatrix_ctrl_t ctrl_bits);
 static bool button_is_recolor(lv_btnmatrix_ctrl_t ctrl_bits);
 static bool button_get_checked(lv_btnmatrix_ctrl_t ctrl_bits);
@@ -51,6 +52,7 @@ static uint16_t get_button_from_point(lv_obj_t * obj, lv_point_t * p);
 static void allocate_btn_areas_and_controls(const lv_obj_t * obj, const char ** map);
 static void invalidate_button_area(const lv_obj_t * obj, uint16_t btn_idx);
 static void make_one_button_checked(lv_obj_t * obj, uint16_t btn_idx);
+static bool has_popovers_in_top_row(lv_obj_t * obj);
 
 /**********************
  *  STATIC VARIABLES
@@ -111,15 +113,8 @@ void lv_btnmatrix_set_map(lv_obj_t * obj, const char * map[])
     lv_coord_t max_w            = lv_obj_get_content_width(obj);
     lv_coord_t max_h            = lv_obj_get_content_height(obj);
 
-    /*Count the lines to calculate button height*/
-    uint8_t row_cnt = 1;
-    uint32_t i;
-    for(i = 0; map[i] && map[i][0] != '\0'; i++) {
-        if(strcmp(map[i], "\n") == 0) row_cnt++;
-    }
-
     /*Calculate the position of each row*/
-    lv_coord_t max_h_no_gap = max_h - (prow * (row_cnt - 1));
+    lv_coord_t max_h_no_gap = max_h - (prow * (btnm->row_cnt - 1));
 
     /*Count the units and the buttons in a line
      *(A button can be 1,2,3... unit wide)*/
@@ -129,7 +124,7 @@ void lv_btnmatrix_set_map(lv_obj_t * obj, const char * map[])
 
     /*Count the units and the buttons in a line*/
     uint32_t row;
-    for(row = 0; row < row_cnt; row++) {
+    for(row = 0; row < btnm->row_cnt; row++) {
         uint16_t unit_cnt = 0;           /*Number of units in a row*/
         uint16_t btn_cnt = 0;            /*Number of buttons in a row*/
         /*Count the buttons and units in this row*/
@@ -144,8 +139,8 @@ void lv_btnmatrix_set_map(lv_obj_t * obj, const char * map[])
             continue;
         }
 
-        lv_coord_t row_y1 = ptop + (max_h_no_gap * row) / row_cnt + row * prow;
-        lv_coord_t row_y2 = ptop + (max_h_no_gap * (row + 1)) / row_cnt + row * prow - 1;
+        lv_coord_t row_y1 = ptop + (max_h_no_gap * row) / btnm->row_cnt + row * prow;
+        lv_coord_t row_y2 = ptop + (max_h_no_gap * (row + 1)) / btnm->row_cnt + row * prow - 1;
 
         /*Set the button size and positions*/
         lv_coord_t max_w_no_gap = max_w - (pcol * (btn_cnt - 1));
@@ -179,6 +174,10 @@ void lv_btnmatrix_set_map(lv_obj_t * obj, const char * map[])
 
         map_row = &map_row[btn_cnt + 1];       /*Set the map to the next line*/
     }
+
+    /*Popovers in the top row will draw outside the widget and the extended draw size depends on
+     *the row height which may have changed when setting the new map*/
+    lv_obj_refresh_ext_draw_size(obj);
 
     lv_obj_invalidate(obj);
 }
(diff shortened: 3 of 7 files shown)


docs/CHANGELOG.md
```diff
@@ -29,6 +29,7 @@
 - fix(examples) don't compile assets unless needed
 - docs(all) Proofread, fix typos and add clarifications in confusing areas
 - fix(zoom) multiplication overflow with zoom calculations on 16-bit platforms
+- feat(btnmatrix): add option to show popovers on button press
```
kisvegabor: Please also update the docs.

src/widgets/lv_btnmatrix.h
```diff
@@ -39,6 +39,7 @@ enum {
     LV_BTNMATRIX_CTRL_CHECKABLE  = 0x0040, /**< The button can be toggled.*/
     LV_BTNMATRIX_CTRL_CHECKED    = 0x0080, /**< Button is currently toggled (e.g. checked).*/
     LV_BTNMATRIX_CTRL_CLICK_TRIG = 0x0100, /**< 1: Send LV_EVENT_VALUE_CHANGE on CLICK, 0: Send LV_EVENT_VALUE_CHANGE on PRESS*/
+    LV_BTNMATRIX_CTRL_NO_POPOVER = 0x0200, /**< Do not show a popover when pressing this key*/
```
kisvegabor: The default should be non popover, so I'd change it to `LV_BTNMATRIX_CTRL_POPOVER`.

kisvegabor (commented):
Hi,

IMO this feature would be really useful if it allowed selecting between related characters. E.g. on pressing `a`, show `áä` in the popover. 
I know that it's much more difficult to implement.

Anyway, I'm ok with merging this as it's still a fancy effect.

src/widgets/lv_btnmatrix.h
Johennes: Hm, are you certain? Even though popovers are disabled by default, I set up the control flags on the default keymap so that popovers work out of the box on the relevant keys as soon as you do `lv_btnmatrix_set_popovers(keyboard, true)`.

Right now it looks like this:

```
static const char * const default_kb_map_lc[] = {"1#", "q", "w", "e", "r", "t", "y", "u", "i", "o", "p", LV_SYMBOL_BACKSPACE, "\n",
    ...
};

static const lv_btnmatrix_ctrl_t default_kb_ctrl_lc_map[] = {
    LV_KEYBOARD_CTRL_BTN_FLAGS | LV_BTNMATRIX_CTRL_NO_POPOVER | 5, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4,  
    ...
};
```

Only the few keys that should not pop up received the `LV_BTNMATRIX_CTRL_NO_POPOVER` and everything else stays as is.

When we flip the semantic of the flag, we'd have to change the map to e.g.

```
static const lv_btnmatrix_ctrl_t default_kb_ctrl_lc_map[] = {
    LV_KEYBOARD_CTRL_BTN_FLAGS | 5, LV_BTNMATRIX_CTRL_POPOVER | 4, LV_BTNMATRIX_CTRL_POPOVER | 4, LV_BTNMATRIX_CTRL_POPOVER | 4, LV_BTNMATRIX_CTRL_POPOVER | 4, LV_BTNMATRIX_CTRL_POPOVER | 4, LV_BTNMATRIX_CTRL_POPOVER | 4, LV_BTNMATRIX_CTRL_POPOVER | 4, LV_BTNMATRIX_CTRL_POPOVER | 4, LV_BTNMATRIX_CTRL_POPOVER | 4, LV_BTNMATRIX_CTRL_POPOVER | 4,  
    ...
};
```

That might make the default map look overly complicated.

Or did you want to totally disable popovers in the default map, meaning users should call `lv_btnmatrix_set_popovers(keyboard, true)` and also provide a custom control map with `LV_BTNMATRIX_CTRL_POPOVER` set wherever they want it?

Johennes:
> IMO this feature would be really useful if it allowed selecting between related characters. E.g. on pressing `a`, show `áä` in the popover.

Yeah, totally agreed, that'd be amazing. I might come back to that one day but the plain popovers already are a huge accessibility improvement for my current use case. :slightly_smiling_face:

src/widgets/lv_btnmatrix.h
Johennes: Hm, on the other hand, if we flip the logic to an `LV_BTNMATRIX_CTRL_POPOVER` option, that would make it easier to marry it with `LV_BTNMATRIX_CTRL_CLICK_TRIG`. I think keys that _do_ show a popover should always trigger the event on release because if the key is already sent while you look at the popover, then the popover isn't very helpful. So we could just say that setting `LV_BTNMATRIX_CTRL_POPOVER` implies `LV_BTNMATRIX_CTRL_CLICK_TRIG` being set as well.

Only issue, as said above, is that the default control map would get a bit length. :thinking:

Johennes: Ok, I updated the PR to use `LV_BTNMATRIX_CTRL_POPOVER`, made it imply `LV_BTNMATRIX_CTRL_CLICK_TRIG` and updated the default keyboard control maps.

Please let me know if this is ok and I'll update the documentation accordingly.

kisvegabor: > Hm, are you certain? 

Good question! :smile: It seems `LV_BTNMATRIX_CTRL_NO_POPOVER` actually means if the certain button is "popover compatible" or not. And if we don't want popovers by default we can set `lv_btnmatrix_set_popovers(btnm ,false)`.

But somehow this approach is still strange to me. Let's talk about just a plain btnmatrix. When the user sets the control map he can decide whether he wants popover on a certain button or not. So there is no need for `lv_btnmatrix_set_popovers()`. 

It's less elegant but it looks more logical to me have an `lv_keyboard_set_popover(kb, enable)` save the `enable` flag in the keyboard, and in `lv_keyboard_set_mode` add  `LV_BTNMATRIX_CTRL_POPOVER` flag to the relevant buttons if the popover is enabled. Maybe rather with `lv_btnmatrix_set_ctrl_all` to add popover to all buttons, and remove from a few. 

What do you think?

Johennes: > Let's talk about just a plain btnmatrix. When the user sets the control map he can decide whether he wants popover on a certain button or not. So there is no need for lv_btnmatrix_set_popovers().

Ah, I think I see what you're saying. The presence (or absence) of ` LV_BTNMATRIX_CTRL_POPOVER` in the control map already says that popovers are enabled (or disabled) so there is no need for a separate `lv_btnmatrix_set_popovers(btnm, true / false)` call.

Instead, we could just derive the value of `btnm->popovers` by looping over the control map and checking if at least one button has `LV_BTNMATRIX_CTRL_POPOVER` set.

> It's less elegant but it looks more logical to me have an lv_keyboard_set_popover(kb, enable) save the enable flag in the keyboard, and in lv_keyboard_set_mode add LV_BTNMATRIX_CTRL_POPOVER flag to the relevant buttons if the popover is enabled. Maybe rather with lv_btnmatrix_set_ctrl_all to add popover to all buttons, and remove from a few.

Hm, I am not sure if that would work generically. The caller could have set a custom map and control map via `lv_keyboard_set_map`. I'm actually doing that in the apps I'm working on to switch between e.g. a French and a German keyboard layout. So with a custom map, when `lv_keyboard_set_mode` is called, I believe, we wouldn't have a way to know what buttons to automatically add the popover flag to.

We could maybe tell if a custom map was set or not by checking if `kb_map[mode]` still points to the default map. Did you mean that?

Alternatively, I think what would work generically is if we inverted the logic you suggest:

1. Add `LV_BTNMATRIX_CTRL_POPOVER` in the default control maps (like currently done in the PR)
2. In `lv_keyboard_set_mode` *remove* the popover flag from all buttons if popovers were previously disabled by calling `lv_keyboard_set_popover(kb, false)`.

That would still have the disadvantage that the default control maps are bloated up a bit by the various `LV_BTNMATRIX_CTRL_POPOVER` flags but maybe that's not too bad.

Let me know what you think and thanks a ton for taking the time to talk this through. :slightly_smiling_face:

kisvegabor: > Alternatively, I think what would work generically is if we inverted the logic you suggest:

You are right, and it sounds good! :slightly_smiling_face: 
The bloating could be slightly avoided by adding a helper macro like 
```c
#define LV_KB_BTN(width) LV_BTNMATRIX_CTRL_POPOVER | width
```

Johennes: >  The bloating could be slightly avoided by adding a helper macro [...]

Oh, great idea!

Ok, I updated the PR accordingly. However, one thing I noticed is that without a dedicated `btnm->popovers` setting, we now have to recompute the extended draw size any time the control map is changed and that means looping over all buttons in the top row and checking if any of them has the `LV_BTNMATRIX_CTRL_POPOVER`.

This is aggravated by the fact that e.g. `lv_btnmatrix_clear_btn_ctrl_all` calls `lv_btnmatrix_clear_btn_ctrl` in a loop and since both functions are public we have to recalculate the extended draw size in both which eventually means we might be doing it quite often.

I'm a bit divided but unless there's a good way to optimize this, I think it might be better to have the `popovers` setting in the button matrix rather than in the keyboard. As you noted before, there will be some duality because users have to set both `LV_BTNMATRIX_CTRL_POPOVER` _and_ `btnm->popovers` but at least this way we'd know exactly when to recompute the extended draw size.

src/widgets/lv_btnmatrix.c
```diff
@@ -232,6 +235,10 @@ void lv_btnmatrix_clear_btn_ctrl(const lv_obj_t * obj, uint16_t btn_id, lv_btnma
 
     btnm->ctrl_bits[btn_id] &= (~ctrl);
     invalidate_button_area(obj, btn_id);
+
+    if (ctrl & LV_BTNMATRIX_CTRL_POPOVER) {
+        lv_obj_refresh_ext_draw_size(obj);
```
embeddedt, line 240: CI doesn't like this, because `obj` is `const lv_obj_t`, but the API expects `lv_obj_t`.

@kisvegabor What is the purpose of `const` on this API, since it modifies the control bits?

Miguel0101:
> Hi,
> 
> IMO this feature would be really useful if it allowed selecting between related characters. E.g. on pressing `a`, show `áä` in the popover.
> I know that it's much more difficult to implement.
> 
> Anyway, I'm ok with merging this as it's still a fancy effect.

I agree.

src/widgets/lv_btnmatrix.h
kisvegabor: > However, one thing I noticed is that without a dedicated btnm->popovers setting, we now have to recompute the extended draw size any time the control map is changed and that means looping over all buttons in the top row and checking if any of them has the LV_BTNMATRIX_CTRL_POPOVER.

IMO it's acceptable to re-implement a custom `clear_all` method in the keyboard for this specific case.

src/widgets/lv_btnmatrix.c
kisvegabor: > What is the purpose of const on this API, since it modifies the control bits?

It seems like a mistake. `lv_btnmatrix_set_btn_ctrl` has non `const` parameter and `lv_btnmatrix_clear_btn_ctrl` should have too.

Johennes: Good catch. I removed the `const` as part of this PR now.

docs/CHANGELOG.md
Johennes: Just did. :slightly_smiling_face:

src/widgets/lv_btnmatrix.h
Johennes: Ok, I tried that but then decided to go a slightly different road. Instead in `lv_keyboard.c` I now copy the control map, remove the `LV_BTNMATRIX_CTRL_POPOVER` on the copied map and finally hand it down to the button matrix. The benefits of this are:

1. The keyboard doesn't need to know about extended draw size recalculations or button invalidations
2. The control map is only set a single time so the extended draw size calculation also only happens once

The drawback is that we have to make a temporary copy of the control map but I think that should be cheap because it's just an integer array.

Let me know what you think. I also updated the docs so everything should be in place now.

docs/widgets/extra/keyboard.md
```diff
@@ -32,6 +32,10 @@ To set the mode manually, use `lv_keyboard_set_mode(kb, mode)`. The default mode
 You can assign a [Text area](/widgets/core/textarea) to the Keyboard to automatically put the clicked characters there.
 To assign the text area, use `lv_keyboard_set_textarea(kb, ta)`.
 
+### Key Popovers
+To enable key popovers on press, like on common Android and iOS keyboards, use `lv_keyboard_set_popovers(kb, true)`. The default control maps are preconfigured to only show the popovers on keys that produce a symbol and not on e.g. space. If you use a custom keymap, set the `LV_BTNMATRIX_CTRL_POPOVER` flag for all keys that you want to show a popover.
+
```
embeddedt, line 37: (I would just add this myself but I don't want to disrupt your force-push.)

I think we should add a note that the selectable extra characters (e.g. accents) are currently not implemented, and this is just a visual effect, to preempt the question being asked on the forum.

Johennes: Good idea. I added a corresponding phrase.

Question for the future: are force-pushes ok or would you rather have me add further commits for follow-up changes?

embeddedt:
It doesn't matter, as we squash PRs on merge anyways.

kisvegabor:
It looks good to me. Thank you, Johannes! :slightly_smiling_face:

PR #2539, opened 2021-09-08
feat(msgbox): omit title label unless needed

Johennes:

Prior to this commit, when the title string was empty and the close button disabled, an extra empty line showed at the top of the message box. This commit prevents adding the title label unless it has content or is needed as a spacer for the close button.

As a positive side effect, this also prevents the default "text" from displaying when `NULL` is passed as the title.

### Checkpoints
- [x] Update the documentation (no update needed, I believe)


kisvegabor:
Sounds reasonable, thank you!

PR #2543, opened 2021-09-09
fix(arc) format code

FASTSHIFT:

Fix code indentation of `lv_arc`.

### Checkpoints

### DIFF
--- a/src/widgets/lv_arc.c
+++ b/src/widgets/lv_arc.c
@@ -234,13 +234,13 @@ void lv_arc_set_mode(lv_obj_t * obj, lv_arc_mode_t type)
     LV_ASSERT_OBJ(obj, MY_CLASS);
     lv_arc_t * arc = (lv_arc_t *)obj;
 
-    int16_t val =arc->value;
+    int16_t val = arc->value;
 
     arc->type = type;
     arc->value = -1; /** Force set_value handling*/
 
-    int16_t bg_midpoint, bg_end =arc->bg_angle_end;
-    if(arc->bg_angle_end <arc->bg_angle_start) bg_end =arc->bg_angle_end + 360;
+    int16_t bg_midpoint, bg_end = arc->bg_angle_end;
+    if(arc->bg_angle_end < arc->bg_angle_start) bg_end = arc->bg_angle_end + 360;
 
     switch(arc->type) {
         case LV_ARC_MODE_SYMMETRICAL:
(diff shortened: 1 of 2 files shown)


FASTSHIFT:
@kisvegabor 
I tried to run `scripts/code-format.sh`. It seems that many files have been changed. Do I need to make uniform changes?

kisvegabor:
> I tried to run scripts/code-format.sh. It seems that many files have been changed. Do I need to make uniform changes?

It wasn't run for a while. Maybe we should have a CI action to run PRs and merges to master automatically and push the changes.
@embeddedt What do you think?

embeddedt:
As far as I know, the standard way of handling this is to have a linting check which rejects new PRs/commits if the formatting is done incorrectly. Usually this can be done by running the formatter on the changed files and comparing the outputs to the original files. If the formatting is correct, the two files should be same.

I'll see if I can set something like this up in the coming days.

kisvegabor:
At first look automatic formatting seems easier to use. I'm afraid rejecting PRs would make contributions more difficult.

embeddedt:
There are two potential issues with automatic formatting:

* The formatter has to be capable of not reformatting its own output. I haven't seen it happen that often, but there are times where I've run a formatter and it keeps finding issues with the file every time.
* It creates extra commits in the Git history just for formatting, instead of the existing commits already being properly formatted.

Another option could be to try and have @lvgl-bot push formatting commits to PR branches, which would alleviate the second issue, since we squash PRs when merging anyways.

kisvegabor:
If it's rejected the PRs could it create a meaningful output (or patch) to show the required changes?

embeddedt:
I haven't checked, but I think that should be possible. You are referring to markers like [this](https://github.com/lvgl/lvgl/pull/2401/files#diff-ee48b60c5e9b9e7fb56d45e6f758d69fcc9c6e9b2c0b83843043267c54dba7ccR138), right?

kisvegabor:
I meant only a `format.patch` to show the problematic lines and simply can be downloaded and applied. 

The marker you have shown would be also great but I think not mandatory.  But there should be a simple way to fix all formatting issues at once instead of error and trial while the CI accepts it.

embeddedt:
I haven't seen any project on GitHub pull this off before, but what if the bot posted the diff as a review comment with GitHub's suggested changes feature? In that case, the patch could be applied without even needing to download it.

kisvegabor:
It's a good idea, but if there are many formatting issues it could be cumbersome to accept the suggestions one by one. 
Anyway, I think it could be very well usable in the vast majority of the cases.

FASTSHIFT:
@kisvegabor  
Does this PR still need to be merged?

kisvegabor:
@embeddedt Shall we open a new issue for the code formatter CI?

embeddedt:
Yes; I'm somewhat busy at the moment, but I do intend to set something up when I am able. In the meantime, I suppose we could just run `scripts/code-format.sh` on `master` manually after merging PRs.

kisvegabor:
> but I do intend to set something up when I am able.

Great! 

I've just run `code-format.sh` and pushed it. Let's do it manually more frequently than in the previous months :slightly_smiling_face:

PR #2549, opened 2021-09-11
fix(lv_textarea.c) Fix crash while delete non-ascii character in pwd …

lmdu-gs:

Steps to reproduce：
* textarea set to password mode.
* input a non-ascii(>= 2 bytes) character.
* delete it.

Reason:
Call _lv_txt_cut with byte length not character length.

### Checkpoints

### DIFF
--- a/src/widgets/lv_textarea.c
+++ b/src/widgets/lv_textarea.c
@@ -259,8 +259,7 @@ void lv_textarea_del_char(lv_obj_t * obj)
     }
 
     if(ta->pwd_mode != 0) {
-        uint32_t byte_pos = _lv_txt_encoded_get_byte_id(ta->pwd_tmp, ta->cursor.pos - 1);
-        _lv_txt_cut(ta->pwd_tmp, ta->cursor.pos - 1, _lv_txt_encoded_size(&ta->pwd_tmp[byte_pos]));
+        _lv_txt_cut(ta->pwd_tmp, ta->cursor.pos - 1, 1);
 
         ta->pwd_tmp = lv_mem_realloc(ta->pwd_tmp, strlen(ta->pwd_tmp) + 1);
         LV_ASSERT_MALLOC(ta->pwd_tmp);


kisvegabor:
The original code looks good to me. If the password is replaced by bullets one character in not one byte so the `_lv_txt_encoded_*` functions needs to be used. 

Anyway, I've tested it and it worked well for me. Could you send a code snippet to reproduce the issue?


FYI, while I was testing this issue I've pushed a fix to a probably unrelated issue. 192419e7b

lmdu-gs:
I've tested this with project lv_sim_vscode_sdl, with modification:
```
--- a/src/lv_demo_widgets/lv_demo_widgets.c
+++ b/src/lv_demo_widgets/lv_demo_widgets.c
@@ -268,6 +268,12 @@ static void profile_create(lv_obj_t * parent)
     lv_textarea_set_one_line(password, true);
     lv_textarea_set_password_mode(password, true);
     lv_textarea_set_placeholder_text(password, "Min. 8 chars.");
+    lv_textarea_set_password_show_time(password, 0);
+    lv_textarea_set_text(password, "жa");
+    LV_LOG_WARN("text area: %s/%s", lv_label_get_text(lv_textarea_get_label(password)), lv_textarea_get_text(password));
+    lv_textarea_set_cursor_pos(password, 1);
+    lv_textarea_del_char(password);
+    LV_LOG_WARN("text area: %s/%s", lv_label_get_text(lv_textarea_get_label(password)), lv_textarea_get_text(password));
     lv_obj_add_event_cb(password, ta_event_cb, LV_EVENT_ALL, kb);
```
The output is:
```
[Warn]  (0.025, +25)     profile_create: text area: ••/жa       (in lv_demo_widgets.c line #273)
[Warn]  (0.040, +15)     profile_create: text area: •/  (in lv_demo_widgets.c line #276)
```
**`lv_textarea_del_char` cut more than one character.**

Why this happen? I watched the code of function `_lv_txt_cut`:
```
void _lv_txt_cut(char * txt, uint32_t pos, uint32_t len)
{

    size_t old_len = strlen(txt);

    pos = _lv_txt_encoded_get_byte_id(txt, pos); /*Convert to byte index instead of letter index*/
    len = _lv_txt_encoded_get_byte_id(&txt[pos], len);

    /*Copy the second part into the end to make place to text to insert*/
    uint32_t i;
    for(i = pos; i <= old_len - len; i++) {
        txt[i] = txt[i + len];
    }
}
```
I think the parameter `pos` and `len` are both number of characters, not number of bytes. So if call it with character pos and byte length, the `len` would be calc twice.

Why not crash? I've noticed that I tested this on lvgl v7.10.1 before. The function _lv_txt_encoded_get_byte_id(lv_txt_utf8_get_byte_id) is:
```
static uint32_t lv_txt_utf8_get_byte_id(const char * txt, uint32_t utf8_id)
{
    uint32_t i;
    uint32_t byte_cnt = 0;
    for(i = 0; i < utf8_id; i++) {
        uint8_t c_size = _lv_txt_encoded_size(&txt[byte_cnt]);
        byte_cnt += c_size > 0 ? c_size : 1;
    }

    return byte_cnt;
}
```
And on v7.11.0, it is:
```
static uint32_t lv_txt_utf8_get_byte_id(const char * txt, uint32_t utf8_id)
{
    uint32_t i;
    uint32_t byte_cnt = 0;
    for(i = 0; i < utf8_id && txt[byte_cnt] != '\0'; i++) {
        uint8_t c_size = _lv_txt_encoded_size(&txt[byte_cnt]);
        byte_cnt += c_size > 0 ? c_size : 1;
    }

    return byte_cnt;
}
```

kisvegabor:
Ah, got it. And you are right. 
This fix is required indeed. 

Merging, thank you!

Issue #2522, opened 2021-09-02
Bug with LV_ARC_MODE_REVERSE


jadonmmiller:
### Describe the bug
It seems when I use LV_ARC_MODE_REVERSE with lv_arc_set_bg_angles and lv_arc_set_value the arc gets drawn erratic ways. It starts the indicator in the middle of the background and draws it off the background.

If I drag the arc handle to the left of its default position, the indicator flips around full circle and draws off the background (see the last screenshot).

Check out the [forum post](https://forum.lvgl.io/t/bug-with-lv-arc-mode-reverse/6655) (so far no responses).

### To Reproduce
```c
        static lv_obj_t* arcBlack;
        arcBlack = lv_arc_create(lv_scr_act());
        lv_arc_set_mode(arcBlack, LV_ARC_MODE_REVERSE);
        lv_arc_set_bg_angles(arcBlack, 0, 90);
        lv_arc_set_value(arcBlack, 40);
```

### Expected behavior
An arc exactly like the non-reversed image below (the first one) but with the indicator on the opposite side of the background.

### Screenshots or video
The code above without the reverse line:
![Working without reverse mode](https://forum.lvgl.io/uploads/default/original/2X/6/64bbbe87758b5948e8a90bcb0b78190caa803d2f.png)

The unmodified arc in reverse mode:
![The arc in reverse mode](https://forum.lvgl.io/uploads/default/original/2X/e/e81ea18a472c3f3ea70c8e80845ee4de2afa4368.png)

The reversed arc above, but dragged backwards:
![Reversed and dragged backwards](https://forum.lvgl.io/uploads/default/original/2X/4/4d393d281e4da78bbc02a31b206a19df8b51affc.png)

kisvegabor:
Thank you for the report. I've pushed a fix.

PR #2551, opened 2021-09-13
Add initial test for checkbox

C47D:

Add initial set of tests for checkbox widget, I choose this widget because it's one of the simplest (?)

@kisvegabor What else should we test? If the checkbox state is `LV_STATE_DISABLED` and we click on it, should it call the `event_handler`?

Also, I wasn't able to use the Unity `setup` and `tearDown` functions, and is it possible to debug the `test_checkbox` object file? I was trying to debug using gdb but couldn't find the generated executables.

### Checkpoints

### DIFF
--- a/tests/src/test_cases/test_bugfix.c
+++ b/tests/src/test_cases/test_bugfix.c
@@ -1,0 +1,29 @@
+#if LV_BUILD_TEST
+#include "../lvgl.h"
+
+#include "unity/unity.h"
+
+void test_bugfix_for_2522(void);
+
+void test_bugfix_for_2522(void)
+{
+    uint16_t expected_start_angle = 36;
+    uint16_t expected_end_angle = 90;
+    int16_t expected_value = 40;
+
+    lv_obj_t *arcBlack;
+    arcBlack = lv_arc_create(lv_scr_act());
+
+    lv_arc_set_mode(arcBlack, LV_ARC_MODE_REVERSE);
+
+    lv_arc_set_bg_angles(arcBlack, 0, 90);
+    
+    /* lv_arc_set_value calls value_update, where the fix is done */
+    lv_arc_set_value(arcBlack, expected_value);
+
+    TEST_ASSERT_EQUAL_UINT16(expected_start_angle, lv_arc_get_angle_start(arcBlack));
+    TEST_ASSERT_EQUAL_UINT16(expected_end_angle, lv_arc_get_angle_end(arcBlack));
+    TEST_ASSERT_EQUAL_INT16(expected_value, lv_arc_get_value(arcBlack));
+}
+
+#endif
--- a/tests/src/test_cases/test_checkbox.c
+++ b/tests/src/test_cases/test_checkbox.c
@@ -8,6 +8,7 @@
 void test_checkbox_creation_successfull(void);
 void test_checkbox_should_call_event_handler_on_click_when_enabled(void);
 void test_checkbox_should_have_default_text_when_created(void);
+void test_checkbox_should_return_dinamically_allocated_text(void);
 
 static lv_obj_t *active_screen = NULL;
 static lv_obj_t *checkbox = NULL;
@@ -54,6 +55,19 @@ void test_checkbox_should_have_default_text_when_created(void)
     checkbox = lv_checkbox_create(active_screen);
 
     TEST_ASSERT_EQUAL_STRING(default_text, lv_checkbox_get_text(checkbox));
+    TEST_ASSERT_NOT_NULL(lv_checkbox_get_text(checkbox));
+}
+
+void test_checkbox_should_return_dinamically_allocated_text(void)
+{
+    const char *message = "Hello World!";
+
+    active_screen = lv_scr_act();
+    checkbox = lv_checkbox_create(active_screen);
+
+    lv_checkbox_set_text(checkbox, message);
+
+    TEST_ASSERT_EQUAL_STRING(message, lv_checkbox_get_text(checkbox));
     TEST_ASSERT_NOT_NULL(lv_checkbox_get_text(checkbox));
 }
 
--- a/tests/src/test_cases/test_checkbox.c
+++ b/tests/src/test_cases/test_checkbox.c
@@ -1,0 +1,48 @@
+#if LV_BUILD_TEST
+#include "../lvgl.h"
+
+#include "unity/unity.h"
+
+#include "lv_test_indev.h"
+
+void test_checkbox_creation_successfull(void);
+void test_checkbox_should_call_event_handler_on_click_when_enabled(void);
+
+static lv_obj_t *active_screen = NULL;
+static lv_obj_t *checkbox = NULL;
+
+static volatile bool event_called = false;
+
+static void event_handler(lv_event_t *e)
+{
+    lv_event_code_t code = lv_event_get_code(e);
+
+    if (LV_EVENT_VALUE_CHANGED == code) {
+        event_called = true;
+    }
+}
+
+void test_checkbox_creation_successfull(void)
+{
+    active_screen = lv_scr_act();
+    checkbox = lv_checkbox_create(active_screen);
+
+    TEST_ASSERT_NOT_NULL(checkbox);
+}
+
+void test_checkbox_should_call_event_handler_on_click_when_enabled(void)
+{
+    active_screen = lv_scr_act();
+    checkbox = lv_checkbox_create(active_screen);
+    
+    lv_obj_add_state(checkbox, LV_STATE_CHECKED);
+    lv_obj_add_event_cb(checkbox, event_handler, LV_EVENT_ALL, NULL);
+
+    lv_test_mouse_click_at(checkbox->coords.x1, checkbox->coords.y1);
+
+    TEST_ASSERT_TRUE(event_called);
+    
+    event_called = false;
+}
+
+#endif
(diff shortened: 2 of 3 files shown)


kisvegabor:
I've read the article you have [linked](https://github.com/lvgl/lvgl/issues/2337#issuecomment-917057222) and I like the long descriptive names. If the test fails anyone sees from the log what could go wrong. With `test_xxx_2` one needs to look at the asserts. Using long names also implies that the test should be very specific and have many short test cases instead of a few complex ones. It's also a good direction IMO.

> What else should we test?

The `lv_checkbox_set/get_text` function also can be tested. `lv_mem_monitor` can be used here to detect memory leaks especially when `lv_checkbox_set_text_static` is also used.

The rest is rather some rendering tests. It's hard to test with `get`  functions whether the checkbox used the padding correctly, or positioned the text correctly to the check indicator.

> Also, I wasn't able to use the Unity setup and tearDown functions

@cmumford replaced Unity with `ctest` (part of CMake) but to be honest I haven't immersed in the details of the current underlying logic. :sweat_smile: 


> and is it possible to debug the test_checkbox object file? 

Good question. The test are organized and built by `main.py` and the `CMakeList.txt` file. I don't know how complicated would it be to make possible to build the tests one by one to get an executable that can be debugged with gdb.

embeddedt:
> lv_mem_monitor can be used here to detect memory leaks especially when lv_checkbox_set_text_static is also used.

It would probably be cleanest to have a common assertion function for the tests that can be used to check whether the memory usage grew by more than an expected amount.

kisvegabor:
E.g. `TEST_MEMORY_USAGE(500)` where 500 is the max expected memory usage?

cmumford:
> @cmumford replaced Unity with ctest (part of CMake) but to be honest I haven't immersed in the details of the current underlying logic. sweat_smile

**no**, we're still using Unity. I'm only using ctest to execute the Unity based tests and to report their results. I did question the use of Unity, as there are other more widely used testing systems (like googletest), but Unity is designed to run on the MCU. We don't currently do this, but I'm guessing the goal is to do so at some point.

The generated executables are inside the `build` folder. I'm going by memory, but something like `build/test_dropdown`, etc. There used to be only one build executable which was overwritten for the next test, but that was changed to allow for parallel builds.

cmumford:
OK. The tests are all built in `tests/build_test/`, and are named the same as the test file basename. So `tests/src/test_cases/test_font_loader.c` compiles into `tests/build_test/test_font_loader`. Windows cmake puts them into a subdirectory. I think it's something like `tests/build_test/Debug/test_font_loader`.

I think that adding an option to `tests/main.py` to build a single test, or maybe print out the paths to the compiled tests might be a good idea. If you want this changed or enhanced at all feel free to create an issue and assign it to me.

C47D:
> Windows cmake puts them into a subdirectory. I think it's something like tests/build_test/Debug/test_font_loader.

Thanks for the tip, I'm using WSL Ubuntu 20, will try to find those files later today.

I think using Unity is better than nothing, I personally have been using CPPUTest and their mocks, it's quite easy to use, we can continue using Unity until we hit a huge blocker.

> There used to be only one build executable which was overwritten for the next test, but that was changed to allow for parallel builds.

At the beginning I was a bit confused by parallel builds 😄 

> no, we're still using Unity.

Did you tried using the `setup` and `tearDown` functions? They come handy when doing the same at the beginning and end of all the tests. Maybe I was doing something wrong but I wasn't able to use them, a segfault was thrown when running this particular test executable, that's why I was looking for it in order to debug it.

C47D:
I had to apply the following patch to allow me build the tests with symbols and debug the individual test (test_checkbox in my case), @cmumford could we enable building with debug information? Maybe disabled as default, I didn't knew where in the `main.py` to add that option.

```diff
diff --git a/tests/CMakeLists.txt b/tests/CMakeLists.txt
index c7cd3e8a7..3933c239e 100644
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -235,6 +235,7 @@ set(COMPILE_OPTIONS
     -Wundef
     -Wuninitialized
     -Wunreachable-code
+    -g
     ${BUILD_OPTIONS}
 )
```
---

The following test fails, I was expecting the event handler not being called because the checkbox is disabled:
```c
static volatile bool event_called = false;

static void event_handler(lv_event_t *e)
{

     if (LV_EVENT_VALUE_CHANGED == code) {
         event_called = true;
     }
}

void test_checkbox_should_not_call_event_handler_on_click_when_disabled(void)
{
    active_screen = lv_scr_act();
    checkbox = lv_checkbox_create(active_screen);

    lv_obj_add_state(checkbox, LV_STATE_DISABLED);
    lv_obj_add_event_cb(checkbox, event_handler, LV_EVENT_ALL, NULL);

    lv_test_mouse_click_at(checkbox->coords.x1, checkbox->coords.y1);

    TEST_ASSERT_FALSE(event_called);

    event_called = false;
}
```
---
> Did you tried using the setup and tearDown functions?

I've placed a brakpoint in the `setup` function and it was hitted only once when the `test_checkbox` ran, instead of the expected two times (one per test).

---
[debug_tests.patch.txt](https://github.com/lvgl/lvgl/files/7166046/debug_tests.patch.txt)

C47D:
@kisvegabor I've accidentally pushed a test for a bugfix (in a separated file), what should I do?

---

I think I don't fully understand the static text capability, I was expecting a chunk of memory being allocated to hold the static text, so the available memory would be reduced.

imagen

I guess I could simulate the checkbox being dissapeared by placing it inside an smaller scope, when the scope is exited, then the statically allocated memory for the static text should be released? Or simply by calling `lv_obj_del` on the checkbox?

cmumford:
> could we enable building with debug information? Maybe disabled as default, I didn't knew where in the `main.py` to add that option.

Sure. You can do that in `main.py` here: https://github.com/lvgl/lvgl/blob/master/tests/main.py#L198. I would just do:

```python
build_type = 'Debug'
```

IMO we're better off running these tests in debug mode. Maybe when we start doing performance tests or running on device (and need the RAM) we can run a release build.

tests/src/test_cases/test_checkbox.c
```diff
@@ -0,0 +1,114 @@
...
+    lv_mem_monitor_t m1;
+    lv_mem_monitor(&m1);
+
+    active_screen = lv_scr_act();
+    checkbox = lv_checkbox_create(active_screen);
+
+    initial_available_memory = m1.free_size;
+
+    lv_checkbox_set_text_static(checkbox, static_text);
+
```
kisvegabor, line 89: `lv_mem_monitor(&m1);` should be called here to get the new memory usage.

tests/src/test_cases/test_checkbox.c
```diff
@@ -0,0 +1,114 @@
...
+    active_screen = lv_scr_act();
+    checkbox = lv_checkbox_create(active_screen);
+
+    lv_checkbox_set_text_static(checkbox, static_text);
+
+    available_memory = m1.free_size;
+ 
+    lv_checkbox_set_text_static(checkbox, NULL);
+
```
kisvegabor: Here too.

tests/src/test_cases/test_bugfix.c
```diff
@@ -0,0 +1,29 @@
+#if LV_BUILD_TEST
+#include "../lvgl.h"
+
+#include "unity/unity.h"
+
+void test_bugfix_for_2522(void);
+
+void test_bugfix_for_2522(void)
```
embeddedt: We need to decide on the grouping for these. IMO, the test should be named `test_arc_angles_when_reversed` and should be in `test_arc.c`. (There can be a comment before/inside the function mentioning #2522.)

Otherwise, we start having tests for a single widget type being spread across many files.

kisvegabor: I agree.

C47D: Also agree, this file wasn't meant to be in this PR, I had it as a test for testing, and I noticed it was included in the commit after pushing it.

I can fix it with your suggestion tho.

C47D:
> IMO we're better off running these tests in debug mode. Maybe when we start doing performance tests or running on device (and need the RAM) we can run a release build.

OK, I will send a PR with your suggestion. I also agree running test with debug info enabled is very useful.

tests/src/test_cases/test_bugfix.c
C47D: Fixed in d425fe062

tests/src/test_cases/test_checkbox.c
C47D: Done in 29c674c

tests/src/test_cases/test_checkbox.c
C47D: Done in 29c674c

C47D:
I'm not sure about both static text tests, let me know if you want me to improve them.

C47D:
> Sure. You can do that in `main.py` here: https://github.com/lvgl/lvgl/blob/master/tests/main.py#L198. I would just do:
> 
> ```python
> build_type = 'Debug'
> ```

I did so but didn't got the symbol information when loading the executables into gdb.

kisvegabor:
> I'm not sure about both static text tests, let me know if you want me to improve them.

`NULL` shouldn't be a valid parameter in `lv_checkbox_set_text_static`. I've just updated the comment. So I think `test_checkbox_should_free_memory_when_static_text_is_refreshed` is not required.

C47D:
> So I think test_checkbox_should_free_memory_when_static_text_is_refreshed is not required.

Test removed

C47D:
Do we need more tests for now?

kisvegabor:
Looks god to me. Thank you, Carlos! 

Merging.

Issue #2560, opened 2021-09-16
General purpose dialog widget


mariotaku:
### Introduce the problem

We have `lv_msgbox`, which is a good widget to display a modal popup. It would be better to be more customizable to add something else, like a `lv_textarea` for example.

### Examples and cases

Desired result:

![image](https://user-images.githubusercontent.com/830358/133634561-ef48ace1-f3c4-4440-b823-bb4be91dc946.png)

### Suggested solution

This is already implemented. I'll create PR shortly after #2484.

rzr:
Sounds nice, Isn't this an extra widget built on core ones ?


FYI, I have tried my own for:

https://forum.lvgl.io/t/dialog-lvgl-ui-utility-to-use-along-scripts/6808


![https://git.ostc-eu.org/rzr/dialog-lvgl](https://forum.lvgl.io/uploads/default/original/2X/4/4a95749b4eb5a107d2f2f7ea17ea3a92b026f3fa.png)

mariotaku:
This is a modified version of `lv_msgbox`. And they share similar styles.

mariotaku:
Good work! Looks like yours is a standalone tool for other apps. While mine is a widget used inside LVGL based app itself.

kisvegabor:
What about considering `lv_msgbox` as a modular widget with functions like
```c
lv_msgbox_add_textarea();
lv_msgbox_add_img();
lv_msgbox_add_label();
lv_msgbox_add_container(); 
lv_msgbox_add_btnmatrix(); 
etc..
```

All elements would have the same width as the message box (width = 100%), and placed below each other.

mariotaku:
Currently it looks like `lv_win`. It has `lv_dialog_get_content()`. Yours is also a good option but this means user have to set styles of common parts on their own (btnmatrix etc). What do you think?

rzr:
I meant should those new widgets be optional to keep lvgl smaller as possible ? that said I wont be against reimplementing this extra dialog sample app with more advanced widgets, I'll keep an eye on it.

mariotaku:
@rzr Don't worry. Each widget are controlled by macro to turn it on or off. If it's not enabled, it will not be included into target binary.

kisvegabor:
> Currently it looks like lv_win

What is the added value of `lv_dialog` compared to `lv_win` and manually adding arbitrary widgets?

mariotaku:
[...]

PR #2561, opened 2021-09-16
added lv_dialog

mariotaku:

Added general purpose dialog as mentioned in #2560.

A `msgbox` like widget, but you can put any content into it.

image

### Checkpoints


kisvegabor:
Unless I miss something it's the same as `lv_msgbox` except adding a "content box", right?
If so, IMO we should rather improve the message box with it. It shouldn't break anything.

src/extra/widgets/msgbox/lv_msgbox.c
```diff
@@ -87,10 +103,15 @@ lv_obj_t * lv_msgbox_create(lv_obj_t * parent, const char * title, const char *
...
+    bool has_txt = txt && strlen(txt) > 0;
+    if (has_txt) {
+        mbox->text = lv_label_create(mbox->content);
+        lv_label_set_text(mbox->text, txt);
+        lv_label_set_long_mode(mbox->text, LV_LABEL_LONG_WRAP);
+        lv_obj_set_width(mbox->text, lv_pct(100));
+    }
 
     if(btn_txts) {
         mbox->btns = lv_btnmatrix_create(obj);
```
kisvegabor, line 117: Shouldn't it be created on `content` too?

mariotaku: Text is now placed within `mbox->content`. What do you think?

kisvegabor: I mean the btnmatrix.

mariotaku: @kisvegabor IMO it's better to put buttons outside. If the user want to, they can add complex layout inside `content` while still having consistent buttons. Also this is how Android design its dialog.

src/extra/themes/default/lv_theme_default.c
```diff
@@ -955,8 +955,15 @@ static void theme_apply(lv_theme_t * th, lv_obj_t * obj)
     else if(lv_obj_check_type(obj, &lv_msgbox_class)) {
         lv_obj_add_style(obj, &styles->card, 0);
         lv_obj_add_style(obj, &styles->msgbox_bg, 0);
+        lv_obj_set_style_flex_main_place(obj, LV_FLEX_ALIGN_START, 0);
```
kisvegabor: In the theme, I'd prefer adding only style. Local style needs more RAM and can't be overwritten by "normal" styles later. 

However doesn't it work to use  the default `LV_FLEX_ALIGN_START` on each align?

src/extra/themes/default/lv_theme_default.c
```diff
@@ -955,8 +955,15 @@ static void theme_apply(lv_theme_t * th, lv_obj_t * obj)
...
+        lv_obj_set_style_flex_cross_place(obj, LV_FLEX_ALIGN_CENTER, 0);
+        lv_obj_set_style_flex_track_place(obj, LV_FLEX_ALIGN_START, 0);
         return;
     }
+    else if(lv_obj_check_type(obj, &lv_msgbox_backdrop_class)) {
+        lv_obj_set_style_bg_color(obj, lv_palette_main(LV_PALETTE_GREY), 0);
```
kisvegabor: Here `lv_obj_add_style` would be really better.

src/extra/widgets/msgbox/lv_msgbox.c
kisvegabor: Makes sense.

src/extra/themes/default/lv_theme_default.c
mariotaku: I see. Default style should work. Let me try that one.

kisvegabor:
Looks good, thank you!

PR #2563, opened 2021-09-17
fixed OOB while deleting CJK character in textarea

mariotaku:

### Checkpoints

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Changelog
 
 ## v8.1.0 (In progress)
+- fix(txt) out of bound while deleting CJK character in textarea (#2563)
 - added sample lv_example_list_2.py
 - lv_obj_move_up(obj) and lv_obj_move_down(obj) added. (#2461)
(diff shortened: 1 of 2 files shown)


kisvegabor:
Can you send a code snippet in which the original code wasn't working well?

mariotaku:
@kisvegabor It can be produced in this way:
1. Send CJK 2 characters by `lv_indev_data_t.key`, that does not exist in font `lv_textarea` is currently using. 
2. Press backspace to delete characters in `lv_textarea`, until cursor reaches start
3. You will notice that placeholder still didn't appear
4. Press keyboard right for few times then press backspace
5. OOB in lv_txt.

Deleting CJK character should delete things in `char[]` by 3 bytes. Somehow, there is one left, causing `old_len < len`, then `pos` got integer overflow, causing illegal access to `txt`.

mariotaku:
So I think there is root cause to this which needs to be fixed.

mariotaku:
Oh I see! I think the root cause is a bug in my input driver. It sends 4 bytes instead of 3, causing this issue. Sorry. I think this can be closed.

PR #2565, opened 2021-09-17
Chart example: add gap between old and new data 

kisvegabor:

Demonstrate how to add gap between the newest and oldest points of a circular line chart.

@amirgon I was trying to add the MP version too but couldn't figure out how to convert this to get an array:
```c
    lv_coord_t * a = lv_chart_get_y_array(chart, ser);
```

### Checkpoints

### DIFF
--- a/examples/widgets/chart/lv_example_chart_9.c
+++ b/examples/widgets/chart/lv_example_chart_9.c
@@ -1,0 +1,46 @@
+#include "../../lv_examples.h"
+#if LV_USE_CHART && LV_DRAW_COMPLEX && LV_BUILD_EXAMPLES
+
+
+static void add_data(lv_timer_t * t)
+{
+    lv_obj_t * chart = t->user_data;
+    lv_chart_series_t * ser = lv_chart_get_series_next(chart, NULL);
+
+    lv_chart_set_next_value(chart, ser, lv_rand(10, 90));
+
+    uint16_t p = lv_chart_get_point_count(chart);
+    uint16_t s = lv_chart_get_x_start_point(chart, ser);
+    lv_coord_t * a = lv_chart_get_y_array(chart, ser);
+
+    a[(s + 1) % p] = LV_CHART_POINT_NONE;
+    a[(s + 2) % p] = LV_CHART_POINT_NONE;
+    a[(s + 2) % p] = LV_CHART_POINT_NONE;
+
+    lv_chart_refresh(chart);
+}
+
+/**
+ * Circular line chart with gap
+ */
+void lv_example_chart_9(void)
+{
+    /*Create a stacked_area_chart.obj*/
+    lv_obj_t * chart = lv_chart_create(lv_scr_act());
+    lv_chart_set_update_mode(chart, LV_CHART_UPDATE_MODE_CIRCULAR);
+    lv_obj_set_size(chart, 200, 150);
+    lv_obj_center(chart);
+
+    lv_chart_set_point_count(chart, 30);
+    lv_chart_series_t * ser = lv_chart_add_series(chart, lv_palette_main(LV_PALETTE_RED), LV_CHART_AXIS_PRIMARY_Y);
+    /*Prefill with data*/
+    uint32_t i;
+    for(i = 0; i < 30; i++) {
+        lv_chart_set_next_value(chart, ser, lv_rand(10, 90));
+    }
+
+    lv_timer_create(add_data, 300, chart);
+
+}
+
+#endif
--- a/examples/widgets/lv_example_widgets.h
+++ b/examples/widgets/lv_example_widgets.h
@@ -58,6 +58,7 @@ void lv_example_chart_5(void);
 void lv_example_chart_6(void);
 void lv_example_chart_7(void);
 void lv_example_chart_8(void);
+void lv_example_chart_9(void);
 
 void lv_example_checkbox_1(void);
 
--- a/src/extra/widgets/chart/lv_chart.c
+++ b/src/extra/widgets/chart/lv_chart.c
@@ -976,7 +976,7 @@ static void draw_series_line(lv_obj_t * obj, const lv_area_t * clip_area)
                         lv_draw_line(&p1, &p2, &series_mask, &line_dsc_default);
                     }
 
-                    if(point_w && point_h && ser->y_points[p_act] != LV_CHART_POINT_NONE) {
+                    if(point_w && point_h && ser->y_points[p_prev] != LV_CHART_POINT_NONE) {
                         lv_draw_rect(&point_area, &series_mask, &point_dsc_default);
                     }
 


amirgon:
> @amirgon I was trying to add the MP version too but couldn't figure out how to convert this to get an array:
> 
> ```c
>     lv_coord_t * a = lv_chart_get_y_array(chart, ser);
> ```

Currently returning an entire array (by returning a pointer to the first element of the array) is not supported in Micropython Bindings.  
(The other direction is supported - you can set an array by sending a Python List as an argument)

That's because - 
- It is not clear how many elements are in the array. How do we prevent the user from accessing past the last element? Currently in case of overflow the system would crash of behave unexpectedly.
- It is rarely used. In LVGL such array is usually (always?) set by the user, so the user already has access to that array without the need to read it from LVGL API.

Could you explain why reading chart coords is needed? (Or why some other array getter is needed?)  

If this is important, I may be able to implement this under the assumption that the user would never try to access elements out of bound, or preferably, if we have some way to detect the size of the array.

---

btw, how is the missing bullet fix related to demonstrating how to add gaps?

kisvegabor:
> If this is important, I may be able to implement this under the assumption that the user would never try to access elements out of bound, or preferably, if we have some way to detect the size of the array.

IMO it'd be better than nothing. We could implement an `lv_array_t` or so but IMO it'd be overkill and very cumbersome for a very little gain. 


> Could you explain why reading chart coords is needed? (Or why some other array getter is needed?)

To manually set some empty points as gap.


> btw, how is the missing bullet fix related to demonstrating how to add gaps?

I forgot the push the really relevant commit. :sweat_smile:

amirgon:
> > Could you explain why reading chart coords is needed? (Or why some other array getter is needed?)
> 
> To manually set some empty points as gap.

I still don't get it.
The user is the one who sets the array coords, right?
So the user knows (and can keep reference of) the array coords of his chart.
So why does the user need to read the array coords from LVGL if he already knows them, because he was the one who set them  in the first place?

kisvegabor:
> The user is the one who sets the array coords, right?

By default LVGL allocates and reallocates the "coord arrays" based on the set point count of the chart. It's only an option to use an external array e.g. if
- you already have the data in an array
- you store the array in a "special memory" e.g. external RAM or flash or so

amirgon:
> By default LVGL allocates and reallocates the "coord arrays" based on the set point count of the chart.

But it's still the user who sets the data, so he could keep track of it.

Anyway, If I implement this, could we agree on a way to tell the array size? Maybe by some new convention?  
For example, a new C function with `_array_size` suffix for each function that returns an array:
```c
size_t lv_chart_get_y_array_size(const lv_obj_t * obj, lv_chart_series_t * ser);
```

That way the binding could query array size for each array it is handling and verify the user is not overflowing.

kisvegabor:
> But it's still the user who sets the data, so he could keep track of it.

Yes, but usually one by one. Only set the next value. So don't have to keep track with the whole array. E.g. you measure the temperature in every 5 minutes and just push the measured value to the chart.

> Anyway, If I implement this, could we agree on a way to tell the array size? Maybe by some new convention?
For example, a new C function with _array_size suffix for each function that returns an array:

Unfortunately it's not that straight forward here. Both the x and y arrays have `point_count` size
 set by `lv_chart_set_point_count()`. So  I don't see a good logical way to add function(s) to get the array size.

I'd go with the unsafe version (no boundary check) or I can implement it as @uraich did [here](https://sim.lvgl.io/v8.1/micropython/ports/javascript/index.html?script_startup=https://raw.githubusercontent.com/lvgl/lvgl/3df2a7444758d2df023f321ccb5931de44af2a48/examples/header.py&script=https://raw.githubusercontent.com/lvgl/lvgl/3df2a7444758d2df023f321ccb5931de44af2a48/examples/widgets/chart/lv_example_chart_7.py).

amirgon:
> Unfortunately it's not that straight forward here. Both the x and y arrays have `point_count` size
> set by `lv_chart_set_point_count()`. So I don't see a good logical way to add function(s) to get the array size.

Could you please explain this?
I don't understand the problem. Why can't `lv_chart_get_y_array_size(...` return whatever was set by `lv_chart_set_point_count()`?

kisvegabor:
It could, but it'd be redundant to add 2 new functions with exactly the same purpose as  `lv_chart_set_point_count()`.

amirgon:
I'm looking into array access (without boundary checks).
Initial work on https://github.com/lvgl/lv_binding_micropython/pull/179, still requires some work to be usable.

kisvegabor:
Great!

amirgon:
@kisvegabor I've implemented array access.  
See details here: https://github.com/lvgl/lv_binding_micropython/pull/179#issuecomment-924456323

Please try to add the MP example you were working on with array access and let me know if there are any issues.

kisvegabor:
Great, thank you! It's still not available in the online simulator right?

amirgon:
> It's still not available in the online simulator right?

Actually it is available.  

@embeddedt took care of automatic deployment whenever lv_micropython is updated.

kisvegabor:
Great!  I made a quick test and it was working: 
[LINK](https://sim.lvgl.io/v8.1/micropython/ports/javascript/index.html?script_startup=https://raw.githubusercontent.com/lvgl/lvgl/51f3310592b488bf6225a13a7b74291751ea8912/examples/header.py&script=https://raw.githubusercontent.com/lvgl/lvgl/51f3310592b488bf6225a13a7b74291751ea8912/examples/widgets/chart/lv_example_chart_1.py&script_direct=17484d3920e0e349dd6ddcf5dc774785b18152c9)

kisvegabor:
@amirgon I've implemented the Python version [here](http://sim.lvgl.io/v8.1/micropython/ports/javascript/index.html?script_direct=e0582c4435640158c1917f4c42053a504d9e9e61).

Can you take look if it looks Pythonic enough? :slightly_smiling_face:

uraich:
Looks nice to me. In fact, these calls make treatment of the point array quite a bit simpler. When I did the examples, I had to treat the point array myself, outside the chart.
So, I think this is quite an improvement.

amirgon:
> @amirgon I've implemented the Python version [here](http://sim.lvgl.io/v8.1/micropython/ports/javascript/index.html?script_direct=e0582c4435640158c1917f4c42053a504d9e9e61).
> 
> Can you take look if it looks Pythonic enough?

Looks good to me .

I just don't understand why you need three assignment to CHART_POINT. NONE

Karijn:
Hmmm, I think that I was to fast with my pull request for MicroPython sample [lv_example_chart_9.py](https://github.com/lvgl/lvgl/pull/2604)

amirgon:
> my pull request for MicroPython sample

Ok, added question there.

kisvegabor:
Okay, I merge it then. Let's continue the MP part in #2604

Issue #2456, opened 2021-08-17
What do you think about having a single repo for LVGL?


kisvegabor:
### Introduce the problem

Now it's quite difficult (and not really solved) to handle the versioning, branching, and testing of repos other than the main lvgl repo. For example lv_demos, lv_lib_png, lv_lib_split_jpeg, lv_lib_qrcode, etc.

It'd be important to
- make it clear which version of the other repos should be used with a given version of lvgl
- have the same branching and release strategy for all repos
- be sure a change in lvgl doesn't break the other repos

### Suggested solution

On solution can be to merge the demos, drivers, and l library repos into the lvgl repo (and create a larger monorepo). This way everything would in one place, in sync, have the same branching, releases, etc. Besides the contributors would be concentrated into one common repo which is IMO good because get closer the the library itself. 

Finally, only the `lv_port_...` repos would remain as external repos.

Please, share you ideas, concerns, experiences with this monorepo concept.

embeddedt:
@kisvegabor FYI: I changed the title of the issue, as the original one sounded like we were creating an extra repository, not reducing the number.

kisvegabor:
Ah typo. I wanted to write "one mono repo" :slightly_frowning_face: 
Thanks!

embeddedt:
> Finally, only the lv_port_... repos would remain as external repos.

Can we make an exception for the Emscripten port, which has a very tiny amount of platform-specific code in it, as it is used directly by the docs? If we did, [this extra hack](https://github.com/lvgl/lvgl/blob/8ae894ebd4c6fda92dcd4e11d30483525bd9230b/scripts/build_html_examples.sh#L4-L13) would not be needed.

Miguel0101:
I agree to merge all repositories into one.

kisvegabor:
@embeddedt 

> Can we make an exception for the Emscripten port, which has a very tiny amount of platform-specific code in it, as it is used directly by the docs? If we did, this extra hack would not be needed.

Would it mean adding only a script to the lvgl repo?

@Miguel0101
> I agree to merge all branches into one.

It's about repositories, not branches.

Miguel0101:
Oh, i'm sorry.

cmumford:
I am very much in favor of a mono repo. One concern of mine is to provide a better workflow (certainly no worse) for the various ways in which LVGL is consumed. Specifically:

* Desktop (Ubuntu, Widows, macOS, etc.)
* ESP-IDF
[...]

PR #2569, opened 2021-09-19
[WIP] Add small 3rd party libs to lvgl 

kisvegabor:
PR for #2456

### Checkpoints


amirgon:
Do you think the Micropython bindings should bind these libs by default?  
Or just LVGL-core as today?
Perhaps this could be controlled through some configuration macros.

If we want this, we would need to consider the additional program size, and also make sure these libs conform to the naming conventions for callbacks etc.

kisvegabor:
> Do you think the Micropython bindings should bind these libs by default?

Definitely, yes.

> Perhaps this could be controlled through some configuration macros.

You can enable these libs in [lv_conf.h](https://github.com/lvgl/lvgl/pull/2569/files#diff-0be6707528460cdc1dd90dad35b2597054e790c6b811c3ed5f9d3ce41c105ed0R540-R577) as usual.

> If we want this, we would need to consider the additional program size, and also make sure these libs conform to the naming conventions for callbacks etc.

Sure, but these libs has very simple API:
- bmp, sjpg, png: No API, just registers a decoder
- fs: no API, just registers drivers for Posix, STDIO and FATFS APIs 
- gif: `lv_gif_create(parent)`, `lv_gif_set_src(obj, src)`
- qrcode: `lv_qrcode_create(parent)`, `lv_qrcode_update(obj, "Hello", 5)`. (I'll remove `lv_qrcode_delete`)
- freetype: It might need some work if needed in MP. See [here](https://github.com/lvgl/lvgl/pull/2569/files#diff-9cc8952b2e57ae51d454c79da05f54123add85d299d74e40661f9233bdca791a).

amirgon:
Added https://github.com/lvgl/lv_binding_micropython/issues/180 to track Micropython Bindings for these libraries.

kisvegabor:
I planning to merge this. Do you agree?

docs/libs/bmp.md
```diff
@@ -0,0 +1,31 @@
...
+:github_url: |github_link_base|/libs/bmp.md
+```
+
+# BMP decoder 
+
+Allow the use of BMP images in LVGL. 
+This implementation uses [bmp-decoder](https://github.com/caj-johnson/bmp-decoder) library.
+The pixel are read on demand (not the whole image is loaded) so using BMP imgages requires very little RAM.
+
+Compatible with LVGL's `feat/new-fs-api` branch.
```
embeddedt: Is this true still?

kisvegabor: No, it isn't. I've just updated all the related docs.

kisvegabor:
I merge this. We can add fixes later if required.

amirgon:
> > Do you think the Micropython bindings should bind these libs by default?
> 
> Definitely, yes.
> 
> >

The following 3rd party libs are now available on Micropython: BMP, SJPG, GIF, QRCODE.
@uraich It could be useful to add Micropython examples and make them available [on the docs](https://docs.lvgl.io/master/libs/index.html).

kisvegabor:
@amirgon what do you think about using the internal png lib instead of the external one?

amirgon:
> what do you think about using the internal png lib instead of the external one?

I'm not sure about that.
The PNG module we use today with Micropython has the following advantages:
- It's a lodepng submodule so it's really easy to update to latest version, either by us or by the user who builds lv_micropython
- It's configured to use Micropython's gc-ram, which is allocated on SPIRAM on ESP32. This is very important because lodepng consumes lots of RAM and ESP32 internal RAM is scarce. I would say lodepng is really useful only with SPIRAM.

kisvegabor:
Fair points. 

> It's configured to use Micropython's gc-ram, which is allocated on SPIRAM on ESP32. 

We can make the malloc configurable from lv_conf,h,

---

If someone wants to use the internal PNG lib, does he need to do anything apart from enabling it in lv_conf.h?

amirgon:
> If someone wants to use the internal PNG lib, does he need to do anything apart from enabling it in lv_conf.h?

He would need to remove the other "external" lodepng otherwise he'll have compilation/linking issues. It's the same lodepng library compiled twice.

kisvegabor:
> He would need to remove the other "external" lodepng otherwise he'll have compilation/linking issues. It's the same lodepng library compiled twice.

I see, thanks.

uraich:
I wanted to have a look into the lib examples and I tried to compile the bmp example for the simulator.
However, I only get:
undefined reference to `lv_example_bmp_1'
What am I missing?
I tried to add :
/*---------------
 * 3rd party libs
 *--------------*/
#define LV_FS_USE_POSIX 'A'
to lv_conv.h
Is this correct? and what is going to be the root directory of A: ? Are the drivers for STDIO and POSIX available and included?
Do I need anything else there to enable the libs?

I have the feeling, I don't really understand how this thing is supposed to work.

amirgon:
> I have the feeling, I don't really understand how this thing is supposed to work.

Hi Uli!

The BMP library is [enabled by default on lv_micropython](https://github.com/lvgl/lv_binding_micropython/blob/e7e4a1ab663b9f7bebb31bcd965d44c2b5e052b4/lv_conf.h#L553-L554).

[According to the docs](https://github.com/lvgl/lvgl/blob/master/docs/libs/bmp.md#:~:text=if%20enabled%20in%20lv_conf.h%20by%20lv_use_bmp%20lvgl%20will%20register%20a%20new%20image%20decoder%20automatically%20so%20bmp%20files%20can%20be%20directly%20used%20as%20image%20sources): 
> If enabled in lv_conf.h by LV_USE_BMP LVGL will register a new image decoder automatically so BMP files can be directly used as image sources. For example:

> `lv_img_set_src(my_img, "S:path/to/picture.bmp");`

You would first need to register the Micropython File System driver, [like this](https://github.com/lvgl/lv_binding_micropython/blob/e7e4a1ab663b9f7bebb31bcd965d44c2b5e052b4/examples/Dynamic_loading_font_example.py#L26-L27):

```python
import fs_driver
fs_drv = lv.fs_drv_t()
fs_driver.fs_register(fs_drv, 'S')
```

And then all you really need to do is to load a BMP file to an image.

uraich:
Hi Amir,
Hmmm, ... and then it works!
image
Thanks !

uraich:
Is there a way to use freetype fonts in MicroPython?
I see that they are disabled in lv_conf.h and if I enable them I get
../../lib/lv_bindings/lvgl/src/extra/libs/freetype/lv_freetype.c:12:10: fatal error: ft2build.h: No such file or directory
when trying to re-compile lv_micropython
bmp: ok
sjpg: ok
qrcode: ok
The other examples need some more work.

amirgon:
> Is there a way to use freetype fonts in MicroPython?

Currently freetype is not supported.
We could probably add this in the future, but that would require some work to build the freetype library itself in lv_micropython.  
The freetype library is large so I'm not sure how practical/useful freetype is on microcontrollers. Possibly we could enable it only for the unix port.

We also don't support the 3rd part FS drivers (LV_USE_FS_STDIO_*) because in Micropython we simply use `fs_driver` and don't care about the underlying file system.

uraich:
@kisvegabor : I tried to compile the libs examples on the simulator but did not succeed. 
I get: main.c:(.text+0xc4): undefined reference to `lv_example_bmp_1'
I see that the source has been compiled, but it seems that the object file is not taken into account. Maybe you also need some additions to lv_conf.h? What about the posix drivers? Are they available?
I have ported the bmp, sjpg and the qrcode example to MicroPython and uploaded them to 
https://github.com/uraich/lv_mpy_examples_v8
for inspection.
freetype is not implemented in lv_micropython, therefore I will skip this example
png is handled differently in micropython as discussed above. I could add these examples using the micropython way if you wish.
rlottie and gif need some more work.

uraich:
@amirgon : As far as I can see, the bmp and the gif examples are almost identical. I just replaced the bmp by the gif file. This however does not work. The window stays white. What do I miss?

amirgon:
> As far as I can see, the bmp and the gif examples are almost identical. I just replaced the bmp by the gif file. This however does not work. The window stays white. What do I miss?

https://github.com/lvgl/lvgl/blob/18f61c5f7743b8a8636ada287927c0de9b92a267/examples/libs/bmp/lv_example_bmp_1.c#L9-L12

https://github.com/lvgl/lvgl/blob/18f61c5f7743b8a8636ada287927c0de9b92a267/examples/libs/gif/lv_example_gif_1.c#L16-L19


Not exactly the same.  

On BMP you create an **image object** and call `lv_img_set_src`.  
On GIF you create a **special Gif object** and call `lv_gif_set_src`.

amirgon:
Regarding rlottie on Micropython:  I can't enable `LV_USE_RLOTTIE` by default on lv_micropython's "lv_conf.h" because it would require rlottie to always be installed on all micropython ports.  

One idea is to use `__has_include` to check if rlottie headers exist and enable `LV_USE_RLOTTIE` only if they are, but I'm not sure how robust that would be, as we would also need to link with librlottie.
What do you think?  

If this make sense, why not add this `__has_include` to `lv_conf_template.h` on LVGL?

uraich:
@amirgon : You were again right! In fact all I needed to do was replacing 
img = lv.**img**(lv.scr_act()) by img = lv.**gif**(lv.scr_act())
image
rlottie I did not try yet. If I can easily create a lv_micropython version with rlottie enabled, then I will try.
I ported the png example using the lodepng method.
Should I already create a pull request for the 
bmp, gif, sjpg, png and qrcode examples?

kisvegabor:
> rlottie I did not try yet. If I can easily create a lv_micropython version with rlottie enabled, then I will try.

If you `make` and `install` rlottie before building MP, the compiler should find and link the installed lib. `lv_rloitte` is just a wrapper around the rlottie lib to use it as a widget.

uraich:
Since I am running Ubuntu 20.04 I installed the librlottie-dev package. When re-compiling lv_micropython I get:
../../lib/lv_bindings/lvgl/src/extra/libs/rlottie/lv_rlottie.c: In function ‘lv_rlottie_constructor’:
../../lib/lv_bindings/lvgl/src/extra/libs/rlottie/lv_rlottie.c:104:26: error: conversion from ‘double’ to ‘size_t’ {aka ‘long unsigned int’} may change value [-Werror=float-conversion]
  104 |     rlottie->framerate = lottie_animation_get_framerate(rlottie->animation);
      |                          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
../../lib/lv_bindings/lvgl/src/extra/libs/rlottie/lv_rlottie.c:123:26: error: assignment to ‘const uint8_t *’ {aka ‘const unsigned char *’} from incompatible pointer type ‘uint32_t *’ {aka ‘unsigned int *’} [-Werror=incompatible-pointer-types]
  123 |     rlottie->imgdsc.data = rlottie->allocated_buf;
      |                          ^
../../lib/lv_bindings/lvgl/src/extra/libs/rlottie/lv_rlottie.c:107:16: error: unused variable ‘parent’ [-Werror=unused-variable]
  107 |     lv_obj_t * parent = lv_obj_get_parent(obj);
      |                ^~~~~~
I must have a closer look into this

kisvegabor:
In [Line 104](https://github.com/lvgl/lvgl/blob/749d1b3ec31ec2ef27f594ed0a4af93edb2c10f0/src/extra/libs/rlottie/lv_rlottie.c#L104) there a cast to `(size_t)` for me.

uraich:
Ok, I had to add a few casts to lv_rlottie.c and lv_mpy.c, and I had to add librlottie.so to the libraries in Makefile to get lv_micropython compiled with rlottie. Unfortunately lv_mpy.c is generated such that the modification will be overwritten the next time you build from scratch. 
With these modifications, things seem to work. I tried example 2:
image

amirgon:
> Ok, I had to add a few casts to lv_rlottie.c and lv_mpy.c, and I had to add librlottie.so to the libraries in Makefile to get lv_micropython compiled with rlottie. Unfortunately lv_mpy.c is generated such that the modification will be overwritten the next time you build from scratch.

@uraich - I've updated lv_micropython to fix these problems and automatically enable rlottie if it's available.
Currently this works only for the unix port.

@embeddedt - I wonder how hard it would be to support rlottie also on the JS port?

embeddedt:
I can probably get rlottie itself working without too much trouble, once I have some time. Loading animations from a file looks like it will be an issue, however, because according to the docs, it uses the standard C file API and would thus bypass our custom file logic.

amirgon:
> because according to the docs, it uses the standard C file API and would thus bypass our custom file logic.

Interesting.  
So what actually does happen when the JS port access the posix file system in Emscrpiten?  
I thought that our custom logic was actually plugged into the posix file system API, doesn't it? The Micropython file system is using posix API on the unix port, how did you replace it with the custom logic?

embeddedt:
The upstream JavaScript port doesn't have any file support. [pmp-p](https://github.com/pmp-p) duplicated the Unix port's file system logic and [hacked it at MicroPython level](https://github.com/lvgl/lv_micropython/blob/619672d0c30ca2408f7de931854f181aa8398f78/ports/javascript/file.c#L212-L213) to be able to open online files as well. However, that bypasses the POSIX API Emscripten provides.

It's possible we could override Emscripten's `open` implementation with our own and move this logic in there.

amirgon:
> It's possible we could override Emscripten's `open` implementation with our own and move this logic in there.

That sounds cleaner. Keep a unix-port-like file system and plug in on the posix level on Emscripten.

---

I tried installing freetype-2.11.0, to try it with Micropython.
It builds and installs correctly but there is something wrong with the include directories.
It looks like it expects the user to add include paths and does not rely on `/usr/local/include`. For example, `ft2build.h` contains:
```
#include <freetype/config/ftheader.h>
```
but "freetype" is under `/usr/local/include/freetype2/` and not under `/usr/local/include` so it's not found without adding that include path, unless I'm missing something here.

amirgon:
@kisvegabor, @uraich  - lv_micropython now supports both rlottie and freetype.

I've tested them both with a simple example.
If you want to try them, make sure you install the library first (rlottie, freetype) with regular `make install`, before building lv_micropython.  `lv_conf.h` and the Makefile first check for the library availability and enable the library on Micropython only if it's available on the machine.

Currently these are supported only on the unix port.

uraich:
@amirgon : I checked lv_conf.h and the freetype library was simply disabled (not checking the include file). I enabled it and lv_micropython compiled without problems with freetype included.
Apart from this it doesn't look bad. Here is the port of the freetype example:
image

amirgon:
> I checked lv_conf.h and the freetype library was simply disabled (not checking the include file).

It checks if the include file exists. 
Are you using [latest version](https://github.com/lvgl/lv_binding_micropython/blob/d3451c5cdf6acd1f6c5d82b879a890e61a15ebf5/lv_conf.h#L572) ? (and updated submodules recursively?)
```c
/*FreeType library, if available*/
#if defined __has_include
#  if __has_include (<freetype2/ft2build.h>)
#    define LV_USE_FREETYPE 1
#  else
#    define LV_USE_FREETYPE 0
#  endif
#else
#define LV_USE_FREETYPE 0
#endif
```

uraich:
Maybe I was too fast upgrading, and you did not upload yet when I pulled. Everything is fine now.
I am done with all the lib exercises and I will prepare the pull request now.

PR #2571, opened 2021-09-20
fix(btnmatrix): make ORed values work correctly with lv_btnmatrix_has_btn_ctrl

Johennes:

This commit replaces the current `actual & expected` check in `lv_btnmatrix_has_btn_ctrl` with `(actual & expected) == expected`. This is required to make the function work with ORed control flags because otherwise a parity in *any* bit will result in a return value of `true` even if not all expected bits are set.

Here is an example demonstrating the issue:

```
lv_btnmatrix_ctrl_t actual = LV_BTNMATRIX_CTRL_CHECKED;
lv_btnmatrix_ctrl_t expected = LV_BTNMATRIX_CTRL_CHECKED | LV_BTNMATRIX_CTRL_DISABLED;

// will print 1 (true) even though LV_BTNMATRIX_CTRL_DISABLED is not set
printf("%d\n", actual & expected ? true : false);

// will print 0 (false)
printf("%d\n", ((actual & expected) == expected) ? true : false);
```

### Checkpoints

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -32,6 +32,7 @@
 - fix(zoom) multiplication overflow with zoom calculations on 16-bit platforms
 - feat(msgbox): omit title label unless needed
 - feat(msgbox): add function to get selected button index
+- fix(btnmatrix): make ORed values work correctly with lv_btnmatrix_has_btn_ctrl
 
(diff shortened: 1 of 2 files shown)


kisvegabor:
Good catch! Thank you!

PR #2572, opened 2021-09-20
fix compiling of examples when cmake is used

JojoS62:

the lvgl repo fails to compile when cmake is used and target is not ESP, Zephyr or Micropython.
In this case, lvgl is built for a core and an example library. The core builds, the examples fail due to wrong include path in some asset files. Some asset.c files use correct ../../lvgl.h include, some are set to use the LV_LVGL_H_INCLUDE_SIMPLE define and use lvgl/lglv.h, which does not point to the include in the lvgl repo.

This PR chages the .c files in assets to use the LV_LVGL_H_INCLUDE_SIMPLE and ../../lvgl.h

Maybe it is even better that the files in assets use also the ../lv_examples.h include.

### Checkpoints

### DIFF
--- a/examples/assets/animimg001.c
+++ b/examples/assets/animimg001.c
@@ -1,8 +1,4 @@
-#if defined(LV_LVGL_H_INCLUDE_SIMPLE)
-#include "lvgl.h"
-#else
 #include "../../lvgl.h"
-#endif
 
 #if LV_BUILD_EXAMPLES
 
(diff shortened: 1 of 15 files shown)


kisvegabor:
As the example has fixed place using only `#include "../../lvgl.h"` is enough. What do you think?

JojoS62:
yes, I think also this will we enough. Maybe the define was used for some historical reasons?

kisvegabor:
Thank you, merging!

PR #2577, opened 2021-09-21
Child group support

mariotaku:

We can use multiple groups to control focus scope. Sometimes we want to let all focusable objects under a container to be added into one group automatically. Of course we could set default group while creating subviews, but it would be troublesome.

With `lv_obj_set_child_group`, you can assign a group to any object, but not adding it into one. Then any new objects created will be added into this group. If no child group specified, it will use default one.

We can further extend msgbox, by assigning child group upon its creation, and specify indev to send key events to it. Then this dialog will consume all key events exclusively.

### Checkpoints

### DIFF
--- a/src/core/lv_group.c
+++ b/src/core/lv_group.c
@@ -97,16 +97,6 @@ lv_group_t * lv_group_get_default(void)
     return default_group;
 }
 
-lv_group_t * lv_group_get_child_group(lv_obj_t * obj)
-{
-    lv_group_t *group = NULL;
-    do {
-        group = lv_obj_get_child_group(obj);
-        obj = lv_obj_get_parent(obj);
-    } while(group == NULL && obj != NULL);
-    return group;
-}
-
 void lv_group_add_obj(lv_group_t * group, lv_obj_t * obj)
 {
     if(group == NULL) return;
--- a/src/core/lv_group.c
+++ b/src/core/lv_group.c
@@ -97,6 +97,16 @@ lv_group_t * lv_group_get_default(void)
     return default_group;
 }
 
+lv_group_t * lv_group_get_child_group(lv_obj_t * obj)
+{
+    lv_group_t *group = NULL;
+    do {
+        group = lv_obj_get_child_group(obj);
+        obj = lv_obj_get_parent(obj);
+    } while(group == NULL && obj != NULL);
+    return group;
+}
+
 void lv_group_add_obj(lv_group_t * group, lv_obj_t * obj)
 {
     if(group == NULL) return;
--- a/Kconfig
+++ b/Kconfig
@@ -128,6 +128,16 @@ menu "LVGL configuration"
                     shadow size is `shadow_width + radius`.
                     Caching has LV_SHADOW_CACHE_SIZE^2 RAM cost.
 
+            config LV_CIRCLE_CACHE_SIZE
+                int "Set number of maximally cached circle data"
+                depends on LV_DRAW_COMPLEX
+                default 4
+                help
+                    The circumference of 1/4 circle are saved for anti-aliasing
+                    radius * 4 bytes are used per circle (the most often used
+                    radiuses are saved).
+                    Set to 0 to disable caching.
+
             config LV_IMG_CACHE_DEF_SIZE
                 int "Default image cache size. 0 to disable caching."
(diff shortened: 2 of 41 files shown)


src/core/lv_group.c
```diff
@@ -97,6 +97,16 @@ lv_group_t * lv_group_get_default(void)
     return default_group;
 }
 
+lv_group_t * lv_group_get_child_group(lv_obj_t * obj)
```
embeddedt: I think this API name may confuse the Micropython bindings generator. (cc @amirgon)

mariotaku: Good point. I think this can be moved to lv_obj.

amirgon: 👍

kisvegabor:
> Of course we could set default group while creating subviews, but it would be troublesome.

By "troublesome" do you mean incontinent or are there any issues with it?

Can you mention a few use-cases when this feature comes in handy?

mariotaku:
@kisvegabor By troublesome I mean one must call them each time they want to assign a bunch of views to another group, and restore to the previous group after they finish that.

In my implementation, I assign a separate group for each msgbox in my own theme:

```c
if (lv_obj_check_type(obj, &lv_msgbox_class)) {
    lv_group_t *group = lv_group_create();
    group->user_data = app_input_get_group();
    lv_obj_set_child_group(obj, group);
    app_input_set_group(group);
    lv_obj_add_event_cb(obj, msgbox_destroy, LV_EVENT_DELETE, NULL);
}
```

And restore to the previous one when this msgbox gets destroyed:

```c
lv_group_t *group = lv_obj_get_child_group(lv_event_get_current_target(event));
if (app_input_get_group() == group) {
    app_input_set_group(group->user_data);
}
lv_group_remove_all_objs(group);
lv_group_del(group);
```

By doing this, all buttons and inputs will be added into this msgbox, so when you press LV_KEY_NEXT, it will only cycle through those widgets while the dialog is there.

kisvegabor:
Hm, interesting. I'd approach this problem differently. What if there were a global `obj_created_cb(obj)`. In this you can check if the widget is on a msgbox and assign it to a group. A callback like this 
- is more versatile because not limited only to the groups
- doesn't consume extra memory

Some use cases for a create callback:
- add fade in animation automatically on certain objects
- customize the style without touching the theme
- position in a special way depending on other variables (e.g. special layout)
- and of course add the widget to a group in a custom way.
 
Similarly there could be  `obj_deleted_cb` too that is called automatically when the widget is being deleted.


To achieve what you have described an `lv_obj_has_parent_with_class(obj, &lv_class_msgbox)` can be useful.

What do you think?

mariotaku:
I have another use case about this feature, is to have 2 panes.

image

Keys are handled differently in 2 groups. If we use a centralized `obj_created_cb`, then this callback will have to be strongly coupled with code in this screen. Which is not ideal when organizing code.

Besides, I think this change wouldn't introduce too many ram usage even for microcontrollers.

kisvegabor:
> If we use a centralized obj_created_cb, then this callback will have to be strongly coupled with code in this screen. Which is not ideal when organizing code.

It's true.


My gut feeling says that there is a more generic missing feature here. Basically the task is to trigger the parent when new children are created on it. What if we added `LV_EVENT_CHILD_CREATED/DELETED` events for the container. 

It'd allow higher flexibility for `group`s too. What if you want to add all children to the group, except one, or except on type? `child_group` will add all the object with `lv_obj_is_group_def() == true`.

With events on the parent, code organization becomes much easier. 

To make possible to react on children on any level `LV_EVENT_CHILD_CREATED/DELETED` should be sent to all parents (bubble).

It also opens the door for many interesting features. E.g. if there is a light theme but there is a dark element, all children of the dark element their style can be tweaked when the children are created on it. 

What do you think?

mariotaku:
@kisvegabor Event for child creation/deletion would be really nice. I didn't look deep into event system so I'll be looking forward to this feature in the future.

Miguel0101:
> I have another use case about this feature, is to have 2 panes.
> 
> image

Very nice UI!

kisvegabor:
I've opened #2618 with `LV_EVENT_CHILD_CREATE/DELETE` support. Let's continue there.

PR #2579, opened 2021-09-22
fix(snapshot) snapshot is affected by parent's style because of wrong coords

XuNeo:
### Coords of object for snapshot should be reset to position (0,0).

Assume an _objectB_ has parent _objectA_, when _objectA_ has a style of pad_all set to non-zero value like 10, _objectA_'s coords should be manually set to 0,0 before taking snapshot. Otherwise, _objectA_'s style will affect snapshot of _objectB_.

To solve the problem, manually set the coordinates to (0,0), keep size unchanged, and mark object layout as dirty when finishing snapshot.

The generated image size is also updated to real size drawn to buffer.

Below code on project lv_sim_eclipse_sdl can be used to reproduce the problem.
```c
void snapshort_demo(void)
{
  lv_obj_t* root = lv_obj_create(lv_scr_act());
  lv_obj_remove_style_all(root);
  lv_obj_set_size(root, 800, 100);
  lv_obj_set_style_pad_all(root, 40, 0);

  lv_obj_t* container = lv_obj_create(root);
  lv_obj_set_size(container, 800, 100);
  lv_obj_set_style_bg_color(container, lv_color_black(), 0);
  lv_obj_set_style_bg_opa(container, LV_OPA_50, 0);

  lv_img_dsc_t *dsc = lv_snapshot_take(container, LV_IMG_CF_TRUE_COLOR_ALPHA);
  lv_obj_t* img = lv_img_create(lv_scr_act());
  lv_img_set_src(img, dsc);
  lv_obj_align(img, LV_ALIGN_BOTTOM_MID, 0, 0);
}
```
### Checkpoints

### DIFF
--- a/src/extra/others/snapshot/lv_snapshot.c
+++ b/src/extra/others/snapshot/lv_snapshot.c
@@ -61,6 +61,9 @@ uint32_t lv_snapshot_buf_size_needed(lv_obj_t * obj, lv_img_cf_t cf)
     /*Width and height determine snapshot image size.*/
     lv_coord_t w = lv_obj_get_width(obj);
     lv_coord_t h = lv_obj_get_height(obj);
+    lv_coord_t ext_size = _lv_obj_get_ext_draw_size(obj);
+    w += ext_size * 2;
+    h += ext_size * 2;
 
     uint8_t px_size = lv_img_cf_get_px_size(cf);
     return w * h * ((px_size + 7) >> 3);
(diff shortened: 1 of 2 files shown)


kisvegabor:
Hi,

When testing it I found two other potential bug. If the object (that is subject of the snapshot) is larger then the screen its image will be clipped. E.g. if there is a text very long text (1000px tall) on a 320x240 screen the the max height of the snapshop will be 240.

So I agree to move the object to (0;0) but the resolution of the dummy screen should be the size of the object.

The other issue is that if the widget has shadow it will be clipped from the snapshot because the shadow is drawn out of the "real coordinates". To solve this we need to increase the size of the snapshot image by `obj->spec_attr->ext_draw_size`.

What do you think about these?

XuNeo:
I agree. I have made the changes. Could you please check if that would work? I haven't made some test cases.

src/extra/others/snapshot/lv_snapshot.c
```diff
@@ -138,6 +143,12 @@ lv_res_t lv_snapshot_take_to_buf(lv_obj_t * obj, lv_img_cf_t cf, lv_img_dsc_t *
     obj->parent = screen;
 
     disp->inv_p = 0;
+
+    obj->coords.x2 = lv_area_get_width(&obj->coords) - 1;
+    obj->coords.x1 = 0;
```
kisvegabor: It should start on `ext_size` else the shadow will be clipped.

Where do you restore the original coordinates?

XuNeo: I have corrected coords start value to ext_size.

Originally, I just mark obj layout as dirty and invalidate it. I have updated code to backup original coords and restore it at end.

kisvegabor:
Just tested and work very well for shadows too.

Thank you!

PR #2583, opened 2021-09-22
feat(disp): Enable rendering to display subsection

Johennes:

This change introduces new fields on `lv_disp_drv_t` that allow to specify the size of the full display and the offset of the display
subsection that is being rendered to. The values are used to transform the drawing area before calling `flush_cb` so that only the desired part of the full display is being rendered to.

Relates to: lvgl/lv_drivers#166

### Checkpoints

This is a draft for evaluating the proposal made in https://github.com/lvgl/lv_drivers/issues/166#issuecomment-924820582. I tried using this in my framebuffer application to only draw my UI in the lower third of the screen and leave the upper two thirds untouched:

```
    /* Query display size */
    uint32_t hor_res;
    uint32_t ver_res_full;
    fbdev_get_sizes(&hor_res, &ver_res_full);

    /* Position display at the bottom of the screen */
    uint32_t ver_res = ver_res_full / 3;
    uint32_t ver_off = 2 * ver_res;

    /* Prepare display buffer */
    const size_t buf_size = hor_res * ver_res / 10; /* At least 1/10 of the display size is recommended */
    lv_disp_draw_buf_t disp_buf;
    lv_color_t *buf = (lv_color_t *)malloc(buf_size * sizeof(lv_color_t));
    lv_disp_draw_buf_init(&disp_buf, buf, NULL, buf_size);    

    /* Initialise display driver */
    static lv_disp_drv_t disp_drv;
    lv_disp_drv_init(&disp_drv);
    disp_drv.draw_buf = &disp_buf;
    disp_drv.flush_cb = fbdev_flush;
    disp_drv.hor_res = hor_res;
    disp_drv.ver_res = ver_res;
    disp_drv.physical_hor_res = hor_res;
    disp_drv.physical_ver_res = 3 * ver_res;
    disp_drv.offset_x = 0;
    disp_drv.offset_y = 2 * ver_res;
    lv_disp_drv_register(&disp_drv);
```

This does seem to work really well in my testing so far.


src/hal/lv_hal_disp.h
```diff
@@ -77,8 +77,13 @@ typedef enum {
  */
 typedef struct _lv_disp_drv_t {
 
-    lv_coord_t hor_res;         /**< Horizontal resolution.*/
-    lv_coord_t ver_res;         /**< Vertical resolution.*/
+    lv_coord_t hor_res;         /**< Horizontal resolution of the display portion that is being rendered to.*/
+    lv_coord_t ver_res;         /**< Vertical resolution of the display portion that is being rendered to.*/
+
+    lv_coord_t display_hor_res;      /**< Horizontal resolution of the full display. Set to -1 for fullscreen mode.*/
```
embeddedt: I'd prefer using more standard terms for the hardware's width/height if possible (e.g. I think this should really be called `stride`, not sure about `display_ver_res` though), as otherwise these will be confused with `hor_res` and `ver_res`.

src/core/lv_refr.c
```diff
@@ -964,5 +964,16 @@ static void call_flush_cb(lv_disp_drv_t * drv, const lv_area_t * area, lv_color_
...
                color_p);
-    drv->flush_cb(drv, area, color_p);
+
+    lv_coord_t hor_res = drv->display_hor_res > 0 ? drv->display_hor_res : drv->hor_res;
+    lv_coord_t ver_res = drv->display_ver_res > 0 ? drv->display_ver_res : drv->ver_res;
+
+    lv_area_t offset_area = {
+        .x1 = LV_MAX(0, LV_MIN(area->x1 + drv->display_offset_x, hor_res - 1)),
+        .y1 = LV_MAX(0, LV_MIN(area->y1 + drv->display_offset_y, ver_res - 1)),
+        .x2 = LV_MAX(0, LV_MIN(area->x2 + drv->display_offset_x, hor_res - 1)),
```
kisvegabor: At this stage we can't truncate the area because `color_p` still contains `drv->hor_res` lines. So if `area` is different than `color_p` was rendered width to user won't be able to find the original dimensions of `color_p`.

Area should be truncated much earlier in the pipeline, but let's postpone this problem a little :slightly_smiling_face: 

Here I suggest only moving the area by an offset and ignore that case if it's out of the display.

src/hal/lv_hal_disp.h
kisvegabor: `stride` sounds reasonable at fist look but thinking more about it it's also confusing. Let's say LVGL has 320 px hor_res to draw. The display has 400 px hor res. Bur for optimal memory access the the frame buffer has N x 32 px hor res = 416 px. In this case the "real" stride is 416. And 400 is only the "real" resolution of the display.  

However, if LVGL works in `disp_drv->full_refresh = 1` and the stride is different from the LVGL's hor res, LVGL will render incorrectly as LVGL can't handle stride now. (But it should.) 

So it seems to me `stride` is an new parameter on top of `hor/ver_res` and `display_hor/ver_res`. 

`display_hor_res` is really a little bit unfortunate name because it's easy to mix with `hor_res`. Something like `physical_hor_res` or `real_hor_res`?

src/core/lv_refr.c
Johennes: Ah, great catch, I missed this. :facepalm: 

Changed it to only apply the offset here without any truncation.

src/hal/lv_hal_disp.h
Johennes: Agree that the naming gets confusing with two displays. :disappointed: 

I changed it to `physical_hor_res` / `physical_ver_res` and `offset_x` / `offset_y` and also updated the comments / docs a little to make the meaning more clear. Let me know if that seems ok.

docs/porting/display.md
```diff
@@ -69,6 +69,10 @@ LVGL might render the screen in multiple chunks and therefore call `flush_cb` mu
 
 ### Optional fields 
 There are some optional display driver data fields:
+- `physical_hor_res` horizontal resolution of the full / physical display in pixels. Only relevant when _not_ using fullscreen mode, otherwise set to -1.
```
kisvegabor: This comment suggests that the user needs to set it to `-1` is unused. I'd rather say: "...set it only if not using the full screen..."

kisvegabor (commented):
Just a minor thing in the comment.

docs/porting/display.md
Johennes: Ah, you're right, yes. I changed it.

kisvegabor (approved):

Miguel0101 (approved):

embeddedt (commented):
Hmm... correct me if I'm wrong, but `physical_hor_res` and `physical_ver_res` are ignored by LVGL, and all drivers have to have references to `hor_res` and `ver_res` replaced with the physical equivalents for this feature to work properly, right?

If so, do we also need physical versions of [these functions](https://github.com/lvgl/lvgl/blob/d57eb7614dcfd6d72fb6bae6b28bf52f843897ca/src/hal/lv_hal_disp.h#L243-L255) to be added for MicroPython-based drivers?

Johennes:
> Hmm... correct me if I'm wrong, but `physical_hor_res` and `physical_ver_res` are ignored by LVGL, and all drivers have to have references to `hor_res` and `ver_res` replaced with the physical equivalents for this feature to work properly, right?

Yes, that's correct but I'm not sure if the physical resolution will be needed in _all_ drivers. The original problem that led to this PR is specified in https://github.com/lvgl/lv_drivers/issues/166. In a non-full-screen app, touchscreens don't work correctly with the libinput driver because the driver maps absolute screen coordinates to absolute display coordinates. Pointer devices, in turn, still work correctly in this case because they only rely on relative coordinate changes.

> If so, do we also need physical versions of [these functions](https://github.com/lvgl/lvgl/blob/d57eb7614dcfd6d72fb6bae6b28bf52f843897ca/src/hal/lv_hal_disp.h#L243-L255) to be added for MicroPython-based drivers?

I can do that if you want me to. I'm not using MicroPython currently though so I can't judge if this will be needed or not.

Johennes:
Ok, I added physical versions of the functions in question now. Let me know if anything else is needed.

Johennes:
Merged `master` in (only had a conflict in the changelog).

kisvegabor:
Looks good to me. @embeddedt ?

embeddedt (approved):

embeddedt:
Looks good to me as well. Merged, thank you!

PR #2593, opened 2021-09-24
fix(snapshot): make coords of screen of fake display same as obj.

XuNeo:
### Snapshot may behave strangly if screen coords is not same as obj for snapshot.

Not sure if it's related to relative layouts, but I have met the problem that snapshot image has wrong layout for children in obj, with this patch it works.

### Checkpoints

### DIFF
--- a/src/extra/others/snapshot/lv_snapshot.c
+++ b/src/extra/others/snapshot/lv_snapshot.c
@@ -123,8 +123,10 @@ lv_res_t lv_snapshot_take_to_buf(lv_obj_t * obj, lv_img_cf_t cf, lv_img_dsc_t * 
 
     lv_disp_drv_init(&driver);
     driver.draw_buf = &draw_buf;
-    driver.hor_res = w;
-    driver.ver_res = h;
+
+    /*Make the display big enough to involve the objects on its original places. */
+    driver.hor_res = obj->coords.x1 + w;
+    driver.ver_res = obj->coords.y1 + h;
     lv_disp_drv_use_generic_set_px_cb(&driver, cf);
 
     disp = lv_disp_drv_register(&driver);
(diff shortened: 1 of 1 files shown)


kisvegabor:
I'm not 100% sure but the issue could be that if you just change the coordinates of an object its children won't move because LVGL stores absolute coordinates. I wasn't think of that earlier... :slightly_frowning_face: 

I have any idea:
The gist is to not move the object but adjust the display size in a smart way. E.g. there is an 20x30 object on 50;100. So the coordinates of the object are: 50;100 70;130. We need a 20x30 buffer for sure and a display with 100x130 size to involve the object on its original place. `lv_obj_invalidate()` will invalidate only a 20x30 area. I hope it makes sense and works :slightly_smiling_face:

XuNeo:
Yes, you are right. This works.

Another issue is that In case of object bigger than its parent, snapshot image is also clipped by size of its parent. 
When we shift object coords by ext_size, it also changes the snapshot since ext_size is added in all directions. Do you think it's acceptable? 

The bottom object is snapshot image. Snapshot is taken from the label. All objects has a border width of 1.
image

kisvegabor:
> Another issue is that In case of object bigger than its parent, snapshot image is also clipped by size of its parent.

I haven't tried with it but if we move the object to a new screen which is large enough to hold the object and its ext_size the original parent shouldn't matter.

XuNeo:
I think screen isn't the probelm.
Without doing full layout update on new display, reusing obj's current coords will have this clipped issue. Considering a simpler case: object's (x1, y1) is negative.

kisvegabor:
Having negative coordinates is really a problem. It seems we can not avoid changing the position of the object. :slightly_frowning_face:

PR #2603, opened 2021-09-25
feat(menu): Add menu widget

HX2003:

I created a new widget to easily create multi-level menus. If there is any interest, I will follow up on this.

Animation
### Checkpoints


Miguel0101:
Very good!

embeddedt:
Looks very nice! I will try to review the code in detail soon.

amirgon:
Hi @HX2003 ,
Do you plan to add Micropython example?

HX2003:
> Hi @HX2003 ,
> Do you plan to add Micropython example?

Yes, it is in the works now.

kisvegabor:
Looks very useful!

I couldn't check you implementation in details so I have only 2 comments:
- Please add comments to the enums and functions  `lv_menu.h`
- ~1 year ago I thought that having functions like `lv_menu_item_add_btn` is useful but I needed to conclude that its very unlikely that people need a menu item exactly like this. So I rather suggest having a lean API for the menu widget and show how to add special menu items in examples. It makes easier to maintain, document and learn the widget. An the examples help people to learn how to customize the menu items to their needs.

HX2003:
@kisvegabor Makes sense.

kisvegabor:
Hi,

I've tested the widget. It looks very flexible but at first a few things have puzzled me:
- `lv_menu_item_create` creates a new "menu page", right? Maybe it's only me but in my mind "menu item" is rather like a button in the list. 
-  For the user's point of view `lv_menu_cont_create` and `lv_menu_item_add_obj` has slightly overlapping functionality. Would it be possible to combine them somehow?
- `LV_MENU_SECTION_ON/OFF`: What if the user explicitly created the sections with a `lv_menu_add_section()` function and later add and arbitrary objects to the section or the menu itself. The content of the menu can be arranged by a colum layout automatically. This way - ideally - `lv_menu_generate` wouldn't be required. 
- The basic idea is to open a submenu on clicking a menu item, right? In the end it's a click event that loads the new page of the menu (`lv_menu_traverse_event_cb`). Using `lv_menu_item_set_obj` and `lv_menu_item_add_obj` I couldn't easily understand in which case shall I use which. An API like `lv_menu_item_set_target_page(menu, btn, menu_page)`.
- I haven't immersed in header and sidebar yet.

In summary - as a user - a very minimal API like this seemed very inuitive:
```c
lv_obj_t * menu = lv_menu_create(parent);
lv_obj_t * page1 =  lv_menu_create_page(menu);
lv_obj_t * page2 =  lv_menu_create_page(menu);

lv_obj_t * btn1 = lv_btn_create(page1);
lv_obj_add_event_cb(btn1, lv_menu_load_page_event, page2);

lv_obj_t * label1 = lv_label_create(page2);
```

Of course it doesn't mention the header and sidebar, only the menu itself. 

Maybe I missed something important and oversimplified it. So let me know it doesn't make sens :slightly_smiling_face:

HX2003:
@kisvegabor I just looked at your snippet and I think you're on to something. It seems like there could be potential for massive simplification of the code, by using `lv_obj_t` for a page, maybe there's not a need for `lv_menu_item_t` anymore.

kisvegabor:
It's glad that you like the idea :)

src/extra/widgets/menu/lv_menu.h
```diff
@@ -0,0 +1,238 @@
...
+ * @return pointer to the created menu page
+ */
+lv_obj_t * lv_menu_page_create(lv_obj_t * parent);
+
+/**
+ * Create a menu cont object
+ * @param parent pointer to an object, it will be the parent of the new menu cont object
+ * @return pointer to the created menu cont
+ */
+lv_obj_t * lv_menu_cont_create(lv_obj_t * parent);
```
kisvegabor, line 112: Is it required? It seems to me the user can add any object and configure the layouts, flags as he needs. 

E.g. the user can add sliders, labels, etc directly to the page, can't he?

Same question for `lv_menu_section_constructor` and `lv_menu_separator_create`.

kisvegabor (commented):
Thank you for the updates. It looks more straight forward to me. 

Still one thing is not clear for ma about the sidebar: what is the unique "power" of the side bar? Why can't it be just an other menu placed on the right?

src/extra/widgets/menu/lv_menu.h
HX2003: You are correct that it is not absolutely required to use these. Layout can even be a grid pattern. I added in because it makes for theming as in using lv_theme_default easier. In the default theme this container has styles for checked and clicked state.

HX2003:
@kisvegabor Sidebar is unique since the page is always displayed. What is clicked will have also checked state applied. I don't how several menus object would interact with each other.

I might have accidentally clicked the re-request a review button.

kisvegabor:
I see.
I was looking into the 5th example,  however wasn't clear to me how the content of the side bar is generated. Could you tell a few words about it?

HX2003:
@kisvegabor The content in the sidebar is generated the same way like any other page. Once sidebar mode is enabled, the first menu page will be placed in the sidebar.

kisvegabor:
Ah, got it. So the the first call of `lv_menu_set_page()` sets the sidebar. 

What about explicitly setting a sidebar like: `lv_menu_set_sidebar(menu, page1)`. It'd allow easily change the sidebar and `lv_menu_set_mode_sidebar()` wouldn't be required.

Regarding the header: I wonder if there are some simplification opportunities here. I see that it makes sens to allow having header in the main area and the sidebar too. But the appearance of the header could be very different in case of different UIs, i.e. there is no generic header that works for >99% of the users. What about having only
```c
lv_menu_get_header(menu);           //User can create any kind of buttons and labels here as needed
lv_menu_get_sidebar_header(menu);   //Same here
lv_menu_back_event();               //Can be called from the back button's event
```

Sorry for nitpicking that much. I think it's a cool widget and want to be sure it's as intuitive and versatile as possible.

HX2003:
> What about explicitly setting a sidebar like: `lv_menu_set_sidebar(menu, page1)`. It'd allow easily change the sidebar and `lv_menu_set_mode_sidebar()` wouldn't be required.

What would be the advantage of doing so? I consider it as a mode because the sidebar container and header is not created if sidebar is not used.

> Regarding the header: I wonder if there are some simplification opportunities here. I see that it makes sens to allow having header in the main area and the sidebar too. But the appearance of the header could be very different in case of different UIs, i.e. there is no generic header that works for >99% of the users. What about having only
> 
> ```c
> lv_menu_get_header(menu);           //User can create any kind of buttons and labels here as needed
> lv_menu_get_sidebar_header(menu);   //Same here
> lv_menu_back_event();               //Can be called from the back button's event
> ```

I updated example 3 to show how the header can be modified, let me know what you think. I have ```lv_menu_set_main_header_back_btn``` so that the back btn (either the entire header or just a back button inside the header can be automatically hidden, leaving other text and icons still showing), the click event bubbles up to the header.
```lv_menu_back_event``` can be added also, but I can't see when it would be useful

**Comment by Karijn** [comment]
Looks good, but not all samples have a MicroPython variant. When needed i can help adding those later on if you want me to.

**Comment by kisvegabor** [comment]
> What would be the advantage of doing so? I consider it as a mode because the sidebar container and header is not created if sidebar is not used.


I meant only a minor change.  `lv_menu_set_mode_sidebar` can be changed to
```c
void lv_menu_set_sidebar_page(lv_obj_t * obj, lv_obj_t * page)
{
  lv_menu_t * menu = (lv_menu_t *)obj;
  menu->sidebar_page = page;   //Used instead of mode_sidebar: if not NULL enabled else disabled
}
```

---

What do you think about this:
```c
lv_menu_page_create(menu, "Title");
```
The title can be added automatically to the header too.

> I updated example 3 to show how the header can be modified, let me know what you think. 

Thank you. I've played with header modes too and now understand how it really works. IMO your approach fine.

---

I have some minor ideas for naming but let's leave it until everything else is set.

HX2003:
@kisvegabor 
> The title can be added automatically to the header too.

I think that's ok. So this title should be shown even when the back button icon is not. Since the entire header is the 'back button' by default (icon difficult to click on small screens), there needs to be a differentiation between the  'back button' and the 'back icon'.
For customization, same as the header back button, there will be something like 'lv_menu_set_main_header_title_label' for the label object (not the text), right?

kisvegabor:
IMO it's fine to hide the whole header on the root page (as now) if this feature is enabled.

`lv_obj_set_ext_click_area` can be used to make the click area of the back icon larger. Or maybe the icon can be a button. If the user wishes he can make the background transparent to keep only the back symbol.

HX2003:
Great, title can be set now. The default behaviour remains unchanged. I made it so that if title is present it will be shown even if the back icon is hidden.
title

kisvegabor:
Great, thanks!

In `lv_menu_example_3`  ASAN reports a "use after free" error. I've tracked that it is caused by the deletion of back icon in `lv_obj_del(lv_obj_get_child(header, 0));` but the `menu->main_header_back_icon` is not `NULL`ed.

What about keeping track only with `main_header_back_btn` and `main_header_back_icon` can be its only child. IMO the button is the gist. (by default the button can be transparent, hidden by the user, or change its content, size, etc) 

On smartphones I usually see that the title is aligned next to the button. What do think about using a flex layout on the header like this:
```c
lv_obj_set_flex_flow(main_header, LV_FLEX_FLOW_ROW);
lv_obj_set_flex_align(main_header, LV_FLEX_ALIGN_START, LV_FLEX_ALIGN_CENTER, LV_FLEX_ALIGN_CENTER);
```

HX2003:
> What about keeping track only with `main_header_back_btn` and `main_header_back_icon` can be its only child. IMO the button is the gist. (by default the button can be transparent, hidden by the user, or change its content, size, etc)

As I see, if I follow your approach, it means I will lose ability to set the whole header (with multiple children) as a 'button', am I right? I know this behaviour is not commonly seen, so let me think about it. lv_obj_set_ext_click_area does not achieve the same effect.

> On smartphones I usually see that the title is aligned next to the button. What do think about using a flex layout on the header like this:

I shall change it:) Anything else?

kisvegabor:
> As I see, if I follow your approach, it means I will lose ability to set the whole header (with multiple children) as a 'button', am I right? I know this behaviour is not commonly seen, so let me think about it. lv_obj_set_ext_click_area does not achieve the same effect.

Yes, I imagined something like:
```c
lv_obj_t * btn = lv_btn_create(header);
lv_obj_set_size(btn, LV_SIZE_CONTENT, lv_pct(100)); //Full height
lv_label_set_text(label, LV_SYMBOL_LEFT "Back");

...

lv_menu_get_back_btn(menu);
```

---

One minor refactoring thing: I'd remove the `MODE` word from `LV_MENU_MODE_HEADER_...` and `LV_MENU_MODE_ROOT_BACK_BTN_...` because IMO it has the same meaning without it but slightly shorter.

HX2003:
@kisvegabor I see, that works too. Since I'm busy, I will revisit this at a later time.

kisvegabor:
Great, no problem.

kisvegabor:
It looks good to me. Let me know if I can merge it.

HX2003:
@kisvegabor Sure!

kisvegabor:
Great, merging! Thank you for this useful and versatile widget!

bosiden:
Thanks for this, very useful. Any idea how to make it work with rotary input? Whats the best way to add menu items to a group?

bosiden:
Thanks for this, it doesn't work with rotary encoder input out of the box right? What would be the best way to have the menu items added to an lv_group?

HX2003:
@bosiden Encoder is not supported by default, but I'm looking into it. 
` lv_group_t * g = lv_group_create();`
`lv_group_set_default(g);`
`lv_indev_set_group(indev, g);`

Then add this for the menu cont you want to be able to focus.
`lv_group_add_obj(g, menu_cont);`
`lv_obj_clear_flag(menu_cont, LV_OBJ_FLAG_SCROLLABLE);`
`lv_obj_add_flag(menu_cont, LV_OBJ_FLAG_SCROLL_ON_FOCUS);`

That way things like sliders and switches which are supported by default will work too.

Issue #2534, opened 2021-09-06
WIP: Support bitbake 


rzr:
Hi,
I am creating this ticket to track my progress to support LVGL along Bitbake (Yocto/OE).
I plan to upstream some fixes and build recipes somewhere, and will refer to this issue for cooperation purposes

Here is a SDL2 version running on RPi3 on AllScenariOS:

![](https://forum.ostc-eu.org/uploads/default/optimized/1X/7be1f53c210dde35e06150f5a4ec43a4307f2cce_2_690x388.jpeg)

Relate-to: https://git.ostc-eu.org/OSTC/planning/core-os/-/issues/219

kisvegabor:
Thanks for opening this issue. Let us know if you need help form our side.

rzr:
So far it goes fine thx,
I am also about to publish a simple app that can be used as reference,  and reference image.
It's working with sdl2 backend which is still under review ( https://github.com/lvgl/lvgl/pull/2484 )
I'll update this ticket for anyone who want to use it, feedback will be appreciated.
Then some hints about wayland backend would be welcome, but one problem at a time.

Regards

kisvegabor:
> Then some hints about wayland backend would be welcome, but one problem at a time.

Have you already seen its [README](https://github.com/lvgl/lv_drivers/tree/master/wayland)?

kisvegabor:
Not stale

rzr:
yes I plan to upstream recipes to OE, meanwhile check status at:

https://git.ostc-eu.org/rzr/dialog-lvgl/-/wikis/home

k3wals:
Thanks for doing this. I have a primitive recipe for integrating LVGL in Yocto.

In case it's helpful, here's what I added to weston.ini to hide the panel to run fullscreen

panel-position=none

rzr:
You're welcome, feel free to review:
https://github.com/openembedded/meta-openembedded/pull/447

yes I have been using weston as kiosk for an other project (maybe a simple compositor would be desirable too)

rzr:
It's merged in master branch
https://github.com/openembedded/meta-openembedded/commit/0924d41ac4cf968fc302e4f7e52d969cd3899423

Should I keep this issue open for tracking ? 

Thanks for support anyway

rzr:
Let me crosslink this issue about support in dunfell branches:

https://github.com/openembedded/meta-openembedded/pull/453

Meanwhile it can be imported downstream it worked for me too as tracked at:

[...]

PR #2621, opened 2021-09-29
build: Add install rule

rzr:
This can be help to install lvgl on systems,
for clients applications.

Relate-to: https://github.com/lvgl/lvgl/issues/2534
Signed-off-by: Philippe Coval <philippe.coval@astrolabe.coop>

### Checkpoints

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -94,6 +94,8 @@ else()
 
 endif()
 
+option(install "Enable install to system" OFF)
+if(install)
 include_directories(${CMAKE_SOURCE_DIR})
 
 file(GLOB LVGL_PUBLIC_HEADERS
@@ -122,4 +124,4 @@ set_target_properties(${CMAKE_PROJECT_NAME} PROPERTIES
 install(TARGETS ${CMAKE_PROJECT_NAME}
     ARCHIVE DESTINATION "${LIB_INSTALL_DIR}"
     PUBLIC_HEADER DESTINATION "${INC_INSTALL_DIR}")
-
+endif(install)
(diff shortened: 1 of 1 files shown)


kisvegabor:
Won't it mess up hardware specific cases? E.g. `ESP_PLATFORM`, `ZEPHYR_BASE`, `MICROPY_DIR`.

rzr:
I dont think so f other platforms are not using "make install" this will be ignored 
and if they do, then issue can be solved later or cmake should do it better than us.

We could make it more linux specific but it would add more noise.
I can surround by if linux if it help to get it merged

kisvegabor:
To be honest I'm not an expert of CMake so can't really judge if it's ok or not.

@embeddedt, @xiaoxiang781216, @amirgon could you take a look?

amirgon:
@rzr I'm not sure this works with ESP32 Micropython port.
Did you test it on this configuration?

rzr:
No I didnt but if "make install" is not invoked it should not cause any harm, but if you want to make sure let's add if (UNIX) test

amirgon:
> No I didnt but if "make install" is not invoked it should not cause any harm, but if you want to make sure let's add if (UNIX) test

I think it makes sense to ensure it doesn't break anything, even when "install" is not called.

@embeddedt @kisvegabor a CI test for Micropython ESP32 port could be useful since it relies on LVGL cmake script, if you think it is worth the extra time LVGL tests run.

kisvegabor:
@amirgon I agree with both. 

@rzr Please, add an `if` for "install".

rzr:
well I made a quick test with platformio (which invoke cmake) and I can confirm there is no regression,
but I can add an option to enable this part , like if unix if you want to make sure...

kisvegabor:
> if unix if you want to make sure...

It'd be great, thank you. IMO it also improves readability.

rzr:
I isolated the "if(install)" change

kisvegabor:
Thank you, merging!

PR #2623, opened 2021-09-29
test(txt) initial unit tests and general code cleanup/fixes

C47D:

What the title says, if you need me to add any special test please let me know 😸

### Checkpoints

### DIFF
--- a/src/misc/lv_txt.c
+++ b/src/misc/lv_txt.c
@@ -484,7 +484,7 @@ void _lv_txt_encoded_letter_next_2(const char * txt, uint32_t * letter, uint32_t
 /**
  * Give the size of an UTF-8 coded character
  * @param str pointer to a character in a string
- * @return length of the UTF-8 character (1,2,3 or 4). O on invalid code
+ * @return length of the UTF-8 character (1,2,3 or 4), 0 on invalid code.
  */
 static uint8_t lv_txt_utf8_size(const char * str)
 {
@@ -496,7 +496,7 @@ static uint8_t lv_txt_utf8_size(const char * str)
         return 3;
     else if((str[0] & 0xF8) == 0xF0)
         return 4;
-    return 0; /*If the char was invalid tell it's 1 byte long*/
+    return 0;
 }
 
 /**
@@ -679,7 +679,8 @@ static uint32_t lv_txt_utf8_get_byte_id(const char * txt, uint32_t utf8_id)
     uint32_t byte_cnt = 0;
     for(i = 0; i < utf8_id && txt[byte_cnt] != '\0'; i++) {
         uint8_t c_size = _lv_txt_encoded_size(&txt[byte_cnt]);
-        byte_cnt += c_size > 0 ? c_size : 1;
+        /* If the char was invalid tell it's 1 byte long*/
+        byte_cnt += c_size ? c_size : 1;
     }
 
     return byte_cnt;
--- a/src/misc/lv_txt.c
+++ b/src/misc/lv_txt.c
@@ -78,6 +78,12 @@
  *      MACROS
  **********************/
 
+#define IS_ASCII(value)             ((value & 0x80U) == 0x00U)
+#define IS_2BYTES_UTF8_CODE(value)  ((value & 0xE0U) == 0xC0U)
+#define IS_3BYTES_UTF8_CODE(value)  ((value & 0xF0U) == 0xE0U)
+#define IS_4BYTES_UTF8_CODE(value)  ((value & 0xF8U) == 0xF0U)
+#define IS_INVALID_UTF8_CODE(value) ((value & 0xC0U) != 0x80U)
+
 /**********************
  *   GLOBAL FUNCTIONS
  **********************/
@@ -488,13 +494,13 @@ void _lv_txt_encoded_letter_next_2(const char * txt, uint32_t * letter, uint32_t
  */
 static uint8_t lv_txt_utf8_size(const char * str)
 {
-    if((str[0] & 0x80) == 0)
+    if(IS_ASCII(str[0]))
         return 1;
-    else if((str[0] & 0xE0) == 0xC0)
+    else if(IS_2BYTES_UTF8_CODE(str[0]))
         return 2;
-    else if((str[0] & 0xF0) == 0xE0)
+    else if(IS_3BYTES_UTF8_CODE(str[0]))
         return 3;
-    else if((str[0] & 0xF8) == 0xF0)
+    else if(IS_4BYTES_UTF8_CODE(str[0]))
         return 4;
     return 0;
 }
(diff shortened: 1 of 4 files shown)


tests/src/test_cases/test_txt.c
```diff
@@ -0,0 +1,21 @@
+#if LV_BUILD_TEST
+#include "../lvgl.h"
+
+#include "unity/unity.h"
+
+void test_txt_should_identify_empty_string_txt_parameter_when_trying_to_get_next_line(void);
+
+/* See #2615 for more information */
+void test_txt_should_identify_empty_string_txt_parameter_when_trying_to_get_next_line(void)
```
embeddedt: It seems a bit long to me... how about this?

```suggestion
void test_txt_next_line_handles_empty_string(void)
```

C47D: That's much better, thanks for the suggestion.

C47D:
For this tests it would be nice to test the decoders with some strings? What do you think?

kisvegabor:
> For this tests it would be nice to test the decoders with some strings? What do you think

Yes, it'd be great. IMO these functions are well-testable:
- `_lv_txt_encoded_next`
-  `_lv_txt_cut`
- `_lv_txt_ins`

C47D:
I'm writing a test for `_lv_txt_ins`, it's using the `lv_txt_utf8_get_byte_id` function, is there a way to also test the `LV_TXT_ENC_ASCII` static functions?

Also, I think this patch needs to be added to `_lv_txt_ins`, passing `NULL` to `strlen` seems to be undefined behavior.
**EDIT**
We should also add a note about the `_lv_txt_ins` parameters `txt_buf` and `ins_txt`, both must be NULL terminated, otherwise we get undefined behavior when calling `strlen` on them.

```diff
diff --git a/src/misc/lv_txt.c b/src/misc/lv_txt.c
index ca51e0b81..4d586d23d 100644
--- a/src/misc/lv_txt.c
+++ b/src/misc/lv_txt.c
@@ -440,6 +440,8 @@ bool _lv_txt_is_cmd(lv_text_cmd_state_t * state, uint32_t c)
  */
 void _lv_txt_ins(char * txt_buf, uint32_t pos, const char * ins_txt)
 {
+    if(txt_buf == NULL || ins_txt == NULL) return;
+
     size_t old_len = strlen(txt_buf);
     size_t ins_len = strlen(ins_txt);
     if(ins_len == 0) return;
```

C47D:
TODO:

- [ ] Handle `lv_txt_utf8_size` returning 0, update it's docs.

C47D:
I've added a comment about `_lv_txt_cut` in `test_txt_cut_len_longer_than_string`, is that the expected behavior?

tests/src/test_cases/test_txt.c
```diff
@@ -0,0 +1,152 @@
...
+void test_txt_should_identify_invalid_start_of_command(void);
+void test_txt_should_identify_scaped_command_in_parameter(void);
+void test_txt_should_skip_color_parameter_in_parameter(void);
+void test_txt_should_reset_state_when_receiving_color_cmd_while_processing_commands(void);
+void test_txt_should_identify_space_after_parameter(void);
+void test_txt_should_insert_string_into_another(void);
+void test_txt_should_handle_null_pointers_when_inserting(void);
+void test_txt_cut_should_handle_null_pointer_to_txt(void);
+void test_txt_cut_happy_path(void);
+void test_txt_cut_len_longer_than_string(void);
```
embeddedt: As much as it's good practice to have prototypes in C, I don't see what benefit they really bring here.

I'm assuming you added them because of warnings/errors, so IMO it'd be preferable to just turn off `-Wmissing-prototypes` when compiling the test files themselves.

C47D: The test compilation broke without them, but you're right, I will try to turn that warning off.

C47D: Fixed with 8b3217d

src/misc/lv_txt.h
```diff
@@ -122,15 +122,15 @@ bool _lv_txt_is_cmd(lv_text_cmd_state_t * state, uint32_t c);
 
 /**
  * Insert a string into an other
- * @param txt_buf the original text (must be big enough for the result text)
+ * @param txt_buf the original text (must be big enough for the result text and NULL terminated)
  * @param pos position to insert (0: before the original text, 1: after the first char etc.)
- * @param ins_txt text to insert
+ * @param ins_txt text to insert, must be NULL terminated
```
kisvegabor:
```suggestion
 * @param ins_txt text to insert, must be '\0' terminated
```

src/misc/lv_txt.h
```diff
@@ -122,15 +122,15 @@ bool _lv_txt_is_cmd(lv_text_cmd_state_t * state, uint32_t c);
...
  */
 void _lv_txt_ins(char * txt_buf, uint32_t pos, const char * ins_txt);
 
 /**
  * Delete a part of a string
- * @param txt string to modify
+ * @param txt string to modify, must be NULL terminated and should point to a heap or stack frame, not read-only memory.
```
kisvegabor:
```suggestion
 * @param txt string to modify, must be '\0' terminated and should point to a heap or stack frame, not read-only memory.
```

tests/CMakeLists.txt
```diff
@@ -221,7 +221,6 @@ set(COMPILE_OPTIONS
     -Wextra
     -Wformat-security
     -Wmaybe-uninitialized
-    -Wmissing-prototypes
```
kisvegabor: Without this tests won't report missing prototypes in lvgl either, right?

embeddedt: Yes. It seems that we need another variable for test file-specific compilation options.

I think that could probably be added [here](https://github.com/lvgl/lvgl/blob/a32b44ecb6304ef642bb394a87493a20923b9762/tests/CMakeLists.txt#L292), although I don't tinker with CMake that much.

C47D: The list `COMPILE_OPTIONS` is the one I've edited in this commit, which is the list of compilation options being passed to the compiler in the line you're pointing out. I've reverted this commit and used the approach in 64909e3, it seems to be working, AFAIK clang also supports that pragma.

embeddedt: Isn't it [the same set of options used for LVGL](https://github.com/lvgl/lvgl/blob/a32b44ecb6304ef642bb394a87493a20923b9762/tests/CMakeLists.txt#L244)?

C47D: Yup, I've updated my comment and also proposed a way to ignore the missing prototype warning in the test files. What do you think?

embeddedt: It can work, but I'm not a huge fan of using pragmas. IMHO it'd be better to add a second variable for test-specific compilation options to the CMake file and combine it with `COMPILE_OPTIONS` for the tests. I tried to do it [here](https://github.com/lvgl/lvgl/pull/2623/commits/2226683570df222ed627cb4e22a13fd3e2e47b94); what do you think? It should be possible to revert https://github.com/lvgl/lvgl/pull/2623/commits/64909e30ed124ca1e8ca390ca0639479c3e34f44 with this.

kisvegabor: I'd also prefer a CMake option as it's cleaner and more compiler independent.

C47D:
I will test your new commit and revert the pragma one later today @embeddedt

tests/CMakeLists.txt
C47D: It seems to be working as expected, I've tested it after reverting 64909e3, can you mark the conversation as resolved?

C47D:
> I've added a comment about `_lv_txt_cut` in `test_txt_cut_len_longer_than_string`, is that the expected behavior?

If so we can add a note about it in the docs.

tests/src/test_cases/test_txt.c
```diff
@@ -0,0 +1,139 @@
...
+}
+
+void test_txt_cut_len_longer_than_string(void)
+{
+    char msg[] = "Hello World";
+
+    _lv_txt_cut(msg, 0, 30);
+
+    // msg ends up being all NULL, len and old_len are equal, so we
+    // set up all of the txt array members to one element past txt len
```
kisvegabor: Not all elements but only `msg[0] = '\0'` as the loop runs only once. I think this behavior makes sense.

C47D: Updated in 7693d79

C47D:
`_lv_txt_encoded_next` test inputs from here https://stackoverflow.com/a/3886015 , let me know what do you think

src/misc/lv_txt.c
```diff
@@ -78,21 +78,16 @@
  *      MACROS
  **********************/
 
+#define IS_ASCII(value)             ((value & 0x80U) == 0x00U)
```
embeddedt: I think these need LV_ prefixes, unfortunately, as there tends to always be a random project which has taken a generic name like this.

C47D: Added in 1a3576c13

embeddedt (approved):

kisvegabor (approved):
Looks good, thank you very much!

Issue #2589, opened 2021-09-23
Calling lv_anim_timeline_del is deleting/stopping an active animation NOT attached to the deleted timeline


tvanfossen:
### Perform all steps below and tick them with [x]
- [ x] Check the related part of the [Documentation](https://docs.lvgl.io/)
- [x ] Update lvgl to the latest version
- [x ] Reproduce the issue in a [Simulator](https://docs.lvgl.io/latest/en/html/get-started/pc-simulator.html)

### Describe the bug

I am working on a gui that has several actions on-going on the screen at any given time and I am using animation timelines to handle the stopping/starting of these updates cleanly as new data is brought onto the screen.

I have a function for View_OnDataChange that gets called every few seconds, or when there is an active update of data (ie. a button is pressed). Any necessary updates to this view are added to the data_change_timeline I am using.

If the data_change_timeline has been started on entry to this function, then I call lv_anim_timeline_stop and lv_anim_timeline_del before starting a fresh timeline with the relevant animations merged into the new timeline. This has worked great so far.

On the end of the data_change_timeline, I add an extra lv_anim_t that just points to callback to indicate when the timeline has been completed such that I can delete it (as otherwise there is a memory leak from lv_anim_timeline_create). When this callback is triggered, I delete the timeline for garbage cleanup. This again works great for the data driven animations.

Now to the actual bug. One of several views, I have a particular icon that I want to fade in/out repeatedly throughtout the life of the screen, but it is not tied to any given timeline. So, once the screen has been created, I start this particular animation with lv_anim_start, but it is never added to ANY timeline as it is meant to run repeatedly. This animation has a callback which then continuously kicks off fresh animations with the relevant fade in/out value (LV_OPA_TRANP -> LV_OPA_COVER and vice verse).

Problem is, this animation stops arbitrarily. If I remove the call to delete the data_change_timeline, the animation continues on as expected.

It would seem that lv_anim_timeline_del(at) is arbitrarily deleting animations that are not attached to the timeline at all, and have been started outside of the timeline.

Is this an intended behavior such that there can only be 1 active timeline, or no active timeline and any number of active animations? 

I have some ideas for alternative ways to perform this same behaviour, but this still seemed a very apparent bug to me that the animation timeline should stop a animation which has never been attached to it

### To Reproduce

```c
//Example of bug,

static void anim_set_arc_value_cb(void * var, int32_t v)
{
    lv_arc_set_value(var, v);
};

[...]

PR #2628, opened 2021-10-03
fix(anim_timeline) avoid calling lv_anim_del(NULL, NULL)

FASTSHIFT:

This PR is about the resolution of https://github.com/lvgl/lvgl/issues/2589.
Avoid calling lv_anim_del(NULL, NULL) to cause all animations to be deleted.

### Checkpoints


src/misc/lv_anim_timeline.c
```diff
@@ -114,7 +114,11 @@ void lv_anim_timeline_stop(lv_anim_timeline_t * at)
 
     for(uint32_t i = 0; i < at->anim_dsc_cnt; i++) {
         lv_anim_t * a = &(at->anim_dsc[i].anim);
-        lv_anim_custom_del(a, (lv_anim_custom_exec_cb_t)a->exec_cb);
+
+        /*Avoid calling lv_anim_del(NULL, NULL) to cause all animations to be deleted*/
+        if (a->var != NULL || a->exec_cb != NULL) {
+            lv_anim_custom_del(a, (lv_anim_custom_exec_cb_t)a->exec_cb);
```
kisvegabor: Why can't we use simple `lv_anim_del`?

FASTSHIFT: OK

kisvegabor:
So far I thought that an animation doesn't make much sens if neither `var` nor `exec_cb` is set. But setting only the `ready_cb` might be also valid use case, especially in case of anim timeline. 

An other solution could be to add a dummy callback to `exec_cb`  in `lv_anim_timeline_add` if it's `NULL`.

BTW, won't the current solution (in this PR) cause memory leak?

FASTSHIFT:
> So far I thought that an animation doesn't make much sens if neither nor is set. But setting only the might be also valid use case, especially in case of anim timeline.`var``exec_cb``ready_cb`
> 
> An other solution could be to add a dummy callback to in if it's .`exec_cb``lv_anim_timeline_add``NULL`
> 
> BTW, won't the current solution (in this PR) cause memory leak?

Why is there a memory leak, is it `lv_anim_timeline` or `lv_anim`?

kisvegabor:
> Why is there a memory leak, is it lv_anim_timeline or lv_anim?


Because if both `var` and `exec_cb` are `NULL` the animation won't be deleted.

FASTSHIFT:
> > Why is there a memory leak, is it lv_anim_timeline or lv_anim?
> 
> Because if both `var` and `exec_cb` are `NULL` the animation won't be deleted.

`lv_anim` will automatically delete the animation when the animation reaches the set end time. No need to actively call `lv_anim_del` to delete.

kisvegabor:
But if an animation is not deleted in `lv_anim_timeline_stop` it will keep running and it will call its `ready_cb`. However, the user intention was to make all animations stop.

FASTSHIFT:
You are right. I thought of a solution: When it is detected that `var` and `exec_cb` are NULL, use the pointer of `lv_anim_timeline` as `var`, so that the animation can be completely deleted when `lv_anim_timeline_stop` is called.

kisvegabor:
Looks good, thank you!

PR #2649, opened 2021-10-09
Reduce the stack usage in log function

xiaoxiang781216:

- feat(printf): support %pV format specifier
- fix(log): remove 768B temp buffer if LV_LOG_PRINTF == 1
- fix(log): save 256B temp buffer if LV_LOG_PRINTF == 0

### Checkpoints

### DIFF
--- a/src/misc/lv_log.c
+++ b/src/misc/lv_log.c
@@ -72,9 +72,6 @@ void _lv_log_add(lv_log_level_t level, const char * file, int line, const char *
     if(level >= LV_LOG_LEVEL) {
         va_list args;
         va_start(args, format);
-        char msg[256];
-        lv_vsnprintf(msg, sizeof(msg), format, args);
-        va_end(args);
 
         /*Use only the file name not the path*/
         size_t p;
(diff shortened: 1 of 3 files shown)


kisvegabor:
Shouldn't lv_log.c be also updated?

xiaoxiang781216:
> Shouldn't lv_log.c be also updated?

Yes, it's in the next PR. I can't add it in this PR because MicroPython CI complain lv_format_t can't be found. I guess that CI can't support the dependent patches in one PR?

kisvegabor:
CI should handle if a problem is fixed in a new commit because it always runs on the last commit. Please, push it here to see what happens.

xiaoxiang781216:
OK, let's what happen now.

xiaoxiang781216:
Here is the error(https://github.com/lvgl/lvgl/pull/2649/checks?check_run_id=3859062425):
```
Error: ../../lib/lv_bindings/lvgl/src/misc/lv_log.c:84:9: error: unknown type name ‘lv_vaformat_t’
   84 |         lv_vaformat_t vaf = {format, &args};
      |         ^~~~~~~~~~~~~
```

kisvegabor:
It fails because the [Micropython binding comes](https://github.com/lvgl/lv_binding_micropython/blob/af7522d7314fc785a72f7de0640716be239e8b37/lv_conf.h#L239) with `LV_SPRINTF_CUSTOM` enabled and `lv_vaformat_t` is typedfed in a `#if LV_SPRINTF_CUSTOM == 0` block.

So `lv_log.c` should support standard `sprintf` too.

xiaoxiang781216:
Ok, look like I need find a better solution to support both case.

xiaoxiang781216:
@kisvegabor @embeddedt could you review this PR which:

- avoid call fwrite
- try to reduce the stack usage in various case

src/misc/lv_log.c
```diff
@@ -85,23 +82,32 @@ void _lv_log_add(lv_log_level_t level, const char * file, int line, const char *
...
-void lv_log(const char * buf)
-{
 #if LV_LOG_PRINTF
-    fwrite(buf, 1, strlen(buf), stdout);
+        printf("[%s]\t(%" LV_PRId32 ".%03" LV_PRId32 ", +%" LV_PRId32 ")\t %s: ",
+               lvl_prefix[level], t / 1000, t % 1000, t - last_log_time, func);
+        vprintf(format, args);
+        printf(" \t(in %s line #%d)\n", &file[p], line);
+#else
+        char buf[512];
```
kisvegabor: The whole block from here can be in `if(custom_print_cb) { ... }` to avoid allocating `buf` and `msg` if there is no `custom_print_cb`.

src/misc/lv_log.c
```diff
@@ -85,23 +82,32 @@ void _lv_log_add(lv_log_level_t level, const char * file, int line, const char *
...
+        lv_snprintf(buf, sizeof(buf), "[%s]\t(%" LV_PRId32 ".%03" LV_PRId32 ", +%" LV_PRId32 ")\t %s: %pV \t(in %s line #%d)\n",
+                    lvl_prefix[level], t / 1000, t % 1000, t - last_log_time, func, &vaf, &file[p], line);
+#endif
+        if(custom_print_cb) custom_print_cb(buf);
 #endif
-    if(custom_print_cb) custom_print_cb(buf);
 
+        last_log_time = t;
+        va_end(args);
+    }
```
kisvegabor, line 112: Why was `lv_log(const char * buf)` removed?

xiaoxiang781216: Restore

src/misc/lv_log.c
xiaoxiang781216: Done.

kisvegabor:
Merging, thanks!

PR #2655, opened 2021-10-11
fix(log): replace printf with fwrite to save the stack size

xiaoxiang781216:

since fwrite consume much small stack than printf

### Checkpoints

### DIFF
--- a/src/misc/lv_log.c
+++ b/src/misc/lv_log.c
@@ -98,7 +98,7 @@ void _lv_log_add(lv_log_level_t level, const char * file, int line, const char *
 void lv_log(const char * buf)
 {
 #if LV_LOG_PRINTF
-    printf("%s", buf);
+    fwrite(buf, 1, strlen(buf), stdout);
 #endif
     if(custom_print_cb) custom_print_cb(buf);
 


kisvegabor:
Thanks

embeddedt:
Note that there are probably some projects which have `printf` but not `fwrite` (I can think of a few of my own). This would break out-of-the-box compatibility on those systems. I don't know if it matters, since one can always register a custom print callback, but it does add one more hurdle to porting.

xiaoxiang781216:
or we call puts?

embeddedt:
That's a great idea, actually. I know GCC will often optimize a `printf` call like this one into a `puts` call anyway, so it's very likely that `puts` would be available.

xiaoxiang781216:
If the user enable printf, an other solution is directly call printf piece by piece in _lv_log_add which could save both stack and code space.

kisvegabor:
> If the user enable printf, an other solution is directly call printf piece by piece in _lv_log_add which could save both stack and code space.

Good idea, but can we sure that `vsnprintf` (required to create `msg`) is available on all platforms where there is `printf`?
It might be a reasonable to assume though. If in rare cases it's missing the user still can register a `print_cb` a let LVGL compose the pieces of the log message.

xiaoxiang781216:
> > If the user enable printf, an other solution is directly call printf piece by piece in _lv_log_add which could save both stack and code space.
> 
> Good idea, but can we sure that `vsnprintf` (required to create `msg`) is available on all platforms where there is `printf`? It might be a reasonable to assume though. If in rare cases it's missing the user still can register a `print_cb` a let LVGL compose the pieces of the log message.

Here is my plan:

1. Call printf in _lv_log_add directly if user enable LV_LOG_PRINTF, so we save both lv_snprintf code space and two big stack buffer
2. Call lv_snprintf with %pV enhancement if user disable LV_LOG_PRINTF, so we can save 256B stack buffer

xiaoxiang781216:
Fixed in this https://github.com/lvgl/lvgl/pull/2655

PR #2656, opened 2021-10-11
Allow CMake users to modify LV_LVGL_H_INCLUDE

higaski:
This PR allows CMake users to specify the preprocessor define LV_LVGL_H_INCLUDE_SIMPLE by using an option with the same name. Furthermore another option called LV_LVGL_H_INCLUDE_PATH can be used to specify the location where the "lvgl.h" header file can be found. This allows arbitrary folder layouts without any assumptions about the users own folder structures.

Example usage
```CMake
set(LV_LVGL_H_INCLUDE_SIMPLE ON CACHE BOOL "" FORCE)
set(LV_LVGL_H_INCLUDE_PATH ${CMAKE_SOURCE_DIR}/config CACHE STRING "" FORCE)
FetchContent_Declare(lvgl GIT_REPOSITORY https://github.com/lvgl/lvgl.git)
FetchContent_MakeAvailable(lvgl)
```

In case everyone agrees with those changes I'd also create PRs for the [lv_drivers](https://github.com/lvgl/lv_drivers) (which has the option LV_CONF_INCLUDE_SIMPLE) and [lv_demos](https://github.com/lvgl/lv_demos) (LV_DEMO_CONF_INCLUDE_SIMPLE)  repos.

We should maybe also consider to create a cmake folder in root for further cleanup. Currently there are 4x different configurations in a single file which have (almost) nothing in common:
- ESP
- Zephyr
- Micropython
- "Normal"

For regular code we'd probably call this a violation of the SRP. ;)

Personally I also think that the hardcoded locations of "lv_conf.h" and "lvgl.h" in the installation part of the current CMakeLists.txt file aren't a good idea. I have not touched that part yet because I'm not willing to break someones else's build (specially if this has something to do with [integration to Yocto](https://github.com/lvgl/lvgl/issues/2534)), but there definitely should be some discussion about this.

### Checkpoints

/edit
PS: I also propose using [cmake-format](https://github.com/cheshirekow/cmake_format). For this PR I tried to keep to the existing style, but I'm always in favor of auto-formatting if there are tools out there.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -88,8 +88,16 @@ else()
         # default linux build uses this scope
         add_library(lvgl STATIC ${SOURCES})
         add_library(lvgl_examples STATIC ${EXAMPLE_SOURCES})
+        add_library(lvgl::lvgl ALIAS lvgl)
+        add_library(lvgl::examples ALIAS lvgl_examples)
 
-        include_directories(${CMAKE_SOURCE_DIR})
+        # Allow users to define their own path to lvgl.h
+        option(LV_LVGL_H_INCLUDE_SIMPLE "Use #include \"lvgl.h\" instead of #include \"../../lvgl.h\"")
+        option(LV_LVGL_H_INCLUDE_PATH "Include path of folder containing lvgl.h")
+
+        target_compile_definitions(lvgl PUBLIC $<$<BOOL:${LV_LVGL_H_INCLUDE_SIMPLE}>:LV_LVGL_H_INCLUDE_SIMPLE>)
+        target_include_directories(lvgl SYSTEM PUBLIC ${CMAKE_CURRENT_SOURCE_DIR} $<$<BOOL:${LV_LVGL_H_INCLUDE_SIMPLE}>:${LV_LVGL_H_INCLUDE_PATH}>)
+        target_link_libraries(lvgl_examples PUBLIC lvgl)
 
         # Lbrary and headers can be installed to system using make install
         file(GLOB LVGL_PUBLIC_HEADERS


kisvegabor:
Hi,

Thank your for this improvement.

I'm not good enough with Cmake to tell what could be best direction, so I cc some people how know it much better than me :slightly_smiling_face:  
@embeddedt @amirgon @cmumford @rzr

ValentiWorkLearning:
Let's refactor CMake library sources collection with this PR also. Probably, we can get rid off FILE GLOB and use the target approach instead.

C47D:
This is also a great opportunity to also improve the lvgl_esp32_drivers and lv_port_esp32 cmake files learning from y'all.

higaski:
> Let's refactor CMake library sources collection with this PR also. Probably, we can get rid off FILE GLOB and use the target approach instead.

I've thought about that as well and I absolutely agree that it would be the cleaner CMake approach. The problem is though that it kinda contradicts with all the _LV_USE_*_ macros inside the _lv_conf.h_ file. Those macros are currently used to enable/disable libraries from the config file (e.g. `#define LV_USE_BMP 0`). Since this is happening in the preprocessor phase it's too late to exclude library sources from the build. The compiler might get a more or less empty source file if the corresponding library is disabled inside _lv_conf.h_  but it gets the source file nonetheless.

I see no benefits from splitting the lvgl CMake target as long as conditional compilation is handled through the configuration header. I'd even argue that the way it is handled now benefits a library mostly targeting embedded system programmers. Having a configuration header to enable/disable libraries works no matter if you copy/paste the source into your own project or if you're having some fancy CMake build.

cmumford:
I don't see a problem with this patch - caveat: I'm no CMake expert.

However, I've been thinking that the way LVGL treats it's headers is overly complex, and I've often fanaticized about simplifying this process. In my fantasy, `LV_LVGL_H_INCLUDE_SIMPLE` would be deleted. I **don't want to derail** the discussion in this PR, so I've opened up a new issue (https://github.com/lvgl/lvgl/issues/2671). Please feel free to shoot down my proposal there.

higaski:
Is there any chance this is going to move forward? If not it would be fine with me as well, but please don't let it hang in limbo.

JojoS62:
@higasi do you have an usage example? I don't have problems with lvgl.h, but with lv_conf.h. I want to have lvgl and some lvglDriver outside my project as common lib and have also set LV_CONF_INCLUDE_SIMPLE

```
        add_subdirectory(${LIB_PATH}lvgl ${CMAKE_CURRENT_BINARY_DIR}/lvgl)
        target_compile_definitions(lvgl PRIVATE LV_CONF_INCLUDE_SIMPLE)
        target_include_directories(lvgl PRIVATE ${CMAKE_CURRENT_SOURCE_DIR})
        target_compile_options(lvgl PRIVATE -Os)

       add_subdirectory(${LIB_PATH}lvglDriver ${CMAKE_CURRENT_BINARY_DIR}/lvglDriver)
       target_compile_definitions(lvglDriver PRIVATE LV_CONF_INCLUDE_SIMPLE)
       target_compile_definitions(lvglDriver PRIVATE LV_CONF_INCLUDE_SIMPLE)
```
now compilation is stuck at lv_examples (which should be optional). When I comment the lv_example creation, then it works.

higaski:
Well sadly no. I have posted an explanation in issue #2671 and what I think needs to be done. But without some sort of "OK" that those changes are welcome I'm reluctant to work on it. Generally I think that this whole CMake issue needs some more discussion... There seems to be a lot of confusion on this topic and I know from experience that CMake is mostly anarchy.

Now to your problem. I also set LV_CONF_INCLUDE_SIMPLE (and LV_DEMO_CONF_INCLUDE_SIMPLE for that matter) in my builds. To make all those things work together I currently use my own Find*.cmake files which define proper targets with dependencies to lvgl.
[find_cmake.zip](https://github.com/lvgl/lvgl/files/7400237/find_cmake.zip)

Notice that these Find*.cmake files only work because I've included the lvgl root directory for the lvgl target.
```CMake
target_include_directories(lvgl SYSTEM PUBLIC ${CMAKE_CURRENT_SOURCE_DIR})
```

The CMakeLists.txt file of the current master branch does not do this.

How exactly lv_examples (you ment lvgl_examples I guess?) should create a problem for you is beyond me. Even if the target configuration is somehow flawed it wouldn't fail at compilation unless you link with lvgl_examples.

JojoS62:
>  I also set LV_CONF_INCLUDE_SIMPLE (and LV_DEMO_CONF_INCLUDE_SIMPLE for that matter)

yes, that is working now.
```
        add_subdirectory(${LIB_PATH}lvgl lvgl)
        target_compile_options(lvgl PRIVATE -Os)
        target_compile_definitions(lvgl PRIVATE LV_CONF_INCLUDE_SIMPLE )
        target_compile_definitions(lvgl_examples PRIVATE  LV_CONF_INCLUDE_SIMPLE LV_DEMO_CONF_INCLUDE_SIMPLE)
        target_include_directories(lvgl PRIVATE ${CMAKE_CURRENT_SOURCE_DIR})
```

rzr:
something that would be useful too is to generate .h file using cmake params to override default values.
Today I am using sed tricks to change allocator and wayland resolutions at build time.

higaski:
Closed until some sort of consensus is reached on this topic.

JojoS62:
I think this PR should be merged. It is a versatile solution for cmake based projects and does not affect existing solutions.
If something may change, a label ‚next Release‘ can announce that it it will be merged later or in a pre release branch for testing.

higaski:
I'm honestly not sure about this PR myself anymore. I've posted my concerns in #2671.

TL;DR
In my opinion the current default of having LV_LVGL_H_INCLUDE undefined and therefor use `#include "lvgl/lvgl.h"` isn't a sane CMake option. There is no guarantee that this folder will exist whatsoever. It's not within the current CMakeLists.txt files own scope. How should CMake handle this case then? In my opinion CMake absolutely must default define LV_LVGL_H_INCLUDE.

For LV_CONF_INCLUDE_SIMPLE and LV_DEMO_CONF_INCLUDE_SIMPLE things get simpler. I've just realized that there are LV_CONF_PATH and LV_DEMO_CONF_PATH definitions which can be uses to point to specific locations. This could be directly translated to CMake.

JojoS62:
Yes, lvgl.h should be the include to avoid assumptions about a directory structure. If this can be set by a define, it’s ok and doesn’t break other solutions.
But the target_include_directories should be added to make the configuration for an app easier and safer for future changes.

PR #2657, opened 2021-10-11
fix(obj):use content coords as draw mask of child obj

guoweilkd:

According to the box model， The drawing field of the child obj should be the content field of the parent obj.

### Checkpoints


kisvegabor:
I believe in HTML it's not true. See https://codepen.io/kisvegabor/pen/RwZNXop

PR #2661, opened 2021-10-12
fix(lv_png) fix compile error when LV_COLOR_DEPTH == 8

calumr:

lv_png fails to compile if LV_COLOR_DEPTH == 8.

### Checkpoints

### DIFF
--- a/src/extra/libs/png/lv_png.c
+++ b/src/extra/libs/png/lv_png.c
@@ -225,9 +225,9 @@ static void convert_color_depth(uint8_t * img, uint32_t px_cnt)
        lv_color_t c;
        uint32_t i;
        for(i = 0; i < px_cnt; i++) {
-           c = lv_color_make(img_argb[i].red, img_argb[i].green, img_argb[i].blue);
-           img[i*2 + 1] = img_argb[i].alpha;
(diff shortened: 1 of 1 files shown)


embeddedt:
@kisvegabor It seems that we need to add the extra libraries to the CMakeLists file. I suspect they're not being checked by CI, otherwise this should have been caught in your original PR.

kisvegabor:
@embeddedt It should gather all the C files, shouldn't it? 
https://github.com/lvgl/lvgl/blob/master/CMakeLists.txt#L41

embeddedt:
Hmm, that's true. Strangely the original code seems to be compiling [on `master`](https://github.com/lvgl/lvgl/runs/3870449756#step:5:119) without issues.

It almost looks like our CI is treating the struct as if it's an anonymous union, where you can access the members as though they're part of the parent.

calumr:
I think LV_USE_PNG isn't turned on for the 8bpp test?

kisvegabor:
@calumr 
Yes, that was it. I've added all 3rd party libs to all build tests.

As there were similar bugs in the other libs too I fixed them and this one as well.

PR #2662, opened 2021-10-13
Sync Kconfig with the latest lv_conf_template.h

xiaoxiang781216:

and fix the typo and sequence issue.
- fix(conf): correct LV_USE_EXTERNAL_RENDERER and LV_USE_GPU_SDL related setting
- fix(Kconfig): typo error in the font and theme related setting
- fix(Kconfig): Add the missing misc config
- fix(Kconfig): sync widget and theme in Kconfig with lv_conf_template.h
- fix(Kconfig): add the option for 3rd party library and examples

### Checkpoints


kisvegabor (approved):
Looks good to me!

kisvegabor:
@C47D, could you also take a look?

kisvegabor:
@xiaoxiang781216 
f8fe5366b also affects lv_conf_internl.h and Kconfig. 

I needed to revert it due to some unexpected issues. I will investigate tomorrow.

C47D:
@kisvegabor I think you missed the `CONFIG_x` prefix when checking `LV_COLOR_DEPTH` and defining `LV_KCONFIG_PRESENT`

xiaoxiang781216:
> @kisvegabor I think you missed the `CONFIG_x` prefix when checking `LV_COLOR_DEPTH` and defining `LV_KCONFIG_PRESENT`

Why not check LV_CONF_SKIP directly which is more natural than LV_COLOR_DEPTH?
BTW, this approach has a bad side effect that the update of Kconfig is normally behind the lv_conf_template.h with this approach we lose the default config set by lv_conf_template.h.

xiaoxiang781216:
How about we change bool to int with range 0 to 1? So Kconfig generator never remove false(0)  setting from config.h.

xiaoxiang781216:
> @xiaoxiang781216 [f8fe536](https://github.com/lvgl/lvgl/commit/f8fe5366bb051cd5090e4a06658eb0d32decc0b3) also affects lv_conf_internl.h and Kconfig.
> 
> I needed to revert it due to some unexpected issues. I will investigate tomorrow.

It's more simpler to merge my patch first and then fixing bool issue together, I think.

C47D:
I think the same.

One point I was being aware of, is using lvgl in Kconfig capable systems using `lv_config.h` possible or something we should have?

kisvegabor:
> Why not check LV_CONF_SKIP directly which is more natural than LV_COLOR_DEPTH?

> One point I was being aware of, is using lvgl in Kconfig capable systems using lv_config.h possible or something we should have?

It's possible to use lv_conf.h and Kconfig at the same time. E.g. Zephyr sets `LV_MEM_CUSTOM_ALLOC/FREE` in lv_conf.h as they can't be set in Kconfig. In this case lv_conf.h can be almost empty. 


> How about we change bool to int with range 0 to 1? So Kconfig generator never remove false(0) setting from config.h.


It's fine for me.

Issue #2668, opened 2021-10-15
arc obj:color display error when color opa was used


guoweilkd:
### Describe the bug

arc color display error when arc color opa was used: like this:
```c
    lv_obj_t * arc = lv_arc_create(lv_scr_act());
    lv_obj_remove_style(arc, NULL, LV_PART_KNOB);
    lv_obj_set_style_arc_color(arc, lv_palette_main(LV_PALETTE_RED), LV_PART_MAIN);
    lv_obj_set_style_arc_opa(arc, LV_OPA_10, LV_PART_MAIN);
    lv_obj_set_style_arc_color(arc, lv_palette_main(LV_PALETTE_BLUE), LV_PART_INDICATOR);
    lv_obj_set_style_arc_opa(arc, LV_OPA_50, LV_PART_INDICATOR);

    lv_obj_set_size(arc, 150, 150);
    lv_arc_set_rotation(arc, 135);
    lv_arc_set_bg_angles(arc, 0, 270);
    lv_arc_set_value(arc, 40);
    lv_obj_center(arc);
```
The result:
![image](https://user-images.githubusercontent.com/35251456/137432588-fe7188a5-49a8-4264-a318-7262f3f544fc.png)

The reason for this is the overlapping blending that occurs when drawing rounded corners. I made some fixes, see #2669 , But the results are still not good. The result after repair is as follows:
![image](https://user-images.githubusercontent.com/35251456/137433293-1f0ac98e-741e-42a6-8ade-ac54063baa35.png)

There will still be color errors at both ends of the angle mask.I found some clues in [lv_draw_mask_angle](https://github.com/lvgl/lvgl/blob/master/src/draw/lv_draw_mask.c#L893~#L990).The point (abs_y == p->cfg.vertex_p.y) may or may not contains  to the mask, We should make it consistent, either include it or not. besides, we need to be able to control whether edge points are included in the mask.

kisvegabor:
Thanks, let's continue in the PR.

PR #2669, opened 2021-10-15
fix(arc):color display error when color opa was used

guoweilkd:

see #2668

### Checkpoints

### DIFF
--- a/src/draw/lv_draw_arc.c
+++ b/src/draw/lv_draw_arc.c
@@ -173,6 +173,10 @@ void lv_draw_arc(lv_coord_t center_x, lv_coord_t center_y, uint16_t radius,  uin
     if(mask_in_id != LV_MASK_ID_INV) lv_draw_mask_remove_id(mask_in_id);
 
     if(dsc->rounded) {
+        if (cir_dsc.bg_opa < LV_OPA_COVER) {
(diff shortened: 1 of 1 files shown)


kisvegabor:
It's a good idea but it can always happen that 2 anti aliased edges doesn't cover each other perfectly. 

Imagine this: 
- The arc has a pixel with 50% opacity on the edge. 
- The arc is black, the background is white so the result will be 0x808080. 
- If all calculation are accurate and the mask tells the the circle also has 50% opacity at the same point, now 0x808080 will be mixed with black resulting in 0x404040. However we expected pure black at the joint of the circle and the arc. 


I was also thinking about this issue earlier and find a different approach. Now the masks are "destructive". It means that they can only remove parts. But with a "builder" type we could build the back the circles.  What do you think?

guoweilkd:
> But with a "builder" type we could build the back the circles
A good idea， Is there a specific release date for the function?

kisvegabor:
Unfortunately I have no time to add it now, but I added it to my todo list for v8.2.
(I'm planning to release v8.1 next week.)

embeddedt:
It sounds like this PR was ultimately not the best approach and it has conflicts; can it be closed?

kisvegabor:
I think yes. I'm thinking about how to fix it whit the new draw architecture.

PR #2673, opened 2021-10-15
Improve LV_FORMAT_ATTRIBUTE usage

xiaoxiang781216:

- fix(format): add LV_FORMAT_ATTRIBUTE to lv_snprintf and lv_log 
- fix(format): remove LV_FORMAT_ATTRIBUTE from the function body
- fix(log): always reference the arguments in all log level

### Checkpoints

### DIFF
--- a/src/misc/lv_log.h
+++ b/src/misc/lv_log.h
@@ -16,6 +16,8 @@ extern "C" {
 #include "../lv_conf_internal.h"
 #include <stdint.h>
 
+#include "lv_types.h"
+
 /*********************
  *      DEFINES
  *********************/
@@ -77,7 +79,8 @@ void lv_log(const char * buf);
  * @param format    printf-like format string
  * @param ...       parameters for `format`
  */
-void _lv_log_add(lv_log_level_t level, const char * file, int line, const char * func, const char * format, ...);
+void _lv_log_add(lv_log_level_t level, const char * file, int line,
+                 const char * func, const char * format, ...) LV_FORMAT_ATTRIBUTE(5, 6);
 
 /**********************
  *      MACROS
--- a/src/misc/lv_mem.c
+++ b/src/misc/lv_mem.c
@@ -63,7 +63,7 @@ static uint32_t zero_mem = ZERO_MEM_SENTINEL; /*Give the address of this variabl
  *      MACROS
  **********************/
 #if LV_LOG_TRACE_MEM
-    #define MEM_TRACE(...) LV_LOG_TRACE( __VA_ARGS__)
+    #define MEM_TRACE(...) LV_LOG_TRACE(__VA_ARGS__)
 #else
     #define MEM_TRACE(...)
 #endif
@@ -122,7 +122,7 @@ void lv_mem_deinit(void)
  */
 void * lv_mem_alloc(size_t size)
 {
-    MEM_TRACE("allocating %d bytes", size);
+    MEM_TRACE("allocating %lu bytes", (unsigned long)size);
     if(size == 0) {
         MEM_TRACE("using zero_mem");
         return &zero_mem;
@@ -139,7 +139,7 @@ void * lv_mem_alloc(size_t size)
 #endif
 
     if(alloc == NULL) {
-        LV_LOG_ERROR("couldn't allocate memory (%d bytes)", size);
+        LV_LOG_ERROR("couldn't allocate memory (%lu bytes)", (unsigned long)size);
         lv_mem_monitor_t mon;
         lv_mem_monitor(&mon);
         LV_LOG_ERROR("used: %6d (%3d %%), frag: %3d %%, biggest free: %6d",
@@ -180,7 +180,7 @@ void lv_mem_free(void * data)
  */
 void * lv_mem_realloc(void * data_p, size_t new_size)
 {
-    MEM_TRACE("reallocating %p with %d size", data_p, new_size);
+    MEM_TRACE("reallocating %p with %lu size", data_p, (unsigned long)new_size);
     if(new_size == 0) {
         MEM_TRACE("using zero_mem");
         lv_mem_free(data_p);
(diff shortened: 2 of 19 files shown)


src/widgets/lv_label.c
```diff
@@ -144,7 +144,7 @@ void lv_label_set_text(lv_obj_t * obj, const char * text)
     lv_label_refr_text(obj);
 }
 
-LV_FORMAT_ATTRIBUTE(2, 3) void lv_label_set_text_fmt(lv_obj_t * obj, const char * fmt, ...)
+void lv_label_set_text_fmt(lv_obj_t * obj, const char * fmt, ...)
```
kisvegabor, line 147: Why was it removed from here?

src/misc/lv_mem.c
```diff
@@ -139,7 +139,7 @@ void * lv_mem_alloc(size_t size)
 #endif
 
     if(alloc == NULL) {
-        LV_LOG_ERROR("couldn't allocate memory (%d bytes)", size);
+        LV_LOG_ERROR("couldn't allocate memory (%zu bytes)", size);
```
kisvegabor: Can we be sure the `zu` is well supported?

E.g. Visual studio doesn't support it: [LINK](https://stackoverflow.com/questions/15610053/correct-printf-format-specifier-for-size-t-zu-or-iu)

kisvegabor (commented):
Thanks, only a few questions.

src/misc/lv_mem.c
xiaoxiang781216: > Can we be sure the `zu` is well supported?
> 

zu is defined by C99. Since LVGL already require the compiler newer than C99, it isn't a new reqiurement.

> E.g. Visual studio doesn't support it: [LINK](https://stackoverflow.com/questions/15610053/correct-printf-format-specifier-for-size-t-zu-or-iu)

zu is supported from VS2013, do we want to support the old version of VS? If need, the code has to add a cast to unsigned long.

BTW, it look like CI for STM32 is broken:
https://github.com/lvgl/lvgl/pull/2673/checks?check_run_id=3909646994

src/widgets/lv_label.c
xiaoxiang781216: Since I never saw any libc implementation add format attribute to function body. Here is the example from glibc:
https://github.com/bminor/glibc/blob/master/include/stdio.h#L42-L44
https://github.com/bminor/glibc/blob/master/stdio-common/snprintf.c#L21-L25

kisvegabor: So it's enough to keep it in the header. Alright.

src/misc/lv_mem.c
kisvegabor: I'd rather `cast` here instead of introducing a possible problem for such a minor gain.

kisvegabor: > BTW, it look like CI for STM32 is broken:
https://github.com/lvgl/lvgl/pull/2673/checks?check_run_id=3909646994

cc @amirgon @embeddedt

xiaoxiang781216: > I'd rather `cast` here instead of introducing a possible problem for such a minor gain.

Done.

src/misc/lv_log.h
```diff
@@ -77,52 +79,43 @@ void lv_log(const char * buf);
...
-#define LV_LOG_TRACE(...) do {}while(0)
-#define LV_LOG_INFO(...) do {}while(0)
-#define LV_LOG_WARN(...) do {}while(0)
-#define LV_LOG_ERROR(...) do {}while(0)
-#define LV_LOG_USER(...) do {}while(0)
-#endif /*LV_USE_LOG*/
+#define LV_LOG_TRACE(...)                                                               \
+    do {                                                                                \
+        if (LV_LOG_LEVEL <= LV_LOG_LEVEL_TRACE)                                         \
+            _lv_log_add(LV_LOG_LEVEL_TRACE, __FILE__, __LINE__, __func__, __VA_ARGS__); \
```
kisvegabor: What is the advantage of this change?

I'm not sure the compiler will drop the log messages if the log level is larger.

xiaoxiang781216: > What is the advantage of this change?
> 

Ensure format string get check even the log/assert turn off, you can see the change catch some wrong usage in PR.

> I'm not sure the compiler will drop the log messages if the log level is larger.

The modern compiler should be enough smart to remove the dead code when the optimization is enabled.

kisvegabor: > The modern compiler should be enough smart to remove the dead code when the optimization is enabled.


Finally it'll be something like:
```c
if(3 <= 4) { _lv_log_add(...)}
```

Seems really straight forward to drop. I cc @embeddedt to hear an other opinion too.

embeddedt: In my personal opinion, independently of the issue of dead code elimination, the previous version of the header was easier to read and understand as the macro definitions were a lot shorter. I'm not sure it's worth making it more complicated just for the format string to be checked all the time. I think logging & assertions are enabled on the CI, so any issues in PRs should be getting caught by that, right?

However, that's just my opinion.

kisvegabor:
> I think logging & assertions are enabled on the CI, so any issues in PRs should be getting caught by that, right?

Not all levels and tracing now, but we can change it.

xiaoxiang781216:
If you don't like to always check the format string, let me remove them from the patchset. But, I always prefer catch the potential issue as soon as possible:).

embeddedt:
>  I always prefer catch the potential issue as soon as possible:).

I completely agree. I just think it might be cleaner to make sure the CI has logging enabled instead of introducing the dead-code elimination issue and growing the macro.

kisvegabor:
I've pushed a fix to run one of the tests with `LV_LOG_LEVEL_TRACE` but it seems CI doesn't catch the warnings fixed in this PR.

src/misc/lv_log.h
xiaoxiang781216: Ok, I remove the change of trace macro from the patch, please take a look.

kisvegabor:
Looks good, thanks!

PR #2674, opened 2021-10-16
feat(rt-thread): implement rt-thread sconscript

mysterywolf:

Hi,

Sorry for the late second PR, because these days I dived into the task of RT-Thread new version release.

This PR implements the Sconscript which is very essential to let RT-Thread to organize the file structure of LVGL.
Meanwhile, there is a **good news**: LVGL has been registered successfully in the RT-Thread software packages center and every RT-Thread user can search the LVGL official package in the RT-Thread official website. https://packages.rt-thread.org/en/detail.html?package=lvgl

I still need to several PRs to complete this RT-Thread port. Thanks ahead for your PRs reviewing. 

If you want to release new version in recent days, just keep your original plan. Don't be waiting for my part.

Meco

### Checkpoints

### DIFF
--- a/rt-thread/SConscript
+++ b/rt-thread/SConscript
@@ -1,15 +1,61 @@
-# RT-Thread building script for bridge
-
-import os
 from building import *
+import rtconfig
+import os
+import shutil
 
+# get current dir path
 cwd = GetCurrentDir()
-objs = []
-list = os.listdir(cwd)
 
+src = []
+inc = []
+LOCAL_CCFLAGS = ''
+
+lvgl_cwd = cwd + '/../'
+lvgl_src_cwd = lvgl_cwd + 'src/'
+
+src = src + Glob('*.c')
+inc = inc + [cwd]
+inc = inc + [lvgl_src_cwd]
+
+for root, dirs, files in os.walk(lvgl_src_cwd):
+    for dir in dirs:
+        src = src + Glob(os.path.join(root,dir,'*.c'))
+        inc = inc + [os.path.join(root,dir)]
+
+if rtconfig.CROSS_TOOL == 'gcc':
+    LOCAL_CCFLAGS += ' -std=c99'
+elif rtconfig.CROSS_TOOL == 'keil':
+    LOCAL_CCFLAGS += ' --c99 --gnu -g -W'
+
+group = DefineGroup('LVGL', src, depend = ['PKG_USING_LVGL'], CPPPATH = inc, LOCAL_CCFLAGS = LOCAL_CCFLAGS)
+
+list = os.listdir(cwd)
 for d in list:
     path = os.path.join(cwd, d)
     if os.path.isfile(os.path.join(path, 'SConscript')):
-        objs = objs + SConscript(os.path.join(d, 'SConscript'))
+        group = group + SConscript(os.path.join(d, 'SConscript'))
 
-Return('objs')
+#
+# try:
+#     shutil.rmtree(os.path.join(lvgl_cwd, '.github'))
+#     shutil.rmtree(os.path.join(lvgl_cwd, 'docs'))
+#     shutil.rmtree(os.path.join(lvgl_cwd, 'scripts'))
+#     shutil.rmtree(os.path.join(lvgl_cwd, 'tests'))
+#     shutil.rmtree(os.path.join(lvgl_cwd, 'zephyr'))
+#     os.remove(os.path.join(lvgl_cwd, '.codecov.yml'))
+#     os.remove(os.path.join(lvgl_cwd, '.editorconfig'))
+#     os.remove(os.path.join(lvgl_cwd, '.gitignore'))
+#     os.remove(os.path.join(lvgl_cwd, '.gitmodules'))
+#     os.remove(os.path.join(lvgl_cwd, 'CMakeLists.txt'))
+#     os.remove(os.path.join(lvgl_cwd, 'component.mk'))
+#     os.remove(os.path.join(lvgl_cwd, 'idf_component.yml'))
+#     os.remove(os.path.join(lvgl_cwd, 'Kconfig'))
+#     os.remove(os.path.join(lvgl_cwd, 'library.json'))
+#     os.remove(os.path.join(lvgl_cwd, 'library.properties'))
+#     os.remove(os.path.join(lvgl_cwd, 'lv_conf_template.h'))
+#     os.remove(os.path.join(lvgl_cwd, 'lvgl.mk'))
+# except:
+#     pass
+#
+
+Return('group')
--- a/rt-thread/lv_rt_thread_conf.h
+++ b/rt-thread/lv_rt_thread_conf.h
@@ -1,0 +1,104 @@
+/*
+ * Copyright (c) 2006-2021, RT-Thread Development Team
+ *
+ * SPDX-License-Identifier: MIT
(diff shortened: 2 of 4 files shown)


kisvegabor:
Thank you! 

> Meanwhile, there is a good news: LVGL has been registered successfully in the RT-Thread software packages center

Amazing! Are there any public statistics on the packages? E.g. download count? 


--- 

At first look only `LV_MEM_CUSTOM_...` was changed in `rt-thread/lv_conf.h`. If so I suggest either adding this defines into `lv_conf_kconfig.h` or as a build parameter. It avoids duplication and if `rt-thread/lv_conf.h` will be included somewhere you won't be able to overwrite the values with Kconfig. Take a look at [lv_conf_internal.h](https://github.com/lvgl/lvgl/blob/master/src/lv_conf_internal.h). It's automatically generated to handle undefined configs. It either uses the value from Kconfig or to a default value.

mysterywolf:
image


Hi, Here is the download statistics.
Because I have not finished this part, thus, RT-Thread has not used official social media to announce this. Once I finished , they will use Twitter, Facebook, Youtube, Linkedin to announce this new LVGL official software to all of the RT-Thread users.

https://www.youtube.com/channel/UCdDHtIfSYPq4002r27ffqPw

https://twitter.com/rt_thread

https://www.facebook.com/RT-Thread-IoT-OS-110395723808463

https://www.linkedin.com/company/rt-thread-iot-os/

--------------

In terms of `rt-thread/lv_conf.h`, I considered your suggestions carefully, but I still think I need to keep `rt-thread/lv_conf.h`, because:
- this PR only is a beginning. In the next PRs, I will change them maybe a lot.
- RT-Thread also use Kconfig to config , but the configuration's method is different from that of LVGL. By the way, the Kconfig file will not be put in LVGL repository. RT-Thread has a particular repository to manage all of the software packages at here: https://github.com/RT-Thread/packages/blob/master/README_en.md

kisvegabor:
> Here is the download statistics.

Oh, I missed that, thanks.

> Once I finished , they will use Twitter, Facebook, Youtube, Linkedin to announce this new LVGL official software to all of the RT-Thread users.

It'd be great. We will also post it on Forum, Blog and LinkedIn.

>  this PR only is a beginning. In the next PRs, I will change them maybe a lot.

Okay. In this case I still suggest adding only the defines you have modified and let `lv_conf_internal.h` handle the rest. 

> RT-Thread also use Kconfig to config , but the configuration's method is different from that of LVGL. By the way, the Kconfig file will not be put in LVGL repository. 

Is it possible to reuse lvgl's Kconfig with `source` (or somehow)? IMO it'd be important to avoid maintaining a second Kconfig file too when we add/change configs.

mysterywolf:
> Is it possible to reuse lvgl's Kconfig with source (or somehow)? IMO it'd be important to avoid maintaining a second Kconfig file too when we add/change configs.

Yes, this is a problem.

mysterywolf:
Sorry , gives me a second to change this as you said, I maybe find a better way.

mysterywolf:
Hi,
I changed the structure as following your suggestions. This is much much better. Thank you for insisting your suggestion!!😁

mysterywolf:
I have done. Please review again.

mysterywolf:
Hi, all set. I still keep `thread/lv_conf.h`, but it just contains some configurations that RT-Thread has to change, and rest of them will be taken over by `lv_conf_internal.h`, just like you mentioned.

Meanwhile, RT-Thread can also change these configurations by: `lv_config.h` ->`rtthread.h` -> `rtconfig.h`
`rtconfig.h` is generated by Kconfig of RT-Thread side, which means RT-Thread users can use RT-Thread Kconfig to config the LVGL's features, just like other software packages.


All in all:
`lv_conf.h` will take over the fixed configurations part, which is related to RT-Thread itself and users don't need to change them;
`rtconfig.h` will take over the dynamic configurations part which will be set by users;
`lv_conf_internal.h` will take over the rest of them as default configurations.

Again, thanks for your suggestions!

embeddedt:
Hi @mysterywolf, first of all, thanks for your contribution!

Could you explain a bit how the RT-Thread project structure works for external packages? Maybe it's just me, but it seems a little strange to have a downstream project's full build script/logic in our own repository. Normally I would think that these files would be in your repository and LVGL would either be a submodule or there would be logic to download a specific commit/version at build time.

mysterywolf:
Hi, @embeddedt , the `rt-thread` folder which is in LVGL repository does not only simply contain a script file. Because this PR is just part of them and I have not finished the whole port task yet, it gives you an illusion that rt-thread folder only contains script file. Actually, this folder's main job is to contain the port *.c and *.h files. Just like in Micropython, there are some operating system's port folder, such as Linux, zephyr and so on..

In the next PR, I will take a video to demonstrate how to use LVGL in RT-Thread projects. Because this PR has not merged, it does not work right now. That's why this PR is essential.

rt-thread/lv_conf.h
```diff
@@ -0,0 +1,105 @@
...
+#define LV_CONF_H
+
+#define LV_RTTHREAD_INCLUDE <rtthread.h>
+#include LV_RTTHREAD_INCLUDE
+
+/*====================
+   COLOR SETTINGS
+ *====================*/
+
+#ifdef PKG_LVGL_ENABLE_COLOR_16_SWAP
```
kisvegabor: Why is it needs to be added?

rt-thread/lv_conf.h
```diff
@@ -0,0 +1,105 @@
...
+ *
+ * SPDX-License-Identifier: MIT
+ *
+ * Change Logs:
+ * Date           Author       Notes
+ * 2021-10-15     Meco Man     First version
+ */
+
+#ifndef LV_CONF_H
+#define LV_CONF_H
```
kisvegabor: Could you remove this file differently to avoid possible confusion? E.g. `lv_rt_thread_conf.h` 

Is is possible to add guards like:
```c
#ifdef RT_THREAD
...
#endif //EOF
```

Else if the user tries to find the declaration of e.g `LV_MEM_CUSTOM` the IDE will offer the defines in this file too.

kisvegabor (commented):
Thank you! Looks good, I had only a few comments.

rt-thread/lv_conf.h
mysterywolf: Thank you for your reviewing. 😊
This is needed. It will let RT-Thread users to use RT-Thread Studio IDE to enable some features by using GUI method.

RT-Thread's configuration method is a little bit different from that of LVGL:
- LVGL use 1 to represent enable, 0 to represent disable.
- RT-Thread uses to define this micro to represent enable, not to define to represent disable.

Thus, it needs to do a transition.

rt-thread/lv_conf.h
mysterywolf: Thank you, I revised them.

kisvegabor: Perfect, thank you! :slightly_smiling_face:

rt-thread/lv_conf.h
kisvegabor: All lvgl config needs to be converted to be Rt-thread compatible? 

Anyway, I merge this to let you make test and we can discuss the details in other PRs.

mysterywolf:
Thank you very much!😊

PR #2692, opened 2021-10-18
fix(rt-thread): include the rt-thread configuration header file

mysterywolf:

Hi,

This PR fixed the header file's error.
Meanwhile, the structure is more reasonable. I leave `<lv_config.h>` to users so that they can add some definitions manually if they want, 

>All lvgl config needs to be converted to be Rt-thread compatible?

which also means the `lv_rt_thread_conf.h` just needs to transition only a few definitions, so you don't need to worry about this any more.

`lv_config.h` will take over the users' manual settings;
`lv_rt_thread_conf.h` will take over the RT-Thread OS related settings (such as systicks, memory) and users' settings that come from RT-Thread Kconfig side (only a few, not too many);
`lv_conf_internal.h` will handle the rest of them.

Meco

### Checkpoints

### DIFF
--- a/rt-thread/lv_rt_thread_conf.h
+++ b/rt-thread/lv_rt_thread_conf.h
@@ -16,22 +16,11 @@
 #define LV_RTTHREAD_INCLUDE <rtthread.h>
 #include LV_RTTHREAD_INCLUDE
 
-/*====================
-   COLOR SETTINGS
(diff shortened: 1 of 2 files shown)


kisvegabor:
Sounds good :slightly_smiling_face:

Issue #121, opened 2018-02-22
Problem with antialiasing


KarolAbego:
Hello.
I have another problem with antialiasing. When antialiasing is off text in objects e.g buttons is bigger that it should (I'm currently testing library with tutorials that You've provided), so I assume that everything is done well. I've tested this issue with two functions put_pixel wich draws graphics pixel by pixel and draw_lcd wich uses DMA2D and copies buffer into the memory. In both cases result is the same.
When I turn on the antialiasing data on screen corrupts (random lines are visible) but text look better. 
I've been trying to fix that by changing VDB_SIZES as well in internal and external memory with FMA. And it doesn't gave any effects. 
The pictures of my problem are attached here -> [Antialias.zip](https://github.com/littlevgl/lvgl/files/1748222/Antialias.zip)

Thank You in advance.

kisvegabor:
Hi,

Anti aliasing makes font to half sized. So the size changing is normal. Learn more here: https://littlevgl.com/basics#drawing-and-rendering

Have you tried the anti aliased version with pixel based frame buffer write instead of DMA2D?

KarolAbego:
Yes I've tried (also I read this artcile) it but the results are the same as in the pictures that are provided in attachment. Also I've tried to draw simple rectangles etc. and drawing functions works fine. Have you got any idea why it corrupts.

kisvegabor:
I see.

Please attache your disp_flush() function and lv_conf.h

By the way, what about your previous issues? Can you give some feedback in the related topics?

KarolAbego:
Flickering of screen was a hardware problem. But about corrupting data on screen I can't say anything because I think that could be connected with antialiasing problem. 
disp_flush and lv_conf are here -> [disp_flush.zip](https://github.com/littlevgl/lvgl/files/1748331/disp_flush.zip)

kisvegabor:
LV_VDB_SIZE is already in pixel count so you don't need to multiply it by 4

To eliminate some possible issues try this way.

```c
#define LV_VDB_SIZE         (480*10)  /*Size of VDB in pixel count*/
#define LV_VDB_ADR          (0)            /*Place VDB to a specific address (e.g. in external RAM) (0: allocate automatically  into RAM)*/

/* Use two Virtual Display buffers (VDB) parallelize rendering and flushing (optional)
 * The flushing should use DMA to write the frame buffer in the background*/
[...]

PR #2694, opened 2021-10-19
lv_obj_draw_part_dsc_t.text_length added (lv_demos issue #121)

tdjastrzebski:
`lv_obj_draw_part_dsc_t` struct contains a pointer to the text buffer (`text`) used during drawing but no buffer length is available.
Fix: add `text_length` containing buffer length.
See [#121](/lvgl/lv_demos/issues/121)


kisvegabor:
Looks good, thanks!

PR #2703, opened 2021-10-20
fix(kconfig): change CONFIG_LV_THEME_DEFAULT_FONT to CONFIG_LV_FONT_DEFAULT

### LINKED ISSUE #2702: The default font is always lv_font_montserrat_14.
I risk expelled from the community, but this is a matter of principle.
Instead of sleeping peacefully, I searched all night for a reason why I can not choose the default built-in font as it was in version `7.XX`.
Found.

### Introduce the problem
I am unable to choose the default font for a trivial reason.
There is no selection menu for built-in default font. It just doesn't exist physically.
There is selection menu  for default font  for `title` the theme and that's it. Although there is a proud `LV_FONT_DEFAULT` macro in `lvgl\Kconfig`, ostensibly intended to globally select the default font. Not only does this definition of `LV_FONT_DEFAULT` do not work in the` ESP-IDF` environment, `choice LV_FONT_DEFAULT` refers to the default font selection menu for the theme's` title`.

### Examples and cases

In file `lvgl\src\lv_conf_internal.h`
```c
/*Always set a default font*/
#ifndef LV_FONT_DEFAULT // Never defined
#  ifdef CONFIG_LV_FONT_DEFAULT
#    define LV_FONT_DEFAULT CONFIG_LV_FONT_DEFAULT
#  else
#    define LV_FONT_DEFAULT &lv_font_montserrat_14 // Therefore the font is always lv_font_montserrat_14.
#  endif
#endif
```

### Suggested solution
In this community, I am an outcast, on the verge of exile, so I will not even try to offer something as a solution for this problem.

Thanks for reading.

xiaoxiang781216:

so the user can overwrite LV_FONT_DEFAULT

### Checkpoints

### DIFF
--- a/src/lv_conf_kconfig.h
+++ b/src/lv_conf_kconfig.h
@@ -46,8 +46,8 @@ extern "C" {
  *******************/
 
 /**
- * NOTE: In Kconfig instead of `LV_THEME_DEFAULT_FONT`
- *       `CONFIG_LV_THEME_DEFAULT_FONT_<font_name>` is defined
+ * NOTE: In Kconfig instead of `LV_DEFAULT_FONT`
+ *       `CONFIG_LV_FONT_DEFAULT_<font_name>` is defined
  *       hence the large selection with if-s
  */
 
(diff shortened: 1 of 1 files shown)


kisvegabor:
Thank you!

PR #2705, opened 2021-10-20
changes aimed at eliminating compiler -Wformat warnings

tdjastrzebski:
Proposed changes aimed at eliminating bunch of compiler -Wformat warnings like
`warning: format '%d' expects argument of type 'int', but argument 4 has type 'uint32_t' {aka 'long unsigned int'} [-Wformat=]`


kisvegabor:
`%d` is platform dependent and means `int`. So IMO change the variable types to `int` in these problematic cases would be easier.

BTW, which platform do you use? I haven't seen these warnings yet.

tdjastrzebski:
It is GCC compiler for STM32 - probably other 32b compilers complain as well but I do not  know for fact.

kisvegabor:
Probably, I just haven't used enabled logging with GCC on embedded devices. :sweat_smile: 

> So IMO change the variable types to int in these problematic cases would be easier.


What do you think about it?

tdjastrzebski:
On the other hand, as far as I know, using platform-specific types for embedded is generally discouraged but I consider myself C novice, so probably our guess is better than mine. Feel free to just reject it.

kisvegabor:
> On the other hand, as far as I know, using platform-specific types for embedded is generally discouraged

In general I agree with this. However in these examples the problematic variables are only simple counters and have very small value. So an `int` even on 16 bit systems should work well.

mysterywolf:
Hi, @tdjastrzebski you may try to use "%u" to instead "%d", rather than use "%ld" to instead "%d"
%u is for unsigned
%d is for signed

tdjastrzebski:
@mysterywolf, @kisvegabor How about fix like this:
`lv_label_set_text_fmt(label, "%d", v);` -> `lv_label_set_text_fmt(label, "%d", (int)v);`
I mean, casting to (int)

mysterywolf:
@tdjastrzebski Hi, I try to use `unsigned int` to replace `uint32_t`, and use `%u` to replace `%d` to solve this ugly problem. https://github.com/lvgl/lvgl/pull/2722 😂. If my PR has been merged, you may submit and fix again. I try ahead.

PR #2722, opened 2021-10-23
fix compiler warnings due to integer type mismatch

mysterywolf:

This PR will silence the warning:
```
lv_refr.c:283:43: warning: format '%d' expects argument of type 'int', but argument 3 has type 'uint32_t {aka long unsigned int}'

lv_refr.c:283:43: warning: format '%d' expects argument of type 'int', but argument 4 has type 'uint32_t {aka long unsigned int}'
```

This is an awkward situation: some compilers think that uint32_t is `long unsigned int` (which means %lu should be used), and some of those think that uint32_t is `unsigned int` (which means %u should be used).

Thus, I think the best solution is to use `unsigned int` to replace `uint32_t`, and use `%u` to replace `%d`. Every compiler is happy.😂

### Checkpoints

### DIFF
--- a/examples/widgets/table/lv_example_table_2.c
+++ b/examples/widgets/table/lv_example_table_2.c
@@ -86,13 +86,13 @@ void lv_example_table_2(void)
     lv_mem_monitor_t mon2;
     lv_mem_monitor(&mon2);
 
-    uint32_t mem_used = mon1.free_size - mon2.free_size;
+    long unsigned int mem_used = mon1.free_size - mon2.free_size;
 
-    uint32_t elaps = lv_tick_elaps(t);
+    unsigned int elaps = lv_tick_elaps(t);
 
     lv_obj_t * label = lv_label_create(lv_scr_act());
-    lv_label_set_text_fmt(label, "%d items were created in %d ms\n"
-                                  "using %d bytes of memory",
+    lv_label_set_text_fmt(label, "%d items were created in %u ms\n"
+                                  "using %lu bytes of memory",
                                   ITEM_CNT, elaps, mem_used);
 
     lv_obj_align(label, LV_ALIGN_BOTTOM_MID, 0, -10);
--- a/examples/event/lv_example_event_1.c
+++ b/examples/event/lv_example_event_1.c
@@ -5,10 +5,10 @@ static void event_cb(lv_event_t * e)
 {
     LV_LOG_USER("Clicked");
 
-    static uint32_t cnt = 1;
+    static unsigned int cnt = 1;
     lv_obj_t * btn = lv_event_get_target(e);
     lv_obj_t * label = lv_obj_get_child(btn, 0);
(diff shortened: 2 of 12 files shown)


kisvegabor:
I like this solution, thank you!

Could you apply it to the other files too that are mentioned in https://github.com/lvgl/lvgl/pull/2705?

xiaoxiang781216:
> I like this solution, thank you!
> 
> Could you apply it to the other files too that are mentioned in #2705?

we can catch more similar warnings if we enable CI for arm/xtensa target.

kisvegabor:
> we can catch more similar warnings if we enable CI for arm/xtensa target

It'd be great to have ARM/Xtensa CI tests too. Contribution would be welcome on it :slightly_smiling_face:

mysterywolf:
> I like this solution, thank you!
> 
> Could you apply it to the other files too that are mentioned in #2705?

For sure, no problem. 
All set. Please double check.😀

examples/widgets/table/lv_example_table_2.c
```diff
@@ -86,13 +86,13 @@ void lv_example_table_2(void)
     lv_mem_monitor_t mon2;
     lv_mem_monitor(&mon2);
 
-    uint32_t mem_used = mon1.free_size - mon2.free_size;
+    unsigned int mem_used = mon1.free_size - mon2.free_size;
```
embeddedt: @kisvegabor On a 16-bit platform, couldn't this overflow if the heap were large enough?

mysterywolf: I am not familiar with 16-bit platform. I think the memory cannot overflow the max of 16-bit because the address bus of the 16-bit CPU may not handle the space which oversteps 16 bits.

embeddedt: I'm not too familiar with them either, but I believe some have the ability to use 32-bit pointers while still operating with 16-bit registers (and thus `int`s).

mysterywolf: Ohhh, OK I will use `long unsigned int`😂

embeddedt (approved):

kisvegabor:
Thank you!

FASTSHIFT:
@mysterywolf 
I think `%"PRIu32"` should be used for printing of `uint32_t` type (see `inttypes.h` for details).
We should use the standard type provided by `stdint.h`.

mysterywolf:
I think this PR's method is better than both of ours: https://github.com/lvgl/lv_demos/pull/146 . How do you think? By the way, could you please help me to review this PR: https://github.com/lvgl/lvgl/pull/2919 @FASTSHIFT

FASTSHIFT:
> I think this PR's method is better than both of ours: [lvgl/lv_demos#146](https://github.com/lvgl/lv_demos/pull/146) . How do you think? By the way, could you please help me to review this PR: #2919 @FASTSHIFT

I agree with you.

PR #2741, opened 2021-10-28
add custom percent opacity

Miguel0101:
Add custom percent opacity, eg: `lv_opa_pct(45)`.

### DIFF
--- a/src/misc/lv_color.h
+++ b/src/misc/lv_color.h
@@ -458,7 +458,7 @@ static inline uint32_t lv_color_to32(lv_color_t color)
  */
 static inline lv_opa_t lv_opa_pct(lv_opa_t value)
 {
-    return ((value * LV_OPA_COVER) / 100)
+    return ((value * LV_OPA_COVER) / 100);
 }
 
 /**
--- a/src/misc/lv_color.h
+++ b/src/misc/lv_color.h
@@ -57,8 +57,6 @@ enum {
     LV_OPA_COVER  = 255,
 };
 
-#define LV_OPA(opa) ((opa * LV_OPA_COVER) / 100) /* Opacity in percent 0..100*/
-
 #define LV_OPA_MIN 2    /*Opacities below this will be transparent*/
 #define LV_OPA_MAX 253  /*Opacities above this will fully cover*/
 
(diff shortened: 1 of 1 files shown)


kisvegabor:
Good idea, but it should be implemented as a function too because defines are not seen by the MicroPython binding. 

I wonder if `LV_OPA_PCT(45)` would be more intuitive. 
Note that for coordinates we have simply `lv_pct()` which offsets the percentage to a very large value and if LVGL sees such a large coordinate it knows if it's a percentage. 

So `lv_opa(lv_pct(45))` could also work but it seems a little bit cumbersome. 

@embeddedt , @xiaoxiang781216 What do you think?

Miguel0101:
> Good idea, but it should be implemented as a function too because defines are not seen by the MicroPython binding.
> 
> I wonder if `LV_OPA_PCT(45)` would be more intuitive. Note that for coordinates we have simply `lv_pct()` which offsets the percentage to a very large value and if LVGL sees such a large coordinate it knows if it's a percentage.
> 
> So `lv_opa(lv_pct(45))` could also work but it seems a little bit cumbersome.
> 
> @embeddedt , @xiaoxiang781216 What do you think?

Sounds good to me.

embeddedt:
Hmm, I'm not sure about this. We already provide percentage enums in increments of 10 for opacity. What is the advantage of having yet another function/macro, as opposed to someone just picking a value between 0 and 255 if they want more precision?

Miguel0101:
> Hmm, I'm not sure about this. We already provide percentage enums in increments of 10 for opacity. What is the advantage of having yet another function/macro, as opposed to someone just picking a value between 0 and 255 if they want more precision?

You're right, but if the user wants a percentage value that isn't included in the `enum`, he could use the `lv_opa_pct` function to put in a percentage value.

kisvegabor:
My only problem with `lv_opa_pct()` is that it's not consistent with the naming of `lv_pct()`. 

Thinking about it more adding, I'm afraid of confusing the API and naming with it.

If there is no objection I'd rather not merge it. The user can easily use `value * LV_OPA_COVER) / 100);` manually, or add this macro himself if needed.

PR #2745, opened 2021-10-29
fix(anim): add lv_anim_get_playtime

xiaoxiang781216:
### Checkpoints

### DIFF
--- a/src/misc/lv_anim.c
+++ b/src/misc/lv_anim.c
@@ -116,6 +116,27 @@ lv_anim_t * lv_anim_start(const lv_anim_t * a)
     return new_anim;
 }
 
+uint32_t lv_anim_get_playtime(lv_anim_t * a)
+{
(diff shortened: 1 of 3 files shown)


src/misc/lv_anim.c
```diff
@@ -116,6 +116,27 @@ lv_anim_t * lv_anim_start(const lv_anim_t * a)
     return new_anim;
 }
 
+uint32_t lv_anim_get_playtime(lv_anim_t * a)
+{
+    uint32_t time = LV_ANIM_PLAYTIME_INFINITE;
```
kisvegabor: `time` might conflict with C library function. Theoretically it shouldn't matter if we don't include `time.h` but I've seen too many strange platforms to believe in this :laughing: 

So could you change it to `t` or so?

xiaoxiang781216: I change time to playtime.

kisvegabor:
Thanks!

PR #2748, opened 2021-10-30
fix(mem): ALIGN_MASK should equal 0x3 on 32bit platform

xiaoxiang781216:

minor fix

### Checkpoints

### DIFF
--- a/src/misc/lv_mem.c
+++ b/src/misc/lv_mem.c
@@ -34,7 +34,7 @@
     #define ALIGN_MASK       0x7
 #else
     #define MEM_UNIT         uint32_t
-    #define ALIGN_MASK       0x7
+    #define ALIGN_MASK       0x3
 #endif
 
 #define ZERO_MEM_SENTINEL  0xa1b2c3d4


embeddedt:
According to the `malloc` [manpage](https://man7.org/linux/man-pages/man3/realloc.3.html):

> The `malloc()` and `calloc()` functions return a pointer to the allocated memory, which is suitably aligned for any built-in type.

Since `long long` on most/all platforms is a 64-bit integer, I believe this means that `malloc` has to always return a pointer with 8-byte alignment.

I don't know if LVGL's allocator has to follow `malloc` semantics exactly, but it is convenient for compatibility.

xiaoxiang781216:
> According to the `malloc` [manpage](https://man7.org/linux/man-pages/man3/realloc.3.html):
> 
> > The `malloc()` and `calloc()` functions return a pointer to the allocated memory, which is suitably aligned for any built-in type.
> 
> Since `long long` on most/all platforms is a 64-bit integer, I believe this means that `malloc` has to always return a pointer with 8-byte alignment.
> 
> I don't know if LVGL's allocator has to follow `malloc` semantics exactly, but it is convenient for compatibility.

The alignment of TLSF is 4bytes on 32bit platform:
https://github.com/lvgl/lvgl/blob/master/src/misc/lv_tlsf.c#L232-L238
Since lvgl allocator is built on top of TLSF, even we keep ALIGN_MASK as 0x7, the allocated memory still align on 4bytes.

embeddedt:
Ah, okay. In that case, I agree with this change.

embeddedt (approved):

kisvegabor:
I think it was just a typo. 
Thank for the fix.

PR #2753, opened 2021-10-31
fix(CMake) split CMakeLists.txt, add options, includes and dependencies

higaski:
This PR splits the current CMakeLists.txt file into 4 files for each target:
- _./cmake/esp.cmake_ for the ESP target
- _./cmake/zephyr.cmake_ for the Zephyr target
- _./cmake/microphyton.cmake_ for the Micropython target
- _./cmake/baremetal.cmake_ for a "baremetal" target (for the lack of a better name)

In my opinion this makes things generally more clearly and easier to work with and or maintain. Changes to a specific target can be done in its corresponding file rather than in one large one.

I have only touched the CMake configuration for the "Baremetal" target the other ones are just copy/paste from the old CMakeLists.txt. I would appreciate it if someone who depends on those targets could test this PR... just in case.

So what actually changed?
- Added options to modify the following definitions
  - LV_LVGL_H_INCLUDE_SIMPLE, default: ON
  - LV_CONF_INCLUDE_SIMPLE, default: ON
  - LV_CONF_PATH, if defined adds parent folder to include path
    ```CMake
    # Project which includes lvgl sets LV_CONF_PATH
    set(LV_CONF_PATH
        ${CMAKE_CURRENT_SOURCE_DIR}/src/lv_conf.h
        CACHE STRING "" FORCE)
    ```
- Changed include_directories (bad practice) to target_include_directories
- Added dependencies
  - lvgl_exampels depends on lvgl
- Namespaced target aliases
  - lvgl::lvgl
  - lvgl::examples
- Applied cmake-format for a consistent style

Debatable:
Currently all includes set the SYSTEM keyword and therefor suppress warnings for users. I figured that's ok since there aren't any warnings specified anyhow.


higaski:
Oh dear. Looks like the relative includes ./cmake/* fail. Any ideas on why that is?
Also building lvgl as an ESP-IDF component seems to have a problem with CMakes project keyword.

higaski:
Some of the tests fail with
```CMake
CMake Error at /home/runner/work/lvgl/lvgl/cmake/baremetal.cmake:19 (add_library):
  No SOURCES given to target: lvgl_examples
```

This makes me wonder how exactly the tests are invoked?
Previously the sources to lvgl and lvgl_examples were added via
```CMake
file(GLOB_RECURSE SOURCES ${CMAKE_CURRENT_LIST_DIR}/src/*.c)
file(GLOB_RECURSE EXAMPLE_SOURCES ${CMAKE_CURRENT_LIST_DIR}/examples/*.c)
```

This isn't exactly wrong I guess, but it's uncommon. Usually sources get added relative to the currently processed CMakeLists.txt file with CMAKE_CURRENT_SOURCE_DIR. CMake's _include_ does not change this variable, so technically it would still point to the root folder.
```CMake
file(GLOB_RECURSE SOURCES ${CMAKE_CURRENT_SOURCE_DIR}/src/*.c)
file(GLOB_RECURSE EXAMPLE_SOURCES ${CMAKE_CURRENT_SOURCE_DIR}/examples/*.c)
```

However this doesn't work... why?

C47D:
> Also building lvgl as an ESP-IDF component seems to have a problem with CMakes project keyword.

The ESP-IDF project main CMakeLists.txt file also defines a project. So maybe it can be fixed by not defining project in this main CMakeLists.txt file if ESP-IDF is defined.

kisvegabor:
The new test environment was written by @cmumford. Chris, could you take a look at this issue?

What I can tell is that the tests are called from [this CI action](https://github.com/lvgl/lvgl/blob/master/.github/workflows/ccpp.yml).

higaski:
Finally...
I've worked around the weird issue by defining `LVGL_ROOT_DIR` in the roots CMakeLists.txt file instead of depending on `CMAKE_CURRENT_SOURCE_DIR` directly.

kisvegabor:
Great! 

I think it'd be important to document the options with CMake build. Could you add a new page in the [getting started](https://github.com/lvgl/lvgl/tree/master/docs/get-started) section of the docs?

higaski:
Like a "cmake.md" page? I think I could do that yes. Sadly I don't know a lot about the other targets. The options I added are only present in what I called the "baremetal" target. I assume the other targets (Zephyr, Micropython and ESP) are more self contained and don't require any further user configuration.

kisvegabor:
> Like a "cmake.md" page?

Yes, exactly.


I think it's enough to mention that LVGL's CMake file works together with Zephyr, ESP, etc build system but if you build LVGL yourself you have the following options...

kisvegabor:
Wonderful!  Thank you very much!

higaski:
Thank you for merging the PR!
I've noticed though that the [options table](https://docs.lvgl.io/master/get-started/cmake.html#additional-cmake-options) isn't formatted correctly. Any idea why?

kisvegabor:
I've fixed it here: https://github.com/lvgl/lvgl/commit/77287ca7d9a8ed4d4a54f1851a498a7678573a4c

higaski:
Perfect, thanks!

PR #2757, opened 2021-11-02
ci(micropython) fix git fetch

amirgon:
Recent failures in GitHub actions are related to this:  
https://github.blog/2021-09-01-improving-git-protocol-security-github/#no-more-unauthenticated-git

Need to switch from "git" protocol to "https" protocol when running git remote operations (git fetch)


embeddedt:
Thanks!

I had read the article back when it was published, but I was more focused on the key deprecation than the `git://` removal.

PR #2780, opened 2021-11-08
fix(config) Add default trace macros when LV_USE_LOG is disabled

kevinpt:

The LV_LOG_TRACE macros are not defined when LV_USE_LOG is 0. Most of the code referencing these macros use `#if LV_LOG_TRACE...` without checking if the macro is defined first. The reference to missing definitions is flagged by -Wundef.

This adds inactive defaults for all macros flagged by the warning.

### Checkpoints

### DIFF
--- a/src/lv_conf_internal.h
+++ b/src/lv_conf_internal.h
@@ -530,6 +530,17 @@ e.g. "stm32f769xx.h" or "stm32f429xx.h"*/
 #  endif
 #endif
 
+#else /*LV_USE_LOG*/
+#  define LV_LOG_LEVEL            LV_LOG_LEVEL_NONE
+#  define LV_LOG_PRINTF           0
+#  define LV_LOG_TRACE_MEM        0
+#  define LV_LOG_TRACE_TIMER      0
+#  define LV_LOG_TRACE_INDEV      0
+#  define LV_LOG_TRACE_DISP_REFR  0
+#  define LV_LOG_TRACE_EVENT      0
+#  define LV_LOG_TRACE_OBJ_CREATE 0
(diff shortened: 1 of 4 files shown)


src/lv_conf_internal.h
```diff
@@ -530,6 +530,17 @@ e.g. "stm32f769xx.h" or "stm32f429xx.h"*/
 #  endif
 #endif
 
+#else
+#  define LV_LOG_LEVEL            LV_LOG_LEVEL_NONE
```
xiaoxiang781216, line 534: lv_conf_internal.h is generated by python script, you shouldn't modify this file manually.

kisvegabor: It's a little bit odd, but it seems easier to fix it in the source files. E.g.
```c
#if LV_USE_LOG && LV_LOG_TRACE_EVENT
    #define EVENT_TRACE(...) LV_LOG_TRACE(__VA_ARGS__)
#else
    #define EVENT_TRACE(...)
#endif
```

kevinpt: I've modified the script to parse a new LVGL specific pragma in the template so that defines will not be expanded. That lets the header ensure conditional defines have a default value.

kisvegabor: If we add it to `lv_conf_template.h` the `lv_conf.h` of the user will be noisier and longer. 

If you don't like the suggestion in my [previous comment](https://github.com/lvgl/lvgl/pull/2780/files#r745719207) and another option could be to manually add the problematic cases at the bottom of `lv_conf_internal.h` (i.e. generate via the script).

What do you think?

PR #2787, opened 2021-11-10
Added additional checks for consecutive vowels and added simple suppo…

DennisSHCheung:

LVGL currently does not display arabic vowel characters directly top/bottom of the next base character correctly. During development, it was discovered that it's also possible to have 2 consecutive vowels on the same base character. This pull request is to open further discussions on how to perfect vowel character placements.

List of tasks remaining:
- Ensure x and y positions of basic vowels are located correctly
- Fix coding style (lv_draw_label.c and lv_txt_ap.c both need to check for arabic vowels so should put the function a common header and remove temporary static variables (next_letter_temp & next_next_letter_temp) etc)
- Add test to ensure string conjunction is working fine
- %u0651%u064E (w/) and %u064E%u0651 (/w) despite different ordering, should be drawn as (/w) (see below pictures: first and final string)

Expected:
image

Current:
image
image

### Checkpoints

### DIFF
--- a/src/draw/lv_draw_label.c
+++ b/src/draw/lv_draw_label.c
@@ -463,27 +463,32 @@ LV_ATTRIBUTE_FAST_MEM void lv_draw_letter(const lv_point_t * pos_p, const lv_are
         return;
     }
 
-    /*If the letter is an arabic vowel it's x and y positions need to be calculated based on the following character*/
-    if (letter >= 0x064B && letter <= 0x0652)
-    {
+    /*Arabic vowels need special xy calculation*/
+    if(letter >= 0x064B && letter <= 0x0652) {
         lv_font_glyph_dsc_t next_glyph;
         lv_font_get_glyph_dsc(font_p, &next_glyph, next_letter_temp, '\0');
 
-        pos_x = pos_p->x + ((next_glyph.box_w - g.box_w) / 2) - 1;  // Center the vowel character based on the width of the next character
+        uint32_t adv_w = next_glyph.adv_w;
+        int32_t next_y = pos_p->y + (font_p->line_height - font_p->base_line) - next_glyph.box_h - next_glyph.ofs_y;
+        if(next_next_letter_temp) { // Next character is a vowel
+            lv_font_glyph_dsc_t next_next_glyph;
+            lv_font_get_glyph_dsc(font_p, &next_next_glyph, next_next_letter_temp, '\0');
 
-        int32_t next_letter_pos_y = pos_p->y + (font_p->line_height - font_p->base_line) - next_glyph.box_h - next_glyph.ofs_y;
-        if(letter == 0x064D || letter == 0x0650) {  // Kasra and tanween kasra vowel sits below a normal character
-            pos_y = next_letter_pos_y + next_glyph.box_h;
+            adv_w = next_next_glyph.adv_w;
+
+            int32_t next_next_y = pos_p->y + (font_p->line_height - font_p->base_line) - next_next_glyph.box_h - next_next_glyph.ofs_y;
+            next_y = next_next_y - next_glyph.box_h;
         }
-        else {
-            if (next_next_letter_temp) {  // If the next character's also a vowel then the character after it has to be used for calculations
-                lv_font_glyph_dsc_t next_next_glyph;    
-                lv_font_get_glyph_dsc(font_p, &next_next_glyph, next_next_letter_temp, '\0');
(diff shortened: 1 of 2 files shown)


src/draw/lv_draw_label.c
```diff
@@ -449,6 +463,34 @@ LV_ATTRIBUTE_FAST_MEM void lv_draw_letter(const lv_point_t * pos_p, const lv_are
...
+            lv_font_glyph_dsc_t next_next_glyph;
+            lv_font_get_glyph_dsc(font_p, &next_next_glyph, next_next_letter_temp, '\0');
+
+            adv_w = next_next_glyph.adv_w;
+
+            int32_t next_next_y = pos_p->y + (font_p->line_height - font_p->base_line) - next_next_glyph.box_h - next_next_glyph.ofs_y;
+            next_y = next_next_y - next_glyph.box_h;
+        }
+
+        // Add a small offset between base character and vowel unless the base character
```
kisvegabor, line 483: For consistency please use `/* */` comment style.

src/draw/lv_draw_label.c
```diff
@@ -449,6 +463,34 @@ LV_ATTRIBUTE_FAST_MEM void lv_draw_letter(const lv_point_t * pos_p, const lv_are
         return;
     }
 
+    /*Arabic vowels need special xy calculation*/
+    if(letter >= 0x064B && letter <= 0x0652) {
```
kisvegabor, line 467: This block should be wrapped into `#if LV_USE_ARABIC_PERSIAN_CHARS`.

kisvegabor (commented):
Thank you for this PR. I had a few comments related to formatting.

I wonder if you can add some [tests](https://github.com/lvgl/lvgl/tree/master/tests) for `src/misc/lv_txt_ap.c` to be sure the joins and vowel skipping works well. 

The bidi algorithm is another but now unrelated topic to test once in the future.

DennisSHCheung:
Hi, @kisvegabor ,

After actually studying the problem instead of just applying offsets and doing confusing calculations, I spent some time opening up a ttf file with fontforge and understanding how combinations are done. It seems that each of the glyph provides some anchor information which indicates where a mark glyph should be placed if there's any. I believe the current font converter does not retrieve that piece of information and I wonder if you can help me out here.

The anchor information will be very useful not only to arabic but also to languages with accents. Thai is a language that has similar problems. When there's only one accent, because Thai's base character's height is always the same, simply drawing an accent most of the time will result a correct display. However, problems arise when you have 2 accents like this: image. The expected form should be this: image
 This can also be solved with the anchor mark information.

DennisSHCheung:
> Hi, @kisvegabor ,
> 
> After actually studying the problem instead of just applying offsets and doing confusing calculations, I spent some time opening up a ttf file with fontforge and understanding how combinations are done. It seems that each of the glyph provides some anchor information which indicates where a mark glyph should be placed if there's any. I believe the current font converter does not retrieve that piece of information and I wonder if you can help me out here.
> 
> The anchor information will be very useful not only to arabic but also to languages with accents. Thai is a language that has similar problems. When there's only one accent, because Thai's base character's height is always the same, simply drawing an accent most of the time will result a correct display. However, problems arise when you have 2 accents like this: image. The expected form should be this: image This can also be solved with the anchor mark information.

DennisSHCheung:
After studying opentype.js, it seems that gpos table 6 is not supported at the moment, I presume that means there's no way to retrieve the glyph's anchor for now. I think the whole "fix" is not being done right and need much more investigation.

kisvegabor:
Can you send a description you have found useful about gpos table 6?

DennisSHCheung:
https://docs.microsoft.com/en-us/typography/script-development/arabic at Mark to mark positioning section.

If you open an Arabic/Thai ttf with fontforge, you will be able to find these "anchor marks" near each glyph, which is used to position compound word together. 
image

I checked the status of opentype.js on GPOS but I don't believe they have got it working yet.

kisvegabor:
Our font converter [uses FreeType](https://github.com/lvgl/lv_font_conv/tree/master/lib/freetype) not opentype.js but its [docs](https://freetype.org/freetype2/docs/glyphs/glyphs-4.html) says it doesn't support GPOS either. 

I found detailed docs about GPOS [here](https://docs.microsoft.com/en-us/typography/opentype/spec/gpos).

I cc @puzrin, the author of lv_font_conv library.

puzrin:
GPOS is not supported now.

`opentype.js` knows how to emulate `kern` table from `GPOS`. It's used in `lv_font_conv` as companion for `FreeType`. Because `FreeType` does not supports kerning at all.

Font spec allows add new tables, if anyone wish to extend. But GPOS support is not easy.

PR #2793, opened 2021-11-11
fix(draw_mask):will crash if get_width/height < 0

guoweilkd:

 Radius will still be < 0 if  get_width/height < 0 .when the circ_calc_aa4() function is called  after. `lv_mem_alloc(radius * 6 + 6)` will fail due to the large amount of memory that needs to be allocated.

### Checkpoints


kisvegabor:
Thanks! I'd be curious what kind of application do you develop that helps to discover these corner cases :slightly_smiling_face:

Issue #2608, opened 2021-09-27
add font fallback support.


incity:
[https://en.wikipedia.org/wiki/Fallback_font](url)

[https://www.w3schools.com/css/css_font_fallbacks.asp](url)

kisvegabor:
Please send a feature request for it in the forum.
https://forum.lvgl.io/c/feature-request/9

PR #2796, opened 2021-11-11
Supports Freetype fonts in SDL Driver

mariotaku:

* Supports SDL renderer for Freetype fonts (`lv_freetype`) and hopefully others
* Supports fallback fonts - #2608 

### Checkpoints


src/extra/libs/freetype/lv_freetype.c
```diff
@@ -151,6 +154,11 @@ void lv_ft_font_destroy(lv_font_t * font)
 #endif
 }
 
+bool lv_ft_font_check_type(const lv_font_t * font)
```
kisvegabor: What does `type` mean here? It seems like a "validity" test.

src/font/lv_font.c
```diff
@@ -84,6 +85,11 @@ uint16_t lv_font_get_glyph_width(const lv_font_t * font, uint32_t letter, uint32
     else return 0;
 }
 
+const lv_font_t* lv_font_get_normal()
+{
+    lv_theme_t *theme = lv_disp_get_default()->theme;
```
kisvegabor: Ideally fonts shouldn't know about `lv_disp_t` and `lv_theme_t` .

src/draw/lv_draw_label.c
```diff
@@ -101,7 +101,7 @@ LV_ATTRIBUTE_FAST_MEM void lv_draw_label_dsc_init(lv_draw_label_dsc_t * dsc)
     lv_memset_00(dsc, sizeof(lv_draw_label_dsc_t));
     dsc->opa = LV_OPA_COVER;
     dsc->color = lv_color_black();
-    dsc->font = LV_FONT_DEFAULT;
+    dsc->font = lv_font_get_normal();
```
kisvegabor: It seems risk with multiple displays if the displays' themes have different `font_normal`. This way always the current default display's font (set by `lv_disp_set_default(disp)`) will be used on all displays. 

Instead, the user can set a font for the screens in the theme. Now no font is set but we safely set `font_normal` (or `LV_FONT_DEFAULT` if it's `NULL`) in the screen style.

kisvegabor (commented):
Thanks! I had a few comments.

A possible issue could be that LVGL determines the line height from `font->line_height`. If a glyph is missing and the fallback font's `line_height` is larger, that glyph might be clipped. 
A similar problem can happen if the baseline of the two fonts is not aligned.

src/extra/libs/freetype/lv_freetype.c
mariotaku: My idea is to check if this is a lv_freetype font, or a built-in font. If we have some new font implementation, creating such function can help developers to use different implementation for different fonts.

kisvegabor: I see. 

Can this function be `static`?

What about a  name like `is_ft_font`?

mariotaku: I'm afraid this function can't be static. It compares function pointers with static functions inside lv_freetype.c, so itself has to be global.
Since it's a global function, the name convention should better follow other functions.

mariotaku: By the way it's not used currently. Maybe I can remove it and aff when we really need it.

mariotaku:
@kisvegabor Do you have suggestions when sizes and baseline are different?

mariotaku:
After testing for days, I think this can be reviewed and considered for merging now. Although there are some caveats that you've mentioned, but the render works pretty well.

image

Works with UI fonts + CJK font fallback too:

image
image

Those fonts have different baselines, but seems that leaving that as-is looks the best.

kisvegabor:
Okay, I also think that having fallbacks is better than nothing even if there will be some minor issues. 

I wonder if we can/should support it for all fonts. The function can try the fallbacks automatically if the glyph is not found.
What do you think?

mariotaku:
I'll try to produce such case to test if it works for other fonts too.

kisvegabor:
Great, thanks!
The only initial problem I see is the generated fonts `const` by default, but we can figure out something to manage it.

mariotaku:
@kisvegabor So I think for user they can allocate by themself, and modify the fallback pointer because this will make font mutable.

mariotaku:
Done. This fallback mechanism works fine for both font types.

kisvegabor:
> Done. This fallback mechanism works fine for both font types.


Could you push it? :slightly_smiling_face:

mariotaku:
@kisvegabor Oops sorry I pushed to another branch. Done now.

src/font/lv_font.c
```diff
@@ -64,7 +64,17 @@ bool lv_font_get_glyph_dsc(const lv_font_t * font_p, lv_font_glyph_dsc_t * dsc_o
                            uint32_t letter_next)
 {
     LV_ASSERT_NULL(font_p);
-    return font_p->get_glyph_dsc(font_p, dsc_out, letter, letter_next);
+    LV_ASSERT_NULL(dsc_out);
```
kisvegabor, line 68: What about this implementation?
```c
bool lv_font_get_glyph_dsc(const lv_font_t * font_p, lv_font_glyph_dsc_t * dsc_out, uint32_t letter,
                           uint32_t letter_next)
{
    const lv_font_t * f = font_p;
    while(f) {
        if (font_p->get_glyph_dsc(font_p, dsc_out, letter, letter_next)) {
            dsc_out->resolved_font = f;
            return true;
        }
        f = f->fallback;
    }
    return false;
}
```

The main advantage is ` lv_font_glyph_dsc_t g = {.resolved_font = NULL};` is not required.

src/font/lv_font_fmt_txt.c
```diff
@@ -182,6 +185,7 @@ bool lv_font_get_glyph_dsc_fmt_txt(const lv_font_t * font, lv_font_glyph_dsc_t *
     dsc_out->ofs_x = gdsc->ofs_x;
     dsc_out->ofs_y = gdsc->ofs_y;
     dsc_out->bpp   = (uint8_t)fdsc->bpp;
+    dsc_out->missing = false;
```
kisvegabor: What is the exact purpose of this flag?

mariotaku: According to my observation, in freetype implementation, even a glyph is missing, it still can draw fallback glyph (tofu). So it's not enough to simply check if `lv_font_get_glyph_dsc_fmt_txt` can return true.

image

src/draw/lv_draw_label.c
```diff
@@ -449,6 +449,9 @@ LV_ATTRIBUTE_FAST_MEM void lv_draw_letter(const lv_point_t * pos_p, const lv_are
         return;
     }
 
+    if (g.resolved_font) {
```
kisvegabor, line 452: It should be removed too, shouldn't it?

src/font/lv_font_fmt_txt.c
kisvegabor: Got it. I think a name like `placeholder_glyph` would be more expressive. What do you think?

mariotaku: I agree. I changed the name to `is_placeholder` for now.

src/draw/lv_draw_label.c
mariotaku: This is still necessary, as when we get the bitmap, we don't get hold of glyph description. So it must use correct font pointer to get glyph bitmap, unless we change `lv_font_get_glyph_bitmap` signature or create another function with glyph info.

kisvegabor: True, my bad.

kisvegabor:
All looks good thank you! Only one thing left (the necessary bad): please add a few words about this new feature in the [docs](https://docs.lvgl.io/master/overview/font.html) too.

mariotaku:
One last small change. This adds way to use pointer in RAM/ROM to load font. I'll add in the docs as well if there is space for freetype.

mariotaku:
This is ready for merging, please let me know if there are any concerns :)

kisvegabor:
Thank you! RAM/ROM font loading is also great! It'd be better to have it in a separate PR for a cleaner commit history. We try to autogenerate the changelog, so it's important to have meaningful commits. But anyway, it's fine now.

The docs of freetype is here: https://docs.lvgl.io/master/libs/freetype.html

 
(I'll merge it when the CI is ready)

PR #2799, opened 2021-11-12
optimize: add non-null judgment to lv_examplse_obj_2

guoweilkd:

hi，although indev is definitely not NULL in LV_EVENT_PRESSING event . but example is a template, and other events may be modified to mimic this function.

### Checkpoints

### DIFF
--- a/examples/widgets/obj/lv_example_obj_2.c
+++ b/examples/widgets/obj/lv_example_obj_2.c
@@ -6,6 +6,8 @@ static void drag_event_handler(lv_event_t * e)
     lv_obj_t * obj = lv_event_get_target(e);
 
     lv_indev_t * indev = lv_indev_get_act();
+    if(indev == NULL)  return;
+
     lv_point_t vect;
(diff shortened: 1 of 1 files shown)


kisvegabor:
It makes sense. In general, the examples should work like a copy-paste template.

examples/widgets/obj/lv_example_obj_2.c
```diff
@@ -6,6 +6,10 @@ static void drag_event_handler(lv_event_t * e)
     lv_obj_t * obj = lv_event_get_target(e);
 
     lv_indev_t * indev = lv_indev_get_act();
+    if(indev == NULL) {
```
kisvegabor: On other places the one-line format is used:
```c
    if(indev == NULL)  return;
```

For consistency please use it here too.

guoweilkd: done

kisvegabor:
Thanks!

Issue #2733, opened 2021-10-26
More flexible custom drawing interface


FASTSHIFT:
### Introduce the problem
Currently lvgl already supports the use of `LV_USE_EXTERNAL_RENDERER` to switch external renderers, but this means that the default internal software renderer cannot be used.
Is it possible to use function pointers to manage drawing interface functions?

### Examples and cases
```C
/* Define the function pointer of lv_blend_fill */
typedef void (*lv_blend_fill_cb_t)(const lv_area_t * clip_area, const lv_area_t * fill_area,
                                   lv_color_t color, lv_opa_t * mask, lv_draw_mask_res_t mask_res, lv_opa_t opa,
                                   lv_blend_mode_t mode);

/* Define a structure containing all draw related functions */
typedef struct {
    lv_blend_fill_cb_t blend_fill_cb;
    ...
} lv_draw_func_t;

static lv_draw_func_t draw_func;

static void lv_draw_init()
{
    /* Initialize the drawing function structure,
     * all points to the internal software rendering function by default
     */
    draw_func.blend_fill_cb = _lv_blend_fill;
    ...
}

/* Replace the default software rendering */
void lv_draw_set_custom_blend_fill_cb(lv_blend_fill_cb_t callback)
{
    draw_func.blend_fill_cb = callback;
}
```

### Suggested solution
The function pointer points to the internal software rendering function by default. When the user's hardware has certain graphics acceleration capabilities, the function pointer can be modified to point to its own drawing function.

kisvegabor:
`LV_USE_EXTERNAL_RENDERER` does much more than replacing the `blend` functions. It also allows you to  renders gradient, shadow, border, letters, lines, polygons, image transformations, etc with a different rendering engine (now only SDL is supported besides SW rendering). The `blend` functions only blend an image or color (with an optional mask) to the current half-ready image. 

Regarding GPU support we found that it's much more flexible to implement GPU support right into LVGL. Its because the GPUs can have very different features and it's too complicated to abstract all the possible drawing features to make most out of the GPUs, handle fallbacks, etc. 

Can you share your use case in more detail? If we had better understanding of your problem, it'd be easier to find a good solution.

kisvegabor:
@xiaoxiang781216 has highlighted some important points here: https://github.com/lvgl/lvgl/issues/2720#issuecomment-952631467

We are about releasing v8.1. Let's turn back to it after the release.

lishaoenxm:
[...]

PR #2803, opened 2021-11-16
Arch/draw backend abstraction

kisvegabor:
Continue #2733

I've abstracted the SW render. I wasn't thinking much about how to create, register, destroy renderers or fallback although it's an important topic that requires serious planning.

Notes about the current state:
- `lv_draw.h` contains the API that can be extended/replaced. Basically the blending and drawing functions. 
- Seeing these functions next to each other it's clear that their protype should be unified.
- In `draw/sw` you can find the decoupled SW renderer
- I've cleaned up the implementation of the SW blend functions
- Probably masking should be more generic and abstracted too
- The NXP and STM GPUs are removed, for now, they should have their own blend and draw functions later

@mariotaku please take a look at it. Is it seem possible to port the SDL GPU API?

As an additional engine what about [Cairo](https://cairographics.org/samples/)? It's very popular, vector-based, and uses HW acceleration is possible.


mariotaku:
Nice work. I have played around with it, although not be able to run a demo yet, but I have some suggestions and will comment directly in the code.

src/draw/lv_draw.h
```diff
@@ -36,11 +35,74 @@ extern "C" {
 /**********************
  *      TYPEDEFS
  **********************/
+typedef struct _lv_draw_backend_t{
+    struct _lv_draw_backend_t * base;
+
+    void * ctx;
+
+    void (*draw_rect)(const lv_area_t * coords, const lv_area_t * clip, const lv_draw_rect_dsc_t * dsc);
```
mariotaku, line 43: Please add `lv_draw_backend_t` or `ctx` to the end of parameters list. By doing so, I can access SDL renderers and offscreen textures, and stop setting non-buffer pointer to `lv_disp_draw_buf_t`, and leave `lv_disp_drv_t` untouched.

src/draw/lv_draw.h
```diff
@@ -36,11 +35,74 @@ extern "C" {
...
+
+    void (*draw_arc)(lv_coord_t center_x, lv_coord_t center_y, uint16_t radius,  uint16_t start_angle, uint16_t end_angle,
+                     const lv_area_t * clip_area, const lv_draw_arc_dsc_t * dsc);
+
+    void (*draw_img)(const lv_area_t * map_area, const lv_area_t * clip_area,
```
mariotaku, line 48: Please let implementation handle all the logic (The same as `lv_draw_img`). By doing so, for implementations uses texture, they can process `src` on their own.

mariotaku (changes requested):

mariotaku:
Update: Got Music Player demo up and running. I'll create a new branch and show you changes later.

mariotaku:
Please make changes as you want. I'll rebase my changes to reflect upstream until API stabilized.

Several proposals about backend mechanism:

- How to handle per-monitor cases? Should we bind backend to display? For SDL, every window has their own renderer context. So it's not possible to share one backend context for multiple windows.
- There are 2 kinds of ways to have backends chain, and I prefer the first one:
    1. Expose lv_draw_backend_t to `draw_` functions, so they have access to `base` pointer and can call it, like `super` keyword.
    2. Iterate through backends chain and check return value of callback function (and of course, if the function pointer is null).

My changes: https://github.com/lvgl/lvgl/compare/arch/draw-backend-abstraction..mariotaku:draw-backend-sdl

src/draw/lv_draw.h
kisvegabor: Ok, I will.

src/draw/lv_draw.h
kisvegabor: It'd skip LVGL's image decoder interface. E.g. if a user registers a special custom decoder, its result should be passed to the drawers.

kisvegabor:
> How to handle per-monitor cases? Should we bind backend to display? For SDL, every window has their own renderer context. So it's not possible to share one backend context for multiple windows.

It's a big question, Let me organize my thoughts here. These are the options I see:
- 1 renderer - 1 display: Could be anything, SW render, embedded or high-end GPU render.
- N renderer - N display: Each display has its own renderer. 
- 1 renderer - N display: One renderer updates the displays one by one 
- N renderer - 1 display: Multicore SW render or multiple GPU's running in parallel. LVGL needs to delegate tasks to keep the renders busy.

The most simple case is 1-N as it assumes that the render is not strictly related to the display and it can render to any buffer it sees. E.g. the current SW render or the typical embedded GPUs. flush_cb is independent of the renderer, the user can use any periphery to send the rendered buffers to display somehow. 

With 1-1 or N-N probably the renderer needs to know about the display and manage the flush_cb too. (E.g. SDL).
In this case, we need to make the connection between an instance of the renderer and the given display. (Now SDL saves it in the `disp_drv->user_data`).

N-1 is quite complex and not that typical. Renderers are working on the same "surface", or creating small surfaces for the tile they need to render, or just draw into a buffer? 

Anyway, we can have a backend base class that can be extended by the user in runtime and an instance of a backend class (or more instances for N-1 case?) can be assigned to a display. 
Two things:
- Usually, the users use a built-in backend (now mainly SW render). The built-in backend needs to be instancianiated and assigned automatically when a display is created. Later the user can overwrite it with a custom backend if he wishes.
- I can't really see how to handle that some backends need to control the flush_cb and others don't.  Probably the whole display initialization should be handled differently. See the [proposal here](https://github.com/lvgl/lvgl/issues/2720#issue-1033542887). 

> Expose lv_draw_backend_t to draw_ functions, so they have access to base pointer and can call it, like super keyword.

I agree, it looks like a class so we can do an OOP-like approach as @puzrin also suggested.

kisvegabor:
I was reviewing Cario and was thinking about how to rework masking to follow a generic masking concept. 

The goal is to allow creating any number of masks that should be applied during blending (something like this already works). 

The current mask scheme is very limited: it can only remove parts from a "fully transparent (0xff)" masks and can not build back parts. 

Usually, masking works like the opposite: Have "fully blocking (0x00)" masks and draw something on them. It's because LVGL handles only one mask layer and if masks could build back parts they could ignore the effect of the previous mask. A practical example:
1. There is a large circle widget with `clip_corner=1`, i.e. the overflowing parts on the corner are masked out.
2. There is a rounded rectangle that is positioned to be clipped by the circle. (the rounded rectangle also uses a mask)

If the rounded rectangle could draw its mask outside the circle mask it would render the circle mask useless as it was simply overwritten. 

The proposed solution is to introduce the concept of "mask layer". A mask layer starts with 0x00 (fully blocking) and the user can draw anything on it. There can be multiple mask layers that are merged into one and this final mask is used during blending.

For the above example: 
- Mask layer 1: a circle 
- Mask layer 2: a rounded rectangle
- Final mask: where both masks are 0xff

Mask layers can be added and removed dynamically as objects are being drawn. During drawing LVGL gets and merges the masks and passes them to the blend functions. 

The API in `lv_draw_backend_t` would be something like:
- `create/delete_mask_layer`: to create a new mask layer object, a placeholder for masks
- `add/remove_mask_layer()`: add/remove the mask layer to be considered during rendering
- `create/delete_<mask_type>`: create/delete a new mask
- `add/remove_mask`: add/remove the mask to/from a mask layer
- `<mask_type>`s:
    - `fill(alpha)`: fill the mask with this alpha    
    - `rectangle(x1, y1, x2, y2, radius, alpha)`
    - `arc(cx, cy, radius, angle1, angle2, width, alpha)`
    - `line(x1, y1, x2, y2, width, alpha)`
    - `letter(x1, y1, font, alpha)`
    - `gradient(x1, y1, x2, y2, alpha_start, alpha_end)`: SW render would support only horizntal and vertical gradient
    - `path(...)` lines, bezier, arc, like in SVG (not for SW render)    

It seems I can change LVGL's current mechanisms to work like this, and it can be realized with Cairo too. How does it fit into SDL?

mariotaku:
With recent versions of SDL, I draw the shape (ARGB) and the mask (Alpha 8) on 2 textures, and use custom blend mode to modify shape texture then draw on screen. This is due to SDL Renderer is still very primitive and can barely handle complex shapes.

mariotaku:
Cairo might be a very good option and it looks like we can achieve render performance like GTK (as they use Cairo IIRC).
For my use case (Linux with age old dependencies) I'd prefer nanovg as it doesn't have external dependency.

By the way do you think separating different backends to other repos is a good idea?

kisvegabor:
> With recent versions of SDL, I draw the shape (ARGB) and the mask (Alpha 8) on 2 textures, and use custom blend mode to modify shape texture then draw on screen. 

Do you use lvgl's mask functions to draw to the Alpha8 texture, right?

> I'd prefer nanovg as it doesn't have external dependency.

Except for OpenGL, isn't it? 

> By the way do you think separating different backends to other repos is a good idea?

We are trying to move as much stuff to the lvgl repo as possible to make versioning, branching and CI simpler. So I'd prefer to keep them here.

kisvegabor:
I'd like to merge this to master else it will be difficult to merge the patches of the current draw functions. 

So the goal is to make it work to the extent of the current behaviour in master. SW rendering is working, @mariotaku could you add the SDL part here?

mariotaku:
Yep no problem. I'll start this part tomorrow.

mariotaku:
> Do you use lvgl's mask functions to draw to the Alpha8 texture, right?

Yes. I use that to generate alpha8 texture.

> Except for OpenGL, isn't it?

Yes. OpenGL only.

> We are trying to move as much stuff to the lvgl repo as possible to make versioning, branching and CI simpler. So I'd prefer to keep them here.

No problem. I'll continue my patches here.

mariotaku:
So to my understanding, for this merge you will not add function signature changes to lv_draw_backend_t, and do it later.
What you want me to do is to just make HW renderer works, then we do more refactoring later. Is this correct?

kisvegabor:
Yes, exactly. The goal is just to move the files to their new place.

kisvegabor:
I've just made all tests pass.

mariotaku:
OK. Now I'll do refactoring base on this branch and make a pull request.

kisvegabor:
Ok, until that I merge it. Please send a PR when you are ready.

PR #2804, opened 2021-11-17
optimize(qrcode):eliminate blank

guoweilkd:

Extra blank occurs when obj_w % qr_size != 0(see [link](https://github.com/lvgl/lvgl/blob/master/src/extra/libs/qrcode/lv_qrcode.c#L111) ). according to the definition of the QR code, the max value of qr_size  can be 40. In other words, blank ranges from 0 to 39px.this can be challenging to show friendliness. In order to ensure the size of the generated QR code is fixed, so I inserted this blank into the image data. according to the test, the change had almost no effect on the scans.

### Checkpoints

### DIFF
--- a/src/extra/libs/qrcode/lv_qrcode.c
+++ b/src/extra/libs/qrcode/lv_qrcode.c
@@ -85,15 +85,32 @@ lv_res_t lv_qrcode_update(lv_obj_t * qrcode, const void * data, uint32_t data_le
 
     if(data_len > qrcodegen_BUFFER_LEN_MAX) return LV_RES_INV;
 
-    uint8_t * qr0 = lv_mem_alloc(qrcodegen_BUFFER_LEN_MAX);
+    lv_img_dsc_t * imgdsc = lv_canvas_get_img(qrcode);
+
+    int32_t qr_version = qrcodegen_getMinFitVersion(qrcodegen_Ecc_MEDIUM, data_len);
+    if (qr_version <= 0) return LV_RES_INV;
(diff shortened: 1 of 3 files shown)


kisvegabor:
If I understood correctly, some rows/columns are duplicated. Really is this the case? 

I think it works in most of the cases but imagine e.g if qr_size = 40 and obj_widht = 60 then every second row/column will be duplicated which can look ugly. 

What about having a `lv_qrcode_update_strecthed` for the logic you have implemented?

guoweilkd:
> will be duplicated which can look ugly.
I agree, do you have good idea?

> according to the definition of the QR code, the max value of qr_size can be 40. In other words, blank ranges from 0 to 39px.
I made a mistake here. the max value of qr_size can be 177.In other words, margin ranges  from 0 to 176px.So we should pay attention to this problem.

guoweilkd:
Hi, I have an idea. see the latest push code.
1. Peek the smallest version of the QR code required by the text.
2. Calculate margin with the version. If the margin is too large, increase  QRcode version value accordingly. because qr_size increases as the version increases.
3. Use the newly Qrcode version draw img.

we need to add  `qrcodegen_getMinFitVersion()` in qrcodegen.c. what do you think?

kisvegabor:
So it helps to avoid the very ugly cases but a few pixels can remain on the edges, right?

guoweilkd:
> So it helps to avoid the very ugly cases but a few pixels can remain on the edges, right?

yes.

kisvegabor:
Great, merging. Thank you!

PR #2675, opened 2021-10-16
fix(hal tick): add precompile !LV_TICK_CUSTOM for global variables and lv_tick_inc()

mysterywolf:
### Description of the feature or fix

When users enable the LV_TICK_CUSTOM, `sys_time` , `tick_irq_flag` and `lv_tick_inc()` are meaningless. Thus, they should be excluded from compiling.

### Checkpoints
- [x] Follow the [styling guide](https://github.com/lvgl/lvgl/blob/master/docs/CODING_STYLE.md)
- [ ] Update [CHANGELOG.md](https://github.com/lvgl/lvgl/blob/master/docs/CHANGELOG.md)
- [ ] Update the documentation

PR #2810, opened 2021-11-18
fix(log): to reduce unnecessary footprint

mysterywolf:

to reduce unnecessary footprint when LV_LOG_PRINTF is enabled

### Checkpoints

### DIFF
--- a/src/core/lv_obj.c
+++ b/src/core/lv_obj.c
@@ -199,7 +199,7 @@ void lv_deinit(void)
 
     LV_LOG_INFO("lv_deinit done");
 
-#if LV_USE_LOG
+#if LV_USE_LOG && !LV_LOG_PRINTF
     lv_log_register_print_cb(NULL);
 #endif
 }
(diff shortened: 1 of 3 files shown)


embeddedt:
I think we should leave the function there but with an empty body, for binary compatibility. Most linkers will drop it from the final binary anyways.

mysterywolf:
> binary compatibility

I have comfirmed that it will not be dropped in Keil-MDK. 8 Bytes saved😂.  Meanwhile, this also improve the readability - to let programmers easily understand this function cannot be used under this kind of configuration.

mysterywolf:
.. and this is my second time to find this kind of problem #2675. Alought we just can save only a few bytes in each PR😅, it also need to be saved, because this is the embedded system environment. Every byte needs to budget.

kisvegabor:
I think it'd be a little bit ambiguous this way: the option for a custom logger depends on whether the printf logging is enabled. 

IMO having more direct config options for logging would be clearer. E.g.
```
#define LV_LOG_PRINTF 1
#define LV_LOG_FILE "S:/folder/mylog.txt"
#define LV_LOG_CUSTOM 1 /*Allow registering a custom clabback*/
```

PR #2811, opened 2021-11-18
feat(rt-thread): fix a bug of log

mysterywolf:

fix a bug of log

### Checkpoints

### DIFF
--- a/rt-thread/lv_rt_thread_port.c
+++ b/rt-thread/lv_rt_thread_port.c
@@ -23,7 +23,7 @@
 #  include <lv_port_indev.h>
 #endif
 
-#if LV_USE_LOG && LV_LOG_PRINTF
+#if LV_USE_LOG
 static void lv_rt_log(const char *buf)
 {
     LOG_I(buf);
@@ -32,7 +32,7 @@ static void lv_rt_log(const char *buf)
 
 static int lv_port_init(void)
 {
-#if LV_USE_LOG && LV_LOG_PRINTF
+#if LV_USE_LOG
     lv_log_register_print_cb(lv_rt_log);
 #endif
 


kisvegabor:
I think it makes sense to keep both options in parallel. E.g. the user might use printf and log to file simultaneously. 
Although once the user has a custom print function to log into a file he can add a printf there too. 

Anyway, I think we should keep the API the same as much as possible regardless of the config changes unless the changes are very clear and have a huge impact. (E.g. disable a widget). 

What do you think?

mysterywolf:
> I think we should keep the API the same as much as possible regardless of the config changes unless the changes are very clear and have a huge impact. (E.g. disable a widget).

There is a problem if the API is regardless of the configuration. Users cannot easily distinguish which APIs can be used under the current config and which APIs cannot be used under the current config. 
For example, in #2810 , in theory,  `lv_log_register_print_cb` only can be used effectively if `LV_LOG_PRINTF` is disabled (not defined); however, if we don't add any macros to limit, `lv_log_register_print_cb` also can be "used" even if `LV_LOG_PRINTF` is enabled. Users begin to confuse: "why I registered a printf callback function, but it doesn't work?" Especially for beginners who are not very familiar with the internal source code of LVGL, this (regardless of the config changes) will mislead them.

Maybe the document has a clear explanation, but I think we still need to remind users clearly and directly, because users cannot recite every sentence in the document.

On the other hand, memory wasted.

kisvegabor:
>  "why I registered a printf callback function, but it doesn't work?"

I think the root of the confusion comes from here:
```c
void lv_log(const char * buf)
{
#if LV_LOG_PRINTF
    puts(buf);
#else
    if(custom_print_cb) custom_print_cb(buf);
#endif
}
```

It should be
```c
void lv_log(const char * buf)
{
#if LV_LOG_PRINTF
    puts(buf);
#endif
    if(custom_print_cb) custom_print_cb(buf);
}
```

So allow printf and custom_print_cb to work together. IMO it's the cleanest solution.

mysterywolf:
OK I got it😅

kisvegabor:
Merged, thank you!

PR #2815, opened 2021-11-20
test(txt): add new test cases for arabic string

DennisSHCheung:
Created a test to challenge Arabic conjunction calculation

### Checkpoints


kisvegabor:
@cmumford It seems build tests try to compile the test cases too. I think it should be avoided because the tests are written for a specific build config. Do you agree with that?

cmumford:
@kisvegabor yes, all tests are built and run for each PR. I think this test is maybe missing a header file as this looks like `"src/misc/lv_txt.h"` needs to be included.  Can you add more info on what you're suggesting?

kisvegabor:
I mean now we have ["build tests"](https://github.com/lvgl/lvgl/blob/a2e4dfdce21a990d03e72a1128566ebc7110fb3d/tests/CMakeLists.txt#L219-L228) for various configs and ["unit tests"](https://github.com/lvgl/lvgl/blob/a2e4dfdce21a990d03e72a1128566ebc7110fb3d/tests/CMakeLists.txt#L229-L231) with a specific (everything is enabled) config.

I think the problem in this case is that the new test in the PR uses a function (`_lv_txt_ap_proc`) that is disabled in some "build test". Therefore it'd be important to build the unit tests only with `OPTIONS_TEST` as the unit tests might contain function calls that might not be available in "build tests".

embeddedt:
Isn't it easier to just wrap the entire function declaration in `LV_USE_ARABIC_PERSIAN_CHARS`?

kisvegabor:
> Isn't it easier to just wrap the entire function declaration in LV_USE_ARABIC_PERSIAN_CHARS?

Good idea!

DennisSHCheung:
Sorry what needs to be done for _lv_txt_ap_proc to be enabled?

kisvegabor:
> Sorry what needs to be done for _lv_txt_ap_proc to be enabled?

We run CI tests with different `lv_conf.h` settings to see LVGL can be built with these different settings. But in not all cases we have `LV_USE_ARABIC_PERSIAN_CHARS` is enabled. 

So please wrap the test into an
```c
#if LV_BUILD_TEST
#include "../lvgl.h"

...


#endif  /*#if LV_USE_ARABIC_PERSIAN_CHARS*/
#endif /*#if LV_BUILD_TEST*/
```

DennisSHCheung:
After doing what you mentioned, the error didn't go away

Running with ./tests/main.py test:
image

Running with ./tests/main.py build:
image
which I assume is because like you said that LV_USE_ARABIC_PERSIAN_CHARS is not always enabled

Any suggestions?

kisvegabor:
Hi,

Sorry for the late reply. I've just pushed a commit to make it build, but the test itself failed for me locally.
```
./src/test_cases/test_txt_ap.c:24:test_txt_ap_proc_detect_invalid_conjuction_with_fatha:FAIL: Expected -17 Was 1604
```

DennisSHCheung:
Thanks for making it to build. Yes, I made it intentionally to fail to see if it could actually build.

kisvegabor:
Okay, great! :slightly_smiling_face: 

I needed to push an other fix.

PR #2819, opened 2021-11-21
improve lv_obj_fade_in/out

Miguel0101:
Improve `lv_obj_fade_in/out`


kisvegabor:
Makes sense, thank you!

PR #2822, opened 2021-11-21
fix(format): run code-format.sh

xiaoxiang781216:
### Checkpoints

### DIFF
--- a/scripts/built_in_font/generate_all.py
+++ b/scripts/built_in_font/generate_all.py
@@ -84,5 +84,5 @@ print("\nGenerating 16 px unscii")
 os.system("lv_font_conv --no-compress --no-prefilter --bpp 1 --size 16 --font unscii-8.ttf -r 0x20-0x7F --format lvgl -o lv_font_unscii_16.c --force-fast-kern-format")
 
 os.system('sed -i \'s|#include "lvgl/lvgl.h"|#include "../../lvgl.h"|\' lv_font_*.c')
-os.system('astyle --options=../code-format.cfg "lv_font_*.c"')
+os.system('astyle --ignore-exclude-errors --options=../code-format.cfg "lv_font_*.c"')
 os.system('mv lv_font_*.c ../../src/font/')
--- a/scripts/code-format.cfg
+++ b/scripts/code-format.cfg
@@ -23,3 +23,12 @@
 --suffix=none
 --preserve-date
 --formatted
+--exclude=../src/extra/libs/gif/gifdec.c
+--exclude=../src/extra/libs/gif/gifdec.h
+--exclude=../src/extra/libs/png/lodepng.c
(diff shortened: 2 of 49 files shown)


kisvegabor:
I think we should do some automation on it. 
@embeddedt could you add an action similar to the `lv_conf_internal.h` checker? It'd throw an error if `git diff` is not empty after running the formatter.

We need to run it manually once, but now it breaks all the open PR. I'm checking them now and merge as much as possible. 

Probably `code-format.sh` should be rather `code-format.py` for cross-platform compatibility.

xiaoxiang781216:
> I think we should do some automation on it. @embeddedt could you add an action similar to the `lv_conf_internal.h` checker? It'd throw an error if `git diff` is not empty after running the formatter.
> 

Good idea.

> We need to run it manually once, but now it breaks all the open PR. I'm checking them now and merge as much as possible.
>

After this patch, only some 3rd party files still contain the style warning:
```
        modified:   src/extra/libs/gif/gifdec.c
        modified:   src/extra/libs/gif/gifdec.h
        modified:   src/extra/libs/png/lodepng.c
        modified:   src/extra/libs/png/lodepng.h
        modified:   src/extra/libs/qrcode/qrcodegen.c
        modified:   src/extra/libs/qrcode/qrcodegen.h
        modified:   src/extra/libs/sjpg/tjpgd.c
        modified:   src/extra/libs/sjpg/tjpgd.h
        modified:   src/extra/libs/sjpg/tjpgdcnf.h
```
I left these files to make the future upgrade easier.

> Probably `code-format.sh` should be rather `code-format.py` for cross-platform compatibility.

Yes.

kisvegabor:
> After this patch, only some 3rd party files still contain the style warning:

What do you mean by "style warning"?

xiaoxiang781216:
> > After this patch, only some 3rd party files still contain the style warning:
> 
> What do you mean by "style warning"?

code-format.sh will change these files, because I exclude these files from PR.

kisvegabor:
I see. But as these are 3rd party files they should be never formatted.  So we should exclude them in `code-format.cfg`. What do you think?

xiaoxiang781216:
Yes, I think so. That's why I left them untouch.

xiaoxiang781216:
> I see. But as these are 3rd party files they should be never formatted. So we should exclude them in `code-format.cfg`. What do you think?

Done.

xiaoxiang781216:
After this patch get merged, we can add a new ci check to ensure there is no coding style violation.

kisvegabor:
Great, thanks!

embeddedt:
I've added a CI check. It will pass as long as the formatting matches `astyle`'s and will print the diff in the log otherwise.

kisvegabor:
Perfect, thanks!

PR #2828, opened 2021-11-23
fix(doc): spelling

ckielstra:
Just a bunch of simple spelling fixes

And one comment referenced the lv_font_codeCompare function which was renamed to unicode_list_compare. Now the comment is renamed too.


kisvegabor:
Thank you! :slightly_smiling_face:

PR #2835, opened 2021-11-25
fix(keyboard) add missing functions

amirgon:
Related: https://forum.lvgl.io/t/keyobard-get-active-btn-text/5683/9?u=amirgon


kisvegabor:
Thanks!

PR #2839, opened 2021-11-26
add lv_textarea_set_key

Miguel0101:
Add `lv_textarea_set_key` to enable/disable `LV_EVENT_KEY` on `lv_textarea`.

### DIFF
--- a/src/widgets/lv_textarea.c
+++ b/src/widgets/lv_textarea.c
@@ -899,7 +899,7 @@ static void lv_textarea_event(const lv_obj_class_t * class_p, lv_event_t * e)
     }
 }
 
-static void key_event_cb(lv_event_t *e)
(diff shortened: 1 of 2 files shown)


kisvegabor:
What is the motivation behind this update?

Miguel0101:
> What is the motivation behind this update?

Set `LV_EVENT_KEY` on/off (for users using physical keyboard or others indevs that use keys) on `lv_textarea`.
But now I see it doesn't make sense :(

kisvegabor:
> Set LV_EVENT_KEY on/off (for users using physical keyboard or others indevs that use keys) on lv_textarea.

In this case, you can create an indev for the keyboard. E.g. you can use your keyboard to write into the text areas of the [online example](https://docs.lvgl.io/master/widgets/core/textarea.html#simple-text-area).

PR #2840, opened 2021-11-26
migrated to use new backend architecture

mariotaku:

This is a continuation for SDL GPU counterparts of #2803 

### Checkpoints


mariotaku:
This does work now, but to make texture based renders work properly, I have following change requests:

1. Provide absolute coordinate for `blend_*` functions, as they doesn't care about `lv_disp_draw_buf_t`.
2. Let backend handle image opening/closing, because on SDL, bitmaps are only opened once, then bitmap loaded into GPU and `src` is used as cache key. Current architecture will force SDL renderer creating/destroying textures frequently, which will cause heavy performance penalty.

If you are OK with that, I'll try to make some changes to this design based on my need. Also as SDL Renderer is quite primitive, I'm planning to provide basic implementation only and implementing a new renderer with some more powerful solution (like nanovg based renderer)

kisvegabor:
> Provide absolute coordinate for blend_* functions, as they doesn't care about lv_disp_draw_buf_t.

Can you make use of [disp_drv->direct_mode=1](https://docs.lvgl.io/master/porting/display.html#direct-mode)?

> Let backend handle image opening/closing, because on SDL, bitmaps are only opened once, then bitmap loaded into GPU and src is used as cache key. Current architecture will force SDL renderer creating/destroying textures frequently, which will cause heavy performance penalty.

Ok, let's call the top-level image drawer from the backend. 

>  I'm planning to provide basic implementation only and implementing a new renderer with some more powerful solution (like nanovg based renderer)

The more backend options are better we should avoid overlapping solutions. Do you think NanoVG is different enough from the SDL backend?
How complicated would it be to use e.g. OpenGL ES directly?

kisvegabor:
A minor thing: I think it'd be better to move `sdl` to `draw/sdl` from `gpu/sdl` (similarly to `draw/sw`) and use names like `lv_draw_sdl_...`. It's because it's not necessary that a new backend uses any kind of GPU. 

I'll do it with STM nd NXP GPUs too.

mariotaku:
@kisvegabor I see nanovg as an wrapper to OpenGL (standard and ES version). Drawing with OpenGL ES directly should have more performance but that's beyond my ability.

kisvegabor:
>  I see nanovg as an wrapper to OpenGL (standard and ES version). Drawing with OpenGL ES directly should have more performance but that's beyond my ability.

I see, it might make sense on systems where there is no SDL but there is Open GL (ES). 

Now, I'm implementing the class-based mechanism for backend handling.

kisvegabor:
Let me know if it's ready to merge.

mariotaku:
Yes please. I'll make some changes to driver later as well.

kisvegabor:
Great, merging.  I'm trying to find the right place for the components in the new architecture. 

One thing that I don't have a good idea for is how the call e.g. `super->draw_rect(...)`.  Let say we have:
```c
void my_draw_rect(lv_draw_t * ctx, lv_draw_rect_dsc_t * dsc)
{
   /*Option A: too long and error prone*/
  lv_draw_class_t * base = my_draw_class.base_class_p;
  while(base && base->draw_rect == NULL) base = base->base_class_p; //Find a base the implements draw_rect

  if(base) base->draw_rect(ctx, dsc);

  /*Option B: needs a function like this for every method (we already have `lv_obj_event_base` for obj events)*/
  lv_draw_rect_base(&my_draw_class, ctx, dsc);

  /*Option C: not really OOP like*/
  return LV_DRAW_RES_CALL_BASE;
  return LV_DRAW_RES_READY;
}
```

What do you think?

embeddedt:
Option A can probably be turned into a macro if the length is a concern.

mariotaku:
I think A or B should be good. BTW Please look forward to my nanovg based renderer :)

kisvegabor:
> Option A can probably be turned into a macro if the length is a concern.

Good idea!

---

@mariotaku 
Now I store a pointer to a "class descriptor" (similar to `lv_slider_class` but for drawing) in `lv_disp_drv_t` and create an instance from it for rendering. My vision is that:
- the instance of the draw backend class is actually the drawing context 
- the drawing context contains the destination buffers (aka frame buffer) and its current coordinates (i.e. the destination buffer is at (20;30) (200;230) coordinates now)
- It'd be the most flexible to say: "Hi draw backend, here is a destination buffer, initialize yourself to draw into it. " If it works so simply we can create a new draw context on each frame or even on each dirty area (render them in parallel somehow with multiple drawers?) 

Could the "on each dirty area" method work for SDL?

kisvegabor:
Thinking about it more "on each dirty area" would be a too large step at this point. So I wonder whether we should create an instance from the drawer once in display init or on each frame.

mariotaku:
On SDL and upcoming NanoVG backend, my implementation draws on a write only texture. Then draw this texture on screen.
As for the draw buffer, it's actually not used. But drawing on dirty area should work.
In SDL we can set clip rect for any render operation, and in OpenGL we can use scissor feature.

kisvegabor:
Can we say that the whole buffering mechanism of LVGL is related to the backend? I.e. drawing in tiles with small draw buffer(s) and calling `flush_cb` in the end is only how the Sw renderer works. The SDL rendered doesn't need anything from that. 
Cairo would be different too.

kisvegabor:
In other words, what we called `lv_disp_drv_t` so far (with all the special callbacks) is actually a drawing backend?

mariotaku:
> In other words, what we called `lv_disp_drv_t` so far (with all the special callbacks) is actually a drawing backend?

I think for the backend/driver it can be interpreted in following way:

Backend: draws everything to one texture
Drivers: Draw that texture on screen

I'm not an expert to computer graphics, so my explanation might be inaccurate. I'll think how to explain properly.

kisvegabor:
> Backend: draws everything to one texture
Drivers: Draw that texture on screen

Makes sense. 

After trying a lot of ways I concluded I need to drop to strictly follow how classes would work in C++ as it looked very ugly in C. I think I could come up with something simple and C-ish yet extendable, but need some time to put everything together and show it.

zeroxer:
> I think A or B should be good. BTW Please look forward to my nanovg based renderer :)

Is there any example I can try using nanovg backend?

mariotaku:
> > I think A or B should be good. BTW Please look forward to my nanovg based renderer :)
> 
> Is there any example I can try using nanovg backend?

The performance was not good at all that I eventually abandoned it although it's kinda working: https://github.com/mariotaku/lvgl-demo-nanovg/commit/11a90080cedbb43a39d856ce1e1e966cf19b7376

zeroxer:
My [nanovg backend (on top of bgfx)](https://github.com/bkaradzic/bgfx/tree/master/examples/common/nanovg) works on PC with GPU, so the performance maybe fine. I'm try to use lvgl as in-game gui framework in my own game engine. Does your lvgl nanovg backend implementation work for this usage?

mariotaku:
@zeroxer I'm not sure. You can try run the demo I just made public, but I won't make any further changes. lvgl changes: https://github.com/mariotaku/lvgl/tree/nanovg-backend

zeroxer:
Thanks. I'll try.

kisvegabor:
Let me bring up an other topic:
I think v9 is ready to make some experiments with OpenGL/SDL again.  

To keep things simpler I would like to follow a different approach then in v8. It would be great to create a generic layer (or so) which uses SW render to draw the primitives, cache them in OpenGL/SDL as textures, and blend just blend the cached textures. I imagine it as a simple API which can be used by OpenGL, OpenGL ES, NanoVG, SDL, Vulkan, etc.

In some cases where we are quite sure that the GPU can handle a draw task very fast we can ignore this caching mechanism. For example:
- Image drawing
- Image transformation
- Rectangle fill
- Triangle drawing


What do you think about it? If you agree with this direction, are you interested in making some experiments with creating this generic layer?

FYI, the summary of the changes in LVGL v9 in general can be found [here](https://github.com/lvgl/lvgl/issues/4011) and the description of the new rendering architecture is [here](https://github.com/lvgl/lvgl/issues/4016#issue-1599634031).

zeroxer:
Sounds fun, but I don't have enough time for it.

PR #2845, opened 2021-11-29
test(bar) Add unit tests

C47D:

### Checkpoints

### DIFF
--- a/tests/src/test_cases/test_bar.c
+++ b/tests/src/test_cases/test_bar.c
@@ -1,0 +1,26 @@
+#if LV_BUILD_TEST
+#include "../lvgl.h"
+
+#include "unity/unity.h"
+
+static lv_obj_t *active_screen = NULL;
+static lv_obj_t *bar = NULL;
+
+void setUp(void)
+{
+    active_screen = lv_scr_act();
+    bar = lv_bar_create(active_screen);
+}
+
+void tearDown(void)
+{
+}
+
+void test_bar_should_have_valid_default_attributes(void)
+{
+    TEST_ASSERT_EQUAL(1, lv_bar_get_min_value(bar));
+    TEST_ASSERT_EQUAL(100, lv_bar_get_max_value(bar));
+    TEST_ASSERT_EQUAL(LV_BAR_MODE_NORMAL, lv_bar_get_mode(bar));
+}
+
+#endif
--- a/tests/src/test_cases/test_bar.c
+++ b/tests/src/test_cases/test_bar.c
@@ -18,7 +18,7 @@ void tearDown(void)
 
 void test_bar_should_have_valid_default_attributes(void)
 {
-    TEST_ASSERT_EQUAL(1, lv_bar_get_min_value(bar));
+    TEST_ASSERT_EQUAL(0, lv_bar_get_min_value(bar));
     TEST_ASSERT_EQUAL(100, lv_bar_get_max_value(bar));
     TEST_ASSERT_EQUAL(LV_BAR_MODE_NORMAL, lv_bar_get_mode(bar));
 }
--- a/tests/src/test_cases/test_bar.c
+++ b/tests/src/test_cases/test_bar.c
@@ -3,6 +3,8 @@
 
 #include "unity/unity.h"
 
+#include "lv_test_indev.h"
+
 static lv_obj_t *active_screen = NULL;
 static lv_obj_t *bar = NULL;
 
@@ -23,4 +25,21 @@ void test_bar_should_have_valid_default_attributes(void)
     TEST_ASSERT_EQUAL(LV_BAR_MODE_NORMAL, lv_bar_get_mode(bar));
 }
 
+void test_bar_indicator_width_should_track_bar_value(void)
+{
+    lv_bar_t * bar_ptr = (lv_bar_t *) bar;
+
+    /* By default the bar is horizontal, so we get the indicator width */
+    int32_t indicator_width = lv_area_get_width(&bar_ptr->indic_area);
+
+    /* If x1 == x2 then width == 1 */
+    TEST_ASSERT_EQUAL_INT32(1, indicator_width);
+
+    /* Set a new value */
+    lv_bar_set_value(bar, 10, LV_ANIM_OFF);
+    lv_test_indev_wait(50);
+
+    indicator_width = lv_area_get_width(&bar_ptr->indic_area);
+    TEST_ASSERT_EQUAL_INT32(27 /* Value found by debugging */, indicator_width);
+}
 #endif
--- a/tests/src/test_cases/test_bar.c
+++ b/tests/src/test_cases/test_bar.c
@@ -81,6 +81,61 @@ void test_bar_should_update_indicator_right_coordinate_based_on_bar_value(void)
     TEST_ASSERT_EQUAL_INT32(expected_coord, actual_coord);
 }
 
+/*
+ * Bar has two parts, main and indicator, coordinates of the latter are
+ * calculated based on:
+ * - Bar size
+ * - Bar (main part) padding
+ * - Bar value
+ * - Bar coordinates
+ * - Bar base direction
+ * See Boxing model in docs for reference.
+ *
+ * Bar properties assumed:
+ * - mode: LV_BAR_MODE_NORMAL
+ * - min value: 0
+ * - max value: 100
+ */
+void test_bar_rtl_should_update_indicator_left_coordinate_based_on_bar_value(void)
+{
+    lv_bar_t * bar_ptr = (lv_bar_t *) bar;
+
+    static lv_style_t bar_style;
+
+    const lv_coord_t style_padding = 5u;
+    const lv_coord_t bar_width = 200u;
+    const lv_coord_t bar_height = 20u;
+    int32_t bar_value = 10u;
+
+    lv_style_init(&bar_style);
+    lv_style_set_pad_all(&bar_style, style_padding);
+
+    /* Setup new style */
+    lv_obj_remove_style_all(bar);
+    lv_obj_add_style(bar, &bar_style, LV_PART_MAIN);
+
+    /* Set properties */
+    lv_obj_set_size(bar, bar_width, bar_height);
+    lv_bar_set_value(bar, bar_value, LV_ANIM_OFF);
+    lv_obj_set_style_base_dir(bar, LV_BASE_DIR_RTL, 0);
+
+    /* FIXME: Remove wait */
+    lv_test_indev_wait(50);
+
+    int32_t actual_coord = bar_ptr->indic_area.x1;
+
+    /* Calculate current indicator width */
+    lv_coord_t bar_max_value = lv_bar_get_max_value(bar);
+    lv_coord_t indicator_part_width = lv_obj_get_content_width(bar);
+    lv_coord_t right_padding = lv_obj_get_style_pad_right(bar, LV_PART_MAIN);
+    int32_t indicator_width = (bar_value * indicator_part_width) / bar_max_value;
+
+    int32_t expected_coord = (bar_width - right_padding) - indicator_width;
+    expected_coord -= 1;
+
+    TEST_ASSERT_EQUAL_INT32(expected_coord, actual_coord);
+}
+
 void test_bar_indicator_area_should_get_smaller_when_padding_is_increased(void)
 {
     lv_bar_t * bar_ptr = (lv_bar_t *) bar;
--- a/tests/src/test_cases/test_bar.c
+++ b/tests/src/test_cases/test_bar.c
@@ -128,4 +128,14 @@ void test_bar_start_value_should_only_change_when_in_range_mode(void)
     TEST_ASSERT_EQUAL_INT32(new_start_value, lv_bar_get_start_value(bar));
 }
 
+void test_bar_start_value_should_be_smaller_than_current_value_in_range_mode(void)
+{
+    /* Set bar in RANGE mode so we can edit the start value */
+    lv_bar_set_mode(bar, LV_BAR_MODE_RANGE);
+    lv_bar_set_value(bar, 50, LV_ANIM_OFF);
+    lv_bar_set_start_value(bar, 100u, LV_ANIM_OFF);
+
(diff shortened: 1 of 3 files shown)


C47D:
@kisvegabor @embeddedt The `draw_indic` is pretty huge, I think it would be good to extract some of the code as static functions.
Also, the default min value is 0, instead of the documented 1.

embeddedt:
I fixed the documentation. 0 makes more sense and we can't change it anyways without breaking existing applications.

C47D:
I think we should write an initial test plan before writing the tests, once the test list is initially agreed upon we can start writing the tests. We can add more items to the list later as we develop the test.

I've also noticed that sometimes I end up testing implementation instead of behavior, maybe because my lack of knowledge of LVGL internals.

Here's the proposal of the initial test plan for the bar widget:

- [x] Documented defaults values should match the implementation.
- [x] Width of the indicator should be set according to the current value of the bar.
~~- [ ] Vertical bar should be created when the width of the object is smaller than it's height.~~
- [x] Bar in `LV_BAR_MODE_RANGE` mode start value should be smaller than the end value.

Added tests cases - December 6

- [x] Padding on `LV_PART_MAIN` should make the indicator area smaller.
- [x] `indic_area` should be updated when bar is in `LV_BAR_SYMMETRICAL` mode.
- [x] `indic_area` should be updated when bar has LTR and RTL base direction.
- [ ] ~~`indic_area` should be updated when bar changes from `LV_BAR_MODE_NORMAL` to mode `LV_BAR_MODE_SYMMETRICAL`~~.

Added tests cases - January 6

- [x] Bar start value can only be changed when bar mode is `LV_BAR_MODE_RANGE`.
- [x] When in `LV_BAR_MODE_SYMMETRICAL` mode, draw the indicator from the zero value to current value. Requires a negative minimum range and positive maximum range.

Some behaviors I don't currently know how to test:
- [ ] ~~Bar in `LV_BAR_SYMMETRICAL` mode (?)~~
- [ ] ~~LTR and RTL base direction (?)~~
- [ ] ~~If we have a bar in X mode and changed to Y mode, how the parameters should behave?~~

Let me know if you want to add more items.

What do you think @kisvegabor @embeddedt ?

kisvegabor:
> The draw_indic is pretty huge, I think it would be good to extract some of the code as static functions.

It's really long but there are only a few branches so it should be doable to get good coverage on it. If it really looks too complex, I can help to separate it into smaller functions.  

> Here's the proposal of the initial test plan for the bar widget:

Looks good. It should cover most of the tricky things.

> Bar in LV_BAR_SYMMETRICAL mode 
> LTR and RTL base direction (?)
For this and other indicator-related tests too, the `indic_area` property can be used. 

>  If we have a bar in X mode and changed to Y mode, how the parameters should behave?

Normally, only `indic_area` should change here too. 

It might be worth testing that padding on `LV_PART_MAIN` should make the indicator area smaller.

C47D:
Hi, I've had a couple of really busy weeks, will try to work on this whenever possible.

kisvegabor:
No worries, take your time :slightly_smiling_face:

C47D:
Not stale

C47D:
Hi @embeddedt @kisvegabor,

I'm trying to add a test for this statement in 0f04864
imagen

I have some questions:

- `indic_area` member of `lv_bar_t` is not initialized in the constructor, should it?
- I've hard-coded the values found by debugging on the test assertions, but I don't understand why is the indicator width 27 when setting the value to 10, I guess it should be around 10 (?)

kisvegabor:
> indic_area member of lv_bar_t is not initialized in the constructor, should it?

Not critical, but would be better to set it to all zero.

> I've hard-coded the values found by debugging on the test assertions, but I don't understand why is the indicator width 27 when setting the value to 10, I guess it should be around 10 (?)

If the bar's width is let's say 200 px, the indicators max width is around 180 px (due to the paddings). If the range is 0-100, value  = 10 means 10% of the max width -> 18 px.

I think in the test it would be great to set the padding too to make the test independent from the styles coming from the theme.

C47D:
> Not critical, but would be better to set it to all zero.

OK, will add a commit for that.

> If the bar's width is let's say 200 px, the indicators max width is around 180 px (due to the paddings). If the range is 0-100, value = 10 means 10% of the max width -> 18 px.
I think in the test it would be great to set the padding too to make the test independent from the styles coming from the theme.

Sounds reasonable, thanks for the explanation, I will add it as a comment above the test.

C47D:
@kisvegabor there's 1px missing from the calculation, calculated indicator width is 19, actual is 20, but we're getting closer, let me know if there's something missing in the calculation.

Some additional questions:

- Should we add an else branch at the end of this?
https://github.com/lvgl/lvgl/blob/e7ac0e41988e5fda772e17292c05d65bcaf58394/src/widgets/lv_bar.c#L271-L279

- This test seems to cover horizontal bar, I assume we should do a similar one for vertical bars, and add a comment about it in the docs (the picture I previously only talks about hor bars).
- Do you think we should add the test plan at the top of the file?

C47D:
Based on the following text on the modes section, I'm assuming the start value of the bar can only be changed when the bar mode is `LV_BAR_MODE_RANGE`. Is this assumption correct? `test_bar_start_value_should_only_change_when_in_range_mode` is testing that assumption.

imagen

If so, adding a check for the mode at the beginning of `lv_bar_set_start_value` might be a good idea.

```diff
diff --git a/src/widgets/lv_bar.c b/src/widgets/lv_bar.c
index ae1dce8e6..0da2a9875 100644
--- a/src/widgets/lv_bar.c
+++ b/src/widgets/lv_bar.c
@@ -107,6 +107,10 @@ void lv_bar_set_start_value(lv_obj_t * obj, int32_t value, lv_anim_enable_t anim

     lv_bar_t * bar = (lv_bar_t *)obj;

+    if(bar->mode != LV_BAR_MODE_RANGE) {
+        return;
+    }
+
     value = LV_CLAMP(bar->min_value, value, bar->max_value);
     value = value > bar->cur_value ? bar->cur_value : value; /*Can't be greater than the right value*/

```

C47D:
imagen

Differs from the code
https://github.com/lvgl/lvgl/blob/8bbbdf7c07b4eac1c63547c4b8019fd7512272dc/src/widgets/lv_bar.c#L111

should the docs say `The start value always has to be smaller than the current value.` ?

`test_bar_start_value_should_be_smaller_than_current_value_in_range_mode` tests that.

C47D:
For this test case

> When in LV_BAR_MODE_SYMMETRICAL mode, draw the indicator from the zero value to current value. Requires a negative minimum range and positive maximum range.

I think we could add a check on `lv_bar_set_value` to return early if the value is negative and we're not in `LV_BAR_MODE_SYMMETRICAL` mode. Another check would be checking the `min_value` is negative. What do you think?

That or indicate in the docs, that setting a negative value will truncate to `min_value`, e.g., setting value to -30 will truncate to 0, when min_value is 0.

C47D:
Should we test that max is greater than min in `lv_bar_set_range` or add a comment in the docs?

image

C47D:
Any feedback @kisvegabor @embeddedt ?

kisvegabor:
> Should we add an else branch at the end of this?

I don't think so, it checks the limits in these 2 cases. Else the values are fine as they are.

> This test seems to cover horizontal bar, I assume we should do a similar one for vertical bars, and add a comment about it in the docs (the picture I previously only talks about hor bars).

Yes, ultimately, we should have tests for the vertical bar too.

> Do you think we should add the test plan at the top of the file?

Yes, it it seems like a good practice for the upcoming tests too.

> Based on the following text on the modes section, I'm assuming the start value of the bar can only be changed when the bar mode is LV_BAR_MODE_RANGE. Is this assumption correct? 

Yes, correct.

> If so, adding a check for the mode at the beginning of lv_bar_set_start_value might be a good idea.

I agree.

> should the docs say The start value always has to be smaller than the current value. ?

The name `cur_value` makes sens in most of the cases, but really ambiguous in case of range mode (where is should be called `end_value`). As it can have only one name, and `cur_value` fits in most of the cases, I'd keep it.
I think the docs is correct as there is no "current value" in range mode, only "start value" and "end value".

> I think we could add a check on lv_bar_set_value to return early if the value is negative and we're not in LV_BAR_MODE_SYMMETRICAL mode. Another check would be checking the min_value is negative. What do you think?

Having negative `min_value` is valid. Imagine a thermometer with -30C..120C range.
I'd rather add `LV_LOG_WARN` if `min_value` is not negative or `mas_value` is not positive in case of symmetrical mode.  A warning log is hard to test though.
 
 > Should we test that max is greater than min in lv_bar_set_range or add a comment in the docs?

Do you mean if min=30 and you try to set max=20, max will be 30 instead? If so, I agree. Once min and max can be set only together with `lv_bar_set_range` we need to decide which value keep and which on the change. I think neither is better than the other, but keeping min feels better to me.

C47D:
@kisvegabor Thanks for the reply, there's one test failing, maybe I'm calculating the expected indicator area wrong, any thoughts? b1b4161

> @kisvegabor there's 1px missing from the calculation, calculated indicator width is 19, actual is 20, but we're getting closer, let me know if there's something missing in the calculation.

kisvegabor:
>  expected_width = bar_value * (bar_width - sides_padding) / bar_max_value;

Actually it doesn't tell the expected width but the expected right x coordinate (assuming the left coordinate is 0). Imagine that if `bar_value` would be `bar_max_value` then `expected_width = 200`. But if `x1=0; x2=200` width is `201`. It's because width calculation counts the start and end pixel too. E.g. `x1=5`, `x2=7` has `w=3`. Or  `x1=10`, `x2=10` has `w=1`.
 See how [lv_area_get_width()](https://github.com/lvgl/lvgl/blob/d6c0d791f70af8a04bf9a7289c15823e89a78cca/src/misc/lv_area.h#L123-L126) works. 

So in the specific case `19` is correct and if `indic_area.x1=100` `indic_area.x2=119` but its width is 20.

In short all your need to do is:
```c
expected_width = bar_value * (bar_width - sides_padding) / bar_max_value + 1;
```

C47D:
Great, thanks for the explanation. I will update the test!

kisvegabor:
Please try to rebase to master. Hopefully it fixes the failing Micropython CI.
```c
git rebase master
git push -f
```

C47D:
@kisvegabor I've rewritten the indicator calculation test from scratch in 216e6ab, I would like to know what do you think about it.

C47D:
8b9fd5a tries to cover the following case, what do you think?

> When in LV_BAR_MODE_SYMMETRICAL mode, draw the indicator from the zero value to current value. Requires a negative minimum range and positive maximum range.

C47D:
I've marked it as ready for review so we can have a general idea of what do we have and what tests are missing.

kisvegabor:
> 8b9fd5a tries to cover the following case, what do you think?

Not pixel perfect testing as for the normal case, but I think it's good enough.

> I've marked it as ready for review so we can have a general idea of what do we have and what tests are missing.

Only these 2 left?
> indic_area should be updated when bar has LTR and RTL base direction.
 indic_area should be updated when bar changes from LV_BAR_MODE_NORMAL to mode LV_BAR_MODE_SYMMETRICAL.

C47D:
Yes, those two are missing. I did drafted one test for base directions but I couldn't get it working.

tests/src/test_cases/test_bar.c
```diff
@@ -0,0 +1,198 @@
...
+ * - Bar value
+ * - Bar coordinates
+ * - Bar base direction
+ * See Boxing model in docs for reference.
+ *
+ * Bar properties assumed:
+ * - mode: LV_BAR_MODE_NORMAL
+ * - min value: 0
+ * - max value: 100
+ * - base direction: RTL
```
kisvegabor: Isn't it LTR?

tests/src/test_cases/test_bar.c
```diff
@@ -0,0 +1,198 @@
...
+    lv_bar_set_value(bar, new_value, LV_ANIM_OFF);
+    TEST_ASSERT_EQUAL_INT32(min_value, lv_bar_get_value(bar));
+}
+
+/** When in symmetrical mode, the bar indicator has to be drawn towards the min
+ * range value. Requires a negative min range value and a positive max range
+ * value.
+ *
+ * Bar properties assumed:
+ * - base direction: RTL
```
kisvegabor: Here too.

kisvegabor:
> Yes, those two are missing. I did drafted one test for base directions but I couldn't get it working.

Let me know if I can help.

tests/src/test_cases/test_bar.c
C47D: Fixed in d7789b3

tests/src/test_cases/test_bar.c
C47D: Fixed in d7789b3

C47D:
> > Yes, those two are missing. I did drafted one test for base directions but I couldn't get it working.
> 
> Let me know if I can help.

Sure, when changing the base direction to RTL, is the x1 point of the indic_area the one updated while x2 stays at the right-mst side of the indicator?

C47D:
> indic_area should be updated when bar changes from LV_BAR_MODE_NORMAL to mode LV_BAR_MODE_SYMMETRICAL.

I think this no longer apply, the indic_area should not being updated when changing modes. What do you think?

kisvegabor:
> Sure, when changing the base direction to RTL, is the x1 point of the indic_area the one updated while x2 stays at the right-mst side of the indicator?

Yes. Basically, it looks like the min value is on the right and the max value is on the left.

> I think this no longer apply, the indic_area should not being updated when changing modes. What do you think?

True, as it's updated during rendering, so the previous mode and the order of settings shouldn't matter.

C47D:
> Yes. Basically, it looks like the min value is on the right and the max value is on the left.

Tested in 5c3f7a0

I think it's now ready to be reviewed and merged

kisvegabor:
Looks good. 
Let me know if it's ready to merge.

C47D:
It is ready to merge.

kisvegabor:
Great, merging then. :slightly_smiling_face:  
Thank you very much!

PR #2850, opened 2021-11-29
fix(lv_hal_indev): update lv_indev_drv_update to free the read_timer

geert-KLA-BE:
The read timer was not freed and recreated. The display was also not checked to be NULL.

### Checkpoints

### DIFF
--- a/src/hal/lv_hal_indev.c
+++ b/src/hal/lv_hal_indev.c
@@ -112,7 +112,7 @@ void lv_indev_drv_update(lv_indev_t * indev, lv_indev_drv_t * new_drv)
     lv_timer_del(indev->driver->read_timer);
 
     LV_ASSERT_NULL(new_drv);
-    if (new_drv->disp == NULL) {
+    if(new_drv->disp == NULL) {
         new_drv->disp = lv_disp_get_default();
     }
     if(new_drv->disp == NULL) {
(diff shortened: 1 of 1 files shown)


kisvegabor:
Looks good, thanks. 
Just please run code-format.py from the scripts folder.

geert-KLA-BE:
Is it possible to run this on windows? Since that is my dev. environment.

kisvegabor:
Yes, astyle can be installed on Windows too, but it seems you managed to run it :slightly_smiling_face: 

Thank you, merging.

geert-KLA-BE:
I did it manually still :).
You are welcome.

embeddedt:
I suppose I could have lvgl-bot open a PR to the contributor's fork with the diff applied... but that could get spammy if they keep making commits and it opens a new PR for each one.

kisvegabor:
Can't the bot just push a "format" commit? We squash the commits of PRs anyway so it won't ass any noise.

embeddedt:
I suppose it could... but:

* I'm not sure I like the idea of pushing to contributors' repositories without them expecting it, especially for those who just use their `master` branch instead of making a dedicated branch for the change.
* If they don't force push, there could end up being multiple `style(format) fix formatting` commits every time they apply a review comment, if the formatting is incorrect again. It would clutter the commit history for those who split their PR into incremental commits. Although, they end up having to make a commit like that manually, so perhaps this point is moot.

kisvegabor:
I think it's true for both points that either the contributor or the bot needs to add the "format" commit. It's really might look intrusive to make the bot commit directly to the master branch but IMO it's an acceptable discomfort. In the end, the PR will the squashed anyway so when he updates again the history will be clean again.

embeddedt:
I had another idea: how about pushing the commit if the contributor writes "@lvgl-bot format"? It's less intrusive and still very intuitive. We can add a note about it in the PR checklist template.

kisvegabor:
Ohh, it would be amazing!

PR #2856, opened 2021-11-30
Add support for ARGB32 to RGBA5658 format conversion required by LVGL

### LINKED ISSUE #3: Missing header file img_conf.h
img_add.c reporting error of a missing file
`#include "img_conf.h"`

I think there is no template file img_conf_tmpl.h which can tell what goes in img_conf.h

X-Ryl669:

This fixes #3.

rlottie renders as ARGB32 in its buffer, and not as what LV_COLOR is set (likely RGB565).
So account for this change and convert the format before it's sent to LVGL for blending.

This allows to display correct rlottie on RGB565 displays.

### DIFF
--- a/src/extra/libs/rlottie/lv_rlottie.c
+++ b/src/extra/libs/rlottie/lv_rlottie.c
@@ -14,6 +14,7 @@
 *      DEFINES
 *********************/
 #define MY_CLASS &lv_rlottie_class
+#define LV_ARGB32   32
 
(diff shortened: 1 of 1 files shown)


X-Ryl669:
Please notice that if merging #2857, then this PR can be ignored.

X-Ryl669:
#2857 was merged, this is not required anymore.

PR #2857, opened 2021-11-30
Add animation control for rlottie

### LINKED ISSUE #4: Improvement in memory management
If a system already provide malloc/free functions then there is no need to use misc/mem module. This will save lot of code space. You can move memory management to HAL instead so people can either write their own dm_* functions or enable USE_DYN* flag if system does not have malloc/free function.

Create a separate dm_mem.h file to define only the function prototypes of needed functions:

dm_init()
dm_alloc()
dm_free()
dm_realloc()

X-Ryl669:

Fix #4 in lv_lib_rlottie.
This adds 2 functions to control rlottie animation:
```cpp
void lv_rlottie_set_play_mode(lv_obj_t * rlottie, const lv_rlottie_play_control_t ctrl);
void lv_rlottie_set_current_frame(lv_obj_t * rlottie, const size_t goto_frame);
```
to be used like this:
```cpp
  lv_obj_t * lottie = lv_rlottie_create_from_file(scr, 128, 128, "test.json");
  lv_obj_center(lottie);
  // Pause to a specific frame
  lv_rlottie_set_current_frame(lottie, 50);
  lv_rlottie_set_play_mode(lottie, lv_rlottie_pause);
  // Play backward and loop
  lv_rlottie_set_play_mode(lottie, lv_rlottie_play | lv_rlottie_backward | lv_rlottie_loop);
  // Play forward once (no looping)
  lv_rlottie_set_play_mode(lottie, lv_rlottie_play | lv_rlottie_forward);
```
Also triggers an EVENT_READY at end of animation (only if not looping, obviously).

### DIFF
--- a/src/extra/libs/rlottie/lv_rlottie.c
+++ b/src/extra/libs/rlottie/lv_rlottie.c
@@ -19,6 +19,7 @@
 /**********************
 *      TYPEDEFS
 **********************/
+#define LV_ARGB32   32
 
 /**********************
  *  STATIC PROTOTYPES
@@ -81,6 +82,13 @@ lv_obj_t * lv_rlottie_create_from_raw(lv_obj_t * parent, lv_coord_t width, lv_co
     return obj;
 }
 
+void lv_rlottie_set_play_mode(lv_obj_t * obj, const lv_rlottie_play_control_t ctrl, size_t param)
+{
+    lv_rlottie_t * rlottie = (lv_rlottie_t *) obj;
+    rlottie->dest_frame = param; /* Must be set before the control */
+    rlottie->play_ctrl = ctrl;
+}
+
 /**********************
  *   STATIC FUNCTIONS
  **********************/
@@ -123,6 +131,9 @@ static void lv_rlottie_constructor(const lv_obj_class_t * class_p, lv_obj_t * ob
 
     lv_img_set_src(obj, &rlottie->imgdsc);
 
+    rlottie->play_ctrl = lv_rlottie_forward;
+    rlottie->dest_frame = 0;
+
     rlottie->task = lv_timer_create(next_frame_task_cb, 1000 / rlottie->framerate, obj);
 
     lv_obj_update_layout(obj);
@@ -146,6 +157,8 @@ static void lv_rlottie_destructor(const lv_obj_class_t * class_p, lv_obj_t * obj
     if(rlottie->task) {
         lv_timer_del(rlottie->task);
         rlottie->task = NULL;
+        rlottie->play_ctrl = lv_rlottie_forward;
+        rlottie->dest_frame = 0;
     }
 
     if(rlottie->allocated_buf) {
@@ -197,10 +210,30 @@ static void next_frame_task_cb(lv_timer_t * t)
 {
     lv_obj_t * obj = t->user_data;
     lv_rlottie_t * rlottie = (lv_rlottie_t *) obj;
-    if(rlottie->current_frame == rlottie->total_frames)
-        rlottie->current_frame = 0;
-    else
-        ++rlottie->current_frame;
+
+    switch(rlottie->play_ctrl)
+    {
+    case lv_rlottie_forward:
+        if(rlottie->current_frame == rlottie->total_frames)
+            rlottie->current_frame = 0;
+        else
+            ++rlottie->current_frame;
+        break;
+    case lv_rlottie_pause:
+        return;
+    case lv_rlottie_goto:
+        rlottie->current_frame = rlottie->dest_frame <= rlottie->total_frames ? rlottie->total_frames - 1 : rlottie->dest_frame;
+        rlottie->play_ctrl = lv_rlottie_pause;
+        break;
+    case lv_rlottie_backward:
+        if(rlottie->current_frame == 0)
+            rlottie->current_frame = rlottie->total_frames - 1;
+        else
+            --rlottie->current_frame;
+        break;
+    default:
+        break;
+    }
 
     lottie_animation_render(
         rlottie->animation,
--- a/src/extra/libs/rlottie/lv_rlottie.c
+++ b/src/extra/libs/rlottie/lv_rlottie.c
@@ -6,7 +6,6 @@
 /*********************
  *      INCLUDES
  *********************/
-
 #include "lv_rlottie.h"
 #if LV_USE_RLOTTIE
 
@@ -82,12 +81,13 @@ lv_obj_t * lv_rlottie_create_from_raw(lv_obj_t * parent, lv_coord_t width, lv_co
     return obj;
 }
 
-void lv_rlottie_set_play_mode(lv_obj_t * obj, const lv_rlottie_play_control_t ctrl)
+void lv_rlottie_set_play_mode(lv_obj_t * obj, const lv_rlottie_ctrl_t ctrl)
 {
     lv_rlottie_t * rlottie = (lv_rlottie_t *) obj;
     rlottie->play_ctrl = ctrl;
 
-    if((rlottie->play_ctrl & lv_rlottie_pause) == lv_rlottie_play && rlottie->task) {
+    if(rlottie->task && (rlottie->dest_frame != rlottie->current_frame ||
+                         (rlottie->play_ctrl & LV_RLOTTIE_CTRL_PAUSE) == LV_RLOTTIE_CTRL_PLAY)) {
         lv_timer_resume(rlottie->task);
     }
 }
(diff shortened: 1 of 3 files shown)


src/extra/libs/rlottie/lv_rlottie.c
```diff
@@ -155,14 +178,84 @@ static void lv_rlottie_destructor(const lv_obj_class_t * class_p, lv_obj_t * obj
...
+            A bit better:
+                        r = ((c & 0xF80000) >> 8)
+                        g = ((c & 0xFC00) >> 5)
+                        b = ((c & 0xFF) >> 3)
+                        rgb565 = r | g | b
+                        a = c >> 24;
+            That's 3 mask, 3 bitshifts and 2 or operations */
+        for(size_t x = 0; x < width; x++) {
+            uint32_t in = src[x];
+            uint16_t r = (uint16_t)(((in & 0xF80000) >> 8) | ((in & 0xFC00) >> 5) | ((in & 0xFF) >> 3));
```
kisvegabor, line 208: You can use `lv_color_make(r, g, b)`. It handles `LV_COLOR_SWAP_16` too.

src/extra/libs/rlottie/lv_rlottie.c
```diff
@@ -155,14 +178,84 @@ static void lv_rlottie_destructor(const lv_obj_class_t * class_p, lv_obj_t * obj
...
+
+            memcpy(dest, &r, sizeof(r));
```
kisvegabor:
```suggestion
            lv_memcpy(dest, &r, sizeof(r));
```

src/extra/libs/rlottie/lv_rlottie.h
```diff
@@ -25,6 +25,14 @@ extern "C" {
 /**********************
  *      TYPEDEFS
  **********************/
+typedef enum {
+    lv_rlottie_forward  = 0,
```
kisvegabor: Please use upper case fields.

I also suggest `LV_RLOTTIE_CTRL_...` and `lv_rlottie_ctrl_t` for consistency.

kisvegabor (commented):
Thank you!

I had a few minor notes.

src/extra/libs/rlottie/lv_rlottie.c
X-Ryl669: Well, it's not the same semantic here. `lv_color_make` creates a triplet of bytes (R, G, B) shifted by the required amount and then it need to be concatenated (even if it's hidden by the bitfield, the compiler will do it). Writing this way is like the naive way commented above, it's not efficient. 
Here, I'm making the mask & shift & concatenation at the same time so it's a lot faster. 

To give an insight, with my code, the cumulative clock count is 6650 clocks, while with this code:
```
lv_color_t c = lv_color_make((in & 0xFF0000) >> 16, (in & 0xFF00) >> 8, (in & 0xFF));
uint16_t r = c.full;
```
it's 13318 clocks, so ~ twice slower.
Since we are working on a large buffer, the impact is far from negligible.

src/extra/libs/rlottie/lv_rlottie.c
X-Ryl669: Right, corrected, thanks!

src/extra/libs/rlottie/lv_rlottie.c
kisvegabor: Ah, I see. We have `lv_color_hex(in)` too which seems to fit better.

src/extra/libs/rlottie/lv_rlottie.h
X-Ryl669: Ok, done.

src/extra/libs/rlottie/lv_rlottie.c
X-Ryl669: `lv_color_hex` is also not good. It expands almost exactly to the code line above and still call `lv_color_make`.
 
While looking at the code, it seems a more efficient `lv_color_hex` could be written (using my trick above to merge shift & mask at the same place and get a x2 performance boost), but I think it'll have to be in its own PR, since it'll imply everything calling it, not only this code.

kisvegabor: For the sake of simplicity let's stick with your version :slightly_smiling_face: 

The was hell until we could come up with color manipulation macros that make all compiler happy. `lv_color_hex` is not really used in performance-critical places so I wouldn't touch it. 

Just please add a different case for `LV_COLOR_SWAP_16 == 1`. In this case, the endianness of RGB565 is changed.

src/extra/libs/rlottie/lv_rlottie.c
```diff
@@ -217,15 +221,18 @@ static void next_frame_task_cb(lv_timer_t * t)
     lv_rlottie_t * rlottie = (lv_rlottie_t *) obj;
 
     if((rlottie->play_ctrl & lv_rlottie_pause) == lv_rlottie_pause) {
-        if(rlottie->current_frame == rlottie->dest_frame)
```
xaowang96: We pause the animation to display the current frame statically, and may not need to be reset to the destination frame.

X-Ryl669: Hi Xao Wang, 

  The logic here is if one is setting a destination frame and then call pause, it'll be displayed **and then** the animation (and timer) will pause. You are reviewing an outdated code (or I changed it while you wrote your message).

It's better this way because else, if you pause the animation while it hasn't even started (like in your constructor), it'll not display anything. 
In my code, it'll display **one** frame exactly and then pause.

X-Ryl669: Ok, I get it. You want to call pause without changing the current frame index in the animation.

X-Ryl669: Should be better now. Let me know what you think.

src/extra/libs/rlottie/lv_rlottie.c
X-Ryl669: Done.

src/extra/libs/rlottie/lv_rlottie.c
xaowang96: It's better now, this is what i want.
Also, i have an idea, i want to refresh a frame after creating lottie, that is to say, call next_frame_task_cb in the constructor.

X-Ryl669: The code will actually do that, see:
1. You create lottie.
2. (task is created in lv_rlottie_constructor)
3. Optionally you call set_current_frame (`current_frame` is updated)
4. Optionally you change the mode to pause
4. (timer is run, since `dest_frame` is not `current_frame`, the frame will be rendered)
5. (timer is run again, but now, `dest_frame` is `current_frame`, so the timer is paused)

You'll get your frame rendered and the timer is no more running.
If you don't change the mode to pause, it'll run from the frame you've set (the timer will run).

src/extra/libs/rlottie/lv_rlottie.c
```diff
@@ -122,6 +140,9 @@ static void lv_rlottie_constructor(const lv_obj_class_t * class_p, lv_obj_t * ob
 
     lv_img_set_src(obj, &rlottie->imgdsc);
 
+    rlottie->play_ctrl = LV_RLOTTIE_CTRL_FORWARD | LV_RLOTTIE_CTRL_PLAY | LV_RLOTTIE_CTRL_LOOP;
+    rlottie->dest_frame = (size_t) -1; /* invalid destination frame so it's possible to pause on frame 0 */
```
kisvegabor: Are you sure it's valid for all compilers? I think an `INT_MAX` constant would be safer.

kisvegabor (commented):
Looks good, only one thing.

src/extra/libs/rlottie/lv_rlottie.c
X-Ryl669: Should be better now.

src/extra/libs/rlottie/lv_rlottie.c
xaowang96: I want execute `next_frame_task_cb ` once synchronously in the constructor, so that it can be updated immediately.

X-Ryl669: Even if you execute `next_frame_task_cb`, it won't be updated immediately. The image is only *invalidated* in the timer callback, it'll only be drawn at the rendering time which will happen after the timer is run anyway, if I understand correctly. So, in reality it won't change anything except for making the code more complex.

xaowang96: If we call `next_frame_task_cb `, rlottie's image source is confirmed, then we can call `lv_refr_now` to refresh the screen.

X-Ryl669: I'm really not fond of object triggering a full screen refresh when constructed. If the later objects in your function takes longer to construct you'll see an incomplete screen. If you want to only display the first frame, use this code, it'll work with almost no overhead:
```
lv_obj_t * rlottie = lv_rlottie_create_from_file(obj, 128, 128, "test.json");
lv_rlottie_set_play_mode(rlottie, LV_RLOTTIE_PAUSE); 
```

kisvegabor:
Looks good, merging. Thank you very much!

PR #2867, opened 2021-12-02
fix(refr): avoid swapping buffers every area in direct mode

gh2o:

When using `direct_mode` and 2 buffers, direct mode is not useful because half of the invalidated areas are drawn to buf1 and half are drawn to buf2. This PR ensures that the buffer swap does not occur until a given buffer is fully redrawn.

### Checkpoints

### DIFF
--- a/src/core/lv_refr.c
+++ b/src/core/lv_refr.c
@@ -979,7 +979,7 @@ static void draw_buf_flush(void)
             call_flush_cb(disp->driver, &draw_buf->area, color_p);
         }
     }
-    /*If there 2 buffer swap them. With direct mode swap only on the last area*/
+    /*If there are 2 buffers swap them. With direct mode swap only on the last area*/
     if(draw_buf->buf1 && draw_buf->buf2 && (!disp->driver->direct_mode || draw_buf->flushing_last)) {
         if(draw_buf->buf_act == draw_buf->buf1)
             draw_buf->buf_act = draw_buf->buf2;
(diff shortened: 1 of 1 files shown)


src/core/lv_refr.c
```diff
@@ -979,7 +979,7 @@ static void draw_buf_flush(void)
             call_flush_cb(disp->driver, &draw_buf->area, color_p);
         }
     }
-    if(draw_buf->buf1 && draw_buf->buf2) {
+    if(draw_buf->buf1 && draw_buf->buf2 && (!disp->driver->direct_mode || draw_buf->flushing_last)) {
```
kisvegabor, line 983:
```suggestion
    /*If there 2 buffer swap them. With direct mode swap only on the last area*/
    if(draw_buf->buf1 && draw_buf->buf2 && (!disp->driver->direct_mode || draw_buf->flushing_last)) {
```

kisvegabor (approved):
Makes sense, thank you! 


I've just added an extra comment.

src/core/lv_refr.c
```diff
@@ -979,7 +979,8 @@ static void draw_buf_flush(void)
             call_flush_cb(disp->driver, &draw_buf->area, color_p);
         }
     }
-    if(draw_buf->buf1 && draw_buf->buf2) {
+    /*If there 2 buffer swap them. With direct mode swap only on the last area*/
```
kisvegabor:
```suggestion
    /*If there are 2 buffers swap them. With direct mode swap only on the last area*/
```

PR #2869, opened 2021-12-02
fix(arc): Free memory when drawing full-circle arc

DCSBL:
### Description of the feature or fix

Please see #2868 

### Checkpoints
- [x] Follow the [styling guide](https://github.com/lvgl/lvgl/blob/master/docs/CODING_STYLE.md)
- [x] Run `code-format.py` from the `scripts` folder. [astyle](http://astyle.sourceforge.net/install.html) needs to be installed.
- [x] Update the documentation

PR #2870, opened 2021-12-02
DO NOT MERGE: Formatting action test

embeddedt:
**Do not merge this.** I am just testing a tweak to the formatting action.

### DIFF
--- a/.github/workflows/check_style.yml
+++ b/.github/workflows/check_style.yml
@@ -28,12 +28,12 @@ jobs:
             echo "Please apply the preceding diff to your code or run scripts/code-format.sh"
             exit 1
           fi
-      - name: Add comment
+      - name: Comment PR
+        uses: thollander/actions-comment-pull-request@v1
         if: github.event_name == 'pull_request' && failure()
-        uses: peter-evans/create-or-update-comment@a35cf36e5301d70b76f316e867e7788a55a31dae
         with:
-          issue-number: ${{ github.event.issue.number }}
-          body: |
+          message: |
             Your PR needs its formatting corrected before it can be merged upstream.
             
             Please comment `@lvgl-bot format` if you would like our bot to commit this change for you.
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/check_style.yml
+++ b/.github/workflows/check_style.yml
@@ -21,4 +21,16 @@ jobs:
         run: python code-format.py
         working-directory: scripts
       - name: Check that repository is clean
-        run: git diff --exit-code --color=always || (echo "Please apply the preceding diff to your code or run scripts/code-format.sh"; false)
+        shell: bash
+        run: |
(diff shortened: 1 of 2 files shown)


embeddedt:
@kisvegabor (and anyone else interested), this stems from an issue I noted with #2869. The [diff my action provided there](https://github.com/lvgl/lvgl/runs/4394955928?check_suite_focus=true) was not very helpful. Anyone without access to `code-format.py` would have a very hard time figuring out what the issue was unless they already knew that there shouldn't be whitespace on empty lines.

I've tried uploading a patch as an artifact, but it's too difficult to find to be useful in most cases. 

I also tried to see if there was a way of creating gists using @lvgl-bot, but there doesn't seem to be a clean way of doing this, at least not from my quick research.

Any ideas on what could be done to improve the developer experience here?

kisvegabor:
> I've tried uploading a patch as an artifact, but it's too difficult to find to be useful in most cases.


What was the problem with a simple `git apply artifact.patch`?

embeddedt:
Finding the patch file itself in GitHub's artifact UI is difficult, and they don't provide a way of getting the URL at upload time and printing it, which would be an acceptable solution otherwise.

kisvegabor:
I see. I've checked some threads and SO posts but seemingly there is a consensus about it's not supported. I really wonder why such simple things (along with downloading repos with submodules) are not supported.

Is it possible to make `lvgl-bot` comment on failure? Although it's really difficult a little bit to find the artifact, but I think we can write a useable guide with some screenshots to help users find it. It's still more than nothing.

embeddedt:
@lvgl-bot format

embeddedt:
@lvgl-bot format

embeddedt:
@lvgl-bot format

embeddedt:
@lvgl-bot format

embeddedt:
Ok, it works on our own branches, now to see if it works coming from a fork.

embeddedt:
@kisvegabor I'm back to the drawing board on this. I realized that it's dangerous to have @lvgl-bot pull a contributor's branch and then operate on it for security reasons (not going into details on GitHub, however I can elaborate in a PM if you want).

However, I remembered that GitHub has a feature that allows you to trigger an appropriately-configured workflow on a desired branch from the web. If we put a workflow like this that would format the branch it runs on, we'd just need to provide instructions for a user to navigate to that page. What do you think?

image

kisvegabor:
> (not going into details on GitHub, however I can elaborate in a PM if you want)

I'd be curious what problems it can cause. So please elaborate in a PM. :slightly_smiling_face: 

> However, I remembered that GitHub has a feature that allows you to trigger an appropriately-configured workflow on a desired branch from the web. If we put a workflow like this that would format the branch it runs on, we'd just need to provide instructions for a user to navigate to that page. What do you think?

Sounds good. I believe in most of the cases we (you, me, or other maintainer) can start formatting too.

Issue #2862, opened 2021-12-01
Add dithering support in the code


X-Ryl669:
### Perform all steps below and tick them with [x]
- [X] Check the related part of the [Documentation](https://docs.lvgl.io/)
- [X] Update lvgl to the latest version
- [X] Reproduce the issue in a [Simulator](https://docs.lvgl.io/latest/en/html/get-started/pc-simulator.html)

### Describe the bug
Currently, there's no dithering algorithm implemented or used in LVGL for small color depth resulting in unpleasant color banding when rendering gradients.
@kisvegabor proposed to use images [here](https://github.com/lvgl/lvgl/issues/611#issuecomment-445491855) and @beibean  proposed a generic solution that's based on style [here](https://github.com/lvgl/lvgl/issues/258#issuecomment-390573927).

The former solution implies a trick (you have to compute the dithering out of LVGL and use additional images, that consume resources). Also, this does not work for round corner gradient (like a shadow or glow effect) since the 1px wide dithering pattern stored in the image can only be stretched in one axis.

The latter solution is better, IMHO, since at least, it can be used everywhere a gradient is used, and does not require an additional object to embed and deal with in the scene.
However, the fixed dither pattern in that solution is ok for small display, but will not work very well on large display.

Ideally a true dithering algorithm like [Floyd Steinberg](https://en.wikipedia.org/wiki/Floyd%E2%80%93Steinberg_dithering) should be used instead, since it's producing the best result for not much additional work. 
[Here's a comparison of different dithering algorithms](https://surma.dev/things/ditherpunk/).

### To Reproduce
See any of the issue reported above for examples of color banding.

### Expected behavior
An *average* correct gradient on low color depth display.

### Screenshots or video
ditto

kisvegabor:
Let's continue in #2872.

PR #2872, opened 2021-12-02
Gradient dithering

X-Ryl669:

This is for low color depth display (those with less than 24bits per pixel) like RGB565.
In those cases, when a gradient is selected for rendering (either from background or shadow), it's ugly:

Vertical gradient

Horizontal gradient

There is an issue #2862 resuming the different workarounds for this ugliness. Those are only workaround and requires compile-time building and embedding a pseudo dithered picture to approximate the gradient itself. It's not perfect and quite painful step to add in your build process
 
This PR adds dithering in the draw engine so we get this instead:

Horizontal dithering
 and this:

Vertical ordered dithering with 8x8 matrix

This is still a WIP but at least it's functional.

The implementation is doing this in pseudo code:
1. Compute the gradient map in full color instead of current `lv_color_t` format.
2. For horizontal gradient, a error map is also added to compute error diffusion when mapping the current theorical color to the displayed color. 
3. Error is diffused using Floyd Steinberg algorithm, with few optimizations to remove the need for multiplies and divide.
4. For vertical gradient, an error diffusion scheme can not work since the draw algorithm is a scanline based renderer and error would diffuse from column to column (meaning that a pixel at column N is impacted by a pixel at row N due to diagonal diffusion)
5. So instead, I've implemented an ordered dithering instead. In that case, a complete *dithered* line is computed and blended. It's a bit less nice than FS, but I don't think it's possible to implement FS without breaking the renderer and a huge performance penalty.

Few optimizations I had to made:
In the current code, the rendering loop was doing this:
```
for (y in blend area)  {
   if (grad = none) do_something
  else if (grad = horz) do_some_other_thing
  else if (grad = vert) do_another_thing
}
/\ This code is repeated 4 times
```
I've changed it to use pointers to blend functions instead that are set once for all. This reduces the code complexity and allows to update the behavior of the rendering depending on the conditions, and also removes 3 tests per iteration in the multiple for.

## Remarks
* I've encapsulated my code in `LV_DITHER_GRADIENT` macro that should be moved to `lv_conf.h`, since not every one will want a dither algorithm.
* The CPU impact currently is not negligible. Maybe it'd be worth to save the computed gradient map in the `lv_obj_t` to avoid recomputing it every draw. 
* The FS algorithm is slower than the ordered dithering algorithm, but produces much better looking results (see above). It's only available for horizontal gradient through. If the speed is absolutely required, maybe an ordered version for the horizontal gradient could be used too.


X-Ryl669:
I've found a way to compute an error diffusion dithering algorithm in vertical. It gives this kind of results:
VertGradFSSecondTry

Not as good as horizontal FS, but it's a lot better visually than the ordered dithering, IMHO.

For those curious, the implementation is based on a modified Floyd Steinberg algorithm removing the diagonal error coefficient and spreading the missing energy on the other terms. Like the horizontal algorithm, it uses an error buffer of exactly one line, so there's no performance penalty compared to horizontal version.

kisvegabor:
Wow, it looks promising! Thank you!
I can review it in detail next week (hopefully Monday).

X-Ryl669:
Ok, no problem. 
Do you have an idea about the right way to store a gradient map inside the obj using it ? (To avoid recomputing it each draw)
Also, right now, the dither algorithm depends on `LV_DITHER_GRADIENT` which is a compile time configuration option. Maybe it's possible to set up a style instead like `style->antialias_gradient = 1`. But then, how to pass it down to the draw operation ?

src/draw/sw/lv_draw_sw_rect.c
```diff
@@ -96,6 +117,159 @@ void lv_draw_sw_rect(const lv_area_t * coords, const lv_area_t * clip, const lv_
 /**********************
  *   STATIC FUNCTIONS
  **********************/
+#if DITHER_GRADIENT
+LV_ATTRIBUTE_FAST_MEM static inline uint8_t clamp(int a)
+{
+    return (uint8_t)(a < 0 ? 0 : (a > 255 ? 255 : a));
```
kisvegabor: We already have `LV_CLAMP(min, value, max)`.

src/draw/sw/lv_draw_sw_rect.c
```diff
@@ -19,9 +19,24 @@
 #define SHADOW_ENHANCE          1
 #define SPLIT_LIMIT             50
 
+#define LV_DITHER_GRADIENT 1
```
kisvegabor: It should be in `lv_conf.h`.

src/draw/sw/lv_draw_sw_rect.c
```diff
@@ -96,6 +117,159 @@ void lv_draw_sw_rect(const lv_area_t * coords, const lv_area_t * clip, const lv_
...
  **********************/
+#if DITHER_GRADIENT
+LV_ATTRIBUTE_FAST_MEM static inline uint8_t clamp(int a)
+{
+    return (uint8_t)(a < 0 ? 0 : (a > 255 ? 255 : a));
+}
+
+
+
+LV_ATTRIBUTE_FAST_MEM static inline void dither_fs(const lv_color32_t * src, lv_color_t * out, scolor24_t * next_line,
```
kisvegabor: If it's for horizontal what about `dither_fs_hor`?

src/draw/sw/lv_draw_sw_rect.c
```diff
@@ -154,23 +328,54 @@ LV_ATTRIBUTE_FAST_MEM static void draw_bg(const lv_area_t * coords, const lv_are
...
+    if(grad_dir == LV_GRAD_DIR_NONE)
+        blend_func = &blend_fill_line;
+    else if(grad_dir == LV_GRAD_DIR_HOR)
+        blend_func = &blend_map_line;
+    else if(grad_dir == LV_GRAD_DIR_VER) {
+#if DITHER_GRADIENT
+        blend_func = &blend_map_line;
+#else
+        blend_func = &blend_fill_vert_line;
+#endif
```
kisvegabor: I'm working on improving the architecture of the drawing part and made a similar simplification. https://github.com/lvgl/lvgl/blob/699cf2d63cb477f9ec76a788a4be26b9e60c4577/src/draw/sw/lv_draw_sw_rect.c#L184-L190

src/draw/sw/lv_draw_sw_rect.c
```diff
@@ -402,6 +600,43 @@ LV_ATTRIBUTE_FAST_MEM static inline lv_color_t grad_get(const lv_draw_rect_dsc_t
     return lv_color_mix(dsc->bg_grad_color, dsc->bg_color, mix);
 }
 
+/*Simple mathematic trick to avoid using a divide or multiplier */
+LV_ATTRIBUTE_FAST_MEM static inline uint8_t divBy255(uint16_t a)
```
kisvegabor: We have `LV_UDIV255`.

kisvegabor (commented):
Thanks a lot! I added some minor comments.

A potential issue came to my mind. If there is a 100x100 rectangle with a gradient and only a 10x10 area on its center needs to be redrawn won't the gradient calculation give a different result because the errors are not calculated from the beginning of the 100x100 rectangle?

For storing the gradient: for v8.2 I'm thinking about something more general. It'd be great if the user could make any objects "buffered" which means LVGL crates a  snapshot about them and just copies this snapshot image during rendering. What do you think about it?

src/draw/sw/lv_draw_sw_rect.c
X-Ryl669: Right, will use that

src/draw/sw/lv_draw_sw_rect.c
X-Ryl669: Yes. On the last commit only, since it'll break CI if I change it, right?

src/draw/sw/lv_draw_sw_rect.c
X-Ryl669: Not the same. UDIV255 multiplies and shift (but works for any number). Here, I know that the product will always be < 65535, so I can use a shift + add solution only (see comment). Not sure it'll make a significant difference although, if the gradient map is cached.

src/draw/sw/lv_draw_sw_rect.c
X-Ryl669: Not sure I'm following you here on your refactor. `lv_draw_sw_blend` as I'm reading it is not a pointer to the right function. It's a pseudo generic function that'll contain something like:
```
if (some mode) do something
else if (some other mode) do some other thing 
else ...
```
Since it's inside a for loop, and it's a function that's not visible to the compiler (different compile unit), it can't be unrolled, so it'll be very heavy (for each iteration, you'll have to pay for the same tests again and again). IIUC, you're hiding the mess of the *current* code under the rug (in its own file). 

It's good for the readability of the code through.

In my solution above, I'm setting the "function" to use depending on some flags that are uniform/constant for the whole loop. It allows having a different line drawing function based on if the gradient is horizontal, vertical, with dithering, etc...

It's a shame that it's C code. In C++, you'd write a lambda and have the compiler write the optimal code for you here without the cost of calling a function. Here, in both your and my solution, we have to pay for calling a function per line.
In your solution, you also pay for finding out what to do in that function (with tests) while in mine, I'm not paying this cost.

I think a better way to follow on your refactoring idea would be to mode the loop inside the `lv_draw_sw_blend` function, so you can have a code like this instead:
```
   if (some mode) for(y loop) blend func()
  else if (some other mode) for (y loop) another blend func()
```

X-Ryl669:
> A potential issue came to my mind. If there is a 100x100 rectangle with a gradient and only a 10x10 area on its center needs to be redrawn won't the gradient calculation give a different result because the errors are not calculated from the beginning of the 100x100 rectangle?

Very good remark, indeed.
In the screenshot above, the icon is a rlottie file. It's transparent and drawn over the background object. Only the area of the icon is drawn each frame and it fits perfectly (but it was a pain to get right). So TL/DR: it works.

So to answer your remark, I'll try to explain why it works with partial area:
1. For horizontal gradient, the whole gradient line is computed/dithered based on the gradient size. Only the partial area is blended. The next error line is also computed completely, so the error is constant, and independent of the area to draw.
2. There is no guarantee that the error will be exactly the same from rows to rows (if starting at a different row), but it'll be exactly the same for columns to columns. Since it's a dithering of an horizontal gradient, a different *error* row actually improve the dithering.
2. For vertical gradient, I'm using the fact that the rendering loop is processing line by line. So the first line to be blended, the error diffusion is computed from a perfect previous line based on the 24 bit perfect vertical gradient map. Then the error diffusion is computed for a complete line and only the required partial area is blended. So again, a different error row for the first line would improve the dithering.
3. The ordered dithering (first commit) does not need the error line and only depends on the X/Y position to know what to draw. So in that case, it's perfectly deterministic and fast since you only need to pass the top left X/Y position to figure the dithering pattern. But it's a bit more grainy, less nice to the eye. Since it does not require an error line it uses less memory. 

Typically, in error diffusion dithering, I'm computing to `grad_map` (which is gradient's width size) and only blending from the partial area's x offset into the map.

As you see, my algorithm takes a lot of time recomputing a complete gradient lines (and not the partial area) so caching would really be required for high performance here. 

There are multiple way to improve the runtime's performance, like deciding if creating a limited sized version of the dithering to an (cached) image and tiling the image would give better result at the cost of higher memory consumption for faster cpu/DMA blitting, or caching the first error line once the objects on top of the gradient are positioned and drawn (so instead of storing a 100 signed pixels error line you'd only store a 10 pixel error line for the icon).
Depending on the situation each method has its pros and cons. 

Right now, the PR is only implementing the feature, optimization would come later, IMHO. 

Also, I'm wondering if the dithering shouldn't be part of the style and let the use select the dither algorithm to use at runtime and per object (so the tradeoff is user selectable between none/ordered/error diffusion).

X-Ryl669:
> For storing the gradient: for v8.2 I'm thinking about something more general. It'd be great if the user could make any objects "buffered" which means LVGL crates a snapshot about them and just copies this snapshot image during rendering. What do you think about it?

I don't know. In my situation, I'm using a ESP32 and I don't have that much RAM to spend. *Maybe* there is a way to compute what would be the minimum size to snapshot to ensure a correct rendering of the dither pattern (like I did for vertical dithering, I can do the same for horizontal dithering and limit the length of the diffusion algorithm to only few rows & cols). In that case, only that part can be snapshot.

But for a general snapshotting algorithm (that would capture the complete object's rendering buffer), then you'll be very close to what X11 is doing and I think it'll be complex to get right (think damage and so on) and will not fit on embedded system well unless it's big and in that case, using X11 will be easier ???

X-Ryl669:
Here it is (many changes):
1. I have made the LV_DITHER_GRADIENT a lv_conf macro. 
2. Added a LV_DITHER_ERROR_DIFFUSION macro to enable the building of the error diffusion code (by default, the ordered diffusion is used, since it's faster and use less memory)
3. Added a style "dither_mode" to allow runtime selection of the dithering for an object's background gradient.
4. Reworked the ordered dithering algorithm to perform better (now it's visually acceptable, almost as good as the error diffusion algorithm). Sorry no screenshot, but a test code so you can run it by yourself.
5. Applied the requested changes to the code. I haven't integrated in the new refactoring since it's not public yet.


Test code:
```cpp
void lv_demo_rlottie() 
{
  lv_disp_t * disp = lv_disp_get_next(NULL);
  disp->driver->monitor_cb = monitor_cb;
  lv_disp_set_bg_color(disp, lv_color_hex(0x000000));

  lv_obj_t * scr = lv_scr_act();

  static lv_style_t style;
  lv_style_init(&style);
  lv_style_set_radius(&style, 5);


  /*Make a gradient*/
  lv_style_set_bg_opa(&style, LV_OPA_COVER);
  lv_style_set_bg_color(&style, lv_color_hex(0x000000));
  lv_style_set_bg_grad_color(&style, lv_color_hex(0x542293));
  lv_style_set_bg_grad_dir(&style, LV_GRAD_DIR_VER);
  lv_style_set_bg_dither_mode(&style, LV_DITHER_ORDERED);


  /*Shift the gradient to the bottom*/
  lv_style_set_bg_main_stop(&style, 0);
  lv_style_set_bg_grad_stop(&style, 480);

  /*Create an object with the new style*/
  lv_obj_t * obj = lv_obj_create(scr);
  lv_obj_add_style(obj, &style, 0);
  lv_obj_set_size(obj, 480, 320);
  lv_obj_center(obj);

  lv_obj_t * lottie = lv_rlottie_create_from_file(obj, 128, 128, "test.json");
  lv_obj_center(lottie);
}
```

src/draw/sw/lv_draw_sw_rect.c
kisvegabor: Yes, it's only a wrapper to use a unified function. Basically, we moved the `if` from `lv_draw_rect` to `lv_draw_sw_blend`.

I think this extra if doesn't matter much compared to the overall operation to render the UI.

kisvegabor:
> Very good remark, indeed.
In the screenshot above, the icon is a rlottie file. It's transparent and drawn over the background object. Only the area of the icon is drawn each frame and it fits perfectly (but it was a pain to get right). So TL/DR: it works.

Thanks for the explanation.

> I don't know. In my situation, I'm using a ESP32 and I don't have that much RAM to spend. Maybe there is a way to compute what would be the minimum size to snapshot to ensure a correct rendering of the dither pattern (like I did for vertical dithering, I can do the same for horizontal dithering and limit the length of the diffusion algorithm to only few rows & cols). In that case, only that part can be snapshot.

It's also ok to build a "local cache" for gradients in a custom way. 
As I see this X11-like rendering becomes more important in the future as MCU gets more complex, faster and feature-rich. Some people also use LVGL on desktop as a QT replacement. It'd great to target this market too. It also allows features like rotating/zooming/blurring an SW rendered button.

> Added a LV_DITHER_ERROR_DIFFUSION macro to enable the building of the error diffusion code (by default, the ordered diffusion is used, since it's faster and use less memory)

By default `LV_DITHER_NONE` should be used, isn't it?

X-Ryl669:
> By default `LV_DITHER_NONE` should be used, isn't it?

It's the default style yes. If you need dithering, you'll need to change this via `set_style` like function.
I meant that if you don't define `LV_DITHER_ERROR_DIFFUSION` (only `LV_DITHER_GRADIENT`) in your `lv_conf` then the error diffusion algorithm is not built, the error cache line is not allocated and the ordered dithering is used instead. If you define it, then you can select the dithering algorithm via `set_style(LV_DITHER_ORDERED)`  or `set_style(LV_DITHER_ERR_DIFF)` at runtime.

kisvegabor:
Got it, sounds good. From my side it's ready to merge but it will have huge conflicts with the new draw architecture in https://github.com/lvgl/lvgl/tree/arch/draw-restructure

So either you or I need to refactor the dithering code for the changes there. 
I don't want to push it to you but as functionally the draw_rect code is the same, only a few API were changed, I think it'd be safer and faster if you could update the dithering code as you oversee it much better. 

So would you have time/interest to rebase it for the `arch/draw-restructure` branch?

On a side note, meanwhile, I got questions about supporting radial, skew, and conic gradients too. In the future maybe we should have a more generic gradient manager with drawing caching, etc

X-Ryl669:
Can you explain your restructure idea please ? 
The branch diverged significantly and I'm not sure to understand how it works in the branch ?

kisvegabor:
I summarized the changes here: https://github.com/lvgl/lvgl/issues/2894 Please comment in the mentioned issue if you have any question or remarks.

X-Ryl669:
I'll wait until it's stabilized (merged ?) to port my changes on this branch.
Technically, it's completely possible to reuse the code for radial and conic gradient too (don't know what is a skew gradient). It would even be possible to use this on image too, since the dithering code does not care about what it's dithering.
:thinking:

kisvegabor:
> I'll wait until it's stabilized (merged ?) to port my changes on this branch.

I can merge it to master, but I consider new architecture stable. No big refactoring is expected.

> Technically, it's completely possible to reuse the code for radial and conic gradient too (don't know what is a skew gradient).

It's great! By skew gradient I mean linear gradient with an angle. E.g. 
image

>  It would even be possible to use this on image too, since the dithering code does not care about what it's dithering.

On images dithering can be applied during converting them C array or binary format. If e.g.  Although if PNG image is opened by LVGL and converted from ARGB8888 format  to ARGB8565 there won't be dithering now.

X-Ryl669:
I meant, I'll port it to the new architecture when you've merged the new architecture in master.

kisvegabor:
Ah, got it. Sounds great!

Miguel0101:
In the future, will LVGL have a diagonal gradient?

kisvegabor:
Yes, it'd be great!

X-Ryl669:
Recent changes:
 1. Split the code in multiple files to make it more maintainable
 2. Merged on new draw backend

## Multiple files
Now, there're `lv_draw_gradient.h/c` files to describe a gradient. Right now, I've limited to 2 stops per gradient, but having more stops is a define away since the interpolation function is not limited to any number of stops.
Gradient are now objects too, and are cached. It means that the gradient map will not be computed each time an object is drawn, but only when there's no more space in the cache.
If we need to support new gradient type, it'll be possible (although the drawing implementation is not trivial), like diagonal gradient and so on, since once the map is computed, only the drawing code will have to handle where to put what color. 

There're also a `lv_draw_dither.h/c` files to describe a dithering algorithm. Nothing special here, that's the minimum to support the features (see below). Adding dithering to picture would be very easy too since it's using a map to work on. 

## Merged on new draw backend
I've merged the code to the new draw backend. It was not easy :sob: .
Here are the results with the different options:

**Horizontal gradient with ordered dithering**
HorzOrd

**Horizontal gradient with error diffusion**
HorzFS

**Vertical gradient with ordered dithering**
VertOrd

**Vertical gradient with error diffusion**
VertFS

X-Ryl669:
~~I've submitted a PR for the failing lv_binding code [here](https://github.com/lvgl/lv_binding_micropython/pull/201)
Basically changing the `dsc.bg_color` to `dsc.bg_grad.stops[0].color` instead.~~

EDIT: Not required anymore.

lv_conf_template.h
```diff
@@ -119,6 +119,16 @@
     * 0: to disable caching */
     #define LV_CIRCLE_CACHE_SIZE 4
 
+    /*Allow dithering gradient (to achieve visual smooth color gradients on limited color depth display)
+    *LV_DITHER_GRADIENT implies consuming one or two lines of the object's rendering surface
```
kisvegabor: How do you mean `consuming`?

src/draw/lv_draw_dither.c
```diff
@@ -0,0 +1,213 @@
...
+    }
+    /* Prevent overwriting */
+    for(; j < w; j++) {
+        grad->map[j] = grad->map[j & 7];
+    }
+}
+
+
+#if LV_DITHER_ERROR_DIFFUSION == 1
+LV_ATTRIBUTE_FAST_MEM void lv_dither_fs_hor(lv_gradient_cache_t * grad, lv_coord_t xs, lv_coord_t y, lv_coord_t w)
```
kisvegabor: What does `fs` mean in the function's name?

src/draw/lv_draw_dither.h
```diff
@@ -0,0 +1,71 @@
...
+ *      INCLUDES
+ *********************/
+#include "../core/lv_obj_pos.h"
+
+
+/*********************
+ *      DEFINES
+ *********************/
+#if LV_COLOR_DEPTH < 32 && LV_DRAW_COMPLEX == 1 && LV_DITHER_GRADIENT == 1
+#define DITHER_GRADIENT 1
```
kisvegabor: It should be prefixed with an underscore to avoid possible conflicts with user code.

src/draw/lv_draw_gradient.c
```diff
@@ -0,0 +1,316 @@
...
+
+/**********************
+ *   STATIC FUNCTIONS
+ **********************/
+union void_cast {
+    const void * ptr;
+    const uint32_t value;
+};
+
+uint32_t compute_key(const lv_gradient_t * g, lv_coord_t size, lv_coord_t w)
```
kisvegabor: These functions are in the `STATIC FUNCTIONS` section. So isn't there missing `static` keywords here?

src/draw/lv_draw_gradient.h
```diff
@@ -0,0 +1,113 @@
...
+/** Set the gradient cache size */
+void lv_grad_set_cache_size(size_t max_bytes);
+
+/** Get a gradient cache from the given parameters */
+lv_gradient_cache_t * lv_grad_get_from_cache(const lv_gradient_t * gradient, lv_coord_t w, lv_coord_t h);
+
+/** Evict item from the gradient cache (not used anymore).
+ *  This bypass the life counter on the item to remove this item.
+ */
+void lv_grad_evict_from_cache(const lv_gradient_t * gradient, lv_coord_t w, lv_coord_t h);
```
kisvegabor: `evict`->`pop`?

src/draw/lv_draw_gradient.c
```diff
@@ -0,0 +1,316 @@
...
+{
+    lv_mem_free(grad_cache_mem);
+    grad_cache_mem = grad_cache_end = NULL;
+    grad_cache_size = 0;
+}
+
+void lv_grad_set_cache_size(size_t max_bytes)
+{
+    lv_mem_free(grad_cache_mem);
+    grad_cache_end = grad_cache_mem = lv_mem_alloc(max_bytes);
```
kisvegabor: Please add `LV_ASSERT_MALLOC(grad_cache_mem)` here.

src/misc/lv_math.h
```diff
@@ -51,6 +51,23 @@ static inline LV_ATTRIBUTE_FAST_MEM int16_t lv_trigo_cos(int16_t angle)
...
+static inline LV_ATTRIBUTE_FAST_MEM uint8_t lv_udiv_255(uint16_t x)
+{
+    /* Mathematical trick here:
+        Since (2^n - 1) = (a - b), and since (a-b)(a+b) = a^2 - b^2, then
+             x / (2^n - 1) = x * (2^n + 1) / [(2^n + 1)*(2^n - 1)] = (x*2^n + x) / (2^(2n) - 1) ~= (x<<n + x) >> (2*n)
+
+        Since 2^(2n) - 1 ~= 2^2n, we commit a smaller error than if we divided by shifting right initially by n.
+        So, for example, for x in [0-65536] range, and dividing by 255, this approximation gives no error */
+    uint32_t num = ((uint32_t)x << 8) + x;
+    return (uint8_t)(num >> 16);
```
kisvegabor: Is it really faster than `LV_UDIV255`? Here there are 2 shifts and 1 addition.

`LV_UDIV255` is 1 multiplication (1 instruction) and 1 shift.

src/draw/lv_draw_dither.h
```diff
@@ -0,0 +1,71 @@
+/**
```
kisvegabor: I think the dithering related files should go into the `src/draw/sw` folder.

kisvegabor (commented):
Thank you for your high quality work! It looks really great!

I didn't try to understand how dithering itself works but rather focused on the whole picture and API level.

> I've merged the code to the new draw backend. It was not easy

What was the main problem? The many refactoring or there were issues with the new backend API?

---

It'd be great to keep the simplicity of just setting a bg. color in `draw_dsc.bg_color` and keep backward compatibility too.

I have an idea! What about keeping `bg_color` and adding `lv_gradient_t * grad` to `lv_draw_rect_dsc_t`. If `grad` is `NULL`, `bg_color` is used, else `bg_color` is ignored and `grad` is used. Later we can move this gradient handling mechanism to styles, i.e. the user can set a pointer to a gradient descriptor in styles. What do you think?

lv_conf_template.h
X-Ryl669: sorry, french word. I meant "using as much memory as"

src/draw/lv_draw_dither.c
X-Ryl669: fs is for the name of the inventors "Floyd" and "Steinberg". We can change to "ed" instead for error diffusion.

src/draw/lv_draw_dither.h
X-Ryl669: Ok

src/draw/lv_draw_gradient.c
X-Ryl669: Well, don't want to start a debate here, but you don't need to repeat `static` if the prototype contains the `static` intenal linkage. 

## C11: 6.2.2 Linkages of identifiers (p4):

For an identifier declared with the storage-class specifier extern in a scope in which a prior declaration of that identifier is visible, **if the prior declaration specifies internal or external linkage, the linkage of the identifier at the later declaration is the same as the linkage specified at the prior declaration**. [...]

But to be consistent with the rest of the code, I'll add static here too.

src/draw/lv_draw_gradient.h
X-Ryl669: Ok

src/misc/lv_math.h
X-Ryl669: On ESP32, mul is 2 cycles, while add is one cycle, shift is one cycle, so it ends up the same. Div is 4 cycles BTW, so the gain is low anyway.

On ARM cortex M0, multiply are 32 cycles while add & shift are only one cycle, so it's definitively faster here.
On ARM cortex M3, multiply takes from 1 to 3 cycles (add & shift = 1 cycle). So whether it's faster or slower depends on the chip implementation.
On ARM cortex M7, same as M3/M4 here.

So in the end, if it's faster, it's by a very short margin. I wonder what the compiler is doing for a plain /255 in the code...

src/draw/lv_draw_dither.h
X-Ryl669: It's not software related, is it ?
The dithering algorithm builds a pixel map that's blitted by the backend. So a GPU backend will just optimize this, IIUC.
The map is constructed by a software algorithm, but I had the impression that the sw folder was for sw rendering algorithms, not for higher level sw drawing functions. Please advise.

X-Ryl669:
> I have an idea! What about keeping bg_color and adding lv_gradient_t * grad to lv_draw_rect_dsc_t. If grad is NULL, bg_color is used, else bg_color is ignored and grad is used. Later we can move this gradient handling mechanism to styles, i.e. the user can set a pointer to a gradient descriptor in styles. What do you think?

The only issue I can think of is that it requires managing the memory (who will allocate, who will free it?). And since the pointer is 4 to 8 bytes anyway, we can almost store a grad structure directly here instead (since it's 8 bytes also).
In the gradient obj, we have a stop count that can be 0 if unused and we'd fallback to bg_color in that case. Let's see what I come to.

src/draw/lv_draw_dither.c
kisvegabor: `ed` seems more intuitive, but `err_diff` or so sounds even better to me.

src/draw/lv_draw_gradient.c
kisvegabor: I didn't know that, thanks. 

> But to be consistent with the rest of the code, I'll add static here to

Thanks.

src/draw/lv_draw_dither.h
kisvegabor: IMO it's related to SW renderer. E.g. ThorVG would have its own gradient functions.

src/misc/lv_math.h
kisvegabor: Oh, I thought mul is 1 i.c. on every modern MCU. Thanks.

> I wonder what the compiler is doing for a plain /255 in the code...

Probably compiler/optimization dependent. 

Anyway, let's leave your option too!

kisvegabor:
> In the gradient obj, we have a stop count that can be 0 if unused and we'd fallback to bg_color in that case. Let's see what I come to.

Sounds good!

X-Ryl669:
Found a way to keep `bg_color` and `bg_grad.stops[0].color` on the same memory place. So best of both world.

src/draw/lv_draw_dither.h
X-Ryl669: Ok, changed.

src/misc/lv_math.h
X-Ryl669: The gain does not justify the maintenance burden, so using LV_UDIV255 now.

X-Ryl669:
The CI failure doesn't seem related to my code. I haven't touched a `draw_polygon` function

src/draw/lv_draw_rect.h
```diff
@@ -33,12 +34,11 @@ typedef struct {
     lv_blend_mode_t blend_mode;
 
     /*Background*/
-    lv_color_t bg_color;
-    lv_color_t bg_grad_color;
-    uint8_t bg_main_color_stop;
-    uint8_t bg_grad_color_stop;
     lv_opa_t bg_opa;
-    lv_grad_dir_t bg_grad_dir : 3;
+    union {
```
kisvegabor, line 39: We needed to get rid of nameless unions due to some compiler issues. I don't remember exactly what was the problem. Maybe @embeddedt remembers.

src/core/lv_refr.c
```diff
@@ -610,7 +610,7 @@ static void lv_refr_area_part(lv_draw_ctx_t * draw_ctx)
             lv_draw_rect_dsc_init(&dsc);
             dsc.bg_img_src = disp_refr->bg_img;
             dsc.bg_img_opa = disp_refr->bg_opa;
-            dsc.bg_color = disp_refr->bg_color;
+            dsc.bg_grad.stops[0].color = disp_refr->bg_color;
```
kisvegabor: Could you revert the occurrences of `.bg_grad.stops[0].color` to `.bg_color` where applicable?

src/draw/lv_draw_rect.h
embeddedt: Anonymous unions are supported in C11, but not C99. In C99 they only work because of GNU extensions.

X-Ryl669: So? What is the conclusion? Can I use it? 
I don't see a way to have both gradient and bg_color working at the same time without anonymous union. 
Are we still restricted to 23 years old standard ?
BTW, I don't think bool was in C99 either, but we use them in the code too.

X-Ryl669: To be honest the only other solution is to use something a lot more ugly:
```cpp
struct {
[...]
    lv_color_t bg_color;
    uint8_t     _reserved[sizeof(lv_gradient_t) - sizeof(lv_color_t)];
    lv_gradient_t * grad; /**< Alias on bg_color above */
[...]
 
In constructor:
  grad = (lv_gradient_t*)&bg_color;
```

But, it's **SO** ugly compared to the union code.

kisvegabor: > Are we still restricted to 23 years old standard ?

Even with C99 it's challenging to support the wide variety of embedded compilers that are out there. It'd be great to move to C11, but I'm afraid we can't. :slightly_frowning_face: 

> What is the conclusion? Can I use it?

No. :slightly_frowning_face: 

> I don't see a way to have both gradient and bg_color working at the same time without anonymous union.

You suggestion seemed good to me:
```c
//In lv_draw_rect_dsc_t
lv_color_t bg_color;
lv_gradient_t grad; //stop_count is initialized to 0, and if it's 0 bg_color is used else grad
```

X-Ryl669: This won't work well. Because if one decide to set the second stop later on, it's very likely he will forget to change stop_count and it'll fail. Also, it means he has to copy the color from bg_color to grad's stop. That's one more step to do so it's adding friction to the code, I don't like it.

*If only the field could be private so it would only be set via a method, but, eh, it's C, not C++.*
I also don't like the pointer in the structure like in the example above since all other field are not pointer, it's distracting.

And if I change the union not to be anonymous, it'll break API anyway:
```cpp
struct {
[...]
  union {
    lv_color_t color;
    lv_gradient_t grad;
  } bg;
[...]

// Ends up used like:
dsc.bg.color = ...
// or
dsc.bg.grad.stops[0].color = ...
```

kisvegabor: > This won't work well. Because if one decide to set the second stop later on, it's very likely he will forget to change stop_count and it'll fail. Also, it means he has to copy the color from bg_color to grad's stop. That's one more step to do so it's adding friction to the code, I don't like it.

Really not the best but I'd be ok with that as a compromise. In 99% of the cases people set only the `bg_color`. If rarely someone touches the gradients, he can take care a little.  If there were more stops the user needs to set the stop_count anyway.

X-Ryl669: Ok, just commited a change. I'm using the `grad.dir` member as the test for the gradient presence. So setting gradients only require setting the 2 stops now and the direction (unlike previous code that required only setting the 2nd stop), not the stop count. It'll work for 99% of the cases, let's say it's ok.

src/draw/lv_draw_gradient.c
```diff
@@ -0,0 +1,316 @@
+/**
+ * @file lv_draw_gradient.c
+ *
+ */
+
+/*********************
```
kisvegabor: I've noticed that this file should be also in the `draw/sw` folder

src/extra/widgets/chart/lv_chart.c
```diff
@@ -1202,7 +1202,8 @@ static void draw_series_bar(lv_obj_t * obj, lv_draw_ctx_t * draw_ctx)
     lv_draw_rect_dsc_t col_dsc;
     lv_draw_rect_dsc_init(&col_dsc);
     lv_obj_init_draw_rect_dsc(obj, LV_PART_ITEMS, &col_dsc);
-    col_dsc.bg_grad_dir = LV_GRAD_DIR_NONE;
+    col_dsc.bg_grad.dir = LV_GRAD_DIR_NONE;
+    col_dsc.bg_grad.dither = LV_DITHER_NONE;
```
kisvegabor: Is this line required if `bg_grad.dir = LV_GRAD_DIR_NONE;`?

kisvegabor (commented):
Looks great, only 2 minor comments.

src/draw/lv_draw_gradient.c
X-Ryl669: Moved.

src/extra/widgets/chart/lv_chart.c
X-Ryl669: No, it's a copy & paste error. Removed.

src/draw/sw/lv_draw_sw_gradient.h
```diff
@@ -0,0 +1,113 @@
...
+#include "../../misc/lv_style.h"
+#include "lv_draw_sw_dither.h"
+
+/*********************
+ *      DEFINES
+ *********************/
+/** Currently, only 2 stops are allowed per gradient. Increase this to allow more stops.
+ *  This adds (sizeof(lv_color_t) + 1) bytes per additional stop
+ */
+#define LV_GRADIENT_MAX_STOPS    2
```
kisvegabor: One last thing: I think this should configurable in `lv_conf.h`.

kisvegabor (commented):
Only one list things from my side. Let me know if it's ready to be merged!

X-Ryl669:
Example with 3 stops:
```cpp
  static lv_gradient_t grad;
  grad.dir = LV_GRAD_DIR_VER;
  grad.dither = LV_DITHER_ORDERED;
  grad.stops_count = 3;
  grad.stops[0].color = lv_color_hex(0x000000);
  grad.stops[0].frac  = 0;
  grad.stops[1].color = lv_color_hex(0x542293);
  grad.stops[1].frac  = 128;
  grad.stops[2].color = lv_color_hex(0x000000);
  grad.stops[2].frac  = 255;
  lv_style_set_bg_gradient(&style, &grad);
```
gives:
dualGrad

src/draw/sw/lv_draw_sw_gradient.h
X-Ryl669: Ok done.

kisvegabor:
Looks good, merging!

Thank you for this great improvement!

tamisoft:
Hey guys, this commit seems to break the spinner element. I have just finished bisecting. Is there a new flag to set to keep the original behavior? The effect I am seeing is that none of the colors for the arcs are being applied, seems to stick with white.
The code snippet is:
`ui->screen_spinner0 = lv_spinner_create(ui->screen[idx], 1000, 30);

lv_style_t style_screen_spinner0_bg;
lv_style_init(&style_screen_spinner0_bg);
lv_style_set_bg_color(&style_screen_spinner0_bg, lv_color_make(0x20, 0x20, 0x20));
lv_style_set_bg_grad_color(&style_screen_spinner0_bg, lv_color_make(0x00, 0x00, 0x00));
lv_style_set_bg_grad_dir(&style_screen_spinner0_bg, LV_GRAD_DIR_NONE);
lv_style_set_bg_opa(&style_screen_spinner0_bg, 0);
lv_style_set_line_color(&style_screen_spinner0_bg, lv_color_make(0x25, 0x26, 0x2e));
lv_style_set_line_width(&style_screen_spinner0_bg, 16);
lv_obj_add_style(ui->screen_spinner0, &style_screen_spinner0_bg, LV_PART_MAIN);

lv_style_t style_screen_spinner0_indic;
lv_style_init(&style_screen_spinner0_indic);
lv_style_set_line_color(&style_screen_spinner0_indic, lv_color_make(0x80, 0x80, 0xff));
lv_style_set_line_width(&style_screen_spinner0_indic, 8);
lv_obj_add_style(ui->screen_spinner0, &style_screen_spinner0_indic, LV_PART_INDICATOR);
lv_obj_set_pos(ui->screen_spinner0, 58, 10);
lv_obj_set_size(ui->screen_spinner0, 60, 60);`

Used to display a lighter blue spinner on a dark gray background, now it is all white, animation is running, but only minor pixel details are visible as it goes around like this:
ezgif com-gif-maker

alec-2018:
#2997 I had a similar problem with the latest master

X-Ryl669:
Right, that's strange. The code is setting a gradient color, but also no gradient (LV_GRAD_DIR_NONE).  I'll check this and fix it up.

X-Ryl669:
What is your platform for the code ?

tamisoft:
> What is your platform for the code ?

I am writing this for ESP32, and display is RGB565

Also, awesome work you did here :)

X-Ryl669:
Works for me in the simulator. I had to set the style as static (else it's destructed upon leaving the scope and the reference is not working anymore). I'll try on my ESP32 board.

tamisoft:
> Works for me in the simulator. I had to set the style as static (else it's destructed upon leaving the scope and the reference is not working anymore). I'll try on my ESP32 board.

correct, those are statically stored outside, just forgot to include definition...

X-Ryl669:
```
lv_style_set_bg_opa(&style_screen_spinner0_bg, 0);
```
Makes the background completely transparent, so the bg color is ignored and none of the style are copied to the draw descriptor. This means the error as nothing to do with the *new* background drawing code, IIUC. 
It seems line color is not working however, so the culprit might be the draw arc stuff.

X-Ryl669:
Ok, found the culprit, please confirm it works for you:
line 158 of lv_draw_sw_rect.c should read:
```
    blend_dsc.color = bg_color; //dsc->bg_grad.stops[0].color;
```

alec-2018:
> 好的，找到了罪魁祸首，请确认它是否适合您： lv_draw_sw_rect.c 的第 158 行应为：
> 
> ```
>     blend_dsc.color = bg_color; //dsc->bg_grad.stops[0].color;
> ```
it works very nice，thank you!!!

tamisoft:
> Ok, found the culprit, please confirm it works for you: line 158 of lv_draw_sw_rect.c should read:
> 
> ```
>     blend_dsc.color = bg_color; //dsc->bg_grad.stops[0].color;
> ```

This solves it indeed. Thanks

kisvegabor:
I found an other issue which is seemingly related to this feature:
```c
    lv_obj_t * panel = lv_obj_create(lv_scr_act());
    lv_obj_set_height(panel, 450);
    lv_obj_set_width(panel, lv_pct(100));
    lv_obj_set_style_bg_color(panel, lv_color_hex(0xF8F8F9), LV_PART_MAIN | LV_STATE_DEFAULT);
    lv_obj_set_style_bg_opa(panel, 255, LV_PART_MAIN | LV_STATE_DEFAULT);
    lv_obj_set_style_bg_grad_color(panel, lv_color_hex(0xDCE5ED), LV_PART_MAIN | LV_STATE_DEFAULT);
    lv_obj_set_style_bg_grad_dir(panel, LV_GRAD_DIR_VER, LV_PART_MAIN | LV_STATE_DEFAULT);
```
image

This is with 16 bit (wrong on 32 bit too) color depth on 480x272 display. The issue is visible only with vertical gradient. The first incorrect line is close to y=256 so maybe it's on overflow somewhere. 
@X-Ryl669 could you take look at it?

EDIT:
Now I see that the issue starts on the 270th pixel which is the width of the rectangle. Maybe swapped width/height somewhere?

X-Ryl669:
Right, I'll take a look. It's likely an issue with the gradient map size. You are not using dithering here, right ?

kisvegabor:
Thank you! Yes, it's without dithering.

X-Ryl669:
See #3034 for a fix to the issue.

PR #2880, opened 2021-12-04
Fix(examples) Remove semicolons from Python code

ckielstra:
A semi-colon in Python denotes separation, rather than termination.
This commit removes all semicolons at line endings in the Python examples.

Also a few other trivial Python style changes like removing extra empty lines at file endings, missing spaces after an operator, double spaces after an operator.


embeddedt:
Thanks! A lot of these were probably just missed during translation from C to Python.

PR #2881, opened 2021-12-05
LV_LIST: Hide icon/image creation behind #if LV_USE_IMG 

FlashForger:
In case images are not used in projects but lists are

### DIFF
--- a/src/extra/widgets/list/lv_list.c
+++ b/src/extra/widgets/list/lv_list.c
@@ -87,7 +87,7 @@ lv_obj_t * lv_list_add_btn(lv_obj_t * list, const char * icon, const char * txt)
         lv_img_set_src(img, icon);
     }
 #endif
-    
+
     if(txt) {
         lv_obj_t * label = lv_label_create(obj);
         lv_label_set_text(label, txt);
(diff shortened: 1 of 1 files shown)


FlashForger:
I don't understand what is the problem

src/extra/widgets/list/lv_list.c
```diff
@@ -81,11 +81,13 @@ lv_obj_t * lv_list_add_btn(lv_obj_t * list, const char * icon, const char * txt)
...
 
+#if LV_USE_IMG == 1
     if(icon) {
         lv_obj_t * img = lv_img_create(obj);
         lv_img_set_src(img, icon);
     }
-
+#endif
+    
     if(txt) {
```
kisvegabor, line 91: What about adding `#if LV_USE_LABEL` here?

kisvegabor (commented):
Thanks! 
Following your idea, I had a suggestion.

src/extra/widgets/list/lv_list.c
FlashForger: I can hardly imagine a scenario where someone would use Lists without Labels (I can hardly imagine anyone using this library without labels). It'll mean only using lists with buttons, that also have no labels, i.e. empty buttons or with images. But technically possible so it's your call.

kisvegabor: Makes sense :slightly_smiling_face: 

The CI complains about some code formatting issues. Please run `scripts/code-format.py` (astyle needs to be installed)

FlashForger: I ran it, but I don't understand, what should I correct.

embeddedt: Once you run it, it should have reformatted your source code. Now you just have to run `git commit` and push it to your fork.

(If you'd rather just have me correct the formatting, I can also do that.)

FlashForger: Ok, this time it worked

embeddedt:
Thanks!

PR #2882, opened 2021-12-06
Fix set px cb crash

XuNeo:
### disp->driver->set_px_cb write buffer overflow

The coords in below code means the dest_buf includes clip_area data, thus should not modify dest_buf in line423.
```c
disp->driver->set_px_cb(disp->driver, (void *)dest_buf, dest_stride, clip_area->x1 + x, clip_area->y1 + y, src_buf[x],
                                        opa);
```

And buffer width should not overflow `clip_w`.

Include lv_gpu_sdl.mk optionally, to avoid compiling error.

### Checkpoints

### DIFF
--- a/src/draw/sw/lv_draw_sw_blend.c
+++ b/src/draw/sw/lv_draw_sw_blend.c
@@ -420,8 +420,6 @@ static void map_set_px(lv_color_t * dest_buf, lv_coord_t dest_stride, const lv_a
 
     int32_t src_stride = lv_area_get_width(src_area);
 
-    dest_buf += dest_stride * clip_area->y1 + clip_area->x1;
-
     src_buf += src_stride * (clip_area->y1 - src_area->y1);
     src_buf += (clip_area->x1 - src_area->x1);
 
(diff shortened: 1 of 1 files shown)


src/gpu/lv_gpu.mk
```diff
@@ -9,4 +9,4 @@ VPATH += :$(LVGL_DIR)/$(LVGL_DIR_NAME)/src/gpu
 
 CFLAGS += "-I$(LVGL_DIR)/$(LVGL_DIR_NAME)/src/gpu"
 
-include $(LVGL_DIR)/$(LVGL_DIR_NAME)/src/gpu/sdl/lv_gpu_sdl.mk
\ No newline at end of file
+-include $(LVGL_DIR)/$(LVGL_DIR_NAME)/src/gpu/sdl/lv_gpu_sdl.mk
```
embeddedt: I believe this is fixed already in https://github.com/lvgl/lvgl/commit/43729d1502dad0ca797b4b6fb8c69a48c81a2af7, so this change is no longer needed.

XuNeo: Got it. I have just removed it.

kisvegabor:
Thanks!

PR #2883, opened 2021-12-06
fix(snapshot): make fake display size big enough to avoid align issue.

XuNeo:
### Snapshot may be different than content on display

As discussed in #2593 , we cannot avoid this issue, but can improve most cases to make snapshot as close to real content on display.

Below is the result of making this change using lv_example_snapshot_1()

image

### Checkpoints


kisvegabor:
It still doesn't solve the issue if the object is clipped on the left or top, but surely fixes some common use cases. 

Thank you!

PR #2889, opened 2021-12-10
Add support for animated buttons with rlottie

X-Ryl669:

This PR adds a new widget: **Animation button**.
An *animation button* is a button whose state are described as part of a rlottie animation. 

Unlike an `lv_imgbtn`, it's not static (it can be) and instead, it plays part of the animation when in a specific state.
For example, the RELEASED state will play animation from frame 0 to frame 1, and the CHECKED state will play animation from frame 12 to frame 60 in loop, etc...

So, why it's useful ?

1. Even if used as static picture, you only need one "small" file to describe a complex button (instead of multiple files with an image button)
2. You get dynamic interface for free with almost no scripting. Your interface is alive, everything is animated when appropriate.
3. You can draw your buttons in After Effects, Glaxnimate, Synfig, or whatever other vector editing software with animation support and you don't need to render to pictures in your build process. Save as lottie and you're done.
4. Vector rendering meaning it's independent of the button size (ahem, except when you don't have enough RAM, but that's another discussion). Design your button once, no need to support multiple export to different png sizes.
 

### Checkpoints

### DIFF
--- a/src/extra/widgets/animbtn/lv_animbtn.c
+++ b/src/extra/widgets/animbtn/lv_animbtn.c
@@ -169,10 +169,6 @@ static void lv_animbtn_event(const lv_obj_class_t * class_p, lv_event_t * e)
         case LV_EVENT_PRESS_LOST:
             apply_state(obj);
             break;
-        case LV_EVENT_DRAW_MAIN:
-            //draw_main(e);
-            lv_obj_invalidate(animbtn->lottie);
-            break;
         case LV_EVENT_COVER_CHECK: {
                 lv_cover_check_info_t * info = lv_event_get_param(e);
                 if(info->res != LV_COVER_RES_MASKED) info->res = LV_COVER_RES_NOT_COVER;
--- a/src/extra/widgets/animbtn/lv_animbtn.c
+++ b/src/extra/widgets/animbtn/lv_animbtn.c
@@ -25,7 +25,6 @@
  **********************/
 static void lv_animbtn_constructor(const lv_obj_class_t * class_p, lv_obj_t * obj);
 static void lv_animbtn_event(const lv_obj_class_t * class_p, lv_event_t * e);
-static void draw_main(lv_event_t * e);
 static void apply_state(lv_obj_t * animbtn);
 static void loop_state(lv_obj_t * animbtn);
 static lv_animbtn_state_t suggest_state(lv_obj_t * animbtn, lv_animbtn_state_t state);
@@ -61,7 +60,7 @@ lv_obj_t * lv_animbtn_create(lv_obj_t * parent, lv_obj_t * anim)
     lv_obj_t * obj = lv_obj_class_create_obj(MY_CLASS, parent);
     lv_obj_class_init_obj(obj);
     /*Capture the animation picture*/
-    ((lv_animbtn_t*)obj)->lottie = anim;
+    ((lv_animbtn_t *)obj)->lottie = anim;
     lv_obj_set_parent(anim, obj);
     lv_obj_add_flag(anim, LV_OBJ_FLAG_EVENT_BUBBLE);
 
@@ -160,7 +159,6 @@ static void lv_animbtn_event(const lv_obj_class_t * class_p, lv_event_t * e)
     lv_event_code_t code = lv_event_get_code(e);
     lv_obj_t * obj = lv_event_get_current_target(e);
     lv_animbtn_t * animbtn = (lv_animbtn_t *)obj;
-    int is_same = obj == lv_event_get_target(e);
     switch(code) {
         case LV_EVENT_READY:
             /*Should we loop?*/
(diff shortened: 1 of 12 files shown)


X-Ryl669:
It's a WIP, don't merge for now. I need to fix the CI and write tests.

X-Ryl669:
Should be ready for review now.

src/extra/widgets/animbtn/lv_animbtn.c
```diff
@@ -0,0 +1,309 @@
...
+
+    lv_obj_set_size(obj, lv_obj_get_width(anim), lv_obj_get_height(anim));
+    return obj;
+}
+
+/*=====================
+ * Setter functions
+ *====================*/
+
+/**
```
kisvegabor, line 75: This comment should be removed.

src/extra/widgets/animbtn/lv_animbtn.c
```diff
@@ -0,0 +1,309 @@
...
+            loop_state(obj);
+            break;
+        case LV_EVENT_PRESSED:
+        case LV_EVENT_RELEASED:
+        case LV_EVENT_PRESS_LOST:
+            apply_state(obj);
+            break;
+        case LV_EVENT_DRAW_MAIN:
+            //draw_main(e);
+            lv_obj_invalidate(animbtn->lottie);
```
kisvegabor: Why is it needed? An invalidation during drawing can mess up things. 

Lottie animations should invalidate themself as needed.

src/extra/widgets/animbtn/lv_animbtn.h
```diff
@@ -0,0 +1,118 @@
...
+} lv_animbtn_state_t;
+
+/**********************
+ *      TYPEDEFS
+ **********************/
+/*State status for anim button*/
+typedef struct {
+    size_t first_frame;
+    size_t last_frame;
+    size_t control; /* A lv_rlottie_ctrl_t instance and only checking Forward/Backward and Loop here */
```
kisvegabor: Really size_t is the appropriate type here?

src/extra/widgets/animbtn/lv_animbtn.h
```diff
@@ -0,0 +1,118 @@
...
+
+/*=====================
+ * Setter functions
+ *====================*/
+
+/**
+ * Set animation for a state of the animation button
+ * @param animbtn pointer to an animation button object
+ * @param state for which state set the new animation
+ * @param desc description of what to do when the button is in this state
```
kisvegabor, line 76: Please mention if it can be a local variable that can go out of the scope or not.

src/extra/widgets/animbtn/lv_animbtn.c
```diff
@@ -0,0 +1,309 @@
...
+ * @return pointer to the created image button
+ */
+lv_obj_t * lv_animbtn_create(lv_obj_t * parent, lv_obj_t * anim)
+{
+    LV_LOG_INFO("begin");
+    lv_obj_t * obj = lv_obj_class_create_obj(MY_CLASS, parent);
+    lv_obj_class_init_obj(obj);
+    /*Capture the animation picture*/
+    ((lv_animbtn_t *)obj)->lottie = anim;
+    lv_obj_set_parent(anim, obj);
```
kisvegabor, line 64: Why can't we use the rlottie object directly? I.e. why is it required to wrap it into a lv_obj?

Without it, we could you same API as lv_rlottie_create does.

Just thinking aloud: if it worked maybe we could even move all these into `lv_rlottie`. Hm... But probably keeping basic animation and the button-like functionalities separate is a better idea.

kisvegabor (commented):
I think it's an awesome idea to use parts of the same lottie files for the different states of a button.

Once this widget works only with rlottie, what about calling it `rlottiebtn`? `animbtn` feels too generic to me.

I'm thinking about a use case like this:
- in released state a red button pulses in a loop (frame 0..20 looped)
- in pressed state the blue button pulses in a loop (frame 30..50 looped)
- there is a transition from released to pressed (frame 20..30)
- there is a transition from pressed to released (frame 60..70)

I see 2 possible problems:
1. how to play the transition part and loop only the pulsing part. I.e. 20->30->50->30->50
2. If I press the button on frame 12 it will jump to frame 20 (beginning of the transition) and starts to loop, So 8 frames are skipped causing a flickering.

I'd suggest extending this:
```c
typedef struct {
    size_t start_frame; //start from here, if -1 start from the current frame
    size_t first_frame;  //loop from here
    size_t last_frame;
    size_t control; 
} lv_animbtn_state_desc_t;
```

What do you think?

X-Ryl669:
> Once this widget works only with rlottie, what about calling it rlottiebtn? animbtn feels too generic to me.

Honestly, when I started implementing, I was really lurking on Thorvg + Rive. These have multiple pros:
1. Rive support transition animation and state with many advanced animation features
2. Thorvg uses it own format that's extracted from SVG making very very small vector based file size
3. It's still maintained (unlike rlottie)

But the cons were against this:
1. Thorvg is not using CMake, so we'd have had to rewrite the build procedure to use CMake
2. There's only one application that's making Rive files and it's not free
3. Thorvg is quite big, so it'd be redundant with rlottie, IMHO
4. In fact, if using Thorvg, then a large part of LVGL renderer could be dropped. It does not make sense to have Thorvg render to a memory buffer and then blit it, since it's consuming a large portion of the limited RAM. Instead, it would be better to let Thorvg directly render to the display (or the small portion of it that's currently active). This implies not using Thorvg as a image decoder, but as a new rendering engine. And since it's a lot more featureful than LVGL's own rendering capabilities (like bezier, arc, mask, stencil, etc...), would it make sense to keep LVGL renderer? Yet, it's a lot of work.

So in the end, I used rlottie. Sure we can rename to rlottiebtn, but I *secretly* hope someone to write a thorvg version :crossed_fingers: 


> how to play the transition part and loop only the pulsing part. I.e. 20->30->50->30->50 ?

Honestly, the semantic of transiting from one state to another state does not exist in LVGL. We could fake it by using the PRESSED state to trigger a "transition" animation (since the press action is supposedly short), but there is no such thing currently in a obj's state. 
Please notice that rlottie is not really made for handling state (although they have been attempts to do so). Rive is better suited for this, IMHO. 

> If I press the button on frame 12 it will jump to frame 20 (beginning of the transition) and starts to loop, So 8 frames are skipped causing a flickering.

I was thinking about this when I started implementing and I came to this conclusion: don't let the animation finish.
Let me explain why:
   1. Let's say you have a loop of 50 frames (so 2s at 25fps). 
   2. If the user pressed the button when the animation was at frame 1, she would have to wait until the animation ends to see the effect of her actions
   3. But 2s is a long time for a "visual" reaction, so it's very likely she'll think "it didn't work, let's retry"
   4. Pressing the button again meanwhile will revert the state to the initial state and even if the animation finishes, it'll loop and play again
   5. In the end, the user will think "well, it's an endless animation, it's not a button"

So I really think it's better to glitch the animation to directly transit to the new state animation.

> I'd suggest extending this: // adding start frame
> What do you think?

Yes that's a good idea. When entering a state we can play a large part of the animation once and then only loop to a smaller part.

src/extra/widgets/animbtn/lv_animbtn.c
X-Ryl669: Copy and paste woes... Will do.

src/extra/widgets/animbtn/lv_animbtn.c
X-Ryl669: Indeed, I've just tested it and it's not required. Will remove it.

src/extra/widgets/animbtn/lv_animbtn.h
X-Ryl669: It's a classic chicken and egg here. I can't declare the type as `lv_rlottie_ctrl_t` since it's an enum and you can't forward declare an enum. If I include `lv_rlottie.h`, it includes `lvgl.h` which in turns include `lv_animbtn.h`. Since all files are protected by ifdef/define guard, so in the end, it does not work either. 

A better design would be to recreate the control we actually support making sure they match those of `lv_rlottie_ctrl_t`

src/extra/widgets/animbtn/lv_animbtn.h
X-Ryl669: I don't get you. There's no pointer (except for the animbtn instance, obviously), so the usual copy by value rules apply here.

src/extra/widgets/animbtn/lv_animbtn.c
X-Ryl669: This button is just that, a wrapper that handles event and sets the the animation state. It can be done from outside, but it's troublesome to get right, hence the object. I'm using a object here because I have to capture the PRESS/CLICKED events and match them to button state, which would not make sense for an image IMHO. Also, I'm capturing the animation's EVENT_READY event here to restart the loop upon termination, and I haven't found a way to do that cleanly without becoming a parent of it.

src/extra/widgets/animbtn/lv_animbtn.h
```diff
@@ -30,14 +30,17 @@ typedef enum {
     _LV_ANIMBTN_STATE_NUM,
 } lv_animbtn_state_t;
 
-/**********************
- *      TYPEDEFS
- **********************/
+typedef enum {
+    LV_ANIMBTN_CTRL_FORWARD  = 0,
+    LV_ANIMBTN_CTRL_BACKWARD = 1,
+    LV_ANIMBTN_CTRL_LOOP     = 8,
```
kisvegabor: Why don't we use the values from `lv_rlottie_ctrl_t`, like `LV_RLOTTIE_CTRL_LOOP`?

X-Ryl669: Sure, let's do that. 
I'm not using `lv_rlottie_ctrl_t` since not all value are valid here (only forward, backward and loop)

kisvegabor:
I also know and see a big potential in ThorVG. I had a [discussion](https://github.com/Samsung/thorvg/issues/1014) with a developer who said they will move Rlottie to ThorVG. So maybe we only needs to be patient. :slightly_smiling_face: 

> Thorvg is not using CMake, so we'd have had to rewrite the build procedure to use CMake

Yes, it's Meson, but we hopefully we can use Meson to build a lib from ThorVG and use that ready-to-use lib with the rest of the projects. Not ideal but acceptable IMO.

> There's only one application that's making Rive files and it's not free

Yes, it's a pity. If they integrated Rlottie, this problem would disappear.

 > Thorvg is quite big, so it'd be redundant with rlottie, IMHO

In the end, seeming ThorVG will be superior to Rlottie.

> In fact, if using Thorvg, then a large part of LVGL renderer could be dropped.

Oh yes, it was one of my main drivers to make LVGL's drawing architecture more abstract, i.e. to use ThorVG,  [NanoSVG](https://github.com/memononen/nanosvg) or [Cairo](https://www.cairographics.org/) as a super advanced draw engine. I'm so happy that you also find it a good step :slightly_smiling_face: 

> And since it's a lot more featureful than LVGL's own rendering capabilities (like bezier, arc, mask, stencil, etc...), would it make sense to keep LVGL renderer? Yet, it's a lot of work.

IMO, the vector renderer and the current renderer should exist in parallel and the user should be able to select which one to use. It's because the current engine is:
- probably faster (as it's simpler and uses only integers)
- needs less memory
- easy to build
- we have full control over it
- work together with the MCUs GPUs to blend and fill

But for larger MCUs, e.g. an SMT32H7 or NXP iMX RT1060 ThorVG (or so) it would be a very interesting choice. 

We were talking about the widget buffering feature in the dither PR, but it can come handy here too. If you have a lot of SVG buttons you can buffer one, use it as an image on multiple places. If you have more memory you can buffer more stuff.

> I was thinking about this when I started implementing and I came to this conclusion: don't let the animation finish.
Let me explain why:

For a 2s long animation it'd be really ugly. But I'd leave the option for the user handle it as he wishes. If the animation is short (e.g. 200 ms) he can use this trick. Anyway, I don't insist on it. :slightly_smiling_face:

X-Ryl669:
Can't refer to `LV_RLOTTIE_CTRL_XXX` as it breaks building (see previous commit message for the error). 
As for the improvement you've requested, I agree. I'll make another PR after this one so we can discuss the best option to implement them. I think having a threshold time to figure out if we finish the animation (or maybe accelerating it by skipping frames so it plays in 200ms everytime?) might be better. I need to make some tests here and show the results so we can decide what looks (and feel) better.

kisvegabor:
I played removing the wrapper obj, and it's really not straight forward sue to to non standard prototype of `lv_rlottie_create`.

But meanwhile, I solved the recursive include issue with:
```diff
diff --git a/src/extra/libs/rlottie/lv_rlottie.h b/src/extra/libs/rlottie/lv_rlottie.h
index 5c3a7a222..b311895f4 100644
--- a/src/extra/libs/rlottie/lv_rlottie.h
+++ b/src/extra/libs/rlottie/lv_rlottie.h
@@ -13,7 +13,7 @@ extern "C" {
 /*********************
  *      INCLUDES
  *********************/
-#include "../../../lvgl.h"
+#include "../../../widgets/lv_img.h"
 #if LV_USE_RLOTTIE
 
 /*********************
diff --git a/src/extra/widgets/animbtn/lv_animbtn.h b/src/extra/widgets/animbtn/lv_animbtn.h
index 8deea0bdb..a55d9d16c 100644
--- a/src/extra/widgets/animbtn/lv_animbtn.h
+++ b/src/extra/widgets/animbtn/lv_animbtn.h
@@ -13,7 +13,6 @@ extern "C" {
 /*********************
  *      INCLUDES
  *********************/
-#include "../../../lvgl.h"
 #include "../../libs/rlottie/lv_rlottie.h"
 
 #if LV_USE_RLOTTIE != 0 && LV_USE_ANIMBTN != 0
```

kisvegabor:
Thanks for the updates. I suggest waiting a little with this PR until the image decoder PR stabilizes.

X-Ryl669:
Not stale

kisvegabor:
Can it be closed as #2904 contains `lv_animbtn` too?

X-Ryl669:
Yes. I'll need to add some "transition" description too and I'll consider it done.

PR #2897, opened 2021-12-14
fix(qrcodegen): fix coding style/build error

brgl:
The way the brackets around this if-else are structured doesn't look
broken but it causes the following build error on some compilers (first
observed on arm-zephyr-eabi-gcc (crosstool-NG 1.24.0.378_e011758) 10.3.0):

./src/extra/libs/qrcode/qrcodegen.c:953:2: error: 'else' without a previous 'if'
  953 |  else if (assignVal < (1 << 7)) {
      |  ^~~~

Fix it by placing brackets consistently.

### DIFF
--- a/src/extra/libs/qrcode/qrcodegen.c
+++ b/src/extra/libs/qrcode/qrcodegen.c
@@ -948,9 +948,9 @@ struct qrcodegen_Segment qrcodegen_makeEci(long assignVal, uint8_t buf[]) {
 	result.mode = qrcodegen_Mode_ECI;
 	result.numChars = 0;
 	result.bitLength = 0;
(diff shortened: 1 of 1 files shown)


embeddedt:
Is it the brackets or is it the platform defining assert as an empty macro?

brgl:
Eek, you're right. This still helps though, because at least there are empty brackets now to tell the compiler it should still take this branch.

embeddedt:
I've merged the change, but I'm not sure if there are other parts of the codebase with the same issue. It would be safer for `assert` to be defined as `do {} while(0)`, as that would still be optimized out by the compiler and counts as a statement, avoiding the problem with missing brackets.

PR #2898, opened 2021-12-14
lv_txt:_lv_txt_get_next_line() add used_width parameter

guoweilkd:

Usually, when calling _lv_txt_get_next_line(..., flag == LV_TEXT_FLAG_NONE), the line width can be obtained. So there is no need to call lv_txt_get_width() to recalculate the line width. For some dynamically loaded fonts (eg: FreeType), this optimization may be useful. It can reduce the call of get_glyph_dsc() interface. 

In addition, is it necessary for the following conditions to return used_width？
```c
uint32_t _lv_txt_get_next_line(xxx)
{
.....
    /*If max_width doesn't mater simply find the new line character
     *without thinking about word wrapping*/
    if((flag & LV_TEXT_FLAG_EXPAND) || (flag & LV_TEXT_FLAG_FIT)) {
        uint32_t i;
        for(i = 0; txt[i] != '\n' && txt[i] != '\r' && txt[i] != '\0'; i++) {
            /*Just find the new line chars or string ends by incrementing `i`*/
        }
        if(txt[i] != '\0') i++;    /*To go beyond `\n`*/
        return i;
    }
.....
```

### Checkpoints

### DIFF
--- a/src/draw/lv_draw_label.c
+++ b/src/draw/lv_draw_label.c
@@ -141,13 +141,13 @@ LV_ATTRIBUTE_FAST_MEM void lv_draw_label(const lv_area_t * coords, const lv_area
         pos.y += hint->y;
     }
 
-    uint32_t line_end = line_start + _lv_txt_get_next_line(&txt[line_start], font, dsc->letter_space, w, dsc->flag);
+    uint32_t line_end = line_start + _lv_txt_get_next_line(&txt[line_start], font, dsc->letter_space, w, NULL, dsc->flag);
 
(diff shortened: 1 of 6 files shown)


kisvegabor:
It seems like a useful change.

> In addition, is it necessary for the following conditions to return used_width？

Actually the point of handling this special case is to avoid actually calculating the line's real width. So we should not return `used_width` here. 

But something like this could be added:
```c
 if(used_width) *used_width = -1;
```

kisvegabor:
Looks good, thanks!

PR #2902, opened 2021-12-15
feat(lottie) move rlottie_capi.h to lv_rlottie.c

xaowang96:

There is no need to include rlottie_capi.h in lv_rlottie.h. so i move it to lv_rlottie.c. 
After deleting lv_rlottie obj, the memory of the image source will be released,  but it is not invalidated, causing a crash.

### Checkpoints

### DIFF
--- a/src/extra/libs/rlottie/lv_rlottie.h
+++ b/src/extra/libs/rlottie/lv_rlottie.h
@@ -16,8 +16,6 @@ extern "C" {
 #include "../../../lvgl.h"
 #if LV_USE_RLOTTIE
 
-#include <rlottie_capi.h>
-
 /*********************
  *      DEFINES
  *********************/
@@ -33,9 +31,11 @@ typedef enum {
     LV_RLOTTIE_CTRL_LOOP     = 8,
 } lv_rlottie_ctrl_t;
 
+/** definition in lottieanimation_capi.c */
+struct Lottie_Animation_S;
 typedef struct {
     lv_img_t img_ext;
-    Lottie_Animation * animation;
+    struct Lottie_Animation_S * animation;
     lv_timer_t * task;
     lv_img_dsc_t imgdsc;
     size_t total_frames;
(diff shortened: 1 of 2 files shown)


X-Ryl669:
Right now, I'm working on a rlottie image decoder to allow for partial rendering. For example, if you have a 64x64 picture, it'll allocate only 16 lines and proceed to 4 renders (each time setting the rendering clip area correctly).
This way, we don't need the huge memory requirement for the rendering buffer that it currently has, exactly like LVGL is doing for the rendering of other objects.

If you are working with rlottie on ESP32, like I'm guessing it, it might completely change the experience, not requiring a PSRAM anymore. I'll keep you informed about this progress.

src/extra/libs/rlottie/lv_rlottie.h
```diff
@@ -33,9 +31,11 @@ typedef enum {
     LV_RLOTTIE_CTRL_LOOP     = 8,
 } lv_rlottie_ctrl_t;
 
+/** definition in lottieanimation_capi.c */
+struct Lottie_Animation_S;
 typedef struct {
     lv_img_t img_ext;
-    Lottie_Animation * animation;
+    struct Lottie_Animation_S * animation;
```
X-Ryl669, line 38: You need to forward declare the structure to use a pointer to it. You can add a `struct Lottie_Animation;` above this declaration. Else the compiler will error here.

xaowang96:
> Right now, I'm working on a rlottie image decoder to allow for partial rendering. For example, if you have a 64x64 picture, it'll allocate only 16 lines and proceed to 4 renders (each time setting the rendering clip area correctly). This way, we don't need the huge memory requirement for the rendering buffer that it currently has, exactly like LVGL is doing for the rendering of other objects.
> 
> If you are working with rlottie on ESP32, like I'm guessing it, it might completely change the experience, not requiring a PSRAM anymore. I'll keep you informed about this progress.

That is a good idea to reduce memory consumption.

kisvegabor:
@xaowang96 Both make sense, thanks! The forward declaration issue is solved, so I merge it. 

@X-Ryl669 
> If you are working with rlottie on ESP32, like I'm guessing it, it might completely change the experience, not requiring a PSRAM anymore. I'll keep you informed about this progress.

It would be huge! Feel free to open an issue for it you need to discuss something about it.

X-Ryl669:
See #2904 for the proposed implementation. The memory usage reduction is huge.

PR #2907, opened 2021-12-16
fix(label): do not add DOT when the width of label is enough

xaowang96:

When the width of the label is enough, there is no need to add DOT at the end of its text.

### Checkpoints

### DIFF
--- a/src/widgets/lv_label.c
+++ b/src/widgets/lv_label.c
@@ -202,7 +202,8 @@ void lv_label_set_long_mode(lv_obj_t * obj, lv_label_long_mode_t long_mode)
     label->offset.x = 0;
     label->offset.y = 0;
 
-    if(long_mode == LV_LABEL_LONG_SCROLL || long_mode == LV_LABEL_LONG_SCROLL_CIRCULAR || long_mode == LV_LABEL_LONG_CLIP || long_mode == LV_LABEL_LONG_DOT)
+    if(long_mode == LV_LABEL_LONG_SCROLL || long_mode == LV_LABEL_LONG_SCROLL_CIRCULAR || long_mode == LV_LABEL_LONG_CLIP ||
+       long_mode == LV_LABEL_LONG_DOT)
         label->expand = 1;
     else
         label->expand = 0;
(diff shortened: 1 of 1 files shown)


kisvegabor:
Can you add an example to reproduce the issue this PR solves?

xaowang96:
Here is the code and running results.

```C
void lvgl_test(void)
{
    lv_obj_t* screen = lv_scr_act();
    lv_obj_set_style_bg_color(screen, lv_color_black(), 0);

    lv_obj_t* text = lv_label_create(screen);
    lv_label_set_long_mode(text, LV_LABEL_LONG_DOT);
    lv_label_set_text(text, "Hello World");
    lv_obj_set_size(text, 400, 30);
    lv_obj_set_style_text_color(text, lv_color_white(), 0);
    lv_obj_set_style_text_font(text, &lv_font_montserrat_28, 0);
    lv_obj_set_style_text_align(text, LV_TEXT_ALIGN_LEFT, 0);
    lv_obj_center(text);

    lv_obj_set_style_border_color(text, lv_color_hex(0xFF0000), 0);
    lv_obj_set_style_border_width(text, 1, 0);
    lv_obj_set_style_border_opa(text, LV_OPA_100, 0);
}
```
截屏2021-12-17 上午11 17 24

 But if the height of the label is enough, the performance is different.

```C
lv_obj_set_size(text, 400, 40);
```

截屏2021-12-17 上午11 21 07

xaowang96:
> Can you add an example to reproduce the issue this PR solves?

The comment above reproduced the problem.

kisvegabor:
There dots because the height of the object is too small, and LVGL things that the text can't fit vertically to the object and adds dots to indicate the missing lines. 

It's more visible with:
```c
    lv_label_set_text(text, "Hello World 1\n"
                             "Hello World 2\n"
                             "Hello World 3\n"
                             "Hello World 4\n");
     lv_obj_set_size(text, 300, 65);
```

xaowang96:
> There dots because the height of the object is too small, and LVGL things that the text can't fit vertically to the object and adds dots to indicate the missing lines.
> 
> It's more visible with:
> 
> ```c
>     lv_label_set_text(text, "Hello World 1\n"
>                              "Hello World 2\n"
>                              "Hello World 3\n"
>                              "Hello World 4\n");
>      lv_obj_set_size(text, 300, 65);
> ```

This is correct for multiple lines of text, but if there is only one line of text, it may be better to show the complete text as much as possible.

kisvegabor:
I agree and pushed a commit to make it behave like this.

PR #2925, opened 2021-12-20
fix lv_dropdown_get_selected_str

paulh002:
This fix checks for nullptr when the dropbox does not have content
Added a check for nullptr on dropdown->options if NULL then return buf[0]='\0';


kisvegabor:
Looks good thanks.

Just please run  `scripts/code-format.py` or format the code like [this](https://github.com/lvgl/lvgl/runs/4583375375?check_suite_focus=true#step:6:28).

paulh002 (commented):
updated formatting

paulh002:
have run scripts/code-format.py

kisvegabor:
Merging, thank you!

Issue #2922, opened 2021-12-20
V8 couple of functions with parameters different in header


xennex22:
https://github.com/lvgl/lvgl/blob/01daed356a4eb540de77956f6c6d0ba86c4969b9/src/extra/layouts/flex/lv_flex.c#L57-L58
https://github.com/lvgl/lvgl/blob/01daed356a4eb540de77956f6c6d0ba86c4969b9/src/extra/layouts/flex/lv_flex.c#L317-L318

https://github.com/lvgl/lvgl/blob/01daed356a4eb540de77956f6c6d0ba86c4969b9/src/extra/widgets/tileview/lv_tileview.h#L51
https://github.com/lvgl/lvgl/blob/01daed356a4eb540de77956f6c6d0ba86c4969b9/src/extra/widgets/tileview/lv_tileview.c#L65

kisvegabor:
Hi,

Thanks for reporting it. Could you open a PR?

PR #2929, opened 2021-12-20
Minor fixes to function parameter order

xennex22:
Fixes for various problems from https://github.com/lvgl/lvgl/issues/2922  I believe I have the col/row order correct.

Also changed several calls to functions taking `lv_anim_enable_t` as a parameter which were passing a bool.

Also changed `lv_tlsf_free` to use a `const void *` rather than `void *`.  This fixes compiler warnings from `lv_mem_free`.


kisvegabor:
I've already fixed the perf monitor issue. Could you remove the last commit?

kisvegabor:
Thank you!

PR #2931, opened 2021-12-21
fix(player demo): encoding problem

mysterywolf:
Some compilers don't recognize • because of different encoding solutions (e.g. vs2019's encoding is 936 by default if using Chinese Language Windows OS)
I suggest that use FULL ASCII to write demos and examples to avoid encoding issues, in order to let different areas and countires people to run demos and examples easily and perfectly.
image

### Checkpoints


kisvegabor:
Makes sense, thank you.

Issue #2934, opened 2021-12-21
Custom blend mode about commit db53ea92


mariotaku:
### Introduce the problem

> @mariotaku I've just discovered this new blend mode. Is it really required? I'm asking because it's much slower to draw with a non-normal blending mode and it'd important to draw these screen sized parts as fast as possible.

### Examples and cases

Let's assume this case:
One app is an OSD app. It has translucent background. For each frame, the part where redrawn is needed will have to clear the whole buffer, then draw `lv_disp_t.bg_color` with opacity (or completely transparent).

### Suggested solution

 For this case, drawing backend doesn't know if the rect is the bottom-most display background, so it has to be drawn in this way, or, introduce another field called `is_disp_bg` for example for drawer to handle.

kisvegabor:
Thanks, let's continue in the PR.

PR #2935, opened 2021-12-21
BG redraw patch for #2934

mariotaku:
reduced usage of blend mode
introduced draw_bg for draw_ctx

### Checkpoints


kisvegabor:
I see what the problem is. In an earlier LVGL version the draw buffers was cleared with `memset(buf, 0x00)` before rendering so blending worked well. But as the buffer can be a texture too we can't do anything like this. Besides this way we'd write the background twice. 

Anyway, I like the idea of `draw_bg`! :slightly_smiling_face:

PR #2940, opened 2021-12-23
extra/fragment (UI Controller concept)

mariotaku:

Fragment is a concept copied from [Android](https://developer.android.com/guide/components/fragments).

It represents a reusable portion of your app's UI. A fragment defines and manages its own layout, has its own lifecycle, and can handle its own events. Like Android's Fragment that must be hosted by an activity or another fragment, Fragment in LVGL needs to be hosted by an object, or another fragment. The fragment’s view hierarchy becomes part of, or attaches to, the host’s view hierarchy.

Such concept also has some similarities to [UiViewController on iOS](https://developer.apple.com/documentation/uikit/uiviewcontroller).

## Usage

Enable `LV_USE_FRAGMENT` in `lv_conf.h`.

### Create Fragment Class

```c
struct sample_fragment_t {
    /* IMPORTANT: don't miss this part */
    lv_fragment_t base;
    /* States, object references and data fields for this fragment */
    const char *counter;
};

const lv_fragment_class_t sample_cls = {
        /* Initialize something needed */
        .constructor_cb = sample_fragment_ctor,
        /* Create view objects */
        .create_obj_cb = sample_fragment_create_obj,
        /* IMPORTANT: size of your fragment struct */
        .instance_size = sizeof(struct sample_fragment_t)
};
```

### Use `lv_fragment_manager`

```c
/* Create fragment instance, and objects will be added to container */
lv_fragment_manager_t *manager = lv_fragment_manager_create(container, NULL);
/* Replace current fragment with instance of sample_cls, and init_argument is user defined pointer */
lv_fragment_manager_replace(manager, &sample_cls, init_argument);
```

### Fragment Based Navigation

```c
/* Add one instance into manager stack. View object of current fragment will be destroyed,
 * but instances created in class constructor will be kept.
 */
lv_fragment_manager_push(manager, &sample_cls, NULL);

/* Remove the top most fragment from the stack, and bring back previous one. */
lv_fragment_manager_pop(manager);
```

### Checkpoints


mariotaku:
What do you think about this mechanism?

kisvegabor:
Hi,

I've already took a quick look, but still need some time immerse in it. I can review it in a few days.

examples/others/controller/lv_example_controller_1.c
```diff
@@ -0,0 +1,42 @@
...
+
+#if LV_USE_OBJ_CONTROLLER && LV_BUILD_EXAMPLES
+
+static void sample_controller_ctor(lv_obj_controller_t *self, void *args);
+
+static lv_obj_t *sample_controller_create_obj(lv_obj_controller_t *self, lv_obj_t *parent);
+
+struct sample_controller_t {
+    lv_obj_controller_t base;
+    const char *counter;
```
kisvegabor: Why is it called `counter`? `title` or so seems more reasonable to me.

src/extra/others/controller/lv_obj_controller.c
```diff
@@ -0,0 +1,372 @@
...
+typedef struct manager_stack_t {
+    const lv_obj_controller_class_t * cls;
+    lv_obj_controller_t * controller;
+    bool obj_created;
+    bool destroying_obj;
+    bool dialog;
+    struct manager_stack_t * prev;
+} manager_stack_t;
+
+struct lv_controller_manager_t {
```
kisvegabor: In other parts of LVGL this convention is followed:
```c
struct _lv_controller_manager_t {
    lv_obj_t * container;
    lv_obj_controller_t * parent;
    manager_stack_t * top;
}lv_controller_manager_t ;
```

Please use it here too for consistency.

src/extra/others/controller/lv_obj_controller.c
```diff
@@ -0,0 +1,372 @@
...
+ *      MACROS
+ **********************/
+
+
+
+/**********************
+ *   GLOBAL FUNCTIONS
+ **********************/
+
+lv_controller_manager_t * lv_controller_manager_create(lv_obj_t * container, lv_obj_controller_t * parent)
```
kisvegabor: Which API function can return an `lv_obj_controller_t * ` that can be used as `parent`?

src/extra/others/controller/lv_obj_controller.c
```diff
@@ -0,0 +1,372 @@
...
+    return controller;
+}
+
+void lv_obj_controller_pop(lv_obj_controller_t * controller)
+{
+    LV_ASSERT(controller);
+    lv_controller_manager_t * manager = controller->manager;
+    LV_ASSERT(manager);
+    LV_ASSERT(manager->top->controller == controller);
+    lv_controller_manager_pop(manager);
```
kisvegabor: I found a names a little bit incontinent and hard to follow in some places.
- `manager` and `controller` have a similar meaning. It's hard to imagine what is difference
- Some functions start with `lv_controller` and other with `lv_obj_controller`. It should be identical to the name of the file.

src/extra/others/controller/lv_obj_controller.c
```diff
@@ -0,0 +1,372 @@
...
+
+lv_obj_controller_t * lv_controller_manager_top_controller(lv_controller_manager_t * manager)
+{
+    LV_ASSERT(manager);
+    manager_stack_t * top = manager->top;
+    if(!top)return NULL;
+    return top->controller;
+}
+
+lv_obj_controller_t * lv_controller_manager_parent(lv_controller_manager_t * manager)
```
kisvegabor: In the name of functions like this there should be a `get` word. E.g. `lv_controller_manager_get_parent`.

Shouldn't there be a `set` counterpart?

src/extra/others/controller/lv_obj_controller.h
```diff
@@ -0,0 +1,216 @@
...
+ * @param cls Controller class which must return valid lv_msgbox instance
+ * @param args
+ */
+void lv_controller_manager_show(lv_controller_manager_t * manager, const lv_obj_controller_class_t * cls, void * args);
+
+lv_obj_controller_t * lv_controller_manager_top_controller(lv_controller_manager_t * manager);
+
+lv_obj_controller_t * lv_controller_manager_parent(lv_controller_manager_t * manager);
+
+lv_obj_controller_t * lv_obj_controller_class_create_unmanaged(const lv_obj_controller_class_t * cls, lv_obj_t * parent,
```
kisvegabor: It's not clear to me what `unmanaged` exactly mean. Could you add more details to the docs to describe as much features as possible?

kisvegabor (commented):
I've add some comments. I have some other questions too but would like to discuss these first. 

I don't have any mobile app development background and probably that's why I'm not familiar with the terminology. However, we can't assume that LVGL users know anything about Android or iOS development either. So we should be very verbose about how these things work.

mariotaku:
Thank you for your review! I'll process those comments.

examples/others/controller/lv_example_controller_1.c
mariotaku: Changed.

src/extra/others/controller/lv_obj_controller.c
mariotaku: Do you mean rename `struct lv_controller_manager_t` to `struct _lv_controller_manager_t` while keeping the typedef same? I have one question as clang-tidy complains about name starting with `_` is reserved.

src/extra/others/controller/lv_obj_controller.c
mariotaku: Functions in controller lifecycle will have access to `lv_obj_controller_t *`. e.g. `lv_obj_controller_class_t.constructor_cb`.

src/extra/others/controller/lv_obj_controller.c
mariotaku: * What about replacing `manager` with `navigator`?
* `lv_controller_manager_` will be changed to `lv_controller_navigator` then.

src/extra/others/controller/lv_obj_controller.c
mariotaku: Changed. This property is read only and specified upon creation so it can't have `set` counterpart.

src/extra/others/controller/lv_obj_controller.c
kisvegabor: > Do you mean rename struct lv_controller_manager_t to struct _lv_controller_manager_t while keeping the typedef same? 

Yes!

> I have one question as clang-tidy complains about name starting with _ is reserved.

We violate this rule on other places too :slightly_smiling_face:

src/extra/others/controller/lv_obj_controller.c
kisvegabor: I mean how can I get pointer that can be passed as `parent`?

src/extra/others/controller/lv_obj_controller.c
kisvegabor: I'm not sure I fully understand the roles of manager and controller. `manager` can replace, push or pop `controller`s, right?

src/extra/others/controller/lv_obj_controller.c
kisvegabor: Thanks!

src/extra/others/controller/lv_obj_controller.c
mariotaku: Yes. `manager` is a stack to manage controllers.

kisvegabor: I see, slowly I'm starting to wrap my head around this concept. :slightly_smiling_face:  It seems really useful to manage life cycles and deal with states. 

I think the word `manager` is fine, but `obj_controller` is strange to me. What does it control? 

Form your first comment: 
> UiViewController on iOS or Fragment on Android

I think Fragment describes the purpose better, as it's really a UI fragment. `view_controller` is a little bit long but precise. 

If I imagine myself into the place of a new user, from `fragment` and `fragment_manager` I'd imagine what it does and what is the relation between the 2 things. (Assuming I know fragment means a UI fragment)

src/extra/others/controller/lv_obj_controller.c
mariotaku: like this:

```c
void my_obj_created_cb(struct lv_obj_controller_t * self, lv_obj_t * container) {
    my_controller * controller = self;
    controller->child_manager = lv_controller_manager_create(controller->container_obj, self);
}
```

src/extra/others/controller/lv_obj_controller.c
mariotaku: I see. I'll rename this one.

src/extra/others/controller/lv_obj_controller.c
mariotaku: I'm OK with fragment naming thing. So let's rename it to fragment and fragment_manager.

src/extra/others/controller/lv_obj_controller.c
kisvegabor: I see! So a manager can have only one child?

src/extra/others/controller/lv_obj_controller.c
kisvegabor: Great!

src/extra/others/controller/lv_obj_controller.h
mariotaku: I have added some descriptions.

src/extra/others/controller/lv_obj_controller.c
mariotaku: Actually they can have whatever they want. Nested manager however, doesn't persist stack yet. If there is need then I'll implement in the future.

kisvegabor: > Actually they can have whatever they want. 

How can I add two children then?

mariotaku:
```c
void my_obj_created_cb(struct lv_obj_controller_t * self, lv_obj_t * container) {
    my_controller * controller = self;
    controller->child_manager1 = lv_controller_manager_create(controller->container_obj1, self);
    controller->child_manager2 = lv_controller_manager_create(controller->container_obj2, self);
}
```

Such use case can be found when there are multiple panes.

examples/others/fragment/lv_example_fragment.h
```diff
@@ -0,0 +1,37 @@
+/**
+ * @file lv_example_fragment.h
+ */
+#ifndef LV_EXAMPLE_fragment_H
+#define LV_EXAMPLE_fragment_H
```
kisvegabor:
```suggestion
#ifndef LV_EXAMPLE_FRAGMENT_H
#define LV_EXAMPLE_FRAGMENT_H
```

src/extra/others/fragment/lv_fragment.c
```diff
@@ -0,0 +1,392 @@
...
+ **********************/
+
+lv_fragment_manager_t * lv_fragment_manager_create(lv_obj_t * container, lv_fragment_t * parent)
+{
+    LV_ASSERT(container);
+    lv_fragment_manager_t * instance = lv_mem_alloc(sizeof(lv_fragment_manager_t));
+    instance->parent = parent;
+    instance->container = container;
+    instance->top = NULL;
+    return instance;
```
kisvegabor: It'd be great to add an `LV_EVENT_DELETED` event to `container` to automatically remove the `manager` when the `container` is deleted.

src/extra/others/fragment/lv_fragment.c
```diff
@@ -0,0 +1,392 @@
...
+}
+
+void lv_fragment_manager_del(lv_fragment_manager_t * manager)
+{
+    LV_ASSERT(manager);
```
kisvegabor:
```suggestion
    LV_ASSERT_NULL(manager);
```

src/extra/others/fragment/lv_fragment.c
```diff
@@ -0,0 +1,392 @@
...
+    if(old) {
+        item_destroy_obj(manager, old);
+        item_destroy_fragment(old);
+        lv_mem_free(old);
+    }
+    manager->top = top;
+    item_create_obj(manager, top, manager->container, NULL);
+}
+
+size_t lv_fragment_manager_size(lv_fragment_manager_t * manager)
```
kisvegabor:
```suggestion
size_t lv_fragment_manager_get_size(lv_fragment_manager_t * manager)
```

src/extra/others/fragment/lv_fragment.c
```diff
@@ -0,0 +1,392 @@
...
+{
+    size_t size = 0;
+    for(manager_stack_t * cur = manager->top; cur; cur = cur->prev) {
```
kisvegabor: I fond that in XC16 compiler (Microchip PIC) this declaration doesn't work. 
```suggestion
    manager_stack_t * cur;
    for(cur = manager->top; cur; cur = cur->prev) {
```

src/extra/others/fragment/lv_fragment.c
```diff
@@ -0,0 +1,392 @@
...
+    manager_stack_t * top = manager->top;
+    if(!top) return false;
+    lv_fragment_t * instance = top->instance;
+    if(!instance || !top->cls->event_cb) return false;
+    return top->cls->event_cb(instance, which, data1, data2);
+}
+
+#if LV_USE_MSGBOX
+
+void lv_fragment_manager_show(lv_fragment_manager_t * manager, const lv_fragment_class_t * cls, void * args)
```
kisvegabor: What is purpose of this?

src/extra/others/fragment/lv_fragment.c
```diff
@@ -0,0 +1,392 @@
...
+    item_destroy_obj(manager, top);
+    item_destroy_fragment(top);
+    lv_mem_free(top);
+    if(!dialog && prev) {
+        item_create_obj(manager, prev, manager->container, NULL);
+    }
+    manager->top = prev;
+}
+
+bool lv_fragment_manager_dispatch_event(lv_fragment_manager_t * manager, int which, void * data1, void * data2)
```
kisvegabor: Why `data1` and `data2`? Why not only 1 data argument?

src/extra/others/fragment/lv_fragment.c
```diff
@@ -0,0 +1,392 @@
...
+    return top->instance;
+}
+
+lv_fragment_t * lv_fragment_manager_get_parent(lv_fragment_manager_t * manager)
+{
+    LV_ASSERT(manager);
+    return manager->parent;
+}
+
+lv_fragment_t * lv_fragment_class_create_unmanaged(const lv_fragment_class_t * cls,
```
kisvegabor: What about this slight change in the API:

Instead of:
```c
    lv_fragment_manager_t *manager = lv_fragment_manager_create(root, NULL);
    lv_fragment_manager_replace(manager, &sample_cls, "Fragment");
```

This:
```c
    lv_fragment_manager_t *manager = lv_fragment_manager_create(root, NULL);
    lv_fragment_t * f1 = lv_fragment_create(manager, &sample_cls, "Fragment");
    lv_fragment_manager_replace(manager, f1);
```
For unmanaged:
```c 
  lv_fragment_t * f1 = lv_fragment_create(NULL, &sample_cls, "Fragment");
```

It's a little bit more verbose but cleaner IMO.

src/extra/others/fragment/lv_fragment.c
```diff
@@ -0,0 +1,392 @@
...
+    if(cls->obj_deleted_cb) {
+        cls->obj_deleted_cb(fragment, fragment->obj);
+    }
+    if(cls->destructor_cb) {
+        cls->destructor_cb(fragment);
+    }
+    lv_mem_free(fragment);
+}
+
+void lv_fragment_pop(lv_fragment_t * fragment)
```
kisvegabor: Do we really need this? 
This seems more flexible to me:
```c
lv_fragment_manager_pop(lv_fragment_get_manager(fragment));
```

src/extra/others/fragment/lv_fragment.c
```diff
@@ -0,0 +1,392 @@
...
+    LV_ASSERT(manager->top->instance == fragment);
+    lv_fragment_manager_pop(manager);
+}
+
+void lv_fragment_recreate_obj(lv_fragment_t * fragment)
+{
+    LV_ASSERT(fragment);
+    // Disable CB first
+    lv_obj_remove_event_cb(fragment->obj, obj_cb_delete);
+    fragment_destroy_obj(fragment);
```
kisvegabor: Shouldn't it be public too once `recreate` is public?

src/extra/others/fragment/lv_fragment.h
```diff
@@ -0,0 +1,232 @@
...
+    lv_fragment_manager_t * manager;
+    /**
+     * lv_obj returned by create_obj_cb
+     */
+    lv_obj_t * obj;
+
+    /**
+     * Private internal states
+     */
+    manager_stack_t * priv_item;
```
kisvegabor: It seems `manager_stack_t` and the fragment is tightly coupled. What is the point of `manager_stack_t` as pointer instead of having its fields directly in `lv_fragment_t`?  I'd help to unify many static functions: one works on `item` and other on `fragment`.

src/extra/others/fragment/lv_fragment.h
```diff
@@ -0,0 +1,232 @@
...
+     */
+    const lv_fragment_class_t * cls;
+    /**
+     * Manager this fragment instance belongs to
+     */
+    lv_fragment_manager_t * manager;
+    /**
+     * lv_obj returned by create_obj_cb
+     */
+    lv_obj_t * obj;
```
kisvegabor, line 54: Is it for the user in case he wants to store an obj?

kisvegabor (commented):
Thank you for the updates.

I've added some comments and questions. 
Sorry for nitpicking that much. I find this update really important as it can be a proposed way how people build UIs with LVGL, so I'd like to understand all the details. :slightly_smiling_face:

src/extra/others/fragment/lv_fragment.h
mariotaku: There are 2 cases.
1: User returns obj, and it will be managed along with the fragment
2: User doesn't return obj, means they will have to manage the views they added manually

For case 2 this is null.

src/extra/others/fragment/lv_fragment.c
mariotaku: I prefer to let users do by themselves. Because we can still keep fragment instance after view destroyed, so considering we may support keeping nested manager instance, we should leave this to the user.

src/extra/others/fragment/lv_fragment.h
mariotaku: There are too many internal states in this manager_stack_t. If we expose them all, I'm afraid users will be confused when they see too many fields at once.

src/extra/others/fragment/lv_fragment.c
mariotaku: This makes it more like Android Fragment which really makes sense. But in this way I think I will have to separate fragment creation and object attachment.

src/extra/others/fragment/lv_fragment.h
kisvegabor: In other parts of LVGL we provide API function to access struct members, so if something needs to be hidden simply there is no API function for it.

But if you'd like to keep it separate I think it'd be better to add `manager_stack_t` as a field of `lv_fragment_t` (e.g. `lv_fragment_priv_t priv_data`) and not a pointer in order to really manage them together.

src/extra/others/fragment/lv_fragment.c
kisvegabor: `object attachment` means creating object?

src/extra/others/fragment/lv_fragment.c
mariotaku: It's `lv_fragment_del_obj` now.

src/extra/others/fragment/lv_fragment.h
mariotaku: This internal state only available when fragment is managed, so I'll stay with pointer. But yes, the struct can be made non-opaque. It's just not meant to be modify by user, but fragment manager only.

src/extra/others/fragment/lv_fragment.c
mariotaku: Done.

src/extra/others/fragment/lv_fragment.c
mariotaku: It means creating `lv_obj_t`s upon containers. It's separated now.

mariotaku:
New todos:

- [ ] Update examples to include all public functions
- [x] Update docs
- [ ] Stress test / memleak test
- [x] Proper design of fragment states. `lv_fragment_managed_states_t * managed` maybe?
- [x] Review functions not needed

src/extra/others/fragment/lv_fragment.c
mariotaku: It's for showing msgbox. When a msgbox is added, it should not remove existing view, so it has a separate stack.

src/extra/others/fragment/lv_fragment.c
mariotaku: It's copied from SDL. I'll update to use one userdata.

mariotaku:
Sorry that I have made some big changes. Now the API should be stabilized, could you take a look when you have time? I'll do finish TODOs listed above later.

docs/others/fragment.md
```diff
@@ -0,0 +1,72 @@
...
+.. include:: /header.rst 
+:github_url: |github_link_base|/others/fragment.md
+```
+
+# Fragment
+
+This is something like [Fragment on Android](https://developer.android.com/guide/components/fragments)
+or [UiViewController on iOS](https://developer.apple.com/documentation/uikit/uiviewcontroller) - to help developers
+better handles code structure, and object lifecycle.
+
```
kisvegabor, line 17: Some more info wold be great here: e.g. 
- what is a fragment
- it consists of a fragment manager that can create/delete fragments attached to it...

src/extra/others/fragment/lv_fragment_manager.c
```diff
@@ -0,0 +1,346 @@
...
+                                 lv_obj_t * const * container)
+{
+    lv_fragment_t * top = lv_fragment_manager_get_top(manager);
+    if(top != NULL) {
+        lv_fragment_manager_remove(manager, top);
+    }
+    lv_fragment_manager_push(manager, fragment, container);
+}
+
+void lv_fragment_manager_recreate_obj(lv_fragment_manager_t * manager, lv_fragment_t * fragment)
```
kisvegabor: Isn't it redundant with `lv_fragment_recreate_obj()`?

src/extra/others/fragment/lv_fragment.h
```diff
@@ -0,0 +1,349 @@
...
+ * @param args Arguments assigned by fragment manager
+ */
+void lv_fragment_manager_show(lv_fragment_manager_t * manager, lv_fragment_t * fragment);
+
+/**
+ * Send event to top-most fragment
+ * @param manager Fragment manager instance
+ * @param which User-defined ID of event
+ * @param data1 User-defined data
+ * @param data2 User-defined data
```
kisvegabor: Only has `user_data`

kisvegabor (commented):
Thanks, look much cleaner to me!

The popup part is not clear to me.

> It's for showing msgbox. When a msgbox is added, it should not remove existing view, so it has a separate stack.

Is it like "branching" a new manager that runs in popup or so?

src/extra/others/fragment/lv_fragment.h
mariotaku: Fixed.

src/extra/others/fragment/lv_fragment_manager.c
mariotaku: `lv_fragment_recreate_obj` is a shorthand to `lv_fragment_manager_recreate_obj`. How about to use inline function instead?

mariotaku: \* seems I shouldn't. I think it's OK to leave this as is.

mariotaku:
@kisvegabor Yes. It's like branching a manager. Its usage is quite limited so I removed it. User can use unmanaged fragment for such use case instead.

docs/others/fragment.md
mariotaku: I will copy many sentences from Android Developers 😆  the concept is identical. https://developer.android.com/guide/fragments

kisvegabor: No problem with that. To goal is to have a clear easy to digest description about what this feature is.

kisvegabor:
> @kisvegabor Yes. It's like branching a manager. Its usage is quite limited so I removed it. User can use unmanaged fragment for such use case instead.

Okay, simpler API is better :slightly_smiling_face:

mariotaku:
@kisvegabor Do you think the API is good now? I'll add some use cases later.

mariotaku:
BTW test is failing since something in micropython might broke...

docs/others/fragment.md
```diff
@@ -0,0 +1,78 @@
...
+# Fragment
+
+Fragment is a concept copied from [Android](https://developer.android.com/guide/fragments).
+
+It represents a reusable portion of your app's UI. A fragment defines and manages its own layout, has its own lifecycle,
+and can handle its own events. Like Android's Fragment that must be hosted by an activity or another fragment, Fragment
+in LVGL needs to be hosted by an object, or another fragment. The fragment’s view hierarchy becomes part of, or attaches
+to, the host’s view hierarchy.
+
```
kisvegabor, line 14: I think the manager should be mentioned here too.

src/extra/others/fragment/lv_fragment.h
```diff
@@ -0,0 +1,337 @@
...
+    /**
+     * Managed fragment states. If not null, then this fragment is managed.
+     *
+     * @warning Don't modify values inside this struct!
+     */
+    lv_fragment_managed_states_t * managed;
+    /**
+     * Child fragment manager
+     */
+    lv_fragment_manager_t * child_manager;
```
kisvegabor, line 50: Why is it called  `child_manager`? Why not simply `manager`?

src/extra/others/fragment/lv_fragment.h
```diff
@@ -0,0 +1,337 @@
...
+     * @return Created object, NULL if multiple objects has been created
+     */
+    lv_obj_t * (*create_obj_cb)(struct lv_fragment_t * self, lv_obj_t * container);
+
+    /**
+     *
+     * @param self Fragment instance
+     * @param obj lv_obj returned by create_obj_cb
+     */
+    void (*obj_created_cb)(struct lv_fragment_t * self, lv_obj_t * obj);
```
kisvegabor: Can we replace these callbacks with events?

examples/others/fragment/lv_example_fragment_1.c
```diff
@@ -0,0 +1,43 @@
+/**
```
kisvegabor, line 1: `lv_example_fragment_1` results in empty screen for me.

examples/others/fragment/lv_example_fragment_2.c
```diff
@@ -0,0 +1,104 @@
+/**
```
kisvegabor, line 1: It crashes if I press "Push" and "Pop" after that. (I.e. when the last fragment is popped)

src/extra/others/fragment/lv_fragment.h
```diff
@@ -0,0 +1,337 @@
...
+ * @return Fragment manager instance
+ */
+lv_fragment_manager_t * lv_fragment_manager_create(lv_fragment_t * parent);
+
+/**
+ * Destroy fragment manager instance
+ * @param manager Fragment manager instance
+ */
+void lv_fragment_manager_del(lv_fragment_manager_t * manager);
+
```
kisvegabor, line 179: Please add the missing API comments.

kisvegabor (commented):
Thanks for the updates. I added some comments.

src/extra/others/fragment/lv_fragment.h
mariotaku: It's for managing child fragments. If user needs to create fragments under this fragment, they must use `child_manager`.

src/extra/others/fragment/lv_fragment.h
mariotaku: `lv_fragment` is not an object. I don't think it's applicable to current event system lvgl has.

kisvegabor: I meant the fragment's event system.

examples/others/fragment/lv_example_fragment_1.c
mariotaku: Done, got it fixed.

examples/others/fragment/lv_example_fragment_2.c
mariotaku: Sorry for this issue, it's fixed.

src/extra/others/fragment/lv_fragment.h
mariotaku: I think it should be separated. This callback is related to lifecycle, while event callback is related to user-defined events.

mariotaku:
Now that the API has been simplified, and issues has been fixed, I think it can be merged if no other issues found.

kisvegabor:
Thank you! In the 2nd example I noticed that if I pop the last fragment too and push again the state (counter value) is lost. Is this the intended behavior?

  
**Events**
> I think it should be separated. This callback is related to lifecycle, while event callback is related to user-defined events.

So I mean these callbacks:
```c
    void (*attached_cb)(struct lv_fragment_t * self);
    void (*detached_cb)(struct lv_fragment_t * self);
    lv_obj_t * (*create_obj_cb)(struct lv_fragment_t * self, lv_obj_t * container);
    void (*obj_created_cb)(struct lv_fragment_t * self, lv_obj_t * obj);
    void (*obj_will_delete_cb)(struct lv_fragment_t * self, lv_obj_t * obj);
    void (*obj_deleted_cb)(struct lv_fragment_t * self, lv_obj_t * obj);
```

IMO an event can be anything: we can send ''system events" but the the user can send custom "user events" too. These are just the things happening with fragment.
I see the following advantages of converting these callbacks into events:
- using a generic interface (the event) is easier to digest and learn
- `lv_fragment_class_t` get's smaller, cleaner and easier to read
- more maintainable to extend an event enum than a callback list. 

**API suggestions**
Now there is
```c
lv_fragment_manager_recreate_obj
```
and
```c
lv_fragment_recreate_obj
```

When is it useful to use `lv_fragment_manager_recreate_obj` instead of `lv_fragment_recreate_obj`?  It seems to me `lv_fragment_recreate_obj` is more flexible as can do the same with less arguments. Same for `del_obj` and `create_obj`.

`lv_fragment_remove_self` rename to `lv_fragment_detach`?

`lv_fragment_manager_dispatch_event` -> `lv_fragment_manager_send_event`? (Just for consistency with `lv_obj_set_event`)

**Question**
`lv_fragment_get_parent`: So a fragments can be attached to managers (in this case there is navigation with push/pop) but to other fragments too. (unmanaged?). How to attach a fragment to an other fragment? Or how to create a manager on an other fragment? Is there API for it?

mariotaku:
Thanks for your review.

> if I pop the last fragment too and push again the state (counter value) is lost. Is this the intended behavior?

Yes. If a fragment was popped, then it's completely destroyed.

### Regarding callback functions in fragment class

I would disagree on using one single events callback for callbacks above. I think those functions are very important, and should enforce function signature in language level just like `lv_draw_backend_t`.

---

For other reviews I'll carefully review and make relevant changes.

mariotaku:
* `lv_fragment_manager_recreate_obj` has been removed and will use `lv_fragment_recreate_obj` instead.
* `lv_fragment_remove_self` has been removed replaced by `lv_fragment_del`.
* `lv_fragment_manager_dispatch_event` was renamed to `lv_fragment_manager_send_event`
---

To attach `Fragment A` to `Fragment B`, use `child_manager` inside `Fragment B`. One fragment can either be added to navigation stack, or be added to object hierarchy without adding to the stack (so it's just a wrapper to one object).

mariotaku:
@kisvegabor What do you about last change?

src/extra/others/fragment/lv_fragment.h
```diff
@@ -0,0 +1,338 @@
...
+/**********************
+ *      TYPEDEFS
+ **********************/
+
+typedef struct _lv_fragment_manager_t lv_fragment_manager_t;
+
+typedef struct lv_fragment_t lv_fragment_t;
+typedef struct lv_fragment_class_t lv_fragment_class_t;
+typedef struct lv_fragment_managed_states_t lv_fragment_managed_states_t;
```
kisvegabor: For consistency please refactor `struct lv_fragment_...` to `struct _lv_fragment_...`

src/extra/others/fragment/lv_fragment.h
```diff
@@ -0,0 +1,338 @@
...
+ * Remove the top-most fragment for stack
+ * @param manager Fragment manager instance
+ * @return true if there is fragment to pop
+ */
+bool lv_fragment_manager_pop(lv_fragment_manager_t * manager);
+
+/**
+ * Replace fragment. Old item in the stack will be removed.
+ * @param manager Fragment manager instance
+ * @param fragment Fragment instance
```
kisvegabor, line 225: `@param container` is missing.

kisvegabor (commented):
Sorry for the late reply. 

I've added only 2 minor comments.

In [this test](https://github.com/lvgl/lvgl/blob/master/tests/CMakeLists.txt#L191) please enable this widget to make the CI build it.

examples/others/fragment/lv_example_fragment.h
```diff
@@ -0,0 +1,37 @@
+/**
```
kisvegabor, line 1: The CI fails because this file is not included in `lv_example_others.h`

mariotaku: I think it is properly included.

kisvegabor:
I've pushed a fix for the failing CI. 
All looks good to me so I merge it. We still have plenty of time until next release so we can add fixes later too.

Thank you very much!

amirgon:
I'm not sure if there's any value to Micropython bindings here, but it's worth noting that this API does not follow the conventions so this can't at current state be used with Micropython.

kisvegabor:
What are the problematic parts? At first looks it seems similar to any widget's API.

amirgon:
> What are the problematic parts? At first looks it seems similar to any widget's API.

`user_data` is missing in `lv_fragment_class_t`, but that seems easy to fix.  
The more fundemental issue is that in Micropython we cannot extend fragment from `lv_fragment_t` base, at least not the way this is done in the C examples. Casting `void* args` is another problem.

But really, I'm not sure about the value of fragments with Micropython.  
Instead of fragments it seems much easier to simply define Python classes that represent reusable UI parts. That would be cleaner and much more "Pythonic".  
Could you explain what are the benefits of fragments over Python classes? I can see some functions for "pushing" and "poping" fragments from "fragment manager" but I don't really understand how this is useful.

Was there any attempt to create Micopython examples next to the C examples?  
Trying to create Micropython examples is the best way to find binding issues and where the API is incompatilble.

mariotaku:
How about to ignore python binding for this component and create a pure object orientated version with python? I think it should be much easier to do so.

PR #2941, opened 2021-12-23
SDL draw updates

mariotaku:

SDL_RenderSetClipRect will slow render down especially on low end devices. This is for a little bit more performance improvement.
Also, it adds polygon drawing logic

### Checkpoints


src/draw/lv_draw_mask.c
```diff
@@ -1218,6 +1250,79 @@ LV_ATTRIBUTE_FAST_MEM static lv_draw_mask_res_t lv_draw_mask_map(lv_opa_t * mask
     return LV_DRAW_MASK_RES_CHANGED;
 }
 
+LV_ATTRIBUTE_FAST_MEM static lv_draw_mask_res_t lv_draw_mask_polygon(lv_opa_t * mask_buf, lv_coord_t abs_x,
```
kisvegabor, line 1253: Does it support concave polygons like this?
image

kisvegabor (commented):
Thanks!

src/draw/lv_draw_mask.c
mariotaku: It supports convex polygon only. Should LVGL support concave polygons?

kisvegabor: It should, but doesn't support at this moment :laughing: 

Supporting convex only is fine, as it's in pair with current features.

kisvegabor:
Please fix the code formatting issue, and I'll merge it! :slightly_smiling_face:

mariotaku:
done :)

src/draw/lv_draw_mask.c
mariotaku: I'll support this in later patches. So users can use mask directly anywhere (without point limit).

kisvegabor:
Merging thanks.

PR #2955, opened 2022-01-03
kconfig: fixes for Zephyr integration

brgl:
Here are two Kconfig fixes for bugs I noticed when integrating LVGL's Kconfig with Zephyr's.

### DIFF
--- a/Kconfig
+++ b/Kconfig
@@ -839,22 +839,19 @@ menu "LVGL configuration"
 
     menu "3rd Party Libraries"
         config LV_USE_FS_STDIO
-            int "File system on top of stdio API"
-            default 0
+            bool "File system on top of stdio API"
         config LV_FS_STDIO_PATH
             string "Set the working directory"
             depends on LV_USE_FS_STDIO
 
         config LV_USE_FS_POSIX
-            int "File system on top of posix API"
-            default 0
+            bool "File system on top of posix API"
         config LV_FS_POSIX_PATH
             string "Set the working directory"
             depends on LV_USE_FS_POSIX
 
         config LV_USE_FS_WIN32
-            int "File system on top of Win32 API"
-            default 0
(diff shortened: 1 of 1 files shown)


Kconfig
```diff
@@ -839,22 +839,19 @@ menu "LVGL configuration"
 
     menu "3rd Party Libraries"
         config LV_USE_FS_STDIO
-            int "File system on top of stdio API"
-            default 0
+            bool "File system on top of stdio API"
```
kisvegabor, line 842: Here you need to set an ASCII character (i.e. a driver letter) on which LVGL will reach this file system.
Finally, a file path will look like `"S/my/folder/file.txt"`.

brgl: Right. Is this the same for the other two too? Can we just add a sub-option for the filesystem letter?

kisvegabor: > Is this the same for the other two too

Yes, same for the others too. 

> Can we just add a sub-option for the filesystem letter?

I'm ok with that. User can choose from ASCII upper case letters and `\0` to disable.

rzr: or set default prefix set to "" for path starting with separator "/"

kisvegabor: @rzr It's only a character, not a string :slightly_frowning_face: 
E.g. `'A'`

rzr: yes but my point that string would be more generic like "A:" (or "df0:" for amiga os ;) )

kisvegabor: Yes, it'd be really better but would be an API break. We can do it only in v9.

kisvegabor:
@brgl 
I've fixed the anonymous choice in 71c739cc2 and `LV_FS_` related parts in abba1c351.

So I close this PR. Feel free to open a new PR is you find any other issues with Kconfig.

PR #2968, opened 2022-01-06
feat(span):lv_spangroup_get_expand_width() adds a parameter

guoweilkd:

feat(span):lv_spangroup_get_expand_width() adds a parameter

### Checkpoints

### DIFF
--- a/src/extra/widgets/span/lv_span.c
+++ b/src/extra/widgets/span/lv_span.c
@@ -86,11 +86,6 @@ const lv_obj_class_t lv_spangroup_class  = {
  *   GLOBAL FUNCTIONS
  **********************/
 
-/**
- * Create a spangroup object
- * @param par pointer to an object, it will be the parent of the new spangroup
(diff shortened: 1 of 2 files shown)


src/extra/widgets/span/lv_span.c
```diff
@@ -434,10 +434,13 @@ lv_coord_t lv_spangroup_get_max_line_h(lv_obj_t * obj)
 }
 
 /**
- * get the width when all span of spangroup on a line. not included spangroup pad, border width.
+ * get the text content width when all span of spangroup on a line.
```
kisvegabor: Now both comments say:
> get the text content width when all span of spangroup on a line.


BTW, in other files we added API comments only in the header files to avoid duplication. Please follow this pattern with the span widget too?

kisvegabor:
Thanks! 
Although technically it's an API breaking change but I think we can be flexible here as probably this function is very rarely used.

PR #2970, opened 2022-01-06
fix(snapshot): fix memory leak

xaowang96:
Free draw_ctx after snapshot is over.

### Checkpoints


embeddedt:
Thanks. I added a test to catch any similar bugs in the future.

PR #2973, opened 2022-01-06
Add guards to LV_SYMBOL_*

tyler-gilbert:
### Add Guards to LV_SYMBOL_* definitions

This will allow users to pre-define any `LV_SYMBOL_` value. This will allow users to use any iconset with LVGL rather than just FontAwesome. It also allows users to customize the keyboard icons without using custom code to change the mappings.


kisvegabor:
Looks good, thank you!

PR #2979, opened 2022-01-08
Changes adopted by the PR 2401

eismeraldo:
### feat(fs) add caching

This PR is based on the [PR 2401] (https://github.com/lvgl/lvgl/pull/2401) from PGNetHun. At the moment this PR is not functional. The further procedure must be discussed with those responsible for LVGL.

### Checkpoints


kisvegabor:
Hi,

Thanks for opening this PR. I've made some fixes and clean up and now it works well for me. 

Note the changes in the [configuration of FS drivers](https://github.com/lvgl/lvgl/pull/2979/files#diff-0be6707528460cdc1dd90dad35b2597054e790c6b811c3ed5f9d3ce41c105ed0).

kisvegabor:
@brgl The changes in the PR affect yours about Kconfig. #2955
Finally I took your approach here about using `LV_USE_FS_...` explicitly. 
Please review it and confirm if it works this way for Kconfig too.

brgl:
> @brgl The changes in the PR affect yours about Kconfig. #2955 Finally I took your approach here about using `LV_USE_FS_...` explicitly. Please review it and confirm if it works this way for Kconfig too.

@kisvegabor Thanks for the heads-up. I'm focusing on the zephyr fork currently, once the lvgl v8 support is upstream, I will try to get it up to date and send any downstream changes your way.

kisvegabor:
Great, thanks!

kisvegabor:
@GorgonMeducer
Does [these changes](https://github.com/lvgl/lvgl/pull/2979/files#diff-0be6707528460cdc1dd90dad35b2597054e790c6b811c3ed5f9d3ce41c105ed0) affect cmsis-pack?

GorgonMeducer:
@kisvegabor YES, I need to re-run the gen_pach.sh and that's all.

kisvegabor:
Great, I can do that too :slightly_smiling_face: Anyway I'll ping you before we merge it to confirm if it look good.

@eismeraldo did you have a chance to try it out?

eismeraldo:
@kisvegabor  I could test if the fixed PR works. Under which URL can I download the fixed source?

eismeraldo:
I noticed that I was up to date. Everything compiles without errors. The behavior has not changed, 200 bytes are still read instead of 512.
I found out what the reason is: It uses twice the width of the 100x100 BMP, i.e. 200 bytes. The block should actually be independent of the size of the BMP, but this is probably not that easy to solve.

kisvegabor:
> I could test if the fixed PR works. Under which URL can I download the fixed source?

You can `git pull` your `master` branch. `git log -3` should show the last commits are from me. 

How does the config look like in your lv_conf.h now?

eismeraldo:
I have activated the various options in the File System section, but only the two options LV_USE_FS_STDIO and LV_USE_FS_POSIX compile without errors. But as already mentioned, they do not change the behavior.
[lv_conf.txt](https://github.com/lvgl/lvgl/files/7861498/lv_conf.txt)

eismeraldo:
I took some pictures of the stack trace.
In Image1.png you can see that the width of the image is being read.
In Image2.png you can see that the color depth 16 divided by 8 bits results in 2 bytes multiplied by the image width of 100 bytes results in 200 bytes.
In Image3.png you can see that btr is now exactly 200 and that the cache_size is not set.
And in Image4.png the user read function is now called with the btr of 200.
Image 1
Image 2
Image 3
Image 4

I also attached the log that documents the open, close, read, seek statements
[platformio-device-monitor-220113-095525.log](https://github.com/lvgl/lvgl/files/7861854/platformio-device-monitor-220113-095525.log)

kisvegabor:
> In Image3.png you can see that btr is now exactly 200 and that the cache_size is not set.

Why isn't it set? The cache size should be set [here](https://github.com/lvgl/lvgl/pull/2979/files#diff-4c08eb6a734735fadedb3f775d4528c3b5d3609fab1aad9aa6c84948eba79b19R75). Could you debug what happens with the set value?

eismeraldo:
It's maddening, yesterday I tried to debug the ESP32 all day, the craziest error messages were displayed. Today I tried again and after a few hours the PC died. I'm typing here from my wife's PC, which is just good enough to run LibreOffice and accounting. I will order a new one, but that will take a few days.

eismeraldo:
I was allowed to test the PR on my friend's PC. I have now discovered the following:
1) The fs_drv.cache_size is set correctly in lv_fs_stdio.c.
2) However, in the lv_fs_read function in lv_fs.c, file_p→drv→cache_size is set to 0.

I found this out with logging. I tried to print out the addresses of the two objects and they are different, maybe I'm doing something wrong. The logs are:
lv_fs_stdio.c line 69
     LV_LOG_USER("==== cache_size: %i, address %p", fs_drv.cache_size, &fs_drv.cache_size);
and in
lv_fs.c line 211
     LV_LOG_USER("----- cache_size: %i, address %p", file_p->drv->cache_size, &file_p->drv->cache_size);

kisvegabor:
Hi, 

I made some clean up and checked it again but it still works well for me. 

I've searched for `cache` in lvgl to were we assign a value to it, but its only the initialization of the driver.
Can you track where `cache_size` loses its value?

kisvegabor:
We are planning to release v8.2 soon and I'd like to make it part of the release. 

It seems working for me, but if you find the root your issue, please comment here.

src/extra/libs/fsdrv/lv_fs_fatfs.c
```diff
@@ -62,7 +66,9 @@ void lv_fs_fatfs_init(void)
     lv_fs_drv_init(&fs_drv);
 
     /*Set up fields...*/
-    fs_drv.letter = LV_USE_FS_FATFS;
+    fs_drv.letter = LV_FS_FATFS_LETTER;
+    fs_drv.cache_size = LV_FS_FSTFS_CACHE_SIZE;
```
papadkostas, line 70: There's a typo here.

fs_drv.cache_size = **LV_FS_FSTFS_CACHE_SIZE**;

lv_conf_template.h
```diff
@@ -559,18 +559,38 @@
...
+    #define LV_FS_WIN32_PATH ""         /*Set the working directory. File/directory path ill be appended to it.*/
+    #define LV_FS_WIN32_CACHE_SIZE 0    /*>0 to cache this number of bytes in lv_fs_read()*/
+#endif
 
-#define LV_USE_FS_FATFS '\0'        /*Uses f_open, f_read, etc*/
+/*API for FATFS (needs to be added separately). Uses f_open, f_read, etc*/
+#define LV_USE_FS_FATFS  0
+#if LV_USE_FS_FATFS
+    #define LV_FS_FATFS_LETTER '\0'     /*Set an upper cased letter on which the drive will accessible (e.g. 'A')*/
+    #define LV_FS_FATSF_CACHE_SIZE 0    /*>0 to cache this number of bytes in lv_fs_read()*/
```
papadkostas, line 592: Here as well.
    #define **LV_FS_FATSF_CACHE_SIZE** 0    /*>0 to cache this number of bytes in lv_fs_read()*/

src/extra/libs/fsdrv/lv_fs_fatfs.c
kisvegabor: Thank you, both are fixed here: 74091c42f

eismeraldo:
My new PC is now set up and I was able to debug. The first driver is created based on lv_conf.h and the second is mine based on   https://docs.lvergleich.io/master/overview/file-system.html . I have now deactivated the driver in lv_conf.h and now everything runs perfectly.

kisvegabor:
And BMPs are loaded faster too? :slightly_smiling_face:

eismeraldo:
I haven't measured the running time, but it feels a little better. Now 40 read and seek operations are performed instead of 100. Are the seek and open close operations really necessary? If I read in the same BMP by hand, it gives Open → read until end → close

kisvegabor:
Some of them yes as LVGL needs to collect some info about the image on various places. 
Try setting `LV_IMG_CACHE_DEF_SIZE 1`. I think it will reduce the open count a little.

eismeraldo:
I took measurements and found something surprising.
LV_IMG_CACHE_DEF_SIZE 0 throughput time = 1,196
LV_IMG_CACHE_DEF_SIZE 1 throughput time = 5.11
In this variant, two files are somehow read in overlapping. But the display works perfectly. Thus, the time measured between the first open and the last close means nothing.

Now I made measurements with different cache_size:
cache_size = 0 throughput time = 1,029
cache_size = 512 throughput time = 1,196
cache_size = 2048 throughput time = 2.07

What happened here? I noticed that after every read of the size of the cache_size, a subsequent seek is called, which moves the position by exactly 200 bytes instead of the cache_size. That's why the cache_size is now counterproductive.

Now I have implemented a dummy function that does not seek to the SD card.
Without seek, cache_size = 2048 throughput time = 0.721
The throughput time is now very good, unfortunately the images are no longer displayed correctly.
In the case of cache_size > 0, the seek would probably have to be handled by the cache functionality.

kisvegabor:
`lv_fs_seek` should handle seek in there is cache.

At this moment I have no time to debug it, but if you find something please send a PR or comment here.

eismeraldo:
To me it looks like the BMP is read from back to front, but the data is stored in the cash from front to back.
In the first pass, image 1, position(p) is set to 19938.
Image01

As a result, the driver's seek is executed (Figure 2) because file_p→cache→end = 2048 and file_p→cache→file_position = 19938.
Image02

The next time the position is 19738, then 19538 etc. See LOG
[log5_seek_problem.txt](https://github.com/lvgl/lvgl/files/8025983/log5_seek_problem.txt)

kisvegabor:
> To me it looks like the BMP is read from back to front, but the data is stored in the cash from front to back.


Ahh, it really results in a lot of seek operations and renders cache useless. Even worse with a larger cache LVGL reads more data for nothing...

What if you covert the BMP image to a binary [here](https://lvgl.io/tools/imageconverter)?
The result file is just like a BMP image but not mirrored. So it should be much faster.

eismeraldo:
The conversion works perfectly and the images are rendered in a fraction of the variant via SD card. My problem was a memory problem, I'm not moving on a PC but on an ESP32 with limited memory. For this reason, I tried to import the background images from the SD card, which represent different schemes of a heating system, and render them. In the meantime, I've decided to forget the graphical schemes and limit myself to the on-board resources of LVGL.
So thanks again for the help and the great LVGL.

kisvegabor:
Anyway, it was a meaningful lesson. I learnt a lot from it. Thank you for all the tests and debugging. :+1:

PR #3000, opened 2022-01-13
fix(font):may crash if f->fallback == NULL

guoweilkd:

dsc_out->resolved_font should point to the previous font if f->fallback == NULL

### Checkpoints

### DIFF
--- a/src/draw/sw/lv_draw_sw_letter.c
+++ b/src/draw/sw/lv_draw_sw_letter.c
@@ -97,13 +97,28 @@ void lv_draw_sw_letter(lv_draw_ctx_t * draw_ctx, const lv_draw_label_dsc_t * dsc
 {
     lv_font_glyph_dsc_t g;
     bool g_ret = lv_font_get_glyph_dsc(dsc->font, &g, letter, '\0');
-    if(g_ret == false)  {
+    if(g_ret == false) {
         /*Add warning if the dsc is not found
          *but do not print warning for non printable ASCII chars (e.g. '\n')*/
         if(letter >= 0x20 &&
            letter != 0xf8ff && /*LV_SYMBOL_DUMMY*/
            letter != 0x200c) { /*ZERO WIDTH NON-JOINER*/
-            LV_LOG_WARN("lv_draw_letter: glyph dsc. not found for U+%X", (unsigned int)letter);
+            LV_LOG_WARN("lv_draw_letter: glyph dsc. not found for U+%" PRIX32, letter);
+
+            /* draw placeholder */
+            lv_area_t glyph_coords;
+            lv_draw_rect_dsc_t glyph_dsc;
+            lv_coord_t begin_x = pos_p->x + g.ofs_x;
+            lv_coord_t begin_y = pos_p->y + g.ofs_y;
+            lv_area_set(&glyph_coords, begin_x, begin_y, begin_x + g.box_w, begin_y + g.box_h);
+            lv_draw_rect_dsc_init(&glyph_dsc);
+            glyph_dsc.bg_opa = LV_OPA_MIN;
+            glyph_dsc.outline_opa = LV_OPA_MIN;
+            glyph_dsc.shadow_opa = LV_OPA_MIN;
+            glyph_dsc.bg_img_opa = LV_OPA_MIN;
+            glyph_dsc.border_color = dsc->color;
+            glyph_dsc.border_width = 1;
+            draw_ctx->draw_rect(draw_ctx, &glyph_dsc, &glyph_coords);
         }
         return;
     }
--- a/src/font/lv_font.c
+++ b/src/font/lv_font.c
@@ -66,18 +66,41 @@ bool lv_font_get_glyph_dsc(const lv_font_t * font_p, lv_font_glyph_dsc_t * dsc_o
 {
     LV_ASSERT_NULL(font_p);
     LV_ASSERT_NULL(dsc_out);
-    dsc_out->resolved_font = NULL;
+    const lv_font_t * placeholder_font = NULL;
     const lv_font_t * f = font_p;
-    bool found = false;
+
+    dsc_out->resolved_font = NULL;
+
     while(f) {
-        found = f->get_glyph_dsc(f, dsc_out, letter, letter_next);
-        if(found && !dsc_out->is_placeholder) {
-            dsc_out->resolved_font = f;
-            break;
+        bool found = f->get_glyph_dsc(f, dsc_out, letter, letter_next);
+        if(found) {
+            if(!dsc_out->is_placeholder) {
+                dsc_out->resolved_font = f;
+                return true;
+            }
+            else if(placeholder_font == NULL) {
+                placeholder_font = f;
+            }
         }
         f = f->fallback;
     }
-    return found;
+
+    if(placeholder_font != NULL) {
+        placeholder_font->get_glyph_dsc(placeholder_font, dsc_out, letter, letter_next);
+        dsc_out->resolved_font = placeholder_font;
+        return true;
+    }
+
+    dsc_out->resolved_font = NULL;
+    dsc_out->box_w = font_p->line_height / 2;
+    dsc_out->adv_w = dsc_out->box_w + 2;
+    dsc_out->box_h = font_p->line_height;
+    dsc_out->ofs_x = 0;
+    dsc_out->ofs_y = 0;
+    dsc_out->bpp   = 1;
+    dsc_out->is_placeholder = true;
+
+    return false;
 }
 
 /**
@@ -91,10 +114,8 @@ uint16_t lv_font_get_glyph_width(const lv_font_t * font, uint32_t letter, uint32
 {
     LV_ASSERT_NULL(font);
     lv_font_glyph_dsc_t g;
-    bool ret;
-    ret = lv_font_get_glyph_dsc(font, &g, letter, letter_next);
-    if(ret) return g.adv_w;
-    else return 0;
+    lv_font_get_glyph_dsc(font, &g, letter, letter_next);
+    return g.adv_w;
 }
 
 /**********************
--- a/src/draw/sdl/lv_draw_sdl_label.c
+++ b/src/draw/sdl/lv_draw_sdl_label.c
@@ -77,6 +77,21 @@ void lv_draw_sdl_draw_letter(lv_draw_ctx_t * draw_ctx, const lv_draw_label_dsc_t
            letter != 0xf8ff && /*LV_SYMBOL_DUMMY*/
            letter != 0x200c) { /*ZERO WIDTH NON-JOINER*/
             LV_LOG_WARN("lv_draw_letter: glyph dsc. not found for U+%X", letter);
+
+            /* draw placeholder */
+            lv_area_t glyph_coords;
+            lv_draw_rect_dsc_t glyph_dsc;
+            lv_coord_t begin_x = pos_p->x + g.ofs_x;
+            lv_coord_t begin_y = pos_p->y + g.ofs_y;
+            lv_area_set(&glyph_coords, begin_x, begin_y, begin_x + g.box_w, begin_y + g.box_h);
+            lv_draw_rect_dsc_init(&glyph_dsc);
+            glyph_dsc.bg_opa = LV_OPA_MIN;
+            glyph_dsc.outline_opa = LV_OPA_MIN;
+            glyph_dsc.shadow_opa = LV_OPA_MIN;
+            glyph_dsc.bg_img_opa = LV_OPA_MIN;
+            glyph_dsc.border_color = dsc->color;
+            glyph_dsc.border_width = 1;
+            draw_ctx->draw_rect(draw_ctx, &glyph_dsc, &glyph_coords);
         }
         return;
     }


X-Ryl669:
It seems wrong to me. This will pick the first font searched for even if it doesn't have the glyph. 
Said differently, `resolved_font` **should** be null if no font contains the searched glyph. Then if it's crashing later on, then it's this code that needs fixing (either by making it draw a square when `resolved_font` is null, or ignore the glyph, whatever).
**OR** LVGL should have a default font that answer "yes" to all glyph and draw the same path for all and then you can set `resolved_font` to that font.

guoweilkd:
> either by making it draw a square when `resolved_font` is null, or ignore the glyph, whatever

for TTF font files, it returns square glyph when current letter glyph is not found. 

See this [link](https://developer.apple.com/fonts/TrueType-Reference-Manual/RM06/Chap6glyf.html) for details

kisvegabor:
The API docs says:
> return true: descriptor is successfully loaded into `dsc_out`.
>           false: the letter was not found, no data is loaded to `dsc_out`

And it's really handled in LVGL. If this function returns `false` LVGL prints warnings and simply skips that letter. So there is no `□` drawn now.

> This will pick the first font searched for even if it doesn't have the glyph.

Nitpicking: I think this patch returns the last fallback.

---

So IMO using NULL as font if the glyph is not found is fine.

guoweilkd:
> and simply skips that letter. 

At present it seems to be while(1)  in lv_font_get_glyph_bitmap(). (LV_ASSERT_NULL->LV_ASSERT_HANDLER ->while(1)).

It is very familiar for font files to have missing glyphs, and the user certainly does not want to stop the program because of this. 
Using □ instead of missing glyphs is a convention and we should follow that.  Ignoring missing glyphs will confuse the user. for example, I'm not sure -> I'm sure if not have glyphs of `n`,`o`,`t`, the meaning changes completely.
Each font file may have a different style, □ should follow the style. So the display of □ should be handed over to the font file.

kisvegabor:
> It is very familiar for font files to have missing glyphs, and the user certainly does not want to stop the program because of this.

It won't stop because it returns earlier of the glyph is not found. This assert handles if glyph info can be retrieved but there is no bitmap. It indicates some kind of error.

> Using □ instead of missing glyphs is a convention and we should follow that. Ignoring missing glyphs will confuse the user.

So far I haven't dealt with the handling of □ fallback character to keep things as simple as possible. I see two possible problems:
1. If the user converts a ttf font to C array including only ASCII characters he also need to add the □ character. The user needs to know it and shouldn't forget to do it. 
2. E.g. a special monochrome or 7 segment font  might not contain □ glyph. 

Anyway, I agree that it'd be useful to visualize missing glyphs. However, I think it's specific to the given font engine and the handling of replacement character should be hidden in `lv_font_get_glyph_dsc`.

embeddedt:
Why don't we just render the pixels manually for □ instead of relying on a glyph?

kisvegabor:
> Why don't we just render the pixels manually for □ instead of relying on a glyph?

□ might look different in different fonts. But we can ignore this and just use the width of the `space` character and draw a border with let's say `min(font_height/20, 1)` width.

embeddedt:
We could also use it from the font if it exists, and only use an LVGL fallback if the glyph is missing.

kisvegabor:
> We could also use it from the font if it exists, and only use an LVGL fallback if the glyph is missing.

Sounds good. `lv_font_get_glyph_dsc` still needs to return the info of a □-like character to layout the texts (e.g. return space). 

`lv_font_glyph_dsc_t` already has an [is_placeholder](https://github.com/lvgl/lvgl/blob/e24f58962b97a92bdfc9db6689a4b855a535a64e/src/font/lv_font.h#L47) field to indicate exactly this case.

guoweilkd:
> This assert handles if glyph info can be retrieved but there is no bitmap. It indicates some kind of error.

Currently, when letter glaph is not found, get_glyph_dsc() set is_placeholder = 1 and return true (because of it returns □ glyph) in FreeType interface.This behavior may lead to resolved_font = NULL and return true in lv_font_get_glyph_dsc(). the result lead to invoke lv_font_get_glyph_bitmap() with resolved_font == NULL in lv_draw_sw_letter(). And then the code crashes.

Which of FreeType->get_glyph_dsc() or lv_font_get_glyph_dsc() should we fix to avoid this problem?


> We could also use it from the font if it exists
This PR can solve this problem, or do you have a better way?

kisvegabor:
Based on the discussion I see these scenarios:
1. The character is found in a font or in a fallback font. 
   - `lv_font_get_glyph_dsc()` returns `true`, `dsc` is set the metrics of the glyph and LVGL calls `lv_font_get_glyph_bitmap()`.  
2. The character is not found and none of the fonts has □ either. 
   -  `lv_font_get_glyph_dsc()` returns `false` but sets the `dsc` to the metrics of a `space` character (or any character). If LVGL sees `false` return value it draws the □ as a rectangle border and won't try to call `lv_font_get_glyph_bitmap`.
3. The character is not found but  □ is found.
   - `lv_font_get_glyph_dsc()` can return `true` and set the `dsc` of □. So LVGL doesn't have to know that the character is replaced by □. It just draws as if `á` looked like `□`.    `lv_font_get_glyph_bitmap()` is called which should return the bitmap of □.


Do you agree?

embeddedt:
Yes; that sounds clean to me.

guoweilkd:
> * return value it draws the □ as a rectangle border and won't try to call `lv_font_get_glyph_bitmap`.

here have a problem, the user layer uses adv_w to determine the coordinates of the next letter, which doesn't seem to be possible in lv_draw_sw_letter()

Is it possible to use the following scheme：
```c
static LV_ATTRIBUTE_LARGE_CONST const uint8_t glyph_bitmap[] = {
    0xff, 0x81, 0x81, 0x81, 0x81, 0x81, 0x81, 0xff
};

static const lv_font_t lv_font_missing_glyph = {
    .get_glyph_dsc = get_glyph_dsc_cb,
    .get_glyph_bitmap = get_glyph_bitmap_cb,
    .line_height = 4,          /*The maximum line height required by the font*/
    .base_line = 0,             /*Baseline measured from the bottom of the line*/
#if !(LVGL_VERSION_MAJOR == 6 && LVGL_VERSION_MINOR == 0)
    .subpx = LV_FONT_SUBPX_NONE,
#endif
#if LV_VERSION_CHECK(7, 4, 0) || LVGL_VERSION_MAJOR >= 8
    .underline_position = -1,
    .underline_thickness = 1,
#endif
    .dsc = NULL           /*The custom font data. Will be accessed by `get_glyph_bitmap/dsc` */
};

static bool get_glyph_dsc_cb(const lv_font_t * font, lv_font_glyph_dsc_t * dsc_out,
    uint32_t unicode_letter, uint32_t unicode_letter_next)
{
    dsc_out->adv_w = 8;
    dsc_out->box_h = 8;
    dsc_out->box_w = 8;
    dsc_out->ofs_x = 0;
    dsc_out->ofs_y = 0;
    dsc_out->bpp   = 1;
    dsc_out->is_placeholder = false;

    return true;
}

static const uint8_t * get_glyph_bitmap_cb(const lv_font_t * font, uint32_t unicode_letter)
{
    return (const uint8_t *)glyph_bitmap;
}

bool lv_font_get_glyph_dsc()
{
    LV_ASSERT_NULL(font_p);
    LV_ASSERT_NULL(dsc_out);
    const lv_font_t * f = font_p;
    bool found = false;
    while(f) {
        found = f->get_glyph_dsc(f, dsc_out, letter, letter_next);
        dsc_out->resolved_font = f;
        if(found && !dsc_out->is_placeholder) {
            break;
        }
        f = f->fallback;
    }

    //add 
    if (found == false) {
        f = &lv_font_missing_glyph;
        f->get_glyph_dsc(f, dsc_out, letter, letter_next);
        dsc_out->resolved_font = f;
    }

    return true;
}
```
The disadvantage is that the size of missing_glyph cannot change with font size.
Of course we can also make it dynamic, but bitmap will consume some memory.

kisvegabor:
> The disadvantage is that the size of missing_glyph cannot change with font size.

I think drawing a rectangle is more flexible.  

---

For the sake of simplicity let's ignore that case when "LVGL draws a rectangle border" and simply return a manually set glyph dsc like:
```c

    dsc_out->adv_w = font_height / 2;
    dsc_out->box_h = 0;
    dsc_out->box_w = 0;
    dsc_out->ofs_x = 0;
    dsc_out->ofs_y = 0;
    dsc_out->bpp   = (uint8_t)fdsc->bpp;
    dsc_out->is_placeholder = true;
``` 

As `adv_w` is set all text layouting functions work, but due to `box_w = 0` LVGL will return early in `lv_draw_sw_letter`.

src/draw/sw/lv_draw_sw_letter.c
```diff
@@ -98,13 +98,20 @@ void lv_draw_sw_letter(lv_draw_ctx_t * draw_ctx, const lv_draw_label_dsc_t * dsc
     lv_font_glyph_dsc_t g;
     bool g_ret = lv_font_get_glyph_dsc(dsc->font, &g, letter, '\0');
     if(g_ret == false)  {
-        /*Add warning if the dsc is not found
-         *but do not print warning for non printable ASCII chars (e.g. '\n')*/
-        if(letter >= 0x20 &&
-           letter != 0xf8ff && /*LV_SYMBOL_DUMMY*/
-           letter != 0x200c) { /*ZERO WIDTH NON-JOINER*/
```
kisvegabor: I think these non-printable characters should silently ignored

src/draw/sw/lv_draw_sw_letter.c
```diff
@@ -98,13 +98,20 @@ void lv_draw_sw_letter(lv_draw_ctx_t * draw_ctx, const lv_draw_label_dsc_t * dsc
...
-            LV_LOG_WARN("lv_draw_letter: glyph dsc. not found for U+%X", (unsigned int)letter);
```
kisvegabor, line 106: I'd still add a warning if a character is not found.

src/font/lv_font.c
```diff
@@ -71,12 +71,23 @@ bool lv_font_get_glyph_dsc(const lv_font_t * font_p, lv_font_glyph_dsc_t * dsc_o
...
+        dsc_out->resolved_font = f;
         if(found && !dsc_out->is_placeholder) {
-            dsc_out->resolved_font = f;
             break;
         }
         f = f->fallback;
     }
+
+    if(found == false) {
+        dsc_out->adv_w = font_p->line_height / 2;
```
kisvegabor: Add `dsc_out->resolved_font = NULL` to not point to font which doesn't contain the glyph.

src/font/lv_font.c
```diff
@@ -71,12 +71,23 @@ bool lv_font_get_glyph_dsc(const lv_font_t * font_p, lv_font_glyph_dsc_t * dsc_o
     bool found = false;
     while(f) {
         found = f->get_glyph_dsc(f, dsc_out, letter, letter_next);
+        dsc_out->resolved_font = f;
         if(found && !dsc_out->is_placeholder) {
```
kisvegabor: There can be 2 cases:
1. The glyph is found in a font -> great, we break and return;
2. Only a placeholder is found in a font -> We should save the font in which a placeholder was found, and if there is no better option use that.

src/font/lv_font.c
```diff
@@ -71,12 +71,23 @@ bool lv_font_get_glyph_dsc(const lv_font_t * font_p, lv_font_glyph_dsc_t * dsc_o
...
-            dsc_out->resolved_font = f;
             break;
```
kisvegabor:
```suggestion
            return true;
```

kisvegabor (commented):
Thanks for the update. I added a few comments.

src/font/lv_font.c
```diff
@@ -66,18 +66,41 @@ bool lv_font_get_glyph_dsc(const lv_font_t * font_p, lv_font_glyph_dsc_t * dsc_o
...
+            }
+            else if(placeholder_font == NULL) {
+                placeholder_font = f;
+            }
         }
         f = f->fallback;
     }
-    return found;
+
+    if(placeholder_font != NULL) {
```
kisvegabor, line 88: I meant this. :slightly_smiling_face: 

A minor optimization issue is that now `get_glyph_dsc` is called twice on the placeholder font. So `dsc_out` can be saved too, but it doesn't matter much as probably there are only a few missing characters.

kisvegabor (commented):
Thanks for the updates!

src/font/lv_font.c
guoweilkd: A second call to `get_glyph_dsc()` is necessary for FreeType interface. Before calling `lv_font_get_glyph_bitmap()`, need make sure that the `get_glyph_dsc()` of the current glyph is called.  `font->fallback()`  may erase the bitmap corresponding to the previous graph. So we have to call `get_glyph_dsc()`  again at the end.

kisvegabor: I see, okay.

kisvegabor:
Al looks good to me. Just please rebase to master to see if the CI error disappear.

guoweilkd:
Is CI error related to this PR?

kisvegabor:
I don't think so. I merge it and we will see. :)

Issue #2977, opened 2022-01-07
Add assertions levels


C47D:
### Introduce the problem
There are currently multiple asserts, but none to check if a condition is evaluated to be true

```
/*Enable asserts if an operation is failed or an invalid data is found.
 *If LV_USE_LOG is enabled an error message will be printed on failure*/
#define LV_USE_ASSERT_NULL          1   /*Check if the parameter is NULL. (Very fast, recommended)*/
#define LV_USE_ASSERT_MALLOC        1   /*Checks is the memory is successfully allocated or no. (Very fast, recommended)*/
#define LV_USE_ASSERT_STYLE         0   /*Check if the styles are properly initialized. (Very fast, recommended)*/
#define LV_USE_ASSERT_MEM_INTEGRITY 0   /*Check the integrity of `lv_mem` after critical operations. (Slow)*/
#define LV_USE_ASSERT_OBJ           0   /*Check the object's type and existence (e.g. not deleted). (Slow)*/

/*Add a custom handler when assert happens e.g. to restart the MCU*/
#define LV_ASSERT_HANDLER_INCLUDE <stdint.h>
#define LV_ASSERT_HANDLER while(1);   /*Halt by default*/
```

### Examples and cases

`lv_bar_set_range` doesn't validate if `max` is greater than `min`

```c
void lv_bar_set_range(lv_obj_t * obj, int32_t min, int32_t max)
{
    LV_ASSERT_OBJ(obj, MY_CLASS);

    lv_bar_t * bar = (lv_bar_t *)obj;

    if(bar->min_value == min && bar->max_value == max) return;

    bar->max_value = max;
    bar->min_value = min;

    if(lv_bar_get_mode(obj) != LV_BAR_MODE_RANGE)
        bar->start_value = min;

    if(bar->cur_value > max) {
        bar->cur_value = max;
        lv_bar_set_value(obj, bar->cur_value, LV_ANIM_OFF);
    }
    if(bar->cur_value < min) {
        bar->cur_value = min;
        lv_bar_set_value(obj, bar->cur_value, LV_ANIM_OFF);
    }
    lv_obj_invalidate(obj);
}
```

A test in the simulator, min is greater than max, and the bar is empty even when we set the value, without any clue why is that.
![image](https://user-images.githubusercontent.com/11037705/148614886-fab3de1d-7dab-4593-88d1-226fcfc85906.png)

### Suggested solution

[...]

PR #3001, opened 2022-01-13
Add LV_ASSERT_VERIFY

C47D:

See #2977 for context

### Checkpoints


C47D:
LVGL funtions using C-string manipulation functions that can lead to UB and `LV_ASSERT_VERIFY` could be used:

`strcpy`

- [ ] `lv_table_set_cell_value`
- [ ] `lv_label_set_text`
- [ ] `fs_dir_open`
- [ ] `fs_dir_read`
- [ ] `lv_tabview_add_tab`
- [ ] `lv_checkbox_set_text`
- [ ] `lv_dropdown_add_option`
- [ ] `lv_dropdown_set_options`
- [ ] `lv_textarea_set_text`
- [ ] `lv_textarea_set_placeholder_text`
- [ ] `lv_textarea_set_password_mode`
- [ ] `lv_img_set_src`
- [ ] `lv_img_decoder_open`
- [ ] `lv_roller_set_options`
- [ ] `lv_span_set_text`

---

In the else branch of `lv_span_set_text`, should we set `span->txt` to the return value of `lv_mem_realloc` (as in the if branch)? Maybe assert that it's not NULL?

```c
void lv_span_set_text(lv_span_t * span, const char * text)
{
    if(span == NULL || text == NULL) {
        return;
    }

    if(span->txt == NULL || span->static_flag == 1) {
        span->txt = lv_mem_alloc(strlen(text) + 1);
    }
    else {
        lv_mem_realloc(span->txt, strlen(text) + 1);
    }
    span->static_flag = 0;
    strcpy(span->txt, text);

    refresh_self_size(span->spangroup);
}
```

embeddedt:
> should we set span->txt to the return value of lv_mem_realloc (as in the if branch)?

Yes, I think that is a very real bug (good catch). There's no guarantee that `realloc` returns the same value that was provided.

C47D:
It would be great to be able to make realloc fail and see the bug being triggered.

I think a mock or a stub would do that.

> Yes, I think that is a very real bug (good catch). There's no guarantee that realloc returns the same value that was provided.

I think a separated PR is better, waiting for this one to be finished will take a lot longer. See #3005

C47D:
Some questions about `lv_checkbox_set_text`. Docs says the following:

imagen

1. What does "refresh with the current text" means?
2. When passing `NULL` to `txt`, do we need to do the allocations? It currently would allocate 1 byte (`strlen` of `NULL` returns 0) and copy `NULL` to `cb->txt`.
3. Do you think we could add a symbol named `LV_OBJ_REFRESH_WITH_CURRENT_TEXT` (or something similar) as an alias to `NULL` to make it more readable, (see 44e0c56, I didn't noticed `txt` being `NULL` is valid).

```c
{
    size_t len = _lv_txt_ap_calc_bytes_cnt(txt);
#else
    size_t len = strlen(txt);
#endif

    if(!cb->static_txt) cb->txt = lv_mem_realloc(cb->txt, len + 1);
    else  cb->txt = lv_mem_alloc(len + 1);
#else
    strcpy(cb->txt, txt);
#endif

    cb->static_txt = 0;

    lv_obj_invalidate(obj);
}
```
---
`test_checkbox_should_refresh_text` gives the following `SEGFAULT` cb50ccb, it seems like it's going on the `LV_USE_ARABIC_PERSIAN_CHARS` condition, ~~woudn't be `#if LV_USE_ARABIC_PERSIAN_CHARS == 1`?~~ just noticed the test configuration sets that symbol to 1.

imagen

C47D:
This solved the issue with `lv_checkbox_set_text`, let me know if it's correct and you want me to commit it, I also had to edit the test configuration, it was setting the `LV_USE_ARABIC_PERSIAN_CHARS` to 1, all the text are still passing.

```diff
diff --git a/src/widgets/lv_checkbox.c b/src/widgets/lv_checkbox.c
index 76097c41e..a4817b9ed 100644
--- a/src/widgets/lv_checkbox.c
+++ b/src/widgets/lv_checkbox.c
@@ -71,25 +71,27 @@ void lv_checkbox_set_text(lv_obj_t * obj, const char * txt)

     lv_checkbox_t * cb = (lv_checkbox_t *)obj;

+    if (NULL != txt) {
 #if LV_USE_ARABIC_PERSIAN_CHARS
-    size_t len = _lv_txt_ap_calc_bytes_cnt(txt);
+        size_t len = _lv_txt_ap_calc_bytes_cnt(txt);
 #else
-    size_t len = strlen(txt);
+        size_t len = strlen(txt);
 #endif

-    if(!cb->static_txt) cb->txt = lv_mem_realloc(cb->txt, len + 1);
-    else cb->txt = lv_mem_alloc(len + 1);
+        if(!cb->static_txt) cb->txt = lv_mem_realloc(cb->txt, len + 1);
+        else cb->txt = lv_mem_alloc(len + 1);

-    LV_ASSERT_MALLOC(cb->txt);
-    if (cb->txt == NULL) return;
+        LV_ASSERT_MALLOC(cb->txt);
+        if (cb->txt == NULL) return;

 #if LV_USE_ARABIC_PERSIAN_CHARS
-    _lv_txt_ap_proc(txt, cb->txt);
+        _lv_txt_ap_proc(txt, cb->txt);
 #else
-    strcpy(cb->txt, txt);
+        strcpy(cb->txt, txt);
 #endif

-    cb->static_txt = 0;
+        cb->static_txt = 0;
+    }

     lv_obj_refresh_self_size(obj);
     lv_obj_invalidate(obj);
```

kisvegabor:
> What does "refresh with the current text" means?

Probably it was copied from `lv_label_set_text`:
image

I added this a long time ago for hacks if the `label->text` is modified manually passing `NULL` makes the label recalculating its stuff without reallocating the buffer. 

I case of the checkbox it's a mistake in docs.

 > LVGL funtions using C-string manipulation functions that can lead to UB and LV_ASSERT_VERIFY could be used:



Could you show an example how would a `strcpy` with `LV_ASSERT_VERIFY` look like?

embeddedt:
Maybe I'm missing something, but IMO it makes far more sense to use `strncpy` then to add an assert on `strcpy`.

C47D:
> Maybe I'm missing something, but IMO it makes far more sense to use `strncpy` then to add an assert on `strcpy`.

Yes, I think the same, I didn't remembered that function until now. I can send a PR with this change later today.

> > What does "refresh with the current text" means?
> 
> Probably it was copied from `lv_label_set_text`: image
> 
> I added this a long time ago for hacks if the `label->text` is modified manually passing `NULL` makes the label recalculating its stuff without reallocating the buffer.
> 
> I case of the checkbox it's a mistake in docs.

How should the docs be updated?

> > LVGL funtions using C-string manipulation functions that can lead to UB and LV_ASSERT_VERIFY could be used:
> 
> Could you show an example how would a `strcpy` with `LV_ASSERT_VERIFY` look like?

I think @embeddedt proposal is much better but with `LV_ASSERT_VERIFY` you should assert the destination pointer can hold the contents of the source pointer.

```c
LV_ASSERT_VERIFY(len_of_dst_pointer >= len_of_src_pointer);
strcpy(dst, src);
```

**EDIT**

It seems like we still have some possible UB

image

kisvegabor:
What about adding `lv_strncpy` in which we can check for possible problems and assert if required?

C47D:
> What about adding `lv_strncpy` in which we can check for possible problems and assert if required?

Were should it be added?

---

Maybe this PR should be only about adding this new assert, and creating new ones for using it might be more effective. What do you think?

kisvegabor:
>> What about adding lv_strncpy in which we can check for possible problems and assert if required?

> Were should it be added?

`lv_txt.c` or `lv_mem.c`. I'd rather choose `lv_mem.c` as it's similar to `lv_memcpy`.

>  adding this new assert, and creating new ones

How do you mean creating new ones? Are we about creating new asserts besides `LV_ASSERT_VERIFY`?

C47D:
> How do you mean creating new ones? Are we about creating new asserts besides LV_ASSERT_VERIFY?

No, I meant using this PR to include the `LV_ASSERT_VERIFY` option, and use other new PRs to introduce it's usage in the codebase.

> lv_txt.c or lv_mem.c. I'd rather choose lv_mem.c as it's similar to lv_memcpy.

This could be the first PR where we use `LV_ASSERT_VERIFY`, inside `lv_strncpy`. Introducing a new file named `lv_str` would be ok?

kisvegabor:
> Introducing a new file named lv_str would be ok?

I'm ok with that too.

C47D:
Test added in cb50ccb3494ffccbf1a09bf8bf8262c590a98c1f is failing because of the text refreshing, how should the docs be updated to make it pass?

lv_conf_template.h
```diff
@@ -225,6 +225,7 @@
 #define LV_USE_ASSERT_STYLE         0   /*Check if the styles are properly initialized. (Very fast, recommended)*/
 #define LV_USE_ASSERT_MEM_INTEGRITY 0   /*Check the integrity of `lv_mem` after critical operations. (Slow)*/
 #define LV_USE_ASSERT_OBJ           0   /*Check the object's type and existence (e.g. not deleted). (Slow)*/
+#define LV_USE_ASSERT_VERIFY        0   /*Verify conditions*/
```
```suggestion
#define LV_USE_ASSERT_VERIFY        0   /*Verify conditions. Fast, recommended.*/
```
?

src/misc/lv_str.c
```diff
@@ -0,0 +1,58 @@
...
+/**********************
+ *   GLOBAL FUNCTIONS
+ **********************/
+
+char * lv_strncpy(char * dst, const char * src, size_t count)
+{
+    LV_ASSERT_VERIFY(dst);
+    LV_ASSERT_VERIFY(src);
+
+    char *ret = strncpy(dst, src, count);
```
kisvegabor: With the current code the string is iterated twice: once in `strncpy` and once in `strlen`. 
What about something like this:
```c
LV_ASSERT_VERIFY(count > 0);
size_t i;
for(i = 0; i < count - 1 && src[i] != '\0'; i++) {
   dst[i] = src[i];
}
dst[i] = '\0';
```

src/misc/lv_str.h
```diff
@@ -0,0 +1,52 @@
...
+
+/**********************
+ * GLOBAL PROTOTYPES
+ **********************/
+
+/**
+ * Wrapper for the standard strncpy
+ * @param dst pointer to the character array to copy to
+ * @param src pointer to the character array to copy from
+ * @param count maximum number of character to copy
```
kisvegabor:
```suggestion
 * @param count maximum number of character to copy (i.e. the size of `dst`)
```

src/misc/lv_str.h
```diff
@@ -0,0 +1,52 @@
...
+/**********************
+ *      TYPEDEFS
+ **********************/
+
+/**********************
+ * GLOBAL PROTOTYPES
+ **********************/
+
+/**
+ * Wrapper for the standard strncpy
```
kisvegabor:
```suggestion
 * Same as the standard `strncpy` but it always null-terminates the destination string
```

kisvegabor (commented):
> Test added in cb50ccb is failing because of the text refreshing, how should the docs be updated to make it pass?

"NULL to refresh with the current text." should be simply removed.

src/misc/lv_str.c
C47D: With this code the test in ade7126 failed, I had to replace `count -1` with `count`. See 1f38348

src/misc/lv_str.h
C47D: See 1f38348

src/misc/lv_str.h
C47D: See 1f38348

lv_conf_template.h
C47D: See 1f38348

C47D:
Currently if `LV_ASSERT_NULL` gets triggered (or any other assert), ie. when passing `text` as `NULL` to `lv_checkbox_set_text` the tests hangs, I guess because of the default handler (while(1)). I was thinking on how can we make sure the asserts gets triggered when intended.

That's why I had to remove the test in d47d3d4

C47D:
@kisvegabor Do you want me to change all `strcpy` usage with `lv_strncpy` in this PR or may I open a new one? Any other remaining work for this PR?

**EDIT**
I've just saw this other implementation of something similar https://github.com/openmv/openmv/commit/927854257cb136f11ba9b14c67062f70622e1722

kisvegabor:
> Do you want me to change all strcpy usage with lv_strncpy in this PR or may I open a new one? Any other remaining work for this PR?

Yes, it'd be great!

> I've just saw this other implementation of something similar openmv/openmv@9278542

Great idea!

C47D:
@kisvegabor Ready to be merged

kisvegabor:
I found something interesting. [Tutorialspoint](https://www.tutorialspoint.com/c_standard_library/c_function_strncpy.htm) says 
image
```c
char dst[128];
const char * src = "Hello";
strncpy(dest, src, 50); //Will it copy junk after the 6th byte? 
```

But `cplusplus.com` says:
image

[linux.die.net](https://linux.die.net/man/3/strncpy):
image
```c
char dst[128];
const char * src = "Hello";
strncpy(dest, src, sizeof(dest)); //Will it copy extra 122 zeros? 
```

I think only **Maximum number of characters to be copied** makes sens as it helps to respect the size of `dest` and do not add  `\0`s.

It it really works as `linux.die.net` says, I think we should add our own `strncpy` implementation. 
What do you think?

C47D:
> It it really works as linux.die.net says, I think we should add our own strncpy implementation.
What do you think?

So we avoid filling the extra spaces in `dst` with '\0'?

kisvegabor:
> So we avoid filling the extra spaces in dst with '\0'?

Yes, exactly.

embeddedt:
IMHO, if we're going to implement the function ourselves, it's better to implement a version of [`strlcpy`](https://linux.die.net/man/3/strlcpy). It expects the source to be a true C string and you just have to pass the destination's maximum size (it calculates the source size automatically).

kisvegabor:
I usually verify the existence of "exotic" functions in XC16 (16 bit PIC) and there was no `strlcpy` there. :slightly_frowning_face: 

There are these functions in `string.h`
```c
int memcmp(const void *, const void *, size_t);
void * memcpy(void *, const void *, size_t);
void * memmove(void *, const void *, size_t);
void * memset(void *, int, size_t);
char * strcat(char *, const char *);
int strcmp(const char *, const char *);
int strcoll(const char *, const char *);
char * strcpy(char *, const char *);
size_t strcspn(const char *, const char *);
char * strerror(int);
size_t strlen(const char *);
char * strncat(char *, const char *, size_t);
int strncmp(const char *, const char *, size_t);
char * strncpy(char *, const char *, size_t);
size_t strspn(const char *, const char *);
char * strtok(char *, const char *);
size_t strxfrm(char *, const char *, size_t);
void * memchr(const void *, int, size_t);
char * strchr(const char *, int);
char * strpbrk(const char *, const char *);
char * strrchr(const char *, int);
char * strstr(const char *, const char *);
```

embeddedt:
It's a BSD function, so it won't be in most standard libraries. We'd have to implement it ourselves. I just mentioned it because it solves the issue of not wasting time filling the destination with `\0`s, and it's a lot less confusing to use than `strncpy`, since it figures out the source length automatically.

If it were implemented using `memcpy` under the hood, performance shouldn't be an issue.

kisvegabor:
So basically do you suggest calling this function `lv_strlcpy`? If so, I'm fine with that.

embeddedt:
Exactly.

C47D:
@kisvegabor @embeddedt could you please do the update? I'm currently stuck with work stuff. I think you should be able to push to this branch.

kisvegabor:
No problem. :slightly_smiling_face: 

@embeddedt What about using [this function](https://github.com/lattera/freebsd/blob/master/sys/libkern/strlcpy.c) as it is? Of course with the correct attribution.

embeddedt:
Sounds good to me. I'm happy to update it as soon as I can, but I'm also a bit busy at the moment.

C47D:
@kisvegabor which is more urgent, this or the table tests? I can try to finish the realloc_safe today and follow with the next one.

kisvegabor:
Sorry, for my late reply. I was quite busy with finalizing [SquareLine](https://forum.lvgl.io/t/edgeline-khm-squreline-is-here/8134/4). 

> which is more urgent, this or the table tests?

Let's finish the table widget. It seems only a few things left there.

C47D:
Congratulations on the release of SquareLine, the demo looks really nice. I'm sure it took great job to get it done!

PR #3003, opened 2022-01-13
Add preprocessing and stop bubbling features for events

### LINKED ISSUE #2985: Pre-processing Event
### Introduce the problem
Right now, it's not possible to pre-process an event, so whatever we attach to an object, it'll always happen after the object as dealt with it.
For example, let's say you don't want a switch button not to go into its checked state when pressed and released, the only option is to maintain your own state tracking of the button.

In other GUI frameworks, you usually attach an event listener that's executed first and then prevent the event from continuing processing since you've dealt with it. (à la `e.preventDefault()` or `stopPropagation()`)

It won't be hard to add in LVGL so I'm wondering why it's not there.

### Examples and cases
Right now, the event processing code is doing this:
```cpp
lv_res_t lv_obj_event_base(const lv_obj_class_t * class_p, lv_event_t * e)
{
    const lv_obj_class_t * base;
    if(class_p == NULL) base = e->current_target->class_p;
    else base = class_p->base_class;

    /*Find a base in which call the ancestor's event handler_cb if set*/
    while(base && base->event_cb == NULL) base = base->base_class;

    if(base == NULL) return LV_RES_OK;
    if(base->event_cb == NULL) return LV_RES_OK;

    /*Call the actual event callback*/
    e->user_data = NULL;
    base->event_cb(base, e);

    lv_res_t res = LV_RES_OK;
    /*Stop if the object is deleted*/
    if(e->deleted) res = LV_RES_INV;

    return res;
}

static lv_res_t event_send_core(lv_event_t * e)
{
    EVENT_TRACE("Sending event %d to %p with %p param", e->code, (void *)e->current_target, e->param);

    /*Call the input device's feedback callback if set*/
    lv_indev_t * indev_act = lv_indev_get_act();
    if(indev_act) {
        if(indev_act->driver->feedback_cb) indev_act->driver->feedback_cb(indev_act->driver, e->code);
        if(e->deleted) return LV_RES_INV;
    }

    lv_res_t res = LV_RES_OK;
    res = lv_obj_event_base(NULL, e);

    lv_event_dsc_t * event_dsc = res == LV_RES_INV ? NULL : lv_obj_get_event_dsc(e->current_target, 0);

    uint32_t i = 0;
    while(event_dsc && res == LV_RES_OK) {
        if(event_dsc->cb && (event_dsc->filter == LV_EVENT_ALL || event_dsc->filter == e->code)) {
            e->user_data = event_dsc->user_data;
            event_dsc->cb(e);

            /*Stop if the object is deleted*/
            if(e->deleted) return LV_RES_INV;
        }

        i++;
        event_dsc = lv_obj_get_event_dsc(e->current_target, i);
    }

    if(res == LV_RES_OK && e->current_target->parent && event_is_bubbled(e)) {
        e->current_target = e->current_target->parent;
        res = event_send_core(e);
        if(res != LV_RES_OK) return LV_RES_INV;

    }

    return res;
}
```
In pseudo code, it's checking if there's a device filter and then it calls the class's event processing loop and finally the event descriptor's list.

### Suggested solution
We can add a preprocess member to the event list to indicate if it has to be processed first, and slighty change the loop like this:
```cpp
static lv_res_t event_send_core(lv_event_t * e)
{
    EVENT_TRACE("Sending event %d to %p with %p param", e->code, (void *)e->current_target, e->param);

    /*Call the input device's feedback callback if set*/
    lv_indev_t * indev_act = lv_indev_get_act();
    if(indev_act) {
        if(indev_act->driver->feedback_cb) indev_act->driver->feedback_cb(indev_act->driver, e->code);
        if(e->deleted) return LV_RES_INV;
    }

    lv_res_t res = LV_RES_OK;
    lv_event_dsc_t * event_dsc = res == LV_RES_INV ? NULL : lv_obj_get_event_dsc(e->current_target, 0);
    
    /*Preprocess event set first*/
    uint32_t i = 0;
    while(event_dsc && event_dsc->pre && res == LV_RES_OK) {
        if(event_dsc->pre && event_dsc->cb && (event_dsc->filter == LV_EVENT_ALL || event_dsc->filter == e->code)) {
            e->user_data = event_dsc->user_data;
            event_dsc->cb(e);

            /*Stop if the object is deleted*/
            if(e->deleted) return LV_RES_INV;
        }

        i++;
        event_dsc = lv_obj_get_event_dsc(e->current_target, i);
    }

    /*Call class processing*/
    res = lv_obj_event_base(NULL, e);

    /*Continue processing postprocess events*/
    while(event_dsc && res == LV_RES_OK) {
        if(event_dsc->cb && (event_dsc->filter == LV_EVENT_ALL || event_dsc->filter == e->code)) {
            e->user_data = event_dsc->user_data;
            event_dsc->cb(e);

            /*Stop if the object is deleted*/
            if(e->deleted) return LV_RES_INV;
        }

        i++;
        event_dsc = lv_obj_get_event_dsc(e->current_target, i);
    }

    if(res == LV_RES_OK && e->current_target->parent && event_is_bubbled(e)) {
        e->current_target = e->current_target->parent;
        res = event_send_core(e);
        if(res != LV_RES_OK) return LV_RES_INV;

    }

    return res;
}
```

A new function needs to be added to add preprocess events (it must be pushed to the head of the list), or normal events (added to the tail of the list).
Also, maybe it's worth changing the signature of the preprocess event callback to have them return a `lv_res_t` (with `LV_RES_OK` meaning continue processing and `LV_RES_INV` to stop processing immediately on return), as it'll be cleaner than setting `e->deleted` to 1

What do you think ?

X-Ryl669:

Fix #2985 

Example code:
```cpp
lv_obj_t * toggle = 0;

static void event_handler(lv_event_t * e)
{
    lv_obj_t * obj = lv_event_get_target(e);
    if (obj != toggle && toggle && !(lv_obj_get_state(toggle) & LV_STATE_CHECKED) && code == LV_EVENT_PRESSED) {
      /*Don't process this event, so the button will be inert unless the other button is toggled*/
      lv_event_stop_processing(e);
      LV_LOG_USER("Pressed, but ignored");
      return;
    }

    if(code == LV_EVENT_CLICKED) {
        LV_LOG_USER("Clicked");
    }
    else if(code == LV_EVENT_VALUE_CHANGED) {
        LV_LOG_USER("Toggled");
    }
}

void example_event(void)
{
    lv_obj_t * label;

    lv_obj_t * btn1 = lv_btn_create(lv_scr_act());
    lv_obj_add_event_cb(btn1, event_handler, LV_EVENT_PRESSED | LV_EVENT_PREPROCESS, NULL);
    lv_obj_align(btn1, LV_ALIGN_CENTER, 0, -40);

    label = lv_label_create(btn1);
    lv_label_set_text(label, "Button");
    lv_obj_center(label);

    lv_obj_t * btn2 = lv_btn_create(lv_scr_act());
    lv_obj_add_event_cb(btn2, event_handler, LV_EVENT_ALL, NULL);
    lv_obj_align(btn2, LV_ALIGN_CENTER, 0, 40);
    lv_obj_add_flag(btn2, LV_OBJ_FLAG_CHECKABLE);
    lv_obj_set_height(btn2, LV_SIZE_CONTENT);

    label = lv_label_create(btn2);
    lv_label_set_text(label, "Toggle");
    lv_obj_center(label);
    toggle = btn2;
}
```

### Checkpoints

### DIFF
--- a/src/core/lv_event.c
+++ b/src/core/lv_event.c
@@ -62,6 +62,8 @@ lv_res_t lv_event_send(lv_obj_t * obj, lv_event_code_t event_code, void * param)
     e.user_data = NULL;
     e.param = param;
     e.deleted = 0;
+    e.stop_bubbling = 0;
+    e.stop_processing = 0;
 
     /*Build a simple linked list from the objects used in the events
      *It's important to know if an this object was deleted by a nested event
@@ -115,7 +117,7 @@ lv_obj_t * lv_event_get_current_target(lv_event_t * e)
 
 lv_event_code_t lv_event_get_code(lv_event_t * e)
 {
-    return e->code;
+    return e->code & ~LV_EVENT_PREPROCESS;
 }
 
 void * lv_event_get_param(lv_event_t * e)
@@ -126,6 +128,16 @@ void * lv_event_get_param(lv_event_t * e)
 void * lv_event_get_user_data(lv_event_t * e)
 {
     return e->user_data;
+}
+
+void lv_event_stop_bubbling(lv_event_t * e)
+{
+    e->stop_bubbling = 1;
+}
+
+void lv_event_stop_processing(lv_event_t * e)
+{
+    e->stop_processing = 1;
 }
 
 
@@ -419,20 +431,42 @@ static lv_res_t event_send_core(lv_event_t * e)
     lv_indev_t * indev_act = lv_indev_get_act();
     if(indev_act) {
         if(indev_act->driver->feedback_cb) indev_act->driver->feedback_cb(indev_act->driver, e->code);
+        if(e->stop_processing) return LV_RES_OK;
         if(e->deleted) return LV_RES_INV;
     }
 
     lv_res_t res = LV_RES_OK;
-    res = lv_obj_event_base(NULL, e);
-
     lv_event_dsc_t * event_dsc = res == LV_RES_INV ? NULL : lv_obj_get_event_dsc(e->current_target, 0);
 
     uint32_t i = 0;
     while(event_dsc && res == LV_RES_OK) {
-        if(event_dsc->cb && (event_dsc->filter == LV_EVENT_ALL || event_dsc->filter == e->code)) {
+        if(event_dsc->cb  && ((event_dsc->filter & LV_EVENT_PREPROCESS) == LV_EVENT_PREPROCESS)
+           && (event_dsc->filter == (LV_EVENT_ALL | LV_EVENT_PREPROCESS) ||
+               (event_dsc->filter & ~LV_EVENT_PREPROCESS) == e->code)) {
             e->user_data = event_dsc->user_data;
             event_dsc->cb(e);
 
+            if(e->stop_processing) return LV_RES_OK;
+            /*Stop if the object is deleted*/
+            if(e->deleted) return LV_RES_INV;
+        }
+
+        i++;
+        event_dsc = lv_obj_get_event_dsc(e->current_target, i);
+    }
+
+    res = lv_obj_event_base(NULL, e);
+
+    event_dsc = res == LV_RES_INV ? NULL : lv_obj_get_event_dsc(e->current_target, 0);
+
+    i = 0;
+    while(event_dsc && res == LV_RES_OK) {
+        if(event_dsc->cb && ((event_dsc->filter & LV_EVENT_PREPROCESS) == 0)
+           && (event_dsc->filter == LV_EVENT_ALL || event_dsc->filter == e->code)) {
+            e->user_data = event_dsc->user_data;
+            event_dsc->cb(e);
+
+            if(e->stop_processing) return LV_RES_OK;
             /*Stop if the object is deleted*/
             if(e->deleted) return LV_RES_INV;
         }
@@ -445,7 +479,6 @@ static lv_res_t event_send_core(lv_event_t * e)
         e->current_target = e->current_target->parent;
         res = event_send_core(e);
         if(res != LV_RES_OK) return LV_RES_INV;
-
     }
 
     return res;
@@ -453,6 +486,8 @@ static lv_res_t event_send_core(lv_event_t * e)
 
 static bool event_is_bubbled(lv_event_t * e)
 {
+    if(e->stop_bubbling) return false;
+
     /*Event codes that always bubble*/
     switch(e->code) {
         case LV_EVENT_CHILD_CREATED:
--- a/src/core/lv_event.h
+++ b/src/core/lv_event.h
@@ -84,7 +84,11 @@ typedef enum {
     LV_EVENT_LAYOUT_CHANGED,      /**< The children position has changed due to a layout recalculation*/
     LV_EVENT_GET_SELF_SIZE,       /**< Get the internal size of a widget*/
 
-    _LV_EVENT_LAST                /** Number of default events*/
+    _LV_EVENT_LAST,               /** Number of default events*/
+
+
+    LV_EVENT_PREPROCESS = 0x80,   /** This is a flag that can be set with a event so it's processed
+                                      before the class default event processing */
 } lv_event_code_t;
 
 typedef struct _lv_event_t {
@@ -95,6 +99,8 @@ typedef struct _lv_event_t {
     void * param;
     struct _lv_event_t * prev;
     uint8_t deleted : 1;
+    uint8_t stop_processing : 1;
+    uint8_t stop_bubbling : 1;
 } lv_event_t;
 
 /**
@@ -183,6 +189,19 @@ void * lv_event_get_param(lv_event_t * e);
  */
 void * lv_event_get_user_data(lv_event_t * e);
 
+/**
+ * Stop the event from bubbling.
+ * This is only valid when called in the middle of a event processing chain.
+ * @param e     pointer to the event descriptor
+ */
+void lv_event_stop_bubbling(lv_event_t * e);
+
+/**
+ * Stop processing this event.
+ * This is only valid when called in the middle of a event processing chain.
+ * @param e     pointer to the event descriptor
+ */
+void lv_event_stop_processing(lv_event_t * e);
 
 /**
  * Register a new, custom event ID.


src/core/lv_event.c
```diff
@@ -419,20 +431,41 @@ static lv_res_t event_send_core(lv_event_t * e)
...
     lv_res_t res = LV_RES_OK;
-    res = lv_obj_event_base(NULL, e);
-
     lv_event_dsc_t * event_dsc = res == LV_RES_INV ? NULL : lv_obj_get_event_dsc(e->current_target, 0);
 
     uint32_t i = 0;
     while(event_dsc && res == LV_RES_OK) {
-        if(event_dsc->cb && (event_dsc->filter == LV_EVENT_ALL || event_dsc->filter == e->code)) {
+        if(event_dsc->cb  && ((event_dsc->filter & LV_EVENT_PREPROCESS) == LV_EVENT_PREPROCESS)
+           && (event_dsc->filter == LV_EVENT_ALL || (event_dsc->filter & ~LV_EVENT_PREPROCESS) == e->code)) {
```
embeddedt: Doesn't it need to be `(event_dsc->filter & ~LV_EVENT_PREPROCESS) == LV_EVENT_ALL`?

src/core/lv_event.c
```diff
@@ -419,20 +431,41 @@ static lv_res_t event_send_core(lv_event_t * e)
...
             e->user_data = event_dsc->user_data;
             event_dsc->cb(e);
 
+            if(e->stop_processing) return LV_RES_OK;
```
embeddedt, line 449: I'm not sure I like the idea of one event handler being able to cancel another one. IMO `stop_processing` should be named `prevent_default_processing` (or `prevent_def_proc`) and should only prevent `lv_obj_event_base` from being called.  Other custom event handlers can read the value of `e->prevent_def_proc` and skip running their logic if it's considered "default".

The current behavior allows one event handler to completely disable another one which seems dangerous and hard to predict.

embeddedt (commented):
The implementation is very easy to follow; thank you! I just have a few comments.

src/core/lv_event.c
X-Ryl669: Oops. You're right.

src/core/lv_event.c
X-Ryl669: I understand your point of view. However, the current implementation is more or less 1:1 with how browsers are dealing with events ([preventDefault](https://developer.mozilla.org/en-US/docs/Web/API/Event/preventDefault), [stopPropagation](https://developer.mozilla.org/en-US/docs/Web/API/Event/stopPropagation), [stopImmediatePropagation](https://developer.mozilla.org/en-US/docs/Web/API/Event/stopImmediatePropagation)), so I would say it's more likely the current implementation will match what a user might expect the function to do. 

Since LVGL appends event handlers in the end of an array, when you add an event, it'll not bother with previous events. 
So there's no way to prevent, in the current code, the processing of a previous event handler without having this "preprocessing" scheme.

Typically, this code can be used to filter events (in order to reduce the behavior of an existing scheme, like in the example, I'm preventing a button to react to **pressed** event if some external condition is not valid), like `preventDefault`. It allows processing and preprocessing event, so you can skip `lv_obj_event_base` as you said (`preventDefault`), but you can also avoid skipping it (simply append the event without preprocessing) and stop propagating further (`stopPropagation`). So best of both world here.

There's nothing that prevent an event handler to look at `e->stop_processing`, but since most (if not all) event handler are not going to be modified to act this way, it'll not work and the user will not understand. I think it's better to return in that condition. 

Also, skipping `lv_obj_event_base` processing but still processing other user event is a recipe for disaster, because the user handler make a lot of assumption of the expected state of the object in its handler. For example, if the user installed a handler for the **clicked** (or **released**) event, she will expect the button to change state when the event happen. But if one other handler had a "skip default processing" flag, this expectation is wrong and the user will go crazy to understand why she's seeing the event but the object doesn't change state. 
On the opposite, if the event processing stopped (like in this PR), her handler will never be called and everything will be correct, no expectations broken.

X-Ryl669: Also, one can still send an event to resume processing of an event. You just need to capture the event (in the preprocessing step), copy it. When some other condition happens, you'll set a flag somewhere in your code, and send the event you've captured & copied. It'll resume where it left, calling your preprocessing code (that will ignore doing anything since the flag is set), and then the real (usual) processing will happen. 

This is like async/await in Javascript.

kisvegabor: I agree with @X-Ryl669. Let's say I add preprocessed click event to a button and on the first click I create a pop up saying "Are you sure?" and prevent other click events to run.

kisvegabor (commented):
Looks good to me too.

src/core/lv_event.c
embeddedt: Fair enough; let's do it this way.

embeddedt (approved):

kisvegabor (approved):

kisvegabor:
Merging, thank you!

ym-terada:
Hello!
I recently started using LVGL v8.3 and was looking for a way to override event handlers.
I wanted to override LV_EVENT_KEY to change the cursor behavior, but I was having trouble because lv_obj_event() always handles LK_KEY_LEFT, LK_KEY_RIGHT. This is exactly the fix we needed!

But it doesn't seem to work well with this modification, at least on VisualStudio.

Even if I set LV_EVENT_PREPROCESS, the value of 'event_dsc->filter' gets sign extended to 0xffffff80. As a result, it cannot be preprocessed correctly.

I think the reason is that the 'filter' variable in lv_event_dsc_t is defined as enum, but also uses bit fields.
The size of the enum is not specified in the standard, and the behavior of the bit field depends on the build environment. Therefore, I think it would be more reliable to fix it as follows.

```
typedef struct _lv_event_dsc_t {
    lv_event_cb_t cb;
    void * user_data;
    uint8_t filter;		/* uint8_t to guarantee the size that lv_event_code_t expects. */
} lv_event_dsc_t;
```

What do you think?

kisvegabor:
Hey,

It makes sense. Please send a Pull request!

FYI, [in v9](https://github.com/lvgl/lvgl/blob/baf7fedc824b1f70bb3237a391478c6dd3b1e772/src/misc/lv_event.h#L36) it's already resolved as you have suggested.

PR #3012, opened 2022-01-15
Rounded img support

mariotaku:

This patch enables ability for `lv_img` to draw rounded image. This won't introduce any changes to API, instead, it makes `lv_img` aware of `lv_obj_set_style_radius`.

Software rendering:
image

Accelerated rendering with SDL (yes, it's visually identical):
image

On SDL it caches 4 rounded corners, and draw other parts directly without composition, so performance is pretty decent.

### Checkpoints


mariotaku:
See this feature work in action:

Radius works with shadows, and zoom.

image

src/widgets/lv_img.c
```diff
@@ -661,6 +663,13 @@ static void draw_img(lv_event_t * e)
             img_max_area.x2 -= pright;
             img_max_area.y2 -= pbottom;
 
+            int16_t radius_mask_id = LV_MASK_ID_INV;
+            lv_draw_mask_radius_param_t radius_param;
```
kisvegabor: The CI fails because masks are enabled only if `LV_DRAW_COMPLEX` is enabled.

kisvegabor (commented):
Thanks, good features. 

Before that you needed a wrapper object with `clip_corner=1` and some radius to achieve the same effect.

Please also update docs [here](https://docs.lvgl.io/master/widgets/core/img.html#parts-and-styles) to mention that the radius is also considered.

mariotaku:
Added fix & docs. Do you think this should be enabled only when clip corner option is set?

mariotaku:
Update: just noticed that `lv_style_set_radius` and `lv_style_set_clip_corner` combination works. So this PR will be a SDL performance patch only.

kisvegabor:
> just noticed that lv_style_set_radius and lv_style_set_clip_corner combination works.

Ah, you are right, I just forgot about it.

kisvegabor (approved):
Looks good, merging!

mariotaku:
Yes, please :)

PR #2993, opened 2022-01-11
feat: add cmsis-pack for lvgl

GorgonMeducer:
### Description of the feature or fix

Add cmsis-pack for LVGL 8.1.0

### Checkpoints
- [x] Follow the [styling guide](https://github.com/lvgl/lvgl/blob/master/docs/CODING_STYLE.md)
- [ ] Run `code-format.py` from the `scripts` folder. [astyle](http://astyle.sourceforge.net/install.html) needs to be installed.
- [ ] Update the documentation

PR #3021, opened 2022-01-18
feat: update the cmsis-pack to 0.8.3

GorgonMeducer:

- Update cmsis-pack with the latest LVGL
- Add README.md to the cmsis-pack folder, it shows the steps for generating cmsis-pack using existing gen_pack.sh
- Add benchmark as a module to cmsis-pack. It enables people to run benchmark immediately after a porting. 
- Typo fix

### Checkpoints


cmsis-pack/lv_demos.h
```diff
@@ -0,0 +1,42 @@
+/**
+ * @file lv_demos.h
```
kisvegabor: Would it be possible to copy this file by `gen_pack.sh`?

GorgonMeducer: No, this one is a patch for the original one.

GorgonMeducer: I tried to save the size of the cmsis-pack and only included benchmark and widgets.

GorgonMeducer: The original lv_demos.h include header files from other demos, especially from the music which is very big. So I have made this patch.

embeddedt:
Now that I look at it, it's a bit ugly to keep both the logic to generate the cmsis-pack and the final pack itself in the same tree. Typically I'd suggest using a CI action to generate the pack alongside a release. However, it seems like the pack can be updated more frequently than just every time there's a LVGL release, so that doesn't really work.

I really should have kept up with my emails better and mentioned this concern on #2993, sorry about that. I wonder if it'd be better to have a separate repository (`lv_cmsis_pack`) with the contents of this folder that can have its own releases?

GorgonMeducer:
@embeddedt it is also a proposal I mentioned before. @kisvegabor decided to keep just one release in the master repo.

image

kisvegabor:
> Now that I look at it, it's a bit ugly to keep both the logic to generate the cmsis-pack and the final pack itself in the same tree

Can we solve it by simply adding a `build` or `packs` folder? Or a `script` folder for the build scripts and extra files?

But TBH, the duplicated files worries me more. It'd be great to copy the original files in build time and update them with `sed` as needed. 

>  I wonder if it'd be better to have a separate repository (lv_cmsis_pack) with the contents of this folder that can have its own releases?

Wouldn't it go against our new monorepo policy?

GorgonMeducer:
> > Now that I look at it, it's a bit ugly to keep both the logic to generate the cmsis-pack and the final pack itself in the same tree
> 
> Can we solve it by simply adding a `build` or `packs` folder? Or a `script` folder for the build scripts and extra files?
> 
> But TBH, the duplicated files worries me more. It'd be great to copy the original files in build time and update them with `sed` as needed.
> 
> > I wonder if it'd be better to have a separate repository (lv_cmsis_pack) with the contents of this folder that can have its own releases?
> 
> Wouldn't it go against our new monorepo policy?

A more simple way would be adding macros to protect against those #includes in lv_demos.h, e.g.

```c
#if LV_USE_DEMO_BENCHMARK
    #include "benchmark/lv_demo_benchmark.h"
#endif
```

What do you think? @kisvegabor

GorgonMeducer:
Hi @kisvegabor ,

I have made the update, please review it.

cmsis-pack/lv_demos.h
GorgonMeducer: Solved in the latest push https://github.com/lvgl/lvgl/pull/3021/commits/1703d5fc8569525b32ac4e707632ff5d72b0685e

kisvegabor:
Thanks. I merge it to not block the "moving to `env_support`" topic but let's continue the discussion about cleaning up the structure. 


> Now that I look at it, it's a bit ugly to keep both the logic to generate the cmsis-pack and the final pack itself in the same tree. Typically I'd suggest using a CI action to generate the pack alongside a release. 

I was also thinking about using a CI action for a release. Finally, we needed to trigger it manually? If so, we can have a manually triggered CI action for cmsis-pack update too, can't we. And the release CI could reuse the cmsis-pack generator's CI in some way.

GorgonMeducer:
OK, let me move cmsis-pack to env_support first.

PR #3024, opened 2022-01-19
Fix/lru mem functions

mariotaku:

### Checkpoints

### DIFF
--- a/src/misc/lv_lru.c
+++ b/src/misc/lv_lru.c
@@ -292,7 +292,7 @@ static void lv_lru_remove_item(lv_lru_t * cache, lruc_item * prev, lruc_item * i
     cache->key_free(item->key);
 
     // push the item to the free items queue
-    lv_memset(item, 0, sizeof(lruc_item));
+    lv_memset_00(item, sizeof(lruc_item));
     item->next = cache->free_items;
     cache->free_items = item;
 }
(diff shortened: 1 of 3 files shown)


src/misc/lv_lru.c
```diff
@@ -82,14 +81,15 @@ lv_lru_t * lv_lru_new(size_t cache_size, size_t average_length, lv_lru_free_t *
     cache->free_memory = cache_size;
     cache->total_memory = cache_size;
     cache->seed = time(NULL);
```
kisvegabor: `time` is not provided by every platforms, so let's use something more generic. 
```suggestion
    cache->seed = lv_rand();
```

kisvegabor (commented):
Thanks, only one thing.

src/misc/lv_lru.c
mariotaku: Thanks for your suggestion. I have updated it also naming convention.

kisvegabor:
Thank you. The renaming changes look good too. 
(I missed that this file is out of the SDL folders)

mariotaku:
It's common if we need more cache features later. For example if we implement OpenGL based renderer, this will also be reused.

kisvegabor:
I agree!

PR #2841, opened 2021-11-27
add lv_chart_refresh() to the functions which modify the data

Miau09:
### Description of the feature or fix

Like mentioned here: https://github.com/lvgl/lvgl/issues/2817
I would suggest to add the lv_chart_refresh() to some more functions.
It was already in place for the lv_chart_set_next_value2(), maybe it was forgotten for the others?

It will then be in line with the wiki, because it does not mention to call the lv_chart_refresh() once the other functions are called.

### Checkpoints
- [x ] Follow the [styling guide](https://github.com/lvgl/lvgl/blob/master/docs/CODING_STYLE.md)
- [ ] Run `code-format.py` from the `scripts` folder. [astyle](http://astyle.sourceforge.net/install.html) needs to be installed.
- [ x] Update the documentation

PR #3028, opened 2022-01-20
replace lv_chart_refresh by invalidate_point so the chart will be updated faster.

ali-rostami:
Unfortunately in the PR #[2841 ](https://github.com/lvgl/lvgl/pull/2841) `lv_chart_refresh` was added by mistake.
Because `invalidate_point` was used before, to update the chart faster by only invalidating the area of the newly added data.

### DIFF
--- a/src/extra/widgets/chart/lv_chart.c
+++ b/src/extra/widgets/chart/lv_chart.c
@@ -539,7 +539,6 @@ void lv_chart_set_next_value(lv_obj_t * obj, lv_chart_series_t * ser, lv_coord_t
     invalidate_point(obj, ser->start_point);
     ser->start_point = (ser->start_point + 1) % chart->point_cnt;
     invalidate_point(obj, ser->start_point);
-    lv_chart_refresh(obj);
 }
 
 void lv_chart_set_next_value2(lv_obj_t * obj, lv_chart_series_t * ser, lv_coord_t x_value, lv_coord_t y_value)
@@ -556,11 +555,8 @@ void lv_chart_set_next_value2(lv_obj_t * obj, lv_chart_series_t * ser, lv_coord_
 
     ser->x_points[ser->start_point] = x_value;
     ser->y_points[ser->start_point] = y_value;
-    invalidate_point(obj, ser->start_point);
     ser->start_point = (ser->start_point + 1) % chart->point_cnt;
     invalidate_point(obj, ser->start_point);
-    lv_chart_refresh(obj);
-
 }
 
 void lv_chart_set_value_by_id(lv_obj_t * obj, lv_chart_series_t * ser, uint16_t id, lv_coord_t value)
@@ -571,7 +567,7 @@ void lv_chart_set_value_by_id(lv_obj_t * obj, lv_chart_series_t * ser, uint16_t 
 
     if(id >= chart->point_cnt) return;
     ser->y_points[id] = value;
-    lv_chart_refresh(obj);
+    invalidate_point(obj, id);
 }
 
 void lv_chart_set_value_by_id2(lv_obj_t * obj, lv_chart_series_t * ser, uint16_t id, lv_coord_t x_value,
@@ -589,7 +585,7 @@ void lv_chart_set_value_by_id2(lv_obj_t * obj, lv_chart_series_t * ser, uint16_t
     if(id >= chart->point_cnt) return;
     ser->x_points[id] = x_value;
     ser->y_points[id] = y_value;
-    lv_chart_refresh(obj);
+    invalidate_point(obj, id);
 }
 
 void lv_chart_set_ext_y_array(lv_obj_t * obj, lv_chart_series_t * ser, lv_coord_t array[])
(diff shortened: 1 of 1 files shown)


src/extra/widgets/chart/lv_chart.c
```diff
@@ -556,11 +555,8 @@ void lv_chart_set_next_value2(lv_obj_t * obj, lv_chart_series_t * ser, lv_coord_
 
     ser->x_points[ser->start_point] = x_value;
     ser->y_points[ser->start_point] = y_value;
-    invalidate_point(obj, ser->start_point);
```
ali-rostami, line 559: I removed this because `lv_chart_set_next_value2` is used only for scatter type, and inside the `invalidate_point` for the scatter type we invalidate the whole chart. 
I don't use scatter type, but I think it would be good to update only the needed area for this type too.

src/extra/widgets/chart/lv_chart.c
```diff
@@ -571,7 +567,7 @@ void lv_chart_set_value_by_id(lv_obj_t * obj, lv_chart_series_t * ser, uint16_t
 
     if(id >= chart->point_cnt) return;
     ser->y_points[id] = value;
-    lv_chart_refresh(obj);
+    invalidate_point(obj, id);
```
ali-rostami, line 570: I think for the `LV_CHART_TYPE_BAR` we need to add this line:
`invalidate_point(obj, id + 1);`
too but I'm not sure, I would appreciate you @kisvegabor to check it.

src/extra/widgets/chart/lv_chart.c
```diff
@@ -589,7 +585,7 @@ void lv_chart_set_value_by_id2(lv_obj_t * obj, lv_chart_series_t * ser, uint16_t
     if(id >= chart->point_cnt) return;
     ser->x_points[id] = x_value;
     ser->y_points[id] = y_value;
-    lv_chart_refresh(obj);
+    invalidate_point(obj, id);
```
ali-rostami, line 588: I know they both do the same, but I think It's better to keep it like this. As it is more consistent and also maybe someone works on improving the `invalidate_point` for `LV_CHART_TYPE_SCATTER` too.

kisvegabor:
I agree to use `invalidate_point` in non-scatter types. 
But for the sake of simplicity I'd keep `lv_chart_refresh()` for the scatter in order to fix the issues with the more common types as quick as possible.

src/extra/widgets/chart/lv_chart.c
kisvegabor: I've tested and it worked for me with `invalidate_point(obj, id);`. Why would it be required to use `+1`?

ali-rostami: > I've tested and it worked for me with `invalidate_point(obj, id);`. Why would it be required to use `+1`?

I'm saying it, just because of the way we do it in `lv_chart_set_next_value`, as you know we call invalidate_point 2 times. Not sure why!

kisvegabor: Hm, `lv_example_chart_2()` seems working well if `lv_chart_set_next_value` is modified like:
```c
    lv_chart_t * chart  = (lv_chart_t *)obj;
    ser->y_points[ser->start_point] = value;
    invalidate_point(obj, ser->start_point);
    ser->start_point = (ser->start_point + 1) % chart->point_cnt;
//    invalidate_point(obj, ser->start_point);
//    lv_chart_refresh(obj);
```

And it seems logical too. Can you confirm it?

ali-rostami: I tested `lv_example_chart_2()`, but even when I comment the first `invalidate_point` it still works!
```c
lv_chart_t * chart  = (lv_chart_t *)obj;
    ser->y_points[ser->start_point] = value;
 //   invalidate_point(obj, ser->start_point);
    ser->start_point = (ser->start_point + 1) % chart->point_cnt;
//    invalidate_point(obj, ser->start_point);
```
So I couldn't find where the chart gets invalidate. any idea?
don't you remember why the second `invalidate_point` was added before?
PS: I'm using the code blocks project and it seems should be updated to v8.2.0.

kisvegabor: It's not working for me with no `invalidate_point`. Let use one `invalidate_point` and merge it to see if any issues comes up.

ali-rostami: O.K, I removed it. So if it is O.K please merge.🙏

src/extra/widgets/chart/lv_chart.c
```diff
@@ -538,8 +538,6 @@ void lv_chart_set_next_value(lv_obj_t * obj, lv_chart_series_t * ser, lv_coord_t
     ser->y_points[ser->start_point] = value;
     invalidate_point(obj, ser->start_point);
     ser->start_point = (ser->start_point + 1) % chart->point_cnt;
-    invalidate_point(obj, ser->start_point);
-    lv_chart_refresh(obj);
 }
```
kisvegabor, line 542:
```suggestion
    invalidate_point(obj, ser->start_point);
}
```

kisvegabor:
I've just realized that the second `invalidate_point(obj, ser->start_point);` is required in `lv_chart_set_next_value` with `LV_CHART_UPDATE_MODE_CIRCULAR`.

PR #3036, opened 2022-01-23
fix(gradient) general cleanup and fix for alignment issues

X-Ryl669:

@kisvegabor Please test this PR if it fix the ASAN issue. Cheers!

If ok, we might need to find what is the minimum required alignment for the current platform instead of the hard-coded 8 bytes. Maybe 4 is enough. 2 seems not enough from the last error report.


embeddedt:
I have added ASAN support to the CI and added @kisvegabor's reproduction case as a test (so `master` currently fails). I can confirm that ASAN no longer reports any issues locally with this change.

X-Ryl669:
Does the CI runs on a CPU with alignment constraints (like ARM?)

embeddedt:
AFAIK it's x86-64.

kisvegabor:
Thanks for the PR!

I've tested it but locally it still failed for me after 2-3 sec.

So I looked deeper into the code and I think I found the problem. In `next_in_cache`:
```c
if((uint8_t *)item + s > grad_cache_end) return NULL;
```

should be:
```c
if((uint8_t *)item + s >= grad_cache_end) return NULL;
```

This way it work well for me but while I was testing with various configs I updated some other parts too:
- The current version automatically set the cache size to a large value. (Although the user could change it later). I moved the default cache size into lv_conf.h. Ideally it should be 0 by default. Also rearranged the configs a little and updated Kconfig too
- But the current code assumes the cache is large enough for any gradient. So I added to allocate non cached items if they can't fit into the cache size.
- Renamed `lv_gradient_t` -> `lv_grad_dsc_t` and `lv_gradient_cache_t` -> `lv_grad_t`. I found that these names are more consistent with other parts of LVGL and  as `lv_gradient_cache_t` can be non cached too this name is misleading.
- `lv_draw_sw_rect()` crashed with NULL pointer errors in some cases so I fixed them too.
- In `test_style()` I called `lv_timer_handler()` in loop the trigger some drawings.

docs/overview/style-props.md
```diff
@@ -252,8 +252,8 @@ Set the point from which the background's gradient color should start. 0 means t
 <li style='display:inline; margin-right: 20px; margin-left: 0px'><strong>Ext. draw</strong> No</li>
 </ul>
 
-### bg_gradient
-Set the gradient definition for the body. The pointed instance must exist while the object is alive. NULL to disable
+### bg_grad
+Set the gradient definition. The pointed instance must exist while the object is alive. NULL to disable. It wraps `BG_GRAD_COLOR`, `BG_GRAD_DIR`, `BG_MAIN_STOP` and `BG_GRAD_STOP` into one descriptor and allows craeting gradients with more colors too.
```
X-Ryl669, line 256: creating

src/draw/sw/lv_draw_sw_gradient.c
```diff
@@ -148,134 +156,154 @@ static lv_res_t kill_oldest_item(lv_gradient_cache_t * c, void * ctx)
...
+
     item->key = compute_key(g, size, w);
     item->life = 1;
     item->filled = 0;
     item->alloc_size = map_size;
     item->size = size;
-    item->map = (lv_color_t *)(grad_cache_end + sizeof(*item));
+    if(item->not_cached) {
+        uint8_t * p = (uint8_t *)item;
+        item->map = (lv_color_t *)(p + (ALIGN(sizeof(*item))));
```
X-Ryl669: I don't think this is required. Seems like the PR has shown it isn't. 
That's because sizeof(item) is aligned already.

src/draw/sw/lv_draw_sw_gradient.c
```diff
@@ -148,134 +156,154 @@ static lv_res_t kill_oldest_item(lv_gradient_cache_t * c, void * ctx)
...
+#if _DITHER_GRADIENT
+        item->hmap = (lv_color32_t *)(p + ALIGN(sizeof(*item) + ALIGN(map_size * sizeof(lv_color_t))));
```
X-Ryl669: Here, I have more doubts.

src/draw/sw/lv_draw_sw_gradient.c
kisvegabor: It's not sure. E.g. -Os can pack the structs automatically, so its size can be e.g. 27 bytes

X-Ryl669: No that's not possible. Else, if you had an array of those structures, the 2nd element in the array would cause unaligned access. The compiler must align the structure field to *at least* the most constrained field. It can pad the structure size (so if you do sizeof(*item) on a 27 bytes structure, it'll be 28 not 27, even if only the first 27 bytes are used in the structure).

X-Ryl669: In all cases, I don't think it's an issue, since if you ALIGN something that's already aligned, it'll give you the same number. It just make the code harder to follow.

src/draw/sw/lv_draw_sw_gradient.c
```diff
@@ -62,45 +69,44 @@ union void_cast {
...
+static size_t get_cache_item_size(lv_grad_t * c)
 {
-    size_t s = sizeof(*c) + c->alloc_size * sizeof(lv_color_t)
+    size_t s = ALIGN(sizeof(*c)) + ALIGN(c->alloc_size * sizeof(lv_color_t));
 #if _DITHER_GRADIENT
-               + c->size * sizeof(lv_color32_t)
+    s += ALIGN(c->size * sizeof(lv_color32_t));
 #if LV_DITHER_ERROR_DIFFUSION == 1
-               + c->w * sizeof(lv_scolor24_t)
+    s += ALIGN(+ c->w * sizeof(lv_scolor24_t));
```
X-Ryl669: Extra + in there ?

src/draw/sw/lv_draw_sw_gradient.c
X-Ryl669: The parenthesis looks wrong (it's not the same as the compute_size below).

I think that's the major issue with your architecture: lv_scolor24_t is 24 bits so it's clearly not aligned and your CPU is likely to require 32 bits alignment (or 16 bits). So this part must be the last in the structure.

X-Ryl669:
> In next_in_cache `if((uint8_t *)item + s > grad_cache_end) return NULL;`
> should be: `if((uint8_t *)item + s >= grad_cache_end) return NULL;`

Yes, good catch!

src/draw/sw/lv_draw_sw_gradient.c
embeddedt: > No that's not possible. Else, if you had an array of those structures, the 2nd element in the array would cause unaligned access. 

GCC is a little too smart sometimes. I don't know if it happens specifically in this case, but if it knows the target CPU (e.g. Cortex-M7) can support unaligned accesses, and you don't explicitly tell it not to use them with `-mno-unaligned-access`, I have absolutely seen it generate code that uses unaligned accesses like this.

tests/CMakeLists.txt
```diff
@@ -228,6 +235,7 @@ set(LVGL_TEST_OPTIONS_TEST_DEFHEAP
     ${LVGL_TEST_OPTIONS_TEST_COMMON}
     -DLVGL_CI_USING_DEF_HEAP
     -DLV_MEM_SIZE=2097152
+    -fsanitize=address
```
embeddedt, line 238: I intentionally left it out on `DEFHEAP`, because I don't think AddressSanitizer is nearly as useful when it doesn't also control heap allocations.

kisvegabor: I usually use with `LV_MEM_CUSTOM=0` and still see ASAN errors.  So I thought why not to use it here too.

embeddedt: Interesting. If it helps, let's definitely keep it.

src/draw/sw/lv_draw_sw_gradient.c
kisvegabor: I don't understand. How do you mean it?

embeddedt: The leading + is redundant before `c->w`, I think.

src/draw/sw/lv_draw_sw_gradient.c
kisvegabor: I fixed the parenthesis. 

I have a i5 8250U CPU. Nothing special.

src/draw/sw/lv_draw_sw_gradient.c
kisvegabor: I remember that we had issues with GCC packing structs that it wasn't able to read later. That's why we have introduced `LV_ATTRIBUTE_MEM_ALIGN` in `lv_conf.h`.

My experience is that there are too many exotic compilers and hardware so we should do all we can to support special cases too. Especially if it's so simple.

src/draw/sw/lv_draw_sw_gradient.c
kisvegabor: Ups, fixed. I'm a little bit blind sometimes :laughing:

src/draw/sw/lv_draw_sw_gradient.c
X-Ryl669: If the CPU can support unaligned access, then there's no alignment constraint anymore, right?
 
Damn, this makes my head hurt. You are right, there's the standard on one side with all its rules and there's GCC that bending them to its will and its strong knowledge of each platform. I'm 99% sure that because of "array like" access, structs must be aligned. Even an empty struct has a non zero sizeof because of this, the second element's address must be different than the former.

I don't know all architecture LVGL builds upon. But on ARM and ESP32 (which are both platform with alignment constraint I know about), their rules are quite simple, to access a 16 bits variable, it must be on (at least) a 16 bits boundary, and so on.
So, on those platform, the `ALIGN()` above will do nothing  since it'll happen for both pointers if the item is aligned (which is ensured by my initial PR's ALIGN usage). The last map (`error_acc` might be problematic since its 24 bits wide), but since it's the last, it won't push the other map to unaligned positions. This last map might require alignment to 32 bits or 16 bits, I don't know, probably need testing on a platform where alignment is an issue.

X-Ryl669: @kisvegabor If I guess correctly, you are using a ARM cpu (a mac ?) so you can assert my claims by removing those `ALIGN` lines and it should still work.

Please notice that keeping them, at worst, waste memory since any 16bits variable (`lv_color_t` get pushed to a 32bits or 64bits boundary). So at best, we are wasting 6 bytes per gradient (3 maps x 32 bits alignment) and 12 bytes at worst (3 maps x 64 bits alignment) for this *maybe* useless alignment constraint. Since we can have many of those when animating/occluding/etc... it might worth double checking to avoid this waste.

kisvegabor: I can't test it on an ARM based MAC at this moment but I'd accept the waste for higher reliability. 

We could save much more if we didn't allocate  `error_acc` and `hmap` if `LV_DITHER_NONE` is used. 

Anyway, I'd like to merge this fix as soon as possible to not have broken code in master and we can think about improvement options later.

X-Ryl669: > Anyway, I'd like to merge this fix as soon as possible to not have broken code in master and we can think about improvement options later.

Sure, feel free to merge. We can change later on to allocate upon requirement (by passing the dither argument to the gradient_get function)

embeddedt (approved):
I have an idea for adding arm32 and arm64 CIs - will look into it after this is merged.

kisvegabor:
> I have an idea for adding arm32 and arm64 CIs - will look into it after this is merged.

Amazing!

kisvegabor:
Merging.

PR #3040, opened 2022-01-23
Unit tests for table widget

C47D:

Unit tests for table widget.

### Checkpoints


C47D:
@kisvegabor @embeddedt Any thoughts on the new helpers on abbbbc3?

src/widgets/lv_table.c
```diff
@@ -38,6 +38,8 @@ static lv_coord_t get_row_height(lv_obj_t * obj, uint16_t row_id, const lv_font_
                                  lv_coord_t cell_left, lv_coord_t cell_right, lv_coord_t cell_top, lv_coord_t cell_bottom);
 static void refr_size(lv_obj_t * obj, uint32_t strat_row);
 static lv_res_t get_pressed_cell(lv_obj_t * obj, uint16_t * row, uint16_t * col);
+static size_t bytes_to_allocate(const char * txt);
```
kisvegabor:
```suggestion
static size_t get_cell_txt_len(const char * txt);
```

src/widgets/lv_table.c
```diff
@@ -38,6 +38,8 @@ static lv_coord_t get_row_height(lv_obj_t * obj, uint16_t row_id, const lv_font_
...
+static void copy_skip_format_byte(char *dst, const char *txt);
```
kisvegabor:
```suggestion
static void copy_cell_txt(char *dst, const char *txt);
```

kisvegabor (commented):
Looks clearer. I just suggested different names.

src/widgets/lv_table.c
C47D: Yep, your naming is much better, I will commit both suggestions later today.

src/widgets/lv_table.c
C47D: Ditto

src/widgets/lv_table.c
kisvegabor: Thanks ^^

src/widgets/lv_table.c
C47D: Done in 73d1309

src/widgets/lv_table.c
C47D: Done in 73d1309

kisvegabor:
Thank you Carlos! 

To be honest, my goal would be to have at least minimal unit tests for all widgets. So I think having these short unit tests for all widgets is better than having fewer but more detailed tests. It's because it's much easier to to add new test cases if something is already there. 

Anyway, it's great that you contribute with the tests, so if this approach feels better to you,  please keep doing it. :slightly_smiling_face:

C47D:
> Thank you Carlos!
> 
> To be honest, my goal would be to have at least minimal unit tests for all widgets. So I think having these short unit tests for all widgets is better than having fewer but more detailed tests. It's because it's much easier to to add new test cases if something is already there.
> 
> Anyway, it's great that you contribute with the tests, so if this approach feels better to you, please keep doing it. 🙂

Hi, I also don't like the elaborated tests. I will think of more test cases for the table, I'm using the opened issues as inspiration 😺

C47D:
In the static function `get_row_height` can we remove the cell sizes parameters and get them inside the function?

**EDIT**
I've started refactoring it, let me know what do you think. It being an static function didn't helped me adding tests for it.

C47D:
@kisvegabor Drawing tests are a bit more time consuming to do, I did only the easy ones 😸 
Ready to be reviewed, please validate both refactored static functions.

src/widgets/lv_table.c
```diff
@@ -671,18 +658,18 @@ static void draw_main(lv_event_t * e)
...
-                    if(next_cell_data) ctrl = next_cell_data[0];
-                    if(ctrl & LV_TABLE_CELL_CTRL_MERGE_RIGHT)
-                        if(rtl) cell_area.x1 -= table->col_w[col + col_merge + 1];
-                        else cell_area.x2 += table->col_w[col + col_merge + 1];
-                    else {
-                        break;
-                    }
-                }
-                else {
+                uint16_t index = cell + col_merge;
```
kisvegabor: `index` can be reserved word. I suggest `idx` instead.

src/widgets/lv_table.c
```diff
@@ -822,44 +814,45 @@ static lv_coord_t get_row_height(lv_obj_t * obj, uint16_t row_id, const lv_font_
     lv_coord_t h_max = lv_font_get_line_height(font) + cell_top + cell_bottom;
 
     for(cell = row_start, col = 0; cell < row_start + table->col_cnt; cell++, col++) {
-        if(table->cell_data[cell] != NULL) {
-            txt_w              = table->col_w[col];
-            uint16_t col_merge = 0;
-            for(col_merge = 0; col_merge + col < table->col_cnt - 1; col_merge++) {
+        /* NOTE: When can we have a cell_data pointing to NULL? Is continuing the best
```
kisvegabor: > When can we have a cell_data pointing to NULL? 

If the cell is empty, i.e. you haven't set any values yet.

kisvegabor: >  Is continuing the best way to handle it? 

We can wrap the reaming part into an `if` but `continue` is cleaner IMO.

tests/src/test_cases/test_table.c
```diff
@@ -0,0 +1,81 @@
...
+
+    TEST_ASSERT_EQUAL_STRING(value, lv_table_get_cell_value(table, row, column));
+}
+
+void test_table_should_grow_columns_automatically_when_setting_formatted_cell_value(void)
+{
+    uint16_t original_column_count = lv_table_get_col_cnt(table);
+    uint16_t expected_column_count = original_column_count + 1;
+
+    lv_table_set_cell_value_fmt(table, 0, 1, "LVGL %s", "Rocks!");
```
kisvegabor, line 38: Here you assume that the `original_column_count` is 0. 

Shall we add something like `TEST_ASSERT_UINT16(0, lv_table_get_col_cnt(table);)` too?

kisvegabor (commented):
> Drawing tests are a bit more time consuming to do, I did only the easy ones

No problem, having a ~5x5 table with a couple of "cell merging" and other cell controls are perfect as a first step.

src/widgets/lv_table.c
C47D: I think the same.

src/widgets/lv_table.c
C47D: See 318d0af

tests/src/test_cases/test_table.c
C47D: Actually I was assuming `original_column_count` is 1. I've added more comments and the assertion you've suggested in ab28670

C47D:
> No problem, having a ~5x5 table with a couple of "cell merging" and other cell controls are perfect as a first step.

Will work on it tomorrow, if we have 1 cell merging command, how many columns the table will end up having? Rows? I haven't reviewed cells commands yet.

kisvegabor:
> if we have 1 cell merging command, how many columns the table will end up having?

Merging doesn't affect the number of columns returned by `lv_table_get_col_cnt()`.

> Rows? 

There is no row merge now.

C47D:
@kisvegabor what should we be testing on the 5x5 table? Any particular property?

kisvegabor:
My ideas:
- Merging 2 cells
- Merging 5 cells
- Text align
- Multi line text (with '\n')
- Long text wrapped automatically
- `LV_TABLE_CELL_CTRL_TEXT_CROP` on long text

C47D:
~~@kisvegabor If we have table with one cell in `0,0` and want to add ctrl on the cell `0,5`, it currently only adds 1 column, so we end up with cells in `0,0` and `0,1`, is this correct or should it add cells in `0,0`, `0,1`, `0,2`, `0,3`, `0,4`?~~

Added tests marked with `x`, I don't know how to test the other ones, any suggestion?
- [ ] Merging 2 cells
- [ ] Merging 5 cells
- [ ] Text align
- [x] Multi line text (with '\n')
- [x] Long text wrapped automatically
- [ ] LV_TABLE_CELL_CTRL_TEXT_CROP on long text

For the merging cells I've found that the `LV_TABLE_CELL_CTRL_MERGE_RIGHT` is mostly used in `get_row_height` and it's used to calculate the return value of that function (`h_max`). So the test would be based on verifying `get_row_height` works as expected?

`LV_TABLE_CELL_CTRL_TEXT_CROP ` is only used in `draw_main` and it's a bit more difficult to test it, in my opinion. I will do more thinking about this one.

For text alignment, wouldn't be possible to test that in the label widget?

kisvegabor:
> For the merging cells I've found that the LV_TABLE_CELL_CTRL_MERGE_RIGHT is mostly used in get_row_height and it's used to calculate the return value of that function (h_max). So the test would be based on verifying get_row_height works as expected?

It also affects the drawing as no borders should be drawn between the merged cells.


> LV_TABLE_CELL_CTRL_TEXT_CROP is only used in draw_main and it's a bit more difficult to test it, in my opinion. I will do more thinking about this one.


I think all these things can be tested very well with screenshot comparison.
See the [dropdown's test](https://github.com/lvgl/lvgl/blob/master/tests/src/test_cases/test_dropdown.c#L423) as an example.  

> For text alignment, wouldn't be possible to test that in the label widget?


It's really hard to test as the texts of the cell are not label widgets, just texts drawn by `lv_draw_label`. So the result is drawn to the draw buffer.

embeddedt:
> I think all these thinks can be tested very well with screenshot comparison.

I would agree. The screenshots are probably a lot faster to make, too.

C47D:
Can we have those tests on some sort of todo list? The expected workload of my day job is expected to increase considerably and I would like to close as many open PR's as possible to don't let them unmaintained.

C47D:
~~Hi, so we would be testing the drawing of the table instead of the table functionality itself? If so, I think it would be better to add those kind of tests on a "drawing test file", just to keep things separated. One file for testing the drawing and another file to test behavior. What do you think?~~

**EDIT**
I'm not familiar with the drawing of widgets. The tests added are testing behavior based on the docs and sometimes in the code itself. 

Someone more familiar with how the table gets drawn should add test for that.

kisvegabor:
I added the rendering test.

kisvegabor:
Hmm, CI reports an ASAN error for the rendering test. I'll track it tomorrow.

C47D:
The heap overflow is in the `txt_w += table->col_w[idx + 1];` in `get_row_height`

```c
if(ctrl & LV_TABLE_CELL_CTRL_MERGE_RIGHT)
    txt_w += table->col_w[idx + 1];
```

kisvegabor:
Thanks,I've pushed a fix. Now there is no ASAN error but the rendering test still fails. Interestingly the table test is not executed locally for me:
```
Test project /home/kisvegabor/projects/littlevgl/eclipse-workspace/lv_sim_eclipse_sdl_master/lvgl/tests/build_test_defheap
      Start  5: test_demo_stress
      Start 12: test_style
      Start  7: test_dropdown
      Start  6: test_demo_widgets
      Start  3: test_checkbox
      Start 13: test_switch
      Start  1: test_arc
      Start  9: test_font_loader
 1/14 Test  #3: test_checkbox ....................   Passed    0.26 sec
      Start 11: test_snapshot
 2/14 Test  #1: test_arc .........................   Passed    0.29 sec
      Start  2: test_bar
 3/14 Test  #9: test_font_loader .................   Passed    0.30 sec
      Start  8: test_event
 4/14 Test #13: test_switch ......................   Passed    0.33 sec
      Start 10: test_obj_tree
 5/14 Test  #6: test_demo_widgets ................   Passed    0.36 sec
      Start 14: test_txt
 6/14 Test #11: test_snapshot ....................   Passed    0.19 sec
      Start  4: test_config
 7/14 Test  #2: test_bar .........................   Passed    0.27 sec
 8/14 Test  #8: test_event .......................   Passed    0.28 sec
 9/14 Test #14: test_txt .........................   Passed    0.25 sec
10/14 Test #10: test_obj_tree ....................   Passed    0.28 sec
11/14 Test  #4: test_config ......................   Passed    0.28 sec
12/14 Test  #7: test_dropdown ....................   Passed    1.01 sec
13/14 Test #12: test_style .......................   Passed    4.26 sec
14/14 Test  #5: test_demo_stress .................   Passed   11.92 sec
```

Are there any required extra steps to make it run?

kisvegabor:
Passing `--clean` made the table test run.

Most of the CI tests passed. :crossed_fingers:

C47D:
@kisvegabor Feel free to merge when you think it's ok, thanks for adding the rendering test 👍

kisvegabor:
Merging, thank you Carlos.

Issue #3039, opened 2022-01-23
V8.1 SDL GPU canvas does not display


iankellogg:
### Perform all steps below and tick them with [x]
- [ x] Read the [FAQ](https://docs.lvgl.io/master/intro/index.html#faq)
- [ x] Check the related part of the [Documentation](https://docs.lvgl.io/)
- [ x] Update lvgl to the latest version
- [ x] Reproduce the issue in a [Simulator](https://docs.lvgl.io/latest/en/html/get-started/pc-simulator.html)

### Describe the bug
After doing the fix in #3038 
Canvas crashes simulator when using SDL GPU
I added 
    #if LV_USE_GPU_SDL
        sdl_gpu_disp_drv_init(disp->driver);
    #else
    lv_disp_drv_init(disp->driver);
    #endif
to init_fake_disp
to fix the canvas crash but now canvas does not display in program
Images and other objects display just fine

### Expected behavior
Expected to see rendered canvas object in program when using SDL GPU. 

### Screenshots or video
![image](https://user-images.githubusercontent.com/19454690/150696799-bd8eca91-4543-4c7b-81be-205ee206b96e.png)
when using SDL GPU

![image](https://user-images.githubusercontent.com/19454690/150696841-85252d91-a5ff-4d8d-9903-5d6d33a0e28a.png)
when using SDL

kisvegabor:
cc @mariotaku

embeddedt:
@iankellogg I believe both of your issues are fixed in `master`.

kisvegabor:
As there was no activity here for a while  I suppose the issue is solved.

If it's not working feel free the comment here and reopen this issue.

Issue #3038, opened 2022-01-23
V8.1 SDL GPU crash with any draw


iankellogg:
### Perform all steps below and tick them with [x]
- [x ] Read the [FAQ](https://docs.lvgl.io/master/intro/index.html#faq)
- [ x] Check the related part of the [Documentation](https://docs.lvgl.io/)
- [ x] Update lvgl to the latest version
- [x ] Reproduce the issue in a [Simulator](https://docs.lvgl.io/latest/en/html/get-started/pc-simulator.html)

### Describe the bug
LVGL crashes in simulator when using SDL GPU when any lv_draw is called

In lv_draw_sdl_init_ctx the current repository code calls
draw_ctx_sdl->renderer = ((lv_draw_sdl_drv_param_t *) disp_drv->user_data)->renderer;

this puts the wrong address in draw_ctx_sdl->renderer

I was able to fix this with
draw_ctx_sdl->renderer = (lv_draw_sdl_drv_param_t *) disp_drv->user_data;

kisvegabor:
cc @mariotaku

mariotaku:
Please pass `lv_draw_sdl_drv_param_t` instead of `SDL_Renderer` to `disp_drv->userdata`. This struct is designed to hold more fields when needed.

And which simulator repo you are referring to? `lv_sim_vscode_sdl` right?

iankellogg:
Using something based on the eclipse simulator but has been rebuilt using
latest lv_driver, and latest lvgl with a new main.c
both the eclipse and vscode simulator don't run a version of LVGL that has
SDL_GPU support

the disp_drv->userdata to monitor.renderer in sdl_gpu_disp_drv_init in
sdl_gpu.c
If i change disp_drv->userdata to lv_draw_sdl_drv_param_t in sdl_gpu.c that
would require modifications all over sdl_gpu.c as that uses user_data
everytime for renderer. I can make those changes, but seems excessive
considering lv_draw_sdl_drv_param_t only contains SDL_Renderer.

jonbarnettk:
I agree with Iankellog.

I actually had this problem before and solved it in this way, but then I
had to add my own data to lv_draw_sdl_drv_param_t which then got
overridden when I went to master.

I feel this is inflexible because we lose user_data (those use it). For
example I kept SDL_Window in user_data and also created gl_context, because
it was useful for my purposes.

If this is the design, can we add a user_data to lv_draw_sdl_drv_param_t to
[...]

PR #3045, opened 2022-01-25
force canvas to use sw draw

mariotaku:

Fix for #3038 #3039 .

### Checkpoints

### DIFF
--- a/src/draw/sdl/lv_draw_sdl.h
+++ b/src/draw/sdl/lv_draw_sdl.h
@@ -44,6 +44,7 @@ typedef struct {
      * Render for display driver
      */
     SDL_Renderer * renderer;
+    void * userdata;
 } lv_draw_sdl_drv_param_t;
 
 typedef struct {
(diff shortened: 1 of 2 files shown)


src/draw/sdl/lv_draw_sdl.h
```diff
@@ -44,6 +44,7 @@ typedef struct {
      * Render for display driver
      */
     SDL_Renderer * renderer;
+    void * userdata;
```
kisvegabor: For consistency wit other parts of LVGL:

```suggestion
    void * user_data;
```

kisvegabor (commented):
I assume that the under lying problem is that the SDL backend can't draw into a pure color buffer. Is it correct?

mariotaku:
@kisvegabor Yes, it is that case. While canvas only support drawing into a byte buffer, I think forcing sw render is reasonable.

kisvegabor:
Makes sense, merging. Thanks you.

PR #3055, opened 2022-01-28
lv_mem: Add interface for safe reallocation

### LINKED ISSUE #2194: Missing allocation result check
https://github.com/lvgl/lvgl/blob/04f0ffc7756a5f623a3d3ab916efd82250c36e3a/src/core/lv_obj_class.c#L64

C47D:

Safe memory reallocation, fixes #2194 

### Checkpoints

### DIFF
--- a/demos/README.md
+++ b/demos/README.md
@@ -1,7 +1,7 @@
 # Demos for LVGL
 
 ## Add the examples to your projects
-1. demos can be found in the 'demos' folder once you clone the lvgl. 
+1. demos can be found in the 'demos' folder once you clone the lvgl.
 
 2. In the ***lv_conf.h*** or equivalent places, you can find demo related macros, change its value to enable or disable specified demos:
 
@@ -38,7 +38,7 @@
 ...
 ```
 
-3. If your development environment or toolchain does not add source files inside '***lvgl***' folder automatically, ensure the `demos` folder is included for compilation.  
+3. If your development environment or toolchain does not add source files inside '***lvgl***' folder automatically, ensure the `demos` folder is included for compilation.
 4. Include "***demos/lv_demos.h***" in your application source file, for example:
 
 ```c
@@ -53,7 +53,7 @@
 ## Demos
 
 ### Widgets
-Shows how the widgets look like out of the box using the built-in material theme.  
+Shows how the widgets look like out of the box using the built-in material theme.
 
 See in [lv_demo_widgets](https://github.com/lvgl/lv_examples/tree/master/src/lv_demo_widgets) folder.
 
@@ -62,20 +62,20 @@ See in [lv_demo_widgets](https://github.com/lvgl/lv_examples/tree/master/src/lv_
 For running this demo properly, please make sure **LV_MEM_SIZE** is at least **38KB** (and **48KB** is recommended):
 
 ```c
-#define LV_MEME_SIZE    (38ul * 1024ul)
+#define LV_MEM_SIZE    (38ul * 1024ul)
 ```
 
 
 
 ### Music player
-The music player demo shows what kind of modern, smartphone-like user interfaces can be created on LVGL. It works the best with display with 480x272 or 272x480 resolution. 
+The music player demo shows what kind of modern, smartphone-like user interfaces can be created on LVGL. It works the best with display with 480x272 or 272x480 resolution.
 
 See in [lv_demo_music](https://github.com/lvgl/lv_examples/tree/master/src/lv_demo_music) folder.
 
 <img src="https://github.com/lvgl/lv_examples/blob/master/src/lv_demo_music/screenshot1.gif?raw=true" width=600px alt="Music player demo with LVGL">
 
 ### Keypad and encoder
-LVGL allows you to control the widgets with a keypad and/or encoder without a touchpad. This demo shows how to handle buttons, drop-down lists, rollers, sliders, switches, and text inputs without touchpad. 
+LVGL allows you to control the widgets with a keypad and/or encoder without a touchpad. This demo shows how to handle buttons, drop-down lists, rollers, sliders, switches, and text inputs without touchpad.
 Learn more about the touchpad-less usage of LVGL [here](https://docs.lvgl.io/v7/en/html/overview/indev.html#keypad-and-encoder).
 
 See in [lv_demo_keypad_encoder](https://github.com/lvgl/lv_examples/tree/master/src/lv_demo_keypad_encoder) folder.
@@ -83,12 +83,12 @@ See in [lv_demo_keypad_encoder](https://github.com/lvgl/lv_examples/tree/master/
 <img src="https://github.com/lvgl/lv_examples/blob/master/src/lv_demo_keypad_encoder/screenshot1.png?raw=true" width=600px alt="Keypad and encoder navigation in LVGL embedded GUI library">
 
 ### Benchmark
-A demo to measure the performance of LVGL or to compare different settings. 
+A demo to measure the performance of LVGL or to compare different settings.
 See in [lv_demo_benchmark](https://github.com/lvgl/lv_examples/tree/master/src/lv_demo_benchmark) folder.
 <img src="https://github.com/lvgl/lv_examples/blob/master/src/lv_demo_benchmark/screenshot1.png?raw=true" width=600px alt="Benchmark demo with LVGL embedded GUI library">
 
 ### Stress
-A stress test for LVGL. It contains a lot of object creation, deletion, animations, style usage, and so on. It can be used if there is any memory corruption during heavy usage or any memory leaks. 
+A stress test for LVGL. It contains a lot of object creation, deletion, animations, style usage, and so on. It can be used if there is any memory corruption during heavy usage or any memory leaks.
 See in [lv_demo_stress](https://github.com/lvgl/lv_examples/tree/master/src/lv_demo_stress) folder.
 <img src="https://github.com/lvgl/lv_examples/blob/master/src/lv_demo_stress/screenshot1.png?raw=true" width=600px alt="Stress test for LVGL">
 
--- a/src/misc/lv_mem.c
+++ b/src/misc/lv_mem.c
@@ -187,7 +187,7 @@ void * lv_mem_realloc(void * data_p, size_t new_size)
         lv_mem_free(data_p);
         return &zero_mem;
     }
-    if(data_p == &zero_mem && new_size != 0) {
+    if((data_p == &zero_mem || data_p == NULL) && new_size != 0) {
         return lv_mem_alloc(new_size);
     }
 
--- a/src/misc/lv_mem.c
+++ b/src/misc/lv_mem.c
@@ -204,7 +204,7 @@ void * lv_mem_realloc(void * data_p, size_t new_size)
     return new_p;
 }
 
-bool lv_mem_realloc_safe(void * data, size_t new_size)
+bool lv_mem_realloc_safe(void ** data, size_t new_size)
 {
     void * new_p = NULL;
 
@@ -220,7 +220,8 @@ bool lv_mem_realloc_safe(void * data, size_t new_size)
     if(new_p != NULL) {
         MEM_TRACE("Memory allocated at %p", new_p);
         data = new_p;
-    } else {
+    }
+    else {
         LV_LOG_ERROR("couldn't allocate memory");
     }
 
--- a/src/misc/lv_mem.c
+++ b/src/misc/lv_mem.c
@@ -204,6 +204,29 @@ void * lv_mem_realloc(void * data_p, size_t new_size)
     return new_p;
 }
 
+bool lv_mem_realloc_safe(void * data, size_t new_size)
+{
+    void * new_p = NULL;
+
+    MEM_TRACE("reallocating %p with %lu size", data, (unsigned long)new_size);
+
+    if(data_p == &zero_mem) return lv_mem_alloc(new_size);
+
+#if LV_MEM_CUSTOM == 0
+    new_p = lv_tlsf_realloc(tlsf, data, new_size);
+#else
+    new_p = LV_MEM_CUSTOM_REALLOC(data_p, new_size);
+#endif
+    if(new_p != NULL) {
+        MEM_TRACE("Memory allocated at %p", new_p);
+        data = new_p;
+    } else {
+        LV_LOG_ERROR("couldn't allocate memory");
+    }
+
+    return new_p != NULL;
+}
+
 lv_res_t lv_mem_test(void)
 {
     if(zero_mem != ZERO_MEM_SENTINEL) {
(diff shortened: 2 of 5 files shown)


MaJerle (commented):
`bool lv_mem_realloc_safe(void * data, size_t new_size);` should accept rather `void ** data`

MaJerle (commented):
You need *data = new_ptr.

Has this been tested in reality?

C47D:
@MaJerle tested only by compiling the source code. I guess I can write an unit test that reallocates a bunch of memory until it fails.

MaJerle:
> @MaJerle tested only by compiling the source code. I guess I can write an unit test that reallocates a bunch of memory until it fails.

A test with success and failure would be nice for this function. I see some potential wrong implementation.

tests/src/test_cases/test_mem.c
```diff
@@ -0,0 +1,44 @@
...
+
+/* Non-zero size with null pointer should behave like malloc */
+void test_realloc_safe_behave_malloc_like(void)
+{
+    bool retval = true;
+    size_t allocation_size = 1;
+
+    void * ptr = NULL;
+    void * data = &ptr;
+    retval = lv_mem_realloc_safe(data, allocation_size);
```
MaJerle: Are you sure this works well? Don't you need to pass &data to realloc?

src/misc/lv_mem.c
```diff
@@ -210,22 +210,31 @@ bool lv_mem_realloc_safe(void ** data, size_t new_size)
...
-    }
-    else {
+    if(new_p == NULL) {
         LV_LOG_ERROR("couldn't allocate memory");
+        return false;
     }
 
-    return new_p != NULL;
+    MEM_TRACE("Memory allocated at %p", new_p);
+    data = new_p;
```
MaJerle: Should be rather `*data = new_p`.

src/misc/lv_mem.c
```diff
@@ -204,6 +204,39 @@ void * lv_mem_realloc(void * data_p, size_t new_size)
     return new_p;
 }
 
+bool lv_mem_realloc_safe(void ** data, size_t new_size)
+{
+    void * new_p = NULL;
+
+    MEM_TRACE("reallocating %p with %lu size", *data, (unsigned long)new_size);
+
+    if((*data == NULL) && (0 < new_size)) {
```
kisvegabor: `size_t` is always unsigned so it can't be negative.
https://stackoverflow.com/questions/1089176/is-size-t-always-unsigned

src/misc/lv_mem.c
```diff
@@ -204,6 +204,39 @@ void * lv_mem_realloc(void * data_p, size_t new_size)
...
+        *data = lv_mem_alloc(new_size);
+        return false;
```
kisvegabor: What about returning `LV_RES_OK/INV`? It's cleaner then `true/false`.

src/misc/lv_mem.c
```diff
@@ -204,6 +204,39 @@ void * lv_mem_realloc(void * data_p, size_t new_size)
...
+
+#if LV_MEM_CUSTOM == 0
+    new_p = lv_tlsf_realloc(tlsf, data, new_size);
+#else
+    new_p = LV_MEM_CUSTOM_REALLOC(data, new_size);
+#endif
+    if(new_p == NULL) {
+        LV_LOG_ERROR("couldn't allocate memory");
+        return false;
+    }
```
kisvegabor: All these should be handled in `lv_mem_realloc()`

src/misc/lv_mem.c
C47D: Will replace with a `0 != new_size` check

src/misc/lv_mem.c
C47D: It's ok, I was thinking on returning the status of the allocation. `LV_RES_OK` if the allocation was done, `LV_RES_INV` otherwise. Are those enums already available?

src/misc/lv_mem.c
C47D: So
```c
#if LV_MEM_CUSTOM == 0
    new_p = lv_tlsf_realloc(tlsf, data, new_size);
#else
    new_p = LV_MEM_CUSTOM_REALLOC(data, new_size);
#endif
```
should become?
```c
lv_mem_realloc(data, new_size)
```

src/misc/lv_mem.c
MaJerle: It cant be negative but can be zero. Check seems legit unless there are other code style rules by lib.

src/misc/lv_mem.c
MaJerle: Maybe have a look here and please read func desc. Implementation of safe version shall be super super non-complex. 

https://github.com/MaJerle/lwmem/blob/de2917481ddcb02d4789b188f7c8209af6cdcc17/lwmem/src/lwmem/lwmem.c#L894

C47D:
It seems to me that I don't fully understand double pointers, I can't get the test to pass. I think the prototype is documented similar to the reference implementation. I will leave this in your hands @kisvegabor @MaJerle 

I will follow the updates and try to understand from there.

tests/src/test_cases/test_mem.c
```diff
@@ -0,0 +1,40 @@
...
+
+void tearDown(void)
+{
+    /* Function run after every test */
+}
+
+/* Non-zero size with null pointer should behave like malloc */
+void test_realloc_safe_behave_malloc_like(void)
+{
+    void * ptr = NULL;
```
MaJerle: You need only one variable, either ptr or data, but not both.

tests/src/test_cases/test_mem.c
```diff
@@ -0,0 +1,40 @@
...
+    void * data = &ptr;
+    size_t allocation_size = sizeof(lv_font_t);
+
+    lv_res_t retval = lv_mem_realloc_safe(&data, allocation_size);
```
MaJerle: This looks OK

tests/src/test_cases/test_mem.c
```diff
@@ -0,0 +1,40 @@
...
+    TEST_ASSERT_EQUAL_UINT16(retval, LV_RES_OK);
+
+    /* Don't leak memory */
+    lv_mem_free(ptr);
```
MaJerle: You free either data or ptr, the variable you have, but not both.

tests/src/test_cases/test_mem.c
```diff
@@ -0,0 +1,40 @@
...
+}
+
+/* a zero size with a non-null pointer will behave like free and return LV_RES_OK. */
+void test_realloc_safe_behave_free_like(void)
+{
+    void * font = lv_mem_alloc(1 * sizeof(lv_font_t));
+    void * data = &font;
```
MaJerle: Same here. proper test code should be

```c
void* font = lv_mem_alloc(1 * sizeof(lv_font_t));

//And then to realloc (or free in this case since len is 0)
lv_res_t retval = lv_mem_realloc_safe(&font, 0);

// font should now point to NULL as safe function should set it to NULL as it was free operation
if (font != NULL) {
    //Error
}
```

src/misc/lv_mem.c
```diff
@@ -204,6 +205,36 @@ void * lv_mem_realloc(void * data_p, size_t new_size)
...
+
+    if(data == NULL) {
+        LV_LOG_ERROR("Invalid input");
+        return LV_RES_INV;
+    }
+
+    MEM_TRACE("Safely reallocating %p with %lu size", *data, (unsigned long) new_size);
+
+#if LV_MEM_CUSTOM == 0
+    new_p = lv_tlsf_realloc(tlsf, data, new_size);
```
MaJerle: You need existing address, so it must be `*data` in parameter

src/misc/lv_mem.c
```diff
@@ -204,6 +205,36 @@ void * lv_mem_realloc(void * data_p, size_t new_size)
...
+#else
+    new_p = LV_MEM_CUSTOM_REALLOC(data, new_size);
```
MaJerle: You need existing address, so it must be `*data` in parameter

src/misc/lv_mem.c
```diff
@@ -204,6 +205,36 @@ void * lv_mem_realloc(void * data_p, size_t new_size)
...
+    /* Handle return status for free-like operation */
+    if(new_size == 0) {
+        *data = &zero_mem;
+        return LV_RES_OK;
+    }
+    if(new_p == NULL) {
+        return LV_RES_INV;
+    }
+
+    *data = new_p;
```
MaJerle, line 223: This looks good

src/misc/lv_mem.c
C47D: Done in 503e757

src/misc/lv_mem.c
C47D: Done in 503e757

tests/src/test_cases/test_mem.c
C47D: I've refactored the test a bit, can you check 503e757?

C47D: It currently sets the pointer to `zero_mem` (with value `ZERO_MEM_SENTINEL`), should it set it to NULL instead?

tests/src/test_cases/test_mem.c
C47D: So it is ok to only free `ptr`?

tests/src/test_cases/test_mem.c
C47D: Thanks, fixed in 986a62d

C47D:
Thanks for the review @MaJerle , it seems like all I needed was a bit of guidance and go away from the PC for some hours 😺 . I've addressed most of your comments, only 2 are left opened for your feedback. What timezone are you in? I'm in Mexico.

Also, if you have more test cases we can try to add them.

src/misc/lv_mem.c
```diff
@@ -182,13 +182,14 @@ void lv_mem_free(void * data)
...
     MEM_TRACE("reallocating %p with %lu size", data_p, (unsigned long)new_size);
-    if(new_size == 0) {
+    if(data_p != NULL && new_size == 0) {
         MEM_TRACE("using zero_mem");
         lv_mem_free(data_p);
         return &zero_mem;
     }
-
-    if(data_p == &zero_mem) return lv_mem_alloc(new_size);
+    if(data_p == &zero_mem && new_size != 0) {
```
MaJerle: We need to check if it is NULL (first try) or if it is zero mem.

`if((data_p == &zero_mem || data_p == NULL) && new_size != 0) {`

MaJerle:
I'm in EMEA, CET timezone, like Gabor.

src/misc/lv_mem.c
C47D: Resolved with 48c3333

src/misc/lv_mem.c
C47D: Resolved in 0e0bebd

tests/src/test_cases/test_mem.c
C47D: How does 60b0f318 look? I think we could also move `ZERO_MEM_SENTINEL` to `lv_mem.h` to be able to use it on the tests, any thoughts @kisvegabor ?

C47D:
There's only one open question (I think). What is the plan, merging this and then work on using it through the code base in another PR?

kisvegabor:
It looks great in general. So far I haven't tried the "test-driven-development" and to be honest it feels great. Thank you, Carlos for  introducing this pattern in LVGL. :slightly_smiling_face:   

I've added a few minor changes to the code.  What do you think?

> What is the plan, merging this and then work on using it through the code base in another PR?

Using `lv_mem_realloc_safe` in lvgl should be straightforward so I think we can add them in this PR.

demos/README.md
```diff
@@ -53,11 +53,20 @@
...
+Shows how the widgets look like out of the box using the built-in material theme.  
+
 See in [lv_demo_widgets](https://github.com/lvgl/lv_examples/tree/master/src/lv_demo_widgets) folder.
 
 <img src="https://github.com/lvgl/lv_examples/blob/master/src/lv_demo_widgets/screenshot1.png?raw=true" width=600px alt="Basic demo to show the widgets of LVGL">
 
+For running this demo properly, please make sure **LV_MEM_SIZE** is at least **38KB** (and **48KB** is recommended):
+
+```c
+#define LV_MEME_SIZE    (38ul * 1024ul)
```
X-Ryl669: Seems like a typo.

C47D: Good catch!. It was introduced in 7f4cde9388a649fa31fb286dd1052d1108eb7029, will send the fix.

C47D: See f93ce64

C47D:
@kisvegabor was the free-like test removed by accident?

---

> Using lv_mem_realloc_safe in lvgl should be straightforward so I think we can add them in this PR.

Great, lets do it!, should we add the `LV_ASSERT_RETURN` macro discussed in this comment? https://github.com/lvgl/lvgl/issues/2194#issuecomment-1022855269

---

How does 84e4544 and 7f14085 looks?

C47D:
@kisvegabor @MaJerle Some time ago I used [fff](https://github.com/meekrosoft/fff) to fake functions in this project [WFlash](https://github.com/C47D/WFlash/blob/master/test/test_wflash.c), you can see the project also uses Unity, so they play along nicely. I think we could fake the behavior of `lv_mem_realloc` inside `lv_mem_realloc_safe` and test that the input pointer doesn't get updated, etc.

Also we can define what should happen when reallocation fails and also test it, e.g. what should happen when reallocation fails when calling `lv_checkbox_set_text`?

What do you think?

**EDIT**
Just did a quick test and I think we can't use fff because we wouldn't be able to  include the source file were the 'faked' function is defined.

**EDIT2**
Can't define assert handlers in the tests CMakeList.txt file because CMake thinks we end the command with the assert parenthesis. Adding this to the option lists breaks it.

```
-DLV_ASSERT_HANDLER=user_assert\\(\\)\\;
```

kisvegabor:
> was the free-like test removed by accident?

It is handled in `lv_mem_realloc` (i.e. freeing and returning `&zero_mem`)

> Can't define assert handlers in the tests CMakeList.txt file because CMake thinks we end the command with the assert parenthesis. Adding this to the option lists breaks it.

it'd be really great but from [here](https://cmake.org/cmake/help/latest/prop_dir/COMPILE_DEFINITIONS.html):
> Function-style definitions are not supported

But we can add something like this in `lv_test_conf.h`
```c
#if LV_TEST_CUSTOM_ASSERT
# define LV_ASSERT_HANDLER {extern int lv_assert_counter; lv_assert_counter++;}
#endif
```

C47D:
We can leave testing assertions for later (I think), is the change in checkbox the expected from introducing the new API?

C47D:
@kisvegabor should we continue with this PR?

kisvegabor:
Based on @X-Ryl669's comments in #2194 I'm not sure really this is the best direction. 

Implementing @xiaoxiang781216's "emergency free hook" should solve more problems with less effort.

So I suggest closing this PR for now. We can pick up from here if we really want to add a safe realloc later.

What do you  think?

C47D:
It's ok, I think we are solving two different issues. I will check the information on the comment.

PR #3063, opened 2022-01-30
feat(porting): add a macro lv_run_timer_handler_in_period to simplify porting

GorgonMeducer:

In the past, a recommended way to call `lv_timer_handler()` is 
> ...It's a better approach to simply set a flag or some value in the interrupt, and periodically check it in an LVGL timer (which is run by `lv_timer_handler`)...

A helper macro `lv_run_timer_handler_in_period()` provide a simplified way to call `lv_timer_handler()` repeatedly in a given period. It uses lv_tick_get() to avoid using flags+Timer ISR. 

### Checkpoints

### DIFF
--- a/src/misc/lv_timer.h
+++ b/src/misc/lv_timer.h
@@ -28,21 +28,24 @@ extern "C" {
 
 /**
  * Call it in the super-loop of main() or threads. It will run lv_timer_handler()
- * with a given period in ms.
- * This macro is used to simplify the porting.
+ * with a given period in ms. You can use it with sleep or delay in OS environment.
+ * This function is used to simplify the porting.
  * @param __ms the period for running lv_timer_handler()
  */
-#define lv_run_timer_handler_in_period(__ms)                                    \
-        do {                                                                    \
-            static uint32_t last_tick = 0;                                      \
-            uint32_t curr_tick = lv_tick_get();                                 \
-                                                                                \
-            if ((curr_tick - last_tick) >= (__ms)) {                            \
-                last_tick = curr_tick;                                          \
-                lv_timer_handler();                                             \
-            }                                                                   \
-        } while(0)
         
+static inline LV_ATTRIBUTE_TIMER_HANDLER 
+uint32_t lv_timer_handler_run_in_period(uint32_t ms)
+{
+    static uint32_t last_tick = 0;
+    uint32_t curr_tick = lv_tick_get();
+
+    if ((curr_tick - last_tick) >= (__ms)) {
+        last_tick = curr_tick;
+        return lv_timer_handler();
+    }
+    return 1;
+}
+
 /**********************
  *      TYPEDEFS
  **********************/
--- a/src/misc/lv_timer.h
+++ b/src/misc/lv_timer.h
@@ -26,6 +26,23 @@ extern "C" {
 
 #define LV_NO_TIMER_READY 0xFFFFFFFF
 
+/**
+ * Call it in the super-loop of main() or threads. It will run lv_timer_handler()
+ * with a given period in ms.
+ * This macro is used to simplify the porting.
+ * @param __ms the period for running lv_timer_handler()
+ */
+#define lv_run_timer_handler_in_period(__ms)                                    \
+        do {                                                                    \
+            static uint32_t last_tick = 0;                                      \
+            uint32_t curr_tick = lv_tick_get();                                 \
+                                                                                \
+            if ((curr_tick - last_tick) >= (__ms)) {                            \
+                last_tick = curr_tick;                                          \
+                lv_timer_handler();                                             \
+            }                                                                   \
+        } while(0)
+        
 /**********************
  *      TYPEDEFS
  **********************/
--- a/src/misc/lv_timer.h
+++ b/src/misc/lv_timer.h
@@ -13,6 +13,7 @@ extern "C" {
  *      INCLUDES
  *********************/
 #include "../lv_conf_internal.h"
+#include "../hal/lv_hal_tick.h"
 
 #include <stdint.h>
 #include <stdbool.h>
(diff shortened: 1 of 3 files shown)


src/misc/lv_timer.h
```diff
@@ -26,6 +26,23 @@ extern "C" {
 
 #define LV_NO_TIMER_READY 0xFFFFFFFF
 
+/**
+ * Call it in the super-loop of main() or threads. It will run lv_timer_handler()
+ * with a given period in ms.
+ * This macro is used to simplify the porting.
+ * @param __ms the period for running lv_timer_handler()
+ */
+#define lv_run_timer_handler_in_period(__ms)                                    \
```
kisvegabor: I think using a `static inline` function would be better as it's easier to debug and interpret compiler errors.

kisvegabor: What about this name?
```suggestion
#define lv_timer_handler_run_in_period(__ms) 
```

kisvegabor (commented):
Thanks. It's really helpful.

The only issue I see is this version will cause 100% CPU usage on a core as there is no sleep in it.

src/misc/lv_timer.h
GorgonMeducer: Better!!!

GorgonMeducer:
> Thanks. It's really helpful.
> 
> The only issue I see is this version will cause 100% CPU usage on a core as there is no sleep in it.

There is no difference from the original flag + ISR solution in bare-metal env, right?

embeddedt:
I think you may have misunderstood that part of the documentation. I don't think it's referring to calling `lv_timer_handler` based on an ISR tick, but rather to calling an LVGL API (e.g. `lv_label_set_text`) from an interrupt handler, which is not safe. Instead, you'd set a flag and then call `lv_label_set_text` before the next invocation of `lv_timer_handler`.

E.g....

```c
static char * new_text = NULL;
void example_isr(void)
{
    new_text = "test";
}

void main(void)
{
    while(1) {
        if(new_text != NULL) {
            lv_label_set_text(label, new_text);
            new_text = NULL;
        }
        lv_timer_handler();
        usleep(5*1000);
    }
}
```

GorgonMeducer:
> I think you may have misunderstood that part of the documentation. I don't think it's referring to calling `lv_timer_handler` based on an ISR tick, but rather to calling an LVGL API (e.g. `lv_label_set_text`) from an interrupt handler, which is not safe. Instead, you'd set a flag and then call `lv_label_set_text` before the next invocation of `lv_timer_handler`.
> 
> E.g....
> 
> ```c
> static char * new_text = NULL;
> void example_isr(void)
> {
>     new_text = "test";
> }
> 
> void main(void)
> {
>     while(1) {
>         if(new_text != NULL) {
>             lv_label_set_text(label, new_text);
>             new_text = NULL;
>         }
>         lv_timer_handler();
>         usleep(5*1000);
>     }
> }
> ```

I don't think I misunderstood the document. 

https://github.com/lvgl/lvgl/blob/master/docs/porting/task-handler.md

image

This is what I tried to improve. 

In bare-metal env, a good practice is something similar:

```c
void SysTick_Handler(void)
{
    //! every 5ms 
    s_bTimerEvent = true;
    
    
    /*! \note please do not put following code here
     *!
     *!  lv_tick_inc(5);
     *!  
     *!     Use a custom tick source that tells the elapsed time in milliseconds.
     *!     It removes the need to manually update the tick with `lv_tick_inc()`)
     *!     #define LV_TICK_CUSTOM 1
     *!     #if LV_TICK_CUSTOM
     *!         extern uint32_t SystemCoreClock;
     *!         #define LV_TICK_CUSTOM_INCLUDE "perf_counter.h"  
     *!         #define LV_TICK_CUSTOM_SYS_TIME_EXPR                            \
     *!                        (get_system_ticks() / (SystemCoreClock / 1000ul))
     *!     #endif
     */
}



    while(1) {
        
        bool bFlag;
        
        __IRQ_SAFE {
            bFlag = s_bTimerEvent;
            s_bTimerEvent = false;
        }
        
        if (bFlag) {
            lv_timer_handler();
        }
        
    }
```

GorgonMeducer:
@embeddedt I have updated the doc to avoid confusion. 
@kisvegabor rename and static inline.

kisvegabor:
This part of the docs
> It's a better approach to simply set a flag or some value in the interrupt, and periodically check it in an LVGL timer (which is run by `lv_timer_handler`).  

refers to case like this:
```c
void uart_isr(void)
{
   char c = read_uart();
   int v = parse_msg(c);

   lv_label_set_text_fmt(label1, "%d", v); //BAD
}
```

```c
int v = 0;
bool v_updated = false;
void uart_isr(void)
{
   char c = read_uart();
   v = parse_msg(c);
   v_updated = true;
}

void my_timer(lv_timer_t * t)
{
   if(v_updated) {
     disable_isr();
     v_updated = false;
     lv_label_set_text_fmt(label1, "%d", v);
     enable_isr();
}
```

GorgonMeducer:
@kisvegabor YES, I get it. And I have already updated the document:

image

It is a separate paragraph in the Interrupt section. 

Are there anything else I have to update?

kisvegabor:
Sorry my inactivity. We were deeply into the last minute issues with releasing [SquareLine](https://forum.lvgl.io/t/edgeline-khm-squreline-is-here/8134/4).

So, I think all we need to do is updating [this site](https://docs.lvgl.io/master/porting/task-handler.html) in the docs as this change has nothing to do with interrupts. It's just a simpler way to call `lv_timer_handler`.

GorgonMeducer:
@kisvegabor  no worry. 
I have withdrawn the changes about os.md.  Please review it.

kisvegabor:
Looks good to me, thank you for the update. 

The failing CI is unrelated and already fixed in master.

GorgonMeducer:
Thank you! @kisvegabor

PR #3081, opened 2022-02-08
feat(disp): add screen out animations

Viatorus:

Add additional exit/out screen animations.

See: https://forum.lvgl.io/t/screen-exit-transitions/8130

example

### Checkpoints


src/core/lv_disp.c
```diff
@@ -328,11 +330,30 @@ void lv_scr_load_anim(lv_obj_t * new_scr, lv_scr_load_anim_t anim_type, uint32_t
             lv_anim_set_exec_cb(&a_old, set_y_anim);
             lv_anim_set_values(&a_old, 0, lv_disp_get_ver_res(d));
             break;
-
         case LV_SCR_LOAD_ANIM_FADE_ON:
```
kisvegabor: For consistency let's use `LV_SCR_LOAD_ANIM_FADE_IN` and add 
```c
LV_SCR_LOAD_ANIM_FADE_ON = LV_SCR_LOAD_ANIM_FADE_IN, /*For backward compatibility*/
```

src/core/lv_indev.c
```diff
@@ -786,14 +786,24 @@ static void indev_proc_press(_lv_indev_proc_t * proc)
...
+    bool indev_obj_in_prev_screen = false;
     bool new_obj_searched = false;
 
     /*If there is no last object then search*/
     if(indev_obj_act == NULL) {
         indev_obj_act = lv_indev_search_obj(lv_disp_get_layer_sys(disp), &proc->types.pointer.act_point);
         if(indev_obj_act == NULL) indev_obj_act = lv_indev_search_obj(lv_disp_get_layer_top(disp),
                                                                           &proc->types.pointer.act_point);
-        if(indev_obj_act == NULL) indev_obj_act = lv_indev_search_obj(lv_disp_get_scr_act(disp),
+        if (indev_obj_act == NULL && consider_prev_screen) {
```
kisvegabor: I'm not sure about this but I think it's enough to search only on `scr_act`. `scr_act` is changed when the animation starts so the checks on `prev_scr` apply only during animation itself which is typically fast. 

So unless I miss something, I  for the sake of simplicity I'd check only `scr_act`

kisvegabor (commented):
It's really a reasonable idea. I added a few comments.

src/core/lv_indev.c
Viatorus: So what I tested so far:
If you click while a screen is flying away (and prev screen js drawn above act screen) the touch event will reach the not visible act screen which feels akward.

kisvegabor: In [this](https://forum.lvgl.io/t/a-bug-and-a-hack-fix/8280/5?u=kisvegabor) Forum topic I wrote about disabling input device during screen transitions. I think it'd result in the least confusion. What do you think?

Viatorus: I am fine with such solution. I reverted any code from lv_indev.c.

Would be good to have a function to globally disable input devices, not just on screen transition, for the user available.

I don't see this as part of this PR.

src/core/lv_disp.c
Viatorus: Added.

Viatorus:
Sorry for the late reply. I tried to resolve your findings.

src/core/lv_indev.c
kisvegabor: `lv_indev_enable()` is available to enable/disable an input device. However, I think it's not a good idea to use it here. It's because if a user uses  `lv_indev_enable()` and we also use it LVGL internally, the calls can overwrite each other.  

IMO all we need to do is adding a `return` [here](https://github.com/lvgl/lvgl/blob/master/src/core/lv_indev.c#L78) if there if `prev_scr != NULL`

Viatorus: Well, that sounds right. Added.

kisvegabor:
Great, thank you very much!

PR #3082, opened 2022-02-08
fix(conf) work around GCC bug

### LINKED ISSUE #3078: GCC bug 80753 means missing lv_conf.h is silently ignored with LV_CONF_INCLUDE_SIMPLE
- [X ] Read the [FAQ](https://docs.lvgl.io/master/intro/index.html#faq)
- [ X] Check the related part of the [Documentation](https://docs.lvgl.io/)
- [ X] Update lvgl to the latest version
- [X ] Reproduce the issue in a [Simulator](https://docs.lvgl.io/latest/en/html/get-started/pc-simulator.html)

### Describe the bug

https://gcc.gnu.org/bugzilla/show_bug.cgi?id=80753 causes the operations performed in `lv_conf_internal.h` to silently ignore a missing `lv_conf.h`. It uses `__has_include("lv_conf.h")` and then `#include "lv_conf.h"` even if that fails, and GCC will silently ignore the later `#include` not finding a file.

I ran into this due to path confusion in my build system that placed my `lv_conf.h` in the wrong place. Due to this behavior, it took a long time to figure out what was happening.

### To Reproduce

1. Configure with `-DLV_CONF_INCLUDE_SIMPLE=1` and GCC.
2. `make` (without creating a lv_conf.h)

### Expected behavior

Fails to build because the file is missing.

embeddedt:
Fixes #3078

### Checkpoints

### DIFF
--- a/scripts/lv_conf_internal_gen.py
+++ b/scripts/lv_conf_internal_gen.py
@@ -57,6 +57,11 @@ fout.write(
     #else
         #include "../../lv_conf.h"                /*Else assume lv_conf.h is next to the lvgl folder*/
     #endif
+    #if !defined(LV_CONF_H) && !defined(LV_CONF_SUPPRESS_DEFINE_CHECK)
+        /* #include will sometimes silently fail when __has_include is used */
+        /* https://gcc.gnu.org/bugzilla/show_bug.cgi?id=80753 */
+        #pragma message("Possible failure to include lv_conf.h, please read the comment in this file if you get errors")
+    #endif
 #endif
 
 #ifdef CONFIG_LV_COLOR_DEPTH
--- a/src/lv_conf_internal.h
+++ b/src/lv_conf_internal.h
@@ -40,6 +40,11 @@
     #else
         #include "../../lv_conf.h"                /*Else assume lv_conf.h is next to the lvgl folder*/
     #endif
+    #if !defined(LV_CONF_H) && !defined(LV_CONF_SUPPRESS_DEFINE_CHECK)
+        /* #include will sometimes silently fail when __has_include is used */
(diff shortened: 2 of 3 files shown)


bsilver8192 (approved):
LGTM

scripts/lv_conf_internal_gen.py
```diff
@@ -57,6 +57,11 @@
     #else
         #include "../../lv_conf.h"                /*Else assume lv_conf.h is next to the lvgl folder*/
     #endif
+    #if !defined(LV_CONF_H) && !defined(LV_CONF_SUPPRESS_DEFINE_CHECK)
+        /* #include will sometimes silently fail when __has_include is used */
+        /* https://gcc.gnu.org/bugzilla/show_bug.cgi?id=80753 */
+        #pragma message("Possible failure to include lv_conf.h, please read the comment in this file if you get errors")
```
kisvegabor, line 63:
```suggestion
        #warning "Possible failure to include lv_conf.h, please read the comment in this file if you get errors"
```

We already used `#warning` on a lot of places and there were no issues. However, we haven't used `#pragma message` so far and it seems it has no advantage over `#warning`.

embeddedt: The issue is that `-Wpedantic` gives an error that `#warning` is a GCC extension (apparently only `#error` is in the standard?) and this caused the tests to fail before I defined `LV_CONF_SUPPRESS_DEFINE_CHECK`.

kisvegabor: Oh,  I always thought `#warning` is standard. 

I tested `#pragma message` in XC16 (compiler fir 16 bit PICs) and it worked. So I think we can merge it.

PR #3084, opened 2022-02-09
feat(log): improve log

GorgonMeducer:

Problems:

* The original `LV_LOG_TRACE/INFO/WARN/ERROR/USER(text)` always attach information to `text`.
* The original function `lv_log()` is too simple and almost no one uses it (inside the LVGL).
* The official benchmark prints test results in a list on the target screen. 
    - If a user hasn't yet port an input method for the platform, it is impossible to check the results.
    - If the framerate is too low, it is painful to check the result with a touch screen. 

Solution:

* improve the log capability 
    - enhance the `lv_log()` function and make it printf-like
    - the new `lv_log()` should only output strings without attaching other information, e.g. \_\_func\_\_, \_\_LINE\_\_ etc.
    - Add a new macro `LV_LOG()` as a wrapper for `lv_log()`
* Output benchmark results in csv format via `LV_LOG()` 
    - The output can be used directly in EXCEL. This is very friendly for performance analysis and optimisation. 

image

### Checkpoints


kisvegabor:
These change seem useful. Thank you.

PR #3092, opened 2022-02-15
Add initial pre-commit configuration for code formatting

### LINKED ISSUE #3083: [meta] Use pre-commit to run astyle automatically on staged files when commiting
### Introduce the problem
I tend to forget about formatting my code before pushing it to the remotes, then CI will fail and I will end up having multiple automatic emails complaining about it. I've also pushed commits with test cases that doesn't pass.

### Examples and cases
Any of my PR 😄 

### Suggested solution

I've found a 'tool' named [pre-commit](https://pre-commit.com/#install) in [this Interrupt post](https://interrupt.memfault.com/blog/pre-commit) and I've set it up locally, I'm pretty pleased with how it works, it also only runs astyle only on the staged files, not in all of the files.

It's worth mentioning that we can still keep the code-format scripts. 

I've recorded a explanation video, please forgive the record setup, let me know if you have additional questions https://youtu.be/Utp9GEPjJyU

C47D:

Add pre-commit configuration for code formatting. Fixes #3083 

### Checkpoints


.pre-commit-config.yaml
```diff
@@ -0,0 +1,30 @@
...
+    hooks:
+    -   id: trailing-whitespace
+
+-   repo: local
+    hooks:
+    # Run astyle over the staged files with c and h extension found in the directories
+    # listed in the files regex pattern. Ignoring the files in the exclude pattern.
+    - id: format-source
+      name: Formatting source files
+      entry: astyle --options=scripts/code-format-per-file.cfg
```
embeddedt: Is there a reason that we can't just use `scripts/code-format.cfg`? It would be nice not to have multiple configs with the same information. I was able to unify the config file for the tests by adding `--ignore-exclude-errors` to the options and putting all the exclude cases together.

C47D: Let me try adding the `--ignore-exclude-errors` to the entry command and I will report back, you're right about only having one configuration file.

C47D: I'm getting the following error, the pre-commit hook gets run on every staged file (instead of in all the files) so removing the `--recursive` option in the configuration (like in the new file) seems to be one of the possible options to don't get this error.

```console
Formatting source files..................................................Failed
- hook id: format-source
- duration: 0.02s
- exit code: 1

Recursive option with no wildcard
Did you intend quote the filename
Artistic Style has terminated
```

kisvegabor: We can move the `--recursive` option to `code-format.py` like this:
```c
os.system('astyle --options=code-format.cfg "../src/*.c,*.h" --recursive')
```

C47D: See 5b336a9 and 2deb141, do they look ok?

embeddedt (approved):

C47D:
Should we add the [pre-commit badge](https://pre-commit.com/#badging-your-repository) to the README?

embeddedt:
I would say yes, but we don't have any other badges at the moment, so it might look out of place on its own.

C47D:
Ok, there's no need to have it neither. I will wait for additional feedback

embeddedt:
If @kisvegabor agrees I think we can merge it as-is.

kisvegabor:
LGTM! Thank you, Carlos!

PR #3101, opened 2022-02-17
Get key code from the actual event

FoundationKen:
Currently, if you queue up a key to send to the gridnav at a later time (e.g., because you are performing an animation from a previous key press), by the time you go to send the event, the indev has lost its context and `lv_indev_get_act()` returns `NULL` and the key is not returned, so gridnav does nothing.

This change gets the key code directly from the event, so that it doesn't matter when it is received.

It will also enable programmatic key-driven automation, which we intend to use for some automated behavior while doing EMI testing.

### Checkpoints
- [x] Update the documentation (no change required)


kisvegabor:
It makes a lot of sense. Thank you for the update!

PR #3102, opened 2022-02-17
refactor(style) move flag information into a dedicated table

embeddedt:

Working towards an implementation of the `inherit` and `initial` property values, the first step is to free up the 5 bits in the property ID to be used for per-style storage instead of storing behavioral data about each property.

The diff is not too massive, just don't look at the autogenerated files.

The `_FILTERED` properties were removed entirely and replaced by autogenerated functions which call `apply_color_filter` manually.

### Checkpoints


embeddedt:
The next step is to decide where to put the lookup table data for properties registered at runtime. Currently I've just assumed there are never more than 256 properties and kept it all in one table.

src/misc/lv_style.c
```diff
@@ -25,6 +25,101 @@
  *  GLOBAL VARIABLES
  **********************/
 
+uint8_t lv_style_lookup_table[256] = {
+    [LV_STYLE_WIDTH] =                    LV_STYLE_PROP_LAYOUT_REFR,
```
kisvegabor, line 30: I've never used this syntax but seems working in XC16. Tested with:
```c
int a[10] = {
    [2] = 20,
    [4] = 40,
    [3] = 30,
};
```

src/misc/lv_style.c
```diff
@@ -25,6 +25,101 @@
  *  GLOBAL VARIABLES
  **********************/
 
+uint8_t lv_style_lookup_table[256] = {
```
kisvegabor: What about making it `const` and using an other table in RAM (dynamically allocated?) for custom properties?

src/misc/lv_style.h
```diff
@@ -155,120 +154,110 @@ typedef union {
  * Enumeration of all built in style properties
  */
 typedef enum {
-    LV_STYLE_PROP_INV                = 0,
+    LV_STYLE_PROP_INV               ,
 
     /*Group 0*/
-    LV_STYLE_WIDTH                   = 1 | LV_STYLE_PROP_LAYOUT_REFR,
```
kisvegabor, line 161: At least the each groups' first index should be kept to form groups. 

The idea is to group similar properties (e.g. `text_...`) so [get_prop_core()](https://github.com/lvgl/lvgl/blob/master/src/core/lv_obj_style.c#L555) can skip styles that doesn't have properties related to the one being searched.

src/misc/lv_style.h
```diff
@@ -499,6 +488,14 @@ static inline void lv_style_set_pad_gap(lv_style_t * style, lv_coord_t value)
     lv_style_set_pad_column(style, value);
 }
 
+static inline uint8_t lv_style_lookup_flags(lv_style_prop_t prop)
```
kisvegabor: What about this API?
```c
lv_style_prop_has_flag(prop, LV_STYLE_PROP_PARENT_LAYOUT_REFR)
```

kisvegabor (commented):
Looks great! I really like the idea of generating filtered styles. Setting filtered styles didn't make any sense anyway.

src/misc/lv_style.c
embeddedt: Updated.

src/misc/lv_style.h
embeddedt: True. I also put the indexes back so that it's easy to see when the group overflows.

src/misc/lv_style.h
embeddedt: It's definitely a better API since it doesn't expose the semantics of exactly how the flags are stored. Updated.

embeddedt:
Alright, I've cleaned this up and it should be merge-quality code now. I'll implement `inherit` and `initial` in a separate PR; this diff is large enough as it is.

src/core/lv_obj_style.c
```diff
@@ -183,19 +186,19 @@ void lv_obj_refresh_style(lv_obj_t * obj, lv_style_selector_t selector, lv_style
...
-    if(prop == LV_STYLE_PROP_ANY || (prop & LV_STYLE_PROP_EXT_DRAW)) {
+    if(prop == LV_STYLE_PROP_ANY || is_ext_draw) {
         lv_obj_refresh_ext_draw_size(obj);
     }
     lv_obj_invalidate(obj);
 
-    if(prop == LV_STYLE_PROP_ANY ||
-       ((prop & LV_STYLE_PROP_INHERIT) && ((prop & LV_STYLE_PROP_EXT_DRAW) || (prop & LV_STYLE_PROP_LAYOUT_REFR)))) {
+
+
```
```suggestion

```

src/extra/layouts/flex/lv_flex.c
```diff
@@ -93,10 +93,10 @@ void lv_flex_init(void)
 {
     LV_LAYOUT_FLEX = lv_layout_register(flex_update, NULL);
 
-    LV_STYLE_FLEX_FLOW = lv_style_register_prop();
-    LV_STYLE_FLEX_MAIN_PLACE = lv_style_register_prop() | LV_STYLE_PROP_LAYOUT_REFR;
-    LV_STYLE_FLEX_CROSS_PLACE = lv_style_register_prop() | LV_STYLE_PROP_LAYOUT_REFR;
-    LV_STYLE_FLEX_TRACK_PLACE = lv_style_register_prop() | LV_STYLE_PROP_LAYOUT_REFR;
+    LV_STYLE_FLEX_FLOW = lv_style_register_prop(0);
```
kisvegabor:
```suggestion
    LV_STYLE_FLEX_FLOW = lv_style_register_prop(LV_STYLE_PROP_NONE);
```

I'm suggesting because `LV_STYLE_PROP_NONE` is more descriptive than `0`.

src/misc/lv_style.c
```diff
@@ -25,10 +25,109 @@
...
+    [LV_STYLE_BASE_DIR] =                  LV_STYLE_PROP_INHERIT | LV_STYLE_PROP_LAYOUT_REFR,
+};
+
+uint8_t * lv_style_custom_prop_flag_lookup_table = NULL;
+
 /**********************
  *  STATIC VARIABLES
  **********************/
 
+static uint16_t act_custom_prop_id = (uint16_t)_LV_STYLE_LAST_BUILT_IN_PROP;
```
kisvegabor:
```suggestion
static uint16_t last_custom_prop_id = (uint16_t)_LV_STYLE_LAST_BUILT_IN_PROP;
```

src/misc/lv_style.c
```diff
@@ -67,11 +166,36 @@ void lv_style_reset(lv_style_t * style)
...
 }
 
-lv_style_prop_t lv_style_register_prop(void)
+lv_style_prop_t lv_style_register_prop(uint8_t flag)
+{
+    /*
+     * Allocate the lookup table if it's not yet available.
+     */
+    if(lv_style_custom_prop_flag_lookup_table == NULL) {
+        lv_style_custom_prop_flag_lookup_table = lv_mem_alloc(LV_STYLE_MAX_CUSTOM_PROPS);
```
kisvegabor: Why not `realloc` instead of using a fixed `LV_STYLE_MAX_CUSTOM_PROPS` size?

src/misc/lv_style.c
```diff
@@ -25,10 +25,109 @@
...
+    [LV_STYLE_COLOR_FILTER_OPA] = 0,
+    [LV_STYLE_ANIM_TIME] = 0,
+    [LV_STYLE_ANIM_SPEED] = 0,
+    [LV_STYLE_TRANSITION] = 0,
+    [LV_STYLE_BLEND_MODE] = 0,
+    [LV_STYLE_LAYOUT] =                    LV_STYLE_PROP_LAYOUT_REFR,
+    [LV_STYLE_BASE_DIR] =                  LV_STYLE_PROP_INHERIT | LV_STYLE_PROP_LAYOUT_REFR,
+};
+
+uint8_t * lv_style_custom_prop_flag_lookup_table = NULL;
```
kisvegabor:
```suggestion
uint8_t * _lv_style_custom_prop_flag_lookup_table = NULL;
```

src/misc/lv_style.c
embeddedt: Even with 256 custom properties, this would only be 256 bytes (and the default is 64). In my opinion it will just add a greater risk of fragmentation for not much gain. We allocate a lot more than 256 bytes in other places.

kisvegabor: I see 2 advantages of `reallocing`:
- There is no max prop. count -> one less issue, and apps can scale better 
- One less config option

If fragmentation is a concern what about realloc-ing in 16 or 32 byte blocks?

embeddedt: > There is no max prop. count -> one less issue, and apps can scale better

Fair enough, I'll implement it that way, then.

embeddedt: Implemented.

kisvegabor:
Looks great! Support for `inherit` and `initial` will be added in a next step?

embeddedt:
That's the plan, yes! 🙂

embeddedt:
It seems that the Micropython CI failed, probably because the dynamically allocated lookup table needs to be a GC root. I'll look into it.

kisvegabor:
> It seems that the Micropython CI failed, probably because the dynamically allocated lookup table needs to be a GC root. 

Yes, it should be. It's great the CI spotted this issue.

embeddedt:
@amirgon It seems that [this change](https://github.com/lvgl/lvgl/pull/3102/commits/47355f7b4c1bdf6ebc2bfe6c622c9523346971c6#diff-17c322664695d0bd417a012e00f301dabf1713991a53de0c15cb50ba9f1c239eR19) breaks CI. Is it illegal to include `lv_gc.h` here?

amirgon:
> It seems that [this change](https://github.com/lvgl/lvgl/pull/3102/commits/47355f7b4c1bdf6ebc2bfe6c622c9523346971c6#diff-17c322664695d0bd417a012e00f301dabf1713991a53de0c15cb50ba9f1c239eR19) breaks CI. Is it illegal to include `lv_gc.h` here?

I think that including gc in a public header file causes a cyclic dependency.  
Can you move these inline functions that need LV_GC_ROOT into the C file?

PR #3104, opened 2022-02-18
Unit tests for line widget

C47D:

Unit tests for line widget

### Checkpoints

### DIFF
--- a/tests/src/test_cases/test_line.c
+++ b/tests/src/test_cases/test_line.c
@@ -48,7 +48,7 @@ void test_line_size_should_be_updated_after_adding_points(void)
 
     /* Get the biggest coordinate on both axis */
     uint16_t point_idx = 0;
-    for (point_idx = 0; point_idx < point_cnt; point_idx++) {
+    for(point_idx = 0; point_idx < point_cnt; point_idx++) {
         calculated_width = LV_MAX(points[point_idx].x, calculated_width);
         calculated_height = LV_MAX(points[point_idx].y, calculated_height);
     }
--- a/tests/src/test_cases/test_line.c
+++ b/tests/src/test_cases/test_line.c
@@ -16,7 +16,7 @@ void setUp(void)
 
 void tearDown(void)
 {
-    /* Function run after every test */
+    lv_obj_clean(active_screen);
 }
 
 void test_line_should_have_valid_documented_default_values(void)
--- a/tests/src/test_cases/test_line.c
+++ b/tests/src/test_cases/test_line.c
@@ -6,6 +6,8 @@
 static lv_obj_t * active_screen = NULL;
 static lv_obj_t * line = NULL;
 
+static const uint16_t default_point_num = 0U;
+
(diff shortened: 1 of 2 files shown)


C47D:
I don't see any other behavior to test, maybe adding a rendering test would be useful?

tests/src/test_cases/test_line.c
```diff
@@ -0,0 +1,37 @@
...
+
+void setUp(void)
+{
+    active_screen = lv_scr_act();
+    line = lv_line_create(active_screen);
+}
+
+void tearDown(void)
+{
+    /* Function run after every test */
```
kisvegabor: Adding `lv_obj_clean(lv_scr_act())`?

kisvegabor (commented):
I can imagine two more behavior test: 
1. by default the size should be set according to [this](https://github.com/lvgl/lvgl/blob/master/src/widgets/lv_line.c#L152-L153)
2. the extra draw size should be set according to [this](https://github.com/lvgl/lvgl/blob/master/src/widgets/lv_line.c#L137).

C47D:
For extra draw size I should use `lv_obj_refresh_ext_draw_size` on the line object, right? Or how can we test the calculation?

kisvegabor:
`_lv_obj_get_ext_draw_size(obj);` should tell the current ext draw size.

C47D:
@kisvegabor on the bd4a8fc (test for extra draw size) there's something missing, as far as I understand, when the line style line width is greater than the received value from the event, the extra draw size should be set to the style line width. But I'm getting the initial extra draw size (5) and were expecting (10).

tests/src/test_cases/test_line.c
```diff
@@ -0,0 +1,98 @@
...
+    lv_obj_add_event_cb(line, line_event_cb, LV_EVENT_ALL, NULL);
+    /* Trigger the extra draw size event */
+    lv_obj_refresh_ext_draw_size(line);
+
+    TEST_ASSERT_EQUAL(1U, _lv_obj_get_ext_draw_size(line));
+
+    /* Update line width style */
+    static lv_style_t line_style;
+    lv_style_init(&line_style);
+    lv_obj_set_style_line_width(line, LV_PART_MAIN, final_extra_draw_size);
```
kisvegabor:
```suggestion
    lv_obj_set_style_line_width(line, final_extra_draw_size, LV_PART_MAIN);
```

kisvegabor (commented):
I found the issue, it was only tricky typo.

`line_style` related parts are not required, right?

tests/src/test_cases/test_line.c
C47D: Good catch, fixed in 8c14fbe

C47D:
@kisvegabor Thanks for the help, now all tests are passing. Ready to be merged.

kisvegabor:
Aaand merged, thank you again! :slightly_smiling_face:

PR #3113, opened 2022-02-21
fix(misc): add support of maximum value less than minimum value

xaowang96:

The linear mapping function lv_map should include the case where the maximum value of the input value is less than the minimum. Therefore, i submitted this patch to support it.

### Checkpoints


kisvegabor:
Thanks!

PR #3118, opened 2022-02-22
slider: tests and cleanup

C47D:

Initial unit tests for slider and a bit of cleanup/refactoring

### Checkpoints


src/widgets/lv_slider.c
```diff
@@ -330,30 +330,20 @@ static void draw_knob(lv_event_t * e)
...
+    bool is_rtl = base_dir == LV_BASE_DIR_RTL;
 
+    lv_area_t knob_area;
     /*Horizontal*/
-    if(hor) {
-        if(!sym) {
-            knob_area.x1 = LV_SLIDER_KNOB_COORD(hor, base_dir == LV_BASE_DIR_RTL, slider->bar.indic_area);
-        }
-        else {
-            if(slider->bar.cur_value >= 0) {
```
kisvegabor, line 350: Where is the `sym == true` case handled now?

C47D: This is the previous code:

```c
        if(!sym) {
            knob_area.x1 = LV_SLIDER_KNOB_COORD(hor, base_dir == LV_BASE_DIR_RTL, slider->bar.indic_area);
        }
        else {
            if(slider->bar.cur_value >= 0) {
                knob_area.x1 = LV_SLIDER_KNOB_COORD(hor, base_dir == LV_BASE_DIR_RTL, slider->bar.indic_area);
            }
            else {
                knob_area.x1 = LV_SLIDER_KNOB_COORD(hor, base_dir != LV_BASE_DIR_RTL, slider->bar.indic_area);
            }
        }
```

All the cases is the same assignment:
```c
knob_area.x1 = LV_SLIDER_KNOB_COORD(hor, base_dir == LV_BASE_DIR_RTL, slider->bar.indic_area);
knob_area.x1 = LV_SLIDER_KNOB_COORD(hor, base_dir == LV_BASE_DIR_RTL, slider->bar.indic_area);
knob_area.x1 = LV_SLIDER_KNOB_COORD(hor, base_dir != LV_BASE_DIR_RTL, slider->bar.indic_area);
```

The only thing that changed is the RTL check and it's now using a variable.

C47D:
Hi, some questions,

Is there any helper to do this? I think this is keeping `new_value` within `real_min_value` and `real_max_value` boundaries

```c
        if(new_value < real_min_value) new_value = real_min_value;
        else if(new_value > real_max_value) new_value = real_max_value;
```
---

In `lv_slider.c` `draw_knob` there's a check to know if the 'underlying' bar is symmetrical
```c
bool sym = false;
    if(slider->bar.mode == LV_BAR_MODE_SYMMETRICAL && slider->bar.min_value < 0 && slider->bar.max_value > 0) sym = true;
```

There's a similar check in `lv_bar.c` `draw_indic`:
```c
    bool sym = false;
    if(bar->mode == LV_BAR_MODE_SYMMETRICAL && bar->min_value < 0 && bar->max_value > 0 &&
       bar->start_value == bar->min_value) sym = true;
```

Would adding that check in an static inlined function in `lv_bar.h` make sense? We could then can use it in `lv_slider.c`, and other widgets that use a bar underneath.

src/widgets/lv_slider.c
kisvegabor: Ah, you are right! I missed some parts which are actually handles by `lv_bar`.

kisvegabor:
> Is there any helper to do this? I think this is keeping new_value within real_min_value and real_max_value boundaries

Yes, `x = LV_CLAPM(min, x, max);`

> In lv_slider.c draw_knob there's a check to know if the 'underlying' bar is symmetrical


As you pointed out [here](https://github.com/lvgl/lvgl/pull/3118/files#r811939605) `sym` is not required in `lv_slider` (It seems it can be removed from the vertical case too).

C47D:
> As you pointed out [here](https://github.com/lvgl/lvgl/pull/3118/files#r811939605) sym is not required in lv_slider (It seems it can be removed from the vertical case too).

`sym` is still needed in `draw_knob`

C47D:
Hi @kisvegabor In `test_slider_range_mode_should_not_leave_edit_mode_if_released_with_no_left_knob_focus` there's something changing the value of `slider->left_knob_focus` to `1` between the explicit set to `0` and the event handler, so the assertion `TEST_ASSERT_EQUAL(1U, ptr->left_knob_focus);` fails.

I was expecting to hit `if(slider->left_knob_focus == 0) slider->left_knob_focus = 1;` in the code handler.

Any idea what might be changing it?

---
imagen

Now that we have test for this handler it seems almost ready for cleanup

kisvegabor:
I can test it only on Monday... because I can't find the mouse of my notebook... :smile: 

But looking into the code `LV_EVENT_PRESSED` seems suspicious as there is no `if` for indev types.

Anyway, I'll check it on Monday.

C47D:
@kisvegabor Will add notes to the failing asserts and make them pass as the reason of failure is not very evident.
I'm adding a test for `LV_EVENT_HIT_TEST` event, how can the `right_knob_area` and `left_knob_area` be controlled? I guess they represent the knobs on the slider and they got drawn in `draw_knob`, is this right?

@embeddedt What's the preferred way to get this branch up to date with master? by rebasing?

kisvegabor:
I wonder if it's possible to test on the API level without go into the internals. E.g. `ptr->value_to_set` and `ptr->left_knob_focus`.

For example instead of `test_slider_normal_mode_should_leave_edit_mode_if_released`:
- go to edit mode
- `lv_test_encoder_add_diff()` -> `lv_slider_get_value` 
- click the encoder
- `lv_test_encoder_add_diff()` -> `lv_slider_get_left_value` 

What do you think?

C47D:
> I wonder if it's possible to test on the API level without go into the internals. E.g. ptr->value_to_set and ptr->left_knob_focus.

Sorry, I didn't saw that APIs, will update the tests with them 😸 

What does `lv_test_encoder_add_diff` do?

kisvegabor:
> What does lv_test_encoder_add_diff do?

E.g. `lv_test_encoder_add_diff(3)` emulates 3 encoder rotation steps to the right and `lv_test_encoder_add_diff(-3)` 3 steps to the left.

C47D:
Something like this? If so, it's much better, I had to use `lv_test_encoder_turn` instead of `lv_test_encoder_add_diff`.
Maybe we don't need the final assert, what do you think?

```c
void test_slider_normal_mode_should_leave_edit_mode_if_released(void)
{
    int32_t rotations_right = 5U;
    int32_t rotations_left = 3U;
    int32_t expected_final_value = rotations_right - rotations_left;

    /* Setup group and encoder indev */
    lv_group_t * g = lv_group_create();
    lv_indev_set_group(lv_test_encoder_indev, g);
    lv_group_add_obj(g, sliderNormalMode);

    /* Enter edit mode */
    lv_test_encoder_click();

    lv_test_encoder_turn(rotations_right);
    TEST_ASSERT_EQUAL(rotations_right, lv_slider_get_value(sliderNormalMode));

    lv_test_encoder_turn(-rotations_left);
    TEST_ASSERT_EQUAL(expected_final_value, lv_slider_get_value(sliderNormalMode));

    /* Leave edit mode */
    lv_test_encoder_click();

    /* Group leaved edit mode */
    TEST_ASSERT_FALSE(lv_group_get_editing(g));
}
```

I need to get familiar with the different indevs, maybe adding docs to the test simulated indev might be a good exercise, what do you think?

kisvegabor:
> Something like this? If so, it's much better, I had to use lv_test_encoder_turn instead of lv_test_encoder_add_diff.
Maybe we don't need the final assert, what do you think?

Looks great! I meant something like this.

kisvegabor:
I added `test_ranged_mode_adjust_with_encoder` for testing this complex case from the highest level.

In a similar way can we avoid testing `left_knob_focus` directly?

C47D:
I haven't understood what it is used for, we can remove those checks.

C47D:
@kisvegabor Ready to be reviewed

kisvegabor:
It seems somethings is messed up in the commit history and now we have 61 changes files. 

Can you fix it by rebaseing on master and force pushing?

C47D:
Yep, I noticed ~20 new commits in this branch when I start to work on it again.

> Can you fix it by rebaseing on master and force pushing?

Do you mean this? I'm not very familiar with rebasing, that's why I asked before :smile
```commandline
git switch master
git pull upstream master
git switch test/slider
git merge -- rebase master
git push -f origin test/slider
```

---
It seems like it went down to 4 files changed.

kisvegabor:
> Do you mean this? I'm not very familiar with rebasing, that's why I asked before :smile


Yes, I meant that and just tried it but it get even worse :smile: 

Could you resend the PR? Just to be sure we don't merge something unintentionally.

C47D:
Sure, I will send a new PR later today

C47D:
Closed in favor of #3198

Issue #3041, opened 2022-01-24
Image cache and decoder optimization


FASTSHIFT:
### Introduce the problem
* The current image cache uses the number of images to limit how many images can be opened at the same time. This may cause two problems: if there are many low-resolution images, the memory cannot be fully used to reduce the open/close operations of the decoder; If the user uses a higher resolution image, it will easily run out of memory.

*  I have a GPU on my hardware that can accelerate image rendering, but this GPU requires that the stride of the image must be an integer multiple of 16px. At present, my approach is to decode the image and copy the original image into a blank image aligned to a width of 16px, and pass the copied image to lvgl. I have encountered a problem, when using the bin format image, I can customize the decoder to convert the image. But when I wanted to use png decoding, I had to modify the source code of the `lv_png` decoder, insert and call my own image conversion function to convert the image into a new image with a width aligned to 16 pixels. Is there a better solution?

### Suggested solution
* Limit the total memory size occupied by the image cache, rather than the number of open images. The memory occupied by the decoder that is opened each time can provide statistics in two ways:
1 ) lvgl records the image occupancy directly according to the `width * height * (bytes per pixel)` of the image.
2 ) The memory occupied by the custom decoder report itself. 
The default mode can be the first one.

* Is it possible to add a decoder secondary processing interface, which is called after each deocder `open_cb` is completed, to adapt to the situation that the hardware has special requirements for the image size?

kisvegabor:
Hi,

I agree with both things.

For the cache, I think "The memory occupied by the custom decoder report itself." is more reasonable as in some cases the images are read line-by-line or in block so the memory usage is much smaller than width x height.   
#2904 is a huge rework on the decoder interface. I suggest waiting with this feature until it gets merged.

Regarding the stride, so far I thought this issue is related only to the built-in "byte array" format but you are right, decoded PNG images  also can have stride constraints. Maybe I missed something, but it seems your suggestion will require `width x height` RAM for the normal decode plus `(width + padding) x height` for the alignment and then the first `width x height`  can be freed. 

I have no good idea for it, it'd be great to avoid having two image sized array allocated even for a short time.

X-Ryl669:
> If the user uses a higher resolution image, it will easily run out of memory.

Depends on the decoder. Some (most ?) decoder can deal with this case by only rendering line by lines. Some don't. 

[...]

Issue #3116, opened 2022-02-21
Suggestion: Dynamic Global LVGL Cache


amirgon:
Today LVGL maintains a number of caches for different purposes:
- Shadows
- Circle drawing
- Gradient
- Images
- lv_fs_read

Today these caches are allocated statically (on the data segment) and their size is configurable through macros.  
These statically-allocated caches have number of disadvantages:
- The ram is always "locked out" for other purposes. Some applications might have some modes where more ram is temporariliy needed, at the exspense of gui optimizations.
- On some platforms (such as ESP32), heap can be allocated on external RAM (spi-ram) while the main ram is used for DMA etc. Allocating cache buffers on external ram could save precious ram and still be effective since the cpu-cache backs the external ram as well, so on periods of time when the cache buffers are used extensively they can still be very fast even though they are on external ram.
- In case of Micopython or other dynamic bindings, these caches are preconfigured today per firmware instead of being user-configurable at runtime.

Instead of statically allocated caches, it could be useful to allow the user allocate and configure these caches dynamically on runtime in order to provide the user more control over when they are allocated/deallocated and in which size.

Also, to make things simpler to the user, instead of configuring each cache individually, the user could dynamically define a "global" cache according to the available memory on the device and let LVGL manage all caches internally inside this global cache.  
This way, the user could simply say "use more cache now" or "release all cache now" without worrying which cache he should be twicking.  
Advanced users might want to control how this global cache is split between specific LVGL caches (this can be specified, for example, by using a factor for each individual cache type which would control how much it consumes out of the global cache)

kisvegabor:
I like the idea of using run time configurable cache. Something like:
```c
static uint8_t cache[1024];
set_cache(cache, 1024);
```

will have the behavior as now. 

@xiaoxiang781216 proposed to have hook on malloc "Out of memory" that asks the the user to free some memory and try malloc-ing automatically again. The user could simply reduce the cache size in such a hook.

However, if there is an Out of memory error in the middle of e.g. image decoding and we change the cache size (i.e. drop the cache and start with a smaller empty one) some previously allocated data of the current decoding will be lost too. So there should be a way  to restart it in such event somehow.

About the global cache, it's an interesting idea. I still can't how can it be managed in predictable way but let's see what can we figure out. Just as a topic starter: having something like cache priority and set it like:
```c
lv_cache_set_priority(LV_CACHE_TYPE_GRADIENT, LV_CACHE_PRIORITY_HIGHEST);
[...]

PR #2904, opened 2021-12-15
Partial rlottie

X-Ryl669:
## Rlottie as an image decoder. 
Instead of the current rlottie library that's creating a raw image (complete with all pixels), this PR tries to achieve a similar result with a true image decoder (respecting LVGL image decoder format).
This require a modification to rlottie library (see [here](https://github.com/Samsung/rlottie/pull/512)) to support partial rendering.

This presents multiple advantages:
1. Reduced memory consumption (since only a small buffer is used instead of the whole rendered image)
2. Can be used for images too *as usual* (with `lv_img_set_src`) anywhere where image are used
3. Memory is automagically handled by the cache system (so you can have tens of animation and it'll work even if only 3 are actually allocated)
4. More consistent with the LVGL API 

And some cons:
1. ~~Right now, the image size is read from the lottie file. See below for the required discussion.~~
2. ~~When the cache decide to kill a decoder context, we loose the current progression in the animation. See below~~
3. ~~Does not work with animations yet due to the number 2 above.~~

So, instead of starting an issue, I prefer to speak with some code, it's easier to understand where to start from.

I propose to change the way LVGL is dealing with decoder and images.
Right now, there is no link between decoder and the resulting image object (only via the cache system).
We can do a lot of things in the decoder context, but since the cache system is closing the decoder at any time, we are loosing anything we store in this context.
In my rlottie object, I'm querying the cache system to find the decoder context for the given src, but it's being reset each time the cache decided to do that.
So in order to get everything working, I propose to add a link between the image and its decoder. So instead of using "img_data == NULL" as a signal that a decoder should be used, I propose to add a "dec_ctx" member to the picture instead. This would allow user-defined context in an image where we can store metadata like the animation state.

So typically, the rendering function for an image will change from:
```cpp
lv_draw_img_core(src, drawContext) {
    cacheEntry = find_cache_for_image_src(src);
    if (cdsc->dec_dsc.img_data) drawWholeBuffer();
    else callDecoderToDecodeOneLineAndDraw();
```
I'd like to have instead:
```cpp
lv_draw_img_core(src, drawContext, decContext = img->dec_ctx) {
    cacheEntry = find_cache_for_image_src(src, img->dec_ctx);
    if (cdsc->dec_dsc.img_data) drawWholeBuffer();
    else callDecoderToDecodeOneLineAndDraw(img->dec_ctx);
```
[...]

PR #3124, opened 2022-02-24
feat(img_cache): limit the number of caches based on the total memory

FASTSHIFT:
**Do not merge this PR**

Related discussion: https://github.com/lvgl/lvgl/issues/3041

The practice of this PR is a new idea for cache management. The number of caches is limited according to the memory occupied by the picture to prevent the decoder from occupying too much memory.

I tested with `libs/png` decoder and it seems to work fine so far.

### Checkpoints

### DIFF
--- a/src/draw/lv_img_cache.c
+++ b/src/draw/lv_img_cache.c
@@ -35,13 +35,17 @@
  **********************/
 #if LV_IMG_CACHE_DEF_SIZE
     static bool lv_img_cache_match(const void * src1, const void * src2);
+    static _lv_img_cache_entry_t * find_reuse_entry(void);
+    static _lv_img_cache_entry_t * find_empty_entry(void);
 #endif
 
 /**********************
  *  STATIC VARIABLES
  **********************/
 #if LV_IMG_CACHE_DEF_SIZE
-    static uint16_t entry_cnt;
+    static uint32_t cache_total_size;
+    static uint32_t cache_free_size;
+    static uint16_t cache_entry_cnt;
 #endif
 
 /**********************
@@ -66,7 +70,7 @@ _lv_img_cache_entry_t * _lv_img_cache_open(const void * src, lv_color_t color, i
     _lv_img_cache_entry_t * cached_src = NULL;
 
 #if LV_IMG_CACHE_DEF_SIZE
-    if(entry_cnt == 0) {
+    if(cache_entry_cnt == 0 || cache_total_size == 0) {
         LV_LOG_WARN("lv_img_cache_open: the cache size is 0");
         return NULL;
     }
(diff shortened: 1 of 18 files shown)


src/draw/lv_img_cache.c
```diff
@@ -75,58 +79,111 @@ _lv_img_cache_entry_t * _lv_img_cache_open(const void * src, lv_color_t color, i
...
+        LV_LOG_WARN("can't get image info");
+        return NULL;
     }
 
-    /*Close the decoder to reuse if it was opened (has a valid source)*/
-    if(cached_src->dec_dsc.src) {
-        lv_img_decoder_close(&cached_src->dec_dsc);
-        LV_LOG_INFO("image draw: cache miss, close and reuse an entry");
+    /*Calculate the memory usage of an image*/
+    uint32_t img_req_size = lv_img_buf_get_img_size(header.w, header.h, header.cf);
```
kisvegabor: It's not accurate in all cases.  E.g. if a BMP file is ready from a file only a file descriptor is cached.

It'd be better if the decoder could tell the required memory size, similarly to `time_to_open` but before opening the image.

kisvegabor (commented):
Hi,

Thanks for the PR. I like this approach. I could imagine only a very complicated ways to assign cache buffers to images, manage and free them, etc. Your solution (simply saving the used memory and freeing until the target limit is hit)  is very simple and elegant. 

We 2 other PRs/issues about related topics:
- Global cache: https://github.com/lvgl/lvgl/issues/3116
- Rework decoders: https://github.com/lvgl/lvgl/pull/2904#issuecomment-1047616125

I suggest adding the decoder changes first and think about the cache things later.

src/draw/lv_img_cache.c
FASTSHIFT: Yes, I would like to use `lv_img_decoder_get_info()` to get the occupied memory, but I can only get the information described in `lv_img_header_t`, would you like to extend the parameters that `lv_img_decoder_get_info()` can get?

FASTSHIFT:
@kisvegabor 
Does the use of `lv_canvas` lead to wrong cache occupancy?

kisvegabor:
I'm not aware of that. Can you send a code snippet to reproduce the issue?

FASTSHIFT:
> I'm not aware of that. Can you send a code snippet to reproduce the issue?

For the buffer of canvas, can I think that the memory size occupied by the decoder is 0?

kisvegabor:
> For the buffer of canvas, can I think that the memory size occupied by the decoder is 0?

Yes!

FASTSHIFT:
@kisvegabor 
I modified the parameters passed by `lv_img_decoder get_info()` and used `lv_img_decoder_info_t` to expand the passed info (including the memory usage that the decoder needs to report). 
Do you agree with this modification? I can try to complete other decoders.

FASTSHIFT:
@kisvegabor 
Do you think this change is reasonable?

kisvegabor:
Sorry for the late answer.

I think it's good direction, but I also think we should implement the global cache as described #3116. In the global cache implementation we can handle the images cache as you described here.

kisvegabor:
Sorry for the late answer.

I think it's good direction, but I also think we should implement the global cache as described #3116. In the global cache implementation we can handle the images cache as you described here.

PR #3135, opened 2022-02-26
Don't build examples when running unit tests

### LINKED ISSUE #3117: Examples being built when running tests
### Introduce the problem
The examples are being built when running tests, is this necessary? (also the demos are being built)

### Examples and cases
Run the test cases and see the examples being built.

### Suggested solution
Don't build examples when unit tests are being build.

C47D:

@cmumford @embeddedt I was trying to get the examples not being compiled when running the unit tests, I'm not familiar with CMakeLists.txt so I thought by setting `LV_BUILD_EXAMPLES=0` and removing the `target_link_libraries` when compiling the test would be enough, but it seems like it doesn't, maybe I'm missing something?

Purpose of the change is to run tests faster, I'm planning to try run tests individually or passing a list of tests cases to run to the python script. 

Closes #3117 

### Checkpoints


tests/CMakeLists.txt
```diff
@@ -97,7 +97,7 @@ set(LVGL_TEST_OPTIONS_16BIT
 
 set(LVGL_TEST_OPTIONS_16BIT_SWAP
     -DLV_COLOR_DEPTH=16
-    -DLV_COLOR_16_SWAP=1
```
cmumford, line 100: I think it would be cleaner if the whitespace fixes (which are good) landed in a separate commit titled something like "white fixes". This makes `git blame` a bit more helpful.

tests/CMakeLists.txt
```diff
@@ -328,6 +330,7 @@ add_library(test_common
         src/test_fonts/font_3.c
         unity/unity_support.c
         unity/unity.c
+        src/img_caret_down.c
```
cmumford: Did you copy `examples/assets/img_caret_down.c` into `tests/src` and forget to add it to this commit? An alternative is to change this to:

    ${LVGL_DIR}/examples/assets/img_caret_down.c

I'd be in favor of moving the example asset files into their own library on which examples and tests both depend, but that's outside the scope of this change. @embeddedt WDYT about this?

tests/CMakeLists.txt
C47D: I agree, how can I fix it? Reverting this commit and then making two separate commits?

tests/CMakeLists.txt
C47D: Agree, I can try to update the path instead.

tests/CMakeLists.txt
cmumford: I think the easiest is for you to make a second commit with the whitespace changes reverted. The reviewer, when approving your PR, can use [squash and merge](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/incorporating-changes-from-a-pull-request/about-pull-request-merges#squash-and-merge-your-pull-request-commits). Alternatively you could rebase/squash and then do a force push to your branch.

tests/CMakeLists.txt
cmumford: Actually I tried that a little bit later and because your change, by design, doesn't define `LV_BUILD_EXAMPLES` the `img_caret_down`  symbol is not defined. I wasn't testing correctly so I missed that :-(.

tests/CMakeLists.txt
C47D: Sorry, I thought I was sending the changes from #3133 into this commit.

"Problem" is I'm using a git hook that look for trailing whitespaces and fixes them in the staged files just before the commit, so if we want to prevent that I would need to skip that git hook, do the commit and then force the git hook to fix the trailing whitespaces. Or before any change force the git hook on the file I need to work on.

cmumford: Maybe not worth it then. You could do the whitespace fixup in a separate commit and then rebase your branch on master once it lands, but that's up to you. I'm not a project collaborator with PR approve privileges, I'm just a contributor.  Maybe keep the whitespace in this PR and do that only if requested by the collaborator.

tests/CMakeLists.txt
C47D: Hi, thanks for your patience, the dropdown test file uses that particular asset file. I haven't worked on this since last week, will try to continue this week.

C47D:
So the examples are no longer being build when running unit tests, they are being build when running the build tests. Still trying to figure out how to let the unit tests configuration where's the `img_caret_down` asset.

kisvegabor:
I think the best would be to make the tests use their own assets. I.e. we can add a `test/src/test_assets` folder. What do you think?

C47D:
It is ok for me, should the directories `ref_imgs` and `test_fonts` moved to that new directory?

kisvegabor:
`ref_imgs` are not really assets so I'd keep them in separate folder. 

`test_fonts` actually contains `*.fnt` assets and `*.c` reference fonts. 

What about creating an `assets` and a `refs` folder too? (`*.fnt` would go to the `refs` folder)

C47D:
> What about creating an assets and a refs folder too? (*.fnt would go to the refs folder)

Ok, I will do that

C47D:
So this now builds, but when I tried to validate again the examples are not being build when running tests it seems like they are, so, could you test it in your end?

I was checking the `custom.cmake` file and it seems like it adds lv_examples as dependency for LVGL, I may be wrong about this tho.

kisvegabor:
`./main.py test --clean` builds the examples for me too.

C47D:
Will upload a diff I did on custom.cmake later today

PR #3140, opened 2022-02-28
fix(btnmatrix): set btn_id_act on keypad press

LoveMHz:

This fix is to correct the behavior of the keypad interface and bring it up to the same logic as other interfaces (ie pointer).

While using the btnmatrix component and the keypad interface it's possible that the first LV_EVENT_PRESSED event will fire with the incorrect btn_id_act if a call to lv_btnmatrix_set_btn_ctrl() was made beforehand (as set_btn_ctrl sets btn_id_act but not focus).

### Checkpoints


kisvegabor:
Great catch, thanks!

Issue #3087, opened 2022-02-11
STM32 DMA2D and display driver flush


patricklaf:
### Introduce the problem
I am using LVGL with Arduino on STM32. Prior to last update, I used DMA2D in the display driver with _lv_gpu_stm32_dma2d_copy_ function. With latest 8.2.0, this function doesn't exist anymore and all public functions from DMA2D code are based on _lv_draw_ctx_t_ use.

### Examples and cases
Code working in 8.0.2:
```c
void lvgl_disp_flush(lv_disp_drv_t *display, const lv_area_t *area, lv_color_t *color_p) {
#if LV_USE_GPU_STM32_DMA2D
	lv_gpu_stm32_dma2d_copy(
			(lv_color_t*) tft.getBuffer() + area->y1 * LV_HOR_RES + area->x1,
			(lv_coord_t) LV_HOR_RES, color_p, area->x2 - area->x1 + 1,
			area->x2 - area->x1 + 1, area->y2 - area->y1 + 1);
#else
	tft.drawBuffer(area->x1, area->y1, area->x2 - area->x1 + 1, area->y2 - area->y1 + 1, (uint16_t*)color_p);
#endif
	lv_disp_flush_ready(display);
}
```
### Suggested solution
Is it possible to add a public function to GPU drivers to achieve accelerated buffer copy usable from display driver flush function?

kisvegabor:
> Is it possible to add a public function to GPU drivers to achieve accelerated buffer copy usable from display driver flush function?

I agree to add a buffer copy callback to the `draw_ctx` it's really a generic operation and it can be useful for the user.

I've just pushed d034511b in which `draw_ctx->buffer_copy` and `lv_draw_sw_buffer_copy` is added.

Could you send a Pull request which added this function the DMA2D driver too?

patricklaf:
I am currently away from my office and from my board, so I can't test anything for a week.
I will do it when I am back.

I think that it can be something like this in `lv_gpu_stm32_dma2d.c`:
```c
/**********************
 *   GLOBAL FUNCTIONS
 **********************/

void lv_draw_stm32_dma2d_buffer_copy(lv_draw_ctx_t *draw_ctx, void *dest_buf,
		lv_coord_t dest_stride, const lv_area_t *dest_area, void *src_buf,
		lv_coord_t src_stride, const lv_area_t *src_area) {
	LV_UNUSED(draw_ctx);
[...]

PR #3147, opened 2022-03-05
feat(dma2d) add lv_draw_stm32_dma2d_buffer_copy function

patricklaf:

This allow to use DMA2D accelerated hardware from display driver flush callback.
Tested on STM32F746G_DISCO with Arduino core.

Related to issue #3087.

### Example

```cpp
void lvgl_disp_flush(lv_disp_drv_t* display, const lv_area_t* area, lv_color_t* color_p) {
#if LV_USE_GPU_STM32_DMA2D
#if LV_VERSION_CHECK(8,2,0)
	lv_draw_stm32_dma2d_buffer_copy(NULL,
			(lv_color_t *)tft.getBuffer() + area->y1 * LV_HOR_RES + area->x1,
			(lv_coord_t) LV_HOR_RES, area, color_p, area->x2 - area->x1 + 1, area);
#else
	lv_gpu_stm32_dma2d_copy((lv_color_t *)tft.getBuffer() + area->y1 * LV_HOR_RES + area->x1, (lv_coord_t)LV_HOR_RES, color_p, area->x2 - area->x1 + 1, area->x2 - area->x1 + 1, area->y2 - area->y1 + 1);
#endif
#else
	tft.drawBuffer(area->x1, area->y1, , area->y2 - area->y1 + 1, (uint16_t*)color_p);
#endif
	lv_disp_flush_ready(display);
}
```
Note
- `tft.getBuffer()` return the address of the display frame buffer.
- LV_VERSION_CHECK should be set with the version released with these changes.
- `lv_gpu_stm32_dma2d_copy` was working fine before version 8.2.0.

### DIFF
--- a/src/draw/stm32_dma2d/lv_gpu_stm32_dma2d.c
+++ b/src/draw/stm32_dma2d/lv_gpu_stm32_dma2d.c
@@ -147,9 +147,9 @@ void lv_draw_stm32_dma2d_blend(lv_draw_ctx_t * draw_ctx, const lv_draw_sw_blend_
     if(!done) lv_draw_sw_blend_basic(draw_ctx, dsc);
 }
 
-void lv_draw_stm32_dma2d_buffer_copy(lv_draw_ctx_t *draw_ctx,
-                                     void *dest_buf, lv_coord_t dest_stride, const lv_area_t *dest_area,
-                                     void *src_buf, lv_coord_t src_stride, const lv_area_t *src_area)
+void lv_draw_stm32_dma2d_buffer_copy(lv_draw_ctx_t * draw_ctx,
+                                     void * dest_buf, lv_coord_t dest_stride, const lv_area_t * dest_area,
+                                     void * src_buf, lv_coord_t src_stride, const lv_area_t * src_area)
 {
     LV_UNUSED(draw_ctx);
     lv_draw_stm32_dma2d_blend_map(dest_buf, dest_area, dest_stride, src_buf, src_stride, LV_OPA_MAX);
--- a/src/draw/stm32_dma2d/lv_gpu_stm32_dma2d.c
+++ b/src/draw/stm32_dma2d/lv_gpu_stm32_dma2d.c
@@ -149,7 +149,7 @@ void lv_draw_stm32_dma2d_blend(lv_draw_ctx_t * draw_ctx, const lv_draw_sw_blend_
 
 void lv_draw_stm32_dma2d_buffer_copy(lv_draw_ctx_t *draw_ctx,
                                      void *dest_buf, lv_coord_t dest_stride, const lv_area_t *dest_area,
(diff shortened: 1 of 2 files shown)


embeddedt (commented):
Thank you! The implementation looks good to me.

Since it's intended to be a public function, can you add a note in `lv_gpu_stm32_dma2d.h` that calling `draw_ctx->wait_for_finish(draw_ctx);` is necessary?

It's not clear to me from a cursory glance at the code whether it would be needed, but your comment [here](https://github.com/lvgl/lvgl/issues/3087#issuecomment-1059336527) suggests it is.

patricklaf:
> Since it's intended to be a public function, can you add a note in `lv_gpu_stm32_dma2d.h` that calling `draw_ctx->wait_for_finish(draw_ctx);` is necessary?

In fact, it is not necessary because, it is already called before `flush_cb` is called in `draw_buf_flush`.
It works out of the box with the code example given.

kisvegabor:
Thank you, merging!

GiulioDallaVecchia:
Hi @patricklaf, @kisvegabor 
I've noticed that only if I'm drawing an image LVGL is able to take advantage of DMA2D for managing blending, but for the base shapes not, isn't it?

patricklaf:
Hi @GiulioDallaVecchia, 
Yes, it's a blitter.
You can learn more about Chrome-ART DMA2D [here](https://www.st.com/resource/en/product_training/STM32F7_System_DMA2D.pdf).

GiulioDallaVecchia:
Hi @patricklaf,

but is it not possible to take advantage of DMA2D also for drawing basic shape with blending? (instead of using software blending)

kisvegabor:
Unfortunately, DMA2D can blend a rectangle with a given color. It can only fill a rectangle with a color or blend an image. So to use DMA2D to blend a red rectangle with 50% opacity first we need to draw a red rectangle in the memory and them blend it. It's not that effective.

Besides, according to my measurements LVGL's SW render as as fast as DMA2D, especially for simple fill and blend. I'm also surprised by that.

vahidajalluian:
@kisvegabor I had the same problem as @patricklaf and rolled back some files to have the screen. Is there any fix that has been published yet? Supporting libraries like png and gif of the 8.2 version is really tempting.

embeddedt:
Your options are to use the `master` branch or wait for v8.3 which will be released in July.

vahidajalluian:
Dear @embeddedt 
I shall ask if this issue has been resolved in v8.3 which was recently released?

patricklaf:
Hi @vahidajalluian,
It was merged in master, so yes, it is part of newest release i.e. 8.3.0.

vahidajalluian:
@patricklaf  Thank you for your reply. I could neither find any documentation nor any example on how to use this function. where can I find such items?

patricklaf:
@vahidajalluian, you have an example in the first message of this PR.

PR #3156, opened 2022-03-10
chore(env_support/cmake): Add version.cmake with LVGL version information

### LINKED ISSUE #3152: Getting LVGL version when using it in a CMake-based project
### Introduce the problem
It is not currently possible to know/get the LVGL version (major, minor, patch) when using it in a CMake-based project.

### Examples and cases
In `lv_port_esp32` we have some API implementations that depends on the LVGL version, instead of having `#if`, `#else`, `#endif` blocks I would like to have separated source files and compiling/link them depending on the LVGL version being used.

### Suggested solution
Have a file named `version.cmake` and expoer cmake variables from there,

C47D:

Adds `version.cmake` file to get LVGL version information, to be used in `CMakeLists.txt` files. Does it look ok @tore-espressif ?

Closes #3152 

### Checkpoints


env_support/cmake/version.cmake
```diff
@@ -0,0 +1,4 @@
+set(LVGL_VERSION_MAJOR 8 PARENT_SCOPE)
+set(LVGL_VERSION_MINOR 0 PARENT_SCOPE)
+set(LVGL_VERSION_PATCH 3 PARENT_SCOPE)
+set(LVGL_VERSION_INFO "dev" PARENT_SCOPE)
```
tore-espressif: Could you also please add
```
set(LVGL_VERSION "${LVGL_VERSION_MAJOR}.${LVGL_VERSION_MINOR}.${LVGL_VERSION_PATCH}")
```

This way all CMake version compares can be used https://cmake.org/cmake/help/latest/command/if.html#version-comparisons

tore-espressif (changes requested):
I have just one note about the `PARENT_SCOPE`

So there are two ways we can use this `version.cmake` file. 
1. Either we include it in LVGL's CMakeLists.txt (that means all files in https://github.com/lvgl/lvgl/tree/master/env_support/cmake), in this case we need the `PARENT_SCOPE` to propagate it to user's project CMake scope. (I'm slightly inclined to this solution - it is easier for the user)
2. Or, the user includes it in his project manually and in this case we don't need the `PARENT_SCOPE` 

(Not really a critical thing, just an FYI)

Thank you very much for your effort

C47D:
Ok, I think I got the intention of point 1, the variables will "bubble up" and the user can use the defined variables in their `CMakeLists.txt`, will they need to include anything?

tore-espressif:
I made few experiments and this solutions seems to be the best https://github.com/lvgl/lvgl/commit/dccb684471e63db4dab00f50d9404ee6b81ed5d9. 

Exposing the version in env-variable rather than `PARENT_SCOPE` makes it accessible project-wise.

Example usage:
`message(STATUS "LVGL version is $ENV{LVGL_VERSION}")`

(I didn't expose the major/minor/patch versions, as only the combined `LVGL_VERSION` should be used)

C47D:
It does look much better, do you want to submit that commit as PR?

tore-espressif:
> It does look much better, do you want to submit that commit as PR?

No, not worth it :D you can either cherry-pick it, or just hard-copy it into your PR; you don't need to mention me as co-author

C47D:
@kisvegabor Ready for review

env_support/cmake/version.cmake
```diff
@@ -0,0 +1,6 @@
+set(LVGL_VERSION_MAJOR 8)
+set(LVGL_VERSION_MINOR 0)
+set(LVGL_VERSION_PATCH 3)
+set(LVGL_VERSION_INFO "dev")
```
xiaoxiang781216, line 4: can we do this similar thing like: https://github.com/lvgl/lvgl/blob/master/scripts/find_version.sh
or call find_version.sh directly to simplify the release process.

C47D: I don't know bash, but I guess `$LVGL_VERSION_MAJOR`, `$LVGL_VERSION_MINOR`, etc can be used to automatically create the content `version.cmake` file?

xiaoxiang781216: The key point is that we should just define version number in one place, other location infer the version info from the first place by parsing it. Otherwise, the maintainer may forget to update one of two location in the release process.

C47D: Agree, even this commit has a bug, see MINOR version set in the proposed `version.cmake` file. @kisvegabor @embeddedt where do you think we should auto generate the `version.cmake` file?

kisvegabor: Now store the versions number on many unrelated places:
- `library.json`
- `library.properties`
- `lv_conf_template.h`
- `version.cmake`
- `lvgl.h`
- etc

I can't imagine a good way to make all these files use a common source. Instead we can have a simple script which `sed`s the content of these files.

C47D:
@kisvegabor @xiaoxiang781216 Should we open a separated issue (and PR) for using `sed` to populate all of those files?

kisvegabor:
> Should we open a separated issue (and PR) for using sed to populate all of those files?

I not really meant populate, but to update those files. But, yes, let's open a new issue!

PR #3161, opened 2022-03-11
Part 1. Add support for img control and img src type.

X-Ryl669:

Rework partial lottie pull request in smaller requests.

#### Description of this PR suite 
This first part only implement the interface change in `lv_img` code. It adds the declaration of animation control and decoder caps.
There's no decoder change in this PR, so animation and vector rendering features aren't working (but it shouldn't break anything on existing code).

The next PR will add support for `lv_animbtn` that's making use of this features (again, non working, since there decoder change aren't implemented)

The last PR will add all the decoder & image cache changes (large PR) to make all of these working. This one will be a breaking change, since the interface will drop support for `void*` image source.

#### Description for this PR
I've tried to keep compatibility for a `void*` image's source in this PR. When the first byte pointed by the `void*` source is `0xFF` then it's interpreted as a `lv_img_src_t` type. 

I've deprecated the use of `lv_img_set_src` function (it prints a warning at runtime in the log).

When version 9.0 will be released, we'll be able to remove the check for version 8 that's adding this useless header to the img source type. This also provides a better (compact) form of the source structure.

There is no animation timer in this PR, since it's too hard to implement without the decoder's code. Changing the decoder's code is not possible in a small PR without implication in multiple place of the code base (most of the change in #2904 are due to this).

The idiomatic setting of an image source in future LVGL should change from:
```cpp
  // Case 1: Image as a variable type
  LV_IMG_DECLARE(whatever);
    // Somewhere else, we have lv_img_dsc_t whatever = { .width = 128, .data = {...}, etc }
  lv_img_set_src(img, &whatever);
  // Case 2: Image as a file
  lv_img_set_src(img, "/path/to/file");
  // Case 3: Symbol
  lv_img_set_src(img, LV_SYMBOL_OK);
```
to
```cpp
  // Case 1: Image as a variable type
  const char image_data[] = {...};
  lv_img_accept_src(img, lv_img_src_from_data(image_data, sizeof(image_data)));
  // Case 2: Image as a file
  lv_img_accept_src(img, lv_img_src_from_file("/path/to/file"));
  // Case 3: Symbol
  lv_img_accept_src(img, lv_img_src_from_symbol("OK " LV_SYMBOL_OK));
  // Case 4: Widget/Snapshot
  lv_img_accept_src(img, lv_img_src_from_widget(obj));
```
(a bit more verbose, but typesafe). I've changed the function name to ensure that old code will not build (since `void *` is converted silently to anything, keeping `lv_img_set_src` would build wrongly).


lvgl.h
```diff
@@ -10,14 +10,6 @@
 extern "C" {
 #endif
 
-/***************************
```
kisvegabor: We have intentionally moved the version defines here because people seemed to find it easier here.

src/draw/lv_draw_img.c
```diff
@@ -182,39 +182,6 @@ bool lv_img_cf_has_alpha(lv_img_cf_t cf)
...
 
-/**
- * Get the type of an image source
- * @param src pointer to an image source:
- *  - pointer to an 'lv_img_t' variable (image stored internally and compiled into the code)
- *  - a path to a file (e.g. "S:/folder/image.bin")
- *  - or a symbol (e.g. LV_SYMBOL_CLOSE)
- * @return type of the image source LV_IMG_SRC_VARIABLE/FILE/SYMBOL/UNKNOWN
- */
-lv_img_src_t lv_img_src_get_type(const void * src)
```
kisvegabor, line 193: It's not tightly related to the `lv_img` widget but to the image drawing. So it should be placed to a generic place.  (like this file.)

src/draw/lv_img_decoder.h
```diff
@@ -28,17 +29,44 @@ extern "C" {
 /**********************
  *      TYPEDEFS
  **********************/
+#if defined(LV_IMG_FRAME_INDEX_LARGE_SIZE)
+typedef uint32_t lv_frame_index_t;
```
kisvegabor, line 32: Do we really need this config? It seems we can save only a few bytes but we need to add an other config option. I think 
 always  using `uint32_t` is enough.

src/widgets/lv_img.h
```diff
@@ -87,14 +108,61 @@ lv_obj_t * lv_img_create(lv_obj_t * parent);
...
+ * (like `lv_obj_set_size` for vector format) before calling this
+ *
+ * @param obj           pointer to an image object
+ * @param src           pointer to an image source object
+ * @param skip_cache    if 1, the image will only be parsed to extract its metadata but it'll skip
+ *                      the image cache until it's actually used.
+ *                      You might only need to set this if you are preparing your screen at once on start
+ * @return LV_RES_OK    If the image was correctly parsed
+ */
+lv_res_t lv_img_accept_src(lv_obj_t * obj, const lv_img_src_t * src, int skip_cache);
```
kisvegabor: I like the return value here but I'm not sure about `skip_cache`. I think we should be able to handle it internally.  It'd be great to know what we win by having this option. E.g. what is the different in widget creation time when creating 10 images with our without caching. 

I'd like to avoid solving a non existing (or only theoretical) problem.

---

```suggestion
lv_res_t lv_img_set_src_ext(lv_obj_t * obj, const lv_img_src_t * src, int skip_cache);
```

The word `accept` is not that intuitive for me.

src/widgets/lv_img.h
```diff
@@ -17,41 +17,62 @@ extern "C" {
...
+    uint16_t            angle;      /**< rotation angle in 1/10th of degree of the image*/
+    uint16_t            zoom;       /**< 256 means no zoom, 512 double size, 128 half size*/
+    uint8_t             cf : 5;        /**< Color format from `lv_img_color_format_t`*/
+    uint8_t             antialias : 1; /**< Apply anti-aliasing in transformations (rotate, zoom)*/
+    uint8_t             obj_size_mode: 2; /**< Image size mode if image and object sizes differ */
+
+    uint8_t             ctrl;       /**< The current control flags */
+    lv_img_dec_ctx_t  * dec_ctx;    /**< Additional decoder context */
+    uint8_t             caps;       /**< Decoder capabilities */
+    lv_timer_t     *    anim_timer; /**< The timer task for animated images */
```
kisvegabor, line 71: Odd `*` placement.

src/widgets/lv_img.h
```diff
@@ -87,14 +108,61 @@ lv_obj_t * lv_img_create(lv_obj_t * parent);
...
+
+/**
+ * Set the stop frame for the animation
  * @param obj       pointer to an image object
- * @param src_img   1) pointer to an ::lv_img_dsc_t descriptor (converted by LVGL's image converter) (e.g. &my_img) or
+ * @param index     Index of the frame to stop at. Will emit a EVENT_READY when reached.
+ * @param forward   If 1, the animation plays forward, else plays backward
+ * @warning This will only work once the source of the image is set and the image is animated
+ */
+lv_res_t lv_img_set_stopat_frame(lv_obj_t * obj, const lv_frame_index_t index, const int forward);
```
kisvegabor:
```suggestion
lv_res_t lv_img_set_stop_at_frame(lv_obj_t * obj, const lv_frame_index_t index, const int forward);
```

src/widgets/lv_img.h
```diff
@@ -87,14 +108,61 @@ lv_obj_t * lv_img_create(lv_obj_t * parent);
...
+
+/**
+ * Parse the image source to display on the the object
```
kisvegabor:
```suggestion
 * Set an image source for the image object
```

src/misc/lv_math.h
```diff
@@ -136,6 +136,11 @@ uint32_t lv_rand(uint32_t min, uint32_t max);
 #define LV_SMAX_OF(t) (((0x1ULL << ((sizeof(t) * 8ULL) - 1ULL)) - 1ULL) | (0x7ULL << ((sizeof(t) * 8ULL) - 4ULL)))
 #define LV_MAX_OF(t) ((unsigned long)(LV_IS_SIGNED(t) ? LV_SMAX_OF(t) : LV_UMAX_OF(t)))
 
+/* Bit-flag test true */
+#define LV_BT(v, b) ((v) & (b)) == (b)
+/* Bit-flag test false */
+#define LV_BN(v, b) ((v) & (b)) != (b)
```
kisvegabor:
```suggestion
#define LV_BF(v, b) ((v) & (b)) != (b)
```

Or `BN` is a standard name?

src/draw/lv_img_decoder.h
```diff
@@ -28,17 +29,44 @@ extern "C" {
...
+ * Set by the decoder when extracting info or being opened */
+typedef enum {
+    LV_IMG_DEC_DEFAULT      = 0x00, /**!< Default format, no specificities */
+    LV_IMG_DEC_VECTOR       = 0x01, /**!< Image format is vector based, size independant */
+    LV_IMG_DEC_ANIMATED     = 0x02, /**!< Image format stores an animation */
+    LV_IMG_DEC_SEEKABLE     = 0x04, /**!< Animation is seekable */
+    LV_IMG_DEC_CACHED       = 0x08, /**!< The complete image can be cached (used for rotation and zoom) */
+    LV_IMG_DEC_VFR          = 0x10, /**!< The animation has a variable frame rate */
+    LV_IMG_DEC_LOOPING      = 0x20, /**!< The animation is looping */
+} lv_img_decoder_caps_t;
```
kisvegabor:
```suggestion
} lv_img_dec_caps_t;
```

We used `dec` on other places too.

src/draw/lv_img_src.c
```diff
@@ -0,0 +1,191 @@
...
+lv_img_src_type_t lv_img_src_get_type(const void * src)
+{
+    lv_img_src_type_t img_src_type = LV_IMG_SRC_UNKNOWN;
+
+    if(src == NULL) return img_src_type;
+    const uint8_t * u8_p = src;
+
+    /*The first byte shows the type of the image source*/
+    if(u8_p[0] >= 0x20 && u8_p[0] <= 0x7F) {
+        img_src_type = LV_IMG_SRC_FILE; /*If it's an ASCII character then it's file name*/
```
kisvegabor: Nitpicking: why not `return LV_IMG_SRC_FILE`?

src/draw/lv_img_src.c
```diff
@@ -0,0 +1,191 @@
...
+    }
+
+    if(LV_IMG_SRC_UNKNOWN == img_src_type) {
+        LV_LOG_WARN("lv_img_src_get_type: unknown image type");
+    }
+
+    return img_src_type;
+}
+
+lv_res_t lv_img_src_parse(lv_img_src_t * uri, const void * src)
```
kisvegabor: Shouldn't it have similar prototype to lv_img_src_set_symbol?
E.g. 

```suggestion
lv_res_t lv_img_src_set_dsc(lv_img_src_t * src, const void * src_data)
```

TBH, I think `uri` is not descriptive enough for novice users.

src/widgets/lv_img.c
```diff
@@ -693,17 +733,17 @@ static void draw_img(lv_event_t * e)
                     coords_tmp.x2 = coords_tmp.x1 + img->w - 1;
 
                     for(; coords_tmp.x1 < img_max_area.x2; coords_tmp.x1 += img_size_final.x, coords_tmp.x2 += img_size_final.x) {
-                        lv_draw_img(draw_ctx, &img_dsc, &coords_tmp, img->src);
+                        lv_draw_img(draw_ctx, &img_dsc, &coords_tmp, img->src.uri);
```
kisvegabor: Why no pass `img->src`? As it already contains the source type.

src/draw/lv_img_src.h
```diff
@@ -0,0 +1,154 @@
...
+    size_t          uri_len;        /**< The next URI length in bytes */
+#else
+    size_t          type : 3;       /**< See `lv_img_src_type_t` above */
+#if UINTPTR_MAX == 0xffffffffffffffff
+    size_t          uri_len : 61;   /**< The next URI length in bytes */
+#else
+    size_t          uri_len : 29;   /**< The next URI length in bytes */
+#endif
+#endif
+    const void   *  uri;            /**< A pointer on the given unique resource identifier */
```
kisvegabor:
```suggestion
    const void   *  src_data;            /**< A pointer on the given unique resource identifier */
```

kisvegabor (commented):
Thank you very much!

In PR I couldn't find `lv_img_src_from_data`. Is it planned in a new PR(s)?

In this PR how can the user use the new API to set e.g. an image path as src?

lvgl.h
X-Ryl669: The issue with having version in lvgl.h is that it's including plethora of other header. In the include header chain, there's the header that's trying to use the version macro.  
So any code that's checking for the version should either be outside the standard/common lvgl.h include pool or we have circular dependencies. 

Right now, I'm using LVGL_MAJOR as a way to deprecate & change the structure layout. On LVGL8, since we allow `lv_img_set_src` the `lv_img_src_t` as a "first" byte in the structure set as '0xFF' to be able to be detected by img_get_src_type. But once we drop support for guess-my-type-void* this hack can be dropped and the struct be more compact.

src/draw/lv_draw_img.c
X-Ryl669: Since there's the new `lv_img_src_t` type and the matching lv_img_src.c/h files, it's where it's gone. Everything related to image source is in this header/source file now.

src/draw/lv_img_decoder.h
X-Ryl669: This is a slippage from ffmpeg decoder I'm including earlier in the sequence. For lottie files, a uint16 is more than enough. Since it's used for many member fields (current frame, stop frame, total frame, iteration start & destination frame, etc...), in the end that's a large saving in for the decoder context and animation fields. For ffmpeg however, there might be more than 65535 frames in a movie (that's ~36mn at 30fps) so a uint32 is required.

I don't know what's the best here. Maybe use USE_FFMPEG as the enabling macro to avoid creating a new config option ?

src/widgets/lv_img.h
X-Ryl669: That's a requirement you've made in the former PR. It was made for those GUI where everything is set up/declared upfront (like when creating all screens). We can remove this here and use a caps instead (so in case you want to gain 0.1% of speed compared to hitting the cache, you'll set the decoder's context cap to "SKIP_INITIAL_CACHE" or something), just like we do for lottie files and their initial size.  

Ok for accept/set_src_ex.

My initial idea was to have a different name than `set_src` to avoid LVGL <= 8 code to compile when we drop the `void*` source type. In LVGL 9, it can't be `lv_img_set_src` (else previous code will compile but will crash at runtime). `set_src_ex` without a `set_src` looks awkward too. I don't know.

src/draw/lv_img_src.c
X-Ryl669: That's a copy and paste of the current code in LVGL. I *guess* it's because of the LV_LOG_WARN macro below that make sure at least one source type is detected. But, since there's a `else` in the if/else chain, the log can't be triggered. 
I'll remove it.

src/draw/lv_img_src.c
X-Ryl669: I think there a misunderstanding here. `lv_img_src_parse` is the function used to accept the *old* void* parameter that could point to either a file, a dsc or a symbol and build a `lv_img_src_t`. It's actually parsing the (void*) buffer to figure out what it contains, as lv_img used to in set_src. 
This function is used for compatibility where void* are kept (like bg_img_src, window_icon, etc...). Ideally, it should disappear in LVGL9, but that's a lot of work since it implies modifying the API to remove all traces of void* for image source (and instead replace by `lv_img_src_t*`).

Again we have an issue here because since everything converts to void* a code written for LVGL <= 8 will compile in LVGL9 if we keep the function name the same, but crash at runtime.

So we need to agree on how to rename the functions in order to avoid this issue. I'm not that good at function name, if you have some ideas please suggest!

src/draw/lv_img_src.h
X-Ryl669: Ok

src/misc/lv_math.h
X-Ryl669: No, I don't think so. Ok for change.

src/widgets/lv_img.c
X-Ryl669: That's where I put the limit between PR. The part 3 will modify the lv_draw_img signature to accept a `lv_img_src_t*` instead of void * so yes, that's exactly like you've thought.

lv_draw_img new version use the new image cache code and the new decoder code so I had to decide when to include it in the PR suite.

src/widgets/lv_img.h
X-Ryl669: oops

src/widgets/lv_img.h
X-Ryl669: Ok

src/draw/lv_img_src.c
X-Ryl669: I've tried different strategies to keep the same name but have the compiler to break on the new code:
Ideally we would want to go from `lv_img_set_src(lv_img_t * img, void* src)` to `lv_img_set_src(lv_img_t * img, lv_img_src_t * src)` but since `void*` is (silently) convertible to the `lv_img_src_t*` it'll silently compile invalid code like `lv_img_set_src(img, dsc->bg_img)` where dsc->bg_img is a void * containing a symbol or a file's path.

Here are the strategies I've thought about:
1. Change the type only and add a `-Wc++-compat` flag to the compilation. This will issue a compiler warning on GCC (sorry for other compiler). The code will still build and crash at runtime.
2. Change the signature to add another parameter (like `lv_img_set_src(lv_img_t * img, lv_img_src_t * src, uint32_t flags)`. In that case, users will have to adapt their code (so a giant warning above the function declaration should say "don't use void * anymore") 
3. Change the function name (like `lv_img_set_src_ex(lv_img_t * img,  lv_img_src_t * src)` and remove the previous function. 
4. Add a new function name and keep the previous function that'll go to a backport parsing (like in this PR). However, user will continue to use the old function until it's removed. And this might not work for every place a `void *` is used to store some *guess what I am* source. 
5. Pass src type by value, like (`lv_img_set_src(lv_img_t * img, lv_img_src_t src)`). That's probably the best solution but not as efficient as passing a pointer. Since the `lv_img_src_t` is a small structure (one type, one pointer and a length), that might work ok anyway. The compiler will refuse to compile previous code with a void * and so it should be easy to deal with. Also, we can have functions like this: `lv_img_src_t lv_img_src_from_file(const char* path)` to the final code would look like `lv_img_set_src(img, lv_img_src_from_file("/path/to/img.png"))` which I like because it's self describing.

What do you think is best ?

X-Ryl669:
> In PR I couldn't find `lv_img_src_from_data`. Is it planned in a new PR(s)?

We need to settle on the API first (see above). But yes, it's expected in the last PR.
 
> In this PR how can the user use the new API to set e.g. an image path as src?

```cpp
// Create image object
lv_img_t * img = lv_img_create();

// Create the source object to use for the image
lv_img_src_t src;
lv_img_src_set_file(&src, "/path/to/file");

// Assign the source to the image
lv_img_set_src(img, &src);
```
One can (should?) re-use the source instance in multiple place, since it avoid re-parsing it.

In the last PR it might become:

```cpp
// Create image object
lv_img_t * img = lv_img_create();
// Assign the source to the image
lv_img_set_src(img, lv_img_src_from_file("/path/to/img"));
```

As a side note, I think it's safe to drop the `img` in the name to only get this: `lv_src_from_file("/path/to/file")` or just `lv_from_file` maybe ?

lvgl.h
kisvegabor: I see the problem. But do we really need this version check? In v8 we use the `LVGL_VERSION_MAJOR <= 8` branch and in v9 we change it to `#else` branch's code. LVGL can't have two versions at the same time and all code in LVGL should be compatible only with the current version. 

Version check makes sense in external files which are independent from LVGL.

src/draw/lv_draw_img.c
kisvegabor: My bad, sorry. I though it was moved to `lv_img.c`.

src/draw/lv_img_decoder.h
kisvegabor: > current frame, stop frame, total frame, iteration start & destination frame, etc...), 

So the saving is 10 bytes, let's say 16 on each decoder context on in other object where it is used. I think it doesn't matter compared to the huge RAM usage of image decompression. 

Other topic, but I suggested earlier to introduce an `lv_int_t`  type (and maybe an `lv_uint_t`) and it could be used for coordinates, slider value, chart point count,etc, and for lv_frame_index_t.  And the user could set `lv_int_t` in `lv_conf.h`. What do you think about it?

X-Ryl669: A bit like `lv_coord_t` ? We could use this (saying that the frame index is a coordinate in time...)

src/widgets/lv_img.h
kisvegabor: > so in case you want to gain 0.1% of speed 

It seems much more than 0.1%. E.g. decoding a PNG image vs. reading its first few bytes to get its resolution.

This issue emerged due to the removal of `decoder->info_cb`, right? Are you sure it's really needs to be removed? Just thinking aloud: if getting the info, and decoding takes the same time the decoder can decide to save the decoded image too.

src/draw/lv_img_src.c
kisvegabor: > Pass src type by value, like (lv_img_set_src(lv_img_t * img, lv_img_src_t src)). That's probably the best solution but not as efficient as passing a pointer. Since the lv_img_src_t is a small structure (one type, one pointer and a length), that might work ok anyway.

I also like it the most. The current image converter could generate an `lv_img_src_t` variable too, so in this case:
```c
lv_img_set_src(img, my_img);  //not &my_img
```

It important to keep source type open for extension by external modules. E.g. the SDL engine might provide an `SDL_Texture` image source.

src/widgets/lv_img.h
X-Ryl669: The `decoder_accept` doesn't decode the image (unlike get_info), only the headers. That's true for simple format (like bmp, gif, png, jpeg) but not for ffmpeg (as I haven't found a way to do that) or rlottie.
In this latter case, I'm still skipping rendering/color conversion, so it still faster than full image decoding.

By default, skip_cache is 0, so instead of calling `decoder_accept`, we directly call `decoder_open` that's decoding the image (or part of the image) and store the result in the cache. So no time is wasted if the next action will be to display the image.

When skip_cache is 1, `decoder_accept` is called instead. For simple format, it avoid the cache and is faster than decoding the whole picture. When painting, since the cache is empty, decoder_open will be called. 
For more complex format, the image will be parsed and decoded but image rendering and video decoding will be skipped.
However as soon as the image need painting, we'll have to pay the parsing time again.
 
In current LVGL code, get_info decodes the image and throw the result away. Worst, it's called twice per paint (just before a decoder_open) so it's a lot slower. It's only faster when dealing with SRC_VARIABLE, because it's using the `lv_img_dsc_t`'s fields, but it's like cheating, since another process gave it the exams' results.

In the end, I'm not sure of the utility of skip_cache. If you have plenty of screens you're building at once on program startup, without skip_cache, they'll all compete on the cache, so we might get a lot of evictions, but at the same time, the last screen will have its images cached already. Maybe it's worth adding a caps to the decoder and have the generic screen code API (upcoming for v8.3 ???) set this caps internally. This would gives a good advantage but doesn't modify the API with some useless complexity, IMHO.

src/widgets/lv_img.c
kisvegabor: Clear, and I agree to put the line here.

kisvegabor:
The current API is a little bit cumbersome but `lv_img_set_src(img, lv_img_src_from_file("/path/to/img")); looks much better. 

> As a side note, I think it's safe to drop the img in the name to only get this: lv_src_from_file("/path/to/file") or just lv_from_file maybe ?

IMO, `src` on its own is too broad.

src/widgets/lv_img.h
kisvegabor: I see `skip_cache` is unused in the current PR. TBH, I can't exactly oversee the consequences blindly (i.e. without seeing the actual processes).

I suggest removing the `skip_cache` parameters for now, and reconsider it when the functions using it will be added too.

src/draw/lv_img_src.c
X-Ryl669: Ok. I agree.

Right now, in LVGL9, lv_img_src_t is like this:
```cpp
struct src
{
    uint32_t type: 2;
    uint32_t length: 30; (or 62 on 64 bits machine)
    void * data;
};
```
So we have 4 possibles source type (file, data, symbol, unknown) and 1GB of for the source's data length (for data mainly). I think it's completely safe to reduce to only 512MB here and allow 8 possible source type. Storing a 512MB binary picture inside the final executable would be crazy anyway. 
That way, we could add new source type like "widget snapshot" and GPU textures.

X-Ryl669:
Ok, I agree.

src/draw/lv_img_src.c
kisvegabor: I'd rather choose:
```c
struct src
{
    uint32_t type: 8;
    uint32_t reserved: 24;
    uint32_t lenghth;
    void * data;
};
```

My way of thinking is that we are just redesigning an interface an seemingly we are close to some possible limits:
- having space only for 4 extra source types?
- only 512MB files? It it really that much? Even a 720p movie is larger than that.

X-Ryl669: Length is for embedded data (in the final binary software). File & stream will store the path/uri length here, it's not the file size. So having 512MB of embedded resource in a binary .EXE is huge (IMHO). Even 256MB would be huge in today's standard.

I'm trying to limit the size of the structure since it'll be passed by value, so I think 2 or 3 registers is ok, but having more is not. There's no point in reserving space here, IMHO.  We could also steal bits from the data pointer, since it's always aligned, we can steal the 2 or 3 low bits.

How many type do you think of ?

1. uri (file path at first, can become true url in the future since ffmpeg supports them)
2. embedded data 
3. text (symbols)
4. snapshot (ptr to lv_img_dsc)
5. unknown
6. file descriptor (like GPU textures, DMABUF etc...)
7. ???
8. ???

kisvegabor: > File & stream will store the path/uri length here, 

Okay, I'm convinced :slightly_smiling_face: 


> How many type do you think of ?

I found some candidates. (I don't say all these makes sense and should be differentiated at `src_type` level but it shows that there are other options)

5. Vector path
6. SDL textures
6. OpenGL textures
6. DirectX textures
7. Some kind of stream?
8. Web URL
9. FTP
10. 3D model
10. 3D model in other format

I think web URL's should have it's own type e.g. to allow the PNG decoder to drop them. 

So 16 types seems better to me. >256 MB of embedded date really doesn't make much sense.
```c
struct src
{
    uint32_t type: 4;
    uint32_t length: 28; //256MB 
    void * data;
};
```

Anyway, this `struct` can be modified/adjusted at any time without breaking anything.

src/widgets/lv_img.h
X-Ryl669: In fact, it's the CI's style checker fix for it.

X-Ryl669:
Failed CI test is due to lv_binding repository using `lv_img_decoder_dsc_t` instead of `lv_img_dec_dsc_t`

kisvegabor:
> Failed CI test is due to lv_binding repository using lv_img_decoder_dsc_t instead of lv_img_dec_dsc_t

No problem with that we can update it later.

X-Ryl669:
Ok, I think I've took care of all your concerns. If it's ok, I'll prepare part 2. Don't merge part 1 yet (since it doesn't add anything until all parts are included), the idea is to merge all parts at the same time.

No API breaking changes in this patch set.

src/draw/lv_img_src.c
X-Ryl669: Agreed. Will be in changed in LVGL9 with the change in the `lv_img_set_src` api. Right now, I've removed all version checking so that there isn't any API breakable at this level.

kisvegabor:
I've created the `arch/img-decode-rework` branch. I suggested merging there to have a stable ground.

kisvegabor:
I still would like to clarify one thing:

- `LV_IMG_SRC_SYMBOL` means e.g. `LV_SMYBOL_OK`
- `LV_IMG_SRC_FILE` means e.g. `"S/dir/img.png"`
- `LV_IMG_SRC_VARIABLE` a raw `uint8_t` array containing e.g. a raw PNG image
- `LV_IMG_SRC_OBJ` means `lv_img_dsc_t *`. If so the comment `This means the void* points to a lv_img_src_t` should contain `lv_img_dsc_t`. Can we refactor it to  `LV_IMG_SRC_IMG_DSC_T`? It's odd but more descriptive.

X-Ryl669:
>     * `LV_IMG_SRC_SYMBOL` means e.g. `LV_SMYBOL_OK`

Correct.

> 
>     * `LV_IMG_SRC_FILE` means e.g. `"S/dir/img.png"`

Correct.

>     * `LV_IMG_SRC_VARIABLE` a raw `uint8_t` array containing e.g. a raw PNG image

In my previous PR, it was true but it broke with the current API that's expecting a `lv_img_dsc_t*`
So this PR (for LVGL8), it's a `lv_img_dsc_t*` as before. 

In LVGL9 it'll be like you said.


> 
>     * `LV_IMG_SRC_OBJ` means `lv_img_dsc_t *`. 

No. It's a `lv_img_src_t*`. (which is a struct containing a type by itself and a void * that depends on the type + length).
This was added for compatibility with LVGL8. 

It's expected to be removed in LVGL9 since `lv_img_src_t` will be passed by value.

>If so the comment `This means the void* points to a lv_img_src_t` should contain `lv_img_dsc_t`. Can we refactor it to  `LV_IMG_SRC_IMG_DSC_T`? It's odd but more descriptive.

That's what LV_IMG_SRC_VARIABLE is for. This PR shouldn't break user code, so I can't change `LV_IMG_SRC_VARIABLE` to `LV_IMG_SRC_IMG_DSC_T`. `LV_IMG_SRC_OBJ` is just a placeholder until we increase the major version, to mean that the given pointer already points to a `lv_img_src_t` so no need to parse again it can be copied directly. 

Any type you pass in with `lv_img_set_src(..., void*)` will be converted to a `lv_img_src_t*` stored one of the member of the `lv_img_t` instance. When given to the rest of the API expecting a `void*` so its type will be detected as `LV_IMG_SRC_OBJ` internally. 

Yet, since I'm not touching everywhere a `void *` is expected in this PR, it'll be parsed again (for example, in lv_draw_image) and will be correctly detected as a `lv_img_src_t*`. 

Said differently:
### Before

```
[file]  => lv_img_set_src => lv_img.src = strdup(file)
[symbol] => lv_img_set_src => lv_img.src = strdup(symbol)
[lv_img_dsc_t* p] => lv_img_set_src => lv_img.src = p

Then, when used:
lv_img.src => parsed as file => decode file
lv_img.src => parsed as symbol => draw text
lv_img.src => parsed as lv_img_dsc_t* => decode pointed data
```

### Now

```
[file]  => lv_img_set_src => lv_img.src.type = FILE, lv_img.src.data = strdup(file)
[symbol] => lv_img_set_src => lv_img.src.type = SYMBOL, lv_img.src.data = strdup(symbol)
[lv_img_dsc_t* p] => lv_img_set_src => lv_img.src.type = VARIABLE, lv_img.src.data = p
[lv_img_src_t * p] => lv_img_set_src => lv_img.src.type = OBJ, lv_img.src.data = copy(p)

Then, when used:
lv_img.src.type == file => decode file
lv_img.src.type == symbol => draw text
lv_img.src.type == variable => decode pointed data as lv_img_dsc_t*
```

### LVGL9
```
[file]  => lv_img_set_src(from_file) => lv_img.src.type = FILE, lv_img.src.data = strdup(file)
[symbol] => lv_img_set_src(from_symbol) => lv_img.src.type = SYMBOL, lv_img.src.data = strdup(symbol)
[lv_img_dsc_t* p] => lv_img_set_src(from_descriptor) => lv_img.src.type = DESCRIPTOR, lv_img.src.data = p ???? Do we still need this or simply kill it once for all ?
[lv_img_src_t * p] => lv_img_set_src(direct) => lv_img.src.type = p.type, lv_img.src.data = copy(p)

Then, when used:
lv_img.src.type == file => decode file
lv_img.src.type == symbol => draw text
lv_img.src.type == variable => decode pointed data as (byte*, len)
```

kisvegabor:
Got it, thanks for the explanation.

Please confirm if I can merge it into `arch/img-decode-rework`.

X-Ryl669:
Yes you can. I'll track this branch then.

kisvegabor:
Merged, thank you!

PR #3162, opened 2022-03-11
feat(gpu): Add gpu arm2d

GorgonMeducer:

Add support for arm's 2D image processing library (arm-2d).

### Checkpoints

### DIFF
--- a/Kconfig
+++ b/Kconfig
@@ -207,9 +207,15 @@ menu "LVGL configuration"
         menu "GPU"
             config LV_USE_EXTERNAL_RENDERER
                 bool
-
+                
+            config LV_USE_GPU_ARM2D
+                bool "Enable Arm's 2D image processing library (Arm-2D) for all Cortex-M processors."
             config LV_USE_GPU_STM32_DMA2D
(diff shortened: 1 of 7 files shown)


kisvegabor:
Hi,

Looks great! Thank you! 

BTW, have measured the performance difference between SW rendering and ARM GPU?

GorgonMeducer:
@kisvegabor We use the official benchmark with the following modifications:
a. set `LV_DISP_DEF_REFR_PERIOD` to 2 to force a 100% CPU usage in every test scenario
b. implement `disp_flush()` as an almost empty one to only measure the rendering time (without the overhead of low-level flush.
c. In theory, you can use any Cortex-M processor with the arm-2d library to test the performance. I am using the `https://github.com/lvgl/lv_port_an547_cm55_sim` to test the performance.
d. When using `arm-2d`, please make sure you define a macro `__ARM_2D_HAS_ASYNC__=0` in your project configuration. 

We have some internal performance data and I can share it with you. It would be better to do it via email.
If possible, would you mind ping me via gabriel.wang@arm.com?

kisvegabor:
Sure, email sent.

PR #3163, opened 2022-03-12
fix(draw): src_buf_tmp will be NULL when LV_DRAW_COMPLEX is '0'

GorgonMeducer:

If `LV_DRAW_COMPLEX` is `0`, then the local pointer `src_buf_tmp` will not be assigned with a proper address but used with `NULL`. 

It looks like this code is not well tested with `LV_DRAW_COMPLEX` set to `0`

### Checkpoints


kisvegabor:
Thanks!

PR #3171, opened 2022-03-15
fix(style) fix typo in value line spacing calculation

ckielstra:
### Applicable version
This bug applies to the v7 release branch. The code in v8 is very different and this bug doesn't seem to apply there.

The code for calculating the extra space around a 'value style' property contained a copy/paste error where the 'letter space' instead of the 'line space' was used in the vertical required size calculations.


embeddedt:
Thanks!

PR #3184, opened 2022-03-20
feat(tabview): support vertical scrolling

timstableford:

This makes it so when the tabview has the tabs on the left or right
it will be scrolled up and down rather than left and right in all cases.

### Checkpoints
- [X] Update the documentation - None to update

I couldn't see any reference in tabview that specifically documents the scroll direction to change. Please let me know what I'm meant to do for this case.

I appreciate this changes existing behaviour and I'm happy to modify this so that scroll direction could be passed as an option to keep the current behaviour. I'm also not convinced the code around RTL support when vertically scrolling is correct for my changes in cont_scroll_end_event_cb is correct.


kisvegabor (commented):
Hi,

Makes sense. Only one thing: `base_dir` doesn't affect vertical positioning.

timstableford:
@kisvegabor I believe I've now confined the rtl behaviour to the original horizontal scrolling

kisvegabor:
Looks good, thank you!

PR #3199, opened 2022-03-23
chore(config): change LV_FFMPEG_AV_DUMP_FORMAT to LV_FFMPEG_DUMP_FORMAT

xiaoxiang781216:

ensure all ffmpeg special config start with the prefix(LV_FFMPEG_)

### Checkpoints

### DIFF
--- a/Kconfig
+++ b/Kconfig
@@ -969,8 +969,8 @@ menu "LVGL configuration"
 
         config LV_USE_FFMPEG
             bool "FFmpeg library"
-        config LV_FFMPEG_AV_DUMP_FORMAT
-            bool "Dump av format"
+        config LV_FFMPEG_DUMP_FORMAT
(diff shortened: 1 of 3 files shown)


kisvegabor:
I think it affects only a few people, but it's actually a breaking change. Why is it that important to remove the `AV` part? The config still starts with `LV_FFMPEG_` in both cases.

xiaoxiang781216:
> I think it affects only a few people, but it's actually a breaking change.

Yes, that's why I think the change may acceptable.

> Why is it that important to remove the `AV` part? The config still starts with `LV_FFMPEG_` in both cases.

because AV is also acronym/fork for the same project: 
https://en.wikipedia.org/wiki/Libav
look like we abbreviate twice:).

kisvegabor:
Got it, thanks! :)

PR #3200, opened 2022-03-23
feat(span):add an interface for setting the number of lines

guoweilkd:
#### feat(span):add an interface for setting the number of lines
It is convenient for users to control the maximum height more easily of span obj.

#### fix draw: remove the high overflow judgment code

In many cases, the font-size < line_height. it causes the text to not be displayed under certain circumstances. Therefore, referring to web css, the line-height overflow judgment code is deleted.

### Checkpoints


kisvegabor:
I was thinking about introducing a global size unit, similar to `em` or `rem` on web. The usage could be:
```c
lv_obj_set_height(label, LV_EM(5)); //5 lines
```

Do you think something like this can be used here too?

guoweilkd:
In `LV_SPAN_MODE_BREAK` mode, the height is automatically increased, `lv_spangroup_set_lines(n)` can limit the height to no more than `n` lines. of course, lv_obj_set_max_height() can meet this requirement. However, the line-height of each span in a spangroup may be different. How to deal with this problem?

kisvegabor:
Sorry for the late reply. Number of lines  and `height = N x line_height` is really different in this case.

Let's merge it now and we will see if we can find some simplification option later.

PR #3228, opened 2022-03-30
fix(gradient): assert before dividing by 0

FASTSHIFT:
@X-Ryl669 
Is it possible to divide by 0 during operation?

### Checkpoints


X-Ryl669:
With a smart developer, yes. That would require setting multiple stop gradient (more than 2) and set the fraction the same as the previous value. There's a check for the 2 extrema, but not for internal fractions.

I don't know how to deal with this, if we have to. Since the `frac`tion is set by the developer, is set to something bad, and it asserts, it might be enough. Or maybe we should fix it (but where ?)

So, yes, it's currently useful, thanks!

kisvegabor:
It'd be better to handle this case internally silently but asserting is definitely better than failing div by 0.

PR #3232, opened 2022-03-30
fix(meter): fix infinite loop caused by loop variable type mismatch

FASTSHIFT:
### Checkpoints

### DIFF
--- a/src/extra/widgets/meter/lv_meter.c
+++ b/src/extra/widgets/meter/lv_meter.c
@@ -379,8 +379,6 @@ static void draw_ticks_and_labels(lv_obj_t * obj, lv_draw_ctx_t * draw_ctx, cons
     p_center.x = scale_area->x1 + r_edge;
     p_center.y = scale_area->y1 + r_edge;
 
-    uint8_t i;
-
     lv_draw_line_dsc_t line_dsc;
     lv_draw_line_dsc_init(&line_dsc);
     lv_obj_init_draw_line_dsc(obj, LV_PART_TICKS, &line_dsc);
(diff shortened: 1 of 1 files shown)


src/extra/widgets/meter/lv_meter.c
```diff
@@ -435,7 +433,7 @@ static void draw_ticks_and_labels(lv_obj_t * obj, lv_draw_ctx_t * draw_ctx, cons
         int16_t inner_act_mask_id = LV_MASK_ID_INV; /*Will be added later*/
 
         uint32_t minor_cnt = scale->tick_major_nth ? scale->tick_major_nth - 1 : 0xFFFF;
-        for(i = 0; i < scale->tick_cnt; i++) {
+        for(uint16_t i = 0; i < scale->tick_cnt; i++) {
```
kisvegabor:
```suggestion
        uint16_t i;
        for(i = 0; i < scale->tick_cnt; i++) {
```

kisvegabor:
Thank you for this and all the other fixes. Have you found these with a static code analyzer?

FASTSHIFT:
> Thank you for this and all the other fixes. Have you found these with a static code analyzer?

Yes，I used [Coverity](https://scan.coverity.com) and [TscanCode](https://github.com/Tencent/TscanCode).

kisvegabor:
AFAIK, Coverity can run only online, but seemingly TscanCode can run offline too. If so it'd be great to have a script that runs TscanCode. This way we can run it at least manually time to time.

FASTSHIFT:
> AFAIK, Coverity can run only online, but seemingly TscanCode can run offline too. If so it'd be great to have a script that runs TscanCode. This way we can run it at least manually time to time.

TscanCode has a [linux command line tool](https://github.com/Tencent/TscanCode/tree/master/release/linux/TscanCodeV2.14.2395.linux), which can be called as a shell script.
Consider integrating into CI.

kisvegabor:
I've just tested with `./tscancode src/. 2>result.txt` and it reported 126 problem.

If we integrated it into CI we needed to fix these 126 problems first. :slightly_smiling_face:

PR #3240, opened 2022-03-31
cmake: Add missing "REQUIRES main" for ESP-IDF

huming2207:

Hi there,

I realised that it may be necessary to add `REQUIRES main` in `idf_component_register()` in `env_support/cmake/esp.cmake`. Otherwise if I need to use any external libraries such as `freetype2`, the LVGL adapters like `lv_freetype.c` will complain about missing headers.

For example, I have a project that needs FreeType2, and I wrote a interface library (see here: https://github.com/huming2207/esp-ft2). I make it as a Git submodule and clone it into `components` directory of my ESP-IDF project. Meanwhile I also have a LVGL copy as a submodule in the `components` directory together. Now, if I enable the `LV_USE_FREETYPE` and start playing with the FreeType stuff, the compiler will complain something like this:

```
../components/lvgl/src/extra/libs/freetype/lv_freetype.c:12:10: fatal error: ft2build.h: No such file or directory
```

Even if I get the `main/CMakeLists.txt` to `REQUIRE` both `lvgl` and `freetype`, when compiling, the compiler will still complain the same error.

That is due to the LVGL component doesn't know the existence of my FreeType component. So we need to let `lvgl` component to know the existence of the project's `main` component, and since `main` component knows `freetype`, then `lvgl` component also knows the `freetype`.

Regards,
Jackson

### Checkpoints

### DIFF
--- a/env_support/cmake/esp.cmake
+++ b/env_support/cmake/esp.cmake
@@ -22,7 +22,7 @@ if(LV_MICROPYTHON)
   endif()
 else()
   idf_component_register(SRCS ${SOURCES} INCLUDE_DIRS ${LVGL_ROOT_DIR}
-                         ${LVGL_ROOT_DIR}/src ${LVGL_ROOT_DIR}/../)
+                         ${LVGL_ROOT_DIR}/src ${LVGL_ROOT_DIR}/../ REQUIRES main)
 
   target_compile_definitions(${COMPONENT_LIB} PUBLIC "-DLV_CONF_INCLUDE_SIMPLE")
 


embeddedt:
Hmm... we originally did not specify `REQUIRES main` because of https://github.com/lvgl/lvgl/issues/2403. I think merging this would cause that issue to regress.

huming2207:
> Hmm... we originally did not specify `REQUIRES main` because of https://github.com/lvgl/lvgl/issues/2403. I think merging this would cause that issue to regress.

Okay... what about PRIV_REQUIRES? I'll have a try tomorrow.

huming2207:
@embeddedt `PRIV_REQUIRES` works for me. 

Meanwhile if say both `PRIV_REQUIRES` or `REQUIRES` are both not doable, then what should I do to include the freetype dependency anyway? Maybe add freetype to `main` component? But then I guess LVGL still won't know the existence of the freetype...

embeddedt:
@C47D Could you take a look at this when you have some time?

C47D:
Hi, yes I will take a look at it.

embeddedt:
Not stale.

Issue #3149, opened 2022-03-05
fs_stdio: directory listing fails when using multiple directory handles (WIN32)


Ryzee119:
### Perform all steps below and tick them with [x]
- [x] Read the [FAQ](https://docs.lvgl.io/master/intro/index.html#faq)
- [x] Check the related part of the [Documentation](https://docs.lvgl.io/)
- [x] Update lvgl to the latest version
- [x] Reproduce the issue in a [Simulator](https://docs.lvgl.io/latest/en/html/get-started/pc-simulator.html)

### Describe the bug
Having multiple directory handles and calling lv_fs_dir_read between them will break the filename listings. Only tested in WIN32 backend, but it appears to be isolated here.

### To Reproduce
Use fs_stdio backend with WIN32
```
lv_fs_dir_t dir_a;
lv_fs_dir_t dir_b;

lv_fs_dir_open(&dir_a, "A:folder1");
lv_fs_dir_open(&dir_b, "A:folder2");
char fname[256];
lv_fs_dir_read(&dir_a, fname); //Read 1st file in dir_a;
lv_fs_dir_read(&dir_a, fname); //Read 2nd file in dir_a;

lv_fs_dir_read(&dir_b, fname); //Read 1st file in dir_b;

lv_fs_dir_read(&dir_a, fname); //SHOULD read 3rd file in dir_a; but it returns the 2nd file in dir_b.
```

### Expected behavior
You should be able to read from different directory handles and keep track of the current files in each directory independently.

Failing due to a shared `next_fn` buffer:
https://github.com/lvgl/lvgl/blob/ff0bf6386d702ff9f32528fe6296c8da326c9914/src/extra/libs/fsdrv/lv_fs_stdio.c#L192

kisvegabor:
Hi,

It's great that you have found the reason of this issue. 

Unfortunately, I'm on Linux and I don't really have experience with the WIN32 API so probably others can suggest a good solution. 

@mariotaku do you have any idea for this?

Ryzee119:
My fix was to create a directory handle struct which contained `next_fn` which is allocated on each `dir_open` call.

My diff is here:
https://github.com/Ryzee119/lvgl/commit/6d2b6726a384eb435f412e0ccd628ba4fc0eb156

kisvegabor:
Thanks @Ryzee119! Could you send a PR with the fix?

mariotaku:
[...]

PR #3243, opened 2022-03-31
Fix: Track multiple directory handles with win32 backends

Ryzee119:

Described in https://github.com/lvgl/lvgl/issues/3149.

Due to using a shared static `next_fn` buffer, tracking multiple directories at once was broken. This fix allocates a new `next_fn` buffer for each directory handle. On the win32 backend it also allocates a new `next_error` variable for each directory.
As this code is in the stdio backend and the win32 backend, there is two commits, one for each file.

I could not get the win32 backend to work at all for me, but I included some fixes and it works now. Namely we were calling `GetLastError()` in dir_open even when no error occured, so successful operations were reporting the last unrelated error that occured even if successful. https://docs.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-findfirstfilea implies we should only call this if the function fails or fails to locate files instead of all the time. This PR includes this fix.

### Checkpoints

Test code now works as expected:
```
lv_fs_dir_t dir_a;
lv_fs_dir_t dir_b;
lv_fs_dir_open(&dir_a, "A:folder1");
lv_fs_dir_open(&dir_b, "A:folder2");

char fname[256];
lv_fs_dir_read(&dir_a, fname); //Read 1st file in dir_a;
lv_log("dir_a, first file: %s\n", fname);
lv_fs_dir_read(&dir_a, fname); //Read 2nd file in dir_a;
lv_log("dir_a, second file: %s\n", fname);

lv_fs_dir_read(&dir_b, fname); //Read 1st file in dir_b;
lv_log("dir_b, first file: %s\n", fname);
lv_fs_dir_read(&dir_a, fname); //Read 3rd file in dir_a;
lv_log("dir_a, third file: %s\n", fname);
```
Output on win32 and stdio backends built with mingw:
```
dir_a, first file: 1file1.txt
dir_a, second file: 1file2.txt
dir_b, first file: 2file1.txt
dir_a, third file: 1file3.txt
```
Each folder contains 1file1/2/3.txt and 2file1/2/3.txt for folder1 and folder2 respectively.

### DIFF
--- a/src/extra/libs/fsdrv/lv_fs_stdio.c
+++ b/src/extra/libs/fsdrv/lv_fs_stdio.c
@@ -21,10 +21,19 @@
 /*********************
  *      DEFINES
  *********************/
+#define MAX_PATH_LEN 256
 
 /**********************
  *      TYPEDEFS
  **********************/
+typedef struct {
+#ifdef WIN32
+    HANDLE dir_p;
+    char next_fn[MAX_PATH_LEN];
+#else
+    DIR * dir_p;
+#endif
+} dir_handle_t;
 
 /**********************
  *  STATIC PROTOTYPES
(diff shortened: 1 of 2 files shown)


kisvegabor:
Thanks you very much!

PR #3252, opened 2022-04-05
fix(obj_style): prevent access to class null pointer

FASTSHIFT:
### Checkpoints


kisvegabor:
I don't think it can be NULL in the practice, but let's do it correctly :slightly_smiling_face:

PR #3253, opened 2022-04-05
feat(gif): GIF controls

filipbudisa:

Hello! We've recently started using this library in our products and fell in love with it. However, the need for more control around GIFs has arisen.

This PR adds the following functions for controlling GIF playback:
* lv_gif_start - start GIF playback
* lv_gif_stop - stop GIF playback
* lv_gif_set_loop - set loop mode:
  * DEFAULT - loop as many times as defined in the file
  * SINGLE - loop once, draw last frame after that
  * ON - loop indefinitely

The code is tested and already used in production for one of our devices.

The code already existing in lv_gif.h didn't contain much documentation so I didn't bother with it either. Let me know if I should add documentation for the new functions before you can merge the PR.

### Checkpoints

### DIFF
--- a/src/extra/libs/gif/lv_gif.c
+++ b/src/extra/libs/gif/lv_gif.c
@@ -85,6 +85,7 @@ void lv_gif_set_src(lv_obj_t * obj, const void * src)
     gifobj->imgdsc.header.h = gifobj->gif->height;
     gifobj->imgdsc.header.w = gifobj->gif->width;
     gifobj->last_call = lv_tick_get();
+    gifobj->loop = LV_GIF_LOOP_DEFAULT;
 
     lv_img_set_src(obj, &gifobj->imgdsc);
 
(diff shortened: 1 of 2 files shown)


X-Ryl669:
Unfortunately, we are currently reworking the image decoder and the image format. Please see #2904 for where it's leading.
Typically, everything you added here is included in `lv_img_set_play_mode`. Unlike rlottie (which is also an animation format), GIF does not support seeking in the animation (well, unless we start building a map of frame index => file's position), so not all playmode are supported (like stop at frame, or playing backward). Yet, looping (or not) and restarting works.

filipbudisa:
That sounds great. Does that mean you'll be able to specify a gif file as a source for lv_img, which will then also have controls mentioned here? Stopping at a specific frame isn't something that seems absolutely necessary for gifs, but will there be support for stopping the animation in general, making subsequent draw calls render the frame it stopped on?

X-Ryl669:
> Does that mean you'll be able to specify a gif file as a source for lv_img, which will then also have controls mentioned here?

Yes

> will there be support for stopping the animation in general, making subsequent draw calls render the frame it stopped on?

Yes there is. If you set the control flag in `set_play_mode` not to include `LV_IMG_CTRL_PLAYING` it'll stop and stay here.
You can't set the flag `BACKWARD` and you can't seek to a frame number except 0 for GIF files.

filipbudisa:
Awesome. Looking forward for the release of the new img interface.

Well then, I'll let the maintainers decide whether they want to merge this PR before the new img format is released or not.

Cheers!

X-Ryl669:
See #3294 for implementation.

PR #3254, opened 2022-04-05
Part 2. Add support for animation button widget

X-Ryl669:

This is the second part of the patch serie. This one is quite small since it's only adding a animation button widget.
Unlike what I did in my former PR, this widget support setting animations for both state and transition.

This means that one can play an animation when a widget is in some state (checked, released, etc...) with or without looping, forward or backward and within a specific keyframe range.

It can also play an animation once when transiting from one state to another (forward or backward, but no looping, obviously).
This should gives the best behavior for both intent. 

### Checkpoints

### DIFF
--- a/src/extra/widgets/animbtn/lv_animbtn.c
+++ b/src/extra/widgets/animbtn/lv_animbtn.c
@@ -39,7 +39,7 @@ static bool is_transiting(lv_animbtn_t * animbtn, lv_animbtn_state_t current_sta
  *  STATIC VARIABLES
  **********************/
 const lv_obj_class_t lv_animbtn_class = {
-    .base_class = &lv_obj_class,
+    .base_class = &lv_img_class,
     .instance_size = sizeof(lv_animbtn_t),
     .constructor_cb = lv_animbtn_constructor,
     .destructor_cb = lv_animbtn_destructor,
@@ -54,22 +54,13 @@ const lv_obj_class_t lv_animbtn_class = {
  *   GLOBAL FUNCTIONS
  **********************/
 
-/**
- * Create an image button object
- * @param parent pointer to an object, it will be the parent of the new image button
- * @return pointer to the created image button
- */
-lv_obj_t * lv_animbtn_create(lv_obj_t * parent, lv_obj_t * anim)
+lv_obj_t * lv_animbtn_create(lv_obj_t * parent)
 {
     LV_LOG_INFO("begin");
     lv_obj_t * obj = lv_obj_class_create_obj(MY_CLASS, parent);
     lv_obj_class_init_obj(obj);
-    /*Capture the animation picture*/
-    ((lv_animbtn_t *)obj)->img = anim;
-    lv_obj_set_parent(anim, obj);
-    lv_obj_add_flag(anim, LV_OBJ_FLAG_EVENT_BUBBLE);
 
-    lv_img_t * img = (lv_img_t *)anim;
+    lv_img_t * img = (lv_img_t *)obj;
     lv_obj_set_size(obj, img->w, img->h);
     return obj;
 }
@@ -188,14 +179,6 @@ static bool is_transiting(lv_animbtn_t * animbtn, lv_animbtn_state_t current_sta
 static void lv_animbtn_constructor(const lv_obj_class_t * class_p, lv_obj_t * obj)
 {
     LV_UNUSED(class_p);
-    lv_animbtn_t * animbtn = (lv_animbtn_t *)obj;
-    /*Initialize the allocated 'ext'*/
-    lv_memset_00(animbtn->state_desc, sizeof(animbtn->state_desc));
-    animbtn->img = NULL;
-    animbtn->prev_state = 0;
-    animbtn->trans_count = 0;
-    animbtn->trans_desc = 0;
-
     lv_obj_add_flag(obj, LV_OBJ_FLAG_CLICKABLE);
     lv_obj_add_flag(obj, LV_OBJ_FLAG_CHECKABLE);
 }
@@ -237,7 +220,7 @@ static void lv_animbtn_event(const lv_obj_class_t * class_p, lv_event_t * e)
             }
         case LV_EVENT_GET_SELF_SIZE: {
                 lv_point_t * p = lv_event_get_self_size_info(e);
-                p->x = LV_MAX(p->x, ((lv_img_t *)animbtn->img)->w);
+                p->x = LV_MAX(p->x, ((lv_img_t *)animbtn)->w);
                 break;
             }
         default:
@@ -251,17 +234,18 @@ static void setup_anim(lv_animbtn_t * animbtn, lv_animbtn_state_desc_t * desc)
     int backward = LV_BT(desc->control, LV_IMG_CTRL_BACKWARD);
     if(backward && desc->first_frame < desc->last_frame) {
         /*Play in reverse means start from last to first*/
-        lv_img_set_current_frame(animbtn->img, desc->last_frame);
-        lv_img_set_stop_at_frame(animbtn->img, desc->first_frame, !backward);
+        lv_img_set_current_frame((lv_obj_t *)animbtn, desc->last_frame);
+        lv_img_set_stop_at_frame((lv_obj_t *)animbtn, desc->first_frame, !backward);
     }
     else {
-        lv_img_set_current_frame(animbtn->img, desc->first_frame);
-        lv_img_set_stop_at_frame(animbtn->img, desc->last_frame, !backward);
+        lv_img_set_current_frame((lv_obj_t *)animbtn, desc->first_frame);
+        lv_img_set_stop_at_frame((lv_obj_t *)animbtn, desc->last_frame, !backward);
     }
 }
 
 static void loop_state(lv_obj_t * obj)
 {
+    if(!obj) return;
     lv_animbtn_t * animbtn = (lv_animbtn_t *)obj;
     lv_animbtn_state_t current_state = get_state(obj);
     lv_animbtn_state_t state  = suggest_state(obj, current_state);
@@ -270,7 +254,7 @@ static void loop_state(lv_obj_t * obj)
         animbtn->prev_state = state;
     }
 
-    if(animbtn->prev_state != state || animbtn->img == NULL || !is_state_valid(&animbtn->state_desc[state - 1])) return;
+    if(animbtn->prev_state != state || !is_state_valid(&animbtn->state_desc[state - 1])) return;
 
     /*Set the logic for the current state*/
     if(LV_BT(animbtn->state_desc[state - 1].control, LV_IMG_CTRL_LOOP)) {
@@ -281,6 +265,7 @@ static void loop_state(lv_obj_t * obj)
 
 static void apply_state(lv_obj_t * obj, bool skip_transition)
 {
+    if(!obj) return;
     lv_animbtn_t * animbtn = (lv_animbtn_t *)obj;
     lv_animbtn_state_t current_state = get_state(obj);
     lv_animbtn_state_t state  = suggest_state(obj, current_state);
@@ -296,7 +281,7 @@ static void apply_state(lv_obj_t * obj, bool skip_transition)
         }
     }
 
-    if(state == animbtn->prev_state || animbtn->img == NULL || !is_state_valid(&animbtn->state_desc[state - 1])) return;
+    if(state == animbtn->prev_state || !is_state_valid(&animbtn->state_desc[state - 1])) return;
 
     /*Set the logic for the current state*/
     setup_anim(animbtn, &animbtn->state_desc[state - 1]);
--- a/src/extra/widgets/animbtn/lv_animbtn.c
+++ b/src/extra/widgets/animbtn/lv_animbtn.c
@@ -59,9 +59,6 @@ lv_obj_t * lv_animbtn_create(lv_obj_t * parent)
     LV_LOG_INFO("begin");
     lv_obj_t * obj = lv_obj_class_create_obj(MY_CLASS, parent);
     lv_obj_class_init_obj(obj);
-
-    lv_img_t * img = (lv_img_t *)obj;
-    lv_obj_set_size(obj, img->w, img->h);
     return obj;
 }
 
--- a/src/extra/widgets/animbtn/lv_animbtn.c
+++ b/src/extra/widgets/animbtn/lv_animbtn.c
@@ -1,0 +1,380 @@
+/**
+ * @file lv_animbtn.c
+ *
+ */
+
+/*********************
+ *      INCLUDES
+ *********************/
+
+#include "lv_animbtn.h"
+
+#if LV_USE_ANIMBTN != 0
+
+/*********************
+ *      DEFINES
+ *********************/
+#define MY_CLASS &lv_animbtn_class
+
+/**********************
+ *      TYPEDEFS
+ **********************/
+
+/**********************
+ *  STATIC PROTOTYPES
+ **********************/
+static void lv_animbtn_constructor(const lv_obj_class_t * class_p, lv_obj_t * obj);
+static void lv_animbtn_destructor(const lv_obj_class_t * class_p, lv_obj_t * obj);
+static void lv_animbtn_event(const lv_obj_class_t * class_p, lv_event_t * e);
+static void apply_state(lv_obj_t * animbtn, bool skip_transition);
+static void loop_state(lv_obj_t * animbtn);
+static lv_animbtn_state_t suggest_state(lv_obj_t * animbtn, lv_animbtn_state_t state);
+static lv_animbtn_state_t get_state(const lv_obj_t * animbtn);
+static void setup_anim(lv_animbtn_t * animbtn, lv_animbtn_state_desc_t * desc);
+static int is_state_valid(const lv_animbtn_state_desc_t * state);
+static uint8_t find_trans(lv_animbtn_t * animbtn, lv_animbtn_state_t from, lv_animbtn_state_t to);
+static bool is_transiting(lv_animbtn_t * animbtn, lv_animbtn_state_t current_state);
+
+/**********************
+ *  STATIC VARIABLES
+ **********************/
+const lv_obj_class_t lv_animbtn_class = {
+    .base_class = &lv_obj_class,
+    .instance_size = sizeof(lv_animbtn_t),
+    .constructor_cb = lv_animbtn_constructor,
+    .destructor_cb = lv_animbtn_destructor,
+    .event_cb = lv_animbtn_event,
+};
+
+/**********************
+ *      MACROS
+ **********************/
+
+/**********************
+ *   GLOBAL FUNCTIONS
+ **********************/
+
+/**
+ * Create an image button object
+ * @param parent pointer to an object, it will be the parent of the new image button
+ * @return pointer to the created image button
+ */
+lv_obj_t * lv_animbtn_create(lv_obj_t * parent, lv_obj_t * anim)
+{
+    LV_LOG_INFO("begin");
+    lv_obj_t * obj = lv_obj_class_create_obj(MY_CLASS, parent);
+    lv_obj_class_init_obj(obj);
+    /*Capture the animation picture*/
+    ((lv_animbtn_t *)obj)->img = anim;
+    lv_obj_set_parent(anim, obj);
+    lv_obj_add_flag(anim, LV_OBJ_FLAG_EVENT_BUBBLE);
+
+    lv_img_t * img = (lv_img_t *)anim;
+    lv_obj_set_size(obj, img->w, img->h);
+    return obj;
+}
+
+/*=====================
+ * Setter functions
+ *====================*/
+
+/**
+ * Set images for a state of the image button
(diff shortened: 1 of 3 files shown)


kisvegabor:
> It can also play an animation once when transiting from one state to another (forward or backward, but no looping, obviously).


It's sounds amazing.  So (probably with some work in the event handler) we do:
- loop 0-100 in released
- loop 200-300 in checked
- and move from the current frame to 200 when going to checked

Is it correct?

If so, maybe a special frame like `FRAME_CURRENT` can be added to make all these automatic.

X-Ryl669:
> Is it correct?

Yes it is.
You'll set the button this way:
```cpp
lv_obj_t * img = lv_img_create(parent);
lv_img_set_src(img, "/path/to/lottie.json");
lv_obj_t * anim = lv_animbtn_create(parent, img);
lv_animbtn_state_desc_t desc;

// Set release animation
desc.from = 0;
desc.to = 100;
desc.control = LV_ANIMBTN_CTRL_FORWARD | LV_ANIMBTN_CTRL_LOOP;
lv_anim_set_state_desc(anim, LV_ANIMBTN_STATE_RELEASED,  desc);

// Set checked animation
desc.from = 200;
desc.to = 300;
desc.control = LV_ANIMBTN_CTRL_FORWARD;
lv_anim_set_state_desc(anim, LV_ANIMBTN_STATE_CHECKED_RELEASED,  desc);

// Set transition from released to checked
desc.from = 100;
desc.to = 200;
desc.control = LV_ANIMBTN_CTRL_FORWARD;
lv_anim_set_transition_desc(anim, LV_ANIMBTN_STATE_RELEASED, LV_ANIMBTN_STATE_CHECKED_RELEASED,  desc);
```

You can fetch the current frame with `lv_img_get_current_frame` if you need to (when creating the descriptor). You'll need to listen to EVENT_READY to change the descriptor.

But it's not always a good idea to start from the current frame, if, like I'm doing, you store unrelated animation in the same lottie file, it's much better to have a true start point for the animation (so it's not glitching on the screen), like this:
image

src/extra/widgets/animbtn/lv_animbtn.c
```diff
@@ -0,0 +1,380 @@
...
+static void setup_anim(lv_animbtn_t * animbtn, lv_animbtn_state_desc_t * desc);
+static int is_state_valid(const lv_animbtn_state_desc_t * state);
+static uint8_t find_trans(lv_animbtn_t * animbtn, lv_animbtn_state_t from, lv_animbtn_state_t to);
+static bool is_transiting(lv_animbtn_t * animbtn, lv_animbtn_state_t current_state);
+
+/**********************
+ *  STATIC VARIABLES
+ **********************/
+const lv_obj_class_t lv_animbtn_class = {
+    .base_class = &lv_obj_class,
```
kisvegabor: Can is be based on `lv_img` and have functions like  `lv_animbtn_create(parent)` and `lv_animbtn_set_src()` (wrapper to `lv_ing_set_src`)?

src/extra/widgets/animbtn/lv_animbtn.c
```diff
@@ -0,0 +1,380 @@
...
+
+/**********************
+ *   GLOBAL FUNCTIONS
+ **********************/
+
+/**
+ * Create an image button object
+ * @param parent pointer to an object, it will be the parent of the new image button
+ * @return pointer to the created image button
+ */
```
kisvegabor: API comment should be only in header files.

src/extra/widgets/animbtn/lv_animbtn.h
```diff
@@ -0,0 +1,143 @@
...
+
+#if LV_USE_ANIMBTN != 0
+
+/**********************
+ *      TYPEDEFS
+ **********************/
+typedef enum {
+    LV_ANIMBTN_STATE_RELEASED = 1,          /* 0001 */
+    LV_ANIMBTN_STATE_PRESSED  = 2,          /* 0010 */
+    LV_ANIMBTN_STATE_DISABLED = 3,          /* 0100 */
```
kisvegabor: `3 != /* 0100 */`

src/extra/widgets/animbtn/lv_animbtn.h
```diff
@@ -0,0 +1,143 @@
...
+    LV_ANIMBTN_STATE_CHECKED_RELEASED = 4,  /* 1001 */
+    LV_ANIMBTN_STATE_CHECKED_PRESSED  = 5,  /* 1010 */
+    LV_ANIMBTN_STATE_CHECKED_DISABLED = 6,  /* 1100 */
+    _LV_ANIMBTN_STATE_NUM = 6,
+} lv_animbtn_state_t;
+
+
+typedef enum {
+    LV_ANIMBTN_CTRL_FORWARD  = LV_IMG_CTRL_FORWARD,
```
kisvegabor, line 35: Can't we use `LV_IMG_CTRL_...` directly?

kisvegabor:
>   You can fetch the current frame with lv_img_get_current_frame if you need to (when creating the descriptor). You'll need to listen to EVENT_READY to change the descriptor.

I meant this:
```
desc.from = FRAME_CURRENT;   /*Can be anywhere in 0-100, but will move smoothly to 200*/
desc.to = 200;
desc.control = LV_ANIMBTN_CTRL_FORWARD;
lv_anim_set_transition_desc(anim, LV_ANIMBTN_STATE_RELEASED, LV_ANIMBTN_STATE_CHECKED_RELEASED,  desc);
```


> But it's not always a good idea to start from the current frame, if, like I'm doing, you store unrelated animation in the same lottie file, it's much better to have a true start point for the animation

So I meant it only for transitions. 

> (so it's not glitching on the screen), like this:

Have you created it with LVGL?

src/extra/widgets/animbtn/lv_animbtn.h
X-Ryl669: The bit pattern here is used to track the transition via a graycode (that was my initial idea, to check if a transition applies with a XOR, instead of "if from == prev_state && to == new_state"). But in the end there's only few states so the complexity of gray coding for this optimization is too large for the expected gain. I'll remove the comment, since it doesn't fit anymore.

src/extra/widgets/animbtn/lv_animbtn.h
X-Ryl669: No, I don't think so. There's only a few control in IMG that can be used in ANIMBTN, so if I remove this enum, devs will wonder why some controls don't work. I think it's better to only list those that are expected to work here and have them named differently.

src/extra/widgets/animbtn/lv_animbtn.c
X-Ryl669: Ok

src/extra/widgets/animbtn/lv_animbtn.c
X-Ryl669: Yes that's possible. `lv_animbtn_set_src` will be useless if the first element of the struct is a `lv_img_t`, one can use `lv_img_set_src` directly then.

In that case, I think it's better to use lv_img's API directly, since one my want to change the size, etc... and I don't want to replicate the complete API with `lv_anim_` prefix.

X-Ryl669:
What I meant is that FRAME_CURRENT does not really make sense. Let me explain with an example here:
1.  Let's say you have an animation with 2 button states stored in the animation. Frame from 1 to 100 are for one state (imagine an empty button glowing) and frame from 100 to 200 is another state (a checked button glowing).
2. Transition for one state to another is stored in frame 200 to 250 (a empty button starting to fill up until it's full)
3. When going from state 1 to state 2, you need to play transition animation from 200 to 250
4. When going from state 2 to state 1, you need to play transition animation from 250 to 200 

If you are in the state 2, let's say at frame 150 and the button needs to change state, it's not a good idea to play from 150 to 1 (since you'll be in the middle of the state 2's animation and you'll be playing backward to the first frame, crossing unrelated animations). You'll instead want to jump straight to frame 250 and play back to frame 200, than jump straight to frame 1 and play to 100.

You'll have a glitch from frame 150 to frame 250 (in my example, the button glowing animation will directly skip to the button full), but since it's clearly designed this way, it's very likely it'll fit as expected.

If you look carefully in the GIF above, you'll see what I mean when the guy pushing on his legs scrolls to the top. The legs are extended but jump straight to the leg folded when the animation transit to the non pressed state (when scrolling up). It's very subtle so it's almost perfect.

If we had a FRAME_CURRENT for the transition, we would have see legs playing backward (it would have glitched) so it would have been very disturbing here since the movement is a cycle and it would have inverted. 

I haven't made this animation, it's an example in Samsung rlottie repository. It's completely possible to make it in LVGL with #2904 since we have the vector capability in the decoders (so it's possible to resize the picture perfectly while zooming/scrolling).

src/extra/widgets/animbtn/lv_animbtn.h
kisvegabor: Fair enough.

kisvegabor:
Got it, but imagine this:
- state1: 0-100 (pulsing red)
- transtition12: 100-200 (red to blue)
- state2: 200-300 (pulsing blue)

If we are now in state1 at frame 68 and going to state2: 
```
68->                100->200->               loop(200-300)
finish state 1      state1->state2 trans.    loop state2
``` 

makes sense.

If also we had
- transtition21: 300-400 (blue to red and 400 looks the same image as 0)

 state2->state1 worked as: 
```
248->               300->400->                loop(0-100)
finish state 2      state2->state1 trans.     loop state1
```

X-Ryl669:
BTW, in your example, the transition21 is useless and it's unlikely it'll be stored in the initial file. You can use transition12 and play backward. So in the end, without continuity, you'll get a glitch from 248 to 200 then it's smooth until 100 and then state 1.

But I got your point.

Maybe it ends up in having a "play until current animation is done or play backward until start of animation or skip to frame immediately" feature (3 states to choose from) so it does not depends on the position of the transition frames in the animation. In all cases, it's already possible to do so, one must listen to EVENT_READY event and change the transition to whatever is required.

kisvegabor:
>  it's already possible to do so, one must listen to EVENT_READY event and change the transition to whatever is required.

True, but in practice it's a little bit difficult to handle:
1. Click event received: Save that we need to go to checked state
2. Ready event received: If we are about going to checked state, set new dsc-s and change widget state here.

> "play until current animation is done or play backward until start of animation or skip to frame immediately" 

I agree that the bottom line is this. However somewhere it needs to be stored to"start the transition after the current loop is finished". Maybe it's just a bit like `wait_for_loop_ready_before_transitioning` which calls `apply_state()` in the READY event instead of PRESS/RELEASE etc.

What do you think?

It's really no critical, so we can postpone it.

src/extra/widgets/animbtn/lv_animbtn.c
```diff
@@ -0,0 +1,365 @@
...
+ *   GLOBAL FUNCTIONS
+ **********************/
+
+lv_obj_t * lv_animbtn_create(lv_obj_t * parent)
+{
+    LV_LOG_INFO("begin");
+    lv_obj_t * obj = lv_obj_class_create_obj(MY_CLASS, parent);
+    lv_obj_class_init_obj(obj);
+
+    lv_img_t * img = (lv_img_t *)obj;
```
kisvegabor: Why do you need to an `lv_obj` wrapper?

X-Ryl669: I'm not really sure how the object inheritance works in LVGL.
To be able to use lv_img's API transparently, I made `lv_animbtn` a child of `lv_img`.

I want to create an instance of lv_img_t here by calling its constructor and then just fixing the additional members.
So I've set `MYCLASS` to refer as lv_img's class for the base class, but I'm calling `lv_obj_class_create_obj/class_init_obj` here. Is it right ?

I guess the 2 first lines above will call animbutton_constructor (it should since I've specified it in MYCLASS right ?)
But `lv_obj_class_create_obj` returns a `lv_obj_t*` and I need to access the `lv_img_t`'s width and height member here (which might not be those of the object, for example for vector picture). I've changed this compared to #2904, so it might be modified when I merge the last part of this PR in this serie.

Please correct if I'm missing something here.

X-Ryl669:
> What do you think?
> 
> It's really no critical, so we can postpone it.

You are right, it's probably better done this way. I'll add a "action before transition" member to the transition structure (as an enum as specified above), and a "flag" to the animbtn to tell not to transition immediately (that'll be checked on the EVENT_READY & PRESS/RELEASED). But I'll do that after the next PR which fix the decoders for the new API. So it'll be possible to test & adjust this correctly.

src/extra/widgets/animbtn/lv_animbtn.c
X-Ryl669: Ping ?

kisvegabor: Sorry, for the late reply. I was ill at the end of the previous week and just got back to work.

You can take a look at `lv_slider` which is inherited from `lv_bar`, which is inherited from `lv_obj`. 

>  I'm calling lv_obj_class_create_obj/class_init_obj here. Is it right ?

Yes.

> I guess the 2 first lines above will call animbutton_constructor 

It will call `lv_obj_constructor` first, and then `lv_img_construstor` and finally `lv_animbtn_constructor`.

> But lv_obj_class_create_obj returns a lv_obj_t* and I need to access the lv_img_t's width and height member

You can do this:
```c
lv_animbtn_t * animbtn = (lv_animbtn_t *) obj;
animbtn->img.w;
```

> which might not be those of the object, for example for vector picture

Do you mean `animbtn->img.w` and `animbtn->img.h` won't be set in case of vector images?

X-Ryl669: Before the last change, one has to pass a fully constructed img object to the create function. So the image's width and height were supposedly set (because of previous call to `lv_img_set_src`). The anim button contained an image so the image's size was used to set the button's size.

Now, the image object isn't constructed beforehand, so I must provide a mechanism for this. 
By looking at the image decoder PR, I've removed all image's decoder specific code bu lv_*pic*_init() except for rlottie.

So it's perfectly fine to let the user call `lv_img_set_src` on a `lv_animbtn_t` object. I'll need to add a parameter in the function `lv_img_set_src_ext` to specify a image decoder's context to use while constructing (typically, that's where we set the image's preferred rendering size for vector format, and the image's buffer size and so on).

kisvegabor:
> I'll need to add a parameter in the function lv_img_set_src_ext to specify a image decoder's context to use while constructing (typically, that's where we set the image's preferred rendering size for vector format, and the image's buffer size and so on).

Regarding preferred size: if the animbtn has `LV_SIZE_CONTENT` size can't we use the intrinsic size of the vector (or non-vector) image? If the widget's size is set explicitly (e.g. to 100x200) the vector images can scale accordingly.

What is an "image's buffer size"?

X-Ryl669:
The issue with the size is something we've talked few month ago. If I use img's declared size, there's no need for specifying the size so the default will work fine. However, many (if not all) lottie files have very big native size (because they are designed for high dpi screen, which we don't have in embedded world), so one is likely to prefer a smaller rendering size than the declared size. 

How to set the size:
1. __Scenario 1__ Set the size with `lv_obj_set_width/height`. This is possible but unfortunately, it's not available until the loop run once and process events to compute the true size, IIUC. So if the code call `lv_obj_set_width(...); lv_img_set_src(...)` we don't know the size... yet. If we need the user to call `lv_obj_invalidate` in between, most of the time, it'll be forgotten and will fail
2. __Scenario 2__ Set the size by providing a the decoder context in `lv_img_set_src`. This is the most convenient way. The object is in `SIZE_CONTENT` mode and the decoder will find the context and use the specified size. This scheme is easily extendable for the other parameters (like the rendering buffer size, the first frame index, etc...)
3. __Scenario 3__ Set the size by changing the img's members. It's not really the LVGL way of working to deal with internal members of the structure. This cause a penalty of storing a useless image in the cache (the one with the default size). This works for the size, but not for other parameters (like rendering buffer size...)

As for the rendering buffer size, in rlottie, there's a size limit where the picture is rendered to a buffer that's not few lines high.
If the rendering buffer is large enough to store the complete image (old `img_data` field if you remember), then it's rendered once to this buffer and no line by line rendering is used. Also, in that case, the rotation and zoom works on this cached image.

Ideally, in a future enhancement, we could pass the rotation angle and zoom factor to the rendering engine in rlottie for all cases, but it's the way it's done now. I've thought about this, but it implies changing the size of the image dynamically (when rotating, the width & height changes to follow a circle). I don't see how it can be handled in LVGL.

There's also all the other, per decoder parameters, like the first frame to render on (can be done manually with `lv_img_set_current_frame`), whether to store the rlottie parser structure (parsing the JSON file takes 10x the rendering time, so it's better to cache the parser structure), the ffmpeg av_XXX context structures and so on. 

Providing plenty of functions for this is too much work IMHO, so it's better to embed this into an opaque decoder context's `user_data` pointer, so that's how I've done it in #2904. `lv_rlottie_...` just access this context to store/retrieve info.

kisvegabor:
> Scenario 1 Set the size with lv_obj_set_width/height. This is possible but unfortunately, it's not available until the loop run once and process events to compute the true size, IIUC. So if the code call lv_obj_set_width(...); lv_img_set_src(...) we don't know the size... yet. If we need the user to call lv_obj_invalidate in between, most of the time, it'll be forgotten and will fail

If you need know the final size for some reason you can force LVGL to recalculate to update the layouts by calling `lv_obj_update_layout(img);`. If you can call it only for vector images (i.e. not on all `lv_img_set_src`) it might work. What do you think?

X-Ryl669:
That might work, but IMHO, that's only part of the solution. Right now, the image size is read from the object when rendering so at worst (without dealing with image size at parsing time with update_layout), it'll cause a useless caching of the wrong sized image. Also, there's some policy to define here. If the object has a size, it can come from either being set after creation and before img_set_src, **or** from a previous img_set_src call (if you change the src, should we use the previous object size or overwrite the object size with the image's size ?).

Right now, the default is to use the image own size to set the obj size. I think it's better not to change this. If the user need to specify an image size (for rendering vector images or not), she can do it after calling img_set_src (but it's not optimal) or specify the *additional context* when setting the source (for vector images or ffmpeg's swscale).

kisvegabor:
>  Right now, the image size is read from the object when rendering so at worst (without dealing with image size at parsing time with update_layout), it'll cause a useless caching of the wrong sized image.

The layouts are recalculated **before** rendering. In more detail, before rendering all `lv_pct()` and `LV_SIZE_CONTENT` (using `LV_EVENT_GET_SELF_SIZE`) values and other layout constrains are resolved, and  positions are calculated. When all objects store the correct absolute coordinates in `obj->coords` LVGL starts to render them. So when rendering starts the image size should be already known as it was used  in  `LV_EVENT_GET_SELF_SIZE` earlier.

> Also, there's some policy to define here. If the object has a size, it can come from either being set after creation and before img_set_src, or from a previous img_set_src call (if you change the src, should we use the previous object size or overwrite the object size with the image's size ?).

Calling `lv_img_set_src` shouldn't set the width/height directly (using `lv_obj_set_width/height()`). That is, if the user sets `lv_obj_set_width(img, 200)` it should remain 200 regardless to the `src`. If the user sets `LV_SIZE_CONTENT` width, the width should be updated for every new `src`.

X-Ryl669:
But if the user sets `LV_SIZE_CONTENT` then there no way to specify the rendering width of a vector image explicitly. Right ?

kisvegabor:
Correct! Is it a problem?

X-Ryl669:
Here's a sequence of code that show the problem:

1. lv_obj_t * obj = lv_img_create(parent);
2. lv_obj_set_width&height(obj, w, h);
3. lv_img_set_src(obj, "vectorFile")
3.1 Since it's a vector file, this calls update_layout()
3.2 img.w = coords.width...
3.3 Accept the source and set the dimension, the size hint is w & h
4. lv_obj_set_width&height(obj, size content, size content)
5. lv_img_set_src(obj, "otherLottie")
5.1 update_layout()  => will use the previous image's size here because at this point, the object isn't mutated yet, so the content size is img.w & h
5.2 img.w = coords.w...
5.3 Here dimensions are set to the **previous image size**, not the image intrinsic size

X-Ryl669:
Anyway, this isn't related to this PR, but the next one with the decoder change. We should continue discussing in the next PR I think.

kisvegabor:
My version:
1. lv_obj_t * obj = lv_img_create(parent);
2. lv_obj_set_size(obj, 100, 200);
3. lv_img_set_src(obj, "vectorFile")
  3.1 Accept src, but do nothing special here 
4. LVGL's automatic layout update will set `img->coords` according to the set `width=100`, `height=200`
5. During drawing the image (`LV_EVENT_DRAW_MAIN`) use `img->coords` to set the desired render size
6. lv_obj_set_size(obj, size content, size content)
7. lv_img_set_src(obj, "otherLottie")
  7.1 Accept src, but do nothing special here 
8. LVGL's automatic layout update will set `img->coords` based on `LV_EVENT_GET_SELF_SIZE`
9. During drawing the image (`LV_EVENT_DRAW_MAIN`) use `img->coords` to set the desired render size


Okay, I merge it and we can fine tune later.

PR #3257, opened 2022-04-06
fix(demo): added gaurd to prevent crash if default theme is not used

peterson79:
### Fixes compilation error when LV_USE_THEME_DEFAULT=0

When LV_USE_THEME_DEFAULT=0 "lv_theme_default_init()" is not built.  However, lv_theme_default_init() is referenced in "demos/widgets/lv_demo_widgets.c" without checking if LV_USE_THEME_DEFAULT is set.  A check for this case was added.


kisvegabor:
Thanks!

Issue #3207, opened 2022-03-25
LV_USE_GRIDNAV missing in Kconfig


itavero:
### Describe the bug
I'm using LVGL as a component in ESP-IDF and I noticed that I can't enable Grid navigation via menuconfig.
Upon further investigation, it turns out `LV_USE_GRIDNAV` is missing from `Kconfig`.

kisvegabor:
Hi,

Could you send a Pull request?

itavero:
Sure. I'll create one.

kisvegabor:
Fixed by #3270.

Thank you!

PR #3270, opened 2022-04-13
fix(Kconfig): Add LV_USE_GRIDNAV and LV_USE_FRAGMENT to Kconfig

itavero:
Add `LV_USE_GRIDNAV` to Kconfig, which was missing as mentioned in #3207.
Also noticed that `LV_USE_FRAGMENT` was missing, so I've also added that one.


kisvegabor:
Thank you!

Issue #205, opened 2018-04-13
Does this Gui support touch gestures slide?


qwert1213131:
As stated in the title

kisvegabor:
Some object supports it. For example in `lv_tabview` you can change between tabs with slides.

Try this:
```c
    lv_theme_t * th = lv_theme_alien_init(210, NULL);
    lv_test_theme_1(th);
```

`lv_examples` will be required.

qwert1213131:
Referring to #204, I find lv_page is more suitable for my needs. Thank you.

PR #3278, opened 2022-04-16
feat(cmake-build): Option to allow building shared libraries.

ghost:
Add option in CMake build to specify that shared libraries should be created (as
opposed to static). This is an exclusive option (not enabled by default).

I'm by no means a CMake specialist (or even fan), but I believe this is the (current) recommended method.


kisvegabor:
Thanks! Could you also add note about these config options in the [docs](https://github.com/lvgl/lvgl/blob/master/docs/get-started/platforms/cmake.md)?

ghost:
Updated.

* A similar PR for the same functionality has also been made on the lv_driver repository ([#205](https://github.com/lvgl/lv_drivers/pull/205l))

kisvegabor:
Looks good, thank you!

PR #3283, opened 2022-04-18
chore(ci): add informational Codecov status checks

thomasrockhu-codecov:

Hi, Tom from Codecov here. I noticed that you were using Codecov but weren't actually getting any notifications on pull requests. I figured it would be useful to get some idea if code being changed is being tested, but also not blocking CI/merging. Let me know if this makes sense or if we can do something that would be helpful.

### Checkpoints

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -4,5 +4,9 @@ codecov:
 comment: off
 coverage:
   status:
-    patch: off
-    project: off
+    patch:
+      default:
+        informational: true
+    project:
+      default:
+        informational: true


kisvegabor:
Hi Tom,

Thanks for opening the PR. 

We have disabled notifications because the coverage is still quite low. We were planning to enable them once we have good (~90%) coverage. 

cc @embeddedt

kisvegabor:
I'd like to wait with enabling coverage report until all core functionalities get unit tests. After that it will be reasonable to ask for tests too in PRs.

embeddedt:
I agree, otherwise every contribution will likely be accompanied by the Codecov action reporting reduced coverage. If I understand correctly, once it's enabled, it requires any PR to have the same or increased coverage compared to the base branch.

embeddedt:
Let's wait for some more tests to be added as @kisvegabor suggested and then we can merge this.

PR #3284, opened 2022-04-19
feat(fsdrv) fix issues for win32 backends

MouriNaruto:
In the last changes from win32 backends, some APIs use the `wchar_t` version instead of `char` version. It will cause bugs, so I have fixed it.

Kenji Mouri


MouriNaruto:
@kisvegabor ping for progress

Kenji Mouri

kisvegabor:
Sorry, it seems I missed this PR.

PR #3317, opened 2022-05-02
fix(png):failed to compile using arm-none-eabi-gcc

lihuanhuan:

in arm-none-eabi-gcc ,uint32_t is defined to long unsigned int,it saying "expected 'unsigned int *' but argument is of type 'uint32_t *' {aka 'long unsigned int *'}"

### Checkpoints

### DIFF
--- a/src/extra/libs/png/lv_png.c
+++ b/src/extra/libs/png/lv_png.c
@@ -147,8 +147,8 @@ static lv_res_t decoder_open(lv_img_decoder_t * decoder, lv_img_decoder_dsc_t * 
             }
 
             /*Decode the PNG image*/
-            uint32_t png_width;             /*Will be the width of the decoded image*/
-            uint32_t png_height;            /*Will be the width of the decoded image*/
+            unsigned png_width;             /*Will be the width of the decoded image*/
+            unsigned png_height;            /*Will be the width of the decoded image*/
 
             /*Decode the loaded image in ARGB8888 */
             error = lodepng_decode32(&img_data, &png_width, &png_height, png_data, png_data_size);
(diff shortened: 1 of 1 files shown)


embeddedt (dismissed):
Changing to `unsigned` is not safe as it doesn't specify the exact size of the integer like `uint32_t` does.

Which function expects `unsigned int *`? That should possibly be changed to accept `uint32_t *` instead.

lihuanhuan:
In the lodepng,all unsigned is considered as long unsigned int in arm-none-eabi-gcc
example:
unsigned lodepng_decode32(unsigned char** out, unsigned* w, unsigned* h,const unsigned char* in, size_t insize);

embeddedt:
I see. It indeed appears that lodepng makes extensive use of this convention. I wasn't aware of that when I wrote my previous comment, sorry.

@kisvegabor and/or @xiaoxiang781216, what do you think we should do about this?

xiaoxiang781216:
> I see. It indeed appears that lodepng makes extensive use of this convention. I wasn't aware of that when I wrote my previous comment, sorry.
> 
> @kisvegabor and/or @xiaoxiang781216, what do you think we should do about this?

either uint32_t or unsigned int is fine, but it's wrong to mix both since uint32_t may map to other type(e.g. unsigned long) on some platform.

kisvegabor:
`unsigned` without `int` seems really odd to me but it's standard. At least in `lv_png.c` we should use `unsigned int`.

IMO it's not a problem if it maps to different size on different platforms because any of the possible options provide enough room for the width and height of the image.

So suggest using `unsigned int`. :slightly_smiling_face:

kisvegabor:
Why is it closed?

embeddedt:
It must have been a GitHub issue. I never received a notification about anyone closing it.

kisvegabor:
Hm, strange. 

Anyway, are you ok with `unsigned int`?

embeddedt:
:+1:

embeddedt:
@lihuanhuan Can you please update this to use `unsigned int` instead of `unsigned` for convention reasons? Then this is ready to be merged.

Issue #3052, opened 2022-01-27
Attach the NXP PXP/VGLITE GPUs back to the new draw context (architecture changes)


nicusorcitu:
### Perform all steps below and tick them with [x]
- [x] Read the [FAQ](https://docs.lvgl.io/master/intro/index.html#faq)
- [x] Check the related part of the [Documentation](https://docs.lvgl.io/)
- [x] Update lvgl to the latest version
- [ ] Reproduce the issue in a [Simulator](https://docs.lvgl.io/latest/en/html/get-started/pc-simulator.html)

### Describe the bug
Draw context architecture changed and now the GPU rendering is done by implementing the supported acceleration callbacks.
The NXP GPUs hooks are gone and we need to add the support back.

### To Reproduce
NXP GPUs support is currently broken on the master branch.

### Expected behavior
NXP PXP or VGLITE can be enabled and used as accelerators.

nicusorcitu:
I am currently working on this issue, the PXP attachment is complete and tested.

kisvegabor:
Thank you for opening this issue and working on it. 
Feel free to send a PR so that I can also try it out.

gokhannsahin:
I wonder why PXP and VGlite are limited to 16bits and why rotating is not used with vglite? The performance of IMXRT1170 is lower than expected because the gpu drivers poor.

gokhannsahin:
Also, using CPU to rotate is hungry, why don't use VGLite to rotate any image?

nicusorcitu:
Thank you for your interest to the NXP products. We were about to update the GPU with more functionalities but now we have to rework the patches due to the new architecture.

We will add soon the support for ARGB 32 bit color format on both PXP and VGLITE, hopefully at the end of this ticket. Also the image rotation is in progress.

We will continue to add even more GPU acceleration support so that the performance gets better and better.

gokhannsahin:
That is good news.
Also, multilayer features can be added default so it provides very good performance using LCDifv2 alpha blendig.

nicusorcitu:
We are also studying if chroma key, alpha blending, multi-layers alpha blending can be accelerated using LCDIFv2.

kisvegabor:
@nicusorcitu 
What is the current state of the new drivers?

gokhannsahin:
I'm looking forward to your reply. Please upgrade the driver for new version because we don't use it in v8.2.

nicusorcitu:
[...]

Issue #168, opened 2018-03-16
Letter space rendering is inconsistent


BenQoo:
The rendering letter space could be inconsistent on certain characters.
But, for the most of characters space are fine, so to adjust `letter_space` of `lv_style_t` could not solve this issue properly.

* LVGL version : 5.1
* Font Name : [NotoSans-SemiCondensedMedium.ttf](https://noto-website-2.storage.googleapis.com/pkgs/NotoSans-hinted.zip)
* Height :18 
* [lv_font_notosans_semicondensedmedium_18.zip](https://github.com/littlevgl/lvgl/files/1818188/lv_font_notosans_semicondensedmedium_18.zip)

Please refer to the following screenshots
* Red arrows: space is too tight
* Blue arrows: space is too loose.

Screenshot
![100%](https://user-images.githubusercontent.com/29389982/37507264-9a4c8828-2928-11e8-8258-be5c1aff735d.png)
Screenshot with zoom in 600%
![600%](https://user-images.githubusercontent.com/29389982/37507313-d4c52794-2928-11e8-911d-987f4a277b79.png)

kisvegabor:
Hi,

Thank you for the detailed report!

As I see you have used `style.text.letter_space = 0` so it should be no space between letters. If you look carefully the word "current" between 'r' and 'e' there is no space because 'r' has a very light pixel on the right and 'e' has light pixels on the left. 

This issue is typically solved by [Kerning](https://en.wikipedia.org/wiki/Kerning) but it is not implemented in LittlevGL yet. Kerning uses tables to modify the letter space based on letter pairs.

BenQoo:
Hi,

Thanks for your explanation, I got it.

> As I see you have used style.text.letter_space = 0 so it should be no space between letters.

You're right. Because set `style.text.letter_space = 1` will make the space too loose, I tried to set it to zero before.
Please let me close this issue.

kisvegabor:
Your welcome! :)

Issue #144, opened 2018-03-05
Button object addition idea


8bit-meiko:
Would be awesome to have a button with background image (or animation like a battery graph)

It's a much wanted feature for me, as I am planning to use these kind of icons for a GUI.

I think it's best to create a separate object for this, like lv_obj_btn_anim.

What do you think about this?

kisvegabor:
Hi,

**lv_icon** is a planned object which will acts as button but draws an image

Until lv_icon arrives you can create an image on a button:
```c
lv_obj_t * btn = lv_btn_create(lv_scr_act(), NULL);
lv_btn_set_style(btn, LV_BTN_STYLE_REL, &lv_style_transp_tight);
lv_btn_set_style(btn, LV_BTN_STYLE_PR, &lv_style_transp_tight);
lv_btn_set_action(btn, LV_BTN_ACTION_CLICK, my_action_func);

lv_obj_t * img = lv_img_create(btn, NULL);
lv_img_set_click(img, false);
lv_img_set_src(img, &some_image);
```

PetrenkoIgor:
> Hi,
> 
> **lv_icon** is a planned object which will acts as button but draws an image
> 
> Until lv_icon arrives you can create an image on a button:
> 
> ```c
> lv_obj_t * btn = lv_btn_create(lv_scr_act(), NULL);
> lv_btn_set_style(btn, LV_BTN_STYLE_REL, &lv_style_transp_tight);
> lv_btn_set_style(btn, LV_BTN_STYLE_PR, &lv_style_transp_tight);
> lv_btn_set_action(btn, LV_BTN_ACTION_CLICK, my_action_func);
> 
> lv_obj_t * img = lv_img_create(btn, NULL);
> lv_img_set_click(img, false);
> lv_img_set_src(img, &some_image);
> ```

where does this function come from "lv_img_set_click(img, false);"

embeddedt:
You should use [`lv_obj_set_click`](https://docs.littlevgl.com/en/html/object-types/obj.html#_CPPv416lv_obj_set_clickP8lv_obj_tb), which works on all object types.

PR #3322, opened 2022-05-04
Feature/reattach nxp pxp vglite accelerators v8.2.0 new widget acceleration

nicusorcitu:
Issue #3052


src/draw/nxp/lv_gpu_nxp.c
```diff
@@ -0,0 +1,391 @@
...
+    nxp_draw_ctx->base_draw.draw_rect = lv_draw_nxp_rect;
+    nxp_draw_ctx->base_draw.draw_img_decoded = lv_draw_nxp_img_decoded;
+    nxp_draw_ctx->blend = lv_draw_nxp_blend;
+    //nxp_draw_ctx->base_draw.wait_for_finish = lv_draw_nxp_wait_cb;
+}
+
+void lv_draw_nxp_ctx_deinit(lv_disp_drv_t * drv, lv_draw_ctx_t * draw_ctx)
+{
+    LV_UNUSED(drv);
+    LV_UNUSED(draw_ctx);
```
kisvegabor: `lv_draw_sw_deinit_ctx(drv, draw_ctx)` should be called here.

kisvegabor (commented):
Hi,

Amazing, thank you very much. I can test it tomorrow on hardware.

It seems the `draw_ctx` is used correctly and fallbacks are added to SW rendering.

However, in the callbacks there are checks like 
```c
    /*Fill/Blend only non masked, normal blended*/
    if(dsc->mask_buf == NULL && dsc->blend_mode == LV_BLEND_MODE_NORMAL) {
```
to see whether the GPU can draw a given thing. It's great but with the [new transformation update](https://docs.lvgl.io/master/overview/style.html#opacity-blend-modes-and-transformations) a strange thing can happen. During rendering LVGL might initializes new `draw_ctx`s and start drawing into a separate buffer (called layer). There is no problem so far, but if the content to be rendered has "holes", e.g. rounded corner LVGL temporarily sets the `disp_drv.screen_transp` flag. It means the renderers should draw into an ARGB buffer. With 32 bit color depth I think it's not a big problem but with 16 bit color depth the target pixel format is ARGB8565 which is probably not supported by the GPU. So in short I think the NXP callbacks should fallback to SW rendering if `LV_COLOR_DEPTH != 32 && disp->driver->screen_transp == 1`. The display being refreshed can be get with `_lv_refr_get_disp_refreshing()`.  

And a minor thing: in other files there are no API comments on global functions in both the C and H files but only in the H file to make maintenance simpler. If it's an NXP policy e can leave it as it is but if not I think it'd be better to update it.

nicusorcitu:
Sorry, it seems the makefiles are messed up, I will push a fix for them.

We normally build by MCUX or IAR and not using make.

This time I pushed another fix and I test it by myself:
          working-directory: tests/makefile
          run: make test_file

The build with makefiles shall be fine now.

kisvegabor:
I've just tested it on i.MXRT1170-EVK with the widgets demo and noted two things:
1. To bg_color of the selected tab is incorrect. It should be light blue but its dark gray.
2. There is no significant improvement in performance with PXP and VGLite enabled. I guess the limiting factor is the memory bandwidth to fill large rectangles and even the GPUs can't write the memory faster.  To mitigate this probably more `Lcdif` layers and bg color could be used to avoid constant memory filling of large areas. Some questions to this:
    - Would it be possible to store 1 or 2 FB in the internal RAM of 1176 MCU? 
    - Do you know how fast the external RAM is? Just to make some calculations to see the limitations.

nicusorcitu:
>     1. To bg_color of the selected tab is incorrect. It should be light blue but its dark gray.
There is a known issue while drawing widgets with VGLite, the issue is present since lvgl 8.0.2 (widget acceleration support was added since lvgl 8.0.2). The problem is that the background of VG-Lite is black instead of transparent, drawing full cover does not reveal the problem but once you use opacity the black color will make the image darker. --> Fixed by [4910c7f](https://github.com/lvgl/lvgl/pull/3322/commits/4910c7f7a5aa5e57558474d2882edb954cf86053)
Hence I am not aware of the wrong color issue, gray instead of blue... we will look at it. --> Fixed by [6c0b869](https://github.com/lvgl/lvgl/pull/3322/commits/6c0b869312be99b44ea59f2ec51900a883414f8b)
> 
>     2. There is no significant improvement in performance with PXP and VGLite enabled. I guess the limiting factor is the memory bandwidth to fill large rectangles and even the GPUs can't write the memory faster.  To mitigate this probably more `Lcdif` layers and bg color could be used to avoid constant memory filling of large areas. Some questions to this:
>        
>        * Would it be possible to store 1 or 2 FB in the internal RAM of 1176 MCU?
There is 1MB internal memory and 512MB external. 1MB shall be enough for 2 FB.
>        * Do you know how fast the external RAM is? Just to make some calculations to see the limitations.
Unfortunately I do not have such data, I will ask around.

src/draw/nxp/lv_gpu_nxp.c
nicusorcitu: Ok.

kisvegabor:
Thanks for the updates.

> There is 1MB internal memory and 512MB external. 1MB shall be enough for 2 FB.

1 FB needs 1280 x 720 x 2byte = 1800kB

>>  Do you know how fast the external RAM is? Just to make some calculations to see the limitations.

>Unfortunately I do not have such data, I will ask around.

The [datasheet](https://www.nxp.com/docs/en/nxp/data-sheets/IMXRT1170IEC.pdf) says:
image

So the best case is we have 200 M/sec data rate. With 32 bit interface and 16 bit colors its 400 M px/sec. On 30 FPS 400/30 = 13 Mpx/frame. And one frame is ~1Mpx. Let's say the LCD controller reads a frame (1MPx) with also 30 FPS, and LVGL fills the background (1 Mpx) and draw something (and other 1MPx). So 3 Mpx is used to fill the screen twice and show it on the display. 3/13 = ~25%. I.e. 25% CPU. Does this calculation seem correct?

xiaoxiang781216:
@kisvegabor this is case why many people are very interesting in RGB565+A8 format which can save both 25% memory and bandwidth without the noticeable visual and feature lose.

kisvegabor:
I see.

But really? Not considering that the LCD periphery is reading the frame buffer, we can write with 200 MHz. It's like we have a 200MHz MCU when we fill the whole screen. (Or 400MHz in case of 32 bit RAM, I'm not sure about the data width).

nicusorcitu:
> 1 FB needs 1280 x 720 x 2byte = 1800kB

True :)
I did the same calculation (but on 32 bits color dept) but I shrink by mistake a zero.

> 2. There is no significant improvement in performance with PXP and VGLite enabled.

In my unit tests the performance was really big, the FPS count was twice bigger than using the CPU only.
But I used only images of 128x128, rectangle 200x200 and areas bigger than the limit explained below.

The calls to VGLite/PXP API have an overhead and requires flushing the CPU caches, also the initialization takes some time. So for small pixel areas, using the VGLite/PXP might not be efficient. This is why we added the threshold areas LV_GPU_NXP_VG_LITE[PXP]_BLIT[FILL]_SIZE_LIMIT which triggers GPU/PXP. If the area is smaller, it will fallback to CPU.

The GPU/PXP really accelerates the pixel processing for large areas. Those thresholds are set to ~5000 pixels to help reach the best framerate.

As the LVGL benchmark application uses a mix of small areas (lines) and large areas (filled rectangles), the performance using GPU/PXP might not be as much as expected. But even so I see an improvement of at least 50% for supported features.

[CPU.txt](https://github.com/lvgl/lvgl/files/8660570/CPU.txt)
[PXP.txt](https://github.com/lvgl/lvgl/files/8660572/PXP.txt)
[VG-Lite.txt](https://github.com/lvgl/lvgl/files/8660573/VG-Lite.txt)

If you want to see how many fallback there are due to the threshold limit, you can enable
#define LV_GPU_NXP_VG_LITE_LOG_TRACES 1
or
#define LV_GPU_NXP_PXP_LOG_TRACES 1
It will tear down the system due to logging but you shall see something like this:

_[Error]	(6.215, +0)	 lv_gpu_nxp_vglite_fill: Area size 411 smaller than limit 5000. 	(in lv_draw_vglite_blend.c line #144)
[Error]	(6.215, +0)	 lv_draw_nxp_blend: VG-Lite fill failed. Fallback. 	(in lv_gpu_nxp.c line #168)
[Error]	(6.215, +0)	 lv_gpu_nxp_vglite_fill: Area size 138 smaller than limit 5000. 	(in lv_draw_vglite_blend.c line #144)
[Error]	(6.215, +0)	 lv_draw_nxp_blend: VG-Lite fill failed. Fallback. 	(in lv_gpu_nxp.c line #168)
[Error]	(6.215, +0)	 lv_gpu_nxp_vglite_fill: Area size 138 smaller than limit 5000. 	(in lv_draw_vglite_blend.c line #144)
[Error]	(6.215, +0)	 lv_draw_nxp_blend: VG-Lite fill failed. Fallback. 	(in lv_gpu_nxp.c line #168)
[Error]	(6.215, +0)	 lv_gpu_nxp_vglite_fill: Area size 888 smaller than limit 5000. 	(in lv_draw_vglite_blend.c line #144)
[Error]	(6.215, +0)	 lv_draw_nxp_blend: VG-Lite fill failed. Fallback. 	(in lv_gpu_nxp.c line #168)_

> 1. To bg_color of the selected tab is incorrect. It should be light blue but its dark gray.

Fixed by [6c0b869](https://github.com/lvgl/lvgl/pull/3322/commits/6c0b869312be99b44ea59f2ec51900a883414f8b)

kisvegabor:
>  But even so I see an improvement of at least 50% for supported features.

Have you measured the CPU only performance with -O3 optimization?

nicusorcitu:
> > But even so I see an improvement of at least 50% for supported features.
> 
> Have you measured the CPU only performance with -O3 optimization?

Indeed, with CPU optimization -O3, there is no improvement while using the GPU, but having it enabled will offload the CPU. As I mentioned the GPU become stronger while dealing with bigger areas.
I did not yet implemented the "wait_for_finish" callback. Hence the CPU will wait for the GPU completion before returning from draw context. E.G. In VG-Lite we currently wait for every widget:
```
err = vg_lite_finish();
VG_LITE_ERR_RETURN_INV(err, "Finish failed.");
```
I am working to implement this callback for both VG-Lite and PXP, not sure it will be ready for this patch set. But I believe it will be a big improvement for GPU performance.

kisvegabor:
Okay, let me know when you it's ready to be merged.

nicusorcitu:
Hi Gabor,
I have pushed some fixes as well as changes suggested by you. I did a rebase and run the unit tests again. Everything looks good.
Please merge the PR if everything is fine on your side.

BTW, the issue with the wrong opacity did escaped from my unit tests because I used all the time OPA_50 (255-opa == opa). We adapt the unit tests to use a OPA_20.

When do you intend to release 8.3.0?
Thanks.

kisvegabor:
Hi Nicu,

Awesome! CI passes too, so I merge it.

Thank you very much for this update! 

If something comes up we an open targeted issues.

kisvegabor:
Regarding v8.3. I think beginning of July seems like a reasonable date. @embeddedt, do you see any blocking issue?

nicusorcitu:
Thank you.

nicusorcitu:
Hi Gabor,

Can you please do a small patch with this include header that fixes a warning on 16 bits color depth? I can not do another PR as I have to pass through a long process.

Thanks, Nicu.

```diff
---
 src/draw/nxp/lv_gpu_nxp.c | 3 +++
 1 file changed, 3 insertions(+)

diff --git a/src/draw/nxp/lv_gpu_nxp.c b/src/draw/nxp/lv_gpu_nxp.c
index a02400230..2b0b02948 100644
--- a/src/draw/nxp/lv_gpu_nxp.c
+++ b/src/draw/nxp/lv_gpu_nxp.c
@@ -50,6 +50,9 @@
     #include "vglite/lv_draw_vglite_rect.h"
     #include "vglite/lv_draw_vglite_arc.h"
 #endif
+#if LV_COLOR_DEPTH != 32
+    #include "../../core/lv_refr.h"
+#endif
 
 /*********************
  *      DEFINES
-- 
```

kisvegabor:
Sure, added here: 6be43b83

Issue #3324, opened 2022-05-04
UsageFault Excepetion received when trying to allocate a buffer bigger than LV_MEM_SIZE


nicusorcitu:
### Perform all steps below and tick them with [x]
- [x] Read the [FAQ](https://docs.lvgl.io/master/intro/index.html#faq)
- [x] Check the related part of the [Documentation](https://docs.lvgl.io/)
- [x] Update lvgl to the latest version
- [ ] Reproduce the issue in a [Simulator](https://docs.lvgl.io/latest/en/html/get-started/pc-simulator.html)

### Describe the bug
Trying to allocate a buffer bigger than LV_MEM_SIZE with lv_mem_buf_get() will lead into exception handler.

### To Reproduce
Apply the attached patch with tlsf logs and try to allocate a bigger buffer. The traces below are self explanatory.
```c
[Trace]	(0.057, +0)	 lv_mem_buf_get: begin, getting 22 bytes 	(in lv_mem.c line #280)
[Trace]	(0.057, +0)	 lv_mem_buf_get: returning already allocated buffer (buffer id: 0, address: 0x8070aa04) 	(in lv_mem.c line #302)
[Trace]	(0.057, +0)	 lv_mem_buf_release: begin (address: 0x8070aa04) 	(in lv_mem.c line #334)
[Trace]	(0.057, +0)	 lv_mem_buf_get: begin, getting 60 bytes 	(in lv_mem.c line #280)
[Trace]	(0.057, +0)	 lv_mem_buf_get: returning already allocated buffer (buffer id: 0, address: 0x8070aa04) 	(in lv_mem.c line #302)
[Trace]	(0.057, +0)	 lv_mem_buf_release: begin (address: 0x8070aa04) 	(in lv_mem.c line #334)
[Trace]	(0.057, +0)	 lv_mem_buf_get: begin, getting 64 bytes 	(in lv_mem.c line #280)
[Trace]	(0.057, +0)	 lv_mem_buf_get: returning already allocated buffer (buffer id: 0, address: 0x8070aa04) 	(in lv_mem.c line #302)
[Trace]	(0.057, +0)	 lv_mem_buf_release: begin (address: 0x8070aa04) 	(in lv_mem.c line #334)
[Trace]	(0.059, +2)	 lv_mem_buf_get: begin, getting 65536 bytes 	(in lv_mem.c line #280)
[...]

PR #3325, opened 2022-05-04
Fix TLSF returning the wrong pointer when the requested size is too large

embeddedt:

* Limit test runs to 15 seconds so tests don't hang forever and instead fail.
* Make sure an  `LV_ASSERT_MSG` causes the test to fail instead of entering a `while(1);` loop.
* Add a test for #3324. Unfortunately it seems to pass at the moment.

### Checkpoints

### DIFF
--- a/src/misc/lv_tlsf.c
+++ b/src/misc/lv_tlsf.c
@@ -1208,6 +1208,10 @@ void * lv_tlsf_realloc(lv_tlsf_t tlsf, void * ptr, size_t size)
         const size_t cursize = block_size(block);
         const size_t combined = cursize + block_size(next) + block_header_overhead;
         const size_t adjust = adjust_request_size(size, ALIGN_SIZE);
+        if(size > cursize && adjust == 0) {
+            /* The request is probably too large, fail */
+            return NULL;
+        }
 
         tlsf_assert(!block_is_free(block) && "block already marked as free");
 
--- a/tests/src/test_cases/test_mem.c
+++ b/tests/src/test_cases/test_mem.c
@@ -1,0 +1,26 @@
+#if LV_BUILD_TEST
+#include "../lvgl.h"
+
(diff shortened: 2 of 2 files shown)


embeddedt:
I have realized that the assert statements are an issue in general, because I actually want the test to pass if an assertion happens, and fail otherwise (the opposite of the default logic). The cleanest solution would be if the testing framework had a feature built-in to invert the pass/fail logic, but Unity does not appear to have that. I'll need to think of an alternate solution if we want a CI test for #3324.

kisvegabor:
I believe the same should be tested with `lv_mem_alloc/free/realloc` where we can see the returned value and decide to fail or pass. 

```c
  void * buf = lv_mem_alloc(20);
  printf("%p\n", buf);
  buf = lv_mem_realloc(buf, LV_MEM_SIZE + 16384);
  printf("%p\n", buf);
```

Results in
```
0x56222ce78510
0x56222ce78510
``` 

Anyway we can fix this issue by updating `lv_tlsf.c`. That's we have a copy of these files and not submodules.

embeddedt:
I found the issue. `adjust_request_size` in TLSF returns 0 if the provided size is larger than the maximum block size, and that causes `lv_tlsf_realloc` to think the allocation doesn't need resizing just below [here](https://github.com/lvgl/lvgl/blob/58b75f1175504fc7a020cd3b89937c52d954a5f8/src/misc/lv_tlsf.c#L1210). 

What do you think of my fix?

EDIT: It seems that the issue was mentioned upstream but never addressed. https://github.com/mattconte/tlsf/issues/22

kisvegabor:
LGTM!

> It seems that the issue was mentioned upstream but never addressed. https://github.com/mattconte/tlsf/issues/22

I wanted to suggest reporting it in the upstream.

PR #3326, opened 2022-05-04
feat(demo): add Weighted FPS and Opa speed log output

FASTSHIFT:
### Checkpoints

### DIFF
--- a/demos/benchmark/lv_demo_benchmark.c
+++ b/demos/benchmark/lv_demo_benchmark.c
@@ -760,6 +760,8 @@ static void generate_report(void)
            "LVGL v%d.%d.%d " LVGL_VERSION_INFO
            " Benchmark (in csv format)\r\n",
            LVGL_VERSION_MAJOR, LVGL_VERSION_MINOR, LVGL_VERSION_PATCH);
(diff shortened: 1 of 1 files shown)


GorgonMeducer:
I still think the way this benchmark project calculates FPS is not ideal. 
An ideal one should be:
1) generate specific number of frames and calculate the cycles/time used.
2) calculate the FPS based on the cycles/time 

The current solution is:
1) run 2 sec 
2) see how many frames are rendered.

The key is that you should let the LVGL finish its work...

kisvegabor:
> An ideal one should be:

Also possible to make the CPU run at 100% by calling `lv_timer_handler();` and `lv_tick_inc(50)` in a loop.

> The current solution is:

With a reasonable FPS (> 10) in 2 sec it can introduce only few percentage of error. Which can be really significant in some cases. 

The advantage of this method is that you see in real time and also see the smoothness of the animations. With your suggestion the first benchmark would be ready in fraction of a second and the user wouldn't see anything.  It's not relevant though only if you use an empty display driver. 

It seems to me both modes are useful.

kisvegabor:
Anyway, this PR can be merged. Thanks :slightly_smiling_face:

PR #3331, opened 2022-05-05
feat(anim): add the function of getting global animation refresher timer

FASTSHIFT:
### Checkpoints

### DIFF
--- a/src/misc/lv_anim.c
+++ b/src/misc/lv_anim.c
@@ -73,16 +73,6 @@ void lv_anim_init(lv_anim_t * a)
     a->early_apply = 1;
 }
 
-void lv_anim_set_refr_period(uint32_t period)
-{
-    lv_timer_set_period(_lv_anim_tmr, period);
-}
-
-uint32_t lv_anim_get_refr_period(void)
-{
-    return _lv_anim_tmr->period;
-}
-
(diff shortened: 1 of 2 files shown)


kisvegabor:
Two questions:
1. What is the real-life use case behind this feature?
2. What about `lv_anim_set/get_timer_period`?

FASTSHIFT:
> Two questions:
> 
> 1. What is the real-life use case behind this feature?
> 2. What about `lv_anim_set/get_timer_period`?

If I want to reduce the power consumption of the device, I can use this API to reduce the overall animation frame rate and reduce the number of redraws.

kisvegabor:
Got it. Seems fair. 

API-wise we can consider  `lv_timer_t * lv_anim_get_timer(void)` too. It gives the highest flexibility. And it's similar to `_lv_disp_get_refr_timer(disp)`. What do you think?

FASTSHIFT:
> Got it. Seems fair.
> 
> API-wise we can consider `lv_timer_t * lv_anim_get_timer(void)` too. It gives the highest flexibility. And it's similar to `_lv_disp_get_refr_timer(disp)`. What do you think?

I agree with this.

kisvegabor:
Thank you!

PR #3335, opened 2022-05-05
fix(benchmark): remove redundant string for the small screens 

GorgonMeducer:

When using `lv_demo_benchmark_run_scene()` to run a specified scene, the test result was shown with a redundant string (the same as the title) in the subtle, it is unreadable for small screens. 

By only keeping the "Result: " part, it is readable. 

image

### Checkpoints


kisvegabor:
Makes sense. I added the the name of the previous test to make it clear that it's the result of the previous test. 

But I think it's not that confusing.

PR #3338, opened 2022-05-06
lv_txt_iso8859_1_next(..., NULL): fix returned value

amouiche:
In LV_TXT_ENC_ASCII case, some labels can be are truncated, depending on the label text content.
The root cause is the incorrect lv_txt_iso8859_1_next(..., NULL) behavior which returns the next character instead of the current one.

Width computed by lv_txt_get_width() may become invalid.


kisvegabor:
Thank you!

PR #3352, opened 2022-05-11
feat(benchmark): make lvgl render at the highest frame rate

FASTSHIFT:
The score calculated by the old benchmark program will be affected by the value set by `LV_DISP_DEF_REFR_PERIOD` and cannot reflect the real performance of the system.
The purpose of this PR is to force lvgl to render at the highest frame rate.

### Checkpoints

### DIFF
--- a/demos/benchmark/lv_demo_benchmark.h
+++ b/demos/benchmark/lv_demo_benchmark.h
@@ -34,6 +34,8 @@ void lv_demo_benchmark_run_scene(int_fast16_t scene_no);
 
 void lv_demo_benchmark_set_finished_cb(finished_cb_t * finished_cb);
 
+void lv_demo_benchmark_set_max_speed(bool en);
+
 /**********************
  *      MACROS
  **********************/
--- a/demos/benchmark/lv_demo_benchmark.h
+++ b/demos/benchmark/lv_demo_benchmark.h
@@ -34,6 +34,10 @@ void lv_demo_benchmark_run_scene(int_fast16_t scene_no);
 
 void lv_demo_benchmark_set_finished_cb(finished_cb_t * finished_cb);
 
+/**
+ * Make the benchmark work at the highest frame rate
+ * @param en true: highest frame rate; false: default frame rate
+ */
 void lv_demo_benchmark_set_max_speed(bool en);
 
 /**********************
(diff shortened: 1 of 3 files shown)


FASTSHIFT:
@kisvegabor 
Do I need to add an option to keep the option to use `LV_DISP_DEF_REFR_PERIOD` for the render cycle?

@GorgonMeducer
I plan to add an empty `disp_flush_cb` to bypass the time overhead of the display driver, What do you think?

FASTSHIFT:
New benchmarks at 1080P resolution
(KSAKSRML86G2O8_RC`4$ 5

GorgonMeducer:
@FASTSHIFT I use a different solution and will put a PR soon. 
It is simple:
a. In `disp_flush()` check a static flag before the actual flushing
b. Introduce a call-back function in the benchmark, i.e. `end_of_test_cb`, with which users can 
   * set the flushing flag to false
   * register a call back function 
   * run benchmark
   * in the call back function set the flushing flag to true

With this arrangement, you can read the test results normally and get rid of the LCD latency. 

This method has two benefits:
a. A callback for end-of-test indication is useful and can be used in more than one way. 
b. You don't have to modify existing code too much.

I will do this today, no worry. Just a piece of cake.

GorgonMeducer:
Done in #3353

kisvegabor:
As #3353 is merged and should solve this issue too I close this PR.

FASTSHIFT:
@kisvegabor 
This PR https://github.com/lvgl/lvgl/pull/3353 does not modify the rendering frame rate, which is different from what my PR modified.

kisvegabor:
Got it. In this case I suggest moving the timer adjust part into a separate function which can be optionally called after `lv_demo_becnhmark()`. E.g. `lv_demo_benchmark_max_speed()`. In v9 we can pass it as a parameter to `lv_demo_bechmark()` (along with a range of tests to run).

FASTSHIFT:
> Got it. In this case I suggest moving the timer adjust part into a separate function which can be optionally called after `lv_demo_becnhmark()`. E.g. `lv_demo_benchmark_max_speed()`. In v9 we can pass it as a parameter to `lv_demo_bechmark()` (along with a range of tests to run).

`lv_demo_benchmark_max_speed` has been added, do you agree with this modification?

demos/benchmark/lv_demo_benchmark.h
```diff
@@ -34,6 +34,8 @@ void lv_demo_benchmark_run_scene(int_fast16_t scene_no);
 
 void lv_demo_benchmark_set_finished_cb(finished_cb_t * finished_cb);
```
kisvegabor, line 36: Please add a comment to indicate the purpose of this function.

kisvegabor:
LGTM, thank you!

PR #3353, opened 2022-05-11
feat(demos): add a callback for benchmark

GorgonMeducer:

Add a callback to the benchmark project to indicate the testing is finished.
One application of this callback is using it to switching off flushing temporarily when testing and switch on the flushing when it is finished.

```c
#if LV_USE_DEMO_BENCHMARK
static void on_benchmark_finished(void)
{
    disp_enable();
}
#endif

int main(void)
{
    printf("Hello LVGL!!\r\n");
    
    __cycleof__("Draw strings on LCD") {
        __LL_LCD_PRINT_BANNER("Hello LVGL!!");
    }
    
    lv_init();
    lv_port_disp_init();
    lv_port_indev_init();
    
#if LV_USE_DEMO_BENCHMARK

    __LL_LCD_PRINT(24, 0, "Running LVGL Benchmark...");
    __LL_LCD_PRINT(25, 0, "Please stand by...");
    __LL_LCD_PRINT(28, 0, "NOTE: You will NOT see anything until the end.");

    disp_disable();
    
    lv_demo_benchmark_register_finished_handler(&on_benchmark_finished);
    lv_demo_benchmark();
    
    //lv_demo_benchmark_run_scene(43);      // run scene no 31
#endif
    ...
}
```

### Checkpoints

### DIFF
--- a/demos/benchmark/README.md
+++ b/demos/benchmark/README.md
@@ -15,6 +15,7 @@ On to top of the screen the title of the current test step, and the result of th
 - After `lv_init()` and initializing the drivers call `lv_demo_benchmark()`
 - If you only want to run a specific scene for any purpose (e.g. debug, performance optimization etc.), you can call `lv_demo_benchmark_run_scene()` instead of `lv_demo_benchmark()`and pass the scene number.
 - If you enabled trace output by setting macro `LV_USE_LOG` to `1` and trace level `LV_LOG_LEVEL` to `LV_LOG_LEVEL_USER` or higher, benchmark results are printed out in `csv` format.
+- If you want to know when the testing is finished, you can register a callback function via `lv_demo_benchmark_register_finished_handler()` before calling `lv_demo_benchmark()` or `lv_demo_benchmark_run_scene()`. 
 
 
 ## Interpret the result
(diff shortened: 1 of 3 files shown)


FASTSHIFT:
@GorgonMeducer 
Is it possible to temporarily take over the `flush_cb` callback function at the beginning of the benchmark, and set the `flush_cb` callback function registered by the user back to the `disp->driver` before executing the `report`, so that the user does not need to modify the driver.

GorgonMeducer:
> @GorgonMeducer
> Is it possible to temporarily take over the `flush_cb` callback function at the beginning of the benchmark, and set the `flush_cb` callback function registered by the user back to the `disp->driver` before executing the `report`, so that the user does not need to modify the driver.

Yes. It is possible and you need this finished callback to resume it. But I prefer to use flag to control flushing, It is easy and intuitive. 

Since the driver is also implemented by user, so it won't be a problem. If you implement this flushing control flag at the very beginning, then it is part of the infrastructure and becomes an driver feature.

demos/benchmark/lv_demo_benchmark.c
```diff
@@ -663,6 +665,39 @@ void lv_demo_benchmark(void)
...
+        rnd_reset();
+        scenes[scene_act].create_cb();
+
+        lv_timer_t * t = lv_timer_create(report_cb, SCENE_TIME, NULL);
+        lv_timer_set_repeat_count(t, 1);
+    }
+}
+
+
+void lv_demo_benchmark_register_finished_handler(finished_cb_t * finished_cb)
```
kisvegabor: I think this name would be more similar to other API names:
```suggestion
void lv_demo_benchmark_set_finished_cb(finished_cb_t * finished_cb);
```

GorgonMeducer: OK, I have no problem with your suggestion. Let's make this change.

demos/benchmark/README.md
```diff
@@ -36,25 +37,72 @@ By default, only the changed areas are refreshed. It means if only a few pixels
...
+    lv_port_disp_init();
+    lv_port_indev_init();
+
+    LV_LOG("Running LVGL Benchmark...");
+    LV_LOG("Please stand by...");
+    LV_LOG("NOTE: You will NOT see anything until the end.");
+
+    disp_disable();
+    
+    lv_demo_benchmark_register_finished_handler(&on_benchmark_finished);
```
kisvegabor:
```suggestion
    lv_demo_benchmark_set_finished_cb(&on_benchmark_finished);
```

kisvegabor:
Thanks!

PR #3363, opened 2022-05-18
update doc os: lv_time_handler to task_handler

rafaelcorsi:
`lv_task_handler` as the example code. 

### DIFF
--- a/docs/porting/os.md
+++ b/docs/porting/os.md
@@ -8,7 +8,7 @@ However, in the following conditions it's valid to call LVGL related functions:
 
 
 ## Tasks and threads
-If you need to use real tasks or threads, you need a mutex which should be invoked before the call of `lv_timer_handler` and released after it.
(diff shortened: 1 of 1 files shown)


embeddedt:
`lv_timer_handler` is the new name in v8+. `lv_task_handler` still exists but is just a legacy macro for those transitioning from v7.

rafaelcorsi:
> `lv_timer_handler` is the new name in v8+. `lv_task_handler` still exists but is just a legacy macro for those transitioning from v7.

this appears in a few other places in the doc, I can review everything and update this PR if you think it's necessary.

https://github.com/lvgl/lvgl/blob/a7cfc0b9015a501860ce6456fa1aa80a83750480/docs/intro/index.md#lvgl-doesnt-start-randomly-crashes-or-nothing-is-drawn-on-the-display-what-can-be-the-problem

embeddedt:
As far as I can tell, all the references on that page are using the correct (new) name already, unless I missed something?

embeddedt:
I'm going to close this as I think the current docs are correct.

PR #3372, opened 2022-05-20
fix(benchmark): fix the issue that wrong scene number is used

GorgonMeducer:

The wrong `scene_act` number was used when the result is reported after calling `lv_demo_benchmark_run_scene()` .

### Checkpoints


kisvegabor:
Thanks!

PR #3374, opened 2022-05-20
Add edge callbacks when trying to move focus past beginning or end

ftab:

When using rotary encoder to scroll focus through a group, there is no feedback given when trying to scroll further than the beginning or end. In some situations it can make the system feel unresponsive, compared to the bounce that it gives when scrolling via touchscreen (i.e. "I know you're trying to scroll, I'm just letting you know, you're already at the end").

This PR adds callbacks to the group so that something can be done to indicate there's no further to go. (Scroll bounce, sound, etc.)

See forum topic: https://forum.lvgl.io/t/rotary-encoder-arrow-key-scrolling-bounce/8918

Example usage: https://github.com/ftab/lv_port_pc_vscode/commit/0dbb083f2979c8ce7fe7a833e4bcaa296cae93a6

### Checkpoints

### DIFF
--- a/src/core/lv_group.c
+++ b/src/core/lv_group.c
@@ -24,7 +24,7 @@
 /**********************
  *  STATIC PROTOTYPES
  **********************/
-static void focus_next_core(lv_group_t * group, void * (*begin)(const lv_ll_t *),
+static bool focus_next_core(lv_group_t * group, void * (*begin)(const lv_ll_t *),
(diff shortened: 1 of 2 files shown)


ftab:
Should this be callbacks, or should it be events? I'm new with LVGL, so forgive me...

kisvegabor:
I see that it really can be useful in some situations. And the implementation looks good too, so I merge it. :slightly_smiling_face: 

Thank you!

PR #3380, opened 2022-05-25
feat(GN): Add lvgl.gni

100ask:

Let lvgl be built in GN (generate ninja).

GN:  https://gn.googlesource.com/gn

### Use example

Reference lvgl.gni in BUILD.gn: 

```shell

import("./lvgl/lvgl.gni")

static_library("app_spi_demo") {
   configs = []

   sources = lvgl_sources
   sources += [
      "main.c",
   ]


   include_dirs = board_include_dirs
   include_dirs += lvgl_include_dirs
}

```


kisvegabor:
Isn't there a glob feature to get all the source files?

embeddedt:
Is there an advantage of GN over other build systems like CMake that justifies adding yet another file to the root of the repository? LVGL's source/include file layout is intentionally kept simple so it doesn't need to ship its own build files.

PR #3384, opened 2022-05-26
feat(porting): add flushing control to the template

GorgonMeducer:

Introduce flushing control to `lv_port_disp_template.c`, this should simplify people's porting work. 
The flushing control is especially useful in some of the LVGL benchmark activities. In some application scenarios, a method to temporarily disable low-level flushing (or to create a dummy flushing) is useful. 

### Checkpoints

### DIFF
--- a/examples/porting/lv_port_disp_template.c
+++ b/examples/porting/lv_port_disp_template.c
@@ -144,7 +144,7 @@ static void disp_init(void)
 
 volatile bool disp_flush_enabled = true;
 
-/* Enable the flushing process when disp_flush() is called by LVGL
+/* Enable updating the screen (the flushing process) when disp_flush() is called by LVGL
  */
 void disp_enable_update(void)
 {
--- a/examples/porting/lv_port_disp_template.c
+++ b/examples/porting/lv_port_disp_template.c
@@ -11,10 +11,20 @@
  *********************/
 #include "lv_port_disp_template.h"
 #include "../../lvgl.h"
+#include <stdbool.h>
 
 /*********************
  *      DEFINES
  *********************/
+#ifndef MY_DISP_HOR_RES
+    #warning Please define or replace the macro MY_DISP_HOR_RES with the actual screen width, default value 320 is used for now.
+    #define MY_DISP_HOR_RES    320
+#endif
+
+#ifndef MY_DISP_VER_RES
+    #warning Please define or replace the macro MY_DISP_HOR_RES with the actual screen height, default value 240 is used for now.
+    #define MY_DISP_VER_RES    240
+#endif
 
 /**********************
  *      TYPEDEFS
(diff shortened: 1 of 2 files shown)


kisvegabor:
I think it adds an other layer of complexity to the driver for a specific use case. What about just adding a comment in `disp_flush` like "if you want to test only the rendering speed of LVGL just call `lv_disp_flush_ready(drv)`"

GorgonMeducer:
@kisvegabor Your proposal permanently disable the flushing. In fact, my proposal enables people to check the result on screen after the testing process.  It also helps people who have no trace output capability. 

you said it introduced complexity to driver, here is my defense : 

- It is only introduced to template, people can delete them if they don't like it
- If people leave it alone, by default,  the flushing is enabled, hence it doesn't introduce extra work to original porting work
- This logic of code is so obvious, I don't think people find it hard to understand the code in 30 seconds

Conclusion:  it introduce a new SIMPLE feature without introduce extra steps to the original porting steps, I see no harm but only benefits.

examples/porting/lv_port_disp_template.c
```diff
@@ -132,20 +142,38 @@ static void disp_init(void)
     /*You code here*/
 }
 
+volatile bool disp_flush_enabled = true;
+
+/* Enable the flushing process when disp_flush() is called by LVGL
+ */
+void disp_enable(void)
```
kisvegabor:
```suggestion
void disp_enable_update(void)
```

examples/porting/lv_port_disp_template.c
```diff
@@ -132,20 +142,38 @@ static void disp_init(void)
...
+{
+    disp_flush_enabled = true;
+}
+
+/* Disable the flushing process when disp_flush() is called by LVGL
+ */
+void disp_disable(void)
```
kisvegabor:
```suggestion
void disp_disable_update(void)
```

examples/porting/lv_port_disp_template.c
```diff
@@ -132,20 +142,38 @@ static void disp_init(void)
     /*You code here*/
 }
 
+volatile bool disp_flush_enabled = true;
+
+/* Enable the flushing process when disp_flush() is called by LVGL
```
kisvegabor:
```suggestion
/* Enable updating the screen (the flushing process) when disp_flush() is called by LVGL
```

examples/porting/lv_port_disp_template.c
```diff
@@ -132,20 +142,38 @@ static void disp_init(void)
...
+ */
+void disp_enable(void)
+{
+    disp_flush_enabled = true;
+}
+
+/* Disable the flushing process when disp_flush() is called by LVGL
```
kisvegabor:
```suggestion
/* Disable updating the screen (the flushing process) when disp_flush() is called by LVGL
```

kisvegabor (commented):
Fair enough. I just made some suggestions to approach the goal of these function from a slightly higher level.

GorgonMeducer:
Thank you very much!

kisvegabor:
The [header](https://github.com/lvgl/lvgl/pull/3384/files#diff-62ae52b28130f1038b202c05f5800e4951e98687928b6c6c8acd041343179e3eR34-R41) and the comment of [disp_disable_update](https://github.com/lvgl/lvgl/pull/3384/files#diff-62ae52b28130f1038b202c05f5800e4951e98687928b6c6c8acd041343179e3eR34-R41) still should be updated.

examples/porting/lv_port_disp_template.c
```diff
@@ -132,20 +142,38 @@ static void disp_init(void)
...
+/* Enable updating the screen (the flushing process) when disp_flush() is called by LVGL
+ */
+void disp_enable_update(void)
+{
+    disp_flush_enabled = true;
+}
+
+/* Disable the flushing process when disp_flush() is called by LVGL
+ */
+void disp_disable_update(void)
```
kisvegabor, line 156: This comment and the same in C file are intentionally not updated?

GorgonMeducer: Sorry, I don't understand. Can you give more details?
I have checked that the `lv_port_disp_template.c` and `lv_port_disp_template.h` have the same prototypes and the same comments.

kisvegabor: I mean the comment of the `disp_disable_update` is
```
Disable the flushing process when disp_flush() is called by LVGL
```

But for `disp_enable_update`:
```
Enable updating the screen (the flushing process) when disp_flush() is called by LVGL
```

So they have a different format. 

Sorry for the nitpicking.

GorgonMeducer: Aha, I got it now. I will update them.

kisvegabor:
Thank you!

PR #3385, opened 2022-05-27
fix(style): Reset lookup table after gc sweep

amirgon:
When a device goes through soft reset, all gc allocated memory is sweeped.
In such case, all LV_GC_ROOTs are set to NULL.

Consider the following scenario (related: https://github.com/lvgl/lv_binding_micropython/issues/213):
```
lv.init()
lv.deinit()
machine.soft_reset()
lv.init()
```

LVGL halts on the following assert:
```
[Error] (0.000, +0)      lv_style_register_prop: Asserted at expression: (mp_state_ctx.vm._lv_style_custom_prop_flag_lookup_table) != NULL (NULL pointer)       (in lv_style.c line #190)              
```

because [_lv_style_custom_prop_flag_lookup_table_size is set to 0](
https://github.com/lvgl/lvgl/blob/6be43b83b3dc9340263552167dbbb07c1069bdb0/src/misc/lv_style.c#L124) by static variable initialization, but is not reset again to 0 when `lv_deinit()` is called.  
But in this case the soft reset sweeps all gc allocated memory and `LV_GC_ROOT(_lv_style_custom_prop_flag_lookup_table)` is `NULL`, while `_lv_style_custom_prop_flag_lookup_table_size` is not 0.

To work around this, in case `LV_GC_ROOT(_lv_style_custom_prop_flag_lookup_table)` is `NULL`, we can assume a gc_sweep occured and force `_lv_style_custom_prop_flag_lookup_table_size` to 0.

In general, there may be a bigger issue here: How reliable is `lv_deinit()`?   
I think it should reset the entire state including global LVGL variables such as `_lv_style_custom_prop_flag_lookup_table_size`, and there are probably others.  
In any case, it's not safe to assume that any LV_GC_ROOT is valid after `lv_deinit` is called.

### DIFF
--- a/src/misc/lv_style.c
+++ b/src/misc/lv_style.c
@@ -169,6 +169,9 @@ void lv_style_reset(lv_style_t * style)
 
 lv_style_prop_t lv_style_register_prop(uint8_t flag)
 {
+    if(LV_GC_ROOT(_lv_style_custom_prop_flag_lookup_table) == NULL) {
+        _lv_style_custom_prop_flag_lookup_table_size = 0;
+    }
     /*
      * Allocate the lookup table if it's not yet available.
      */


kisvegabor:
Your fix looks good. `_lv_style_custom_prop_flag_lookup_table` is global so we could zero it in `lv_deinit` too, but it's cleaner to handle it here. 

 > In general, there may be a bigger issue here: How reliable is lv_deinit()?
I think it should reset the entire state including global LVGL variables such as _lv_style_custom_prop_flag_lookup_table_size, and there are probably others.

I can't imagine a better way than checking global variables one-by-one to be sure they are reset.

A related case: after `lv_deinit` the pointers stored in the `lv_style_t` will be invalid too. Is it ok to assume that, after deinit all LVGL related objects are invalid and we can leave wild pointers in the old widgets and styles?

amirgon:
> I can't imagine a better way than checking global variables one-by-one to be sure they are reset.

This sounds very tedious and hard to maintain.  
On the other hand, if we had a single struct that contains all of the "globals" all over LVGL, we could simply memset it to 0 on `lv_deinit`.

> A related case: after `lv_deinit` the pointers stored in the `lv_style_t` will be invalid too. Is it ok to assume that, after deinit all LVGL related objects are invalid and we can leave wild pointers in the old widgets and styles?

The cleanest approach is for lv_deinit to go over all objects, styles, screens, drivers etc. and free them all one by one.  
I don't think lv_deinit does any of that today, so that leaves this responsibility to the user.

kisvegabor:
> This sounds very tedious and hard to maintain.

True. 

> On the other hand, if we had a single struct that contains all of the "globals" all over LVGL, we could simply memset it to 0 on lv_deinit.

Interesting idea...

> The cleanest approach is for lv_deinit to go over all objects, styles, screens, drivers etc. and free them all one by one.
I don't think lv_deinit does any of that today, so that leaves this responsibility to the user.

Also true. The built in `malloc` can be reset so there is no memory leak, but with normal or a custom `malloc` it's not to case. We can `lv_obj_del` all objects and `lv_style_reset` all the style found in them, destroy all display, indevs, groups, animatons, etc, but we cannot do anything with
- styles that are not bound to any object now 
- `lv_obj_t *` pointers saved/used by the user

amirgon:
> but we cannot do anything with
> 
> * styles that are not bound to any object now
> * `lv_obj_t *` pointers saved/used by the user

I think it's reasonable to expect the user who calls `lv_deinit` not to use old styles and object pointers.  
I'd expect `lv_deinit` to be called in a context of cleanup/restart of the system, such as the soft-reset flow in Micropython.

kisvegabor:
> I'd expect lv_deinit to be called in a context of cleanup/restart of the system, such as the soft-reset flow in Micropython.

In this case we shouldn't care much about the memory leak either, as (I assume) it's also reset on the soft-reset.

But global variables of LVGL (stored in C, not necessarily affected by system soft reset) should be really zerod out.

amirgon:
> In this case we shouldn't care much about the memory leak either, as (I assume) it's also reset on the soft-reset.

For Micropython that's correct - soft-reset sweeps gc memories.  
But potentially in other cases, when calling `lv_deinit` from C, I can imagine cases where memory leaks could matter.



> But global variables of LVGL (stored in C, not necessarily affected by system soft reset) should be really zerod out.

Right.

kisvegabor:
I added this PR to the [architecture project](https://github.com/lvgl/lvgl/projects/6#card-82732632) as a reminder.

PR #3388, opened 2022-05-28
feat(textarea): Make it possible to customise the bullet character

Johennes:

This adds functions to allow overriding the default bullet character with a custom one (or even no character at all).

### Checkpoints

### DIFF
--- a/docs/widgets/core/textarea.md
+++ b/docs/widgets/core/textarea.md
@@ -63,8 +63,7 @@ In this mode the height is set automatically to show only one line, line break c
 ### Password mode
 The text area supports password mode which can be enabled with `lv_textarea_set_password_mode(textarea, true)`.
 
-If the `•` ([Bullet, U+2022](http://www.fileformat.info/info/unicode/char/2022/index.htm)) character exists in the font, the entered characters are converted to it after some time or when a new character is entered.
-If `•` not exists, `*` will be used.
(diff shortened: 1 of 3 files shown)


docs/widgets/core/textarea.md
```diff
@@ -63,8 +63,7 @@ In this mode the height is set automatically to show only one line, line break c
 ### Password mode
 The text area supports password mode which can be enabled with `lv_textarea_set_password_mode(textarea, true)`.
 
-If the `•` ([Bullet, U+2022](http://www.fileformat.info/info/unicode/char/2022/index.htm)) character exists in the font, the entered characters are converted to it after some time or when a new character is entered.
-If `•` not exists, `*` will be used.
+By default, if the `•` ([Bullet, U+2022](http://www.fileformat.info/info/unicode/char/2022/index.htm)) character exists in the font, the entered characters are converted to it after some time or when a new character is entered. If `•` does not exist in the font, `*` will be used. You can override the default character with `lv_textarea_set_password_bullet`.
```
kisvegabor:
```suggestion
By default, if the `•` ([Bullet, U+2022](http://www.fileformat.info/info/unicode/char/2022/index.htm)) character exists in the font, the entered characters are converted to it after some time or when a new character is entered. If `•` does not exist in the font, `*` will be used. You can override the default character with `lv_textarea_set_password_bullet(textarea, "x")`.
```

I'm suggesting this because it's more consistent with the other parts of the docs and indicates that a string (not a character) should be passed as 2nd argument.

kisvegabor (commented):
Your implementation looks great. And I agree that this feature can be useful. Only one nitpicking...

docs/widgets/core/textarea.md
Johennes: Ah, good call yes. 👍

kisvegabor:
Merged, thank you! :slightly_smiling_face:

Issue #3086, opened 2022-02-11
Add 'inherit' and 'initial' CSS properties


FoundationKen:
### Perform all steps below and tick them with [x]
- [x] Read the [FAQ](https://docs.lvgl.io/master/intro/index.html#faq)
- [x] Check the related part of the [Documentation](https://docs.lvgl.io/)
- [x] Update lvgl to the latest version (not tested under latest version, but the style inheritance code looks the same)
- [x] Reproduce the issue in a [Simulator](https://docs.lvgl.io/latest/en/html/get-started/pc-simulator.html)

### Describe the bug
If you create a grandparent object and add style with `text_color` set to red, then add a child object to it and add a label to that child, the label should inherit the red text color from the grandparent, but it does not.

It DOES work if the text color prop is in the immediate parent, but does not work if the text color prop is in the grandparent.
 
### To Reproduce
Simulator link with example:

https://sim.lvgl.io/v8.1/micropython/ports/javascript/index.html?script_startup=https://raw.githubusercontent.com/lvgl/lvgl/0b5a1d4b23975b920ff841ea9cd038802f51711b/examples/header.py&script=https://raw.githubusercontent.com/lvgl/lvgl/0b5a1d4b23975b920ff841ea9cd038802f51711b/examples/styles/lv_example_style_8.py&script_direct=8fb1c2b59e4164f867198e062ed23179ae69db58

### Expected behavior
Text color should be inherited from the whole parent chain.

### Screenshots or video
N/A

kisvegabor:
It's not working because `parent` also sets the text color to black. The inheritance algorithm finds the first parent on which the property is set. 

To achieve what you want you can remove all styles from the parent with `parent.remove_style_all()` (it already makes the text red) and add new styles to the parent which doesn't set the text color.

FoundationKen:
Thanks for the explanation.  Makes sense now.

Is there any less aggressive way to clear just the text color prop.  I guess this is a bit tricky since multiple styles could be applied and you don't necessarily want to remove the prop from a style completely (maybe it's shared and the change could affect other objects).

Is there a value you can set for a prop that could mean "inherit from parent" even if other styles are on this object?

That would let you add another style to the object to override any previous styles.  The recursive prop search would find this special value, but instead of returning it, it would recurse to the parent.

kisvegabor:
> Is there a value you can set for a prop that could mean "inherit from parent" even if other styles are on this object?

[...]

PR #3390, opened 2022-05-29
 Add 'inherit' and 'initial' CSS properties

embeddedt:

Implements the style values suggested in #3086.

Concerns:

* The API surface has tripled due to the need for separate APIs to set `inherit` and `initial`. It may be better to go with @FoundationKen's original suggestion and require users to use `lv_style_set_prop_meta` directly. Since there aren't three possible value types, there is no type safety issue there.
* `initial` currently uses the default value of a property, however, it might be more intuitive to derive that from the currently selected theme instead.
* No documentation yet.

### Checkpoints


src/core/lv_obj_style.c
```diff
@@ -224,14 +224,14 @@ lv_style_value_t lv_obj_get_style_prop(const lv_obj_t * obj, lv_part_t part, lv_
 {
     lv_style_value_t value_act;
     bool inherit = lv_style_prop_has_flag(prop, LV_STYLE_PROP_INHERIT);
```
kisvegabor: Do we need to store the  `INHERIT` flag for each property or should it be rather the default value of some properties?

src/misc/lv_style.c
```diff
@@ -155,7 +162,7 @@ void lv_style_reset(lv_style_t * style)
 {
     LV_ASSERT_STYLE(style);
 
-    if(style->is_const) {
+    if(style->prop1 == LV_STYLE_PROP_ANY) {
```
kisvegabor, line 168: What is the purpose of this change?

kisvegabor (commented):
> The API surface has tripled due to the need for separate APIs to set inherit and initial. It may be better to go with @FoundationKen's original suggestion and require users to use lv_style_set_prop_meta directly. Since there aren't three possible value types, there is no type safety issue there.

I agree to use a common function to set meta things. as these settings probably will be only rarely used. It makes the API a little bit inconsistent, but I think it's the best we can do in C. 

> initial currently uses the default value of a property, however, it might be more intuitive to derive that from the currently selected theme instead.

Great idea! I was already to thinking about how to handle the themes in a more clever way. My idea is that there could be 3 level of styles:
- local style: we already know it
- normal styles (like css classes): we also know it
- themes: unlike now, it can be a totally separated layer. If a property wasn't found in local or normal styles LVGL says: "okay theme, your turn, what should it be if no one knows." (i.e. it's the default).

This way a theme is not a huge collection of styles which are added to all objects by default, but just callback like:
```
lv_style_prop_t theme_get_prop(obj, prop, part, state);
``` 

Under the hood it can has a large switch case to find a style from where it gets the property, but it can also just return a color, width, radius, etc.

Moreover, it can return context based values too. E.g. for every 2nd button return gray, and white for the others. Or if the parent of a slider is a tabview, use different values. Basically what CSS selectors can do. It can be like the browser's style sheet. 
We can also let the users to specify a callback which will be called before the theme's callback to easily hook the default theme.

This lead a little bit far, but this way we could handle the styles a more optimal and flexible way. And getting the default value would be obvious too.

src/core/lv_obj_style.c
embeddedt: We could have a hardcoded default value but I suspect the added code will use up most of the space saved by not having a lookup table for built-in properties.

src/misc/lv_style.c
embeddedt: `is_const` had to be removed so that the address of `prop1` can be taken. Before, it was a bitfield. Since `LV_STYLE_PROP_ANY` isn't a valid value to store in `prop1`, I used that as a check for a constant style instead.

src/core/lv_obj_style.c
kisvegabor: I rather concerned about the inconsistency as this way we can set the inherit flag in 2 places:
- in the [prop look up table](https://github.com/lvgl/lvgl/blob/master/src/misc/lv_style.c#L29) 
- in the property value (with this feature)

Once we have this feature we don't need to use the look up table for this as it causes confusion about where inheritance is/should be set. 

Adding ~10 props [here](https://github.com/lvgl/lvgl/blob/master/src/misc/lv_style.c#L361) shouldn't require too much space but at least there were only one clear way to set inherit flag.

src/misc/lv_style.c
kisvegabor: In this case can we add dedicated name for const style? E.g. `_LV_STYLE_PROP_CONST`.

src/core/lv_obj_style.c
embeddedt: These are separate flags. The flag here (`LV_STYLE_PROP_INHERIT`, which will be renamed to `LV_STYLE_PROP_FLAG_INHERIT` in v9) specifies that the property supports inheritance from parents at all (because some like x and y don't). That flag is specified at the property type level.

The other flag (the new one) is effectively a special value for each style that forces `get_prop_core` to immediately check the parent's styles. That is specified in the prop-value pair within a style. It just so happens to be stored with the property ID because there are free bits there.

kisvegabor: >  The flag here (LV_STYLE_PROP_INHERIT, which will be renamed to LV_STYLE_PROP_FLAG_INHERIT in v9) specifies that the property supports inheritance from parents at all (because some like x and y don't). 

I see. Then what about a name like `LV_STYLE_PROP_FLAG_INHERITABLE`?

embeddedt: That name is definitely better, but we can't change it till v9 without breaking user code (unless we keep the old name as well).

Or do you think it's minor enough to change, since it can be fixed with a find-and-replace?

kisvegabor: We can keep it, I just wanted to be sure we are talking about the same thing. I added a note about the new name here: https://github.com/lvgl/lvgl/pull/3390#discussion_r885915769 

I suggest changing at least the variable name:
```suggestion
    bool inheritable = lv_style_prop_has_flag(prop, LV_STYLE_PROP_INHERIT);
```

src/misc/lv_style.c
embeddedt: I've added it, and under the hood it uses the same value for now.

src/core/lv_obj_style.c
embeddedt: Changed as you suggested.

kisvegabor:
Looks good! The only missing thing is to write a few sentences about it in the docs.

PR #3395, opened 2022-06-01
fix(slider): should send LV_EVENT_VALUE_CHANGED event on drag release

martaaay:

Per the documentation here:
https://docs.lvgl.io/8.2/widgets/core/slider.html?highlight=lv_slider_is_dragged
A LV_EVENT_VALUE_CHANGED event should be triggered on release:
"The event is sent continuously while the slider is dragged and once when released."
But it appears this event is not being triggered. I'm very new to this code base, so I'm not certain my fix is right, but it at least fixes the problem I was having. Seen in 8.2 release branch.

### Checkpoints
- [n/a] Update the documentation

### DIFF
--- a/src/widgets/lv_slider.c
+++ b/src/widgets/lv_slider.c
@@ -253,6 +253,9 @@ static void lv_slider_event(const lv_obj_class_t * class_p, lv_event_t * e)
 
         lv_obj_invalidate(obj);
 
+        res = lv_event_send(obj, LV_EVENT_VALUE_CHANGED, NULL);
+        if(res != LV_RES_OK) return;
+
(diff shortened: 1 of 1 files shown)


embeddedt:
If it doesn't introduce inconvenience, I'd rather fix the documentation to match the code. Currently the event fires every time the value really does change, which is more efficient. Firing an extra time on release seems redundant when I think about it. If the goal is to only process the value change once, you can just listen to `LV_EVENT_RELEASED` like the slider itself does.

martaaay:
Oh ok! That makes sense! I'll give that a try and close this pull request, and just reopen if i find a new problem. Thank you!

martaaay:
I'll write up a pull request for the documentation

PR #3397, opened 2022-06-02
fix(slider): fix documentation to match the code where a LV_EVENT_VAL…

martaaay:
…UE_CHANGED event is not throw on slider release, but a LV_EVENT_RELEASED is.

Fixing the documentation. The documentation [here](https://docs.lvgl.io/8.2/widgets/core/slider.html?highlight=lv_slider_is_dragged) specifies:
"The event is sent continuously while the slider is dragged and once when released. Use lv_slider_is_dragged to determine whether the Slider is still being dragged or has just been released."
But the code does not fire an LV_EVENT_VALUE_CHANGED event on release. Instead, a developer should just register for an LV_EVENT_RELEASED event.

### Checkpoints


embeddedt:
Thanks!

PR #3403, opened 2022-06-07
SDL Transformation support

### LINKED ISSUE #3386: SDL Renderer Issue When Zoom Set
- [ ] Read the [FAQ](https://docs.lvgl.io/master/intro/index.html#faq)
- [ ] Check the related part of the [Documentation](https://docs.lvgl.io/)
- [ ] Update lvgl to the latest version
- [ ] Reproduce the issue in a [Simulator](https://docs.lvgl.io/latest/en/html/get-started/pc-simulator.html)

### Describe the bug

After commit https://github.com/lvgl/lvgl/commit/318146a2c25362eabf258470be263a4cfeaefe87, 
SDL renderer will have issue drawing objects has zoom transform. Object is not properly zoomed, and when drawing it invades a part of objects in the front.

image

Sorry that I didn't post more details. I'll be investigating this issue and make patches.

mariotaku:

This PR adds ability for SDL backend to render a widget with transformation (zoom, rotate). This also fixes #3386 

### Checkpoints


kisvegabor:
Thank you for the PR.

I agree to move as much things as possible into a `draw_ctx` callback to provide the highest flexibility. However calling e.g. `lv_obj_redraw` from a draw function little bit messes up the architecture. Ideally drawing shouldn't know about objects. 

Or we can simply say the `draw_ctx` is a "cross-layer" thing and not limited to drawing only, but the whole rendering process.

Thinking about it more I can't imagine a better interface either. E.g. as the SW render part works (refresh in smaller chunks for simple layer) is specific to the nature of the SW render and its goals (i.e. use minimal resources).

So if you say it's ready I can merge it. After merging I'll move [this section](https://github.com/lvgl/lvgl/pull/3403/files#diff-67fd1b48a2bb4b3f48b602ce63408bcc5c729d6193bf05735905a7e74c0fa50cR827) into a `lv_draw_sw_` function, similarly as you did for SDL.

mariotaku:
That's a good point to separate objects and drawing backends completely. To simplify the drawing process, I'll write some pseudo-code.

```
if has_transform:
  t_ctx = prepare_transform_ctx()
  draw_obj(t_ctx)
  draw_transformed_img(t_ctx.snapshot)
  finish_transform_ctx(t_ctx)
else:
  draw_obj(ctx)
endif
```

This should be the case right? On SDL, there is no need to create a new rendering context, but it's better to introduce a state flag, to let backends know when they are rendering in transformation.

Also, **nested** transformation should be also considered. We could either crash on nested transformation, or support by creating multiple render targets for multiple level of transformations.

I think I can figure to do above refactoring, to prevent API changes introduced into master branch. Let's hold this PR while I update it.

kisvegabor:
> This should be the case right? On SDL, there is no need to create a new rendering context, but it's better to introduce a state flag, to let backends know when they are rendering in transformation.


It looks good. I just wonder how to handle the current implementation of SW which can subdivide the ares to redraw in case of LAYER_SIMPLEs to save memory. It's not required in SDL, but useful for SW rendering to avoid allocating large temporary buffers. It'd be the best if such behavior would be related to the backend itself. But the pseudo code you proposed supports only "one-pass" rendering without subdivision.

> Also, nested transformation should be also considered. We could either crash on nested transformation, or support by creating multiple render targets for multiple level of transformations.

SW rendering supports nested transforms.

mariotaku:
I would consider extracting the drawing process, let's say `draw_transformed_img`. So for memory sensitive versions (SW and pixel buffer based renderers), default implementation will be used while SDL will always draw the whole buffer.

mariotaku:
For nested drawing I also come up with some idea. I'll let you know when I'm finished :)

kisvegabor:
> So for memory sensitive versions (SW and pixel buffer based renderers), default implementation will be used while SDL will always draw the whole buffer.

Yes, probably it's the best we can do.

mariotaku:
Made some changes although I'm not 100% satisfied with the naming... Is this design look reasonable to you?

mariotaku:
Let's see if it's possible to implement SW backend with same way.

kisvegabor:
Looks good! Regarding the naming: it's more than just transforming. A new "layer" is created if the object has:
- opa < 100%
- blend mode != normal
- any transformation

So I'd name it somehow with the word "layer". E.g.`draw_ctx->layer_create` and `draw_ctx->layer_destroy`. 

With this API we can still keep the current optimization which subdivides the area. With SDL it can be disabled of course. 

I'll update the SW render to use this API on the weekend (hopefully).

kisvegabor:
I abstracted the SW layer stuff. I needed to slightly modify the [API functions](https://github.com/lvgl/lvgl/pull/3403/files#diff-4d593ddf8b3298897f890f90e370b51fd080fa53110cbd90c947674f4c4582efR46-R73).  `lv_draw_layer_adjust` is not required for you, it's used only when the layer is drawn in multiple parts.

kisvegabor:
@amirgon MP's CI is failing now due [this change](https://github.com/lvgl/lvgl/pull/3403/files#diff-0be6707528460cdc1dd90dad35b2597054e790c6b811c3ed5f9d3ce41c105ed0R137). (This feature wasn't released yet). Shall we do something with it here?

mariotaku:
New interfaces looks good to me! Though there are few fields needed by SDL was not introduced. I'll make changes and push to this branch directly.

mariotaku:
Also I think the variables referenced should be argument in the macro if needed, to prevent building issue.

src/core/lv_refr.c
```diff
@@ -899,8 +912,9 @@ void refr_obj(lv_draw_ctx_t * draw_ctx, lv_obj_t * obj)
         lv_opa_t opa = lv_obj_get_style_opa(obj, 0);
         if(opa < LV_OPA_MIN) return;
 
-        lv_area_t layer_area_full;
-        lv_res_t res = layer_get_area(draw_ctx, obj, layer_type, &layer_area_full);
+        lv_area_t layer_area_ext, layer_area_ext_trans, layer_area_full;
```
kisvegabor: Are you sure `layer_area_ext, layer_area_ext_trans` are really required? My view ways that only the layer area is needed as it will be rendered. The rest should comes naturally from the transformation. 

In other words the "layer module" should not know about the such a thing like an `obj_ext_size`. It just creates a buffer where LVGL can render and then blends/transforms the that buffer.

kisvegabor:
> Also I think the variables referenced should be argument in the macro if needed, to prevent building issue.

What do you mean by that?

src/core/lv_refr.c
mariotaku: There are sometimes 1px misalign, I'll see if I can solve that without draw the whole object.
image

mariotaku:
> > Also I think the variables referenced should be argument in the macro if needed, to prevent building issue.
> 
> What do you mean by that?

I mean This line causes the CI to fail.

```c
#define LV_LAYER_SIMPLE_FALLBACK_BUF_SIZE LV_MAX(lv_area_get_width(&draw_area) * px_size, 2048)
```

mariotaku:
Done! This time it should draw properly, without introducing unnecessary fields.

mariotaku:
This PR is good to go on my side. Please let me know if anything needed.

amirgon:
> @amirgon MP's CI is failing now due [this change](https://github.com/lvgl/lvgl/pull/3403/files#diff-0be6707528460cdc1dd90dad35b2597054e790c6b811c3ed5f9d3ce41c105ed0R137). (This feature wasn't released yet). Shall we do something with it here?

I've update lv_micropython and lv_binding_micropython with this fix to lv_conf.h (https://github.com/lvgl/lv_binding_micropython/commit/09e3cd45828d678c85aa62f997d22da21162d1db)

Please try to re-run the Micropython CI.

kisvegabor (approved):
Looks good to me too! 
Thank you!

PR #3408, opened 2022-06-10
feat(ime_pinyin): add API to use Pinyin IME(Chinese input)

YobieZhou:
1.Low resource usage
2.Support custom thesaurus
3.Support custom styles (Candidate panel, keyboard)
4.Support simple spelling

### Checkpoints


.vscode/settings.json
```diff
@@ -0,0 +1,6 @@
+{
```
kisvegabor: This file shouldn't be added. You can add `.vscode/` to `.gitignore`.

kisvegabor (commented):
Thanks you very much! A few notes:
- What do you think about having a `lv_pinyin_ime_set_keyboard` function for higher modularity? (Instead of creating the keyboard in `lv_pinyin_ime_create`)
- Could you add at least one example?
- Could you write a few words about it in the docs?
- What do you think about using the `lv_img_pinyin_*` prefix instead of `lv_pinyin_ime_*`?  As `ime` is the larger group I'd start with it.

YobieZhou:
> Thanks you very much! A few notes:
> 
> * What do you think about having a `lv_pinyin_ime_set_keyboard` function for higher modularity? (Instead of creating the keyboard in `lv_pinyin_ime_create`)
> * Could you add at least one example?
> * Could you write a few words about it in the docs?
> * What do you think about using the `lv_img_pinyin_*` prefix instead of `lv_pinyin_ime_*`?  As `ime` is the larger group I'd start with it.

1.This is exactly what I want to do next (I've started improving it recently)
2-3.Corresponding documentation and examples will of course be written. I will follow the format document [here](https://docs.lvgl.io/master/others/index.html), and the usage example is also written in the document, what do you think?
4.This looks even better! I will do it.

.vscode/settings.json
YobieZhou: I'll do it

kisvegabor: Please don't forget to remove this file.

examples/others/ime/lv_example_ime_pinyin_1.c
```diff
@@ -0,0 +1,51 @@
...
+    lv_obj_set_style_text_font(ta1, &lv_font_source_han_sans_normal_16, 0);   // set font
+    lv_obj_align(ta1, LV_ALIGN_TOP_LEFT, 10, 10);
+
+    lv_keyboard_set_textarea(kb, ta1);
+
+    lv_obj_add_event_cb(ta1, ta_event_cb, LV_EVENT_ALL, kb);
+
+    /* ta2 */
+    lv_obj_t * ta2 = lv_textarea_create(lv_scr_act());
+    lv_obj_set_style_text_font(ta2, &lv_font_source_han_sans_normal_16, 0);   // set font
```
kisvegabor: Can we use the [built in simsun](https://github.com/lvgl/lvgl/blob/master/lv_conf_template.h#L379) font instead?

kisvegabor (commented):
Sorry for the late reply.

It looks very well, I only had 2 minor notes.

examples/others/ime/lv_example_ime_pinyin_1.c
YobieZhou: > 

This is a better practice! I'll test it before that, and I may need to modify the dictionary according to simsun.

kisvegabor: Great! I tried my best to collect the most common CJK characters in the simsun font, but I might have missed something. Please let me know if we should add some more characters to it.

kisvegabor:
Looks good thank you! To fix the 2 issues of the CI, please do these:
```sh
cd scripts
./code-format.py 
./lv_conf_internal_gen.py 
```

examples/others/ime/lv_example_ime_pinyin_1.c
YobieZhou: I revised the dictionary in ime_pinyin according to simsun. I think simsun is enough for the ime_pinyin example. If simsun is updated later,  the dictionary in ime_pinyin can be updated.

It is not recommended that we provide users with a general Chinese character library and dictionary, because the actual use scenarios are variable.

They can customize according to their actual needs. In a similar way, I provided the method of using Chinese character library and common Chinese characters on the [lvgl Chinese website](http://lvgl.100ask.net/8.2/tools/fonts-zh-source.html) I created.

YobeZhou:
> 

I'm very sorry. I modified it directly on GitHub, ignoring that CI will be triggered many times

YobieZhou:
> Looks good thank you! To fix the 2 issues of the CI, please do these:
> 
> ```shell
> cd scripts
> ./code-format.py 
> ./lv_conf_internal_gen.py 
> ```

I'm very sorry. Recently, I have to modify it directly on GitHub. I checked the commit just now and it should be OK.

kisvegabor:
All tests passed, merging.

Thank you very much!

PR #3416, opened 2022-06-17
fix(benchmark): fix run_max_speed score incorrect

FASTSHIFT:
@GorgonMeducer 

### Checkpoints

### DIFF
--- a/demos/benchmark/lv_demo_benchmark.c
+++ b/demos/benchmark/lv_demo_benchmark.c
@@ -643,12 +643,12 @@ static void benchmark_init(void)
     if(run_max_speed) {
         if(disp->refr_timer) {
             disp_ori_timer_period = disp->refr_timer->period;
-            lv_timer_set_period(disp->refr_timer, 1);
+            lv_timer_set_period(disp->refr_timer, 2);
         }
 
         lv_timer_t * anim_timer = lv_anim_get_timer();
         anim_ori_timer_period = anim_timer->period;
-        lv_timer_set_period(anim_timer, 1);
+        lv_timer_set_period(anim_timer, 2);
     }
 
     lv_obj_t * scr = lv_scr_act();


GorgonMeducer:
@kisvegabor We found a strange thing as shown in the code that instead of 1ms using 2ms can ensure the LVGL runs at the max speed. Do you have any idea why?

kisvegabor:
The issue can be in `lv_timer_exec` but it seems correct to me. Can you check with a debugger what's happening?

GorgonMeducer:
@FASTSHIFT It's your turn now!

FASTSHIFT:
@kisvegabor 
I think benchmark scores as authoritative data should always work in the fastest mode. And don't use the statistical method of accumulating the rendering time of each frame, because this method may cause a large cumulative error under high-performance processors (such as the rendering time of each frame < 1ms). We should let lvgl draw enough frames before recording the total time.

kisvegabor:
In the practice I also find important to see how it really looks. E.g. is there any tearing, curtain effect, etc. 

However, I agree to make the selection between the two modes more direct. E.g. `lv_demo_bechmark(LV_DEMO_BENCHMARK_MODE_FAST/VISUAL)`. 

What do you think?

GorgonMeducer:
@kisvegabor When we want to see how it looks, we don't need a dedicated VISUAL mode but enable the low-level flushing, i.e. call `disp_enable_update()` before `lv_demo_benchmark()`.

kisvegabor:
I think we are talking about two separate features which can be combined:
1. How the measure: 
   - time of 100 rendering (proposed)
   - average rendering time in 2 sec (current)
2. Use of display  
   - enable flush
   - disable flush

Ultimately it seems two modes:
- FAST:  time of 100 rendering with disabled flush
- VISUAL: average rendering time in 2 sec with enabled flush

Other combinations doesn't make much sens to me. 

Disabling the flush can handled by  the demo too. It just needs to set a dummy `flush_cb` and restore the original in the end.

GorgonMeducer:
@kisvegabor OK, the two combinations sound reasonable to me also. And please don't forget, this PR exposes a problem, which is why `2` instead of `1` get the best performance?

kisvegabor:
> why 2 instead of 1 get the best performance

I need to make some test on it. I'll try my best to try it out on Monday or Tuesday.

GorgonMeducer:
Thank you very much! @kisvegabor

kisvegabor:
I've tried out with `max_speed` and 1, 2 and 5 ms periods but got the same result.

Actually it makes sense to get the same result assuming the rendering time of any frame is larger than the timer's period. This way the timer surely expires during rendering. 

In my case the best FPS was 172 -> 5.8ms so until 5 ms there is no difference.

GorgonMeducer:
@kisvegabor Have you disabled disp_flush? There should be significant differences between 1 and 2.
And what kind of platform do you use? We observe differences on MCU platforms.

kisvegabor:
> Have you disabled disp_flush? 

Yes, I did.

> And what kind of platform do you use?

STM32F769

What platform did you use? (if it's public)

GorgonMeducer:
@kisvegabor I am using the MPS3 FPGA prototyping system, which provides an idea condition for testing:
- Zero wait state RAM (ITCM and DTCM)
- No cache

While STM32F769 runs in FLASH (There is a cache for FLASH), and the SRAM is not zero wait state.
I prefer to believe the ideal condition. And the problem is not only observed by me, but also others such as @FASTSHIFT

GorgonMeducer:
@kisvegabor 
I think I found the difference. In our previous testing, macro `LV_DISP_DEF_REFR_PERIOD` was configurable, and we found that 2 was better than 1.

As discussed in the issue https://github.com/lvgl/lvgl/issues/3530, the different values of the new macro `LV_DEF_REFR_PERIOD` should also replicate the same issue.

Here are the test results:

| LV_DEF_REFR_PERIOD | Weighted FPS |
| ------------------ | ------------ |
| 33                 | 223          |
| 2                  | 289          |
| 1                  | 223          |

kisvegabor:
The benchmark set the refr. periods to 1 in fast mode. Do you see the performance difference with the latest master too where `LV_DEF_REFR_PERIOD` is used more consciously only for setting the default timer periods?

GorgonMeducer:
Yes, I am using the latest master. The conclusion is the same, 2 is significantly better than 1.

kisvegabor:
Are there tests where you got >500 or >1000 FPS?

GorgonMeducer:
> Are there tests where you got >500 or >1000 FPS?

Yes, Some  scenes got  > 1000 FPS. 

Similar to this https://github.com/lvgl/lv_port_an547_cm55_sim/blob/master/lvgl_benchmark.md

kisvegabor:
To debug it I can set a lower resolution (e.g. 200x200) in the STM MCU to get > 1000 FPS. I can try it out on Tuesday.

GorgonMeducer:
@kisvegabor I am using 320*240 RGB565

kisvegabor:
I've created a [PR](https://github.com/lvgl/lvgl/pull/3625) with 3 benchmark modes. I think it's much cleaner what is being measured than before. 

Please take a look at it, and let's see if the 1 ms / 2 ms timer period issue still persists. If so, I got an idea: please try calling `lv_anim_refr_now()` at the beginning of `_lv_disp_refr_timer` and disable the animation timer.

GorgonMeducer:
OK, I will have a try and let you know the result.

GorgonMeducer:
@kisvegabor will you merge this PR first, so I can try them?

kisvegabor:
Okay, merged both. (#3625 and this)

GorgonMeducer:
@kisvegabor Please find the test result below.

In `LV_DEMO_BENCHMARK_MODE_RENDER_ONLY` mode:

| LV_DEF_REFR_PERIOD | Weighted FPS |
| ------------------ | ------------ |
| 33                 | 194          |
| 2                  | 212          |
| 1                  | 201          |

>  If so, I got an idea: please try calling lv_anim_refr_now() at the beginning of _lv_disp_refr_timer and disable the animation timer.

Sorry, I am not sure where I should apply this change.

kisvegabor:
I found a bug and fixed it here: 2188fe66b

Does it make any difference?

GorgonMeducer:
@kisvegabor will have a try. But his so-called bug limits the FPS to no more than 50FPS. I have doubts about it.

kisvegabor:
`LV_DEMO_BENCHMARK_MODE_RENDER_ONLY` mode [measures the render time in monitor_cb](https://github.com/lvgl/lvgl/blob/master/demos/benchmark/lv_demo_benchmark.c#L766) so it doesn't matter how long we wait between renders. 

The reason we should wait a little between renders is the have time for the animations to animate. 

Anyway, I have no idea how  `LV_DEF_REFR_PERIOD` affects this as now the whole benchmark spins in a [for loop](https://github.com/lvgl/lvgl/blob/master/demos/benchmark/lv_demo_benchmark.c#L686-L690) independently from `lv_timer_handler()`.

GorgonMeducer:
> Anyway, I have no idea how LV_DEF_REFR_PERIOD affects this as now the whole benchmark spins in a [for loop](https://github.com/lvgl/lvgl/blob/master/demos/benchmark/lv_demo_benchmark.c#L686-L690) independently from lv_timer_handler().

Yeah... this also confuses me... Anyway, I will try and let you know the result.

GorgonMeducer:
@kisvegabor Found an issue about your fixing: https://github.com/lvgl/lvgl/commit/2188fe66bb07d73a0f2f85df7e8d22f3797734d1

The test results become very low, i.e.  < 32FPS

After checking your fix and the test report, I think it is a problem:
```
for(i = 0; i < RENDER_REPEAT_CNT; i++) {
    /*Wait a little to be sure something happens with the animations*/
    uint32_t t = lv_tick_get();
    while(lv_tick_elaps(t) < 20);
    lv_refr_now(NULL);
}
```

Here:
```
    while(lv_tick_elaps(t) < 20);
```
means no test scene can exceed 50 FPS and the test report clearly reflect this:


> Result of "Rectangle": 40 FPS
Result of "Rectangle + opa": 38 FPS
Result of "Rectangle rounded": 39 FPS
Result of "Rectangle rounded + opa": 36 FPS
Result of "Circle": 32 FPS
Result of "Circle + opa": 31 FPS
Result of "Border": 41 FPS
Result of "Border + opa": 41 FPS
Result of "Border rounded": 38 FPS
Result of "Border rounded + opa": 37 FPS
Result of "Circle border": 30 FPS
Result of "Circle border + opa": 30 FPS
Result of "Border top": 39 FPS
Result of "Border top + opa": 39 FPS
Result of "Border left": 39 FPS
Result of "Border left + opa": 39 FPS
Result of "Border top + left": 39 FPS
Result of "Border top + left + opa": 38 FPS
Result of "Border left + right": 38 FPS
Result of "Border left + right + opa": 37 FPS
Result of "Border top + bottom": 38 FPS
Result of "Border top + bottom + opa": 38 FPS
Result of "Shadow small": 26 FPS
Result of "Shadow small + opa": 26 FPS
Result of "Shadow small offset": 27 FPS
Result of "Shadow small offset + opa": 26 FPS
Result of "Shadow large": 19 FPS
Result of "Shadow large + opa": 18 FPS
Result of "Shadow large offset": 19 FPS
Result of "Shadow large offset + opa": 19 FPS
Result of "Image RGB": 48 FPS
Result of "Image RGB + opa": 48 FPS
Result of "Image ARGB": 47 FPS
Result of "Image ARGB + opa": 46 FPS
Result of "Image chorma keyed": 49 FPS
Result of "Image chorma keyed + opa": 48 FPS
Result of "Image indexed": 40 FPS
Result of "Image indexed + opa": 40 FPS
Result of "Image alpha only": 40 FPS
Result of "Image alpha only + opa": 39 FPS
Result of "Image RGB recolor": 47 FPS
Result of "Image RGB recolor + opa": 47 FPS
Result of "Image ARGB recolor": 46 FPS
Result of "Image ARGB recolor + opa": 45 FPS
Result of "Image chorma keyed recolor": 47 FPS
Result of "Image chorma keyed recolor + opa": 46 FPS
Result of "Image indexed recolor": 39 FPS
Result of "Image indexed recolor + opa": 39 FPS
Result of "Image RGB rotate": 39 FPS
Result of "Image RGB rotate + opa": 39 FPS
Result of "Image RGB rotate anti aliased": 36 FPS
Result of "Image RGB rotate anti aliased + opa": 36 FPS
Result of "Image ARGB rotate": 39 FPS
Result of "Image ARGB rotate + opa": 38 FPS
Result of "Image ARGB rotate anti aliased": 29 FPS
Result of "Image ARGB rotate anti aliased + opa": 28 FPS
Result of "Image RGB zoom": 42 FPS
Result of "Image RGB zoom + opa": 43 FPS
Result of "Image RGB zoom anti aliased": 41 FPS
Result of "Image RGB zoom anti aliased + opa": 41 FPS
Result of "Image ARGB zoom": 43 FPS
Result of "Image ARGB zoom + opa": 43 FPS
Result of "Image ARGB zoom anti aliased": 35 FPS
Result of "Image ARGB zoom anti aliased + opa": 35 FPS
Result of "Text small": 19 FPS
Result of "Text small + opa": 19 FPS
Result of "Text medium": 19 FPS
Result of "Text medium + opa": 19 FPS
Result of "Text large": 19 FPS
Result of "Text large + opa": 19 FPS
Result of "Text small compressed": 26 FPS
Result of "Text small compressed + opa": 26 FPS
Result of "Text medium compressed": 26 FPS
Result of "Text medium compressed + opa": 26 FPS
Result of "Text large compressed": 28 FPS
Result of "Text large compressed + opa": 27 FPS
Result of "Line": 30 FPS
Result of "Line + opa": 30 FPS
Result of "Arc think": 33 FPS
Result of "Arc think + opa": 33 FPS
Result of "Arc thick": 33 FPS
Result of "Arc thick + opa": 33 FPS
Result of "Substr. rectangle": 24 FPS
Result of "Substr. rectangle + opa": 22 FPS
Result of "Substr. border": 27 FPS
Result of "Substr. border + opa": 27 FPS
Result of "Substr. shadow": 12 FPS
Result of "Substr. shadow + opa": 11 FPS
Result of "Substr. image": 37 FPS
Result of "Substr. image + opa": 39 FPS
Result of "Substr. line": 23 FPS
Result of "Substr. line + opa": 22 FPS
Result of "Substr. arc": 31 FPS
Result of "Substr. arc + opa": 32 FPS
Result of "Substr. text": 13 FPS
Result of "Substr. text + opa": 13 FPS

> LVGL v9.0.0 dev Benchmark (in csv format)
Weighted FPS: 32
Opa. speed: 96%


(As you can see, the original high-performance scenes can get a result very close to 50 FPS, for example "Image chorma keyed")

I think this is not right. To fix this we can ignore the animation in `LV_DEMO_BENCHMARK_MODE_RENDER_ONLY`, i.e.

modify the code as:

```
for(i = 0; i < RENDER_REPEAT_CNT; i++) {
    lv_refr_now(NULL);
}
```

To be honest, I don't think animation is important in performance evaluation. At least not in this mode `LV_DEMO_BENCHMARK_MODE_RENDER_ONLY`. We already have other modes that takes animation delay into consideration.

kisvegabor:
I looked into it and added a fix: 0d830347f 
It became wrong because meanwhile it was changed how `monitor_cb` reports the elapsed time.

With out the animations nothing will change the scenes and they will be rendered only once when they are loaded.

GorgonMeducer:
OK, I will let you know the result.

GorgonMeducer:
@kisvegabor The https://github.com/lvgl/lvgl/commit/0d830347f622046f686bef3864ee36448764f144 fixed the issue I reported before.

| LV_DEF_REF_PERIOD | Weighted FPS |
| ----------------- | ------------ |
| 33                | 332          |
| 2                 | 332          |
| 1                 | 332          |

The benchmark-refreshing-period-issue disappears.

kisvegabor:
Amazing! Thank you for the feedback!

PR #3422, opened 2022-06-23
chore(docs): add comment in input device part

mysterywolf:
lv_disp_drv_register(&disp_drv) is meaningless in indev part.

### Checkpoints

### DIFF
--- a/docs/porting/indev.md
+++ b/docs/porting/indev.md
@@ -5,6 +5,7 @@
 To register an input device an `lv_indev_drv_t` variable has to be initialized. **Be sure to register at least one display before you register any input devices.**
 
 ```c
+/*Register at least one display before you register any input devices*/
 lv_disp_drv_register(&disp_drv);
 
 static lv_indev_drv_t indev_drv;


embeddedt:
I think it was left in deliberately to repeat the point from the previous sentence about a display needing to be registered first. @kisvegabor?

kisvegabor:
I don't remember what was the original intention. :smile: 

I think we can keep it but let's add a comment too, like 
```c
lv_disp_drv_register(&disp_drv);   /*Register at least one display before you register any input devices*/
```

mysterywolf:
Is there a sequence requirement that we need to register the display device first, then register the input device?

embeddedt:
Yes.

mysterywolf:
OK😁

kisvegabor:
> Is there a sequence requirement that we need to register the display device first, then register the input device?

Each input device is attached to a display. Usually the user has only one display, so if you register the display first LVGL sets that display by default to the new indevs. You can attached a display with `indev_drv.disp = ...`

kisvegabor:
Thank you!

PR #3423, opened 2022-06-23
Fix find_last_focusable in lv_gridnav

JAremko:
# :thinking:


embeddedt:
You're right, it seems that the previous logic would merely increment until it wrapped around to being negative, which is obviously unintended. Thanks.

Issue #3434, opened 2022-06-29
The default LV_MEM_SIZE is too small to finish running benchmark


GorgonMeducer:
The default size of `LV_MEM_SIZE` (which is 64KByte) is too small to finish running the benchmark.

70K - Failed
128K - Failed
200K - Failed
256K - Pass

I highly suspect that there is some memory leaking issue in the newly added layer part (f5759351806d8b14537c0488842e3c7132c72fc9)

I tested it on a 320 * 240 RGB565 env.

### Perform all steps below and tick them with [x]
- [x] Read the [FAQ](https://docs.lvgl.io/master/intro/index.html#faq)
- [x] Check the related part of the [Documentation](https://docs.lvgl.io/)
- [x] Update lvgl to the latest version
- [x] Reproduce the issue in a [Simulator](https://docs.lvgl.io/latest/en/html/get-started/pc-simulator.html)

### Describe the bug

It is blocked in an assert. 
![image](https://user-images.githubusercontent.com/13148491/176562566-0f87e620-9ba1-4e9b-ae6b-30274156a778.png)

Viatorus:
I get the same issue under Linux and ARM.

The problem seems to be:

This isn't called:
https://github.com/lvgl/lvgl/blob/cbff8e83e50fecc2b4b43d661deb91d8d81d6696/src/core/lv_refr.c#L964

Because this returns before:
https://github.com/lvgl/lvgl/blob/cbff8e83e50fecc2b4b43d661deb91d8d81d6696/src/core/lv_refr.c#L918-L922

Why does it return? Because of this:
https://github.com/lvgl/lvgl/blob/cbff8e83e50fecc2b4b43d661deb91d8d81d6696/src/draw/sw/lv_draw_sw_layer.c#L46-L49

So setting `LV_COLOR_SCREEN_TRANSP` to 1 should do the trick.

But the screen doesn't support `LV_COLOR_SCREEN_TRANSP`.... hmm..


**EDIT:**

Just a side note:

Comparing the benchmark results between LVGL 8.2 and LVGL 8.3 nearly all SubStr. benchmark are worse than before.
E.g. Substr. rectangle has 41 fps less than in LVGL 8.2.

GorgonMeducer:
Fixed

kisvegabor:
> Comparing the benchmark results between LVGL 8.2 and LVGL 8.3 nearly all SubStr. benchmark are worse than before.
[...]

PR #3437, opened 2022-06-30
fix(draw): solve memory leaking issue

GorgonMeducer:

Fix issue described in issue
https://github.com/lvgl/lvgl/issues/3434

### Checkpoints

### DIFF
--- a/src/draw/sw/lv_draw_sw_layer.c
+++ b/src/draw/sw/lv_draw_sw_layer.c
@@ -61,7 +61,6 @@ struct _lv_draw_layer_ctx_t * lv_draw_sw_layer_create(struct _lv_draw_ctx_t * dr
             layer_sw_ctx->buf_size_bytes = LV_LAYER_SIMPLE_FALLBACK_BUF_SIZE;
             layer_sw_ctx->base_draw.buf = lv_mem_alloc(layer_sw_ctx->buf_size_bytes);
             if(layer_sw_ctx->base_draw.buf == NULL) {
-                lv_mem_free(layer_ctx);
                 return NULL;
             }
         }
@@ -78,7 +77,6 @@ struct _lv_draw_layer_ctx_t * lv_draw_sw_layer_create(struct _lv_draw_ctx_t * dr
         lv_memset_00(layer_sw_ctx->base_draw.buf, layer_sw_ctx->buf_size_bytes);
         layer_sw_ctx->has_alpha = flags & LV_DRAW_LAYER_FLAG_HAS_ALPHA ? 1 : 0;
         if(layer_sw_ctx->base_draw.buf == NULL) {
-            lv_mem_free(layer_ctx);
             return NULL;
         }
 
--- a/src/draw/sw/lv_draw_sw_layer.c
+++ b/src/draw/sw/lv_draw_sw_layer.c
@@ -43,7 +43,7 @@
 struct _lv_draw_layer_ctx_t * lv_draw_sw_layer_create(struct _lv_draw_ctx_t * draw_ctx, lv_draw_layer_ctx_t * layer_ctx,
                                                       lv_draw_layer_flags_t flags)
 {
-    do{
+    do {
         if(LV_COLOR_SCREEN_TRANSP == 0 && (flags & LV_DRAW_LAYER_FLAG_HAS_ALPHA)) {
             LV_LOG_WARN("Rendering this widget needs LV_COLOR_SCREEN_TRANSP 1");
             break;
--- a/src/draw/sw/lv_draw_sw_layer.c
+++ b/src/draw/sw/lv_draw_sw_layer.c
@@ -43,56 +43,54 @@
 struct _lv_draw_layer_ctx_t * lv_draw_sw_layer_create(struct _lv_draw_ctx_t * draw_ctx, lv_draw_layer_ctx_t * layer_ctx,
                                                       lv_draw_layer_flags_t flags)
 {
-    do{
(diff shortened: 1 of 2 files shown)


src/draw/sw/lv_draw_sw_layer.c
```diff
@@ -43,54 +43,56 @@
 struct _lv_draw_layer_ctx_t * lv_draw_sw_layer_create(struct _lv_draw_ctx_t * draw_ctx, lv_draw_layer_ctx_t * layer_ctx,
                                                       lv_draw_layer_flags_t flags)
 {
-    if(LV_COLOR_SCREEN_TRANSP == 0 && (flags & LV_DRAW_LAYER_FLAG_HAS_ALPHA)) {
-        LV_LOG_WARN("Rendering this widget needs LV_COLOR_SCREEN_TRANSP 1");
-        return NULL;
-    }
+    do {
```
Viatorus: Hmm just my 2 cents but:

What you are basically do with the `do` is introduce an inline function, which early returns on success, otherwise frees the memory and return the failure (null).

I would solve part of the issue as follow:

I would change this line:

https://github.com/lvgl/lvgl/blob/f5759351806d8b14537c0488842e3c7132c72fc9/src/draw/lv_draw_layer.c#L56-L59

to something like:

```c

 lv_draw_layer_ctx_t * init_layer_ctx = draw_ctx->layer_init(draw_ctx, layer_ctx, flags);
 if(init_layer_ctx == NULL) { 
     lv_mem_free(layer_ctx);
     return NULL; 
 } 
 return  init_layer_ctx;
```

And remove all mem_free calls from the layer_init function. This way, the allocation and deallocation stays inside the same function.

src/draw/sw/lv_draw_sw_layer.c
```diff
@@ -43,54 +43,56 @@
...
+        if(LV_COLOR_SCREEN_TRANSP == 0 && (flags & LV_DRAW_LAYER_FLAG_HAS_ALPHA)) {
```
Viatorus: Can you or someone else explain to me the new? meaning of `LV_COLOR_SCREEN_TRANSP`? I remember in LVGL 8.2, this option was only allowed when the display color depth is also 32 bit (which isn't the case for e.g. RGB565 displays)

https://github.com/lvgl/lvgl/blob/58b75f1175504fc7a020cd3b89937c52d954a5f8/docs/overview/display.md?plain=1#L65-L73

src/draw/sw/lv_draw_sw_layer.c
GorgonMeducer: Hi @Viatorus 
Thank you for your suggestion. 
From my perspective, it is reasonable and good! But to be honest, I don't have sufficient information to make such a change and I am not the right guy to make such a decision.

As you can see, 

```c
draw_ctx->layer_init(draw_ctx, layer_ctx, flags);
```
This code is an interface invoking, there is a promise between the interface caller(interface user) and the interface callee (interface implementation). This promise determines what the caller could do and what the callee should do.

Your suggestion is not only fixing this memory leakage issue but also changing this promise, i.e. it is an optimization and refactory of the interface.

As you can see, this memory leakage issue is a serious problem which involves almost everyone. We need a quick fix but not change the behaviours of interface callers and callees.

To make it clear:

- **This PR is to create a quick fix but not optimize or refractory** 
- The change this PR made should not change the behaviour of an existing interface.
- Unless @kisvegabor who designed this interface agrees, I don't have sufficient information to adopt your suggestion

On the other hand, I am only an external contributor, when I making a fix, I prefer to:

- Keep the change minimal and within the scope of the original code
- Fix the issue without crossing interface boundaries

**My fix doesn't stop you to create a separate PR to introduce your optimisation solution later.** 

Please allow me to keep the current solution and wait for the opinion of @kisvegabor.

I hope this is clear.

src/draw/sw/lv_draw_sw_layer.c
GorgonMeducer: I am not the right guy to answer this question, sorry.

src/draw/sw/lv_draw_sw_layer.c
kisvegabor: >  lv_draw_layer_ctx_t * init_layer_ctx = draw_ctx->layer_init(draw_ctx, layer_ctx, flags);
 if(init_layer_ctx == NULL) { 
     lv_mem_free(layer_ctx);
     return NULL; 
 } 
 return  init_layer_ctx;
And remove all mem_free calls from the layer_init function. This way, the allocation and deallocation stays inside the same function.

This solution seems the best to me. Having `lv_mem_free(layer_ctx)` in `lv_draw_sw_layer_create` seems like a mistake to me as `layer_ctx` is not allocated by `lv_draw_sw_layer_create` so it shouldn't free it either.

src/draw/sw/lv_draw_sw_layer.c
kisvegabor: Something like transparent screens are required when LVGL needs to render an ARGB layer. For example render an object which has a radius (the rendered image remains transparent on the corners). 

To allow "render to transparent background" related features `LV_COLOR_SCREEN_TRANSP` needs to be enabled. It's really counter intuitive now and I'll will refactor this part in v9.

src/draw/sw/lv_draw_sw_layer.c
GorgonMeducer: OK, with the interface author's confirmation. I will take your suggestion @Viatorus .

GorgonMeducer:
@kisvegabor Please revew.

kisvegabor:
Perfect, thanks!

PR #3439, opened 2022-07-01
feat(misc): add asynchronous call function cancellation function

FASTSHIFT:

Add the ability to undo asynchronously called functions.

### Checkpoints

### DIFF
--- a/src/misc/lv_async.c
+++ b/src/misc/lv_async.c
@@ -71,27 +71,22 @@ lv_res_t lv_async_call_cancel(lv_async_cb_t async_xcb, void * user_data)
     lv_res_t res = LV_RES_INV;
 
     while(timer != NULL) {
+        /*Find the next timer node*/
+        lv_timer_t * timer_next = lv_timer_get_next(timer);
+
         /*Find async timer callback*/
         if(timer->timer_cb == lv_async_timer_cb) {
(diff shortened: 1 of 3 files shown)


src/misc/lv_async.c
```diff
@@ -65,6 +65,38 @@ lv_res_t lv_async_call(lv_async_cb_t async_xcb, void * user_data)
...
+                /*Record the timer node to be deleted*/
+                lv_timer_t * timer_del = timer;
+
+                /*Find the next timer node*/
+                timer = lv_timer_get_next(timer);
+
+                lv_timer_del(timer_del);
+                lv_mem_free(info);
+                res = LV_RES_OK;
+                continue;
```
kisvegabor: Why do we need this `continue` here? 

I see a problem here (maybe this `continue` is meant to solve it): if a timer is deleted the `timer` variable becomes invalid, so we can not use `lv_timer_get_next(timer);` to get the next timer from it. 

Usually I solve it by getting the next timer in the first line of the loop (where it's still surely valid) and just do `timer = timer_next` at the end of the loop.

FASTSHIFT: Done

kisvegabor:
Merged, thank you!

PR #3445, opened 2022-07-05
fix: make 'lv_font_glyph_dsc_t' local variable initialized to zero

Linjieqiang:

If an object that has automatic storage duration is not initialized explicitly, its value is indeterminate. Indeterminate means it can be anything (including 0). Just because it is zero in the tests you have performed, does not mean it will always be or that you can rely on that behaivour. The behaivour might change even with the same compiler depending on the compilation options and optimization level.

### Checkpoints


kisvegabor:
It's initialized one line [below](https://github.com/lvgl/lvgl/pull/3445/files#diff-d313fc527dbc66b5fd9ad2cda809bce6e7c542b0c4102a45bce59b28c7e6aefdR101) by `lv_font_get_glyph_dsc()`.

Linjieqiang:
Oh. I see. Thanks.

PR #3458, opened 2022-07-07
feat(textarea): add a read-only mode for textarea

hellototoro:
feat(textarea): add a read-only mode for textarea

### Checkpoints

### DIFF
--- a/docs/widgets/core/textarea.md
+++ b/docs/widgets/core/textarea.md
@@ -56,6 +56,9 @@ If `lv_textarea_set_cursor_click_pos(textarea, true)` is applied the cursor will
 ### Hide the cursor
 The cursor is always visible, however it can be a good idea to style it to be visible only in `LV_STATE_FOCUSED` state.
 
+### Read-only mode
+The text area can be configured to be a read-only mode with `lv_textarea_set_readonly_mode(textarea, true)`.
+
 ### One line mode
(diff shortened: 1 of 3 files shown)


kisvegabor:
Hi,

Can't we do the same with `lv_obj_add_state(ta, LV_STATE_DISABLED)`?

hellototoro:
Hi,
If using lv_obj_add_state to replace that, the text area object will not respond to the key events. And the object's color is grey.
Add a read-only mode for the text area not only to display chars normally but also to respond to a key event.
For example, I just want to display a text file, but I don't want to edit it. And the text area object can receive a key event such as ESC.
lv_obj_add_state
read-only mode

kisvegabor:
I see. 

> For example, I just want to display a text file, but I don't want to edit it. And the text area object can receive a key event such as ESC.

In this case why don't you create a simple `lv_label` on an `lv_obj` parent?

hellototoro:
Well, you are right. But should don't the text area has a read-only mode? We can switch between read-write and read-only mode. I think this is useful.

kisvegabor:
I still can't see for what the read-only a text area can be used for where an `lv_label` can't be used. Can you mention some examples?

hellototoro:
Ok, maybe that is redundant.

PR #3459, opened 2022-07-07
fix(draw): compile error with color depth 8

andresovela:

I updated to 8.3 today and I got a compile error because of a non-existent variable. It looks like someone forgot to rename that variable at some point.

### Checkpoints


embeddedt:
Looks like the CI did not pick up on this because none of the runners compile with `LV_COLOR_SCREEN_TRANSP` enabled. It seems that the best option coverage-wise would be to enable it on all of them, since this option adds a lot more code than it disables. @kisvegabor Do you agree?

kisvegabor:
As we are already in v9-dev phase, I'd refactor lv_conf.h a little bit. As part of it `LV_COLOR_SCREEN_TRANSP` should be removed as due to the new layering features it should be always enabled. What do you think?

Issue #3457, opened 2022-07-07
Coordinates and rendered image are out of sync


FASTSHIFT:
### Perform all steps below and tick them with [x]
- [x] Read the [FAQ](https://docs.lvgl.io/master/intro/index.html#faq)
- [x] Check the related part of the [Documentation](https://docs.lvgl.io/)
- [x] Update lvgl to the latest version
- [x] Reproduce the issue in a [Simulator](https://docs.lvgl.io/latest/en/html/get-started/pc-simulator.html)

### Describe the bug
I have a window sliding animation, and I want to load a resource after the window sliding animation ends. This operation of loading resources may take a long time. The code is as follows.
It can be seen that a relatively obvious stutter occurred in the frame before the animation is about to end. Intuitively, the stutter should occur after the animation has completely ended.
This is caused by the rendering mechanism, because after setting the coordinates, lvgl will start rendering in the next cycle.
Is there any way to bypass this problem?

### To Reproduce

```c
static void on_anim_ready(lv_anim_t* a)
{
    LV_LOG_USER("a = %p", a);

    /* Do some time-consuming operations, such as loading resources */
    usleep(500 * 1000);

    LV_LOG_USER("OK");
}

void anim_test(void)
{
    lv_obj_clear_flag(lv_scr_act(), LV_OBJ_FLAG_SCROLLABLE);

    lv_obj_t* ui_root = lv_obj_create(lv_scr_act());
    lv_obj_set_size(ui_root, LV_HOR_RES, LV_VER_RES);

    lv_anim_t a;
    lv_anim_init(&a);
    lv_anim_set_var(&a, ui_root);
    lv_anim_set_exec_cb(&a, (lv_anim_exec_xcb_t)lv_obj_set_y);
    lv_anim_set_values(&a, LV_VER_RES, 0);
    lv_anim_set_time(&a, 1000);
    lv_anim_set_delay(&a, 3000);
    lv_anim_set_ready_cb(&a, on_anim_ready);
    lv_anim_start(&a);
}
```

### Expected behavior
Stuttering occurs after the animation ends instead of the frame before the animation ends.

### Screenshots or video

https://user-images.githubusercontent.com/26767803/177683125-d55c942b-7ddb-44c5-8f31-51db114090ba.mp4

kisvegabor:
I'm not sure is solves the problem, but calling `lv_refr_now()` in `on_anim_ready` might help.

FASTSHIFT:
[...]

PR #3274, opened 2022-04-14
feat(hal): add render_start_cb to disp_drv

FASTSHIFT:
### Description of the feature or fix
Related discussion: https://github.com/lvgl/lvgl/issues/3268

### Checkpoints
- [x] Follow the [styling guide](https://github.com/lvgl/lvgl/blob/master/docs/CODING_STYLE.md)
- [x] Run `code-format.py` from the `scripts` folder. [astyle](http://astyle.sourceforge.net/install.html) needs to be installed.
- [x] Update the documentation

PR #3461, opened 2022-07-07
feat(async): add async call after render

FASTSHIFT:

Related discussion: https://github.com/lvgl/lvgl/issues/3457

### Checkpoints

### DIFF
--- a/src/core/lv_obj.c
+++ b/src/core/lv_obj.c
@@ -132,6 +132,8 @@ void lv_init(void)
     _lv_ll_init(&LV_GC_ROOT(_lv_disp_ll), sizeof(lv_disp_t));
     _lv_ll_init(&LV_GC_ROOT(_lv_indev_ll), sizeof(lv_indev_t));
 
+    _lv_async_init();
+
     /*Initialize the screen refresh system*/
     _lv_refr_init();
 
--- a/src/core/lv_refr.c
+++ b/src/core/lv_refr.c
@@ -11,6 +11,7 @@
 #include "lv_disp.h"
 #include "../hal/lv_hal_tick.h"
 #include "../hal/lv_hal_disp.h"
+#include "../misc/lv_async.h"
 #include "../misc/lv_timer.h"
 #include "../misc/lv_mem.h"
 #include "../misc/lv_math.h"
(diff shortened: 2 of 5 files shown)


kisvegabor:
I think it solves only one particular issue: do something after rendering. But there can be many others, such as do something
- before rendering
- on display flush ready
- before flushing the display
- when logging something
- etc

So I find solving only this case too specific.

FASTSHIFT:
> I think it solves only one particular issue: do something after rendering. But there can be many others, such as do something
> 
> * before rendering
> * on display flush ready
> * before flushing the display
> * when logging something
> * etc
> 
> So I find solving only this case too specific.

Do you mean to complete these events?

kisvegabor:
I think we should leave to the user to manage these triggers as it works the best in his application.

FASTSHIFT:
> I think we should leave to the user to manage these triggers as it works the best in his application.

Something like this interface? https://github.com/lvgl/lvgl/pull/3274

kisvegabor:
Yes, be already have `monitor_cb` for this scenario.

FASTSHIFT:
I need an event that is not related to whether the render is executed or not, but after the render ends, because obj may be outside the screen, lvgl may not be rendered and will not execute `monitor_cb`.

FASTSHIFT:
Consider using a unified event callback function? 
such as:
`disp_refr->driver->render_start_cb(disp_refr->driver);`
change to
`disp_refr->driver->event_cb(disp_refr->driver, LV_DISP_EVENT_RENDER_START, param);`

kisvegabor:
> Consider using a unified event callback function?

Amazing idea. I was planning to do the same with animations too. Are you interested in sending a PR?

FASTSHIFT:
> > Consider using a unified event callback function?
> 
> Amazing idea. I was planning to do the same with animations too. Are you interested in sending a PR?

Yes, but will it cause the current display porting to change too much?

kisvegabor:
Now we are working on v9. This the time to make breaking changes. :slightly_smiling_face:

FASTSHIFT:
I added `refr_finish_cb` to achieve this effect https://github.com/lvgl/lvgl/pull/3502

PR #3466, opened 2022-07-09
fix(cmake-build): Fix example and demo build on esp-idf

wangyz1997:

There are options in `Kconfig` files for users to choose whether to build examples and demos, and `esp-idf` supports `menuconfig`. However, when user enables example or demo to be built, they are actually not built, which is quite confusing. This PR fixes this issue.

### Checkpoints

### DIFF
--- a/env_support/cmake/esp.cmake
+++ b/env_support/cmake/esp.cmake
@@ -21,8 +21,34 @@ if(LV_MICROPYTHON)
                                INTERFACE "-DLV_ATTRIBUTE_FAST_MEM=IRAM_ATTR")
   endif()
 else()
-  idf_component_register(SRCS ${SOURCES} INCLUDE_DIRS ${LVGL_ROOT_DIR}
-                         ${LVGL_ROOT_DIR}/src ${LVGL_ROOT_DIR}/../)
+  if(CONFIG_LV_BUILD_EXAMPLES)
+    file(GLOB_RECURSE EXAMPLE_SOURCES ${LVGL_ROOT_DIR}/examples/*.c)
+  endif()
+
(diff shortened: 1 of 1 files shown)


kisvegabor:
Looks good. I'm just thinking about if it will cause any problem if we enable all the demos by default.  Will ESP ignore the demos (and their assets) during linking if their `*_create()` function is not called?

wangyz1997:
Whether to include code under `demos` folder in the build depends on the new option `LV_BUILD_DEMOS`, which is turned off by default. I have the `LV_USE_DEMO_*` options turned on by default, and they depend on `LV_BUILD_DEMOS`, so when user turns on `LV_BUILD_DEMOS` option, all the `LV_USE_DEMO_*` options are turned on automatically , which I think will be convient.

kisvegabor:
I mean if someone just enables `LV_BUILD_DEMOS` all the demos might require a some MBs of flash if the unused ones are not ignored by the linker. 

As the Kconfig file is used by several platforms I think it's safer to disable all the demos by default and let the users enable what they need.  Once someone is enabling `LV_BUILD_DEMOS` enabling specific demos too shouldn't be a big problem.

wangyz1997:
OK, I'll do that.

wangyz1997:
`Kconfig` file is left unchanged, `LV_BUILD_DEMOS` option is deleted and source file of each demo is added to the build seperately. Looking forward for your review.

kisvegabor:
The only problem I see is that if we create a new demo we shouldn't forget to add it here too.

However, glob-ing the demo sources unconditionally would increase the build time (even if the files are empty due to `#if LV_USE_DEMO_...`s) . Anyway, we don't add demos every day so let's go with this solution. 

 Thank you!

PR #3467, opened 2022-07-10
fix(rt-thread): thread stack align to 8 bytes

mysterywolf:

For  RISC-V chips, thread stack needs to align to 8 bytes when use **float and double** data type variables;
for ARM chips, thread stack needs to align to 8 bytes when use **double** data type variables.

### Checkpoints


kisvegabor:
Thanks!

PR #3468, opened 2022-07-10
chore(make): order the source files in *.mk 

xiaoxiang781216:

- chore(make): sync component.mk with the last directory layout
- chore(make): order the source files in *.mk

### Checkpoints

### DIFF
--- a/component.mk
+++ b/component.mk
@@ -6,24 +6,53 @@ COMPONENT_SRCDIRS := . \
                   src/draw \
                   src/extra \
                   src/font \
-                  src/gpu \
                   src/hal \
                   src/misc \
                   src/widgets \
+                  src/draw/arm2d \
+                  src/draw/nxp \
+                  src/draw/sdl \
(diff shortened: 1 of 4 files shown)


kisvegabor:
For the sake of simplicity I thinking about using [widcards](https://github.com/lvgl/lv_drivers/blob/master/lv_drivers.mk) in `lvgl.mk`. What do you think?

xiaoxiang781216:
> For the sake of simplicity I thinking about using [widcards](https://github.com/lvgl/lv_drivers/blob/master/lv_drivers.mk) in `lvgl.mk`. What do you think?

Yes, it could be, but some template file need to be filter out manually? But anyway it should belong to another patch.

kisvegabor:
Ok, let's merge it now.


> Yes, it could be, but some template file need to be filter out manually? But anyway it should belong to another patch.

All should be disabled by `#if 0`

PR #3478, opened 2022-07-12
fix GCC warning cell_area.x1 and x2 may be used uninitialized in this function

nunojpg:
While the code is correct the rtl depedent initialization can't be tracked by GCC


kisvegabor:
Thank you!

PR #3488, opened 2022-07-15
Fix msg to work with Micropython

amirgon:
- `lv_msg_subscribe_cb_t` does not follow the callback conventions.
Remove the unused `void*` argument to fix that.
- Move `subs_ll` to gc_roots
- Rename "subsribe" to "subscribe"

After this fix, msg can be used in Micropython like this:

```python
import lvgl as lv
lv.init()

MSG = const(42)

class Payload():
  def __init__(self, data):
    self.data=data
  def __repr__(self):
    return self.data

def cb(msg):
  payload = msg.get_payload()
  print(f"Recieved: {payload.__cast__()}")

s = lv.msg_subscribe(MSG, cb, None)

lv.msg_send(MSG, Payload("Hello World!"))
```

@kisvegabor, general comment - when someone adds a new feature with examples and tests, I strongly recommend asking them to add Micropython examples as well.  
Otherwise, the new feature has a good chance of being broken on Micropython.

### DIFF
--- a/docs/others/msg.md
+++ b/docs/others/msg.md
@@ -1,6 +1,6 @@
 # Messaging
 
-Messaging (`lv_msg`) is a classic []publisher subscriber](https://en.wikipedia.org/wiki/Publish%E2%80%93subscribe_pattern) implementation for LVGL.
+Messaging (`lv_msg`) is a classic [publisher subscriber](https://en.wikipedia.org/wiki/Publish%E2%80%93subscribe_pattern) implementation for LVGL.
 
 ## IDs
 Both the publishers and the subscribers needs to know the message identifiers.
@@ -33,9 +33,8 @@ lv_msg_send(MSG_USER_NAME_CHANGED, "John Smith");
 The callback should look like this:
 ```c
 
-static void user_name_subscriber_cb(void * s, lv_msg_t * m)
+static void user_name_subscriber_cb(lv_msg_t * m)
 {
-    /*s: a subscriber obeject, can be used to unscubscribe*/
     /*m: a message object with the msg_id, payload, and user_data (set durung subscription)*/
 
     ...do something...
--- a/examples/others/msg/lv_example_msg_1.c
+++ b/examples/others/msg/lv_example_msg_1.c
@@ -24,7 +24,7 @@ void lv_example_msg_1(void)
     lv_obj_align(label, LV_ALIGN_CENTER, 0, 30);
 
     /*Subscribe the label to a message. Also use the user_data to set a format string here.*/
-    lv_msg_subsribe_obj(MSG_NEW_TEMPERATURE, label, "%d °C");
+    lv_msg_subscribe_obj(MSG_NEW_TEMPERATURE, label, "%d °C");
 }
 
 static void slider_event_cb(lv_event_t * e)
--- a/examples/others/msg/lv_example_msg_1.py
+++ b/examples/others/msg/lv_example_msg_1.py
@@ -3,7 +3,7 @@
 MSG_NEW_TEMPERATURE = const(1)
 
 # Define the object that will be sent as msg payload
-class Temprature:
+class Temperature:
     def __init__(self, value):
         self.value = value
     def __repr__(self):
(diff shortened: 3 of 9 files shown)


kisvegabor:
Thank you! The docs needs to slightly updated too due to the new `lv_msg_subscribe_cb_t`.

> general comment - when someone adds a new feature with examples and tests, I strongly recommend asking them to add Micropython examples as well.
Otherwise, the new feature has a good chance of being broken on Micropython.

True. Actually we should do the same in C as well. Especially now, when we are doing major reworks in v9.

amirgon:
> The docs needs to slightly updated too due to the new `lv_msg_subscribe_cb_t`.

Done on 97a471bb1

amirgon:
@kisvegabor Why did you define `LV_EVENT_MSG_RECEIVED` as an external event, and not part of the `lv_event_code_t` enum?

kisvegabor:
Thank you!

> Why did you define LV_EVENT_MSG_RECEIVED as an external event, and not part of the lv_event_code_t enum?

Because `lv_msg` is just an optional extension and not a core module. I guess it's a problem because of auto-completion and discovering related names in MP. 

---

FYI, I merged #3501 as the first major rework for v9. It work in C with `lv_conf.h` but Kconfig and MP related configs are not updated yet. 

I merged it to not block other features to be merged.

amirgon:
> I guess it's a problem because of auto-completion and discovering related names in MP.

Compare 
```python
slider.add_event_cb(slider_event_cb, lv.EVENT.VALUE_CHANGED, None)
```

to 
```python
label.add_event_cb(label_event_cb, lv.EVENT_MSG_RECEIVED.value, None)
```

So it's supported, but it's awkward. It might be confusing for the user that is one case he needs to use the form `lv.EVENT.VALUE_CHANGED` and in the other case the form `lv.EVENT_MSG_RECEIVED.value`.

> Because `lv_msg` is just an optional extension and not a core module.

So what? What's wrong with adding optional extension events to `lv_event_code_t`? What do you lose by adding them? You can add them in `#ifdef` if you want to disable them when this feature is not enabled, in the same way you disable the call to `lv_msg_init`.

amirgon:
> I merged it to not block other features to be merged.

When you do that - these other features are not tested in Micropython CI.
Then new additional commits will be added that also break Micropython but we won't know about it because the Micropython CI is already broken. And by the time we get to handle those, new commits will be added, that also break Micropython.

I think we might be missing the whole  point of CI when working this way.

I was hoping that if someone adds something to LVGL, that same person would also be responsible of making sure that all CI pipelines are still passing, including Micropython.  
Otherwise, if we give up Micropython CI at this point, that leaves me with a broken LVGL when trying to use it with Micropython while the original contributor of that change believes his job is done since his change was accepted to LVGL. 
 
I'm not sure I'll have the capacity of fixing every LVGL issue that breaks Micropython when working in this manner.

embeddedt:
I have to agree with @amirgon here; I thought the CI breakage was temporary and would be resolved before anything else was merged, but it's sounding like this may not be the case.

kisvegabor:
> So what? What's wrong with adding optional extension events to lv_event_code_t? What do you lose by adding them? You can add them in #ifdef if you want to disable them when this feature is not enabled, in the same way you disable the call to lv_msg_init.

My goal was to create a modular architecture where new widgets/features doesn't have to be integral part of the core LVGL but can be created and maintained independently. E.g. a company can add it's own layout manager, event codes or even style properties without touching LVGL. And my idea was that even in the `lvgl` repo we can separate the core features from user contributed parts. However, I tend to see that it can lead to inconvenient cases an over-complication. So I also agree that once something is in the `lvgl` repo we can consider it as a core feature and we can freely use `#if..#endif`. 

> When you do that - these other features are not tested in Micropython CI.

I also feel that it sounds ugly. I try to add all the major refactoring as quick as I can (probably this week) and after that we will be on a somewhat stable ground. I expect rapid changes in the upcoming days, and I think it'd be wasted time to fix the CI after each iteration. 

I also wanted to merge these changes to `master` to increase the possibility of getting feedback from the direction of the changes.

We really need to a make stronger policy about testing. We need to be more strict about merging features only if at least basics test are added too.

amirgon:
> So I also agree that once something is in the `lvgl` repo we can consider it as a core feature and we can freely use `#if..#endif`.

Done in 0ffebbfb6

kisvegabor:
Thanks, I'll take care of other events and styles.

kisvegabor:
Is it ready to merge?

amirgon:
> Is it ready to merge?

I think it is.

PR #3294, opened 2022-04-25
Part 3. Image decoder rework

X-Ryl669:
### Description of the feature or fix

To follow the 2 previous PR in this suite, this adds the decoder rework for the new image code.
To list the changes:

1. Remove of `img_decoder_get_info` that used to open and decode the encoded image data (but not render). The closest equivalent is `img_decoder_accept` that just say if a given decoder can accept the image's data (no more that a memcmp for most format, either for the extension for files or the magic header for data). This is a lot faster than the previous system to find which decoder to use.
2. Reworked the `img_decoder_decode` to now accept the changes to the image source format and a flag to squeeze decoding time whenever possible when only metadata are required. This should be more or less what get_info used to do in all cases, now it's done only when required and not wasted if the actual action is to decode the picture.
3. Added animation timers to the image class, so both rlottie, ffmpeg, and gif are now natively handled by the image class instead of their own object.
4. Reworked all decoders to remove (most of) their API. Everything is now commonly handler by the image class, so the only function left is **lv_*format*_init()**, that's silently called by the `lv_extra.c` file.
5. Moved the native RGB image decoder to its own decoder library (cleaner) and can probably be optimized away if not used.
6. Fixed few errors reported by valgrind while testing (see first commits in this PR)
7. Reworked the image cache to match what was in #2904. It's still not the best version that accounts for total memory usage. Will be done later, since we haven't conclude about what to do exactly. 

### Checkpoints
- [X] Follow the [styling guide](https://github.com/lvgl/lvgl/blob/master/docs/CODING_STYLE.md)
- [X] Run `code-format.py` from the `scripts` folder. [astyle](http://astyle.sourceforge.net/install.html) needs to be installed.
- [X] Update the documentation

PR #3499, opened 2022-07-18
feat(obj): add object draw cache support.

FASTSHIFT:

Buffer the image after drawing the object to improve the rendering performance of complex objects.

```c
void obj_draw_cache_test(void)
{
    lv_obj_draw_cache_set_enable(obj, true);
    lv_obj_draw_cache_invalidate(obj);
    lv_obj_center(obj);
}
```

### Checkpoints


FASTSHIFT:
Do I need to encapsulate a dedicated memory alloc interface?

src/core/lv_obj_draw_cache.c
```diff
@@ -0,0 +1,189 @@
...
+    # error "Need to enable 'LV_USE_SNAPSHOT'"
+#endif
+
+/**********************
+ *      TYPEDEFS
+ **********************/
+
+typedef struct _lv_obj_draw_cache_t {
+    lv_img_dsc_t img_dsc;
+    bool invalidate;
```
kisvegabor:
```suggestion
    bool invalid;
```

src/core/lv_obj_draw_cache.c
```diff
@@ -0,0 +1,189 @@
...
+    bool draw_normal;
```
kisvegabor:
```suggestion
    bool skip_cache;
```

kisvegabor:
Nice and clean implementation. Looks great! 

I had two suggestions about naming but nothing serious. 

I've already prepared a patch with some refactoring. I'll create in minutes. I'd like to merge that first as it affects to whole library and then we can easily refactor this patch if needed.

kisvegabor:
PR is opened: #3501

kisvegabor:
Although Kconfig and Micropython is not updated yet I merged #3501 to not block features like this. Could you solve the minor conflicts?

FASTSHIFT:
@kisvegabor 
It seems that the codes generated by `lv_conf_internal_gen.py` and `code-format.py` are not consistent

kisvegabor:
@xiaoxiang781216 made some updates on this but it seems it still not perfect.

kisvegabor:
@mariotaku @tanaytekin

If there were an option to automatically snapshot all widgets we could hugely simplify the SDL and OpenGLES engines. We could let LVGL to SW render everything into buffers and when it draws the pre-rendered images the GPUs can save them into texture and use those textures later on.

To make it work effectively maybe it'd be better to the snapshot the widgets without their children. This way when a children is updated the parent doesn't needs to be redrawn. 

@FASTSHIFT 
I took a second look at `lv_obj_draw_cache_invalidate()`. I think it should work the other way around. `lv_obj_invalidate` should call it automatically, to mark the widget for re-snapshoting.  What do you think?

FASTSHIFT:
> @mariotaku @tanaytekin
> 
> If there were an option to automatically snapshot all widgets we could hugely simplify the SDL and OpenGLES engines. We could let LVGL to SW render everything into buffers and when it draws the pre-rendered images the GPUs can save them into texture and use those textures later on.
> 
> To make it work effectively maybe it'd be better to the snapshot the widgets without their children. This way when a children is updated the parent doesn't needs to be redrawn.
> 
> @FASTSHIFT I took a second look at `lv_obj_draw_cache_invalidate()`. I think it should work the other way around. `lv_obj_invalidate` should call it automatically, to mark the widget for re-snapshoting. What do you think?

For moving position and rotation and scaling, there is no need to redraw. We may need to expand the parameters of `lv_obj_invalidate()` to reduce unnecessary redraw operations.

src/core/lv_obj_draw_cache.c
FASTSHIFT: Done

src/core/lv_obj_draw_cache.c
FASTSHIFT: Done

mariotaku:
I have similar concern like @FASTSHIFT . Should a label cache its content including glyphs rendered? Should a button cache every animation frame when it gets pressed? The ability to first render to a buffer then draw it will be very useful for a scrollable list, but extra care should be taken when we decide what to cache and what to render directly...

FASTSHIFT:
I invalidate draw cache by default in `lv_obj_invalidate()`.
Added a `lv_obj_draw_cache_invalidate_cancel()` API cancel invalidate in `set_style_x/y`.

src/core/lv_obj_style_gen.c
```diff
@@ -54,6 +55,9 @@ void lv_obj_set_style_x(struct _lv_obj_t * obj, lv_coord_t value, lv_style_selec
         .num = (int32_t)value
     };
     lv_obj_set_local_style_prop(obj, LV_STYLE_X, v, selector);
+#if LV_USE_OBJ_DRAW_CACHE
```
kisvegabor, line 58: This file is auto generated so we can't edit it.

src/core/lv_obj_pos.c
```diff
@@ -862,7 +863,9 @@ void lv_obj_invalidate(const lv_obj_t * obj)
     obj_coords.y2 += ext_size;
 
     lv_obj_invalidate_area(obj, &obj_coords);
-
+#if LV_USE_OBJ_DRAW_CACHE
+    lv_obj_draw_cache_invalidate(obj);
```
kisvegabor, line 867: It should be added rather to `lv_obj_invalidate_area()` as this the last invalidate-like function in the chain.

src/core/lv_obj_draw_cache.c
```diff
@@ -0,0 +1,226 @@
...
+}
+
+void lv_obj_draw_cache_invalidate_cancel(lv_obj_t * obj)
+{
+    LV_ASSERT_NULL(obj);
+    if(!lv_obj_has_draw_cache(obj)) {
+        return;
+    }
+
+    obj->spec_attr->draw_cache->invalid = false;
```
kisvegabor, line 122: I'm not sure this is the right approach. If an object's `draw_cache` was already correctly invalidated (e.g. style changed) and it's immediately followed by a position change the original invalidation will be cancelled too.

kisvegabor (commented):
I think the best would be to add something like this to the two `lv_obj_invalidate`-s in `lv_obj_pos.c` `lv_obj_move_to()`:
```c
bool draw_cache_inv_old = obj->spec_attr->draw_cache->invalid;
/*Invalidate the original area*/
lv_obj_invalidate(obj);

....
/*Invalidate the new area*/
lv_obj_invalidate(obj);

/*Position change shouldn't cause cache invalidation so restore it*/
obj->spec_attr->draw_cache->invalid = draw_cache_inv_old;
```

kisvegabor:
> The ability to first render to a buffer then draw it will be very useful for a scrollable list, but extra care should be taken when we decide what to cache and what to render directly...

It'd require the implement the whole drawing layer for SDL to support cases when caching is not optimal. Or do you mean just skipping same "texture level caching" when a snapshot is likely to be rendered only once?

tanaytekin:
> @mariotaku @tanaytekin
> 
> If there were an option to automatically snapshot all widgets we could hugely simplify the SDL and OpenGLES engines. We could let LVGL to SW render everything into buffers and when it draws the pre-rendered images the GPUs can save them into texture and use those textures later on.
> 
> To make it work effectively maybe it'd be better to the snapshot the widgets without their children. This way when a children is updated the parent doesn't needs to be redrawn.
> 
> @FASTSHIFT I took a second look at `lv_obj_draw_cache_invalidate()`. I think it should work the other way around. `lv_obj_invalidate` should call it automatically, to mark the widget for re-snapshoting. What do you think?

As far as I know, minimizing the cpu-gpu communication is huge plus. Upload textures once and only set transformation via matrices using shaders when you need to draw. I think for that need to implement shadows - filters etc. (This seems like a very difficult job, I don't know if it's possible) in shader side as well.

kisvegabor:
> I think for that need to implement shadows - filters etc. (This seems like a very difficult job, I don't know if it's possible) in shader side as well.

Why? The idea is to make the SW render draw everything into buffer and the GPU gets a ready to use pixel arrays (bitmaps). So no shadow or other magic needs to be implemented in the GPU, only caching the bitmaps into textures and blending them.

mariotaku:
@kisvegabor I would implement drawing with cached object, as well as the way it draws (primitive elements like rectangles, lines) now. So when the object is rapidly changing, we can still have a good performance.

For other rendering backends, they can also choose to implement one or both. If they only implement drawing with cached object, the object is rendered by CPU. When both are implemented, the object will be rendered by GPU on an off-screen texture.

kisvegabor:
Okay, your call. :slightly_smiling_face: Especially as the SDL rendering is quite full featured. 

What about OpenGL ES? Wouldn't it worth to implement this mixed solution to have something and add more GPU rendering features as a separate step?

mariotaku:
@kisvegabor This would be great. I'll think about a way to introduce "texture" type in lvgl for easier GPU render implementation after this PR finished.

kisvegabor:
#3294 will make it simpler to define new image formats.

tanaytekin:
> > I think for that need to implement shadows - filters etc. (This seems like a very difficult job, I don't know if it's possible) in shader side as well.
> 
> Why? The idea is to make the SW render draw everything into buffer and the GPU gets a ready to use pixel arrays (bitmaps). So no shadow or other magic needs to be implemented in the GPU, only caching the bitmaps into textures and blending them.


> > I think for that need to implement shadows - filters etc. (This seems like a very difficult job, I don't know if it's possible) in shader side as well.
> 
> Why? The idea is to make the SW render draw everything into buffer and the GPU gets a ready to use pixel arrays (bitmaps). So no shadow or other magic needs to be implemented in the GPU, only caching the bitmaps into textures and blending them.

If the filters are changed in a animation for example, need to create textures every frame.(Not ideal)  Other than that, the idea seems pretty good. Also it might be helpful to look at [imgui](https://github.com/ocornut/imgui) they may have implemented it in a similar way.

FASTSHIFT:
@kisvegabor 
Can we restart this PR ? :)

FASTSHIFT:
The [cacheable-container](https://support.touchgfx.com/4.18/docs/development/ui-development/scenarios/achieving-better-performance-with-cacheable-container) of TouchGFX is manually updated by the user, can we also consider a similar approach? This makes the design simpler and easier for users to control.

kisvegabor:
I agree to have this feature. But please wait a little with the implementation. There are too many moving parts now and it's already difficult and error prone to marge the feature branches. 

[Parallel rendering](https://github.com/lvgl/lvgl/issues/3700) improves how "layers" work and I think we should build this feature on top of it. E.g. "layer caching" or so.

PR #3501, opened 2022-07-18
Refactor lv_conf and some API

kisvegabor:

In this PR I updated the draw related section of lv_conf.h to fit better the current vision. I also added a more direct stdlib wrapper and removed `lv_mem_buf_get()` as TLSF is fast enough to allocated/free directly.

Kconfig is not updated yet and some help would be very welcome with it from someone who hay more experience with it and can test it.

Please comment here if you can help with Kconfig and/or open a new PR targeting this branch,

### Checkpoints


xiaoxiang781216:
@kisvegabor for lv_mem_buf_get, my concern isn't speed, but the memory fragment, if we allocate the memory from the global heap for the temporary buffer.

kisvegabor:
I found that actually `lv_mem_buf_get` causes fragmentation. E.g. if a process allocates 10kB, then uses `lv_mem_buf_get` to allocate 20 bytes and the first 10kB is freed, a 10 kB hole remains there because the 20 bytes wasn't freed (as it might be used later by an other process). 

In other words if all methods fully cleaned up after themselves, the fragmentation would be lower as there are no "floating" blocks.   

It was a real problem when I tried to rotate nested objects (not images but [layers](https://docs.lvgl.io/master/overview/style.html#opacity-blend-modes-and-transformations)) and 100kBs of memory were required. But it was out of memory even with 1MB of heap as small `lv_mem_buf_get` block were here and there from the previous methods.

kisvegabor:
I think it's better to merge it as it is to not block other fixes and features. 

TODOs:
- Update Kconfig
- Update MicroPython config

amirgon:
@kisvegabor - I don't get it. Where did `LV_COLOR_16_SWAP` go? We rely on it in ili9xxx drivers and probably on other places. [See lv_micropython README](https://github.com/lvgl/lv_micropython#esp32-port) for example.

kisvegabor:
It will be added back in a different form. We will add a `post_process_cb` to the driver or `draw_ctx` which can convert the rendered image to any format. Probably it will be even faster and more flexible for sure.

amirgon:
> It will be added back in a different form.

I see.
This might be better, but will also require more effort and more time to migrate Micropython to this change.
I cannot align Micropython to latest LVGL until this is resolved.

How about splitting this to a different PR? If you added `LV_COLOR_16_SWAP` back now, I could fix Micropython faster and align it to LVGL master.

kisvegabor:
For first I'd rather try adding `post_process_cb` tomorrow. It should support an equivalent of `LV_COLOR_16_SWAP` out of the box as it's very common.

kisvegabor:
@amirgon 
I added swapped 16 bit support. All you need to do is setting
```c
 disp_drv.color_format = LV_COLOR_FORMAT_NATIVE_REVERSE;
```

Regarding the Micropython tests, please let me know
1. how to run the tests locally?
2. where to send a PR with the fixes?

embeddedt:
> Regarding the Micropython tests, please let me know
> 1. how to run the tests locally?

You can reference the [CI workflow](https://github.com/lvgl/lvgl/blob/master/.github/workflows/build_micropython.yml), but essentially:

1. clone `lv_micropython`
2. run `git submodule update --init --recursive lib/lv_bindings`
3. run `make -C ports/unix submodules`
4. update the LVGL submodule inside `lv_bindings` to point to `master`, 
5. Build the cross compiler: `make -j$(nproc) -C mpy-cross`
6. Build the unix port: `make -j $(nproc) -C ports/unix VARIANT=dev DEBUG=1`
7. Run the actual tests: `lib/lv_bindings/tests/run.sh`

I typed that in a bit of a hurry but it should be the gist of what you need.

> 2. where to send a PR with the fixes?

I believe the fixes will need to be in `lv_binding_micropython`. `lv_micropython` doesn't contain much LVGL-specific code.

amirgon:
> ```c
> disp_drv.color_format = LV_COLOR_FORMAT_NATIVE_REVERSE;
> ```

Actually, it's very good that swapped-16-bit can now be determined on **runtime**.  

This aligns well with dynamic language like Micropython, where you want to build the firmware only once and configure everything on runtime. If only lv_drivers were designed that way, we wouldn't need to create Micropython-specific drivers.

How about `LV_COLOR_DEPTH`? Today we set `LV_COLOR_DEPTH=16` for some displays and `LV_COLOR_DEPTH=32` for others, so we need to build different firmware for different classes of displays.  
But now on v9, would you consider converting color-depth too into a runtime configuration?

> Regarding the Micropython tests, please let me know

@embeddedt summarized this very well! I don't have anything to add.

kisvegabor:
@amirgon 
I've fixed the py examples in LVGL.  In the [dev/v9](https://github.com/lvgl/lv_binding_micropython/tree/dev/v9) branch of `lv_binding_micropython` I updated `lv_conf.h`and added some fixes. Now the tests passed with the unix build. 

>  This aligns well with dynamic language like Micropython, where you want to build the firmware only once and configure everything on runtime. If only lv_drivers were designed that way, we wouldn't need to create Micropython-specific drivers.

I already made some experiments with a new driver architecture integrated in to the lvgl repo. You can read a summary about it [here](https://github.com/lvgl/lvgl/issues/2720#issuecomment-1112515857). It allows you to dynamically create new displays with runtime configured parameters. 

> How about LV_COLOR_DEPTH? Today we set LV_COLOR_DEPTH=16 for some displays and LV_COLOR_DEPTH=32 for others, so we need to build different firmware for different classes of displays.
But now on v9, would you consider converting color-depth too into a runtime configuration?

I'm thinking a lot about it but always come to the conclusion that it'd make the code significantly more complex and slower.

This is how it work now in v9:
- LVGL always render with 8, 16 or 32 bit color depth
- Before flushing the rendered image is converted into `drv->color_format`
- The conversation is done by the `buffer_convert` callback in `draw_ctx`. So the user can override this function to add any custom color conversations. Besides GPUs also can overwrite it to make a better job then SW conversation. 

The current "swap" implementation is surely not the fastest possible. Maybe something like would be better but it needs to be tried out:
```c
buf32[x] = (buf32[x] & 0xff00ff00 >> 8) + (buf32[x] & 0x00ff00ff << 8);
```

amirgon:
> In the [dev/v9](https://github.com/lvgl/lv_binding_micropython/tree/dev/v9) branch of `lv_binding_micropython` I updated `lv_conf.h`and added some fixes. Now the tests passed with the unix build.

Didn't you want to create a PR for this?  A PR could be useful for discussions.   
So, for now I'll write my comments here. If you decide to open a PR we can move them over there:

I can see you removed `LV_COLOR_16_SWAP` but I think there are still pieces missing. For example, on ili9xxx we probably need to set `disp_drv.color_format` on certain cases. Also the README should be fixed.

Without fixing this, the ESP32 would break on runtime. Unfortunately we don't have a way to test this in the CI today. (although I once explored this possibility with Xtensa QEMU but it seemed like a lot of work)

More related questions: 

- What happened to `LV_MEM_BUF_MAX_NUM`, `LV_CIRCLE_CACHE_SIZE`, `LV_LAYER_*`? Is there another caching mechanism in place now? What happened to `LV_USE_FONT_SUBPX`?
- Why did you disable `LV_USE_BIDI` and `LV_USE_ARABIC_PERSIAN_CHARS`? They are currently enabled on lv_micropython. Is this a mistake?
- Why did you disable some widgets as Snapshot, Gridnav, Msg etc.? They were are currently enabled on lv_micropython.

Unrelated question: Why is `LV_COLOR_CHROMA_KEY` configured on compile-time instead of runtime?

> I already made some experiments with a new driver architecture integrated in to the lvgl repo. You can read a summary about it [here](https://github.com/lvgl/lvgl/issues/2720#issuecomment-1112515857). It allows you to dynamically create new displays with runtime configured parameters.

I think this is very interesting for Micropython or any other binding. Using the same LVGL drivers on C and other languages could be very useful instead of maintaining a different set of drivers for each language binding. Today one of the limitations of the Micropython bindings is the limited variety of drivers it provides for combinations of MCUs and displays. We have relatively large number of options for ESP32 but few options for other MCUs.

If all the parameters of the LVGL C drivers could be configured on runtime, and if they had a well defined C API that obeys the code conventions we defined, we could use the binding script to automatically convert the API of these drivers to Micropython and with very little effort use them in Micropython.

I understand this is all just a proof-of-concept for now. If this becomes real I would be very interested in binding those drivers to Micropython and providing a clean automatically generated API for them.

> * The conversation is done by the `buffer_convert` callback in `draw_ctx`. So the user can override this function to add any custom color conversations. Besides GPUs also can overwrite it to make a better job then SW conversation.

So can we always build with color depth of 32, and convert to color depth 16 (when necessary) with `buffer_convert`?  
I'm not sure that's a good idea because it would consume x2 memory for display buffers.... but it might be useful in cases we have enough memory.

> The current "swap" implementation is surely not the fastest possible.

So the move from `LV_COLOR_16_SWAP` to `disp_drv.color_format` has performance penalty on render time?  
Did you measure how bad it is? Maybe it's worth keeping the option of compile-time macro if we want to improve performance?

kisvegabor:
> Didn't you want to create a PR for this? A PR could be useful for discussions.

Sure, but I wasn't sure which branch shall I target with it. Note that `lvgl` is not stable now and won't be for a couple of months.

> I can see you removed LV_COLOR_16_SWAP but I think there are still pieces missing. For example, on ili9xxx we probably need to set disp_drv.color_format on certain cases. Also the README should be fixed.

My first goal was to fix the test only. I can't test on embedded HW now but I believe so far the 2 critical changes are the new `lv_conf.h` and `drv.color_format`.

> What happened to LV_MEM_BUF_MAX_NUM, LV_CIRCLE_CACHE_SIZE, LV_LAYER_*? Is there another caching mechanism in place now? What happened to LV_USE_FONT_SUBPX?

- `LV_MEM_BUF_MAX_NUM`: removed because it increases fragmentation. See [here](https://github.com/lvgl/lvgl/pull/3501#issuecomment-1187666344). I've tested that it doesn't cause performance drop (at least with the built in TLSF allocator).
- `LV_CIRCLE_CACHE_SIZE` renamed to `LV_DRAW_SW_CIRCLE_CACHE_SIZE`
- `LV_LAYER_*` renamed to `LV_DRAW_SW_LAYER_*`
- `LV_USE_FONT_SUBPX` renamed to `LV_DRAW_SW_FONT_SUBPX`

> Why did you disable LV_USE_BIDI and LV_USE_ARABIC_PERSIAN_CHARS? They are currently enabled on lv_micropython. Is this a mistake?
Why did you disable some widgets as Snapshot, Gridnav, Msg etc.? They were are currently enabled on lv_micropython.

These are not enabled only by mistake. (snapshot is not working well yet) 

> Unrelated question: Why is LV_COLOR_CHROMA_KEY configured on compile-time instead of runtime?

I'm planning to completely remove it and move this property to an `style_img_chroma_key_color` or so to make it configurable on image widget level. 

> If all the parameters of the LVGL C drivers could be configured on runtime, and if they had a well defined C API that obeys the code conventions we defined, we could use the binding script to automatically convert the API of these drivers to Micropython and with very little effort use them in Micropython.

Great! As the drivers will be integral part of lvgl (just like widgets) the binding will be generated automatically. Assuming we follow the conventions. :slightly_smiling_face: 

> So can we always build with color depth of 32, and convert to color depth 16 (when necessary) with buffer_convert?
I'm not sure that's a good idea because it would consume x2 memory for display buffers.... but it might be useful in cases we have enough memory.

It also useful when you have multiple display driven by a single MCU and the displays have different color format. E.g. a TFT and an OLED. 

> So the move from LV_COLOR_16_SWAP to disp_drv.color_format has performance penalty on render time?
Did you measure how bad it is? Maybe it's worth keeping the option of compile-time macro if we want to improve performance?

I really need to measure it but here is a quick calculation: assuming it takes 4 instructions to convert 1 pixel, the display has 320x240 resolution, and the MCU is 100 MHz.
```
4 x 320 x 240 / 100M = ~3ms.
```

So 3 ms is lost on the conversation. The worst case is when the screen is very simple (e.g.  just a plain white rectangle) because it this case `LV_COLOR_16_SWAP` is as fast as normal RGB format (as LVGL only fill a buffer with 16 bit integers). So in this case we lose 3 ms.

But if opacity is used, where colors needs to be mixed, `LV_COLOR_16_SWAP` required extra logic to handle. (Almost like having 4 color channels (red, green_high, green_low, blue) instead of 3, so ~33% loss on color mixing).

All things considered I think the performance drop is not significant but we will see.

amirgon:
> Sure, but I wasn't sure which branch shall I target with it. Note that `lvgl` is not stable now and won't be for a couple of months.

Actually, I've [posted about this](https://forum.lvgl.io/t/micropython-v1-19-1-and-lvgl-v8-3/9340) in the forum more than a week ago:
<i>
&nbsp;&nbsp;&nbsp;&nbsp;lv_micropython master branch will now start tracking LVGL v9 development, and may include breaking changes.
&nbsp;&nbsp;&nbsp;&nbsp;For Micropython + LVGL v8, please use the [lv_micropython release/v8](https://github.com/lvgl/lv_micropython/tree/release/v8) branch.
</i>

> My first goal was to fix the test only. I can't test on embedded HW now but I believe so far the 2 critical changes are the new `lv_conf.h` and `drv.color_format`.

I understand that LVGL master is not stable now, but my intention was to keep lv_micropython *relatively* stable while tracking LVGL v9 development.  

This is possible with git submodules - I can update the LVGL submodule only when tests pass and when I'm not aware of special problems.  
But in this case, without fixing the color swap issue, we are knowingly breaking the ESP32 port.  
I don't feel comfortable doing that. I prefer people to use the master branch, report problems etc. and not ignore it because it's broken by definition.

So my preference is to fix the problem we know about now (color swap), and not just do the minimum to make tests pass  leaving the master branch broken.

I'll try fixing that on my side and open that PR soon.

**Update**: Tried [dev/v9](https://github.com/lvgl/lv_binding_micropython/tree/dev/v9) branch now and looks like freetype is broken? `lv_example_freetype_1.py` test crashes with latest LVGL.

> Great! As the drivers will be integral part of lvgl (just like widgets) the binding will be generated automatically. Assuming we follow the conventions.

Looking forward to it! :+1: 

> I really need to measure it but here is a quick calculation: assuming it takes 4 instructions to convert 1 pixel, the display has 320x240 resolution, and the MCU is 100 MHz.
> 
> ```
> 4 x 320 x 240 / 100M = ~3ms.
> ```

Wait, are you taking into account cache misses? Or do you assume the MCU has a really large data cache? Or are cache misses already being accounted for?

kisvegabor:
> Actually, I've [posted about this](https://forum.lvgl.io/t/micropython-v1-19-1-and-lvgl-v8-3/9340) in the forum more than a week ago:

Great! :+1: 

> So my preference is to fix the problem we know about now (color swap), and not just do the minimum to make tests pass leaving the master branch broken.

Fair enough. Kconfig is not updated yet. Is it a blocker for ESP or do you use lv_conf.h anyway?

> Tried [dev/v9](https://github.com/lvgl/lv_binding_micropython/tree/dev/v9) branch now and looks like freetype is broken? lv_example_freetype_1.py test crashes with latest LVGL.

Fixed now it `lvgl` `master`. Snapshot is working again too. 


> Wait, are you taking into account cache misses? Or do you assume the MCU has a really large data cache? Or are cache misses already being accounted for?

Good question. I intentionally left it from the calculations because if draw buffer(s) are not in an internal RAM drawing is awfully slow anyway. 

Meanwhile I made some tests on a 800x480 STM32F769 (200MHz) and optimized the algorithm. It takes 2.3 ms to convert the whole screen (1.2 instructions per pixel). 

On ESP with 320x240 it'd be 320 x 240 x 1.2 / 240M = 0.4ms (assuming the same level of optimization and similarly effective assembly code)

amirgon:
I've opened https://github.com/lvgl/lv_binding_micropython/pull/225 to align lv_binding_micropython to this change.

arbrauns:
> I think it's better to merge it as it is to not block other fixes and features.
> 
> TODOs:
> 
> * Update Kconfig
> * Update MicroPython config

Where's the tracking issue for the Kconfig update? If there isn't one, I'm not very surprised it never happened. The fact that this 2000-line patch set was squash-merged down to a single commit is also a bit of a disaster if you ask me.

kisvegabor:
`master` is under development and Kconfig is not updated yet. Please use `release/v8.3`.

> The fact that this 2000-line patch set was squash-merged down to a single commit is also a bit of a disaster if you ask me.

TBH I don't know which is better:
1. having 50 commits with experiments, trivial fixes messing up the history but being "honest about the development process"
2. having 1 long commit which can be ridiculously large. 

We have chosen 2) mainly because of the cleaner history in which it's easier to find where an issue was introduced and it's easier to generate a change log.

arbrauns:
> TBH I don't know which is better:

I prefer 3. none of the above, and instead cleaning up the commit history before the PR is merged. That way you can have as many tiny fixup commits as you want during development, but the final history that ends up in `master` is still sensible and `git bisect`-able.

kisvegabor:
Ideally it'd work, and really this is the best option. However it needs all the contributors to clean up the history, have nice commit messages, etc. Or if they don't do it, or can't do it we need to help with that.  And unfortunately we don't have enough core developers to fix or force fixing all PRs, so we have chosen the simpler but nicer way.

PR #3510, opened 2022-07-20
feat(img):do not show "no data" if src is NULL

xaowang96:
No need to show "no data" when the source of the image is NULL.

Signed-off-by: wangxuedong <wangxuedong@xiaomi.com>

### Checkpoints


kisvegabor:
It makes sense, thanks.

PR #3536, opened 2022-08-01
Add uint32_t lv_textarea_get_right_char(lv_obj_t * obj);

tryor:

Add a method:
/**
 * Get a the right character from the current cursor position
 * @param obj       pointer to a text area object
 * @return          a the right character or 0
 */
uint32_t lv_textarea_get_right_char(lv_obj_t * obj);

* This method is used in my embedded device, which uses LVGL 7.0 and 8.2

### Checkpoints

### DIFF
--- a/src/widgets/textarea/lv_textarea.c
+++ b/src/widgets/textarea/lv_textarea.c
@@ -208,18 +208,6 @@ void lv_textarea_add_text(lv_obj_t * obj, const char * txt)
     lv_event_send(obj, LV_EVENT_VALUE_CHANGED, NULL);
 }
 
-uint32_t lv_textarea_get_right_char(lv_obj_t * obj)
-{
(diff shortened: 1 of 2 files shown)


kisvegabor:
I see the value in this feature but the word `right` can be misleading when someone types Arabic text with Right to Left writing direction. Also, if the cursor has block style (not just a right border) an exact character is selected. So actually it's not "right" or "left" but rather "current", "selected", or the "character on cursor".

tryor:
I add this method, is to be able to enter alphabetic characters, on a keyboard layout like this.
keypad

tryor:
such an effect:
b70b789c7bca0c31185da9a5e7dc99e8 (1)

kisvegabor:
Looks good, thank you!

tryor:
Thank you for merging my PR. After that, all implementations of this feature, I will consider how to elegantly add it to the project.

PR #3540, opened 2022-08-02
fix(ime_pinyin): fix the problem that lv_ime_pinyin is not hidden together with the keyboard

100ask:
…ether with the keyboard

In fact, the use of lv_ime_pinyin does not need to use the lv_ime_pinyin object itself, so we decided to hide it to avoid some other unnecessary problems.

### Checkpoints


kisvegabor:
Thank you!

PR #3546, opened 2022-08-04
perf(label): add size cache for get self size event

FASTSHIFT:

Reduce the overhead of getting complex long text self size.

### Checkpoints

### DIFF
--- a/src/widgets/label/lv_label.c
+++ b/src/widgets/label/lv_label.c
@@ -763,24 +763,27 @@ static void lv_label_event(const lv_obj_class_t * class_p, lv_event_t * e)
         lv_label_refr_text(obj);
     }
     else if(code == LV_EVENT_GET_SELF_SIZE) {
-        lv_point_t size;
         lv_label_t * label = (lv_label_t *)obj;
-        const lv_font_t * font = lv_obj_get_style_text_font(obj, LV_PART_MAIN);
-        lv_coord_t letter_space = lv_obj_get_style_text_letter_space(obj, LV_PART_MAIN);
-        lv_coord_t line_space = lv_obj_get_style_text_line_space(obj, LV_PART_MAIN);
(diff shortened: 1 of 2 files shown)


FASTSHIFT:
Long text measurement is an overhead that cannot be ignored, especially `lv_draw_label()` re-measure text every time, we should consider adding more cache mechanisms to improve performance.

kisvegabor:
Good idea!

Can you mention a real life example when it was bottleneck and how this change improved the performance?

FASTSHIFT:
@kisvegabor 
For example:
```c
static void long_label_test()
{
    const char* txt = 
        "Lorem ipsum dolor sit amet, consectetur adipiscing elit, "
        "sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. "
        "Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat. "
        "Duis aute irure dolor in reprehenderit in voluptate velit esse cillum dolore eu fugiat nulla pariatur. "
        "Excepteur sint occaecat cupidatat non proident, sunt in culpa qui officia deserunt mollit anim id est laborum.";
    lv_obj_t* label = lv_label_create(lv_scr_act());
    lv_obj_set_width(label, LV_PCT(100));
    lv_obj_set_style_text_font(label, &lv_font_montserrat_20, 0);
    lv_label_set_long_mode(label, LV_LABEL_LONG_WRAP);
    lv_label_set_text_static(label, txt);
}
```

on `STM32F429-Discovery` board:
Before optimization: **17FPS**
After optimization: **38~40FPS**

After measuring `GET_SELF_SIZE` it takes about **3.3ms**. The scroll area is obtained during page scrolling, resulting in frequent calls to `GET_SELF_SIZE`. Adding cache can solve the problem.

kisvegabor:
Got it, thanks!

PR #3549, opened 2022-08-06
fix(hal_disp): incorrect deinit callbacks set

Jackhtk:
This patch correct some incorrectly set hal display callbacks.


kisvegabor:
Nice catch, thank you!

PR #3557, opened 2022-08-10
fix(draw): decrease the LOG level when the letter is not found

FASTSHIFT:

It is common to find letter not found, for example a message sent by a user may contain a word that is not commonly used (not in the font library). 
Lower the LOG level to reduce duplication of printing.

### Checkpoints


kisvegabor:
As LVGL draws □ when the character is not found this log is really less important.

PR #3565, opened 2022-08-11
 fix(lvgl.h): Add macros to reduce include path

supperthomas:
Signed-off-by: supperthomas <78900636@qq.com>

### Checkpoints

### DIFF
--- a/lvgl.h
+++ b/lvgl.h
@@ -76,13 +76,27 @@ extern "C" {
 #include "src/widgets/tileview/lv_tileview.h"
 #include "src/widgets/win/lv_win.h"
 
-#include "src/others/snapshot/lv_snapshot.h"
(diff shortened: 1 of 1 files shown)


embeddedt:
Just curious: have you measured a substantial improvement to build performance when the includes are wrapped by macros? It does make the file slightly harder to read.

supperthomas:
> Just curious: have you measured a substantial improvement to build performance when the includes are wrapped by macros? It does make the file slightly harder to read.

Developers usually work with macros in the lv_conf.h file. By adding macros, it is easy for customers to add header files for unnecessary and infrequently used components when porting to different platforms. Usually there are some infrequently used header files that do not need to be added to the path if the macro is not opened

PR #3576, opened 2022-08-16
SDL Fix: Transformation With Alpha

mariotaku:

When using SDL renderer, if an object has opacity, then program will hang in draw functions forever. This is because max_row_with_alpha was not set.

### Checkpoints


kisvegabor:
LGTM!
Why is it in draft?

mariotaku:
Maybe I was not sure how well it works when I created this PR. It turns out works fine. I'll mark this ready!

embeddedt:
Okay, merging now. Thank you!

PR #3586, opened 2022-08-17
feat(refr): support REAL FPS mode

shuai132:

1. no need volatile
2. calculate FPS should from last draw end to this draw end. **AND SHOULD NOT use begin to end of one draw .**

BTW, a lot of demo declare it's HIGH FPS, which was wrong in fact, because the FPS showing is not correct.

Details:
* before(wrong FPS):
800FPS is not possible! and which not equals/close to 30FPS below. And 30FPS on the bottom right is alway 30!!!

image

* after:
~28FPS on my laptop.

image

### Checkpoints

### DIFF
--- a/docs/get-started/quick-overview.md
+++ b/docs/get-started/quick-overview.md
@@ -9,7 +9,7 @@ You should read this first to get a general impression and read the detailed [Po
 Instead of porting LVGL to embedded hardware straight away, it's highly recommended to get started in a simulator first.
 
 LVGL is ported to many IDEs to be sure you will find your favorite one.
-Go to the [Simulators](/get-started/pc-simulator) section to get ready-to-use projects that can be run on your PC.
+Go to the [Simulators](/get-started/platforms/pc-simulator) section to get ready-to-use projects that can be run on your PC.
 This way you can save the time of porting for now and get some experience with LVGL immediately.
 
 ## Add LVGL into your project
@@ -251,7 +251,7 @@ The theme for your application is a compile time configuration set in `lv_conf.h
 ```
 
 ## Micropython
-Learn more about [Micropython](/get-started/micropython).
+Learn more about [Micropython](/get-started/bindings/micropython).
 ```python
 # Create a Button and a Label
 scr = lv.obj()
--- a/docs/porting/project.md
+++ b/docs/porting/project.md
@@ -25,7 +25,7 @@ include $(LVGL_DIR)/$(LVGL_DIR_NAME)/lvgl.mk
 For integration with CMake take a look this section of the [Documentation](/get-started/platforms/cmake).
 
 ### Other platforms and tools
-The [Get started](/get-started/index.html) section contains many platform specific descriptions e.g. for ESP32, Arduino, NXP, RT-Thread, NuttX, etc.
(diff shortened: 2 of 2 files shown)


GorgonMeducer:
I cannot agree with this idea:
> calculate FPS should from last draw end to this draw end. AND SHOULD NOT use begin to end of one draw .

The measurement's purpose is to evaluate the performance of LVGL but not the whole embedded system. When you start from "the last draw end to this draw end", the whole embedded system latency is involved, introducing a lot of external noise into the performance evaluation of LVGL. 

In fact, we even introduced a method to remove the latency introduced by the LCD driver (i.e. `disp_flush`) in the benchmark project. 

Image that someone reports a performance-related issue, we will kindly ask he/she to run the benchmark project and show us the result. If rather than concentrating on measuring the performance of LVGL but measuring the performance of the whole software system, we have to either 1) ask he/she to use a standard environment/platform (RTOS or bare-metal) and/or 2) unnecessarily ask for more information about the whole software environment. 

Obviously, this is not good for locating the issue or understanding the potential problem.

shuai132:
>  The measurement's purpose is to evaluate the performance of LVGL but not the whole embedded system. 

As you mentioned, there are indeed reasons to only measure FPS for LVGL.
But measuring the real FPS of the entire system is also very useful and necessary. Especially for the real project, it is not meaningful to just look the FPS of LVGL, but also to take into account the impact of other operations on the FPS (noise and LCD driver latency).

for example:
1. Compare the effects of various rendering implementations on FPS
2. Overall FPS performance of the software
3. Want to measure the overall maximum frame rate (including rendering and other logic)

Since both measurements are meaningful, it is recommended to add a configuration item for the user to choose, such as LV_USE_PERF_MONITOR_SYSTEM_FPS in the above submission (of course the definition in config.h is missing)

Then configure LV_DEF_REFR_PERIOD=1 to get a general understanding of system performance.


Finally, I can continue to refine it if needed, otherwise I can close this MR after a further discussion.

GorgonMeducer:
Let's see what @kisvegabor will say.

embeddedt:
I'm not @kisvegabor, but I agree it would be worthwhile to support both options.

shuai132:
Any suggestion？@kisvegabor

kisvegabor:
[Here](https://github.com/lvgl/lvgl/pull/3416#issuecomment-1187571475) we were talking about introducing FAST and VISUAL modes. It makes sense to have a REAL (or so) mode too. 

If someone is interested in it, a PR would be highly appreciated to add these modes.

shuai132:
Yes, the `Real` mode are not only useful for benchmarking, but also important for user products.
This MR is all we need to do! It works both benchmark and user app. It's ready to merge.

Just define `LV_USE_PERF_MONITOR_SYSTEM_FPS` (or `LV_USE_PERF_MONITOR_REAL_FPS` is better?)

I will learn add it to kconfig/v_conf_template.h and commit it later. it will be here:
image

any suggestions? @kisvegabor

shuai132:
By the way, I think `Real` FPS should be enabled by default on LV_USE_MEM_MONITOR.

Because, what most people cared about is the real FPS. But now, it will alway be 33FPS by default, even the code is:
```
while(1) {
  lv_timer_handler();
  my_delay_ms(100);
}
```
Obviously, the real FPS will be ~10.

shuai132:
I open a new PR instead of this one, please review when you have time.
@kisvegabor

Issue #3585, opened 2022-08-17
Code format hook


kisvegabor:
Something happened with pre-commit format hook because somehow the excluded files are note excluded.

I've tried to update `lv_conf_internal.h` by running `lv_conf_internal_gen.py` but on commit `lv_conf_internal.h` is formatted although it's excluded in `code-format.cfg`.

It worked so far and I have no idea what happened. I couldn't find anything related in git history.

@C47D could you take a look?

C47D:
Sure, I will take a look later today.

C47D:
I did a manual change in lv_conf_internal.h and tried to run astyle manually as pre-commit runs it

```commandline
astyle --options=scripts/code-format.cfg --ignore-exclude-errors src/lv_conf_internal.h
```

This command formats `lv_conf_internal.h`.

Then I ran astyle standalone (no pre-commit)

```commandline
astyle --exclude=src/lv_conf_internal.h -v -i --formatted src/lv_conf_internal.h
```

And got the same behavior, `lv_conf_internal.h` ends up being formatted. So I think the exclude option is being "ignored" when we tell astyle to format the excluded file.

I think for a momentary fix we could use `/* *INDENT-OFF* */` at the beginning of the file and `/* *INDENT-ON* */` at the end of the file to prevent astyle formatting it

C47D:
for the whitespace issue reported in #3587 i think we can include the following patch

```diff
diff --git a/.pre-commit-config.yaml b/.pre-commit-config.yaml
index 7c9b2bfd8..5a8353893 100644
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -4,7 +4,10 @@ repos:
 -   repo: https://github.com/pre-commit/pre-commit-hooks
     rev: v3.2.0
     hooks:
-    -   id: trailing-whitespace
+    - id: trailing-whitespace
+      stages: [ commit ]
+      exclude: 'src/lv_conf_internal.h'
 
 -   repo: local
```

The issue is because the trailing-whitespace hook runs on all the files, with the patch it will run on the pre-commit hook and will exclude the `src/lv_conf_internal.h` file.

You can test this patch by applying it, staging `.pre-commit-config.yaml`, then adding whitespace to the `src/lv_conf_internal.h` file and run the command
```commandline
pre-commit run trailing-whitespace
```
You will see `src/lv_conf_internal.h` file being ignored by pre-commit.

embeddedt:
[...]

PR #3587, opened 2022-08-17
fix(lv_arc): add get_rotation getter for lv_arc

tvanfossen-bissell:
Adds a getter function for lv_arcs rotation property - I found this to be useful when animating an arc rotation from an unknown current rotation to a new rotation

Adds a missing getter function for lv_arc rotation property

### Checkpoints
- [x ] Follow the [styling guide](https://github.com/lvgl/lvgl/blob/master/docs/CODING_STYLE.md)
- [x ] Run `code-format.py` from the `scripts` folder. [astyle](http://astyle.sourceforge.net/install.html) needs to be installed.
- [x ] Update the documentation

### DIFF
--- a/src/widgets/arc/lv_arc.h
+++ b/src/widgets/arc/lv_arc.h
@@ -223,6 +223,13 @@ int16_t lv_arc_get_max_value(const lv_obj_t * obj);
  */
 lv_arc_mode_t lv_arc_get_mode(const lv_obj_t * obj);
 
+/**
+ * Get whether the arc is type or not.
+ * @param arc       pointer to an arc object
+ * @return          arc's current rotation
+ */
+int16_t lv_arc_get_rotation(const lv_obj_t * obj)
+
 /*=====================
  * Other functions
  *====================*/
--- a/src/widgets/arc/lv_arc.h
+++ b/src/widgets/arc/lv_arc.h
@@ -224,11 +224,11 @@ int16_t lv_arc_get_max_value(const lv_obj_t * obj);
 lv_arc_mode_t lv_arc_get_mode(const lv_obj_t * obj);
 
 /**
- * Get whether the arc is type or not.
+ * Get the rotation for the whole arc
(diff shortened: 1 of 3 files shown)


src/widgets/arc/lv_arc.h
```diff
@@ -223,6 +223,13 @@ int16_t lv_arc_get_max_value(const lv_obj_t * obj);
  */
 lv_arc_mode_t lv_arc_get_mode(const lv_obj_t * obj);
 
+/**
+ * Get whether the arc is type or not.
```
embeddedt:
```suggestion
 * Get the rotation for the whole arc
```

src/widgets/arc/lv_arc.h
```diff
@@ -223,6 +223,13 @@ int16_t lv_arc_get_max_value(const lv_obj_t * obj);
...
+ * @param arc       pointer to an arc object
+ * @return          arc's current rotation
+ */
+int16_t lv_arc_get_rotation(const lv_obj_t * obj)
```
embeddedt: Looks like there's a missing semicolon here. I think this is the reason for the CI failure.

tvanfossen-bissell: Updated

src/widgets/arc/lv_arc.h
tvanfossen-bissell: Updated

tvanfossen-bissell:
Running lv_conf_internal_gen.py makes a slight whitespace change, pre-commit cancels out the lv_conf_internal_gen.py change, I'm assuming this is not something that should be changed in the context of this PR @embeddedt 

image
Above change wiped out with pre-commit

embeddedt:
Yes, there's an [unrelated issue with pre-commit](https://github.com/lvgl/lvgl/issues/3585) at the moment, don't worry about that.

embeddedt:
Thanks!

PR #3588, opened 2022-08-17
feat(lv_rlottie): Lv rlottie utility functions

tvanfossen-bissell:

Adding a set of lv_rlottie utility functions:
get/set framerate
get/set render height/width
get/set dest_frame
get/set start_frame
get playmode
get current frame

Updating lv_rlottie documentation for the added utilities
Updating lv_rlottie documentation to provide an example for compilation of rlottie lib for ESP-IDF (been sitting on this way too long) - this portion of the documentation may be better provided as a patch file for end users to optionally apply?

### Checkpoints
- [x ] Follow the [styling guide](https://github.com/lvgl/lvgl/blob/master/docs/CODING_STYLE.md)
- [ x] Run `code-format.py` from the `scripts` folder. [astyle](http://astyle.sourceforge.net/install.html) needs to be installed.
- [ x] Update the documentation

### DIFF
--- a/docs/libs/rlottie.md
+++ b/docs/libs/rlottie.md
@@ -24,6 +24,7 @@ And finally add the `-lrlottie` flag to your linker.
 
 On embedded systems you need to take care of integrating Rlottie to the given build system.
 
+ESP-IDF example at bottom
 
 ## Usage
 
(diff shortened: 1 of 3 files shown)


kisvegabor:
https://github.com/lvgl/lvgl/pull/3294 will add a lot of related updates and adds generic animation and vector graphics control to `lv_img`.
I hope we can merge it soon.

@X-Ryl669 probable can comment more about this topic. 


---

However, the ESP build instructions would be very great to have. Could you send a separate pull request with them?

tvanfossen-bissell:
Will do - thinking it might serve better to build a .patch file of the rlottie changes and include a copy of the CMakeLists used at the esp component level

kisvegabor:
> Will do - thinking it might serve better to build a .patch file of the rlottie changes and include a copy of the CMakeLists used at the esp component level

I agree, adding a patch as a text to the docs would be better. Power users can easily create am actual patch from it, but other can also understand it.

PR #3589, opened 2022-08-17
feat(lv_arc): add knob offset utility

tvanfossen-bissell:
Edge case usage for use as decorative arc widget. Allows for the knob to be offset from the arc

This change allows for the knob to be offset from the end of the arc. For most cases, it probably will remain unchanged.

Below is what this change enables:

image

For frame of reference, this is an lv_arc object with the below properties:

rotation: 270
start_angle: 10
end_angle: 350
bg_start_angle: 10
bg_end_angle: 350

LV_PART_KNOB visible
knob_offset: -10
knob_padding: 2

Originally, I wanted to do this with an image, but rotating the image was not smooth so ended up going down a path to enable rotate the arc as a whole, with the rotation of the arc used to indicate value in place of using the actual value - it is not as performant as using lv_arc_value and filling the arc directly, but it does accomplish the desired look and feel of using the knob as an indicator

To further show usage, here is an example animation using the knob as the indicator for the arc:

Clarification: Design intent for this image was to have an circle widget, shown with a dot to indicate a given 0-100 value, with padding between the dot(knob) and the beginning/end of the circle

```
static int convert_value_to_degrees(int value)
{
    float percent = (float)value / 100.0;
    int degrees = (percent * 360);

    return degrees;
}

static void anim_set_arc_rotate_0_to_100_cb(void * var, int32_t v)
{

    int degrees = convert_value_to_degrees(v);
    int start_degree = ARC_START_ROTATION; //this is a define added for my purposes to indicate the initial rotation (270) of the arc
    int actual_rotation = (start_degree + degrees) % 360;
    
    LV_LOG_INFO("anim_set_arc_rotate_0_to_100_cb: act_rot %d\n", actual_rotation);

    lv_arc_set_rotation(var, actual_rotation);
};

lv_anim_t set_arc_rotate_0_to_100(lv_obj_t* obj, int start_rot, int target_rot, uint32_t anim_speed, uint32_t anim_delay, lv_anim_path_cb_t anim_path, uint16_t playcount)
{
    lv_anim_t a;
    printf("start rot for soc rotation: %d\n", start_rot);
    lv_anim_init(&a);
    lv_anim_set_var(&a, obj);
    lv_anim_set_exec_cb(&a, (lv_anim_exec_xcb_t)anim_set_arc_rotate_0_to_100_cb);
    lv_anim_set_path_cb(&a, (lv_anim_path_cb_t)anim_path);
    lv_anim_set_delay(&a, anim_delay);
    lv_anim_set_time(&a, anim_speed);
    lv_anim_set_values(&a, start_rot, target_rot);

    return a;
}

```

### Checkpoints
- [ x] Follow the [styling guide](https://github.com/lvgl/lvgl/blob/master/docs/CODING_STYLE.md)
- [ x] Run `code-format.py` from the `scripts` folder. [astyle](http://astyle.sourceforge.net/install.html) needs to be installed.
- [ x] Update the documentation

### DIFF
--- a/docs/widgets/arc.md
+++ b/docs/widgets/arc.md
@@ -43,7 +43,7 @@ The change rate is defined in degree/second unit and can be set with `lv_arc_set
 
 ### Knob offset
 Changing the knob offset allows the location of the knob to be moved relative to the end of the arc
-The knob offset can be set by `lv_arc_set_knob_offset(arc, offset)`, will only be visible if LV_PART_KNOB is visible
+The knob offset can be set by `lv_arc_set_knob_offset(arc, offset_angle)`, will only be visible if LV_PART_KNOB is visible
 
 
 ### Setting the indicator manually
--- a/docs/widgets/arc.md
+++ b/docs/widgets/arc.md
@@ -41,6 +41,10 @@ The mode can be set by `lv_arc_set_mode(arc, LV_ARC_MODE_...)` and used only if 
 If the arc is pressed the current value will set with a limited speed according to the set *change rate*.
 The change rate is defined in degree/second unit and can be set with `lv_arc_set_change_rage(arc, rate)`
 
+### Knob offset
+Changing the knob offset allows the location of the knob to be moved relative to the end of the arc
+The knob offset can be set by `lv_arc_set_knob_offset(arc, offset)`, will only be visible if LV_PART_KNOB is visible
+
 
 ### Setting the indicator manually
 It's also possible to set the angles of the indicator arc directly with `lv_arc_set_angles(arc, start_angle, end_angle)` function or `lv_arc_set_start/end_angle(arc, start_angle)`.
--- a/src/widgets/arc/lv_arc.c
+++ b/src/widgets/arc/lv_arc.c
@@ -799,7 +799,7 @@ static void get_knob_area(lv_obj_t * obj, const lv_point_t * center, lv_coord_t 
     r -= indic_width_half;
 
     lv_coord_t angle = get_angle(obj);
-    lv_coord_t knob_offset lv_arc_get_knob_offset(obj);
+    lv_coord_t knob_offset = lv_arc_get_knob_offset(obj);
     lv_coord_t knob_x = (r * lv_trigo_sin(knob_offset + angle + 90)) >> LV_TRIGO_SHIFT;
     lv_coord_t knob_y = (r * lv_trigo_sin(knob_offset + angle)) >> LV_TRIGO_SHIFT;
 
--- a/src/widgets/arc/lv_arc.c
+++ b/src/widgets/arc/lv_arc.c
@@ -266,6 +266,14 @@ void lv_arc_set_change_rate(lv_obj_t * obj, uint16_t rate)
     arc->chg_rate = rate;
 }
 
+void lv_arc_set_knob_offset(lv_obj_t * obj, int16_t offset)
+{
+    LV_ASSERT_OBJ(obj, MY_CLASS);
+    lv_arc_t * arc = (lv_arc_t *)obj;
+
+    arc->knob_offset = offset;
+}
+
 /*=====================
  * Getter functions
  *====================*/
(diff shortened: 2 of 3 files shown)


docs/widgets/arc.md
```diff
@@ -41,6 +41,10 @@ The mode can be set by `lv_arc_set_mode(arc, LV_ARC_MODE_...)` and used only if
 If the arc is pressed the current value will set with a limited speed according to the set *change rate*.
 The change rate is defined in degree/second unit and can be set with `lv_arc_set_change_rage(arc, rate)`
 
+### Knob offset
+Changing the knob offset allows the location of the knob to be moved relative to the end of the arc
+The knob offset can be set by `lv_arc_set_knob_offset(arc, offset)`, will only be visible if LV_PART_KNOB is visible
```
kisvegabor:
```suggestion
The knob offset can be set by `lv_arc_set_knob_offset(arc, offset_angle)`, will only be visible if LV_PART_KNOB is visible
```

kisvegabor (commented):
Interesting idea. But can't you realize it like
```c
    lv_obj_t * arc = lv_arc_create(lv_scr_act());
    lv_obj_set_size(arc, 150, 150);
    lv_arc_set_bg_angles(arc, 0, 300);
    lv_arc_set_angles(arc, 330, 330);
    lv_obj_center(arc);
```

image

tvanfossen-bissell:
> Interesting idea. But can't you realize it like
> 
> ```c
>     lv_obj_t * arc = lv_arc_create(lv_scr_act());
>     lv_obj_set_size(arc, 150, 150);
>     lv_arc_set_bg_angles(arc, 0, 300);
>     lv_arc_set_angles(arc, 330, 330);
>     lv_obj_center(arc);
> ```
> 
> image

Definitely right, didn't think about doing it this way - in the case of the knob offset, if the arc value is set, the knob would still move with it - undesirable for this particular use-case - in your example, the knob would be locked in place, and could still utilize the same example animation above, just trying to think of any edge cases where there might be value to having the knob offset rather than doing it as above

kisvegabor:
> just trying to think of any edge cases where there might be value to having the knob offset rather than doing it as above

I was thinking about animations like
- animate up to current value
- shake the current value

without changing `value`. However offset is in degree so it's hard to tell the required degree to get to the current value's angle. 

There could be an `offset_value` property too, but it seems too bloated for a very special use case.

tvanfossen-bissell:
Aah I hadn't thought of the shake idea - could be an interesting interaction for trying to display arc state is locked when a user trys to drag it, for whatever underlying app reasoning

Knob offset angle of 0 would always return it to the current values angle wouldnt it? Offset is only applied in addition to the angle already applied in determining knob angle in get_knob_area, so its always offset some X deg from arc value end?

Or was the thinking of doing an lv_arc_value animation while keeping knob position at the same angle between start/end of the indicator/bg arcs? Feel that use case would want to be handled inside the animation callback itself where knob offset and arc value are changed inside of the same animation callback? User data field could be used to hold the initial value of the knob offset pre-animation

something to the effect of in anim exec cb:

```c
int32_t start_knob_angle = lv_obj_get_user_data(obj);
lv_arc_set_value(obj, v);
int32_t new_arc_angle = get_angle(obj); // as used in determining knob angle before this change
int32_t new_knob_angle = -1*((new_arc_angle + start_knob_angle)%360); 
lv_arc_set_knob_offset(new_knob_angle);
```

For the knob to remain in the same position it should always be the negative of the current arc indicator angle? Assuming knob started at 0 degrees, unless it started at something other than 0 deg.

That said, this route could result in the animation getting shakey unless knob offset is adjusted ahead of lv_arc_set_value (accounting instead for the change in V from current angle of the arc)

kisvegabor:
> Knob offset angle of 0 would always return it to the current values angle wouldnt it? Offset is only applied in addition to the angle already applied in determining knob angle in get_knob_area, so its always offset some X deg from arc value end?

I agree.

The use case I imagined is that the value is set to 30 (min = 0, max=100), and we'd like to animate the knob up to 30 from 0 without changing the value. The offset animation should star from -X and end at 0. The what is  `X`? we need to know the current angle. Or shall we read it from the arc? As it's stored in `arc->end_angle`.

tvanfossen-bissell:
the get_angle function with the arc object passed in should provide -X (or like you said simple grabbing arc->end_angle) as the starting point for the animation.

If the arc_value has been set to 30, and the knob offset has not been set then the knob would appear at arc angle end - could make the knob hidden, set the arc_value, get the end angle, then apply the offset and clear the hidden flag for the knob, then animate the offste from -X obtained from end angle, to 0

kisvegabor:
Okay, I'm convinced. :slightly_smiling_face: 
With `get_end_angle` it's really easy to use and can be useful.

Merging, thank you!

PR #3599, opened 2022-08-19
fix(disp):automatically delete the old screen when not using animation

hellototoro:

When I use lv_scr_load_anim(obj, LV_SCR_LOAD_ANIM_NONE, 0, 0, true) to load a new screen, an old screen doesn't be deleted automatically.
### Checkpoints

### DIFF
--- a/src/core/lv_disp.c
+++ b/src/core/lv_disp.c
@@ -274,6 +274,7 @@ void lv_scr_load_anim(lv_obj_t * new_scr, lv_scr_load_anim_t anim_type, uint32_t
     /*Shortcut for immediate load*/
     if(time == 0 && delay == 0) {
         scr_load_internal(new_scr);
+        if(auto_del) lv_obj_del(act_scr);
         return;
     }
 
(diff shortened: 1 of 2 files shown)


embeddedt:
I am pretty sure the screen not being deleted is intentional. The non-animating version (`lv_scr_load`) also only changed the screen, and did not delete it.

hellototoro:
I don't understand why did author do that intentionally. I think loading a new screen without animation and meanwhile deleting the old screen is also normal.

kisvegabor:
I agree that regardless with `delay=0` and `time=0`, `auto_del=true` still should be considered.

As immediate load is special case, we can simple add: `if(auto_del) lv_obj_del(act_scr);`
here
https://github.com/lvgl/lvgl/blob/bdd520e00c86b8e350876e4434e8345615340121/src/core/lv_disp.c#L277

What do you think?

hellototoro:
You are right! I noticed that without setting the prev_scr's value when the immediate loading.

kisvegabor:
Looks good, thank you!

PR #3600, opened 2022-08-20
feat: support REAL FPS

shuai132:
See: https://github.com/lvgl/lvgl/pull/3586

### Checkpoints

### DIFF
--- a/src/core/lv_refr.c
+++ b/src/core/lv_refr.c
@@ -288,7 +288,6 @@ void _lv_disp_refr_timer(lv_timer_t * tmr)
     REFR_TRACE("begin");
 
     uint32_t start = lv_tick_get();
-    volatile uint32_t elaps = 0;
 
     if(tmr) {
         disp_refr = tmr->user_data;
@@ -303,6 +302,9 @@ void _lv_disp_refr_timer(lv_timer_t * tmr)
     else {
         disp_refr = lv_disp_get_default();
     }
+
+    uint32_t elaps = lv_tick_elaps(disp_refr->last_render_start_time);
+    disp_refr->last_render_start_time = start;
 
     /*Refresh the screen's layout if required*/
     lv_obj_update_layout(disp_refr->act_scr);
@@ -342,8 +344,6 @@ void _lv_disp_refr_timer(lv_timer_t * tmr)
         lv_memzero(disp_refr->inv_areas, sizeof(disp_refr->inv_areas));
         lv_memzero(disp_refr->inv_area_joined, sizeof(disp_refr->inv_area_joined));
         disp_refr->inv_p = 0;
-
-        elaps = lv_tick_elaps(start);
 
         /*Call monitor cb if present*/
         if(disp_refr->driver->monitor_cb) {
--- a/src/core/lv_refr.c
+++ b/src/core/lv_refr.c
@@ -306,57 +306,6 @@ void _lv_disp_refr_timer(lv_timer_t * tmr)
     uint32_t elaps = lv_tick_elaps(disp_refr->last_render_start_time);
     disp_refr->last_render_start_time = start;
 
-    /*Refresh the screen's layout if required*/
-    lv_obj_update_layout(disp_refr->act_scr);
-    if(disp_refr->prev_scr) lv_obj_update_layout(disp_refr->prev_scr);
-
-    lv_obj_update_layout(disp_refr->top_layer);
-    lv_obj_update_layout(disp_refr->sys_layer);
-
-    /*Do nothing if there is no active screen*/
-    if(disp_refr->act_scr == NULL) {
-        disp_refr->inv_p = 0;
-        LV_LOG_WARN("there is no active screen");
-        REFR_TRACE("finished");
-        return;
-    }
-
-    if(disp_refr->driver->direct_mode && disp_refr->driver->draw_ctx->color_format != LV_COLOR_FORMAT_NATIVE) {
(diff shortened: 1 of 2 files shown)


kisvegabor:
Hi,

I've also created a [PR](https://github.com/lvgl/lvgl/pull/3625) with all 3 benchmark modes. Let's address the benchmark first to learn from it, and get back to it after that.

shuai132:
Hi,

I reviewed the code and tested it. These are all about benchmarks. Therefore, I think that after supporting this, it is best to only display the real FPS in the bottom right corner.
What's your opinion?
image

kisvegabor:
I agree with displaying only the real FPS in the corner but I'd like to discuss the benchmark PR first to see if something comes up that should be considered here too.

An other option is to display something like `20 (67) FPS` or make it configurable as did in this PR.

shuai132:
The benchmark also adds a lot of auxiliary code, as if nothing like this is needed here (at least I didn't notice it).

Only showing the real one should be more in line with user needs, and it is not easy to be confused. And most UI frameworks are like that.

Other methods are also fine, in fact, as long as there is a way for users to see the real FPS displayed.

kisvegabor:
In the meantime I merged the benchmark demo updates. 

I think it's not required to use a config option for this and we can keep only the "real" version. Could you update the PR?

shuai132:
OK, I will update PR now.

shuai132:
@kisvegabor 
Done, this PR is ready to merge on CI pass.
ps: some CI error, because network, can ignore it.

kisvegabor:
I've looked into the code again and something doesn't adds up in my mind. 

So, what we are trying to measure? The time between 2 `_lv_disp_refr_timer` calls, right? 

What do we want to see if there was nothing to refresh? If we update the FPS indicator, probably there will be something to refresh.

What do you think?

shuai132:
Yes, I think that test the time between 2 _lv_disp_refr_timer calls is all we need to do.

because one frame render begin on timer `_lv_disp_refr_timer` begin, and end on disp_refr->driver->monitor_cb (it call `draw_buf_flush` and all flush if need).


Maybe I know not much about lvgl, mybe I made mistake on the PR. But we just need test one frame begin to end.
So, any better place(to test one frame begin to end)?

shuai132:
PS:
If there was nothing to refresh, just record the time point.
And `last_refresh_time` may be useful for other things. e.g. calculate real FPS.

And, do you have a better way?

@kisvegabor

kisvegabor:
I agree but why 
```c
        elaps = lv_tick_elaps(disp_refr->last_refresh_time);
        /*Record refresh time immediately*/
        disp_refr->last_refresh_time = lv_tick_get();
```
isn't added at the very beginning of `_lv_disp_refr_timer`? It seems to me the whole `#if LV_USE_PERF_MONITOR` and `#if LV_USE_MEM_MONITOR` section from the end of the timer can be moved to the beginning too. 

What  do you think?

shuai132:
1. `last_refresh_time` means last refreshed finish time. should be added at the very end(because `lv_tick_elaps` need). should change name to `last_refreshed_time`?
2. You are right, the whole #if LV_USE_PERF_MONITOR and #if LV_USE_MEM_MONITOR should be moved to the beginning too.

kisvegabor:
> last_refresh_time means last refreshed finish time. should be added at the very end(because lv_tick_elaps need). should change name to last_refreshed_time?

Why can't we measure the time between the elapsed time between the beginning of the calls?

shuai132:
Both begin and end are fine. but begin should include sleep time, so it just equals to last end time.

So, because it add the time of sleep, I think the name may not better.

kisvegabor:
We can find any name, but I think this order is cleanest:
1. Get elapsed time since last call. E.g. `disp->last_render_start_time`
2. Update the CPU and memory indicators
3. Do the rendering

shuai132:
Fine, we can use this flow.

And I update PR now?

kisvegabor:
> And I update PR now?

It'd be great, yes.

shuai132:
I have update PR.
Ready to review.

@kisvegabor

src/core/lv_refr.c
```diff
@@ -288,7 +288,6 @@ void _lv_disp_refr_timer(lv_timer_t * tmr)
     REFR_TRACE("begin");
 
     uint32_t start = lv_tick_get();
```
kisvegabor, line 290: Do we need this variable?

src/core/lv_refr.c
```diff
@@ -304,6 +303,9 @@ void _lv_disp_refr_timer(lv_timer_t * tmr)
         disp_refr = lv_disp_get_default();
     }
 
+    uint32_t elaps = lv_tick_elaps(disp_refr->last_render_start_time);
+    disp_refr->last_render_start_time = start;
+
```
kisvegabor, line 308: Please also bring PERF and MEM monitor handing here (before rendering).

src/core/lv_refr.c
shuai132: It just want a definite time point.
and record lv_tick_elaps cost. (YES it will be very small value).

remove or keep it? I think both are fine.

src/core/lv_refr.c
shuai132: Done, I push a new commit.


(before, I think show last fps is no big wrong, ofcause move to front is right!)

kisvegabor:
Looks good, thank you!

PR #3601, opened 2022-08-20
feat(file_explorer): add API to use file explorer

YobieZhou:

### Checkpoints


src/others/file_explorer/lv_file_explorer.h
```diff
@@ -0,0 +1,190 @@
...
+ * @param dir   the dir from 'lv_file_explorer_dir_t' enum.
+ */
+void lv_file_explorer_set_quick_access_path(lv_obj_t * obj, lv_file_explorer_dir_t dir, char * path);
+
+/**
+ * Set file_explorer quick access state
+ * @param obj   pointer to a label object
+ * @param state true:display, false: hide
+ */
+void lv_file_explorer_set_quick_access_state(lv_obj_t * obj, bool state);
```
kisvegabor: What about having a `get_quick_access_area` function, and the user can hide, style, etc the returned object?

src/others/file_explorer/lv_file_explorer.h
```diff
@@ -0,0 +1,190 @@
...
+/*=====================
+ * Getter functions
+ *====================*/
+
+/**
+ * Get file explorer Selected file
+ * @param obj   pointer to a file explorer object
+ * @return      pointer to the file explorer selected file
+ */
+const char * lv_file_explorer_get_sel_fn(lv_obj_t * obj);
```
kisvegabor:
```suggestion
const char * lv_file_explorer_get_selected_fn(lv_obj_t * obj);
```

src/others/file_explorer/lv_file_explorer.h
```diff
@@ -0,0 +1,190 @@
...
+
+/**
+ * Get file explorer cur path
+ * @param obj   pointer to a file explorer object
+ * @return      pointer to the file explorer cur path
+ */
+char * lv_file_explorer_get_cur_path(lv_obj_t * obj);
```
kisvegabor:
```suggestion
char * lv_file_explorer_get_current_path(lv_obj_t * obj);
```

src/others/file_explorer/lv_file_explorer.h
```diff
@@ -0,0 +1,190 @@
...
+
+/**
+ * Get file explorer head area obj
+ * @param obj   pointer to a file explorer object
+ * @return      pointer to the file explorer head area obj
+ */
+lv_obj_t * lv_file_explorer_get_head(lv_obj_t * obj);
```
kisvegabor:
```suggestion
lv_obj_t * lv_file_explorer_get_header(lv_obj_t * obj);
```

src/others/file_explorer/lv_file_explorer.h
```diff
@@ -0,0 +1,190 @@
...
+
+/**
+ * Get file explorer path obj(label)
+ * @param obj   pointer to a file explorer object
+ * @return      pointer to the file explorer path obj(label)
+ */
+lv_obj_t * lv_file_explorer_get_path_obj(lv_obj_t * obj);
```
kisvegabor:
```suggestion
lv_obj_t * lv_file_explorer_get_path_label(lv_obj_t * obj);
```

src/others/file_explorer/lv_file_explorer.h
```diff
@@ -0,0 +1,190 @@
...
+/*=====================
+ * Other functions
+ *====================*/
+
+/**
+ * Open a specified path
+ * @param obj   pointer to a file explorer object
+ * @param dir   pointer to the path
+ */
+void lv_file_explorer_open_dir(lv_obj_t * obj, char * dir);
```
kisvegabor:
```suggestion
void lv_file_explorer_open_dir(lv_obj_t * obj, const char * dir);
```

src/others/file_explorer/lv_file_explorer.c
```diff
@@ -0,0 +1,782 @@
...
+            break;
+        case LV_EXPLORER_DOCS_DIR:
+            dir_str = &(explorer->docs_dir);
+            break;
+        //case LV_EXPLORER_MNT_DIR:
+        //    dir_str = &(explorer->home_dir);
+        //    break;
+        //case LV_EXPLORER_FS_DIR:
+        //    dir_str = &(explorer->home_dir);
+        //    break;
```
kisvegabor: Are these still required?

src/others/file_explorer/lv_file_explorer.c
```diff
@@ -0,0 +1,782 @@
...
+    /*Allocate space for the new text*/
+    *dir_str = lv_malloc(len);
+    LV_ASSERT_MALLOC(*dir_str);
+    if(*dir_str == NULL) return;
+
+    strcpy(*dir_str, path);
+}
+
+
+void lv_file_explorer_set_quick_access_state(lv_obj_t * obj, bool state)
```
kisvegabor: Make it `static`?

src/others/file_explorer/lv_file_explorer.c
```diff
@@ -0,0 +1,782 @@
...
+/*=====================
+ * Getter functions
+ *====================*/
+const char * lv_file_explorer_get_sel_fn(lv_obj_t * obj)
+{
+    LV_ASSERT_OBJ(obj, MY_CLASS);
+
+    lv_file_explorer_t * explorer = (lv_file_explorer_t *)obj;
+
+    return explorer->sel_fp;
```
kisvegabor: The function name is `sel_fn` but `explorer->sel_fp`. For consistency we should pick either `fn` or `fp`. It's rather a path, not only the filename, right?

src/others/file_explorer/lv_file_explorer.c
```diff
@@ -0,0 +1,782 @@
...
+
+    lv_obj_set_size(obj, LV_PCT(100), LV_PCT(100));
+    lv_obj_set_flex_flow(obj, LV_FLEX_FLOW_COLUMN);
+
+    explorer->cont = lv_obj_create(obj);
+    lv_obj_set_width(explorer->cont, LV_PCT(100));
+    lv_obj_set_flex_grow(explorer->cont, 1);
+
+#if LV_FILE_EXPLORER_QUICK_ACCESS
+    // Quick access bar area on the left
```
kisvegabor: Please use `/* */` style comments

src/others/file_explorer/lv_file_explorer.c
```diff
@@ -0,0 +1,782 @@
...
+    explorer->file_table = lv_table_create(explorer->browser_area);
+    lv_obj_set_size(explorer->file_table, LV_PCT(100), LV_PCT(86));
+    lv_table_set_col_width(explorer->file_table, 0, LV_PCT(100));
+    lv_table_set_col_cnt(explorer->file_table, 1);
+    lv_obj_add_event_cb(explorer->file_table, browser_file_event_handler, LV_EVENT_ALL, obj);
+    // only scroll up and down
+    lv_obj_set_scroll_dir(explorer->file_table, LV_DIR_TOP | LV_DIR_BOTTOM);
+
+    // Initialize style
+    init_style(obj);
```
kisvegabor, line 338: If `lv_style_init` is called inside of it on every `lv_file_explorer_create`there will be a memory leak as the styles are not reset (freed), just always initialized. 

I suggest having a  `lv_file_explorer_init()` function, called from `lv_init()` and initialize the styles there and here only add the styles to the objects.

src/others/file_explorer/lv_file_explorer.c
```diff
@@ -0,0 +1,782 @@
...
+    lv_file_explorer_t * explorer = (lv_file_explorer_t *)obj;
+
+    if(code == LV_EVENT_VALUE_CHANGED) {
+        //struct stat stat_buf;
+        char file_name[LV_FILE_EXPLORER_PATH_MAX_LEN];
+        const char * str_fn = NULL;
+        uint16_t row;
+        uint16_t col;
+
+        memset(file_name, 0, sizeof(file_name));
```
kisvegabor:
```suggestion
        lv_memzero(file_name, sizeof(file_name));
```

kisvegabor (commented):
Hi,

Great work again!

I have a few minor comments.

src/others/file_explorer/lv_file_explorer.c
YobieZhou: I'll do that

src/others/file_explorer/lv_file_explorer.c
YobieZhou: These are still needed. I will open them at the next commit

src/others/file_explorer/lv_file_explorer.c
YobieZhou: This is a good suggestion, but I think that in a project, generally only one `file_explorer` needs to be used. How about using **[local styles](https://docs.lvgl.io/master/overview/style.html#local-styles)**?

src/others/file_explorer/lv_file_explorer.c
YobieZhou: The file name returned here is the file name selected in the table. It should be changed to **return explorer->sel_fn;** .

The full path of the file is obtained by calling the `lv_file_explorer_get_current_path` function.

src/others/file_explorer/lv_file_explorer.h
YobieZhou: That's a good suggestion. I've done it

src/others/file_explorer/lv_file_explorer.c
YobieZhou: I have deleted it now, and let the user realize the control button (the sorting function is the same). The example has also been modified. What do you think?

src/others/file_explorer/lv_file_explorer.c
kisvegabor: It's good idea to use local styles. This way if a file explorer is closed, it's styles will be freed too.

kisvegabor (commented):
Looks good to me. Only the style related question remained.

src/others/file_explorer/lv_file_explorer.c
YobieZhou: I keep `quick_access_list_btn_style`, which is to prepare for future user expansion.

YobieZhou:
CI checks still has problems. What does this mean?

lv_conf_template.h
```diff
@@ -695,6 +695,17 @@
     #endif // LV_IME_PINYIN_USE_K9_MODE
 #endif
 
+/*1: Enable Pinyin input method*/
```
amirgon: Why does it require to " Enable Pinyin input method"?

lv_conf_template.h
```diff
@@ -695,6 +695,17 @@
...
+/*Requires: lv_table*/
```
amirgon, line 699: If it requires `lv_table`, shouldn't it be conditioned (`#ifdef`) in `LV_USE_TABLE`?

examples/others/file_explorer/index.rst
```diff
@@ -0,0 +1,13 @@
+
```
amirgon, line 1: Missing Micropython examples?

src/others/file_explorer/lv_file_explorer.h
```diff
@@ -0,0 +1,184 @@
...
+ *********************/
+
+/**********************
+ *      TYPEDEFS
+ **********************/
+
+typedef enum {
+    LV_EXPLORER_SORT_NONE,
+    LV_EXPLORER_SORT_KIND,
+} lv_file_explorer_sort_t;
```
amirgon, line 31: Did you consider using a callback function for sort?
This would allow the users to apply any custom sort they like, not only "None" and "Kind".

src/others/file_explorer/lv_file_explorer.h
```diff
@@ -0,0 +1,184 @@
...
+/*=====================
+ * Getter functions
+ *====================*/
+
+/**
+ * Get file explorer Selected file
+ * @param obj   pointer to a file explorer object
+ * @return      pointer to the file explorer selected file
+ */
+const char * lv_file_explorer_get_selected_fn(lv_obj_t * obj);
```
amirgon: If it selects files, why is it called "..._selected_fn"?  
"_fn" sounds to me like the shortcut of "function", which is unrelated here.

lv_conf_template.h
100ask: This is a good suggestion. Here I follow the previous format. If it needs to be modified, I think it should be modified together after file explorer merge. @kisvegabor

src/others/file_explorer/lv_file_explorer.h
100ask: This is a good suggestion, but I don't think you need to use callback functions, so I provide [lv_example_file_explorer_3](https://github.com/lvgl/lvgl/pull/3601/commits/50965f7305de0695cce89351833e13febd7ffa3d#diff-daa3025c3cd16cbe5f269aaf631de8f543cd756d865ec224b157ded7e4ab3008). 
What do you think?

examples/others/file_explorer/index.rst
100ask: Micropython examples not ready yet.

lv_conf_template.h
100ask: modified

src/others/file_explorer/lv_file_explorer.h
100ask: `fn` is the abbreviation of `file name`. When a file in the list is clicked, the file name will be returned. Users need to open files themselves.

lv_conf_template.h
kisvegabor: It really looks like as @100ask did with other widgets too. We can reconsider it, but it should be an other PR.

kisvegabor:
Failing Micropython CI: 
- It seems due to this feature the flash is overflowed, however I can see a way in the workflow file to set the device or any of its paramters.
- I don't know why the ESP CI is failing. IT seems like wrong LVGL config but the other Mictopython CI-s are working

@amirgon, @embeddedt ? 

@uraich or @100ask do you have time for adding Micropython examples?

Please enable the file explorer in the test's [CMakeFile](https://github.com/lvgl/lvgl/blob/master/tests/CMakeLists.txt#L197) to have it compiled with CI and spot warnings

embeddedt:
I will have to disable some more MicroPython features on STM32 when I get a chance. The reason this occured is that the binary was already very close to being too big before this change.

amirgon:
> * It seems due to this feature the flash is overflowed, however I can see a way in the workflow file to set the device or any of its parameters.

I don't think this should be a CI workflow parameter. If a certain board has smaller flash, it should explicitly disable some features by default. Otherwise users could get confused seeing the build passing on the CI but failing locally.  
Currently, this problem seems to be limited to only stm32.

> * I don't know why the ESP CI is failing. IT seems like wrong LVGL config but the other Micropython CI-s are working

When adding to lv_conf.h:
```c
#define LV_USE_FILE_EXPLORER 1
#define LV_FILE_EXPLORER_QUICK_ACCESS 1
```
then Micropython esp32 build passes.

In the case these macros are missing, as they are now, I expected `lv_conf_internal.h` to be in effect - but it's not included from anywhere.  
Is it supposed to be included automatically somehow? Or should we include it explicitly at the beginning of each header file?

src/others/file_explorer/lv_file_explorer.h
amirgon: Ok, maybe it's just my opinion, but `_fn` is usually associated with *function* and not with *filename*.  
Searching Micropython sources, for example, I can find `tcp_connected_fn`, `callback_fn` etc.  
So seeing `_fn` here is a bit confusing to me.

embeddedt:
I gutted the ethernet/networking support from the STM32 builds by default; this brings us to about 888KB out of 1MB used when compiling the binary for this PR, meaning there is still quite a bit of headroom left.

src/others/file_explorer/lv_file_explorer.h
YobieZhou: I got it. Modified

100ask:
> Failing Micropython CI:
> 
> * It seems due to this feature the flash is overflowed, however I can see a way in the workflow file to set the device or any of its paramters.
> * I don't know why the ESP CI is failing. IT seems like wrong LVGL config but the other Mictopython CI-s are working
> 
> @amirgon, @embeddedt ?
> 
> @uraich or @100ask do you have time for adding Micropython examples?
> 
> Please enable the file explorer in the test's [CMakeFile](https://github.com/lvgl/lvgl/blob/master/tests/CMakeLists.txt#L197) to have it compiled with CI and spot warnings

After that, I will add in other PR.

amirgon:
> After that, I will add in other PR.

See my [comment above ](https://github.com/lvgl/lvgl/pull/3601#issuecomment-1229543737) about `lv_conf_internal.h`.

uraich:
I have the MP examples for
gridnav
ffmpeg
msg
ready. Working on the others. Do you want me to create a PR for the examples already done or should I complete the set before? I also have the widget demo almost ready

amirgon:
> Do you want me to create a PR for the examples already done or should I complete the set before? I also have the widget demo almost ready

@uraich Very nice!
In my opinion it's easier to review and test them all together.  

I also noticed that [your last PR](https://github.com/lvgl/lvgl/pull/3634) was merged without Micropython tests passing... from that point all future commits to LVGL fail these tests.  
I was trying to fix some of them in https://github.com/lvgl/lvgl/pull/3635 but I think it's better to make sure all tests are passing and CI is clean before merging in the first place.

YobieZhou:
> > * It seems due to this feature the flash is overflowed, however I can see a way in the workflow file to set the device or any of its parameters.
> 
> I don't think this should be a CI workflow parameter. If a certain board has smaller flash, it should explicitly disable some features by default. Otherwise users could get confused seeing the build passing on the CI but failing locally. Currently, this problem seems to be limited to only stm32.
> 
> > * I don't know why the ESP CI is failing. IT seems like wrong LVGL config but the other Micropython CI-s are working
> 
> When adding to lv_conf.h:
> 
> ```c
> #define LV_USE_FILE_EXPLORER 1
> #define LV_FILE_EXPLORER_QUICK_ACCESS 1
> ```
> 
> then Micropython esp32 build passes.
> 
> In the case these macros are missing, as they are now, I expected `lv_conf_internal.h` to be in effect - but it's not included from anywhere. Is it supposed to be included automatically somehow? Or should we include it explicitly at the beginning of each header file?

LV_USE_FILE_EXPLORER needs to default to 0. Passed

YobieZhou:
> I have the MP examples for gridnav ffmpeg msg ready. Working on the others. Do you want me to create a PR for the examples already done or should I complete the set before? I also have the widget demo almost ready

very nice! I think it is better to commit in other PR

kisvegabor:
> I was trying to fix some of them in https://github.com/lvgl/lvgl/pull/3635 but I think it's better to make sure all tests are passing and CI is clean before merging in the first place.

My bad, okay.


So for this PR both failing CI is unrelated, right? So can it be merged?

src/others/file_explorer/lv_file_explorer.h
```diff
@@ -0,0 +1,190 @@
...
+    char * music_dir;
+    char * pictures_dir;
+    char * video_dir;
+    char * docs_dir;
+    char * fs_dir;
+#endif
+    const char * sel_fn;
+    char   current_path[LV_FILE_EXPLORER_PATH_MAX_LEN];
+    lv_file_explorer_sort_t sort;
+} lv_file_explorer_t;
```
amirgon, line 66: Why is the `lv_file_explorer_t` struct **definition** on the H file?
I would suggest putting it in the C file, only keeping the **declaration** on the H file if needed at all (`typedef struct lv_file_explorer_t;`)
This is also important for keeping the Micropython API smaller because there is no need to provide API and accessor logic for each field in this struct.

YobieZhou: Here follow the previous format. If it needs to be modified, I think it should be modified together after file explorer merge. @kisvegabor

kisvegabor: We expose these "classes" because it might be used by derived classes. So I vote for keeping it public. 

It's a long lasting discussion with @amirgon about how to keep the API smaller. If we want global change, I'm coming around to the idea of having e.g. `lv_btn.h` and `lv_btn_priv.h`. In C `*_priv.h` can be included in C files where needed.

YobieZhou: > > I was trying to fix some of them in #3635 but I think it's better to make sure all tests are passing and CI is clean before merging in the first place.
> 
> My bad, okay.
> 
> So for this PR both failing CI is unrelated, right? So can it be merged?

The two problems of CI are not involved in my PR. I think it should be fixed in other PR.

amirgon: @YobeZhou I believe that if you rebase this PR on latest LVGL these problems will be gone.

src/others/file_explorer/lv_file_explorer.h
```diff
@@ -0,0 +1,190 @@
...
+ * Setter functions
+ *====================*/
+
+#if LV_FILE_EXPLORER_QUICK_ACCESS
+/**
+ * Set file_explorer
+ * @param obj   pointer to a label object
+ * @param dir   the dir from 'lv_file_explorer_dir_t' enum.
+ */
+void lv_file_explorer_set_quick_access_path(lv_obj_t * obj, lv_file_explorer_dir_t dir, char * path);
```
amirgon: I believe it should be `const char * path`, no?

src/others/file_explorer/lv_file_explorer.h
```diff
@@ -0,0 +1,190 @@
...
+ * @return      pointer to the file explorer cur path
+ */
+const char * lv_file_explorer_get_current_path(const lv_obj_t * obj);
+
+/**
+ * Get file explorer head area obj
+ * @param obj   pointer to a file explorer object
+ * @return      pointer to the file explorer head area obj(lv_obj)
+ */
+lv_obj_t * lv_file_explorer_get_header(const lv_obj_t * obj);
```
amirgon: If the input is `const lv_obj_t * obj`, how could the output be a non-const `lv_obj_t *`?  
Same question for `lv_file_explorer_get_quick_access_area`, `lv_file_explorer_get_path_label`, etc.

src/others/file_explorer/lv_file_explorer.h
YobieZhou: > @YobeZhou I believe that if you rebase this PR on latest LVGL these problems will be gone.

If necessary, I will consider resubmitting PR after all discussions are completed. Before that, I wanted to know if I needed to do this? @kisvegabor

src/others/file_explorer/lv_file_explorer.h
YobieZhou: Thank you for your correction. It has been revised.

src/others/file_explorer/lv_file_explorer.h
YobieZhou: Modified.

amirgon (approved):
I have no more comments.  
I think it would be nice to also have a Micropython example, along with the C examples.

YobieZhou:
> I have no more comments. I think it would be nice to also have a Micropython example, along with the C examples.

I plan to finish it in the next PR, together with [lv_ime_pinyin](https://docs.lvgl.io/master/others/ime_pinyin.html#example).

src/others/file_explorer/lv_file_explorer.h
kisvegabor: Let leave the `lv_file_explorer_t` in the header file. If we change it we should change other widgets too.

kisvegabor:
All clear for my side. It'd be amazing if you could add Micropython example for both features.

Merging this now. Thank you @YobeZhou  for the this great widget, and @amirgon for the comments.

PR #3605, opened 2022-08-22
fix(group):be sure the default group pointer points to a proper place

hellototoro:

If the group is the default group, when deleting the group, set the default group as NULL.

### Checkpoints


kisvegabor:
Makes sense, thank you!

PR #3609, opened 2022-08-23
fix(style): Use compile time prop_cnt for const styles

pdgendt:
No need to end with an empty/invalid property in the const list.
This change allows to save space, and doesn't break existing lists.

### DIFF
--- a/src/misc/lv_style.h
+++ b/src/misc/lv_style.h
@@ -49,11 +49,26 @@ extern "C" {
 #define LV_IMG_ZOOM_NONE            256        /*Value for not zooming the image*/
 LV_EXPORT_CONST_INT(LV_IMG_ZOOM_NONE);
 
+// *INDENT-OFF*
 #if LV_USE_ASSERT_STYLE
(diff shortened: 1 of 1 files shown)


kisvegabor:
Good idea than you!

Let's wait for the CI, if it passes I'll merge it. `Verify that lv_conf_internal.h matches repository state / verify-conf-internal` will fail, it's a known issue.

kisvegabor:
There is a minor issue with code formatting. It seems the formatter likes it like [this](https://github.com/lvgl/lvgl/runs/7968579787?check_suite_focus=true).

pdgendt:
> There is a minor issue with code formatting. It seems the formatter likes it like [this](https://github.com/lvgl/lvgl/runs/7968579787?check_suite_focus=true).

And do you like it like that? It looks weird 😶

kisvegabor:
Looking into it again it really looks odd. Maybe the `{` `}` in the `define` confuses it. 

It seems better [disabling](http://astyle.sourceforge.net/astyle.html#_Disable_Formatting) astyle for this two macros.

kisvegabor:
Great, thanks!

pdgendt:
@kisvegabor thanks for the fast responses!

One note; I did create the PR for `release/v8.3`, but it's probably easy to cherry-pick to master?

kisvegabor:
Oh, thanks, I haven't noticed that. I just cherry picked it to `master` too.

PR #3612, opened 2022-08-24
fix(tileview): fix tileview display in predict out of range

ZPY-Shark:

In some low-performance touch devices, the read speed of the touch device is slow, or the read cycle of the input device is long.

When the touch slides too fast, the tile id predicted by the tileview in drag_end_handler may exceed the user-defined position range, and it will be judged as an invalid tile in lv_tileview_set_tile_act. 

At this time, the page may be moving to the screen following the touch position In the middle, if return directly, the screen display may stop in the middle of the two tiles and will not continue to slide, which looks like it is stuck.

I think there should be some protection for this, when the prediction is about to slide to an invalid tile, the next tile should be limited to a valid tile defined by the user, and continue to use the animation to slide.

### DIFF
--- a/src/lv_widgets/lv_tileview.c
+++ b/src/lv_widgets/lv_tileview.c
@@ -221,7 +221,13 @@ void lv_tileview_set_tile_act(lv_obj_t * tileview, lv_coord_t x, lv_coord_t y, l
         }
     }
 
-    if(valid == false) return; /*Don't load not valid tiles*/
+    if(valid == false)  /*if not valid tiles, load the last valid tiles*/
+    {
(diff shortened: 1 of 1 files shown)


kisvegabor:
Sorry for the late answer, I missed this issue somehow. 

Which LVGL version do you use? From "drag_end_handler" I guess it's v7.

ZPY-Shark:
> Sorry for the late answer, I missed this issue somehow.
> 
> Which LVGL version do you use? From "drag_end_handler" I guess it's v7.

Yes, the v7.11.0

kisvegabor:
v7 is already not supported. :(

PR #3621, opened 2022-08-25
fixed child fragment event dispatch

mariotaku:

Before, when sending an event to fragment, only the on on top of the stack will get the event. It was supposed to be sent to all attached child fragments (which has created object).

### Checkpoints


kisvegabor:
Upps, sorry, it seems I missed this PR. Merging now.

PR #3623, opened 2022-08-26
fix(transparency): call flush callback before clearing framebuffer

ambecker:

I use a SoC taht supports different, independent graphical layers.
These layers will be alpha-blenden by the hardware of the SoC and therefor I need a OSD Layer with support of transparency.

After following the documentation I received a fully transparent LVGL Framebuffer without any content although the framebuffer was filled before. Debugging showed me that the active Framebuffer is filled with zeroes **BEFORE** the flush callback was called.

After changing this order, my framebuffer was fine.

I did **not** check all other options or combination of options for possible side-effects of my change, because I am not that deep in LVGL that I could overlook all Feature.

I created the PullRequest for master branch although I use Release 8.3.1.
I was not sure if it is still possible to contribute to 8.3.x.
If yes I would really appreciate if this could be merged to 8.3.x as well.

### DIFF
--- a/src/core/lv_refr.c
+++ b/src/core/lv_refr.c
@@ -1191,6 +1191,23 @@ static void draw_buf_flush(lv_disp_t * disp)
     lv_draw_ctx_t * draw_ctx = disp->driver->draw_ctx;
     if(draw_ctx->wait_for_finish) draw_ctx->wait_for_finish(draw_ctx);
 
+    draw_buf->flushing = 1;
+
+    if(disp_refr->driver->draw_buf->last_area && disp_refr->driver->draw_buf->last_part) draw_buf->flushing_last = 1;
+    else draw_buf->flushing_last = 0;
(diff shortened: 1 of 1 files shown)


kisvegabor:
This part really seems messed up a little bit.  Please try it out with this patch 
[a.patch.zip](https://github.com/lvgl/lvgl/files/9436005/a.patch.zip).


From what I could test on PC it works well.

ambecker:
I tried your patch and it works very well for me.  :-)

kisvegabor:
Great! I fixed in the `release/v8.3` branch. This part will be changed in master soon so I haven't cherry-picked this change yet.

ambecker:
Thanks a lot for your great support.
LVGL works really great for me. :-)

kisvegabor:
Happy to hear that :slightly_smiling_face:

PR #3627, opened 2022-08-26
fix(style): check style count overflow

boraozgen:

`style_cnt` is a 6-bit variable, which overflows after 64 styles. Although it is not realistic to have that many styles, it is possible to overflow this variable due to a leak of `lv_obj_add_style` without cleaning up, which breaks havoc as it is not checked. This PR adds checks for overflow.

### Checkpoints


embeddedt:
Good idea, thanks!

PR #3629, opened 2022-08-26
fix(ffmpeg) Ignore deprecated warning

amirgon:
When building LVGL with Micropython, warnings are treated as errors by default.

This fix works around the following warnings/errors:

```
CC ../../lib/lv_bindings/lvgl/src/libs/ffmpeg/lv_ffmpeg.c
../../lib/lv_bindings/lvgl/src/libs/ffmpeg/lv_ffmpeg.c: In function ‘ffmpeg_open_codec_context’:
../../lib/lv_bindings/lvgl/src/libs/ffmpeg/lv_ffmpeg.c:537:13: error: assignment discards ‘const’ qualifier from pointer target type [-Werror=discarded-qualifiers]
  537 |         dec = avcodec_find_decoder(st->codecpar->codec_id);
      |             ^
../../lib/lv_bindings/lvgl/src/libs/ffmpeg/lv_ffmpeg.c: In function ‘ffmpeg_image_allocate’:
../../lib/lv_bindings/lvgl/src/libs/ffmpeg/lv_ffmpeg.c:767:5: error: ‘av_init_packet’ is deprecated [-Werror=deprecated-declarations]
  767 |     av_init_packet(&ffmpeg_ctx->pkt);
      |     ^~~~~~~~~~~~~~
In file included from /usr/local/include/libavcodec/avcodec.h:45,
                 from ../../lib/lv_bindings/lvgl/src/libs/ffmpeg/lv_ffmpeg.c:12:
/usr/local/include/libavcodec/packet.h:512:6: note: declared here
  512 | void av_init_packet(AVPacket *pkt);
      |      ^~~~~~~~~~~~~~
```

### DIFF
--- a/src/libs/ffmpeg/lv_ffmpeg.c
+++ b/src/libs/ffmpeg/lv_ffmpeg.c
@@ -520,7 +520,7 @@ static int ffmpeg_open_codec_context(int * stream_idx,
     int ret;
     int stream_index;
     AVStream * st;
-    AVCodec * dec = NULL;
+    const AVCodec * dec = NULL;
     AVDictionary * opts = NULL;
 
     ret = av_find_best_stream(fmt_ctx, type, -1, -1, NULL, 0);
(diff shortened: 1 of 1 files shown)


embeddedt:
Thanks! I think @uraich had rewritten this locally not to depend on the deprecated API anymore but this should be a good fix in the meantime.

uraich:
Yes, I tried to correct lv_ffmpeg.c using the av_packet_alloc instead of av_init_packet.
You have to make a few more changes because av_pack_alloc returns pointer to the packet while av_init_packet returns the packet itself.
I uploaded it temporarily to [https://github.com/uraich/lv_mpy_examples_v8/blob/main/libs/ffmpeg/lv_ffmpeg.c](https://github.com/uraich/lv_mpy_examples_v8/blob/main/libs/ffmpeg/lv_ffmpeg.c)
so you can have a look at it. The new example programs already running are also available in this repo.

Issue #3535, opened 2022-07-30
Missing Micropython examples


amirgon:
I'm opening this issue to point out missing Micropython examples:

- [ffmpeg](https://github.com/lvgl/lvgl/tree/master/examples/libs/ffmpeg)
- [fragment](https://github.com/lvgl/lvgl/tree/master/examples/others/fragment)
- [gridnav](https://github.com/lvgl/lvgl/tree/master/examples/others/gridnav)
- [ime](https://github.com/lvgl/lvgl/tree/master/examples/others/ime)
- [imgfont](https://github.com/lvgl/lvgl/tree/master/examples/others/imgfont)
- [msg](https://github.com/lvgl/lvgl/tree/master/examples/others/msg) (only [lv_example_msg_1.py](https://github.com/lvgl/lvgl/blob/master/examples/others/msg/lv_example_msg_1.py) implemented)

Also, [Demos](https://github.com/lvgl/lvgl/tree/master/demos) are missing.  
[lv_demo_widgets.py](https://github.com/lvgl/lvgl/blob/master/demos/widgets/lv_demo_widgets.py) was once implemented, but is now outdated and doesn't work.

When someone adds a new C example, I think it's best asking them to provide Micropython example as well.  
The Micropython examples are [automatically tested in the CI](https://github.com/lvgl/lvgl/blob/master/.github/workflows/build_micropython.yml) so adding a Micropython example can ensure that the Micropython API is correct and the C callback conventions are respected for that feature.

Regarding demos, today these are not tested automatically in the CI but we could easily fix that when they are available.

CC: @kisvegabor @embeddedt @uraich

uraich:
Hi AmirI had already started to implement the widget demos. It is not complete yet but there is very little missing. I will try to complete it in the coming days and create a pull request.UliSent from my Galaxy
-------- Original message --------From: Amir Gonnen ***@***.***> Date: 30/07/2022  23:31  (GMT+01:00) To: lvgl/lvgl ***@***.***> Cc: Uli Raich ***@***.***>, Mention ***@***.***> Subject: [lvgl/lvgl] Missing Micropython examples (Issue #3535) 
I'm opening this issue to point out missing Micropython examples:

ffmpeg
fragment
gridnav
ime
imgfont
msg (only lv_example_msg_1.py implemented)

Also, Demos are missing.
lv_demo_widgets.py was once implemented, but is now outdated and doesn't work.
[...]

PR #3631, opened 2022-08-26
fix(pinyin) expose lv_ime_pinyin_class

amirgon:
Micropython bindings requires the class prototype to be exposed in the header, so it could refer to it.

Related: https://github.com/lvgl/lvgl/issues/3535#issuecomment-1221318545

### DIFF
--- a/src/others/ime/lv_ime_pinyin.h
+++ b/src/others/ime/lv_ime_pinyin.h
@@ -25,8 +25,6 @@ extern "C" {
  *      TYPEDEFS
  **********************/
 
-const lv_obj_class_t lv_ime_pinyin_class;
-
 typedef enum {
     LV_IME_PINYIN_MODE_K26,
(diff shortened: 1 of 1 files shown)


src/others/ime/lv_ime_pinyin.h
```diff
@@ -25,6 +25,8 @@ extern "C" {
  *      TYPEDEFS
  **********************/
 
+const lv_obj_class_t lv_ime_pinyin_class;
```
embeddedt: I think this needs to be `extern` and should be in the "GLOBAL VARIABLES" section.

embeddedt (changes requested):

src/others/ime/lv_ime_pinyin.h
amirgon: Done on 3d78b38b3

PR #3641, opened 2022-08-30
replaced lv_widget_demo.py by new version

uraich:

Repaced lv_widget_demo.py by new version. The old one was outdated and did not work anymore. I had to add the necessary fonts to the assets of the widget demos 

### Checkpoints


demos/widgets/lv_demo_widgets.py
```diff
@@ -1,572 +1,1139 @@
-#!/opt/bin/lv_micropython
+#!/opt/bin/lv_micropython -i
 import lvgl as lv
-import time
-import SDL
-lv.init()
+import display_driver_utils
+import fs_driver
+import time,os,sys
+from imagetools import get_png_info, open_png
```
amirgon: imagetools removed on latest version, so this line can be removed.

demos/widgets/lv_demo_widgets.py
```diff
@@ -1,572 +1,1139 @@
...
-        h.set_width(lv.page.get_width_fit(page))
-        h.set_layout(lv.LAYOUT.COLUMN_MID)
-    elif disp_size <= lv.DISP_SIZE.MEDIUM:
-        h.set_size(lv.obj.get_width(lv.obj.__cast__(cal)), lv.obj.get_height(lv.obj.__cast__(cal)))
-        h.set_layout(lv.LAYOUT.PRETTY_TOP);
+# needed for dynamic font loading
+fs_drv = lv.fs_drv_t()
+fs_driver.fs_register(fs_drv, 'S')
+
+# Register PNG image decoder
```
amirgon: No need to register PNG decoder on latest version, so these lines can be removed.

demos/widgets/lv_demo_widgets.py
```diff
@@ -1,572 +1,1139 @@
...
+    tab_btns.set_style_pad_left(LV_HOR_RES / 2, 0)
+    # logo = lv_img_create(tab_btns);
+    #    LV_IMG_DECLARE(img_lvgl_logo);
+    #    lv_img_set_src(logo, &img_lvgl_logo);
+    #    lv_obj_align(logo, LV_ALIGN_LEFT_MID, -LV_HOR_RES / 2 + 25, 0);
+
+    label = lv.label(tab_btns)
+    label.add_style(style_title, 0)
+    label.set_text("LVGL v8")
+    label.align_to(logo, lv.ALIGN.OUT_RIGHT_TOP, 10, 0)
```
amirgon, line 1059: `logo` above is commented out, so you can't `align_to` it.

demos/widgets/lv_demo_widgets.py
```diff
@@ -1,572 +1,1139 @@
...
+
+    label = lv.label(tab_btns)
+    label.set_text("Widgets demo")
+    label.add_style(style_text_muted, 0)
+    label.align_to(logo, lv.ALIGN.OUT_RIGHT_BOTTOM, 10, 0)
```
amirgon, line 1064: logo undefined

demos/widgets/lv_demo_widgets.py
```diff
@@ -1,572 +1,1139 @@
...
+        grid1_col_dsc = [LV_GRID_FR(1), LV_GRID_FR(1), lv.GRID_TEMPLATE_LAST]
+        grid1_row_dsc = [
+            lv.GRID_CONTENT,  # Title
+            lv.GRID_CONTENT,  # Sub title
+            20,               # Spacer
+            lv.GRID_CONTENT,  # Amount
+            lv.GRID_CONTENT,  # Hint
+            lv.GRID_TEMPLATE_LAST]
+
+        chart3.set_size(lv_pct(100), lv_pct(100))
```
amirgon: Should be `lv.pct`

demos/widgets/lv_demo_widgets.py
```diff
@@ -1,572 +1,1139 @@
...
+        chart3.set_style_pad_column(lv.dpx(30), 0)
+
+        panel1.grid_dsc_array(grid1,col_dsc, grid1_row_dsc)
```
amirgon: `grid_dsc_array` is undefined.

demos/widgets/lv_demo_widgets.py
```diff
@@ -1,572 +1,1139 @@
...
+    btn.set_style_radius(lv.RADIUS_CIRCLE, 0)
+    # btn.add_event_cb(color_changer_event_cb, lv.EVENT-ALL, color_cont)
+    btn.set_style_shadow_width(0, 0)
+    btn.set_style_bg_img_src(lv.SYMBOL.NEW_LINE, 0)
+
+    if disp_size == DISP_SMALL:
+         btn.set_size(lv.dpx(42), lv.dpx(42))
+         btn.align(lv.ALIGN.BOTTOM_RIGHT, -lv.dpx(15), -lv.dpx(15))
+    else :
+         btn.set_size(lv.dpx(50), lv.dpv(50))
```
amirgon: `lv.dpv` is undefined. Did you mean `dpx`?

amirgon (changes requested):
Hi Uli

Nice work! :clap: 

I noticed that the window is a bit small for the content.
Maybe worth configuring a larger window? 
(You can set it when you call `display_driver_utils.driver`)

Screenshot from 2022-08-30 23-17-33

When trying to change it to larger screen I noticed some errors, and the screen didn't look correct.

demos/widgets/lv_demo_widgets.py
```diff
@@ -1,572 +1,1139 @@
...
+        title.set_grid_cell(lv.GRID_ALIGN.START, 0, 1, lv.GRID_ALIGN.START, 0, 1)
+        date.set_grid_cell(lv.GRID_ALIGN.START, 0, 1, lv.GRID_ALIGN.START, 1, 1)
+        amount.set_grid_cell(lv.GRID_ALIGN.START, 0, 1, lv.GRID_ALIGN.START, 3, 1)
+        hint.set_grid_cell(lv.GRID_ALIGN.START, 0, 1, lv.GRID_ALIGN.START, 4, 1)
+        chart3.set_grid_cell(lv.GRID_ALIGN.STRETCH, 1, 1, lv.GRID_ALIGN.STRETCH, 0, 5)
+    elif disp_size == DISP_MEDIUM :
+        grid1_col_dsc = [LV_GRID_FR(1), LV_GRID_FR(1), lv.GRID_TEMPLATE_LAST]
+        grid1_row_dsc = [
+                lv.GRID_CONTENT,  # Title + Date
+                Llv.GRID_CONTENT, # Amount + Hint
```
amirgon: Typo: `Llv`

demos/widgets/lv_demo_widgets.py
```diff
@@ -1,572 +1,1139 @@
...
+            tv.update_layout()   # Be sure the sizes are recalculated
+            tv.set_height(LV_VER_RES - kb.get_height())
+            kb.clear_flag(lv.obj.FLAG.HIDDEN)
+            ta.scroll_to_view_recursive(lv.ANIM.OFF)
+ 
+    elif code == lv.EVENT.DEFOCUSED :
+        kb.set_textarea(None)
+        tv.set_height(LV_VER_RES)
+        kb.add_flag(lv.obj.FLAG.HIDDEN)
+        indev.reset(None)
```
amirgon: `indev` could be `None`.  
I suggest checking this first (`if indev:`...)

demos/widgets/lv_demo_widgets.py
```diff
@@ -1,572 +1,1139 @@
...
+
+    elif code == lv.EVENT.READY or code == lv.EVENT.CANCEL:
+        tv.set_height(LV_VER_RES)
+        kb.add_flag(lv.obj.FLAG.HIDDEN)
+        ta.clear_state(lv.STATE.FOCUSED)
+        indev.reset(None)    # To forget the last clicked object to make it focusable again
```
amirgon: `indev` could be `None`.  
I suggest checking this first (`if indev:`...)

demos/widgets/lv_demo_widgets.py
```diff
@@ -1,572 +1,1139 @@
...
+t2 = tv.add_tab("Analytics")
+t3 = tv.add_tab("Shop")
+
+profile_create(t1)
+analytics_create(t2)
+shop_create(t3)
+
+color_changer_create(tv)
+
+while True:
```
amirgon: I think it would be a good idea to add this demo to the automated testing.  
Today every PR or commit goes through a series of tests (GitHub action CI workflow).   
One of these tests involve building the Micropython unix port **and running all the examples**.  
The advantage is that if some change breaks the example code, we know about this immediately.

Your demo can also run automatically, but not with this `while True:` loop.  
So I suggest removing it.  

We can document that on the unix port this needs to be run with `-i` command line option to prevent the window from closing.

Here is how to run your demo with the test automation:

```bash
ports/unix/micropython-dev lib/lv_bindings/tests/run_test.py lib/lv_bindings/lvgl/demos/widgets/lv_demo_widgets.py
```

When it's ready, I will add it to [the script](https://github.com/lvgl/lv_binding_micropython/blob/6431ec4b1c66c92880a533020eb4802ea43694b0/tests/run.sh#L9) so that would run automatically.

uraich:
Hi Amir,
Thanks for the comments. In fact I saw that I had quite a lot of code commented out, which is the reason that it only worked for the lowest resolution screen. On my local copy this is corrected.
I sill have problems with the event callback on the charts.
Btw: I found that lv_example_chart_4 and 6 don't work anymore. I managed to correct  lv_example_chart_6 but lv_example_chart_4 resists.
I will try a few more days and if I don't manage I will come back to you.

demos/widgets/lv_demo_widgets.py
uraich: loop removed

uraich:
Hi Amir,
In your first comment you said: "Nice work". However, I am not really proud of myself. I think I created the PR prematurely and I should have done quite a bit more testing before.
I saw too late that in fact a lot of code had been commented out and was not ported. This is true for the layout for different screen sizes as well as several event callbacks. The new version is quite a bit more complete and, I hope, a lot better.
I still have a problem with the chart event callback when switching LV_USE_DRAW_MASKS on. I hope to solve this problem soon. 
You can now easily switch screen sizes by setting SCREEN_SIZE to
- SCREEN_SIZE_SMALL
- SCREEN_SIZE_MEDIUM
- SCREEN_SIZE_LARGE
I think that many of your comments should be resolved on this version.

amirgon:
> * I think that many of your comments should be resolved on this version.

Hi Uli.
It surely looks better now! :+1: 

Could you mention which problems were resolved?
I noticed, for example, that my comments about the `imagetools` removal are still valid.

Another problem I'm seeing is related to paths.  
For example, you open `../../assets/font/montserrat-24.fnt` so you have an assumption about the "current directory".  
It would be best if it could work without relying on current directory. You can do that by opening files relative to the script path.  
You can see an example of that on the [PNG example](https://github.com/lvgl/lv_binding_micropython/blob/6431ec4b1c66c92880a533020eb4802ea43694b0/examples/png_example.py#L20-L23) where the path of the PNG file is considered relative to the script dir, regardless of current directory.

One thing that could help automatic testing is protecting `indev` access. The tests send events but can't guarantee there's a valid input device active.
So, when you access it, just check that it's valid like this:

```python
if (not indev) or indev.get_type() != lv.INDEV_TYPE.KEYPAD :
```

(it's evaluated to `True` when indev is invalid to enter the "if" and test it)

uraich:
In fact I found so much code that was commented or removed, that I decided to put all this back in a new step. (more than 300 lines of code were missing!)  I tried your proposal for imagetool but it did not work yet. I think my lv_micropython version, although it was only a few days old, was already outdated. The png_example.py still included imagetools.
I will now go through all your comments and make sure they are all taken into account. I am sure, most of them have already been treated.
I just checked removing imagetools on the new version of lv_micropython and this works. However, the newest change breaks the img examples because imagetools has been removed entirely.

uraich:
I think that I completed the lv_widget_demo but maybe there is something I overlooked? I also believe I treated all of Amir's comments and I managed to get rid of the last problem I saw, that had to do with the fading effect of one of the charts.
Looking at some remaining problems with the MP example programs now.
I will try to get this out of the way by the end of next week because then I will be on travel for a week.

kisvegabor:
> I think that I completed the lv_widget_demo 

It's great! It proves that the MP binding is capable f doing such a complex UI. I definitely agree to add this to the CI.

amirgon:
Hi Uli,

Maybe I miss something but there seem to be some regression.
With `SCREEN_SIZE_LARGE` I see this:

Screenshot from 2022-09-03 23-56-30

And with `SCREEN_SIZE_MEDIUM` there seem to be missing parts:
image

Also, here are some small changes I suggest adding for automatic testing to work correctly:

```diff
diff --git a/demos/widgets/lv_demo_widgets.py b/demos/widgets/lv_demo_widgets.py
index 8055aacda..058667eae 100755
--- a/demos/widgets/lv_demo_widgets.py
+++ b/demos/widgets/lv_demo_widgets.py
@@ -1266,7 +1266,7 @@ def ta_event_cb(e,kb,tv) :
     
     indev = lv.indev_get_act()
     if code == lv.EVENT.FOCUSED :
-        if indev.get_type() != lv.INDEV_TYPE.KEYPAD :
+        if (not indev) or indev.get_type() != lv.INDEV_TYPE.KEYPAD :
             kb.set_textarea(ta)
             kb.set_style_max_height(LV_HOR_RES * 2 // 3, 0)
             tv.update_layout()   # Be sure the sizes are recalculated
@@ -1278,14 +1278,14 @@ def ta_event_cb(e,kb,tv) :
         kb.set_textarea(None)
         tv.set_height(LV_VER_RES)
         kb.add_flag(lv.obj.FLAG.HIDDEN)
-        if (not indev) :
+        if (indev) :
             indev.reset(None)
 
     elif code == lv.EVENT.READY or code == lv.EVENT.CANCEL:
         tv.set_height(LV_VER_RES)
         kb.add_flag(lv.obj.FLAG.HIDDEN)
         ta.clear_state(lv.STATE.FOCUSED)
-        if (not indev) :
+        if (indev) :
             indev.reset(None)    # To forget the last clicked object to make it focusable again
 
 def birthday_event_cb(e):
@@ -1295,7 +1295,7 @@ def birthday_event_cb(e):
 
     if code == lv.EVENT.FOCUSED :
         indev = lv.indev_get_act()
-        if indev.get_type() == lv.INDEV_TYPE.POINTER:
+        if (not indev) or indev.get_type() == lv.INDEV_TYPE.POINTER:
             if calendar == None:
                 top_layer = lv.layer_top()
                 top_layer.add_flag(lv.obj.FLAG.CLICKABLE)
```

uraich:
Did the changes for automatic testing as required.
The other problems I don't see here:
SCREEN_SIZE_LARGE
image
SCREEN_SIZE_MEDIUM
image
Is there a problem that it does not find either the png or the font files on your system?

amirgon:
These are file path problems.
In some places it was missing `script_path`.
(btw you can [apply](https://gist.github.com/nepsilon/22bc62a23f785716705c) this patch below with `git apply`, instead of fixing it manually)

```diff
diff --git a/demos/widgets/lv_demo_widgets.py b/demos/widgets/lv_demo_widgets.py
index d48626cc1..d30ba00d5 100755
--- a/demos/widgets/lv_demo_widgets.py
+++ b/demos/widgets/lv_demo_widgets.py
@@ -812,7 +812,7 @@ def shop_create(parent):
     # Create an image from the png file
     try:
         if osVersion == "unix":
-            with open('assets/clothes.png','rb') as f:
+            with open(script_path + '/assets/clothes.png','rb') as f:
                 png_data = f.read()
         else:
             with open('images/clothes.png','rb') as f:
@@ -988,7 +988,7 @@ else:  # disp_size == DISP_SMALL
         if osVersion == "esp32":
             font_large = lv.font_load("S:/font/montserrat-18.fnt")
         else:
-            font_large = lv.font_load("S:../../assets/font/montserrat-18.fnt")
+            font_large = lv.font_load("S:" + script_path + "/assets/font/montserrat-18.fnt")
         
     try:
         font_normal = lv.font_montserrat_12
@@ -997,7 +997,7 @@ else:  # disp_size == DISP_SMALL
         if osVersion == "esp32":        
             font_normal = lv.font_load("S:/font/montserrat-12.fnt")            
         else:
-            font_normal = lv.font_load("S:../../assets/font/montserrat-12.fnt")            
+            font_normal = lv.font_load("S:" + script_path + "/assets/font/montserrat-12.fnt")            
 
 if not font_large or not font_normal:
     print("font loading failed")
@@ -1034,7 +1034,7 @@ if disp_size == DISP_LARGE:
     # Create an image from the png file
     try:
         if osVersion == "unix":
-            with open('assets/lvgl_logo.png','rb') as f:
+            with open(script_path + '/assets/lvgl_logo.png','rb') as f:
                 png_data = f.read()
         else:
             with open('images/lvgl_logo.png','rb') as f:
```

uraich:
Another quick question:
In order to port the file_explorer examples I need to figure out the user's home directory. Can this be done in MicroPython? Either the home directory itself (e.g. by figuring out $HOME) or the user name.
For the moment I read the contents of /home and take the first entry that is not lost+found.

amirgon:
> In order to port the file_explorer examples I need to figure out the user's home directory. Can this be done in MicroPython?

lv_file_explorer uses [lv_fs](https://github.com/lvgl/lvgl/blob/24dac61abf62e598cec3b42f99d43e29fda1a15e/src/others/file_explorer/lv_file_explorer.c#L555) to list files in a directory.  
Today we have in Micropython [fs_driver.py](https://github.com/lvgl/lv_binding_micropython/blob/master/lib/fs_driver.py) which implement lv_fs only partly: it has the functions for opening and reading/writing files, but not the functions for listing directories.

So we would need some implementation of `dir_open_cb`, `dir_read_cb` and `dir_close_cb`. ([link to docs](https://docs.lvgl.io/master/overview/file-system.html#adding-a-driver)).  

One option is to create a "dummy" fs driver as part of the example, which returns a predefined "faked" list of files for the file_explorer example. This should be pretty easy and remove the need to actually read some directory contents.

amirgon (approved):
The demo looks good now, I have no more comments.
Thanks Uli!

uraich:
> > In order to port the file_explorer examples I need to figure out the user's home directory. Can this be done in MicroPython?
> 
> lv_file_explorer uses [lv_fs](https://github.com/lvgl/lvgl/blob/24dac61abf62e598cec3b42f99d43e29fda1a15e/src/others/file_explorer/lv_file_explorer.c#L555) to list files in a directory. Today we have in Micropython [fs_driver.py](https://github.com/lvgl/lv_binding_micropython/blob/master/lib/fs_driver.py) which implement lv_fs only partly: it has the functions for opening and reading/writing files, but not the functions for listing directories.
> 
> So we would need some implementation of `dir_open_cb`, `dir_read_cb` and `dir_close_cb`. ([link to docs](https://docs.lvgl.io/master/overview/file-system.html#adding-a-driver)).
> 
> One option is to create a "dummy" fs driver as part of the example, which returns a predefined "faked" list of files for the file_explorer example. This should be pretty easy and remove the need to actually read some directory contents.

I just saw that os.getenv('HOME')  works!
I need this to create the file paths for the quick access buttons.

uraich:
One more question:
How do I translate LV_IME_PINYIN_MODE_K9?
I tried lv.IME_PINYIN_MODE.K9 but it does not work. This is what I guessed from lv_mpy.c.
There is a similar value in gridnav: lv.GRIDNAV_CTRL.ROLLOVER and that works.
image
Does this work for you? or am I missing something?

amirgon:
Uli why did you close this PR?

> How do I translate LV_IME_PINYIN_MODE_K9?

`lv.ime_pinyin.PINYIN_MODE.K9`

uraich:
I did something wrong when trying to create the new PRs for the missing examples and then I must have closed this PR, which was not my intention. I re-opened it.
I saw that there is also a new event example (example 4). This uses 
lv_draw_polygon(draw_ctx, &draw_dsc, points, n);
points is an array of points (pointers to lv.point_t) How can I create such an array in Python?

amirgon:
> I saw that there is also a new event example (example 4). This uses
> lv_draw_polygon(draw_ctx, &draw_dsc, points, n);
> points is an array of points (pointers to lv.point_t) How can I create such an array in Python?

By a list of points:

```python
points = [
    lv.point_t({'x': 50,  'y': 50}),
    lv.point_t({'x': 150, 'y': 80}),
    lv.point_t({'x': 80,  'y': 150}),    
]
canvas.draw_polygon(points, len(points), draw_dsc)
```

Here is an [online example](https://sim.lvgl.io/v9.0/micropython/ports/javascript/index.html?script_direct=8f54ba3678b04550c60ca8d145e6bbdbc889f869).

---

In fact, there is another way which is more C-like:

```python
_NUM_POINTS=3
points = lv.point_t(_NUM_POINTS)
points[0] = {'x': 50,  'y': 50}
points[1] = {'x': 150, 'y': 80}
points[2] = {'x': 80,  'y': 150}

canvas.draw_polygon(points, _NUM_POINTS, draw_dsc)
```

Another [online exmaple](https://sim.lvgl.io/v9.0/micropython/ports/javascript/index.html?script_direct=76d008c0712496889ff2e7aaab56fcbce43d33d7)

uraich:
Interesting! This works on the canvas but it does not seem to work on draw_ctx.draw_polygon(draw_ctx, draw_dsc, points, self.n)
Here I get:
image

uraich:
lv_demo_keypad_encoder creates a lv_indev_t * cur_drv = NULL. Then it passes through a linked list of input devices with
cur_drc = lv_indev_get_next(cur_drv) and it picks the keypad and encoder drivers. This I cannot do in Python.
I therefore picked up keyboard from the driver class in display_driver_utils. Is there a more elegant way of doing this?
When running the test I get:
image
However, the program runs just fine:
image

amirgon:
> Interesting! This works on the canvas but it does not seem to work on draw_ctx.draw_polygon(draw_ctx, draw_dsc, points, self.n)

That's because `draw_polygon` receives points arguments as `const lv_point_t * points` instead of `const lv_point_t points[]`.
On C it's the same, but the Micropython binding interprets this differently (a pointer to a single point vs. an array of multiple points).

@kisvegabor - I suggest fixing this on `draw_polygon` and on any other case where the expected input is an *array* of something, instead of a pointer to a single object.  
Can you think of additional cases like this?

> lv_demo_keypad_encoder creates a lv_indev_t * cur_drv = NULL. Then it passes through a linked list of input devices with
> cur_drc = lv_indev_get_next(cur_drv) and it picks the keypad and encoder drivers. This I cannot do in Python.

@uraich - It's a little tricky, but actually you can do it in Python:

```python
null_indev = lv.indev_t.__cast__(None) 
indev = null_indev.get_next()
while (indev):
    if indev.driver.type == lv.INDEV_TYPE.KEYPAD:
       ...
    ...
    indev = indev.get_next()
```

uraich:
ah, you have to play with casts again! Ok, I will try soon. 
Tomorrow I will be on travel for a week and I will have less time to look into lvgl but I will be back in a week's time

uraich:
I tried to replace const lv_point_t *points by const lv_point_t points[] but then I get
image
This happens when I make the change in src/draw/lv_draw.h

kisvegabor:
>  I suggest fixing this on draw_polygon and on any other case where the expected input is an array of something, instead of a pointer to a single object.
Can you think of additional cases like this?

Updated here 985f890

I think we have already fixed most of the prototypes like this. Polygon is rarely used so it could slip through.

kisvegabor:
Do you see what did I do wrong? 
https://github.com/lvgl/lvgl/actions/runs/3044398713/jobs/4904743784

uraich:
Hi Gabor,
Please check my last message on this thread. I had the same problem and I already asked Amir about it.

kisvegabor:
Oh, thanks.

amirgon:
> I tried to replace const lv_point_t *points by const lv_point_t points[] but then I get

@uraich That was a bug in the binding generation script!  
Fixed here: https://github.com/lvgl/lv_binding_micropython/commit/23394fed96be3a2b20633e5304e6ea407c574896

@kisvegabor this should fix https://github.com/lvgl/lvgl/commit/985f890b48ea0a5b080a4afa9192c6776ad96911 now.

uraich:
Yes, it works now. Did you see why lv_example_imgfont_1.py is crashing? As far as I can see, this is the last missing example (apart from the fragment ones). As soon as we get this to work, I will integrate it into the missing examples PR and I think we can close this chapter.
In general, I think it was beneficial to try implementing the missing examples because it exhibited a few errors or inconsistencies left in the code.

amirgon:
> Did you see why lv_example_imgfont_1.py is crashing? As far as I can see, this is the last missing example (apart from the fragment ones). As soon as we get this to work, I will integrate it into the missing examples PR and I think we can close this chapter.

I started looking at that.  
I think the problem there is a mix-up between the `user_data` on the font struct and the `user_data` that is passed as argument to the function. The callback might be selecting the wrong one.
I'll look into it and update.


> In general, I think it was beneficial to try implementing the missing examples because it exhibited a few errors or inconsistencies left in the code.

I agree. Thanks for the hard work Uli!

uraich:
Btw: With the fix for plotting the polygon I also got the MP port of what was lv_example_event_4 to work. We can add it as example_5 if you want. 
Hmmm. in fact it crashes when I go beyond 14 points in the polygon, but so does the C version

kisvegabor:
I think the current example is better because the previous was too complicated to show only how events work.

PR #3649, opened 2022-09-02
fix(lv_obj_style): initialise value_act with default values

douglas-healy:
This change fixes a compiler warning indicating that `value_act` may be used uninitialised.
When declaring `value_act` initialise it with default values.

### DIFF
--- a/src/core/lv_obj_style.c
+++ b/src/core/lv_obj_style.c
@@ -224,7 +224,7 @@ void lv_obj_enable_style_refresh(bool en)
 
 lv_style_value_t lv_obj_get_style_prop(const lv_obj_t * obj, lv_part_t part, lv_style_prop_t prop)
 {
-    lv_style_value_t value_act = { .num = 0 };
+    lv_style_value_t value_act = { .ptr = NULL };
     bool inheritable = lv_style_prop_has_flag(prop, LV_STYLE_PROP_FLAG_INHERITABLE);
     lv_style_res_t found = LV_STYLE_RES_NOT_FOUND;
     while(obj) {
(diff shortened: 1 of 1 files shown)


src/core/lv_obj_style.c
```diff
@@ -224,7 +224,7 @@ void lv_obj_enable_style_refresh(bool en)
 
 lv_style_value_t lv_obj_get_style_prop(const lv_obj_t * obj, lv_part_t part, lv_style_prop_t prop)
 {
-    lv_style_value_t value_act;
+    lv_style_value_t value_act = {.num = 0, .ptr = NULL, .color.full = 0};
```
kisvegabor: It produces a warning. Maybe `lv_style_value_t value_act = { 0 }`

src/core/lv_obj_style.c
```diff
@@ -224,7 +224,7 @@ void lv_obj_enable_style_refresh(bool en)
 
 lv_style_value_t lv_obj_get_style_prop(const lv_obj_t * obj, lv_part_t part, lv_style_prop_t prop)
 {
-    lv_style_value_t value_act;
+    lv_style_value_t value_act = { .num = 0 };
```
```suggestion
    lv_style_value_t value_act = { .ptr = NULL };
```

On 64 bit machines `num` is `int32_t` (4 bytes) but `ptr` is 8 bytes.

src/core/lv_obj_style.c
douglas-healy: Hi @kisvegabor, I have pushed a new commit. I realised that `value_act` is a union and for some reason I was thinking it is a struct. I have now just initialised the `num` member to zero 😊

kisvegabor:
I made a minor change, and all tests pass now. Merging, thank you!

douglas-healy:
@kisvegabor, I agree with the change. Thank you for reviewing.

PR #3654, opened 2022-09-02
fix(examples): resolve string formatting compiler warnings

douglas-healy:

This change resolves compiler warnings generated from string formatting operations in certain example files.

### DIFF
--- a/examples/layouts/flex/lv_example_flex_1.c
+++ b/examples/layouts/flex/lv_example_flex_1.c
@@ -28,7 +28,7 @@ void lv_example_flex_1(void)
         lv_obj_set_size(obj, 100, LV_PCT(100));
 
         label = lv_label_create(obj);
-        lv_label_set_text_fmt(label, "Item: %lu", i);
+        lv_label_set_text_fmt(label, "Item: %"LV_PRIu32"", i);
         lv_obj_center(label);
 
         /*Add items to the column*/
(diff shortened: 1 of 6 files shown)


examples/layouts/flex/lv_example_flex_1.c
```diff
@@ -28,7 +28,7 @@ void lv_example_flex_1(void)
         lv_obj_set_size(obj, 100, LV_PCT(100));
 
         label = lv_label_create(obj);
-        lv_label_set_text_fmt(label, "Item: %u", i);
+        lv_label_set_text_fmt(label, "Item: %lu", i);
```
embeddedt: These should probably just be replaced with the `LV_PRI` macros as well; there is no way to easily get the old-style placeholders to work correctly on every platform.

douglas-healy: Thanks @embeddedt, I will look into that. I see some of the checks for certain platforms did fail so you are right about using the `LV_PRI` macros.

kisvegabor:
All tests pass now. This PR is "draft" so I wonder if there is still something to so.

douglas-healy:
> All tests pass now. This PR is "draft" so I wonder if there is still something to so.

@kisvegabor, this PR is ready for merge.

embeddedt (approved):

PR #3668, opened 2022-09-06
fix(draw): center imgfont's drawing position

FASTSHIFT:
imgfont needs to be centered in the line.

### Checkpoints

### DIFF
--- a/examples/others/imgfont/lv_example_imgfont_1.py
+++ b/examples/others/imgfont/lv_example_imgfont_1.py
@@ -14,7 +14,7 @@ try:
 except NameError: 
     script_path = ''
     
-def get_imgfont_path(font, img_src, length, unicode, unicode_next,user_data) :
+def get_imgfont_path(font, img_src, length, unicode, unicode_next, offset_y, user_data) :
     if unicode < 0xf600:
         return
     if LV_USE_FFMPEG:
@@ -31,6 +31,7 @@ def get_imgfont_path(font, img_src, length, unicode, unicode_next,user_data) :
 imgfont = lv.imgfont_create(80, get_imgfont_path, None)
 if imgfont == None:
     print("imgfont init error")
+    return
 
 imgfont.fallback = LV_FONT_DEFAULT
 
--- a/examples/others/imgfont/lv_example_imgfont_1.py
+++ b/examples/others/imgfont/lv_example_imgfont_1.py
@@ -31,6 +31,7 @@ def get_imgfont_path(font, img_src, length, unicode, unicode_next, offset_y, use
 imgfont = lv.imgfont_create(80, get_imgfont_path, None)
 if imgfont == None:
     print("imgfont init error")
+    sys.exit(1)
 
 imgfont.fallback = LV_FONT_DEFAULT
 
--- a/examples/others/imgfont/lv_example_imgfont_1.py
+++ b/examples/others/imgfont/lv_example_imgfont_1.py
@@ -31,7 +31,6 @@ def get_imgfont_path(font, img_src, length, unicode, unicode_next, offset_y, use
 imgfont = lv.imgfont_create(80, get_imgfont_path, None)
 if imgfont == None:
     print("imgfont init error")
-    return
 
 imgfont.fallback = LV_FONT_DEFAULT
 
--- a/examples/others/imgfont/lv_example_imgfont_1.c
+++ b/examples/others/imgfont/lv_example_imgfont_1.c
@@ -5,28 +5,19 @@
 #if LV_USE_IMGFONT
 
 LV_IMG_DECLARE(emoji_F617)
-static bool get_imgfont_path(const lv_font_t * font, void * img_src,
-                             uint16_t len, uint32_t unicode, uint32_t unicode_next,
+static bool get_imgfont_path(lv_imgfont_param_t * param,
                              void * user_data)
 {
-    LV_UNUSED(font);
-    LV_UNUSED(unicode_next);
-    LV_UNUSED(user_data);
-    LV_ASSERT_NULL(img_src);
-
-    if(unicode < 0xF000) return false;
-
-    if(unicode == 0xF617) {
-        memcpy(img_src, &emoji_F617, sizeof(lv_img_dsc_t));
+    if(param->unicode == 0xF617) {
+        memcpy(param->img_src, &emoji_F617, sizeof(lv_img_dsc_t));
     }
     else {
-        char * path = (char *)img_src;
+        char * path = (char *)param->img_src;
 #if LV_USE_FFMPEG
-        snprintf(path, len, "%s/%04X.%s", "lvgl/examples/assets/emoji", unicode, "png");
+        lv_snprintf(path, param->len, "%s/%04X.png", "lvgl/examples/assets/emoji", param->unicode);
 #elif LV_USE_PNG
-        snprintf(path, len, "%s/%04X.%s", "A:lvgl/examples/assets/emoji", unicode, "png");
+        lv_snprintf(path, param->len, "%s/%04X.png", "A:lvgl/examples/assets/emoji", param->unicode);
 #endif
-        path[len - 1] = '\0';
     }
 
     return true;
@@ -40,13 +31,16 @@ void lv_example_imgfont_1(void)
     lv_font_t * imgfont = lv_imgfont_create(80, get_imgfont_path, NULL);
     if(imgfont == NULL) {
         LV_LOG_ERROR("imgfont init error");
+        return;
     }
 
-    imgfont->fallback = LV_FONT_DEFAULT;
+    static lv_font_t user_font;
+    user_font = *(LV_FONT_DEFAULT);
+    user_font.fallback = imgfont;
 
     lv_obj_t * label1 = lv_label_create(lv_scr_act());
     lv_label_set_text(label1, "12\uF600\uF617AB");
-    lv_obj_set_style_text_font(label1, imgfont, LV_PART_MAIN);
+    lv_obj_set_style_text_font(label1, &user_font, LV_PART_MAIN);
     lv_obj_center(label1);
 }
 #else
--- a/examples/others/imgfont/lv_example_imgfont_1.c
+++ b/examples/others/imgfont/lv_example_imgfont_1.c
@@ -5,18 +5,26 @@
 #if LV_USE_IMGFONT
 
 LV_IMG_DECLARE(emoji_F617)
-static bool get_imgfont_path(lv_imgfont_param_t * param,
-                             void * user_data)
+static bool get_imgfont_path(const lv_font_t * font, void * img_src,
+                             uint16_t len, uint32_t unicode, uint32_t unicode_next,
+                             lv_coord_t * offset_y, void * user_data)
 {
-    if(param->unicode == 0xF617) {
-        memcpy(param->img_src, &emoji_F617, sizeof(lv_img_dsc_t));
+    LV_UNUSED(font);
+    LV_UNUSED(unicode_next);
+    LV_UNUSED(offset_y);
+    LV_UNUSED(user_data);
+
+    LV_ASSERT_NULL(img_src);
+
+    if(unicode == 0xF617) {
+        memcpy(img_src, &emoji_F617, sizeof(lv_img_dsc_t));
     }
     else {
-        char * path = (char *)param->img_src;
+        char * path = (char *)img_src;
 #if LV_USE_FFMPEG
-        lv_snprintf(path, param->len, "%s/%04X.png", "lvgl/examples/assets/emoji", param->unicode);
+        lv_snprintf(path, len, "%s/%04X.png", "lvgl/examples/assets/emoji", unicode);
 #elif LV_USE_PNG
-        lv_snprintf(path, param->len, "%s/%04X.png", "A:lvgl/examples/assets/emoji", param->unicode);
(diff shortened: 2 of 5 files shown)


kisvegabor:
With the current implementation the imgfonts were siting on the baseline and it wasn't possible to correctly align img letters like 'y' or 'g', right?

If so the other option would be to allow filling the `lv_font_glyph_dsc_t` of the imgfonts too. 


With your solution, if we want to write a text like this:
image


All letters should have line_height height. Is that correct?

FASTSHIFT:
@kisvegabor 
This patch is mainly to solve the problem of this scene. I need to use emoji of the same size and fonts of similar height.

```c
void imgfont_test(void)
{
    lv_font_t * imgfont = lv_imgfont_create(32, get_imgfont_path, NULL);
    if(imgfont == NULL) {
        LV_LOG_ERROR("imgfont init error");
    }

    lv_obj_set_flex_flow(lv_scr_act(), LV_FLEX_FLOW_COLUMN);

    static lv_font_t my_font_1;
    static lv_font_t my_font_2;

    my_font_1 = lv_font_montserrat_30;
    my_font_2 = lv_font_montserrat_48;

    my_font_1.fallback = imgfont;
    my_font_2.fallback = imgfont;

    lv_obj_t * label1 = lv_label_create(lv_scr_act());
    lv_label_set_text(label1, "English text\ua060\ua061");
    lv_obj_set_style_text_font(label1, &my_font_1, LV_PART_MAIN);

    lv_obj_t * label2 = lv_label_create(lv_scr_act());
    lv_label_set_text(label2, "English text\ua060\ua061");
    lv_obj_set_style_text_font(label2, &my_font_2, LV_PART_MAIN);
}
```

Fix before:
image


Fix after:
image

FASTSHIFT:
> With the current implementation the imgfonts were siting on the baseline and it wasn't possible to correctly align img letters like 'y' or 'g', right?
> 
> If so the other option would be to allow filling the `lv_font_glyph_dsc_t` of the imgfonts too.
> 
> With your solution, if we want to write a text like this: image
> 
> All letters should have line_height height. Is that correct?

If imgfont is used as a fallback font, it should not need to care about `line_height`.
Is it possible to specify `g.ofs_y` to fit for alignment? Or just align the text when designing the image?

kisvegabor:
> Is it possible to specify g.ofs_y to fit for alignment? Or just align the text when designing the image?

If `.`, `a` and `y` have the same height (e.g. all are placed into a 20px tall rectangle), center align works well. However `.` has a 20 px image instead of 2 px. Letting the user to define `ofs_y` could solve the issue and would allow more control but it requires another (optional?) callback. However, once we have a callback we could allow setting other data too e.g. `adv_w`.

cc @guoweilkd

FASTSHIFT:
@kisvegabor 
Adding callbacks seems like a good solution.
But can we consider implementing a freetype-like imgfont manager, which has an independent cache, independent parameters for each glyph, and compresses the fonts in the same file (reducing frequent open close of the file system).

kisvegabor:
I see no problem with adding an a local cache here. We need to cache only:
- unicode
- width, height
- ofs_y
- img src?
- anything else?

FASTSHIFT:
> I see no problem with adding an a local cache here. We need to cache only:
> 
> * unicode
> * width, height
> * ofs_y
> * img src?
> * anything else?

Nothing else I can think of yet.

kisvegabor:
Great! 
Let's proceed step by step. I suggest adding the callback now and the cache later.

FASTSHIFT:
@kisvegabor 
Can merge the callback of the custom offset into `path_cb`?
This requires a redesign of the `lv_get_imgfont_path_cb_t` interface:

```c
typedef bool (*lv_get_imgfont_path_cb_t)(const lv_font_t * font, void * img_src,
                                         uint16_t len, uint32_t unicode, uint32_t unicode_next,
                                         void * user_data);
```

->

```c
typedef bool (*lv_imgfont_get_dsc_cb_t)(const lv_font_t * font, void * img_src,
                                        uint16_t len, uint32_t unicode, uint32_t unicode_next,
                                        int16_t * offset_x, int16_t * offset_y,
                                        void * user_data);
```

Or we unify the parameters into one structure:
```c
typedef strcut {
    const lv_font_t * font;
    void * img_src;
    uint16_t len;
    uint32_t unicode;
    uint32_t unicode_next;
    int16_t * offset_x;
    int16_t * offset_y;
    void * user_data;
} lv_imgfont_param_t;

typedef bool (*lv_imgfont_get_dsc_cb_t)(lv_imgfont_param_t * param);
```

kisvegabor:
I like the idea of extending the existing callback. Just one remark: I suggest making it clear somehow which are the output parameters.

```c
typedef strcut {
    const lv_font_t * font;
    uint32_t unicode;
    uint32_t unicode_next;
    int16_t * offset_x_out;
    int16_t * offset_y_out;
    void * img_src_out;
    uint16_t img_src_buf_len;
    void * user_data;
} lv_imgfont_param_t;
```

@amirgon  @uraich For MicroPython which would be the best? In short, besides the image itself we want to get one (or two) more parameters about the image font characters. We can either add the extra parameter(s) to the callback as pointers or add a struct to fill.

amirgon:
@kisvegabor I'm not sure I see the whole picture.  
I can see the `lv_imgfont_param_t` struct and the callback prototype `lv_imgfont_get_dsc_cb_t`, but could you explain how the callback registration looks like? (setting a `lv_imgfont_get_dsc_cb_t` function as a callback, how LVGL calls the callback later on, and how `user_data` is passed between them).

kisvegabor:
The user_data is saved here:
https://github.com/lvgl/lvgl/blob/613777f6233f3f2ed9e16afc4756b8b71d46529f/src/others/imgfont/lv_imgfont.c#L61

and the callback is called here
https://github.com/lvgl/lvgl/blob/613777f6233f3f2ed9e16afc4756b8b71d46529f/src/others/imgfont/lv_imgfont.c#L107

`lv_imgfont` is already MP compatible (we have a running example) so I believe there are no issues with the `user_data` part.

amirgon:
> `lv_imgfont` is already MP compatible (we have a running example) so I believe there are no issues with the `user_data` part.

But now you want to pass `lv_imgfont_param_t` instead, right?  
How is `lv_imgfont_param_t` registered? Is it allocated by the user? Or does LVGL creates a new instance of it every time and updates `user_data` like it does with `imgfont_dsc_t`?

kisvegabor:
I believe it will be a local variable whose only purpose is to pass the many parameters in a more compact form.

So LVGL can set the same `user_data` as the user passed to `lv_imgfont_create`.

amirgon:
> So LVGL can set the same `user_data` as the user passed to `lv_imgfont_create`.

I don't think this will work on Micropython.   
`user_data` should be set on callback registration. How do you register `lv_imgfont_get_dsc_cb_t`?

kisvegabor:
In the `create` function: 
```c
lv_font_t * lv_imgfont_create(uint16_t height, lv_get_imgfont_path_cb_t path_cb, void * user_data)
```

amirgon:
> In the `create` function:

So in the registration you pass `user_data` argument but the callback doesn't receive `user_data` argument, but a struct argument instead, with a `user_data` field.   

That's a mix of the two options for a callbacks (one is passing `user_data` as an argument on both registration and callback, the other is having `user_data` field in a struct that is passed as an argument).   

The two options weren't designed to mix. It might work, I'm not sure.

kisvegabor:
I see. In this case I think it'd be better to simply pass the new parameters as `int16_t * offset_x, int16_t * offset_y` and do not use `lv_imgfont_param_t`.

amirgon:
> I see. In this case I think it'd be better to simply pass the new parameters as `int16_t * offset_x, int16_t * offset_y` and do not use `lv_imgfont_param_t`.

You could also pass both `lv_imgfont_param_t` and`user_data` (as the last argument). And remove `user_data` from `lv_imgfont_param_t`.

FASTSHIFT:
It seems that the size of the imgfont cannot exceed the height of the text, otherwise it will be clipped.
image

kisvegabor:
> You could also pass both lv_imgfont_param_t anduser_data (as the last argument). And remove user_data from lv_imgfont_param_t.

I think we will be good with more straightforward  approach (i.e. have separate parameters)

> It seems that the size of the imgfont cannot exceed the height of the text, otherwise it will be clipped.


I think it's because a smaller `font->line_height` was set. How can I exactly reproduce it?

FASTSHIFT:
> > You could also pass both lv_imgfont_param_t anduser_data (as the last argument). And remove user_data from lv_imgfont_param_t.
> 
> I think we will be good with more straightforward approach (i.e. have separate parameters)
> 
> > It seems that the size of the imgfont cannot exceed the height of the text, otherwise it will be clipped.
> 
> I think it's because a smaller `font->line_height` was set. How can I exactly reproduce it?

Run `lv_example_imgfont_1()` to reproduce.

kisvegabor:
> Run lv_example_imgfont_1() to reproduce.

It works well here:
image

FASTSHIFT:
> > Run lv_example_imgfont_1() to reproduce.
> 
> It works well here: image

Did you use the patch of this PR? I modified the order of font fallback, because imgfont may be accessed less frequently than normal font, so it should be placed later.

kisvegabor:
I use this PR but swapping the 2 fonts really makes things different as the "main" font's  line height is considered. 

Maybe we can make `lv_font_get_line_height()` search the highest font among the fallbacks but I don't know how to align the base lines.

FASTSHIFT:
> I use this PR but swapping the 2 fonts really makes things different as the "main" font's line height is considered.
> 
> Maybe we can make `lv_font_get_line_height()` search the highest font among the fallbacks but I don't know how to align the base lines.

Can we ignore this situation? Users are advised not to use imgfont that exceeds the line height.

kisvegabor:
> Can we ignore this situation? 

Absolutely yes.

FASTSHIFT:
@kisvegabor @amirgon 
Any other questions? Can we merge this pr?

kisvegabor:
Please go with the "extra parameter for ofs_y" instead of the `param` struct because it's simpler for the MicroPython binding,

FASTSHIFT:
> Please go with the "extra parameter for ofs_y" instead of the `param` struct because it's simpler for the MicroPython binding,

done

kisvegabor:
Please update `examples/others/imgfont/lv_example_imgfont_1.py` too. It should be straight forward.

FASTSHIFT:
> Please update `examples/others/imgfont/lv_example_imgfont_1.py` too. It should be straight forward.

done

examples/others/imgfont/lv_example_imgfont_1.py
```diff
@@ -31,6 +31,7 @@ def get_imgfont_path(font, img_src, length, unicode, unicode_next,user_data) :
 imgfont = lv.imgfont_create(80, get_imgfont_path, None)
 if imgfont == None:
     print("imgfont init error")
+    return
```
kisvegabor:
```suggestion
```

FASTSHIFT: If imgfont creation fails, execution of the following code should be stopped.

kisvegabor: Maybe we want `sys.exit()`? 

@amirgon @uraich what is the best way to abort on error in MicroPython?

amirgon: > @amirgon @uraich what is the best way to abort on error in MicroPython?

`sys.exit()` should work.  
Or you can simply condition the few lines on the following code.

W-Mai: If we use `return`, it may keep the same behaviour with `C` example version.

kisvegabor: > If we use return, it may keep the same behaviour with C example version.

But this part is not inside of a function, so `return` is invalid.

W-Mai: Ooops, yes.  I am seeing things 🥲

And just `exit()` is ok.

W-Mai (changes requested):
```python
    exit()
```

examples/others/imgfont/lv_example_imgfont_1.py
```diff
@@ -31,7 +31,6 @@ def get_imgfont_path(font, img_src, length, unicode, unicode_next, offset_y, use
 imgfont = lv.imgfont_create(80, get_imgfont_path, None)
 if imgfont == None:
     print("imgfont init error")
```
W-Mai, line 33:
```suggestion
    print("imgfont init error")
    exit()
```

amirgon: "exit" is not the right way to exit a python script. 
You should import "sys" and use "sys.exit()"

W-Mai: well I read some article about this, surely `sys.exit` is more correct.
```suggestion
    print("imgfont init error")
    import sys
    sys.exit(1)
```
Is OK for this?
exit and raise an error code 1

amirgon: Imports are usually grouped together at the beginning of the file.

W-Mai:
```suggestion
    print("imgfont init error")
    sys.exit(1)
```

FASTSHIFT: done

kisvegabor:
Great! Merging, thanks!

PR #3671, opened 2022-09-07
corrected examples to run on Linux

uraich:
added micropython variants

### Checkpoints
The file explorer examples did not work on Linux because the path to the standard home folders were incorrect. The user's home directory is not /home, but /home/username. I get the username from the HOME environment variable and I append it the die driver letter, which was also missing. Like this the examples work for me.
I also added the MicroPython equivalents to the file_explorer folder.

### DIFF
--- a/examples/others/file_explorer/lv_example_file_explorer_1.c
+++ b/examples/others/file_explorer/lv_example_file_explorer_1.c
@@ -1,5 +1,9 @@
 
 #include "../../lv_examples.h"
+#include <stdlib.h>
+#include <unistd.h>
+#include <string.h>
+
 #if LV_USE_TABLE && LV_USE_FILE_EXPLORER && (LV_USE_FS_STDIO || LV_USE_FS_POSIX || LV_USE_FS_WIN32 || LV_USE_FS_FATFS) && LV_BUILD_EXAMPLES
 
 static void file_explorer_event_handler(lv_event_t * e)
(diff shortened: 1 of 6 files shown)


uraich:
I guess the failing test is something for Amir. As far as I can see 
#define LV_USE _FILE_EXPLORER 1
is missing in the lv_conf.h file that was used to create lv_micropython used for the test.

amirgon:
> As far as I can see
> #define LV_USE _FILE_EXPLORER 1
> is missing in the lv_conf.h

That's right.
I've fixed that now on latest lv_micropython.
Please try to re-run the tests (rebase or push some commit to trigger it)

uraich:
Hmmm, it still fails:
image

amirgon:
> Hmmm, it still fails:

Looks like a real failure, on lv_example_file_explorer_2.py line 15, in file_explorer_event_handler.
You are trying to concatenate a variable to a string, but the variable is `None`.

You can concatenate `str(x)` so it would work even if `x` is `None`.

uraich:
That is an interesting one! I never managed to crash the program with this error. I modified the code putting str() around the strings, as you suggested,  and the error while testing goes. I still think that this is an artifact generated by the test procedure.
I did find another error through the test however and corrected it.
I must admit that I do not really understand why in the C code the file name is put into a char array if the total length is smaller than LV_FILE_EXPLORER_PATH_MAX_LEN]  before being printed (line 16 in lv_example_file:explorer.c). In any case this makes no sense in Python, so I removed it.
I saw another odd detail: The root directory is A:/ If I select a file in the root directory, e.g. home I get A://home. This happens in both, the C and the MP version.
image
image

If I click the fast access button, I get the correct file name:
image

amirgon:
Uli are these examples supposed to run only on certain Linux distros? Or Windows?  
On my side, they don't display any files.
It could be useful to run them also on other platforms (generic Linux, ESP32, etc.).  

That's why I [suggested](https://github.com/lvgl/lvgl/pull/3641#issuecomment-1237467880) using a "dummy" FS driver which returns a predefined "faked" file system, which is independent of the real file system.

uraich:
If I understand things right, then file_explorer also passes through fs_drv. You must however connect to a physical file system in the end. In lv_config.h I enabled LV_USE_FS_STDIO and I set the drive letter to "A".
Can you try if this helps?
I cannot test the programs on Windows because I don't have a Windows system running lv_micropython. It should run on any Linux distro however. 
I have never tried to access the ESP32 file system with the fs_drv. The quick access buttons in the example make no sense on the ESP32.
This is what lv_example_file_explorer_1 should give you:
image

amirgon:
> You must however connect to a physical file system in the end

My suggestion was not to connect to the physical file system.
Instead, we could use a "dummy" FS driver which returns a predefined "faked" file system, which is independent of the real file system.
I think this can be useful for an example, instead of using the real file system.  
The problem with a real file system is that it looks different on every platform, so we can't run the example reliably on many platforms.

What do you think?

PR #3672, opened 2022-09-07
a few missing mp examples

uraich:
### Description of the feature or fix

A clear and concise description of what the bug or new feature is.

### Checkpoints
- [x] Follow the [styling guide](https://github.com/lvgl/lvgl/blob/master/docs/CODING_STYLE.md)
- [x] Run `code-format.py` from the `scripts` folder. [astyle](http://astyle.sourceforge.net/install.html) needs to be installed.
- [x] Update the documentation
Added quite a few of the missing micropython examples

PR #3677, opened 2022-09-08
Mis ex problems

uraich:

### Checkpoints
Missing MP examples:
fragment: not MP compatible
monkey: is it needed?
lv_example_event_4 : I don't know how to pass the array of points to the polygon drawing function
imgfont: segfaults
Here is a draft pull request with the programs as they stand. They do not work though.


kisvegabor:
> fragment: not MP compatible

Could you remind me what was the conclusion about it? Are we planning to fix it to be MP compatible? If so @amirgon's checklist about MP compatibility could be be tested with it.

> monkey: is it needed?

Not critical at all. 

> lv_example_event_4 : I don't know how to pass the array of points to the polygon drawing function

This example was added as part of a larger PR and I wasn't noticed it. I think something simpler would be better here. So I simplified it in 451eddb12

uraich:
I also have the keyboard and encoder demo working. For the moment only on SDL and therefore only with the keyboard. I would like to try this on my ESP32 not which I can easily connect an encoder.

amirgon:
> Could you remind me what was the conclusion about it? Are we planning to fix it to be MP compatible? If so @amirgon's checklist about MP compatibility could be be tested with it.

@kisvegabor `LV_USE_FRAGMENT` is currently set to `0` on lv_micropython `lv_conf.h`.  
Main problem there is with callbacks, maybe also other problems.
Further discussion: https://github.com/lvgl/lvgl/issues/3535#issuecomment-1232109792 .

Currently I'm not planning to fix the Fragment feature on LVGL. If someone attempts to fix it I can help by reviewing the PR.

> lv_example_event_4 : I don't know how to pass the array of points to the polygon drawing function

Uli did you try to simply pass a Python list of `lv.point_t` objects?

> I also have the keyboard and encoder demo working. For the moment only on SDL and therefore only with the keyboard. I would like to try this on my ESP32 not which I can easily connect an encoder.

Nice!

amirgon:
Uli, what is the relation between this PR and https://github.com/lvgl/lvgl/pull/3672?  
Why not maintain a single PR for fixing/adding Micropython examples on LVGL?

kisvegabor:
> LV_USE_FRAGMENT is currently set to 0 on lv_micropython lv_conf.h.
Main problem there is with callbacks, maybe also other problems.
Further discussion: https://github.com/lvgl/lvgl/issues/3535#issuecomment-1232109792 .

Thank. Let's leave it for an other pass.

uraich:
> Uli, what is the relation between this PR and #3672?
Maybe I did something wrong? My intention was to just upload  lv_example_imgfont_1.py, which still produces a seg fault. Therefore I made it a draft PR. Once lv_example_imgfont_1 works we can close this one and integrate it into #3672

amirgon:
> My intention was to just upload lv_example_imgfont_1.py, which still produces a seg fault.

Got it.

When running `lv_example_imgfont_1.py` the error I'm getting is 
```
Could not find emoji_F617.png
```

And it's really missing.  
There's `img_emoji_F617.c` but it doesn't help in the Micropython case.

uraich:
Oups, I forgot to add the png file. Did an upload just now. Unfortunately it also segfaults with this file

amirgon:
> Oups, I forgot to add the png file. Did an upload just now. Unfortunately it also segfaults with this file

Did you add it on the wrong PR? (https://github.com/lvgl/lvgl/pull/3672)

uraich:
yes, sorry. But it needs to be in the other one as well, at least as soon as we get the imgfonr to work

amirgon:
> My intention was to just upload lv_example_imgfont_1.py, which still produces a seg fault.

Hi Uli,
I fixed the problem, it was a bug in the binding script related to `user_data` on callbacks. (https://github.com/lvgl/lv_binding_micropython/commit/1be17fca5825be0b04cfeffa3fe74c726418eca1)  

On latest lv_micropython, `lv_example_imgfont_1.py` now stops at a different error (missing user_data argument on `get_imgfont_path` - the callback should receive it although should probably not use it directly)

uraich:
Thanks Amir. I gave it a try and it works slightly better now:
image
It does not display the images though and I believe that the reason for this is parameter passing: in get_imgfont_path the img_src is passed by reference (a void *) and it expects the image source to be returned in this parameter. I don't think I can do this in Python, or am I wrong again?

amirgon:
> It does not display the images though and I believe that the reason for this is parameter passing: in get_imgfont_path the img_src is passed by reference (a void *) and it expects the image source to be returned in this parameter. I don't think I can do this in Python, or am I wrong again?

Actually, you can.  
Use the `__dereference__` function of `img_src` to access a "memoryview" of the data. 

Here is an example: https://github.com/lvgl/lv_binding_micropython/blob/e225c4be2ccb428ecb73e7cf83f6e15c675b91e6/lib/fs_driver.py#L40

uraich:
Wow!
image
In fact I did not use the image defined in C code (img_emoji_F617.c) as it is done with memcopy in the C example but I recreated a png image from the C code with the tools I have developed. It is certainly  possible to convert the C code of emoji_F617.c into Python as well, but this is quite a bit of work and I don't really think it is worth it. emoji_F617.c is created automatically from a png file but I don't know of a tool that does the same thing for Python code.
I have a slight problem though:. This is what happens when I run the CI test on it:
image

kisvegabor:
Closing as requested [here](https://github.com/lvgl/lvgl/issues/3535#issuecomment-1252889945).

Issue #3664, opened 2022-09-06
Lightweight TTF LVGL Integration


codewitch-honey-crisis:
### Integrating my lightweight TTF/OTF orchestration into LVGL

I wrote [htcw_gfx](https://honeythecodewitch.com/gfx) in part because LVGL's TTF support via Freetype didn't fit my needs - it had to be compiled on a linux machine (at least at the time i looked into it) and it basically needed an ESP32 or better to run realistically, with restrictions on the size of the font file size. All of this is me spitballing from memory so forgive me if I got some of the details incorrect regarding Freetype.

I supported truetype by adapting this public domain code: https://github.com/nothings/stb/blob/master/stb_truetype.h and changing it around so it could stream from storage instead of requiring the font to be loaded into RAM. It also assumed gobs of memory was available in other ways, like in its internal custom heap, so I modified that as well.

I'd like to integrate it into LVGL, and I started poking around on the forums which led me here:

https://docs.lvgl.io/master/overview/font.html#add-a-new-font-engine

Update!

Anyway, after hacking around I figured things out mostly but I'm running into a design issue and I'd like some feedback on how to handle it in the most LVGLish way.

You can load a font with my engine, and then render it at any given scale.

However, there's no facility for rendering fonts at different scales within LVGL is there? It would have to be a new lv_font_t ? or whatever, for each different scale right?

If that's the case, is there a way you'd recommend I do this?

To be clear, this is how it sort of needs to work to work with my engine:

To load a font, you must pass a file. I even made a const buffer "file system" so you could stream from a header file with a flash embedded array, but to the load the font you must pass it a file containing a TTF or OTF font.

Once you have that, you can spin off fonts with different scales, somehow. That somehow is the big question. I can think of different ways to structure this, but does anyone have any suggestions?

The other thing is, I have a question about memory allocation. A lot of the LVGL code returns pointers, which is troubling to me. For starters it means whipping the heap, and also it means what about PSRAM? what about DMA capable ram that must be specially allocated? Is there a facility for custom allocators in LVGL, and a memory allocation scheme i should be using that goes with it?

kisvegabor:
> Does LVGL include any kind of stream interface I should be using? What is the best practice in a cross platform manner to connect my engine to a TTF file or program flash?

We have a [file system interface](https://docs.lvgl.io/master/overview/file-system.html)  with optional modules for some exiting file APIs. There is no flash interface in it but I can easily imagine a simple one. 

[...]

Issue #206, opened 2018-04-13
Redraw an an object 


Moo2017:
I that any way to redraw an object after being created? Assume that I create a keyboard, then  once user type the text and insert Enter key, a message box including two option "Edit" and "Save" pops up, if Edit is pressed user is returned to the keyboard , Is that any way to redraw the keyboard?

kisvegabor:
If you mean by "Enter" the "Ok" button then yes. Clicking the "Ok" or "Hide" button will delete the keyboard by default but if you assign an `ok_action` then you can control what to happen on click. You can hide the keyboard or animate it out.

Use: `lv_kb_set_ok_action()`

Moo2017:
@kisvegabor Oh no I mean assume that I create an object which has common area with the previous object. Once I delete the second object , the common area will be cleared so I have to redraw the previous one. I was looking for a way not to delete and create the previous object.

Moo2017:
@kisvegabor I found that! I should focus on the previous object only right?! 
what happened if I focus on an object that I haven't add to the group?

kisvegabor:
The managing of redrawing should be done by the library automatically. Please send an example code to see what happens.

kisvegabor:
It seems the issue is solved so I close it. 
If you have any remarks in this topic feel free to reopen it!

PR #3680, opened 2022-09-10
added cbfs and ttf support

codewitch-honey-crisis:
### Lightweight TTF and a const buffer filesystem

Added lightweight truetype support over files under libs/lv_ttf.h
Added a virtual filesystem over a const buffer array to allow embedding files as headers. Under lib/fsdrv/lv_fs_cbfs.h

### Checkpoints
- [x] the truetype engine needs cleanup. It was ported from a public domain source, and it hasn't been sanitized.
- [x] `code-format.py` is not available. link doesn't work
- [x] I'm gonna need some help regarding updating the docs


src/libs/fsdrv/lv_fs_cbfs.c
```diff
@@ -0,0 +1,202 @@
...
+        return;
+    }
+    int i;
+    #if defined(LV_FS_CBFS_LETTER) && LV_FS_CBFS_LETTER >= 'A' && LV_FS_CBFS_LETTER >= 'Z'
+    i= LV_FS_CBFS_LETTER;
+    #else
+    char drives[27];
+    size_t drive_count = strlen(drives);
+    lv_fs_get_letters(drives);
+    for(i = 'A';i<='Z';++i) {
```
kisvegabor: Auto assign is not needed. We expect that users can set a letter.

src/libs/fsdrv/lv_fs_cbfs.c
```diff
@@ -0,0 +1,202 @@
...
+}
+lv_fs_res_t lv_fs_cbfs_create_file(char* out_path,size_t out_path_size, const void* data, size_t data_size) {
+    if(drive==0) {
+        return LV_FS_RES_NOT_EX;
+    }
+    if(out_path==NULL||data==NULL) {
+        return LV_FS_RES_INV_PARAM;
+    }
+    snprintf(out_path,out_path_size,"%c:/0x%lx-%li",drive,(long)data,(long)data_size );
+    return LV_FS_RES_INV_PARAM;
```
kisvegabor:
```suggestion
    return LV_FS_RES_OK;
```

src/libs/fsdrv/lv_fs_cbfs.c
```diff
@@ -0,0 +1,202 @@
...
+        }
+    }
+    if(i>'Z') {
+        LV_LOG_ERROR("CBFS found no more drives available");
+        return;
+    }
+    #endif
+    drive=(char)i;
+    
+    lv_fs_drv_t* drv=(lv_fs_drv_t*)lv_mem_alloc(sizeof(lv_fs_drv_t));
```
kisvegabor:
```suggestion
   static lv_fs_drv_t drv;
```

src/libs/ttf/lv_ttf.h
```diff
@@ -0,0 +1,48 @@
...
+} lv_ttf_t;
+struct stbtt_fontinfo;
+enum {
+    LV_TTF_RES_OK = 0,
+    LV_TTF_RES_ARG = 1, /* Indicates an invalid argument*/
+    LV_TTF_RES_FMT = 2, /*Typically indicates that the font format is invalid or unsupported*/
+    LV_TTF_RES_MEM = 3, /* Out of memory*/
+};
+typedef uint8_t lv_ttf_res_t;
+lv_ttf_res_t lv_ttf_open(lv_ttf_t* out_ttf, lv_fs_file_t* file);
```
kisvegabor:
```suggestion
lv_ttf_res_t lv_ttf_open(lv_ttf_t* out_ttf, const char * path);
```

src/libs/ttf/lv_ttf.h
```diff
@@ -0,0 +1,48 @@
...
+    lv_fs_file_t *stream;
+    size_t offset;
+    int cursor;
+    int size;
+} lv_ttf_buf_t;
+
+typedef struct lv_ttf {
+    void *userdata;
+    //unsigned char *data; // pointer to .ttf file
+    lv_fs_file_t *stream; // stream of .ttf file
```
kisvegabor:
```suggestion
    lv_fs_file_t f; 
```

Open the file in `lv_ttf_open` to `f`.

src/libs/ttf/lv_ttf.c
```diff
@@ -0,0 +1,4512 @@
...
+    lv_fs_read(stream, &result, 1, &br);
+    return result;
+}
+static stbtt_int8 read_int8(lv_fs_file_t *stream) {
+    int8_t result;
+    uint32_t br;
+    lv_fs_read(stream, (uint8_t *)&result, 1, &br);
+    return result;
+}
+// TODO: Support big endian procs
```
kisvegabor: Not critical, just FYI LVGL has the `LV_BIG_ENDIAN_SYSTEM` config.

src/libs/ttf/lv_ttf.c
```diff
@@ -0,0 +1,4512 @@
...
+   return p[0] * 256 + p[1];
+}*/
+// static stbtt_int16 ttSHORT(stbtt_uint8 *p) { return p[0] * 256 + p[1]; }
+// static stbtt_uint32 ttULONG(stbtt_uint8 *p) { return (p[0] << 24) + (p[1] << 16) + (p[2] << 8) + p[3]; }
+// static stbtt_int32 ttLONG(stbtt_uint8 *p) { return (p[0] << 24) + (p[1] << 16) + (p[2] << 8) + p[3]; }
+
+#define stbtt_tag4(p, c0, c1, c2, c3) ((p)[0] == (c0) && (p)[1] == (c1) && (p)[2] == (c2) && (p)[3] == (c3))
+#define stbtt_tag(p, str) stbtt_tag4(p, str[0], str[1], str[2], str[3])
+
+/*static int stbtt__isfont(stbtt_uint8 *font)
```
kisvegabor: Can we remove the many commented out code?

kisvegabor (commented):
Thank you! I added some comment.

Having basic [docs](https://github.com/lvgl/lvgl/tree/master/docs/libs) and some [examples](https://github.com/lvgl/lvgl/tree/master/examples/libs) would be awesome. I can take care of them if you wish.

kisvegabor:
I'll try to fix the CI issues and some others tomorrow.

src/libs/fsdrv/lv_fs_cbfs.c
```diff
@@ -0,0 +1,233 @@
...
+    ++sz;
+    if(*sz != 'x' && *sz != 'X') {
+        return NULL;
+    }
+    ++sz;
+    const char * sz2 = strchr(sz, '-');
+    if(sz2 == NULL || sz2 == sz + 1) {
+        return NULL;
+    }
+    void * p = (void *)strtol(sz, NULL, 16);
```
embeddedt: We will need to add an `lv_strtol` or tweak this function, as not every platform has a `strtol` (e.g. MicroPython on STM32). I'm still working on understanding what this does. The address and size are stored in the file path?

kisvegabor: > We will need to add an lv_strtol or tweak this function

We can add it in `lv_utils.c`.


> The address and size are stored in the file path?

I believe it's based on my idea [here](https://github.com/lvgl/lvgl/issues/3664#issuecomment-1242707330).

embeddedt: Thank you, I was only skimming the emails and missed it.

src/libs/fsdrv/lv_fs_cbfs.c
codewitch-honey-crisis: but where do we initialize the letter? I put a constant for it you can use in lv_conf_template.h just FYI. If you do assign one, then a drive letter is not automatically initialized. The reason I did it was because I need it once on init to produce a drv instance. I guess I noticed after the fact that you require defines for that. That code can safely be eliminated (between the ifdefs) and the alternative used, requiring the constant or returning NULL (or whatever the error condition is i don't have the code in front of me)

Edit: I wanted to throw in some docs. I do have an example but it's not checked in anywhere because I wasn't sure exactly where I should put it, and set it up. I also meant to ask about that but it got lost in my shuffle. Can we hold of documenting until we get the API finalized? See comments below.

src/libs/fsdrv/lv_fs_cbfs.c
codewitch-honey-crisis: that's what I get for coding on 4 hours of sleep! thank you

Edit: I didn't mean to resolve this. I meant to commit your suggestion

src/libs/ttf/lv_ttf.c
codewitch-honey-crisis: absolutely. The reason I left it in, honestly, is because I've removed support for certain formats and rendering modes to save flash space, and I wasn't sure how many font files that would impact. I left that code in there so I could put like, OS/2 TTF and SVG support back in if necessary. It's fine to remove it. I thought about doing it, but I've been on fence about adding this stuff back in.

src/libs/ttf/lv_ttf.c
codewitch-honey-crisis: I meant to ask about that (hence the TODO). I had one in htcw_gfx as well.

src/libs/ttf/lv_ttf.h
codewitch-honey-crisis: That concerns me. Keep in mind, if you do that you are making the TTF engine responsible for the lifetime of the file (unless I am wrong)

If you're okay with that, that's fine, but it requires closing and opening the file somewhere in the code, and not when you create the font, because several lv_font_t instances can be created from a single open file.

src/libs/ttf/lv_ttf.h
codewitch-honey-crisis: See my comment about file lifetime management. If we put it inside the TTF engine, where does it go? You can't tie a file to a single font because you can spin off multiple fonts from a single file at different scales. Ergo, you need to manage the lifetime of the file outside of the fonts so that you don't have to reference count or something equally messy.

src/libs/fsdrv/lv_fs_cbfs.c
codewitch-honey-crisis: Yes, it was absolutely based on kisvegabor's idea. Some of the code could be commented on reflection. I parse so much it's instinct but I forget commenting it for people who don't!

Edit: I can write a strtol function, not a problem. I just didn't know it wasn't available on some platforms. I've hardly ever used it to be honest.

Edit 2: I'll change the function and provide it to you. It's not a problem to parse those and probably isn't worth making a general purpose utility function for it. I don't want to check in and create a bunch of conflicts, so I think I'll just stick it in the comments here?

kisvegabor:
The lifetime management of a files/fonts affects multiple comments so I comment here. 

Seemingly all we need is `lv_ttf_create_font(&font, "path", 40, NULL);`. It can `malloc` an `lv_ttf_t`, saves it in `font` and open the file. The same file can be used by multiple fonts as the same file can be opened multiple times simultaneously. So it should work:
```c
lv_ttf_create_font(&font1, "path", 30, NULL);
lv_ttf_create_font(&font2, "path", 40, NULL);
``` 

This way the user only needs to manage the lifetime of `font`. When it's not required anymore, the font can be destroyed with `lv_ttf_destroy_font(&font)` which closes the file and `free`s the `lv_ttf_t` structure.

Alternatively 
```c
lv_font_t * lv_ttf_create_font("path", 30);
```
could work too, where `lv_font_t` is also `malloc`ed inside  `lv_ttf_create_font`. And we can have a generic `lv_font_set_fallback(&font1, &font_fallback)` function.

BTW, does your engine support "font styles", e.g. bold, italic, etc?

One more note: `lv_ttf` seems a little bit too generic. Do you have a good name in mind? E.g. `lv_htcwttf`.

codewitch-honey-crisis:
> The same file can be used by multiple fonts as the same file can be opened multiple times simultaneously. 

That's a good point. 

> BTW, does your engine support "font styles", e.g. bold, italic, etc?

I wish it did. I don't know how honestly, I haven't looked into it. What I've done in the past is use fontsquirrel.com and find fonts that were already in the style that I needed.

> lv_font_t * lv_ttf_create_font("path", 30);

This is my preference because it's how you all do it with jpgs and how you do it with other fonts.

One thing I would suggest though (because you know how I am by now ha!) is when you malloc the font, maybe do one malloc and pad the end of it so you can copy the font user data contents in after the font itself. That way you're not doing to tiny mallocs instead of one less tiny malloc, but that's what I would do because I'm absolutely paranoid about heap use on anything with less than 1MB of SRAM. =) 

One thing is, I'm totally willing to help here as I'm able. I just need to get a build environment set up which is a work in progress. I've been at it all day, and I'm ready to take a break, but I want to help.

What I'm doing right now is removing the strtol dependency and then going over some of the other issues so I can have code to provide, but I'm not sure exactly where to provide it with this pull request being up in the air (forgive me, I have limited knowledge of collaborating using github).

codewitch-honey-crisis:
Here, I removed the strtol dependency. Not sure how to check this in without creating conflicts
```cpp
static void* open_cb(lv_fs_drv_t* drv,const char* path, lv_fs_mode_t mode) {
    if(mode!=LV_FS_MODE_RD) {
        return NULL;
    }
    const char* sz = path;
    if(*sz!='/' && *sz!='\\') {
        return NULL;
    }
    ++sz;
    if(*sz!='0') {
        return NULL;
    }
    ++sz;
    if(*sz!='x' && *sz!='X') {
        return NULL;
    }
    ++sz;
    long l = 0;
    char ch=*sz;
    while((ch>='0'&&ch<='9')||(ch>='A'&&ch<='F')||(ch>='a'&&ch<='f')) {
        int i;
        if(ch<='9') {
            i = ch-'0';
        } else if(ch<='F') {
            i=ch-'A'+0xA;
        } else {
            i=ch-'a'+0xa;
        }
        l*=0x10;
        l+=i;
        ++sz;
        ch=*sz;
    }
    if(l==0 || ch!='-') {
        return NULL;
    }
    void* p = (void*)l;
    ++sz;
    l=0;
    ch=*sz;
    while(ch>='0'&&ch<='9') {
        int i = ch-'0';
        l*=10;
        l+=i;
        ++sz;
        ch=*sz;
    }
    size_t size = (size_t)l;
    cbfs_handle_t* h = (cbfs_handle_t*)lv_mem_alloc(sizeof(cbfs_handle_t));
    h->data = p;
    h->size = size;
    h->position = 0;
    return (void*)h;
}
```

codewitch-honey-crisis:
I think i have my build environment prepared. I've updated the code with the suggestions (unless I forgot some) and I'm working on building the source tree on the linux vm and formatting the code. Should I issue a new pull request when I check in? Edit: Never mind. I didn't realize once I established a pull request that it would continue to pull from my fork rather than taking a snapshot. Like I said before, collaborating on github is something i'm pretty new at.

codewitch-honey-crisis:
I had to force that because VS code horked my merge. Anyway, I've cleaned up a lot. I've added placeholder files for the example but there are a lot of unknowns in terms of me baking those and the tests, like how do I go about making sure LV_USE_TINY_TTF is enabled for the examples, and other things. Same with the tests. I'll eventually figure it out on my own, but help is always appreciated.

kisvegabor:
Something strange happened. GitHub say ~all files have been changed:
image

Maybe line ending issue?

embeddedt:
> Maybe line ending issue?

Yes, based on the CI error:

```
/home/runner/work/_temp/5745ff82-f8dd-47cf-8504-28caaa8270d7.sh: scripts/install-prerequisites.sh: /usr/bin/sh^M: bad interpreter: No such file or directory
```

This is possibly also the reason that @codewitch-honey-crisis was unable to run the code formatting or `lv_conf_internal.h` generator scripts.

I seem to have been able to fix it by using `git add --renormalize .` and commiting, but I am on Linux. Possibly another command is needed on Windows to checkout using Unix line endings?

codewitch-honey-crisis:
I am on linux now.

The files probably changed because I was an idiot:
1. I messed up my local resolution of merge conflicts, and finally gave up on trying to merge properly and did a git push --force
because...
2. I didn't realize with github once you submit a PR it connects the repos. I thought it took a snapshot.

I'm sorry! It won't happen again. I know better now.

codewitch-honey-crisis:
I did a pull. I am on linux. Should I build, format and try to push to see if it works now?

Edit: May be related to this: https://www.reddit.com/r/bashonubuntuonwindows/comments/ee2s68/vscode_wsl_git_line_endings_crlflf_i_cant_get_it/

embeddedt:
Yes please. (It might be better to run `scripts/lv_conf_internal_gen.py` as the formatting is apparently already fine.)

codewitch-honey-crisis:
i just ran the lv_conf_internal_gen script and pushed.

embeddedt:
It looks like that's worked for the most part! Can you remove `lv_conf copy.h` and the two `.ipynb` files?

codewitch-honey-crisis:
I'm not finding lv_conf copy.h
Where in the source tree is it?

embeddedt:
It looks like you've checked in both an `lv_conf.h` and `lv_conf copy.h` under the `src` folder.

codewitch-honey-crisis:
Yeah i just found it now and am pushing rn

embeddedt:
There shouldn't be an `lv_conf.h` either.

codewitch-honey-crisis:
One failure

Error: /home/runner/work/lvgl/lvgl/ports/esp32/build-GENERIC/lv_mp.c:34056:95: error: 'lv_fs_cbfs_init' undeclared here (not in a function); did you mean 'lv_fs_drv_init'?

I'm not declaring lv_fs_cbfs_init() in lv_fs_cbfs.h
It's declared in lv_fsdrv.h like the others.

Is there something I'm doing wrong there?

codewitch-honey-crisis:
I'm not sure why this is failing and just for Micropython on the ESP32. I know nothing about MP, much less how you generate the source file it's failing in. I'm out of moves. :(

That being said, I'm also working on a backport of the latest version of the STB truetype engine I ported before. Seems they've improved the rasterizer some, and fixed some other issues. This time I'm going to bring in all the functions of the original library. I've got it working, albeit with a significant rendering bug I'm running down. (As an aside I'm trying to get my backport adopted by STB so I don't have to backport future revisions)

My point is I'll have a significant check in, probably soon but I don't want to mess with this until the ESP32 failure is resolved, so I'm pretty much blocked once I get this bug fixed locally.

embeddedt:
Hmm, I wonder if the generator is choking on `lv_fs_cbfs_init()` not having `void` in the parameter area.

@amirgon Do you see any other callback conventions that might be missed in `lv_fs_cbfs.c` that could be causing this issue?

codewitch-honey-crisis:
ooh, i didn't see that. i'll try to update now. I'm so used to not putting them in (habits!). thanks!

codewitch-honey-crisis:
could someone approve these workflows? thanks!

kisvegabor:
I've modified the approval setting to this:
image

codewitch-honey-crisis:
i give up. I have no idea why it's failing that check. I love the irony. The ESP32 is the machine I test on.

codewitch-honey-crisis:
well, this pull is blocked on someone else figuring out why that check fails. I'll be improving the truetype engine, but with an eye towards supporting htcw_gfx with it until this is resolved.

embeddedt:
Honestly, I would carry on and not worry about the check; the other MicroPython platforms are passing and we can always fix it at the end.

codewitch-honey-crisis:
okay. Well fair enough. So I'll just consider this "passed" with qualifiers I guess.

codewitch-honey-crisis:
This latest checkin follows I plan I have to reintegrate the STB truetype engine directly into LVGL without having to port it first. The idea is I add streaming support to STB's offering https://github.com/nothings/stb/blob/master/stb_truetype.h

and then submit it back to them as a pull request, which I've already proposed. If they okay it, I will integrate my LVGL friendly changes back into STB's source and we can pull the header directly from that repo whenever they update the engine.

Next I'll be exposing some of these features. First and foremost I want to try to tease styles out of it if possible. I honestly have no idea how that works, aside from some constants I've seen. Won't stop me from diving in though. I may need to understand the lv_style_t stuff better than I do? Before I go off half cocked.

codewitch-honey-crisis:
Some followup regarding bold/italic/underline/etc

Apparently those do NOT exist in the font file, and there is no standard tweaks you can make to the rasterization to make them happen, although *some* software can do this. Adobe specifically designed TTF such that you need a different file for the italic versions because the glyphs are so different, and then there's hinting and such which tweaking in software would dump all over. (My code doesn't support hinting, but I'm speaking generally as to Adobe's design decision)

So like, yes, it is possible to tweak a font's render to make it bold. Or to make it italic, or to make it underline, etc. It is not the recommended approach, and is non-trivial besides, and wouldn't work with some or a lot of fonts - italic and bold particularly.

Adobe has TTC files, which are collections of fonts, and in them fonts can be tagged with face names, and flagged with styles, but those are identifiers for particular fonts within the collection, kind of like meta tags to identify a particular font.

The STB library I've ported supports enumerating these fonts from these files, so in theory I could support *that* but I've never seen a TTC in the wild, though I haven't looked that hard. I imagine they aren't used much in software design, probably more in publishing, if anything. So I'm for want of testing material as well as a driving motivator to expose the functionality considering how rarely it would be used.

As I may have mentioned? My current route is to download say "Ariel Bold" from fontsquirrel.com and simply use that, as opposed to trying to make "Ariel" bold, if that makes sense.

codewitch-honey-crisis:
I should note that the STB implementation I ported does support subpixel rendering. I also noticed that LVGL has a uint8_t value for  it you can fill as part of the glyph dsc callback, IIRC. I don't know it works in LVGL. How do I use the subpx field or whatever it's called? With STB I just adjust each render with a fractional shift value usually by keeping a floating x position and then shifting by xpos-floor(xpos) like that.

codewitch-honey-crisis:
Okay, I'm asking before I do this, on the chance I end up blowing something up because I make assumptions about how things work with github:

I found a couple of areas in the STB header that need cleanup. There are some functions that were prototyped but never implemented, and they should be removed.

Before I do that my plan is to pull, change, and then push the changes.

I'm not going to mess anything up on your end am I?

Also I was wrong before about where to fill the subpx field. Apparently it's set when filling lv_font_t.
What does this field do? More specifically, how does it change the behavior of the callbacks?
I ask because again, the STB code supports subpixel rendering but I can't do it due to not having fractional x advances.

embeddedt:
> Before I do that my plan is to pull, change, and then push the changes. I'm not going to mess anything up on your end am I?

Nothing should be broken by that. Also, as long as you don't force-push, we can easily revert the changes if something goes wrong.

codewitch-honey-crisis:
hmmm. i formatted the code. let me try one more time.

codewitch-honey-crisis:
okay I don't understand how that happened. I sort of understand why, but code that isn't mine is breaking due to a formatting issue. I'm not clear on how I managed to pull something that was broken (I pulled before I pushed), unless kisvegabor's force stuck failing code in. Please advise.

amirgon:
Micropython binding (non-esp32) was failing because of a bug. (Fixed now).
See: https://github.com/lvgl/lvgl/pull/3641#issuecomment-1245942184

This, however:

```
Error: /home/runner/work/lvgl/lvgl/ports/esp32/build-GENERIC/lv_mp.c:34040:95: error: 'lv_fs_cbfs_init' undeclared here (not in a function); did you mean 'lv_fs_drv_init'?
```

Is related to something else.  
On lv_micropython `LV_USE_FS_CBFS` is not defined on `lv_conf.h`.   
I think somehow `lv_conf_internal.h` is not included so the build chowks on this:

https://github.com/lvgl/lvgl/blob/4a47414edef4955f53849e32cca5e9e692813263/src/libs/fsdrv/lv_fsdrv.h#L46-L48

When `LV_USE_FS_CBFS` is undefined, `#if LV_USE_FS_CBFS != '\0'` behavior is undefined.  
I'm not sure why `lv_conf_internal.h` is not included there but you can probably work around this by wrapping that in `#ifdef LV_USE_FS_CBFS`

src/libs/fsdrv/lv_fs_cbfs.h
```diff
@@ -0,0 +1,12 @@
+#ifndef LV_FS_CBFS_H
+#define LV_FS_CBFS_H
+#include "../../../lvgl.h"
+#include <stddef.h>
+#ifdef __cplusplus
+extern "C" {
+#endif
+lv_fs_res_t lv_fs_cbfs_create(char * out_path, size_t out_path_size, const void * data, size_t data_size);
```
amirgon: Should probably by `const char * out_path`

codewitch-honey-crisis:
@amirgon it's a path to be filled, hence it can't be const, and that's why it's indicated as out_xxx (a convention I use, but I'm not sure anyone else does but I guess I wish they did?)

codewitch-honey-crisis:
@amirgon can you clarify what I should be wrapping with the #ifdef? I've already done this in lv_fsdrv.h:
 ```cpp
// #ifdef is workaround for micropython binding generator
#ifdef LV_USE_FS_CBFS
#if LV_USE_FS_CBFS != '\0'
void lv_fs_cbfs_init(void);
#endif
#endif
```
Should I put this in the cbfs source file as well? Forgive me, I am totally lost as to the Micropython end of this.

**Comment by amirgon** [comment]
@codewitch-honey-crisis I'm not sure I understand this MP esp32 failure.
Could you try something? Try to include `lv_conf_internal.h` from `lvgl.h`.  

`lvgl.h` is included from both the LVGL source and the generated Micropython bindings, so including it there would ensure it's included everywhere. If the problem persists then it's probably not related to `lv_conf_internal.h` as I thought.

**Comment by codewitch-honey-crisis** [comment]
@amirgon I spotted something that might be causing the failure, and I'm going to try that first. Then I will try what you suggested, since i'm already in progress on my try.

@kisvegabor I added some docs - modifying the fs_drv.md and adding tiny_ttf.md. I still need to add examples and link to them in the docs, but it looked like the gif one (which I copied from) had a rust example? did i see that right? anyway, more importantly, I'm a little lost on the examples so I'm gonna need a minute to feel my away around in the dark over here. =) Then I will add something.

**Comment by codewitch-honey-crisis** [comment]
well that didn't work out. I'm including the lv_conf_internal header like you suggested. let's hope.

**Comment by codewitch-honey-crisis** [comment]
I have half a mind to remove the lv_cbfs_init() function altogether. I could lazy init on first use, not a problem, but I fear the inconsistency in the API, being this FS would be the only one without an initialization function, and I can't really justify the reason. I'd get skewered in review if I said I lazy inited to workaround a bug in micropython binding generation! :)

**Comment by codewitch-honey-crisis** [comment]
You know, technically, Tiny TTF does not need CBFS. The reason I included it was

A) In order to enable a library that expects a file to work from a header with an embedded array
becase
B) It's not often realistic performance-wise to do TTF from a filesystem (heavily depends on what sort of filesystem, interconnect, etc)

What I do in my real world projects, at least on ESP32s is I embed as a header. If it's a WROVER with PSRAM, and I have a TTF file (SPIFFS or SD) I will first load it into PSRAM then work off of that.

Now, all of this suggests I could make Tiny_TTF work on just an array of bytes, and that would cover a lot of cases.

And you'd be right. But it wouldn't cover the fact that you simply don't have enough RAM on most devices to use a TTF in its entirety. In that case you either have to embed in flash as a header, or you have to stick it on some filesystem and accept that it's slow. Which is *fine* in some cases, like e-paper displays. It's all dependent on platform and situation. And then there's MCUs like the Teensy which performance benchmarks I've seen suggest its SDMMC interface is plenty fast enough to use TTF streaming.

Anyway, all of the above is stuff to bear in mind. I'll try to bottom line it.

Keeping it streaming allows maximum flexibility but almost demands CBFS

I can make it use an array of TTF data bytes instead. It's not a big deal, but it can't do both due to flash space limitations and the idea of duplicating all the code twice.

So technically, it doesn't require CBFS, and we have other options.

I'd love to hear your thoughts. Particularly you @kisvegabor since I think you've been following along longest.

Edit: I had an idea. I could ditch CBFS altogether, and emulate a stream in my own internal code allowing me to expose a method that creates from a file or from a stream. I don't know why I hadn't thought of that before.

Now I really think maybe removing CBFS would be a good idea, and I can implement my own streamer. I'll wait to hear from you.

**Comment by codewitch-honey-crisis** [comment]
The Micropython failure appears to be caused by it ignoring my extern "C" linker specification in src/libs/lv_tiny_ttf.h

Which can only mean your scripts are changing my code in some way that nobody here seems to be able to figure out.

It's as if the MP generator is including my source file, and defining LV_USE_TINY_TTF without including my header, and then trying to reference the function. 

But I can't be sure because it's a black box that's just giving me yet another reason to dislike Micropython, and I didn't need more.

There is nothing wrong with my code. There is something wrong with your scripts.

I can't support this codebase folks. Sorry. Your build scripts are too much magic, too little delivery for me. Stick a fork in me guys. I tried, but I'm done.

This is super upsetting. But I don't work with build scripts that nobody understands and can maintain. One more shot at it then I'm closing this pull request.

**Comment by kisvegabor** [comment]
Sorry, but sometimes it takes me a few days to catch up and add comments, but I'm sure we can manage it from here. 

I'm looking into it now.

**Comment by kisvegabor** [comment]
Locally (no CI build magic) I get a  compiler error.
```c
static void stbrp_init_target(stbrp_context * con, int pw, int ph, stbrp_node * nodes, int num_nodes)
```
is not defined. 

An it's really not defined.  as `#ifndef STB_RECT_PACK_VERSION` disables this function [here](https://github.com/lvgl/lvgl/pull/3680/files#diff-9026ff81205b0a88ba57f2e61f50682102c823e5c99971942c1d2ab2374286beR4194).

In `stb_rect_pack.h` there is `#define STB_RECT_PACK_VERSION  1` however nothing related is included  `stb_truetype_htwc.h` from where the `STB_RECT_PACK_VERSION` could be seen. 

From this the MP CI failure makes sense. I wonder how it works on your end? Maybe have you added some defines to the compile command?

codewitch-honey-crisis:
I'm not sure, because my last commit was a shot in the dark. I don't even remember what I changed, or if I tried to build, or just forgot. I was at my wits end and also it was really late, so I don't even remember a lot. I'll take a look.

codewitch-honey-crisis:
I can't build locally because main.py build freezes here:

Scanning dependencies of target lvgl
[  2%] Built target test_common

It used to work. I don't care though. I'll check in and let your CI/CD test it. No more f*cks to give.

codewitch-honey-crisis:
How do I make this stop?
make[2]: *** No rule to make target '/mnt/hgfs/c_shared/lvgl/src/libs/fsdrv/lv_fs_cbfs.c', needed by 'CMakeFiles/lvgl.dir/mnt/hgfs/c_shared/lvgl/src/libs/fsdrv/lv_fs_cbfs.c.o'.  Stop.

I removed that stupid file. And the header. And everything that referenced it. I just searched all text in the project. lv_fs_cbfs does not exist.

So you tell me where this is coming from

codewitch-honey-crisis:
I had to put a dummy lv_cbfs_fs.c file in the scripts. There's a hidden reference to that file in your build scripts. Not my problem as far as I'm concerned. I'm just letting you know.

codewitch-honey-crisis:
/usr/bin/ld: build-dev/build-dev/lvgl/lv_mpy.o:(.data.rel.ro.mp_lv_tiny_ttf_destroy_mpobj+0x18): undefined reference to `lv_tiny_ttf_destroy'
/usr/bin/ld: build-dev/build-dev/lvgl/lv_mpy.o:(.data.rel.ro.mp_lv_tiny_ttf_create_data_mpobj+0x18): undefined reference to `lv_tiny_ttf_create_data'
/usr/bin/ld: build-dev/build-dev/lvgl/lv_mpy.o:(.data.rel.ro.mp_lv_tiny_ttf_create_file_mpobj+0x18): undefined reference to `lv_tiny_ttf_create_file'
make: Leaving directory '/home/runner/work/lvgl/lvgl/ports/unix'
collect2: error: ld returned 1 exit status

Well that was a waste of time. Your build scripts don't work, like I said. It would help if they actually linked my source file, and did it in the context of the header that helpfully defines those functions as extern SO IT CAN FIND THEM.

But I can't work with broken build scripts. The C code is valid.

codewitch-honey-crisis:
Sorry. I'm off for today. I've done everything I can, and I'm frazzled.

kisvegabor:
Okay, now it compiles and work for me locally. It won't pass the CI as there are many warnings. 

To be honest, I'm a little bit puzzled. You have sent pull request against LVGL's `master` branch, however in `master` I renamed `lv_mem_alloc` to `lv_malloc`. I fixed it in [2c3a863](https://github.com/lvgl/lvgl/pull/3680/commits/2c3a86328d172a1e647fd10ec7597116bda292b4) but I don't know how it worked on your end. 

Anyway, now it's working and I can clean it up a little bit and resolve the warnings.

codewitch-honey-crisis:
i'm not sure what happened, but it probably has to do with the fact that I have to test using 8.x version of LVGL and then move the code over to the 9 branch and change those calls. I guess I forgot to do it there.

kisvegabor:
Okay, no  problem :slightly_smiling_face: 

I could fix a lot of warnings but still there are some. However this the smaller issue. As I took a deeper look into the original code I saw the single header approach implementation of the core engine. And well... to be honest... it's one of messiest code I've ever seen.  @embeddedt what's your opinion about it?

codewitch-honey-crisis:
It is pretty bad, but it's also pretty mature, at least more mature than I could make on my own. And the API that surrounds it is easy enough to use. So yeah, it's sausage, but it's stable sausage and is being actively maintained.

Edit: Because it's being actively maintained, it's probably counterproductive to try to sanitize it. As it is I already have to backport my streaming editions any time they update the main because the author didn't want to incorporate streaming support.

My concern to be clear is that I'd like to keep this library abreast of any bugfixes or improvements to the original STB offering which this was derived from. The closer it is to the original, the more realistic that is to incorporate iterative changes from the source. I also am loath to muck about with something that works well, and is in use in production (I use a port from an earlier version of this STB offering in my htcw_gfx library which i use professionally). It has been reasonably stress tested as well, and it doesn't do things like leak.

Part of the issue with the mess is the nature of TTF. The format is extremely unwieldy and overly complex. Imagine getting Microsoft, Apple, and IBM in a room, and getting them all drunk on absinth and then spinning that room around until somebody puked. TTF was the result. The code was either going to be this way, or it was going to be broken out into endless abstractions in order to manage the complexity. Take a look at the spec https://developer.apple.com/fonts/TrueType-Reference-Manual/

This code does an amazing amount in a small amount of space. It's not very intelligible, but the maintenance isn't exactly our problem (for better AND for worse) because STB holds the keys to that kingdom. Unless you really want me to try to rewrite it from the ground up. I don't think the result would be as good as what STB has provided, even if it's easier to read.

So I find myself in the uncomfortable position of defending messy code - and code that isn't even mine! But here we are.

embeddedt:
> As I took a deeper look into the original code I saw the single header approach implementation of the core engine. And well... to be honest... it's one of messiest code I've ever seen. @embeddedt what's your opinion about it?

It seems established (the changelog dates back to 2015) and it's a better solution than bringing in FreeType just for loading TTF files, so I'm in favor of ignoring that for now. We don't need to look into the implementation too much since @codewitch-honey-crisis has kindly done the work for us, and the public APIs seem to make sense. So, +1 for keeping from me.

kisvegabor:
Okay, I made very minimal changes to fix warnings. Now it passes the tests for me locally, let's what we get here.

codewitch-honey-crisis:
same linker errors as before :(

kisvegabor:
I hope I fixed it with [549a369](https://github.com/lvgl/lvgl/pull/3680/commits/549a369374870714f11ef093ec58277c807bcc75)

codewitch-honey-crisis:
Interesting. You aren't using C linkage for your external functions when compiling with C++. I think yours is the first major codebase I've seen like that. I'll have to bear that in mind.

kisvegabor:
AFAIK only this "trick" is required to make C work in C++:
```c
#ifdef __cplusplus
extern "C" {
#endif
```

I think it's pretty common. 

Anyway, it seems working now. 

In summary what is the current status? What are the missing TODOs?

codewitch-honey-crisis:
I must have misread your changes. It's late here. I thought you commented the 'extern "C"' bit out.

Anyway, there are no TODOs of mine directly. STB had some TODOs in their codebase - rasterizer minutae and things like that, but they haven't impacted the performance of the engine that I've seen, like there's nothing critical missing, that I know of.

I've tried as I said to keep the STB files as close to the original as I could, so I'm loath to improve on their engine because that effectively means taking over maintenance of it, and I don't entirely understand it, nor am I as versed at truetype as the author of it. So my plan was to backport my updates every time he makes significant changes. I *did* try to submit my additions to him via a pull request, but he doesn't want the streaming support.

Edit: The bottom line is I think it's ready to go into the master branch. I'll keep my eye out for improvements to the original STB offering. If anything major comes down the pipe I can update the code.

codewitch-honey-crisis:
So I'm curious what the plan is for this. Further review before committing into the branch and closing the request? Mostly I want to know because I imagine conflicts will be brewing as it goes on, although probably nothing breaking. Also I'm kind of excited to see this get incorporated into the latest bits, so there's that too. :)

embeddedt:
There are still a couple files which had their line endings somehow changed earlier in the PR process (see the diff on the unrelated files); those will need to be reverted. I can fix that but not right now. Aside from that we need to:

1. document how this works, or at the very least create a super simple C example (the current file seems to be blank :thinking:)
2. check that the necessary APIs are generated correctly on the MicroPython side to create an example in the future (@amirgon or I should be able to do this after 1)
3. remove the empty stub files related to CBFS. @codewitch-honey-crisis if you are building using the `main.py` script, adding `main.py --clean` should make it rescan what files exist.

codewitch-honey-crisis:
Thanks. I'll put together an example when I'm done putting out fires at work (metaphorically speaking) :)

I actually have one, I was just waiting for things to stabilize, and then I might need a little guidance incorporating it into the existing example code - I'll see how far I can get on my own.

Thanks for the main.py --clean tip. I'll sort that stuff out too.

Edit: I've been meaning to ask but I keep forgetting. Does LVGL have a way to input DPI, or is an em just 1.3 pixels or so?

kisvegabor:
I still would like to take  deeper look on the APIs. I'm doing it now.

kisvegabor:
What was the final conclusion about CBFS? Can't we use it to have a unified file based create function?

kisvegabor:
An other question: how complicated it is to the change the size of an already existing font? E.g. with `lv_tiny_ttf_set_size(font, 24);`

codewitch-honey-crisis:
I removed CBFS, and added a second creation function instead. CBFS does not create a unified interface as such. It emulates a FS over an array, allowing me to use the file creation function with an embedded header. It's obsolete.

Using it required far more code than using one of the creation functions directly. If you *really* want a unified creation function, we can just add a prototype for the main creation function implementation (currently "private") although I don't see why you'd do that as it adds no value.

It's trivial to change the size of the font, but I should add a function for doing it. I had originally intended to do something like that but it got lost in the fray.

I can do a pull and take care of it if you like. Otherwise, it's trivial to do - just change the font desc structure with new values.

codewitch-honey-crisis:
I have most of the set size code in place, but how do I force a redraw of the label? I looked up how to do it and all i found was this 
https://github.com/lvgl/lvgl/issues/206

Edit: I found a workaround but it's less than ideal:
```cpp
lv_tiny_ttf_set_size(ttf_font, 20);
lv_obj_set_style_text_font(label,ttf_font,0);
```

kisvegabor:
> I removed CBFS, and added a second creation function instead. CBFS does not create a unified interface as such. It emulates a FS over an array, allowing me to use the file creation function with an embedded header. It's obsolete.

CBFS could be useful for other libs too. So I'm in favor of adding and using it.

> I have most of the set size code in place, but how do I force a redraw of the label? 

We can't search for all styles using a given font it's the user's responsibility. He/She can use the following functions:
- if the font was used in a local style: `lv_obj_refresh_style(obj, selector, prop);`
- if the font was part of an `lv_style_t`: `lv_obj_report_style_change(style);`

codewitch-honey-crisis:
I'll add CBFS back in. I still think we shouldn't force people to use it for the fonts, so I'm in favor of keeping both creation functions. I've gotten some feedback about CBFS when I was producing a standalone library (before I was integrating here i think? early on in any case) and people pointed out it was counter-intuitive to use at first. The reception wasn't great. I'm still fine with adding it back, I just don't think we should force people to use it for tiny_ttf, if that makes sense.

codewitch-honey-crisis:
actually, how hard would it be to pull CBFS from an earlier check in? The reason I ask, is I don't think the code formatting script is correcting my line endings. For me to test the code I *have to* do it on my windows machine. The VM is so slow I want to get out and push! It's due to my win11 settings disabling virtualization acceleration (why i wanted to use WSL in the first place, but it takes over the hypervisor functions now)

Anyway the point is I can't develop in the linux VM. It's all i can do to build lvgl itself and reformat. So I've been pasting my code in from my windows machine in order to get something I've tested into the codebase. It's not fixing line endings.

It's a mess. My workflow is currently a nightmare because of that. I don't want to check something in that screws everything up and forces you to fix what I apparently can't on my end. :(

kisvegabor:
To keep in more intuitive we can use `lv_tiny_ttf_create(path, ine_height)` and `lv_tiny_ttf_create_data(buf, buf_size, line_height)` which is wrapper to `lv_tiny_ttf_create` with a CBFS file.

I'm concerned about performance, but I don't know the internals. Normal `read` function reads the data into a buffer which means:
- we need a buffer (malloc, free, out of memory error)
- we need to copy the data into the buffer

While when using "flash data" we only need to point to the right part of the memory. No malloc, no copy. 
Does the `lv_tiny_ttf_create_data` implementation uses this kind of optimization?

kisvegabor:
> actually, how hard would it be to pull CBFS from an earlier check in?

No problem with that, I can do it if decide to go with it.

codewitch-honey-crisis:
> Does the lv_tiny_ttf_create_data implementation uses this kind of optimization?

It does not, because in order to do so it would about double the size of the flash footprint for tiny_ttf**. The reason being is I had to make the whole library act on streams in order for it to act on both. Abstracting the other direction is not possible without some sort of virtual memory style pointer virtualization.

It is possible to to compile out the streaming so only the create_data function would be available in which case it won't do all those copies. I don't have a facility in there but I could probably add it without much fuss. The only thing I'd want to know is what's the compile time constant name I should use?

** it would also make backporting new updates from STB a lot more difficult.

codewitch-honey-crisis:
added `LV_TINY_TTF_FILE_SUPPORT` which defaults to zero. When it's not enabled, the tiny_ttf does not support loading from files, but runs arrays more efficiently. HOWEVER, that technique will not work well on AVR platforms due to the fact that they need `PROGMEM` attributes on their data and `pgm_read_byte()`, etc to prevent badness. I don't do any of that in the library. It will work fine on many platforms but not ones that need `PROGMEM` markup in the code (`PROGMEM` being a stand in for a platform specific attribute that keeps the data in flash)

When it is not enabled `lv_tiny_ttf_create_file()` will not be available. When it is enabled it will be.

I've also added the following method:
```cpp
void lv_tiny_ttf_set_size(lv_font_t * font, lv_coord_t line_height);
```

If you really want em sizes in there I can add that, but it's about 1.3 pixels unless LVGL can give me an accurate DPI. setting the line_height directly is generally more convenient/intuitive in general, but it may be you want consistency with the other fonts. I need to know how your em is computed before I make that change.

Aside from that you may want to check line endings since I have to copy code over from my windows box. I ran the formatter on it, but apparently it's not fixing it like I said. Sorry. I wish there was a better way to do this. Edit: I'll look for some kind of tool or python script I can run to correct the line endings.

Edit: Re-performance. In my experience, with the streaming support, a standard ESP32 WROOM is just fast enough as long as the font sizes aren't too big. You can get by with a SAMD21 if you don't mind a little creativity (drawing to intermediary bitmaps so you don't see flicker from the letters drawing). Without the streaming I expect a little better, but not much, particularly because memcpys tend to be pretty fast on modern MCUs, so I wouldn't bank on running this on lower end IoT. I don't think this library is particularly inefficient, but it's TrueType. It was never meant to run a system like this, so it won't run on absolutely everything.

kisvegabor:
> HOWEVER, that technique will not work well on AVR platforms

If a platform needs such trick to read flash I can hardly imagine it's a good choice for GUI. So I think it's an acceptable limitation.  

> If you really want em sizes in there I can add that, but it's about 1.3 pixels unless LVGL can give me an accurate DPI. setting the line_height directly is generally more convenient/intuitive in general, but it may be you want consistency with the other fonts. I need to know how your em is computed before I make that change.

Most importantly `line_height` should be the height a line can fully fit as LVGL uses it to calculate the label's height. 

We have `lv_disp_get_dpi(disp)`. It returns the DPI set in the display driver (the default is 130). However, LVGL supports multiple displays and the same font can be used on any display. So I think we don't need to deal with this at this moment. 

 > with the streaming support, a standard ESP32 WROOM is just fast enough as long as the font sizes aren't too big. 

Can you share some benchmarks? E.g. how fast it is to render
1. an empty screen
2.  a full-screen lorem ipsum with a built in font
3.  a full-screen lorem ipsum with a tiny TTF

To easily measure the performance you can enable `LV_USE_PERF_MONITOR` and modify the main loop like:
```c
  while(1) {
      lv_timer_handler();
      lv_obj_invalidate(lv_scr_act());
      my_sleep(5);
  }
```

codewitch-honey-crisis:
> Can you share some benchmarks?

I'll take care of that tonight once I have some time to make some. I wouldn't expect miracles though, TTF is pretty CPU intensive. :)

I'm currently testing on a 240x135 display because it's what I have my build env set for. I'll set up a 320x240 when I set up the benchmarks

Edit: Keep in mind TTF can render outside of its bounding box. That's by design of TTF. Not sure how LVGL will deal with that when it happens. It's nearly impossible to deal with in a reasonable way. I really dislike that TTF can do that. Basically points can be outside a glyph box. I've got a font that does this, so I'll test that too. On my graphics library I built it works, but sometimes you have to offset the upper left of the font, and then leave extra space for the lower right extents/create extra area

kisvegabor:
> I'll take care of that tonight once I have some time to make some. I wouldn't expect miracles though, TTF is pretty CPU intensive. :)


Great, thank you in advance :slightly_smiling_face: 

> Keep in mind TTF can render outside of its bounding box. 

I saw this issue mainly with italic fonts and I tried to mitigate it with a hack: `lv_label` sets a `line_height / 4` extra draw area around the label object.

codewitch-honey-crisis:
```c:/users/gazto/.platformio/packages/toolchain-xtensa-esp32@8.4.0+2021r2-patch3/bin/../lib/gcc/xtensa-esp32-elf/8.4.0/../../../../xtensa-esp32-elf/bin/ld.exe: .pio\build\m5stack-core2\firmware.elf section `.dram0.bss' will not fit in region `dram0_0_seg'
c:/users/gazto/.platformio/packages/toolchain-xtensa-esp32@8.4.0+2021r2-patch3/bin/../lib/gcc/xtensa-esp32-elf/8.4.0/../../../../xtensa-esp32-elf/bin/ld.exe: DRAM segment data does not fit.
c:/users/gazto/.platformio/packages/toolchain-xtensa-esp32@8.4.0+2021r2-patch3/bin/../lib/gcc/xtensa-esp32-elf/8.4.0/../../../../xtensa-esp32-elf/bin/ld.exe: DRAM segment data does not fit.
c:/users/gazto/.platformio/packages/toolchain-xtensa-esp32@8.4.0+2021r2-patch3/bin/../lib/gcc/xtensa-esp32-elf/8.4.0/../../../../xtensa-esp32-elf/bin/ld.exe: region `dram0_0_seg' overflowed by 33512 bytes```

This is a bit of a problem. I can't change the DRAM segment size.

Why is this requiring so much space? It's completely stumping me. I may have to benchmark with the release version.

For reference see this:
https://github.com/espressif/arduino-esp32/issues/2943

**Comment by codewitch-honey-crisis** [comment]
I'm getting some unusual results.
My TTF font is reporting significantly better benchmark numbers than the stock font, despite being larger. It is however, pegging the CPU at 99%, whereas the stock font runs at 89%
@ 240x135 - SPI/Arduino LVL 8.3.1** It's what I could get working
TTF:
33 FPS @ 99% CPU
Stock font
23 FPS @ 89% CPU

I don't buy it unless maybe it has to do with the antialiasing? But I don't know how to verify the results.
```cpp
// create a font with a line height of 20 from the data
ttf_font = lv_tiny_ttf_create_data(data,data_size,20);
// create simple label
label = lv_label_create( lv_scr_act() );
lv_color16_t col;
col.full = 31<<11;
lv_obj_set_style_text_color(label,col,0);
lv_style_selector_t selector=0;
// uncomment below to set truetype
//lv_obj_set_style_text_font(label,ttf_font,selector);
lv_label_set_text( label, text);
lv_obj_align( label, LV_ALIGN_CENTER, 0, 0 );

```

Anyway, I also forgot to mention, vector font resource usage varies significantly from font to font. Here I was testing with a simple, 20KB font file - Telegrama at fontsquirrel.com

Edit 2: I just tried it with the Ubuntu font at fontsquirrel and I'm getting a much more expected 6fps at 97% CPU usage.

Edit 3: I just figured out how to verify the results. silly me. I just change the color to a random value every loop() call. The first result is not accurate. Maybe it's less than 1fps? not sure but the perf report is wrong.

kisvegabor:
> Edit 3: I just figured out how to verify the results. silly me. I just change the color to a random value every loop() call. The first result is not accurate. Maybe it's less than 1fps? not sure but the perf report is wrong.

The FPS monitor underflows under 4 FPS. I didn't think it will matters :sweat_smile: 

You are trying it on ESP, right? I've found a [few articles](https://blog.classycode.com/esp32-floating-point-performance-6e9f6f567a69) about the `float` support in ESP is not that fast. I'll try on an STM32F769 later today.

Have you enabled any compiler optimization?

codewitch-honey-crisis:
Yeah, it's an ESP32 WROOM

I think it's just doing -Os - the default for the arduino toolchain

I haven't tried putting optimization pragmas in the source. Frankly, I'm not a big believer in bit twiddling for speed. If I can't make algorithmic improvements to code, I just don't see the value in a few last mile improvements by changing compiler optimization settings.

That said, I did not know that ESP32 float support was particularly bad, but the cost to convert this to fixed point makes it impossible. Unless you want to do it. :P

kisvegabor:
I've also made some test on STM32F769 Discovery, 800x480, 200MHz, O3 optimization. 

I think the performance is fine for a few texts, but for full screen texts (not so typical) it's way to slow. Probably glyph caching could help a lot. With 32px height let's say a glyph has 20x20 box on average and there are 70 glyphs to cache. It's only 28kB (not counting the glyph descriptors).

Anyway, if only a few labels needs to be scaled, or rendered from TTF real time, I think it's pretty usable and useful.

---

### Long text

Test code:
```c
	  lv_font_t * f = lv_tiny_ttf_create_data(Ubuntu_data, sizeof(Ubuntu_data), 32);
	  lv_obj_t * obj = lv_label_create(lv_scr_act());
	  lv_obj_set_style_text_font(obj, f, 0);
//	  lv_obj_set_style_text_font(obj, &lv_font_montserrat_12, 0);
	  lv_obj_set_width(obj, lv_pct(100));
	  lv_label_set_text(obj, "Lorem ipsum dolor sit amet, consectetur adipiscing elit. Sed vel dictum augue. Class aptent taciti sociosqu ad litora torquent per conubia nostra, per inceptos himenaeos. Curabitur sagittis commodo tincidunt. Maecenas ac porttitor est. In condimentum condimentum turpis, ac congue ligula laoreet sagittis. Nunc sit amet aliquam erat, fermentum gravida ligula. Curabitur imperdiet ex id tortor sodales dapibus. Proin sollicitudin metus felis, vitae semper neque porttitor in. Vestibulum ante ipsum primis in faucibus orci luctus et ultrices posuere cubilia curae;");
```

- Empty screen: 10ms
- Montserrat built-in 12px: 30ms
- Montserrat built-in 32px: 47ms. Profiler:
image
- Ubuntu, 32 px: 234 ms. Profiler:
image
- Ubuntu, 12 px: 121 ms

---

### Very long text being scrolled

Profiling a lot of text being scrolled with Ubuntu 32px: 3-4 FPS:
image

---

### Buttons with labels

15 button visible at the same time + scrolling

```c
  lv_font_t * f = lv_tiny_ttf_create_data(Ubuntu_data, sizeof(Ubuntu_data), 20);
  lv_obj_set_style_text_font(lv_scr_act(), f, 0);
  lv_obj_set_flex_flow(lv_scr_act(), LV_FLEX_FLOW_ROW_WRAP);
  lv_obj_set_style_pad_all(lv_scr_act(), 80, 0);
  lv_obj_set_style_pad_column(lv_scr_act(), 80, 0);
  lv_obj_set_style_pad_row(lv_scr_act(), 40, 0);

  uint32_t i;
  for(i = 0; i < 30; i++) {
	  lv_obj_t * btn = lv_btn_create(lv_scr_act());
	  lv_label_set_text_fmt(label, "Button %d", i + 1);
  }
```

- Ubuntu 20 px: 72ms, 15-17 FPS while scrolling
- Built-in Montserrat  20 px: 30ms, 30 FPS (80% CPU) while scrolling

codewitch-honey-crisis:
That's not actually that bad, TBH, for a vector font!

I produced Tiny TTF standalone for LVGL 8.3.1+ (maybe other versions but I've tested it with that one)

The reason I'm telling you is that there's some documentation at the readme. It's mostly just code, but it covers how to use it pretty well, IMO, given there are only a handful of functions.

I'll prepare an example as well, I just need to organize and test it.

The code is not quite drop in. I've omitted the `lv_` prefix from everything in this library.

https://github.com/codewitch-honey-crisis/tiny_ttf

codewitch-honey-crisis:
I've been thinking about the glyph caching scheme. I'd be willing to do it, but here's the thing. I'd need my own hashtable based caching scheme, and that code is - while not difficult, it seems heavy handed to have it only be used in tiny_ttf.

Also it could take a significant amount of RAM and you should be able to use your own allocator, so you can use PSRAM. How would one do that in LVGL?

the wrinkle with the cache i see is how to make it generic without templates. void* sort of works but only if you're willing to make heap allocations and that stinks in this case - one for each item in the cache - that's.. not fantastic. do you have ideas? one thing I can think of is having a reserve space allocated with each element, and allowing you to copy your data into it. Cache keys are a little more complicated but might be handled using the same principle.

Maybe I'm overthinking this. Maybe I should just do it for tiny_ttf. If this was C++ this would be cake, but I need templates. :)

codewitch-honey-crisis:
I've added a preliminary LRU caching scheme and got it to double the framerate to 10FPS on lorum ipsum with around 4KB of cache per font/size.

I think it could be faster, but I am not a superstar hashtable implementer. I kept it simple. I'm not checking it in yet, because I want to hear back from you if you have any thoughts about this.

Edit: I think if i cached the glyph dsc callback results it would bring things more in line with other fonts, but I am loath to copy an entire kerning table into memory.

It breaks the API, if only because now there's a cache_size parameter as the final parameter to the create functions.

Edit 2: In an attempt to increase the performance further I attempted to cache glyph descriptions as well, but lost 1fps, probably due to me using the same hashtable for both. I was looking at the callbacks, and LVGL seems to call the glyph description callback quite a bit. I also noticed if anything, fonts in general could use improvements in performance. Might it be worthwhile to cache glyph descriptions deeper in LVGL so it applies to all fonts?

kisvegabor:
I'm also happy with the performance, as I think it's already really usable for a lot of things. Before going deeper into jungle of caching. I suggest marge the current version first. 

TODOs I see:
- Remove CBFS
- Add docs
- Add examples
- Add a basic test (screenshot compare is fine)

codewitch-honey-crisis:
will do

codewitch-honey-crisis:
should i wait for it to pass again so I can pull and add the changes you requested?

kisvegabor:
Great, thank you.  If it's simpler for you I can add the test too. 

It's passing now.

codewitch-honey-crisis:
Yeah I'd appreciate it if you could write the test, as I am completely lost by your test framework. I don't have much experience with Unity. I've just done the basics. I just checked in docs and an example.

Edit: I have the cache system ready to integrate. I'm just waiting on you to give me the go ahead and I'll pull the trigger.

kisvegabor:
Great! I added the test. 

> Edit: I have the cache system ready to integrate. I'm just waiting on you to give me the go ahead and I'll pull the trigger.

Amazing. Let's merge it first and please send a new PR with the cache support.

codewitch-honey-crisis:
will do. I'll wait for the merge.

kisvegabor:
Aaaand all passed, merging :tada: :tada: :tada: 

Thank you very much. This feature will make a huge difference!

codewitch-honey-crisis:
Woohoo! You're very welcome. I'm just glad to contribute. I will work on the caching as soon as I figure out how to make github bend the way I need it to bend to make a second pull request.

codewitch-honey-crisis:
I can’t reply on GitHub because their website is being really slow loading the thread. I warn because replying in email often destroys the formatting. Those results are honestly better than I expected. When I use these in the wild, I am frugal with when and how I draw with them. I’ve never tried to do full screen text on LCD with it – only e-paper. That’s all on an ESP32. Glyph caching may help, but how do I know how much RAM to allocate for it?

kisvegabor:
Feel free to open a new issue to discuss the new questions.

Cluir:
Hello,how is the information of the ubuntu_font array generated? @codewitch-honey-crisis @kisvegabor @embeddedt

codewitch-honey-crisis:
The font was downloaded from fontsquirrel.com, and then encoded into a uint8_t based byte array verbatim from the TTF file.
You can also stream the TTF from a filesystem. It also supports OTF format.
If you select "C++" from the dropdown here: https://honeythecodewitch.com/gfx/converter
It will generate you a header for any font. One thing is you have to unzip it before you drag it on there. It won't work if you try to drag directly from a zip file in the windows shell for some reason. It creates a nearly empty header.

Cluir:
> 字体是从 fontsquirrel.com 下载的，然后从 TTF 文件逐字编码成基于 uint8_t 的字节数组。 您还可以从文件系统流式传输 TTF。它还支持 OTF 格式。 如果您从此处的下拉列表中选择“C++”： https: [//honeythecodewitch.com/gfx/converter](https://honeythecodewitch.com/gfx/converter) 它将为您生成任何字体的标题。一件事是你必须先解压缩它，然后再将其拖到那里。如果您出于某种原因尝试直接从 windows shell 中的 zip 文件拖动，它将不起作用。它创建了一个几乎为空的标题。

I put the ttf file on the external sd card, when I read it, my device becomes very stuck, and the rendering will be very slow. what caused it?My device info is RT1052，528MHz , Cortex M7 , 32MB SDRAM, 16MB FLASH          Is it because of poor performance?

codewitch-honey-crisis:
SD is going to be very slow. What you should do is read the file into SRAM (or PSRAM) and work with it from there. I provided the ability to read from SD and other sources to future proof the code more than anything. Most MCUs these days are not fast enough to make reading it directly from the SD practical. Who knows in a few years? =)

Cluir:
> 字体是从 fontsquirrel.com 下载的，然后从 TTF 文件逐字编码成基于 uint8_t 的字节数组。 您还可以从文件系统流式传输 TTF。它还支持 OTF 格式。 如果您从此处的下拉列表中选择“C++”： https: [//honeythecodewitch.com/gfx/converter](https://honeythecodewitch.com/gfx/converter) 它将为您生成任何字体的标题。一件事是你必须先解压缩它，然后再将其拖到那里。如果您出于某种原因尝试直接从 windows shell 中的 zip 文件拖动，它将不起作用。它创建了一个几乎为空的标题。
编码之后的标题，还具有某些矢量字体特性吗，比如放大缩小。

Cluir:
> The font was downloaded from fontsquirrel.com, and then encoded into a uint8_t based byte array verbatim from the TTF file. You can also stream the TTF from a filesystem. It also supports OTF format. If you select "C++" from the dropdown here: https://honeythecodewitch.com/gfx/converter It will generate you a header for any font. One thing is you have to unzip it before you drag it on there. It won't work if you try to drag directly from a zip file in the windows shell for some reason. It creates a nearly empty header.

Does the encoded header still have the characteristics of vector fonts, such as zooming in and out?

codewitch-honey-crisis:
Yes. Again. It is just a TTF file in an array. These are vector fonts. Not raster fonts.

Cluir:
> 是的。再次。它只是一个数组中的 TTF 文件。这些是矢量字体。不是光栅字体。

oh，how did it do , What information is stored in it, I did not find relevant information about the encoded array. In addition, the volume of my commonly used Chinese character library is too large after encoding. Is there any compression method?

codewitch-honey-crisis:
Yeah, umm, I don't think you'll be able to do east asian fonts in TrueType on most IoT devices. Tiny TTF won't even load some asian font tables, because I have never seen code to do it so I had no reference when I wrote it.

Cluir:
> 是的，嗯，我认为您无法在大多数 IoT 设备上使用 TrueType 制作东亚字体。Tiny TTF 甚至不会加载一些亚洲字体表，因为我从未见过这样做的代码，所以我在编写它时没有参考。

In fact, I have been using tinyttf to render Chinese fonts, but if I store the fonts on the sd card, the device will spend a lot of time reading the font files, if I encode the fonts into headers, because the files are too large, Will take up a lot of flash. Is there a solution that compresses the encoded array for decoding and rendering in tinyttf, sacrificing time for space.

codewitch-honey-crisis:
No, and because of the way it works, it would be impossible to compress it. It works with the file as is off the storage directly to save memory, so that cannot be compressed. When I said you won't be able to do it on most IoT devices I was taking into account the fact that SD would probably be too slow. However, you may get better results using 4-wire SDMMC instead of 1-wire or SPI for the SD. It's still not going to be fast. PSRAM is your best bet, but you'll still need to fit it into about 4MB because I haven't written any himem stream accessors yet. 

I'm pleased to learn it will render Chinese, but I'm not surprised your fonts won't fit in flash. I think your best bet is to use a different font engine, and render just a subset of characters you'll actually be using. Alternatively, you could find a TrueType font editor and remove the characters you won't be using.

PR #3681, opened 2022-09-11
test(led) add unit tests

mmurty:
Add unit tests for lv_led widget.


mmurty:
@kisvegabor Hi,
Please send me the screenshot for the below code:

```
    lv_obj_t * led  = lv_led_create(active_screen);
    lv_obj_align(led, LV_ALIGN_CENTER, -80, 0);
    lv_led_off(led);

    led  = lv_led_create(active_screen);
    lv_obj_align(led, LV_ALIGN_CENTER, 0, 0);
    lv_led_set_brightness(led, 150);
    lv_led_set_color(led, lv_palette_main(LV_PALETTE_RED));

    led  = lv_led_create(active_screen);
    lv_obj_align(led, LV_ALIGN_CENTER, 80, 0);
    lv_led_on(led);
```

tests/src/test_cases/test_led.c
```diff
@@ -0,0 +1,129 @@
...
+
+    /* Test by changing colors randomly. */
+    lv_led_set_color(led, lv_palette_main(LV_PALETTE_AMBER));
+    TEST_ASSERT_EQUAL(lv_palette_main(LV_PALETTE_AMBER).full, ledObj->color.full);
+
+    lv_led_set_color(led, lv_palette_main(LV_PALETTE_GREY));
+    TEST_ASSERT_EQUAL(lv_palette_main(LV_PALETTE_GREY).full, ledObj->color.full);
+
+    lv_led_set_color(led, lv_palette_main(LV_PALETTE_RED));
+    TEST_ASSERT_EQUAL(lv_palette_main(LV_PALETTE_RED).full, ledObj->color.full);
```
embeddedt, line 49: I'm really not sure it's worth unit-testing in this manner. All we are doing here is verifying that `lv_led_set_color` actually changes `ledObj->color.full`. If we ever change the underlying object implementation we need to go back and update all the tests. Additionally, it's possible that due to some future bug the LED no longer chooses its color based on this field, meaning the test would pass even if the actual widget is broken.

I think instead of using object internals we should focus on testing what the user sees. For example, we should just set the color once or twice, and use a screenshot after each change to ensure that the color is changed correctly and that is reflected in the UI. This means we are only testing the public API surface and its apparent behavior to the user, which is a better way of testing in my opinion.

tests/src/test_cases/test_led.c
```diff
@@ -0,0 +1,129 @@
...
+}
+
+void test_led_set_brightness_works(void)
+{
+    lv_obj_t * led  = lv_led_create(active_screen);
+    lv_led_t * ledObj = (lv_led_t *)led;
+    
+    /* Test by changing brightness randomly. */
+    lv_led_set_brightness(led, 150);
+    TEST_ASSERT_EQUAL(150, ledObj->bright);
```
kisvegabor:
```suggestion
    TEST_ASSERT_EQUAL(150, lv_led_get_brightness(led));
```

tests/src/test_cases/test_led.c
```diff
@@ -0,0 +1,129 @@
...
+
+static void event_handler(lv_event_t * e)
+{
+    lv_event_code_t code = lv_event_get_code(e);
+    if(code == exp_evt_code) {
+        event_triggered = true;
+    }
+}
+/* Added this test to increase code coverage. */
+void test_led_event_draw_main_works(void)
```
kisvegabor: The screen shot test will make LVGL render the screen so it will not be required.

tests/src/test_cases/test_led.c
kisvegabor: > This means we are only testing the public API surface and its apparent behavior to the user, which is a better way of testing in my opinion.

I agree with this. Actually there should be an `lv_led_get_color` function too which could be used here. 

Similarly, instead of `ledObj->bright`, `lv_led_get_brightness(led)` should be used.

tests/src/test_cases/test_led.c
mmurty: Without this test the test coverage was 50%. After adding this the coverage increased to 100%.

tests/src/test_cases/test_led.c
embeddedt: It would still be ideal to use a screenshot in addition to just checking that the setters and getters work.

mmurty: @embeddedt I will be adding a screenshot in `test_led_render_1()` once @kisvegabor uploads it here.

kisvegabor: Here it is:
image

kisvegabor:
The CI says it's all good. What is coverage now?

mmurty:
> The CI says it's all good. What is coverage now?

100%

kisvegabor:
Wow, great! :slightly_smiling_face: 

Merging, please send an expense about 25 USD.

Issue #3685, opened 2022-09-13
Wrong behavior of overflow of lv_span


FASTSHIFT:
### Perform all steps below and tick them with [x]
- [x] Read the [FAQ](https://docs.lvgl.io/master/intro/index.html#faq)
- [x] Check the related part of the [Documentation](https://docs.lvgl.io/)
- [x] Update lvgl to the latest version
- [x] Reproduce the issue in a [Simulator](https://docs.lvgl.io/latest/en/html/get-started/pc-simulator.html)

### Describe the bug
When the font placeholder is turned off, and `lv_span` is used to display the mixed text of Chinese + emoji, displaying ellipses is not expected behavior.

### To Reproduce
in lv_conf.h:
`#define LV_USE_FONT_PLACEHOLDER 0`

```c
void span_test(void)
{
    lv_obj_t* spans = lv_spangroup_create(lv_scr_act());
    lv_spangroup_set_align(spans, LV_TEXT_ALIGN_LEFT);
    lv_spangroup_set_overflow(spans, LV_SPAN_OVERFLOW_ELLIPSIS);
    lv_spangroup_set_mode(spans, LV_SPAN_MODE_EXPAND);
    lv_obj_set_style_border_width(spans, 1, 0);
    lv_obj_set_style_border_color(spans, lv_color_black(), 0);
    lv_obj_set_style_text_font(spans, &lv_font_simsun_16_cjk, 0);

    lv_span_t* span = lv_spangroup_new_span(spans);
    lv_span_set_text(span, "a写\ua060");  // ‘\ua060’ is an emoji that does not support display
}

```

### Expected behavior
![image](https://user-images.githubusercontent.com/26767803/189879206-93f4dc59-54df-4750-b4f6-3f58156b822d.png)

### Screenshots or video
![image](https://user-images.githubusercontent.com/26767803/189879108-dcce9b41-999b-4db6-8eff-b4bbc286b78a.png)

FASTSHIFT:
It can be displayed normally when the text order is modified:
```c
lv_span_set_text(span, "写a\ua060");
```

![image](https://user-images.githubusercontent.com/26767803/189892305-ce2c2227-2406-44fc-a0d1-dd7cc8dee673.png)


I tried to compare the drawing process of the two texts and found that `_lv_txt_get_next_line()` in `lv_txt_get_snippet()` returned different values.

For text: "写a\ua060", `ofs = 7`
[...]

PR #3687, opened 2022-09-15
fix(span): fix incomplete text display when disable the placeholder.

FASTSHIFT:

Related discussion: #3685

I tried adding an extra pixel to the text measurement, which made it possible to include an imgfont with a width of 0 after `max_width` was reached, as expected. 
But this is obviously not the best solution, what do you think?
@kisvegabor 

### Checkpoints

### DIFF
--- a/src/widgets/span/lv_span.c
+++ b/src/widgets/span/lv_span.c
@@ -606,7 +606,14 @@ static bool lv_txt_get_snippet(const char * txt, const lv_font_t * font,
         return false;
     }
 
-    uint32_t ofs = _lv_txt_get_next_line(txt, font, letter_space, max_width, use_width, flag);
+    lv_coord_t real_max_width = max_width;
+#if !LV_USE_FONT_PLACEHOLDER
+    /* fix incomplete text display when disable the placeholder. */
+    /* workaround by: https://github.com/lvgl/lvgl/issues/3685 */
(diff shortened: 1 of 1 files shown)


kisvegabor:
Sorry for it took me so long to immerse in it.

I also debugged it and found that the issue is related to that all CJK characters treated as "break chars". If you comment out this line it works normally:
https://github.com/lvgl/lvgl/blob/46cbbd672f8cf95323c2706b40698c337899004c/src/misc/lv_txt.h#L172


Adding 1 to width solves the issue in most of the cases but e.g. if `letter_space != 0` it fails.

So what is the underlying problem? There might be characters with `width==0`, and if they are on the end of the text, they might cause incorrect "ellipsis calculation". 

How can a character have zero width?
- The glyph is not found and `LV_USE_FONT_PLACEHOLDER 0`
- It's a non printable character
- Anything else?

All these seem very special and rare. Your fix (although doesn't seem to solve all problems) should work for most of these rare cases. 

So unless you got any ideas from my comment I'm ok with merging it. 

Just out of curiosity, do you find these issues while work on your hobby projects or while working professionally for a company?

FASTSHIFT:
@kisvegabor 
If you feel reasonable, I agree to merge this pr.
It was found in the company's project.

PR #3703, opened 2022-09-26
added glyph caching to tiny_ttf

codewitch-honey-crisis:
### Added glyph caching to tiny_ttf

Improved performance by adding glyph caching to tiny_ttf

### Checkpoints


codewitch-honey-crisis:
ARGH

Line endings are broken again. Is there an easy fix?

kisvegabor:
Maybe this?
https://docs.github.com/en/get-started/getting-started-with-git/configuring-git-to-handle-line-endings

codewitch-honey-crisis:
that's weird. where's the CI/CD stuff?

codewitch-honey-crisis:
I'm confused. I pulled from LVGL before forking. Yet stb_truetype_htcw.h has regressed. What happened?

More importantly, how do i fix it? Do you want me to override the warnings or do you want me to delete the functions?

codewitch-honey-crisis:
@kisvegabor when you said you were fixing warnings, were you talking about the unused functions in the STB truetype header?

If so, what did you do to resolve them, because it seems to have regressed, and I'm not sure what my next move should be. Is there an earlier version that was fixed that somehow got overwritten? I don't want to touch anything until I get a feel for what's going on.

codewitch-honey-crisis:
Okay I've overridden the warnings in the STB files, but tiny_ttf is now failing its test. I can't run that test locally i think. I'm not sure how I'd go about it. Can you run that down and see what's happening? I've tested the cache code under 8.3.1 but i can't get the latest bits to flash on an esp32.

Edit: I did copy the test code over to an 8.2.0 release and run it there (I had a codebase handy and already hooked to a device) and it appears to be displaying correctly, to my eye. Not sure why the PNG is off.

kisvegabor:
I'll look into it today evening . Meanwhile I added a TTF file based example and nothing is displayed on the screen. It doesn't crash but  the size of he label object is zero. Could you take a look at it?
884faf0a1

codewitch-honey-crisis:
Absolutely I will. That's kind of odd, because the file method was the first one that worked. Maybe some kind of regression.

codewitch-honey-crisis:
Can you do me a favor? I don't have a copy of the specific font you're using? Or maybe you just renamed it. Does the byte array version work?

kisvegabor:
Yes, I [renamed](https://github.com/lvgl/lvgl/blob/master/examples/libs/tiny_ttf/ubuntu_font.h) it as I found the name Ubuntu too generic. 

The byte array version is ver strange. Randomly it gives:
1. Empty screen
2. Normal screen
3. Or this
 
image

Running the file version a few times I realized that it works in the same "random" way. I saw it working once too.

kisvegabor:
I've found the problem and fixed it here: b96f72bdd

codewitch-honey-crisis:
thanks. I had to sleep, so I hadn't gotten to it yet. I still can't understand why it's failing the tests. When I try to make that label w/ the TTF font it makes the image just fine. It should match the png. Like I said, I'm having a rough time actually getting the tests to run. I probably just don't know what I'm doing.

kisvegabor:
I've checked why the CI is failing. It's because a barely visible pixel at the bottom right corner of letter 'h':
image
It seems it's the last pixel of the glyph so I guess it's an off by one copy issue.

codewitch-honey-crisis:
I need to reproduce this, but I can't run those tests. If I can somehow get this to reproduce using my ASCII art truetype renderer I can probably debug the problem in Visual Studio, but since I can't run this test, I don't know that I'll find anything.

Edit, if you can, I'd like you to be my eyes for now. Does this happen on all the letters or just the h?

kisvegabor:
I think I found something. `  lv_font_get_glyph_dsc(font, &dsc, ' ', ' ');` tells "space" has 1x1 box. Is it possible that somehow there is some garbage on that pixel?

codewitch-honey-crisis:
I don't think so. Space is just another glyph I think. Besides, it was passing the tests before, which leads me to suspect that this has something to do with the cache. I fear the worst - that memory is being freed before it is accessed, but I cannot run these tests, so I'm over here blind.

Edit: I got the tests running, so I'm investigating further.

codewitch-honey-crisis:
Is there a way to log output from code during a test run? like will LV_LOG spit to console or a file or something?

kisvegabor:
I think it the problem was solved by fix for the previous [issue](https://github.com/lvgl/lvgl/pull/3703#issuecomment-1260785077). At least locally now it passes for me. I updated the reference image tot the correct one and now all looks good.

codewitch-honey-crisis:
super!

kisvegabor:
I added a minor formatting to `lv_tiny_ttf.c`. Let's see what the CI says. 

In the meantime could you add a few words about the caching to the docs?

codewitch-honey-crisis:
will do now

kisvegabor:
Great! 

I've just noticed that it wasn't added to the docs. As the docs are built only for master we will see if it fails on after merge.

codewitch-honey-crisis:
Oh I forgot about the index. Sorry, sometimes when I'm navigating this maze it's easy for me to overlook something.

kisvegabor:
At least now we have everything in one repo. Imagine when there were an `lvgl`, an `examples` and a `docs` repo... :smile: 

Aaaand merging, thank you very much again!

kisvegabor:
Now it's in the docs. https://docs.lvgl.io/master/libs/tiny_ttf.html

I've just added `lv_example_tiny_ttf_2` it should be there in  few minutes too.

PR #3710, opened 2022-09-29
Feature/release new widget radio button

nicusorcitu:

The Radio button is basically a rectangle with vertical layout to which buttons (modified from [Checkbox](/widgets/core/checkbox)) can be added, and at same time only one button can be selected.

### Checkpoints

### DIFF
--- a/docs/example_list.py
+++ b/docs/example_list.py
@@ -54,6 +54,7 @@ widgets = {
 "menu":"Menu",
 "meter":"Meter",
 "msgbox":"Message box",
+"radiobtn": "Radio button",
 "roller":"Roller",
 "slider":"Slider",
 "span":"Span",
--- a/docs/widgets/radiobtn.md
+++ b/docs/widgets/radiobtn.md
@@ -1,0 +1,71 @@
+# Radio button (lv_radiobtn)
+
+## Overview
+The Radio button is basically a rectangle with vertical layout to which buttons(modified from [Checkbox](/widgets/core/checkbox)) can be added, and at same time only one button can be selected.
+
+## Parts and Styles
+
+**Background**
+- `LV_PART_MAIN` The main part of the radio button that uses all the typical background properties
+
+**Buttons**
+- `LV_PART_MAIN` The main part of the single button that uses all the typical background properties
+- `LV_PART_INDICATOR` The "tick box" is a circle that uses all the typical background style properties.
+By default, its size is equal to the height of the main part's font. Padding properties make the tick box larger in the respective directions.
+- `LV_PART_CUSTOM_FIRST` the inner box inside "tick box" that uses all the typical background style properties, use to show the button being selected.
+
+## Usage
+
+### Buttons
+`lv_radiobtn_add_item(radiobtn, text)` adds an button with text.
+
+The text of button can be modified with the `lv_radiobtn_set_item_text(rb_item, "New text")` function and will be dynamically allocated.
+
+To set a static text,
+use `lv_radiobtn_set_item_static_text(rb_item, txt)`. This way, only a pointer to `txt` will be stored. The text then shouldn't be deallocated while the button exists.
+
+#### Check, uncheck, disable
+You can manually check, un-check, and disable the single button by using the common state add/clear function:
+```c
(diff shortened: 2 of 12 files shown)


nicusorcitu:
@WenbinYuanNXP 
Please have a look and fix the PR failures.
Thanks!

Wenbin-YUAN:
> 
Okay, I already start to fix failure issues

nicusorcitu:
Hi @WenbinYuanNXP

There are other errors you have to fix: https://github.com/lvgl/lvgl/actions/runs/3158949101/jobs/5141569818#step:10:2219

Wenbin-YUAN:
> Hi @WenbinYuanNXP
> 
> There are other errors you have to fix: https://github.com/lvgl/lvgl/actions/runs/3158949101/jobs/5141569818#step:10:2219

Yes, I am working on it.

kisvegabor:
Having radio button functionality might be useful in a lot of cases. So I wonder if we can create a public function which can be used in a generic way. 

For example, add an event handler to the parent similarly to this: https://docs.lvgl.io/master/widgets/checkbox.html#checkboxes-as-radio-buttons

I agree to have a dedicated radio button widget because it needs a unique appearance however it might be useful to make this functionality publicly available. 

An other note, at first look radio button is very similar to checkbox. So have you considered basing it on checkbox? Similarly as bar is based on slider: https://github.com/lvgl/lvgl/blob/master/src/widgets/slider/lv_slider.c#L49

Wenbin-YUAN:
@kisvegabor 

Thanks for your comment. 
At beginning, I did get the idea of radio button from this example of checkbox then refer the list widget during development. 
> For example, add an event handler to the parent similarly to this: https://docs.lvgl.io/master/widgets/checkbox.html#checkboxes-as-radio-buttons
>
> I agree to have a dedicated radio button widget because it needs a unique appearance however it might be useful to make this functionality publicly available.

But I am a little confused about "_make this functionality publicly available_", could you please explain more?
> An other note, at first look radio button is very similar to checkbox. So have you considered basing it on checkbox? Similarly as bar is based on slider: https://github.com/lvgl/lvgl/blob/master/src/widgets/slider/lv_slider.c#L49

Yes, actually the codes of radiobtn_item is modified from checkbox because at early time we thought that decoupling checkbox is a good choice. but you are right based on checkbox might be better, I will try to do it.

kisvegabor:
> But I am a little confused about "make this functionality publicly available", could you please explain more?

I'm not sure what would be the best way to do it either. Maybe besides `lv_radiobutton_create(parent);` we can have `lv_obj_add/remove_radiobutton_manager(parent);` functions too. 

In `lv_radiobutton_create` `lv_obj_add_radiobutton_manager(parent);` is called which adds an event to the `parent` to handle all children with `LV_OBJ_FLAG_CHECKABLE` as radio button. 
 
We should add the event handler only once so `lv_obj_remove_event_cb` also should be called. Or we can add a new function like `lv_obj_has_event_cb(obj, event_cb)` to check if the object already has a given event. 

Regardless to `lv_radiobutton` the user still can use `lv_obj_add/remove_radiobutton_manager(parent);` to make any children behave as radio buttons  regardless to their type.

What do you think about it?

Wenbin-YUAN:
> In `lv_radiobutton_create` `lv_obj_add_radiobutton_manager(parent);` is called which adds an event to the `parent` to handle all children with `LV_OBJ_FLAG_CHECKABLE` as radio button.
> 
> Regardless to `lv_radiobutton` the user still can use `lv_obj_add/remove_radiobutton_manager(parent);` to make any children behave as radio buttons regardless to their type.
> 
Thanks, now I understand what's public function in this case, we could make a function to control not only `lv_radiobutton` but all children behave as radio button.

And yes, I believe it's a good idea to make this more useful, I will continue work on it.

kisvegabor:
Great, let me know if I can help with something. :slightly_smiling_face:

Wenbin-YUAN:
> Great, let me know if I can help with something. 🙂

Thanks, I was worked in my works, now I will start working on this widget.

Wenbin-YUAN:
for past one month I was busy but I tried change base class and create lv_obj_has_event(), its' working. I might continue to work on this 1 week later.

PR #3716, opened 2022-10-01
fix(tiny_ttf): add missing gc root

amirgon:
Replace a static local variable in ttf_get_glyph_bitmap_cb with a gc root, to prevent gc from collecting it prematurely

Related: https://github.com/lvgl/lvgl/issues/3535#issuecomment-1264404099


embeddedt (approved):
Just waiting for CI to confirm.

PR #3727, opened 2022-10-06
fix(style) prevent overflow in the style count

martaaay:

Every time lv_obj_add_style is called on an object it increments style_cnt on that object. Once style_cnt hits 64, it wraps to 0 because 6 bits are allocated for it. In the call to realloc, style_cnt==0 is used in the realloc size calculation and lv_realloc returns the address of code as a sentinel in that case. If the return value from lv_realloc is not checked and data is written to the returned address, then code is stomped. This update adds a quick assert to ensure a fast fail in the event that somebody calls lv_obj_add_style many times (likely already a bug) so that the failure mode is more instant and clear. A further fix would be to audit all lv_realloc (and probably lv_alloc) calls and ensure the data returned is going to cause a fast crash (e.g., return NULL or NULL+1 as a sentinel).

### Checkpoints
- [n/a] Update the documentation

### DIFF
--- a/src/core/lv_obj_style.c
+++ b/src/core/lv_obj_style.c
@@ -75,6 +75,8 @@ void _lv_obj_style_init(void)
 
 void lv_obj_add_style(lv_obj_t * obj, const lv_style_t * style, lv_style_selector_t selector)
 {
+    LV_ASSERT(obj->style_cnt < 63);
+
     trans_del(obj, selector, LV_STYLE_PROP_ANY, NULL);
 
     uint32_t i;


embeddedt:
Most platforms have the code segment as read-only (either because it's in flash, or because it's enforced by memory protection) so normally the `&zero_mem` case should cause a fairly quick crash if a write occurs.

embeddedt (approved):

martaaay:
Yesss, i was wondering about exactly that. I'll check into that. I guess I've only worked on weird platforms where we explicitly wanted to be able to edit the code regions haha (my prior experience was developing the JVM... so hacks on top of hacks all in the name of squeaking more optimizations).

martaaay:
So maybe I've not had enough coffee yet, but looking into the sentinel further...

zero_mem is defined as:
```c
static uint32_t zero_mem = ZERO_MEM_SENTINEL; /*Give the address of this variable if 0 byte should be allocated*/
```
So, wouldn't that be writable? If I pass the address to somebody else, they could then write on this space. (I easily could be not understanding right!)

Looking at the space around zero_mem and printed it as instructions, I see things like _lv_txt_get_encoded_length, which is not code, but a pointer to code. So perhaps I'm stomping on those pointers to code, which then means I'm doomed eventually when i call through one of these squashed function pointers.
```
(gdb) x/8i obj->styles
   0x3fc9d968 <zero_mem>:	ee.vldbc.32.ip	q7, a13, 0x10c
   0x3fc9d96b <zero_mem+3>:	l32r	a10, 0x3fc5db60
   0x3fc9d96e:	.byte 00
   0x3fc9d96f:	addx2	a13, a8, a0
   0x3fc9d972 <_lv_txt_get_encoded_length+2>:	l32r	a0, 0x3fc89a7c
   0x3fc9d975 <_lv_txt_encoded_get_char_id+1>:	xorb	b0, b1, b9
   0x3fc9d978 <_lv_txt_encoded_get_byte_id>:	.byte 0x84
   0x3fc9d979 <_lv_txt_encoded_get_byte_id+1>:	xorb	b0, b1, b9
```

kisvegabor:
Thank for the PR!

A little bit of clarification on `zero_mem`. The idea is to return something even if 0 byte  is allocated. It's because 0 bytes is not that different from 1 byte. It's still some kind of memory just with 0 length (and not `NULL`). Similarly 0 kilometer is still a distance, and not nothing. So `lv_malloc(0)` returns `&zero_mem`, a writable variable. As it's writable we can initialize it to a sentinel value and check if something was written into it by accident. That's it.

PR #3728, opened 2022-10-07
feature: ported arabic keyboard from release 7.10.0

gianlucacornacchia:
### ported arabic keyboard from release 7.10.0

Arabic keyboard support was added to the release 7.10.0, but in the following releases was dropped. This is a porting of that code to the release 8.3.3.

### Checkpoints


kisvegabor:
Looks great! Thank you very much!

embeddedt:
Gabor, shouldn't this change be added to `master`, since it's a new feature?

kisvegabor:
Uh, I missed that it's not for `master`. What do you think about transferring this commit to `master`? I.e. remove from here, and add to `master` only.

embeddedt:
Sounds good. We should be able to cherry-pick and then revert the change here (this way we avoid a force push).

kisvegabor:
Done, as you suggested.

PR #3740, opened 2022-10-12
Fix push to the component registry on tag

kumekay:
Run `esp_upload_component.yml` only when a tag is pushed. 

Previously job failed, because `actions/checkout` by default creates a shallow copy and doesn't fetch tags. 
There are 2 ways to fix it:
1. Use the `fetch-depth: '0'` option for checkout action, but this will result in extra work being done on every commit.
2. A leaner solution is to run the workflow only on tags and get tag names from `github.ref_name`  

+ Versions of actions are now fixed, to avoid breaking changes in the future.


kumekay:
@tore-espressif FYI

kisvegabor:
In think running the workflow only on `v*` tags is fine. It means 1 run/1-2 month.

Issue #3721, opened 2022-10-04
lv_obj_set_content_width & lv_obj_set_content_height don't take border side into account


Johennes:
### Perform all steps below and tick them with [x]
- [x] Read the [FAQ](https://docs.lvgl.io/master/intro/index.html#faq)
- [x] Check the related part of the [Documentation](https://docs.lvgl.io/)
- [x] Update lvgl to the latest version
- [ ] Reproduce the issue in a [Simulator](https://docs.lvgl.io/latest/en/html/get-started/pc-simulator.html)

I haven't spent time reproducing this in a simulator but I'm fairly certain that I see the issue in the source code.

### Describe the bug

The functions [`lv_obj_set_content_width`](https://github.com/lvgl/lvgl/blob/master/src/core/lv_obj_pos.c#L249) and [`lv_obj_set_content_height`](https://github.com/lvgl/lvgl/blob/master/src/core/lv_obj_pos.c#L258) subtract the border width from the object width but they don't check which border sides are hidden. As a result, when only the top border is active, the border width will erroneously be subtracted twice.

### To Reproduce

I spotted this in a keyboard widget, styled with

```
lv_style_set_border_side(&(styles.keyboard), LV_BORDER_SIDE_TOP);
lv_style_set_border_width(&(styles.keyboard), lv_dpx(theme->keyboard.border_width));
```
The button matrix inside the keyboard leaves empty space on the right and bottom side because the available width and height were calculated too low due to the border width subtractions.

### Expected behavior

The border width should only be subtracted when the respective border side is active, e.g.

```
lv_coord_t lv_obj_get_content_width(const lv_obj_t * obj)
{
    LV_ASSERT_OBJ(obj, MY_CLASS);

    lv_coord_t left = lv_obj_get_style_pad_left(obj, LV_PART_MAIN);
    lv_coord_t right = lv_obj_get_style_pad_right(obj, LV_PART_MAIN);
    lv_coord_t border_width = lv_obj_get_style_border_width(obj, LV_PART_MAIN);

    lv_border_side_t border_side = lv_obj_get_style_border_side(obj, LV_PART_MAIN);

    lv_coord_t width = lv_obj_get_width(obj) - left - right;
    
    if (border_side & LV_BORDER_SIDE_LEFT) {
        width -= border_width;
    }
    if (border_side & LV_BORDER_SIDE_RIGHT) {
        width -= border_width;
    }

    return width;
}
```

### Screenshots or video

[...]

PR #3744, opened 2022-10-14
fix: Take border side into account in flex and scrolled objects

Johennes:

This fixes an issue similar to #3721 that I found in flex / scrolled objects. The root cause was that the borer side wasn't taken into account. With the new utility functions added in #3721 this luckily was easy to fix. I did fix a few similar cases along the way. Let me know if you'd want me to take these out.

### Checkpoints


kisvegabor:
Very nice! All looks good, thank you!

With these change the core positioning part already considers the border visibility. Most of the remaining issues should be local to the widgets.

PR #3753, opened 2022-10-21
fix(mem): add missing stdio.h and string.h include

FASTSHIFT:

1. Add missing stdio.h and string.h include.
2. Fix wrong implementation of `lv_strncpy_builtin()`, see: https://cplusplus.com/reference/cstring/strncpy.

### Checkpoints

### DIFF
--- a/lv_conf_template.h
+++ b/lv_conf_template.h
@@ -57,6 +57,8 @@
 #endif  /*LV_USE_BUILTIN_SNPRINTF*/
 
 #define LV_STDLIB_INCLUDE <stdint.h>
+#define LV_STDIO_INCLUDE  <stdint.h>
+#define LV_STRING_INCLUDE <stdint.h>
 #define LV_MALLOC       lv_malloc_builtin
 #define LV_REALLOC      lv_realloc_builtin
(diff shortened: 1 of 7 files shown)


W-Mai:
That's great for me 👍

kisvegabor:
The prototype of `strncpy` is really like as you suggested. Thanks!

GorgonMeducer:
@kisvegabor Could you please review this PR again?
Something doesn't look right.

https://github.com/lvgl/lvgl/commit/648d1cbd03e1bf466aab0f9d8e5f32d1c7035ac3

Maybe we should update the lv_printf.c and use `_lv_vsnprintf` to replace `_vsnprintf`

https://github.com/lvgl/lvgl/blob/39db14d75e612a25cf8d3478d86c7fff9a680b66/src/misc/lv_printf.c#L550

Otherwise, there will be some conflicts between this `_vsnprintf` and the one defined in stdio.h:

image

kisvegabor:
It's strange that neither the CI nor my local builds doesn't produce any issues. Is `_vsnprintf` as non standard API from Keil's std lib?

Anyway, I fixed it as you suggested.

PR #3757, opened 2022-10-22
feat(tlsf): add memory pool expansion support

FASTSHIFT:

It is convenient for users to expand the memory pool, such as PSRAM.

Usage:
```c
#include "lvgl/src/misc/lv_malloc_builtin.h"

void mem_init(void)
{
    /* get external PSRAM information */
    psram_info_t info;
    psram_get_info(&info);

    /* add to memory builtin pool*/
    lv_mem_builtin_add_pool(info.address, info.size);
}
```

### Checkpoints


FASTSHIFT:
@xiaoxiang781216
Does `lv_mem_builtin_add_pool()` need to automatically align memory addresses set by users. Because TLSF checks for memory alignment.

lv_conf_template.h
```diff
@@ -38,6 +38,9 @@
     /*Size of the memory available for `lv_malloc()` in bytes (>= 2kB)*/
     #define LV_MEM_SIZE (128U * 1024U)          /*[bytes]*/
 
+    /*Size of the TLSF pool maximum size(must be greater than LV_MEM_SIZE)*/
+    #define LV_TLSF_MAX_POOL_SIZE (128U * 1024U) /*[bytes]*/
```
kisvegabor: We don't expose that we use TLSF internally. So it should be 
`LV_MEM_MAX_POOL_SIZE`. 

In the comment we should also explain what does it do. What is a "pool"? 

Can we set it to `LV_MEM_SIZE` by default?

kisvegabor:
So more pools can added/removed dynamically and TLSF will see them as one?

lv_conf_template.h
FASTSHIFT: I think we can use the value of `LV_MEM_POOL_EXPAND_SIZE` as the expansion size of `LV_MEM_SIZE`. If the user does not need this function, set `LV_MEM_POOL_EXPAND_SIZE` to 0.

FASTSHIFT:
> So more pools can added/removed dynamically and TLSF will see them as one?

Yes, as a user, it is insensitive to use `lv_malloc()`/`lv_free()`. When the first pool cannot be allocated, it will be automatically allocated from the second pool.

FASTSHIFT:
The current memory monitor does not seem to cover the second memory pool.

lv_conf_template.h
kisvegabor: Seems much clearer to me.

kisvegabor:
> The current memory monitor does not seem to cover the second memory pool.

If I understood correctly it's not about only the "second" memory pool as any number of memory pools can be added later.

I'm thinking about one thing: how mush nondeterminism will be added if the user adds a PSRAM memory pool next to the original on chip SRAM? I mean PSRAM is slower and probably has caching. It might lead to hard to debug performance issues.

FASTSHIFT:
> > The current memory monitor does not seem to cover the second memory pool.
> 
> If I understood correctly it's not about only the "second" memory pool as any number of memory pools can be added later.
> 
> I'm thinking about one thing: how mush nondeterminism will be added if the user adds a PSRAM memory pool next to the original on chip SRAM? I mean PSRAM is slower and probably has caching. It might lead to hard to debug performance issues.

I think this question can be left to the user to choose. In addition, the externally expanded memory pool may also be SRAM, because the SRAM memory addresses of some chips may be discontinuous. Using a single memory manager can reduce the difficulty of management.

kisvegabor:
Okay. All clear from my side. Is it ready to merge? ([This question](https://github.com/lvgl/lvgl/pull/3757#issuecomment-1287662965) still seems open)

FASTSHIFT:
> Okay. All clear from my side. Is it ready to merge? ([This question](https://github.com/lvgl/lvgl/pull/3757#issuecomment-1287662965) still seems open)

@xiaoxiang781216  said the job of memory alignment can be left to the user.

kisvegabor:
Ok, merging, thank you!

PR #3723, opened 2022-10-05
Unit tests for Animation Image widget

pablojr:
Added a set of test cases for the Animation Image widget.
Based on the example for that widget.
Asserts are simple because widget has not getters and I didn´t like to check by directly accessing internal details (which might change in the future).

Code coverage
               | Exec | Total | Coverage
Lines            39 | 39     | 100.0%
Branches        0 |      0 |          -%

PR #3773, opened 2022-10-28
feat(animimg): add getters for properties: duration and repeat count

pablojr:
Add missing getters for properties duration and repeat count in widget animimg

This addition will improve the way unit tests (provided later in PR #3723) can be written (without messing with underlying implementation details)


kisvegabor:
All look good. Thank you!

PR #3778, opened 2022-10-31
feat(libs): add barcode widget

FASTSHIFT:

Ported from: https://github.com/fhunleth/code128

image

### Checkpoints
 
### DIFF
--- a/examples/libs/barcode/lv_example_barcode_1.py
+++ b/examples/libs/barcode/lv_example_barcode_1.py
@@ -1,0 +1,21 @@
+#!/opt/bin/lv_micropython -i
+import lvgl as lv
+import display_driver
+
+bg_color = lv.palette_lighten(lv.PALETTE.LIGHT_BLUE, 5)
+fg_color = lv.palette_darken(lv.PALETTE.BLUE, 4)
+
+barcode = lv.barcode(lv.scr_act())
+barcode.set_height(50)
+barcode.center()
+
+# Set color
+barcode.set_dark_color(fg_color)
+barcode.set_light_color(bg_color)
+
+# Add a border with bg_color
+barcode.set_style_border_color(bg_color, 0)
+barcode.set_style_border_width(5, 0)
+
+# Set data
+barcode.update("https://lvgl.io")
--- a/Kconfig
+++ b/Kconfig
@@ -951,6 +951,9 @@ menu "LVGL configuration"
 		config LV_USE_QRCODE
 			bool "QR code library"
 
+		config LV_USE_BARCODE
+			bool "Barcode library"
+
 		config LV_USE_FREETYPE
 			bool "FreeType library"
 		if LV_USE_FREETYPE
--- a/docs/libs/barcode.md
+++ b/docs/libs/barcode.md
@@ -1,0 +1,30 @@
+
+# Barcode
+
+Barcode generation with LVGL. Uses [code128](https://github.com/fhunleth/code128) by [fhunleth](https://github.com/fhunleth).
+
+## Usage
+
+Enable `LV_USE_BARCODE` in `lv_conf.h`.
+
+Use `lv_barcode_create()` to create a barcode object, and use `lv_barcode_update()` to generate a barcode.
+
+Call `lv_barcode_set_scale()` or `lv_barcode_set_dark/light_color()` to adjust scaling and color, and call `lv_barcode_update()` again to regenerate the barcode.
+
+## Notes
+- It is best not to manually set the width of the barcode, because when the width of the object is lower than the width of the barcode, the display will be incomplete due to truncation.
+- The scale adjustment can only be an integer multiple, for example, `lv_barcode_set_scale(barcode, 2)` means 2x scaling.
+
+## Example
+```eval_rst
+
+.. include:: ../../examples/libs/barcode/index.rst
+
+```
+
+## API
+
+```eval_rst
+
+.. doxygenfile:: lv_barcode.h
+  :project: lvgl
--- a/examples/libs/barcode/index.rst
+++ b/examples/libs/barcode/index.rst
@@ -1,0 +1,6 @@
+Create a Barcode
+"""""""""""""""""""""""""""""""""""""""""""""""
+
+.. lv_example:: libs/barcode/lv_example_barcode_1
+  :language: c
+
--- a/examples/libs/barcode/lv_example_barcode.h
+++ b/examples/libs/barcode/lv_example_barcode.h
@@ -1,0 +1,38 @@
+/**
+ * @file lv_example_barcode.h
+ *
+ */
+
+#ifndef LV_EXAMPLE_BARCODE_H
+#define LV_EXAMPLE_BARCODE_H
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+/*********************
+ *      INCLUDES
+ *********************/
+
+/*********************
+ *      DEFINES
+ *********************/
+
+/**********************
+ *      TYPEDEFS
+ **********************/
+
+/**********************
+ * GLOBAL PROTOTYPES
+ **********************/
+void lv_example_barcode_1(void);
+
+/**********************
+ *      MACROS
+ **********************/
+
+#ifdef __cplusplus
+} /*extern "C"*/
+#endif
+
+#endif /*LV_EXAMPLE_BARCODE_H*/
--- a/examples/libs/barcode/lv_example_barcode_1.c
+++ b/examples/libs/barcode/lv_example_barcode_1.c
@@ -1,0 +1,28 @@
+#include "../../lv_examples.h"
+#if LV_USE_BARCODE && LV_BUILD_EXAMPLES
+
+/**
+ * Create a Barcode
+ */
+void lv_example_barcode_1(void)
+{
+    lv_color_t bg_color = lv_palette_lighten(LV_PALETTE_LIGHT_BLUE, 5);
+    lv_color_t fg_color = lv_palette_darken(LV_PALETTE_BLUE, 4);
+
+    lv_obj_t * barcode = lv_barcode_create(lv_scr_act());
+    lv_obj_set_height(barcode, 50);
+    lv_obj_center(barcode);
+
+    /*Set color*/
+    lv_barcode_set_dark_color(barcode, fg_color);
+    lv_barcode_set_light_color(barcode, bg_color);
+
+    /*Add a border with bg_color*/
+    lv_obj_set_style_border_color(barcode, bg_color, 0);
+    lv_obj_set_style_border_width(barcode, 5, 0);
+
+    /*Set data*/
+    lv_barcode_update(barcode, "https://lvgl.io");
+}
+
+#endif
--- a/examples/libs/lv_example_libs.h
+++ b/examples/libs/lv_example_libs.h
@@ -13,6 +13,7 @@ extern "C" {
 /*********************
  *      INCLUDES
  *********************/
+#include "barcode/lv_example_barcode.h"
 #include "bmp/lv_example_bmp.h"
 #include "ffmpeg/lv_example_ffmpeg.h"
 #include "freetype/lv_example_freetype.h"
--- a/lv_conf_template.h
+++ b/lv_conf_template.h
@@ -633,6 +633,9 @@
 /*QR code library*/
 #define LV_USE_QRCODE 0
 
+/*Barcode code library*/
+#define LV_USE_BARCODE 0
+
 /*FreeType library*/
 #define LV_USE_FREETYPE 0
 #if LV_USE_FREETYPE
--- a/lvgl.h
+++ b/lvgl.h
@@ -84,6 +84,7 @@ extern "C" {
 #include "src/others/ime/lv_ime_pinyin.h"
 #include "src/others/file_explorer/lv_file_explorer.h"
 
+#include "src/libs/barcode/lv_barcode.h"
 #include "src/libs/bmp/lv_bmp.h"
 #include "src/libs/fsdrv/lv_fsdrv.h"
 #include "src/libs/png/lv_png.h"
--- a/src/libs/barcode/code128.c
+++ b/src/libs/barcode/code128.c
@@ -1,0 +1,580 @@
+// Copyright (c) 2013, LKC Technologies, Inc.
+// All rights reserved.
+//
+// Redistribution and use in source and binary forms, with or without
+// modification, are permitted provided that the following conditions are met:
+//
+// Redistributions of source code must retain the above copyright notice, this
+// list of conditions and the following disclaimer. Redistributions in binary
+// form must reproduce the above copyright notice, this list of conditions and
+// the following disclaimer in the documentation and/or other materials
+// provided with the distribution. THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT
+// HOLDERS AND CONTRIBUTORS "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES,
+// INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND
+// FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE
+// COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT,
+// INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+// LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA,
+// OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF
+// LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING
+// NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE,
+// EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+#include "../../../lvgl.h"
+#if LV_USE_BARCODE
+
+#include "code128.h"
+#include <string.h>
+
+#define CODE128_MALLOC      lv_malloc
+#define CODE128_REALLOC     lv_realloc
+#define CODE128_FREE        lv_free
+#define CODE128_MEMSET      lv_memset
+#define CODE128_STRLEN      lv_strlen
+#define CODE128_ASSERT      LV_ASSERT
+
+#define CODE128_QUIET_ZONE_LEN 10
+#define CODE128_CHAR_LEN       11
+#define CODE128_STOP_CODE_LEN  13
+
+#define CODE128_START_CODE_A 103
+#define CODE128_START_CODE_B 104
+#define CODE128_START_CODE_C 105
+
+#define CODE128_MODE_A    'a'
+#define CODE128_MODE_B    'b'
+#define CODE128_MODE_C    'c'
+
+#define CODE128_MIN_ENCODE_LEN (CODE128_QUIET_ZONE_LEN * 2 + CODE128_CHAR_LEN * 2 + CODE128_STOP_CODE_LEN)
+
+static const int code128_pattern[] = {
+    // value: pattern,     bar/space widths
+    1740, //   0: 11011001100, 212222
+    1644, //   1: 11001101100, 222122
+    1638, //   2: 11001100110, 222221
+    1176, //   3: 10010011000, 121223
+    1164, //   4: 10010001100, 121322
+    1100, //   5: 10001001100, 131222
+    1224, //   6: 10011001000, 122213
+    1220, //   7: 10011000100, 122312
+    1124, //   8: 10001100100, 132212
+    1608, //   9: 11001001000, 221213
+    1604, //  10: 11001000100, 221312
+    1572, //  11: 11000100100, 231212
+    1436, //  12: 10110011100, 112232
+    1244, //  13: 10011011100, 122132
+    1230, //  14: 10011001110, 122231
+    1484, //  15: 10111001100, 113222
+    1260, //  16: 10011101100, 123122
+    1254, //  17: 10011100110, 123221
+    1650, //  18: 11001110010, 223211
+    1628, //  19: 11001011100, 221132
+    1614, //  20: 11001001110, 221231
+    1764, //  21: 11011100100, 213212
+    1652, //  22: 11001110100, 223112
+    1902, //  23: 11101101110, 312131
+    1868, //  24: 11101001100, 311222
+    1836, //  25: 11100101100, 321122
+    1830, //  26: 11100100110, 321221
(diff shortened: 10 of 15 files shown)


examples/libs/barcode/lv_example_barcode_1.py
```diff
@@ -0,0 +1,21 @@
+#!/opt/bin/lv_micropython -i
+import lvgl as lv
+import display_driver
+
+bg_color = lv.palette_lighten(lv.PALETTE.LIGHT_BLUE, 5)
+fg_color = lv.palette_darken(lv.PALETTE.BLUE, 4)
+
+barcode = lv.barcodecode(lv.scr_act())
```
kisvegabor:
```suggestion
barcode = lv.barcode(lv.scr_act())
```

kisvegabor (commented):
Very nice! Do you have a our test environment set locally? If you could you add simple [rendering tests](https://github.com/lvgl/lvgl/blob/master/tests/README.md#asserts) with a few bar codes with different colors?

kisvegabor:
@amirgon Do you have any idea why the MP CI fails?

FASTSHIFT:
@kisvegabor 
`LV_USE_BARCODE` is 0 by default. How to automatically enable it to test and generate reference images?

kisvegabor:
Please enable it here:
https://github.com/lvgl/lvgl/blob/b01229e451289ae9c64c51897bfb70045227080b/tests/CMakeLists.txt#L249

amirgon:
> @amirgon Do you have any idea why the MP CI fails?

Yes. `lv_conf.h` didn't enable `LV_USE_BARCODE` therefore "barcode" widget wasn't built for the Micropython CI.
I've updated lv_micropython to enable it, please try to re-run the test.

In the longer term, it might make sense to move Micropython's lv_conf.h to lvgl repo, such that an LVGL PR could update it. Otherwise additional commits to [lv_binding_micropython](https://github.com/lvgl/lv_binding_micropython/commit/c544c51c97b38907709a0cd7ea7ca8ab96410cbb) and [lv_micropython](https://github.com/lvgl/lv_micropython/commit/b8dde900fbe08bde418b7d7c1c51f6c060c355ab) are required.

embeddedt:
> In the longer term, it might make sense to move Micropython's lv_conf.h to lvgl repo, such that an LVGL PR could update it. 

I agree; we seem to have this problem regularly.

kisvegabor:
> In the longer term, it might make sense to move Micropython's lv_conf.h to lvgl repo, such that an LVGL PR could update it. Otherwise additional commits to [lv_binding_micropython](https://github.com/lvgl/lv_binding_micropython/commit/c544c51c97b38907709a0cd7ea7ca8ab96410cbb) and [lv_micropython](https://github.com/lvgl/lv_micropython/commit/b8dde900fbe08bde418b7d7c1c51f6c060c355ab) are required.

The idea is that the content of `lvgl` shouldn't be modified by the user. If we added an `lv_conf_mp.h` here the user needed to edit it directly. What about adding a CMake or Makefile for the MP tests which defines the correct settings for MP tests and `MP's lv_conf.h` wouldn't be not used for the tests?

amirgon:
> The idea is that the content of `lvgl` shouldn't be modified by the user. If we added an `lv_conf_mp.h`

But LVGL should provide some defaults, no?  
If someone adds a new widget, how can we test it if it's not enabled by default?

> What about adding a CMake or Makefile for the MP tests which defines the correct settings for MP tests and `MP's lv_conf.h` wouldn't be not used for the tests?

Do you mean add some CMake or Makefile in LVGL repo? Today all make logic is in the lv_micropython and lv_binding_micropython repos. Does it make sense to add Micropython make logic to LVGL?

How can we not use MP's lv_conf.h for the tests? Where will we take all the configuration from? How do we update the configuration without duplicating it across repos?

kisvegabor:
> If someone adds a new widget, how can we test it if it's not enabled by default?

From [`lv_conf_template.h`](https://github.com/lvgl/lvgl/blob/master/lv_conf_template.h) we generate [`lv_conf_internal.h`](https://github.com/lvgl/lvgl/blob/master/src/lv_conf_internal.h). So the defaults are in `lv_conf_template.h` and if no one set those defines `lv_conf_internal.h` will set the default value.

> Does it make sense to add Micropython make logic to LVGL?

As we have [CI test for MP](https://github.com/lvgl/lvgl/blob/master/.github/workflows/build_micropython.yml) also in the `lvgl` repo an `mk` or `h` file in the [`tests`](https://github.com/lvgl/lvgl/tree/master/tests) folder can be ok too.

> How can we not use MP's lv_conf.h for the tests? Where will we take all the configuration from? How do we update the configuration without duplicating it across repos?

With `-DLV_CONF_PATH=".../my_custom_conf.h"` we can use a custom `lv_conf.h` file. We already maintain a list of test configs  in [CMakeLists.txt](https://github.com/lvgl/lvgl/blob/master/tests/CMakeLists.txt) but it's CMake, not make. So we can't reuse them. :slightly_frowning_face:

amirgon:
> From [`lv_conf_template.h`](https://github.com/lvgl/lvgl/blob/master/lv_conf_template.h) we generate [`lv_conf_internal.h`](https://github.com/lvgl/lvgl/blob/master/src/lv_conf_internal.h). So the defaults are in `lv_conf_template.h` and if no one set those defines `lv_conf_internal.h` will set the default value.

But the default values are false... Here on this PR for example `lv_conf_template.h`:
```c
/*Barcode code library*/
#define LV_USE_BARCODE 0
```
We need it to be 1 for the tests to pass, so `lv_conf_template.h` doesn't help here.

> As we have [CI test for MP](https://github.com/lvgl/lvgl/blob/master/.github/workflows/build_micropython.yml) also in the `lvgl` repo an `mk` or `h` file in the [`tests`](https://github.com/lvgl/lvgl/tree/master/tests) folder can be ok too.

I still don't see the point in adding Micropython specific build scripts in LVGL.  
Wouldn't it make more sense to keep all build scripts in one place, and only pass them parameters?

> With `-DLV_CONF_PATH=".../my_custom_conf.h"` we can use a custom `lv_conf.h` file. We already maintain a list of test configs in [CMakeLists.txt](https://github.com/lvgl/lvgl/blob/master/tests/CMakeLists.txt) but it's CMake, not make. So we can't reuse them.

Why do you keep these test defaults in CMake and not in a simple plain H file?  
If you had something like `lv_conf_test.h` file which configures LVGL for tests, we could have used it directly for both Micropython and LVGL tests without code duplication and without any specialized make scripts.

kisvegabor:
Let me reiterate: instead of having CMake configs we could have an `lv_test_conf.h` file, in which based on the test settings (`-DTEST_TYPE`) we can `#if #endif` some parts. Actually we already have [`lv_test_conf.h`](https://github.com/lvgl/lvgl/blob/master/tests/src/lv_test_conf.h) but it's very simple now.  

The gist is that we use these `lv_test_conf.h` only for the tests (the values defined in `lv_test_conf.h` overwrite the default values).

amirgon:
> Let me reiterate: instead of having CMake configs we could have an `lv_test_conf.h` file, in which based on the test settings (`-DTEST_TYPE`) we can `#if #endif` some parts. Actually we already have [`lv_test_conf.h`](https://github.com/lvgl/lvgl/blob/master/tests/src/lv_test_conf.h) but it's very simple now.

Ok. Having a single `lv_test_conf.h` for both C and MP would make sense, to prevent code duplication.
It would probably need to undef existing defines, since most configurations in `lv_conf.h` are not conditioned.

kisvegabor:
> It would probably need to undef existing defines, since most configurations in lv_conf.h are not conditioned.

We can use only one `lv_conf.h` at a time. For tests it'd be `lv_test_conf.h` so it doesn't interfere with the normal `lv_conf.h`. 

So in summary:
- There is an `lv_test_conf.h` in `lvgl/tests`
- This `h` file has the configs for various test settings include MP tests
- MP test builds shall add `-DLV_CONF_PATH="lvgl/tests/lv_test_conf.h"`

What do you think?

amirgon:
> We can use only one `lv_conf.h` at a time. For tests it'd be `lv_test_conf.h` so it doesn't interfere with the normal `lv_conf.h`.

In case of Micropython we already have [lv_conf.h](https://github.com/lvgl/lv_binding_micropython/blob/master/lv_conf.h) which is custom made for the Micropython bindings. All we really need is to enable some widgets so we could test them.  
That's why I suggested that `lv_test_conf.h` would **override** it, not replace it.

Does it make any sense to maintain two copies of almost identical `lv_conf.h` files?  
We could easily avoid this code duplication in several ways:
- We could override macros, by undefining and redefining them to enable widgets for tests
- We could enable a widget in ifdef in lv_conf.h, so it takes effect only if it wasn't enabled/disabled already. We already use this technique today to allow [setting color depths from outside](https://github.com/lvgl/lv_binding_micropython/blob/c544c51c97b38907709a0cd7ea7ca8ab96410cbb/lv_conf.h#L27-L29) for example.

kisvegabor:
> In case of Micropython we already have [lv_conf.h](https://github.com/lvgl/lv_binding_micropython/blob/master/lv_conf.h) which is custom made for the Micropython bindings. All we really need is to enable some widgets so we could test them.
That's why I suggested that lv_test_conf.h would override it, not replace it.

I see your point. As LVGL has no config to handle multiple `lv_conf` files we could do something like this is `lv_test_conf.h`:
```c
#ifdef MICROPYTHON
  #include "../../../lv_conf.h"  //from lv_binding_micropython

  #undef LV_USE_BTN
  #undef LV_USE_LABEL
  #undef LV_USE_BARCODE
  ...
#endif

#if TEST_TYPE == 1
  #include "test_type_1_conf.h"   
#endif

#if TEST_TYPE == 2
  #include "test_type_2_conf.h"   
#endif
```  

This way all the extra work we need is to `undef` all the option we will set later in  `test_type_*_conf.h`.

What do you think?

amirgon:
> I see your point. As LVGL has no config to handle multiple `lv_conf` files we could do something like this is `lv_test_conf.h`:
> ...
> This way all the extra work we need is to `undef` all the option we will set later in `test_type_*_conf.h`.
> 
> What do you think?

In principle this looks good to me.

But I'm still not sure I understand the details:
- What kind of "test types" would we have? How would `test_type_1_conf.h` look like for example?
- I assume that these `test_type_X_conf.h` would only enable widgets, is that right? Or would they affect other configurations as well?
- Since these "test types" are set on compile time, only one of them would be relevant for Micropython right? Micropython has only a single compile time configuration. How would `TEST_TYPE` be configured when we build Micropython for tests?
- If there's only a single `TEST_TYPE` for Micropython testing, why do we need to undefine these macros globally and not in the context of that specific `TEST_TYPE`?

kisvegabor:
> What kind of "test types" would we have? How would test_type_1_conf.h look like for example?

These:
image




> I assume that these test_type_X_conf.h would only enable widgets, is that right? Or would they affect other configurations as well?

Enable widgets and other features. E.g. `LV_USE_FONT_SUBPX`

> Since these "test types" are set on compile time, only one of them would be relevant for Micropython right? Micropython has only a single compile time configuration. How would TEST_TYPE be configured when we build Micropython for tests?
If there's only a single TEST_TYPE for Micropython testing, why do we need to undefine these macros globally and not in the context of that specific TEST_TYPE?

You are right. Probably it should be better:
```c
#ifdef MICROPYTHON
  #include "../../../lv_conf.h"  //from lv_binding_micropython
#endif

#if TEST_TYPE == 1
  #include "test_type_1_conf.h"   
#endif

#if TEST_TYPE == 2
  #include "test_type_2_conf.h"   
#endif
```

`#undef` is not required as each test type has it' own `lv_conf.h` like H file.

amirgon:
> `#undef` is not required as each test type has it' own `lv_conf.h` like H file.

Are you sure?  
What if both the binding's `lv_conf.h` and the test's `lv_conf.h` define/undefine the same macro?

kisvegabor:
In MP test we use only the bindings `lv_conf.h`. It will serve as the test config. Just like now. 

But I think we lost the purpose (or maybe just me).  The original goal was to enable the new widgets only on one place. And it should be the `test_type_X_conf.h` files. So what about this:
```c
//Platform specific configs
#ifdef MICROPYTHON
  #include "../../../lv_conf.h"     //from lv_binding_micropython, memory config + enable widgets
#endif

#ifdef UNIX
  #include "unix_test_conf.h"        // Just memory and other basic settings 
#endif


#if TEST_TYPE == 1
  #include "test_disable_all.h"   //#undef and disable all features to have a clean start. Can be included form test_type_1_conf.h too
  #include "test_type_1_conf.h"   //#undef what it will overwrite
#endif

#if TEST_TYPE == 2
  #include "test_disable_all.h"   //#undef and disable all features to have a clean start. Can be included form test_type_2_conf.h too
  #include "test_type_2_conf.h"   
#endif
```

The MP test should `-DTEST_TYPE=5` (let's say in 5 everything is enabled).

amirgon:
> In MP test we use only the bindings `lv_conf.h`. It will serve as the test config. Just like now.

The problem is that when a new widget is added and the define is still missing in `lv_conf.h`. That was the original reason MP test failed on this PR.

> The MP test should `-DTEST_TYPE=5` (let's say in 5 everything is enabled).

Yes assuming it would also ` #include "test_disable_all.h"` first.
As a test that enables everything, how about not defining `TEST_TYPE` at all? (then check it with `#ifndef`)

kisvegabor:
> Yes assuming it would also #include "test_disable_all.h" first.

Yes, we'd keep only the memory and other platform specific settings.

> As a test that enables everything, how about not defining TEST_TYPE at all? (then check it with #ifndef)

It could also work. E.g.
```c
#if !defined(TEST_TYPE) || TEST_TYPE == 5
  #include "test_disable_all.h" 
  #include "test_type_5_conf.h"  
#endif
```

FASTSHIFT:
Having trouble with the MP test? Why do many PR tests fail.

amirgon:
> Having trouble with the MP test? Why do many PR tests fail.

Only this test fails here:
```
micropython-dev: ../../lib/lv_bindings/lvgl/src/libs/tiny_ttf/stb_truetype_htcw.h:1781: stbtt_FindGlyphIndex: Assertion `0' failed.
```

Could you try to rebase your PR on latest LVGL? 

---

@kisvegabor Latest fails because you fixed only C and not Python examples:

https://github.com/lvgl/lvgl/commit/a218d7540945612d476f88b3b13d6a7a77b7e4de causes https://github.com/lvgl/lvgl/actions/runs/3502652826/jobs/5867181576

kisvegabor:
> Latest fails because you fixed only C and not Python examples:

For some reason I haven't got email notification about the failing CI. Anyway, I've tried to fix it locally but got a strange error: 
```
$ ./run.sh 
Running the SDL lvgl version
Running ./../lvgl/examples/header.py ...
lvgl obj 
Running the SDL lvgl version
Running ./../lvgl/examples/event/lv_example_event_4.py ...
lvgl obj 
lvgl obj 
  File "./run_test.py", line 97, in run
  File "./run_test.py", line 47, in reraise
  File "lv_utils.py", line 123, in task_handler
  File "<string>", line 34, in event_cb
AttributeError: 'module' object has no attribute 'RADIUS_CIRCLE'
parallel: This job failed:
timeout 5m catchsegv ./../../../ports/unix/micropython-dev ./run_test.py ./../lvgl/examples/event/lv_example_event_4.py
```
I've confirmed that the built MicroPython recognizes `lv.RADIUS_CIRCLE`.

amirgon:
> I've tried to fix it locally but got a strange error:
> ...
> I've confirmed that the built MicroPython recognizes lv.RADIUS_CIRCLE.

I checked out this PR and `lv_example_event_4.py` works fine.  
I'm not sure why you are seeing an error related to `lv_example_event_4.py`. Could you try to clean and rebuild?
Moreover, I don't see your error on any GitHub workflow, not on this PR and not on LVGL master.

What I do see, is an error with `lv_example_meter_3.py` and `lv_demo_widgets.py`:
```
AttributeError: 'meter' object has no attribute 'add_scale'
```
(https://github.com/lvgl/lvgl/actions/runs/3508673019/jobs/5877241151, https://github.com/lvgl/lvgl/actions/runs/3508136393/jobs/5876333063)

and this makes sense, because on https://github.com/lvgl/lvgl/commit/a218d7540945612d476f88b3b13d6a7a77b7e4de you removed `lv_meter_add_scale`

kisvegabor:
> I'm not sure why you are seeing an error related to lv_example_event_4.py. Could you try to clean and rebuild?
Moreover, I don't see your error on any GitHub workflow, not on this PR and not on LVGL master.

It's a local issue for sure, but I don't know why. I've tried to clean and rebuild but didn't help.


> What I do see, is an error with lv_example_meter_3.py and lv_demo_widgets.py:

I've just fixed them. Now all tests pass.

kisvegabor:
@amirgon 
I've added the new header based tests in [feat/test-configs-from-headers](https://github.com/lvgl/lvgl/tree/feat/test-configs-from-headers).

Finally, it became slightly different from what we have discussed because it made more sense in the practice. 
I've added MP's memory configs directly in [`lv_test_conf.h`](https://github.com/lvgl/lvgl/compare/lvgl:3940f53...lvgl:38a9f11#diff-e282720b51ca36e99b2b086a97641a2df42d4c3793485cfeb0e5975e06dbdd51R43-R61).  This way we don't need to `#undef` all the options defined by MP's `lv_conf.h`. I'm not sure I'm right, but I think all we need to do is making the tests on GitHub workflow to not use MP's `lv_conf.h` but [`lv_test_conf.h`](https://github.com/lvgl/lvgl/blob/feat/test-configs-from-headers/tests/src/lv_test_conf.h) with something like
`-DLV_CONF_PATH=${LVGL_TEST_DIR}/src/lv_test_conf.h`.

If a new widget/feature comes we only need to add it in [`lv_test_conf_full.h`](https://github.com/lvgl/lvgl/compare/lvgl:3940f53...lvgl:38a9f11#diff-05b9ed1a8384e4ba1f41d99452a63b732cac4e16faa882d2051a5ff7341978f7).

What do you think?

kisvegabor:
Now it doesn't effect MicroPython so I merged the header based test to `master` in order to not block #3821.

kisvegabor:
@embeddedt After this (test config to header files) change the ARM test on def heap fails with timeout on stress test:
image
https://github.com/lvgl/lvgl/actions/runs/3547541843/jobs/5957774176

Do you have any idea what could be the problem? The tests have 8MB memory.

amirgon:
> I've added MP's memory configs directly in [`lv_test_conf.h`](https://github.com/lvgl/lvgl/compare/lvgl:3940f53...lvgl:38a9f11#diff-e282720b51ca36e99b2b086a97641a2df42d4c3793485cfeb0e5975e06dbdd51R43-R61)

@kisvegabor - That's a code duplication.  
If you do that, we'll always need to keep in mind that there are two different configuration files that need to be kept in sync, in two different repos. In my opinion this is confusing.  
For example, https://github.com/lvgl/lv_binding_micropython/pull/242 changes exacly this duplicated part.... Once it's merge, LVGL will break until we fix `lv_test_conf.h` on LVGL.
Another problem of this approach is dependencies. It requires specific versions of LVGL to match specific versions of lv_binding_micropython. Ideally we would want to be able to change LVGL submodule version in lv_binding_micropython and that it would just work. But if LVGL has Micropython configuration in it, that's another dependency that limits which versions of LVGL can work with which versions of lv_binding_micropython.   

So I would prefer to avoid this duplication and have only a single configuration.  
Since it's Micropython specific, I don't think it should be in LVGL repo.

There is another option, which is to define every configuration symbol only if it was not defined previously:
```c
#ifndef SYMBOL
#define SYNBOL ...
#endif
```
If all configuration macros everywhere were defined that way, it could have been very easy to override configuration just by including some header earlier, without the need to undef anything.

kisvegabor:
What about creating an `lv_conf_mem.h` file in MP which could be reused by the tests and MP itself too?

> There is another option, which is to define every configuration symbol only if it was not defined previously:

If I understand correctly it'd apply to MP's `lv_conf.h` too. It'd make the already long file even longer and harder to understand. 

What is the correct `define` to detect MicroPython environment? I used `#ifdef MICROPYTHON` but it's just placeholder, I don't know if it's correct.

amirgon:
> What about creating an `lv_conf_mem.h` file in MP which could be reused by the tests and MP itself too?

It's possible but doesn't solve the more general dependencies problem.  
If LVGL includes `lv_conf_mem.h` from Micropython then it assumes it exists and in specific path, which may or may not be true for a specific version of lv_binding_micropython.
Today we don't have assumptions in LVGL about Micropython, and this makes sense because LVGL is a library used by Micropython. It's better for a library to not have dependencies on the applications that use it.

> If I understand correctly it'd apply to MP's `lv_conf.h` too. It'd make the already long file even longer and harder to understand.

Yes it will be longer, but I don't think it will be harder to understand.  
In fact, on Micropython's `lv_conf.h` I've already started to do that [for some defines](https://github.com/lvgl/lv_binding_micropython/blob/c544c51c97b38907709a0cd7ea7ca8ab96410cbb/lv_conf.h#L27-L29).
It provides flexibility. For example, a user could provide any configuration parameter with the `-D` option on `CFLAGS`, even on the Makefile command line. [(`LV_CFLAGS="-DLV_COLOR_DEPTH=16"`)](https://github.com/lvgl/lv_micropython#esp32-port)

> What is the correct `define` to detect MicroPython environment? I used `#ifdef MICROPYTHON` but it's just placeholder, I don't know if it's correct.

I don't think we have such a define today.

kisvegabor:
> Today we don't have assumptions in LVGL about Micropython, and this makes sense because LVGL is a library used by Micropython. It's better for a library to not have dependencies on the applications that use it.

What about a `LV_MP_TEST_CONF_PATH` define? Once we run MicroPython tests this dependency seems acceptable to me.

> I don't think we have such a define today.

We could use `LV_MP_TEST_CONF_PATH` for this purpose.

amirgon:
> What about a `LV_MP_TEST_CONF_PATH` define? Once we run MicroPython tests this dependency seems acceptable to me.

What will be included in this `LV_MP_TEST_CONF_PATH`?  Only the macros that are specific to Micropython?  (gc, memory allocation)
What about `LV_EXPORT_CONST_INT` for example?

What you suggest can work, but requires me to split `lv_conf.h` in lv_bindings_micropython into two header files.  

In that case, it could be very useful for me to also change LVGL to have **two** header files instead of a single `lvgl/lv_conf_template.h`.   
That's because on every update I make to LVGL submodule in lv_bindings_micropython I always diff `lv_conf.h` vs. `lv_conf_template.h` to see what was changed or added there, then I update `lv_conf.h` accordingly. It would be harder for me to compare a single `lv_conf_template.h` to two header files in lv_bindings_micropython every time.

kisvegabor:
> What will be included in this LV_MP_TEST_CONF_PATH? Only the macros that are specific to Micropython? (gc, memory allocation)
What about LV_EXPORT_CONST_INT for example?

Yes, `LV_EXPORT_CONST_INT` too. 

Realistically how often do you except these system level defines to change? If once in a year or so I think it's manageable to have  a separate `lv_conf_test.h` where the system level defines are copied from the normal `lv_conf.h`. So it's a duplication as `lv_conf_test.h` is not included  in `lv_conf.h`.  But it's flexible, simple, and doesn't require much change.

FASTSHIFT:
@kisvegabor 
When is it expected to merge this PR? :)

kisvegabor:
Let's merge it now. We can discuss the test updates independently.

PR #3782, opened 2022-11-04
refact(freetype): clean up global variables and infrequent functions

FASTSHIFT:

1.Merge all global variables into `lv_freetype_context_t`.
2.Remove the support of cache size=0, because this performance is not good, it will hardly be used.
3.Clear the font pathname reference count multiplexing, this will not save much memory compared to freetype cache.
4.Canonical function naming `lv_ft_xxx` -> `lv_freetype_xxx`.
5.Remove `lv_ft_info_t` as input and output parameters, this is easy to make user error.
6.Optimize error log output, support viewing specific error causes.
7.Encapsulate `font_manager` on the basis of `lv_freetype`, including path management, `font-family` support, font resource reuse and recycling functions (implemented in the next PR).

### Checkpoints
 
### DIFF
--- a/src/libs/freetype/lv_freetype.h
+++ b/src/libs/freetype/lv_freetype.h
@@ -22,6 +22,7 @@ extern "C" {
 /**********************
  *      TYPEDEFS
  **********************/
+
 typedef enum {
     LV_FREETYPE_FONT_STYLE_NORMAL = 0,
     LV_FREETYPE_FONT_STYLE_ITALIC = 1 << 0,
@@ -32,12 +33,34 @@ typedef enum {
  * GLOBAL PROTOTYPES
  **********************/
 
+/**
+ * Initialize the freetype library.
+ * @param max_faces Maximum number of opened FT_Face objects managed by this cache instance. Use 0 for defaults.
+ * @param max_sizes Maximum number of opened FT_Size objects managed by this cache instance. Use 0 for defaults.
+ * @param max_bytes Maximum number of bytes to use for cached data nodes. Use 0 for defaults.
+ *                  Note that this value does not account for managed FT_Face and FT_Size objects.
+ * @return LV_RES_OK on success, otherwise LV_RES_INV.
+ */
 lv_res_t lv_freetype_init(uint16_t max_faces, uint16_t max_sizes, uint32_t max_bytes);
 
+/**
+ * Uninitialize the freetype library
+ */
 void lv_freetype_uninit(void);
 
-lv_font_t * lv_freetype_font_create(const char * pathname, uint16_t weight, lv_freetype_font_style_t style);
+/**
+ * Create a freetype font.
+ * @param pathname font file path.
+ * @param size font size.
+ * @param style font style(see lv_freetype_font_style_t for details).
+ * @return Created font, or NULL on failure.
+ */
+lv_font_t * lv_freetype_font_create(const char * pathname, uint16_t size, lv_freetype_font_style_t style);
 
+/**
+ * Delete a freetype font.
+ * @param font freetype font to be deleted.
+ */
 void lv_freetype_font_del(lv_font_t * font);
 
 /**********************
--- a/src/libs/freetype/lv_freetype.h
+++ b/src/libs/freetype/lv_freetype.h
@@ -23,52 +23,22 @@ extern "C" {
  *      TYPEDEFS
  **********************/
 typedef enum {
-    FT_FONT_STYLE_NORMAL = 0,
-    FT_FONT_STYLE_ITALIC = 1 << 0,
-    FT_FONT_STYLE_BOLD   = 1 << 1
-} LV_FT_FONT_STYLE;
-
-typedef struct {
-    const char * name;  /* The name of the font file */
-    const void * mem;   /* The pointer of the font file */
-    size_t mem_size;    /* The size of the memory */
-    lv_font_t * font;   /* point to lvgl font */
-    uint16_t weight;    /* font size */
-    uint16_t style;     /* font style */
-} lv_ft_info_t;
+    LV_FREETYPE_FONT_STYLE_NORMAL = 0,
+    LV_FREETYPE_FONT_STYLE_ITALIC = 1 << 0,
+    LV_FREETYPE_FONT_STYLE_BOLD   = 1 << 1
+} lv_freetype_font_style_t;
 
 /**********************
  * GLOBAL PROTOTYPES
  **********************/
 
-/**
- * init freetype library
- * @param max_faces Maximum number of opened FT_Face objects managed by this cache instance. Use 0 for defaults.
- * @param max_sizes Maximum number of opened FT_Size objects managed by this cache instance. Use 0 for defaults.
- * @param max_bytes Maximum number of bytes to use for cached data nodes. Use 0 for defaults.
- *                  Note that this value does not account for managed FT_Face and FT_Size objects.
- * @return true on success, otherwise false.
- */
-bool lv_freetype_init(uint16_t max_faces, uint16_t max_sizes, uint32_t max_bytes);
+lv_res_t lv_freetype_init(uint16_t max_faces, uint16_t max_sizes, uint32_t max_bytes);
 
-/**
- * Destroy freetype library
- */
-void lv_freetype_destroy(void);
+void lv_freetype_uninit(void);
 
-/**
- * Creates a font with info parameter specified.
- * @param info See lv_ft_info_t for details.
- *             when success, lv_ft_info_t->font point to the font you created.
- * @return true on success, otherwise false.
- */
-bool lv_ft_font_init(lv_ft_info_t * info);
+lv_font_t * lv_freetype_font_create(const char * pathname, uint16_t weight, lv_freetype_font_style_t style);
 
-/**
- * Destroy a font that has been created.
- * @param font pointer to font.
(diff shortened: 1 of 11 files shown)


src/libs/freetype/lv_freetype.h
```diff
@@ -23,52 +23,22 @@ extern "C" {
...
-/**
- * init freetype library
- * @param max_faces Maximum number of opened FT_Face objects managed by this cache instance. Use 0 for defaults.
- * @param max_sizes Maximum number of opened FT_Size objects managed by this cache instance. Use 0 for defaults.
- * @param max_bytes Maximum number of bytes to use for cached data nodes. Use 0 for defaults.
- *                  Note that this value does not account for managed FT_Face and FT_Size objects.
- * @return true on success, otherwise false.
- */
-bool lv_freetype_init(uint16_t max_faces, uint16_t max_sizes, uint32_t max_bytes);
+lv_res_t lv_freetype_init(uint16_t max_faces, uint16_t max_sizes, uint32_t max_bytes);
```
kisvegabor, line 44: Please do not remove the API comments.

kisvegabor (commented):
@embeddedt What do you think about adding `freetype` to the CI? The only issue I see is it will be a extra dependencies and it might be complicated to install FreeType on Windows.

embeddedt:
> it might be complicated to install FreeType on Windows

Do you mean for potentially running tests on Windows in the future, or for test case development?

kisvegabor:
I mentioned Windows as some LVGL developers are using Windows and it's already difficult for them to the tests. Or can WSL2  manage `sudo apt-get install freetype` and we can say we support only Linux for testing?

embeddedt:
> Or can WSL2 manage sudo apt-get install freetype and we can say we support only Linux for testing?

This is what I was thinking, yes. I find Linux distros' package management to be much simpler to work with than that of Windows. :laughing: 

Another option would be to detect the presence of FreeType in CMake and only enable the related tests if it is installed. I am unfamiliar with the current scripts used to collect all the test cases, so not sure how feasible that would be.

kisvegabor:
Doh, it needs to be [installed manually](https://docs.lvgl.io/master/libs/freetype.html). Anyway, I hope it still works.

Could you update the CI in `master` to install FreeType? If so I added a basic FreeType test. After that we could rebase this PR on `master`.

src/libs/freetype/lv_freetype.h
FASTSHIFT: done

FASTSHIFT:
> Doh, it needs to be [installed manually](https://docs.lvgl.io/master/libs/freetype.html). Anyway, I hope it still works.
> 
> Could you update the CI in `master` to install FreeType? If so I added a basic FreeType test. After that we could rebase this PR on `master`.

We can directly compile the freetype source code with lvgl, which is the simplest and has no platform restrictions.
Just add in `CMakeLists.txt`:
```cmake
add_subdirectory(${PROJECT_SOURCE_DIR}/freetype2 ${PROJECT_BINARY_DIR}/build_freetype2)

target_link_libraries(
        Simulator PRIVATE
        lvgl
        lvgl::examples
        lvgl::demos
        lvgl::drivers
        freetype
)
```

kisvegabor:
Hmmm, is's great! @embeddedt what do you think?

embeddedt:
It sounds good to me as well. We just need to add some logic to fetch and extract a FreeType source tarball as otherwise we'd need to add it as a submodule, if I understand correctly.

FASTSHIFT:
I tried to tailor the minimum configuration of freetype required by a lvgl, and adapted the file access and memory allocation of freetype to lvgl. 
Put the compressed file:
[freetype_port.zip](https://github.com/lvgl/lvgl/files/9951997/freetype_port.zip)
 in this directory:
image
and add it to `CMakeLists.txt`:
```cmake
add_definitions(
        -DFT2_BUILD_LIBRARY
        -DFT_CONFIG_MODULES_H=<freetype_port/ftmodule.h>
        -DFT_CONFIG_OPTIONS_H=<freetype_port/ftoption.h>
)
```

It takes up about 100KB of FLASH space on the MCU platform.

kisvegabor:
Wao, I didn't know that FT can be configured to replace all memory and file system functions. What a great lib!

I wonder if we should promote this option in the docs too and add the content of `freetype_port.zip` into our `freetype` folder guarded with `LV_FREETYPE_USE_LVGL_PORT` or so.

What do you think?

FASTSHIFT:
> Wao, I didn't know that FT can be configured to replace all memory and file system functions. What a great lib!
> 
> I wonder if we should promote this option in the docs too and add the content of `freetype_port.zip` into our `freetype` folder guarded with `LV_FREETYPE_USE_LVGL_PORT` or so.
> 
> What do you think?

This is a very good choice, I uploaded `lv_ftsystem.c`, but freetype does not call the functions in `lv_ftsystem.c` and uses its own default implementation. Can you help me check it?

kisvegabor:
In the [CMakeList](https://github.com/freetype/freetype/blob/master/CMakeLists.txt) of FreeType I haven't found any mentions about `FT_CONFIG_OPTIONS_H` or `FT_CONFIG_OPTIONS_H` .

It seems it always wants to use its one `ftsystem.c`. See [here](https://github.com/freetype/freetype/blob/0f43a0e7ebfbda303f0b2e8b6d5db63c362233a3/CMakeLists.txt#L426-L432).

FASTSHIFT:
> In the [CMakeList](https://github.com/freetype/freetype/blob/master/CMakeLists.txt) of FreeType I haven't found any mentions about `FT_CONFIG_OPTIONS_H` or `FT_CONFIG_OPTIONS_H` .

freetype will check the custom configuration file in [ftheader.h](https://github.com/freetype/freetype/blob/master/include/freetype/config/ftheader.h#L146), which is also mentioned in the [document](https://github.com/freetype/freetype/blob/master/docs/CUSTOMIZE#L114).

FASTSHIFT:
> > In the [CMakeList](https://github.com/freetype/freetype/blob/master/CMakeLists.txt) of FreeType I haven't found any mentions about `FT_CONFIG_OPTIONS_H` or `FT_CONFIG_OPTIONS_H` .
> 
> freetype will check the custom configuration file in [ftheader.h](https://github.com/freetype/freetype/blob/master/include/freetype/config/ftheader.h#L146), which is also mentioned in the [document](https://github.com/freetype/freetype/blob/master/docs/CUSTOMIZE#L114).

We can recommend users to point the path to the lv_freetype directory:
```cmake
add_definitions(
        -DFT2_BUILD_LIBRARY
        -DFT_CONFIG_MODULES_H=<lvgl/src/libs/freetype/ftmodule.h>
        -DFT_CONFIG_OPTIONS_H=<lvgl/src/libs/freetype/ftoption.h>
)
```

kisvegabor:
Great, thank you! So let's step one square back. What would be the bast way of build the freetype port in CI?

Probably `wget` freetype next to the lvgl folder in CI.  On desktop we can say freetype needs to be installed to run the tests.

FASTSHIFT:
> Great, thank you! So let's step one square back. What would be the bast way of build the freetype port in CI?
> 
> Probably `wget` freetype next to the lvgl folder in CI. On desktop we can say freetype needs to be installed to run the tests.

I think you should try to use the latest version of the test, the installed version of freetype may be older versions not necessarily uniform.

kisvegabor:
In this case we can have a `test_setup_enc.py` or script to install/download all we need. What do you think?

FASTSHIFT:
> In this case we can have a `test_setup_enc.py` or script to install/download all we need. What do you think?

I agree with this.

kisvegabor:
Okay, great. I'd like to merge it and I think we should open a new PR for updating the tests.

Just one thing: I'm not sure if it's really a good idea to use LVGL's FS function with FT. This way the example won't work if someone normally installed  FreeType. So I suggest making the use of LVGL FS optional with a defiine like `LV_FREETYPE_USE_LVGL_FS` and set it to `0` by default. What do you think?

FASTSHIFT:
> Okay, great. I'd like to merge it and I think we should open a new PR for updating the tests.
> 
> Just one thing: I'm not sure if it's really a good idea to use LVGL's FS function with FT. This way the example won't work if someone normally installed FreeType. So I suggest making the use of LVGL FS optional with a defiine like `LV_FREETYPE_USE_LVGL_FS` and set it to `0` by default.

`LV_FREETYPE_USE_LVGL_PORT` is off by default, I can add notes in the document.

kisvegabor:
Ah, I see now. Thank you.  and it's already documented [here](https://github.com/lvgl/lvgl/pull/3782/files#diff-89a31af9fc7085f81fc10827c928ca98e5e279b108792cabceba919b1dc6129dR60).

Ok, I merge it. Thank you @FASTSHIFT.

Are you interested in participating of updating the tests too?

FASTSHIFT:
> Ah, I see now. Thank you. and it's already documented [here](https://github.com/lvgl/lvgl/pull/3782/files#diff-89a31af9fc7085f81fc10827c928ca98e5e279b108792cabceba919b1dc6129dR60).
> 
> Ok, I merge it. Thank you @FASTSHIFT.
> 
> Are you interested in participating of updating the tests too?

ok if I have free time I will do it ;)

kisvegabor:
Great, thank you in advance!

PR #3796, opened 2022-11-10
fix(sdl): crash on lv_draw_sdl

gorogoro2:
lv_draw_sdl_init_ctx will crash if it does not have the user_data

Also; the driver_hor_res and ver_res did not seem quite write (line: 126)


kisvegabor:
Thank you!

PR #3800, opened 2022-11-15
Unit tests for chart widget

C47D:

Unit tests for chart widget

### Checkpoints
 
### DIFF
--- a/src/widgets/chart/lv_chart.c
+++ b/src/widgets/chart/lv_chart.c
@@ -358,7 +358,12 @@ lv_chart_series_t * lv_chart_add_series(lv_obj_t * obj, lv_color_t color, lv_cha
     if(chart->type == LV_CHART_TYPE_SCATTER) {
         ser->x_points = lv_malloc(sizeof(lv_coord_t) * chart->point_cnt);
         LV_ASSERT_MALLOC(ser->x_points);
-        /* NOTE: Should we return if ser->x_points is NULL? Is this an out of memory error? */
+        if(ser->x_points == NULL) {
+            lv_free(ser->y_points);
+            _lv_ll_remove(&chart->series_ll, ser);
+            lv_free(ser);
+            return NULL;
+        }
     }
     else {
         ser->x_points = NULL;
@@ -374,8 +379,8 @@ lv_chart_series_t * lv_chart_add_series(lv_obj_t * obj, lv_color_t color, lv_cha
     ser->start_point = 0;
     ser->y_ext_buf_assigned = false;
     ser->hidden = 0;
-    ser->x_axis_sec = axis & LV_CHART_AXIS_SECONDARY_X;
-    ser->y_axis_sec = axis & LV_CHART_AXIS_SECONDARY_Y;
+    ser->x_axis_sec = axis & LV_CHART_AXIS_SECONDARY_X ? 1u : 0u;
+    ser->y_axis_sec = axis & LV_CHART_AXIS_SECONDARY_Y ? 1u : 0u;
 
     /* Set points on y axis to value LV_CHART_POINT_NONE so they do not get drawn */
     uint16_t i;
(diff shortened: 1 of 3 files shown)


C47D:
Hi @kisvegabor and @embeddedt , I've added some questions to the edited code, it would be really helpful if you can help me clarify them.

I've set the `x_points` pointer of the series on the not scatter type charts, I don't know if it was meant to be kept unassigned.

C47D:
I was thinking on the naming of the tests, let me know what do you think.

- When testing a particular interface: `test_<widget>_<interface>_<scenario>`, e.g. `test_chart_set_new_point_count_as_zero`, here we test `lv_chart_set_point_count` when setting the new count to 0.
- When testing a particular use case: `test_<widget>_<usecase>`, e.g. `test_chart_point_is_added_at_the_end_of_a_serie`, here we test a particular point is added at the end of the serie, we might need to call multiple interfaces as the use is more complex.

src/widgets/chart/lv_chart.c
```diff
@@ -344,34 +344,42 @@ lv_chart_series_t * lv_chart_add_series(lv_obj_t * obj, lv_color_t color, lv_cha
...
     }
 
+    /* Set series properties on successful allocation */
+    ser->color = color;
     ser->start_point = 0;
     ser->y_ext_buf_assigned = false;
     ser->hidden = 0;
-    ser->x_axis_sec = axis & LV_CHART_AXIS_SECONDARY_X ? 1 : 0;
-    ser->y_axis_sec = axis & LV_CHART_AXIS_SECONDARY_Y ? 1 : 0;
+    ser->x_axis_sec = axis & LV_CHART_AXIS_SECONDARY_X;
```
kisvegabor: Why is this change? We want to store 1 or 0 not the result of &-ing.

src/widgets/chart/lv_chart.c
```diff
@@ -344,34 +344,42 @@ lv_chart_series_t * lv_chart_add_series(lv_obj_t * obj, lv_color_t color, lv_cha
...
-
-    ser->color  = color;
+    /* Allocate memory for point_cnt points, handle failure below */
     ser->y_points = lv_malloc(sizeof(lv_coord_t) * chart->point_cnt);
     LV_ASSERT_MALLOC(ser->y_points);
 
     if(chart->type == LV_CHART_TYPE_SCATTER) {
         ser->x_points = lv_malloc(sizeof(lv_coord_t) * chart->point_cnt);
         LV_ASSERT_MALLOC(ser->x_points);
+        /* NOTE: Should we return if ser->x_points is NULL? Is this an out of memory error? */
```
kisvegabor: Yes,  I think so. There is no point to continue in that case. It'd be great to free the other `malloc`s too.

kisvegabor:
> I've set the x_points pointer of the series on the not scatter type charts, I don't know if it was meant to be kept unassigned.


Sure, it's better to have an initial value. 

> When testing a particular interface: test_<widget>_<interface>_<scenario>, e.g. test_chart_set_new_point_count_as_zero, here we test lv_chart_set_point_count when setting the new count to 0.

Why don't we use the exact same "interface" name?  I.e. `test_chart_set_point_count_as_zero`.

> When testing a particular use case: test_<widget>_<usecase>, e.g. test_chart_point_is_added_at_the_end_of_a_serie, here we test a particular point is added at the end of the serie, we might need to call multiple interfaces as the use is more complex.

There was an idea about using "should" too. E.g. `test_chart_next_point_should_be_added_at_the_end_of_a_serie`. I'm ok with both, but the "should" version sounds more human friendly to me.

src/widgets/chart/lv_chart.c
C47D: You're right, I got confused and thought it was && instead of just one, I'll revert it.

src/widgets/chart/lv_chart.c
C47D: Is d769df2d2 ok? Is there a way to simulate malloc failure?

kisvegabor: Yes, looks great!

Issue #3763, opened 2022-10-24
Discussion on the behavior consistency of `lv_flex` and `web flex`


W-Mai:
### Introduce the problem

Will `lv_flex` eventually be mostly compliant with `web flex` standards? 

For now, some features in `lv_flex` are still not quite consistent with `web flex`. 

The `grow` attribute in `lv_flex` simply means to carve up the weight of the remaining space. 

However, the `grow` in web still has a `limit of not less than its own size`. 

At the same time, `lv_flex` does not have the `shrink` attribute in web, but replaces part of the shrink feature in web with the function of `grow` attribute. 

Still, in `lv_flex`, the behavior of `align cross place` is inconsistent with that of web. 

So, will `lv_flex` be mostly compatible with web flex standards in the future?

### Examples and cases

N/A
### Suggested solution

Support Web Standard.

kisvegabor:
Hi,

Thanks for the feedback!

> Will lv_flex eventually be mostly compliant with web flex standards?

It's the goal but to only to reasonable extent. 


> The grow attribute in lv_flex simply means to carve up the weight of the remaining space.
However, the grow in web still has a limit of not less than its own size.
At the same time, lv_flex does not have the shrink attribute in web, but replaces part of the shrink feature in web with the function of grow attribute.

True. Actually, I'm ok with having both `shrink` and `grow`. What causes some headache for me in case of web-flex is `flex-basis`. I'd like to omit it for the sake of simplicity and use the set width/height as basis. 

> Still, in lv_flex, the behavior of align cross place is inconsistent with that of web.

What is the inconsistency here?

W-Mai:
> What is the inconsistency here?

I'm going to try to write a demo to reproduce this situation

W-Mai:
```cpp
auto scr = lv_scr_act();

auto obj0 = lv_obj_create(scr);

lv_obj_set_flex_align(obj0, LV_FLEX_ALIGN_START, LV_FLEX_ALIGN_CENTER, LV_FLEX_ALIGN_START);

lv_obj_set_width(obj0, 800);
lv_obj_set_height(obj0, 500);

for (int i = 0; i < 8; i++) {
    lv_obj_create(obj0);
}
```

When there is only one track, setting the alignment of the cross direction to center should be centered in the web.
And that's what lvgl looks like.
[...]

Issue #20, opened 2017-08-28
Allow background colour (mcolor) for labels


scopa90:
It might be good to allow the background colour of a label to be set via mcolor and inherit the parent object's mcolor if it's not supplied.

kisvegabor:
Hi,
it is a good idea, thank you! 
Do you mean if the label's style is NULL (inherited from parent) use transparent background and use _mcolor_ as background color if label's style is explicitly set?  Maybe the *empty* style attribute can disable this feature.

scopa90:
Or just keep it as transparent unless it's explcitly set?

kisvegabor:
Currently there is no tracking for a unique style attribute (like *mcolor*) whether it is set or not. Only I can see the entire style is inherited or not.

kisvegabor:
If we think it further the label can use all *backround related* style attributes like *mcolor*, *gcolor*, *bcolor*, *bopa* etc. but only if its style is set explicitly.  Else only use the inherited *font*, *ccolor*, *letter space* etc 

What do you think?

kisvegabor:
I added the update to `beta`. Now if a label's style is set by *lv_obj_set_style()* to not NULL then background will be drawn. 
To avoid drawing background set *empty=1* and *bwidth=0* properties or copy LV_STYLE_TRANSP when initializing a new label style. 

Tests are still needed to see how it effect existing objects where labels are used.

scopa90:
Just had chance to test this but only for a straight label.  Looks very good.

kisvegabor:
Thanks for testing!

Issue #19, opened 2017-08-18
New object types in next release - ideas, suggestions, reviews


kisvegabor:
In the next release (v4.3.0) I'm planning to add **new smart phone like object types**. 
I thought about the following types:
 - **lv_split** a hor. or ver. line for decoration purpose
 - **lv_valset** a label with "+" and "-" buttons
 - **lv_tabview** organise content with tabs (support swipe left/right)
 - **lv_btngrp** display more buttons to choose an option from them
 - **lv_switch** turn on/off by tap (a little slider)
 - **lv_roller** roller to select a value (like on smartphones)

If you have any new idea or suggestion don't hesitate to **write a comment**!

kisvegabor:
Tabview, roller,switch and btngrp (from lv_btnm)  is added in v5.0.0

PR #3807, opened 2022-11-17
feat(margin): add `margin` property to lvgl

W-Mai:

After a fierce battle of ideas, as a former front-end developer, I still wanted to add the `margin` attribute to `lvgl`, but of course adding the `margin` attribute has no effect on those who don't use it.

In the end I introduced the `margin` property to `lvgl`. now the following method has been added

```cpp
void lv_obj_set_style_margin_top(struct _lv_obj_t * obj, lv_coord_t value, lv_style_selector_t selector);
void lv_obj_set_style_margin_bottom(struct _lv_obj_t * obj, lv_coord_t value, lv_style_selector_t selector);
void lv_obj_set_style_margin_left(struct _lv_obj_t * obj, lv_coord_t value, lv_style_selector_t selector);
void lv_obj_set_style_margin_right(struct _lv_obj_t * obj, lv_coord_t value, lv_style_selector_t selector);
void lv_obj_set_style_margin_ver(struct _lv_obj_t * obj, lv_coord_t value, lv_style_selector_t selector);
void lv_obj_set_style_margin_hor(struct _lv_obj_t * obj, lv_coord_t value, lv_style_selector_t selector);
void lv_obj_set_style_margin_all(struct _lv_obj_t * obj, lv_coord_t value, lv_style_selector_t selector);

void lv_obj_get_style_margin_top(struct _lv_obj_t * obj, lv_coord_t value, lv_style_selector_t selector);
void lv_obj_get_style_margin_bottom(struct _lv_obj_t * obj, lv_coord_t value, lv_style_selector_t selector);
void lv_obj_get_style_margin_left(struct _lv_obj_t * obj, lv_coord_t value, lv_style_selector_t selector);
void lv_obj_get_style_margin_right(struct _lv_obj_t * obj, lv_coord_t value, lv_style_selector_t selector);
```

And the `lv_obj_align` and `lv_obj_align_to` methods have been enhanced to support the `margin` property.
Similarly, `lv_obj_scroll` has been adjusted to increase their support for `margin`.
Corresponding changes have also been made to the `LV_SIZE_CONTENT` behavior of the component.

Glad if you can accept the margin attribute back into lvgl.

Here is a screenshot of the demo:
without margin:
image
with margin:
image

I tested several demos of `lvgl` and found no impact on the existing behavior.

### Checkpoints
 

W-Mai:
https://github.com/lvgl/lvgl/issues/3763

W-Mai:
If it's acceptable, I'll introduce margin into lv_flex next.

kisvegabor:
I was thinking a lot of whether adding margin or not.  I've also collected some references (other libs, articles, etc) where there is no margin but in the end only one thing matters: do we need it to build UIs? 

So I'm ok with adding it if you can show a few **real-life** UI designs (i.e. not just boxes as demonstration)  where margin is really useful. My experience is that (while creating embedded UI all day) flex and grid (even in their current form) are enough to manage layouts.

W-Mai:
It's hard for me to find a situation that meets your requirements, because the vast majority of websites use margin.
You can see the existence of `margin` on any website.

So I wrote a simple demo to demonstrate the advantages of `margin` in some cases.
The animation is as follows.

https://user-images.githubusercontent.com/24841247/202697380-5a060159-1257-4450-9ed4-85dd4f8e552d.mov

If you use `gap` to implement it, you can't control each item finely and individually, and if you use `transition`, it won't affect the layout of other item.

And offset is not a `style` property, it cannot be used in flex layouts.
At the same time, this example can also easily modify the distribution of item.

The code is also very simple, using `attribute animation` can be achieved.

```cpp
static lv_style_transition_dsc_t dsc;
static const lv_style_prop_t props[] = {
    LV_STYLE_MARGIN_TOP,
    LV_STYLE_MARGIN_BOTTOM,
    LV_STYLE_MARGIN_LEFT,
    LV_STYLE_MARGIN_RIGHT,
    LV_STYLE_HEIGHT,
    LV_STYLE_PROP_INV,
};
static lv_style_t style;
lv_style_init(&style);

lv_style_transition_dsc_init(&dsc, props, lv_anim_path_ease_in_out, 50, 0, nullptr);
lv_style_set_transition(&style, &dsc);
lv_style_set_margin_top(&style, 10);
lv_style_set_margin_left(&style, 10);
lv_style_set_margin_right(&style, 10);
lv_style_set_margin_bottom(&style, 10);
lv_style_set_height(&style, 20);

......

lv_obj_add_style(o, &style, LV_STATE_PRESSED);
```

In addition, `margin` has an advantage over `offset` because `margin` is a property that takes effect after it is **set once**.

As a front-end developer, seeing `margin` is more cordial than other attributes and can adapt to `lvgl` development more quickly.

The design drafts provided by UI designers are also based on `margin`, `padding`, `border` and so on.

To web, iOS 's `UIKit` and `SwiftUI` also have `margin` attributes, which is very easy to use.

You can say that there are other ways to replace `margin` to achieve a similar effect. What I want to say is that if you have `margin`, you don't have to **work around** and you can manipulate each obj in a more straightforward way.

Adding margin **will not** put more burden on `lvgl`, on the contrary, it is a more convenient tool for more developers.
\\(^o^)/~

kisvegabor:
> So I wrote a simple demo to demonstrate the advantages of margin in some cases.

Actually it's a good example.  However a little bit special. Without margin I would solve it by wrapping each colored boxes into a transparent box and change the size of the outer transparent box.

But there is one thing that I don't understand in this demo. At 0:04 you click the pink box it stays in place. But when you blue box that stays in place.  But the length of the track became larger, with the extra margin and when this larger width is centered all elements should have moved a little. It's like the pivot point of the centering is the click position.   What is trick, or am I missing something obvious? :slightly_smiling_face: 

> To web, iOS 's UIKit and SwiftUI also have margin attributes, which is very easy to use.

There are other frameworks where there is no margin. See for example Framer, a very new website designer tool: https://www.youtube.com/watch?v=2JAyFZOXxDw

> The design drafts provided by UI designers are also based on margin, padding, border and so on.

I have a quite negative feeling about the current state of the web development. I think it's quite messy and hacky. So I would afraid of taking it as a  good reference (rather the opposite, we should learn the mistakes of web). I consider it only as a source of ideas which should be taken by a huge grain of salt.

> Adding margin will not put more burden on lvgl, on the contrary, it is a more convenient tool for more developers.

Addig margin actually has consequences even if one doesn't want to use them:
- steeper learning curve on layouts, it can be mixed with padding, longer docs
- more style property to `get` might cause  performance drop if the layouts with a lot of widgets need to be arrange.  
- makes flex and grid algorithm more complex (more possible bugs, harder to maintain, corner cases)
- mixing padding and margin can result in hard to debug layouts 
- margin collapse?

My opinion is that `margin` is a conceptually bad idea because it acts out of the element. E.g. you create a style for a button with `margin-top:10px`. It means literally whatever is on top of the button I'd like to keep 10px distance. How do you know that it will true in general? Keep distance from whatever? Doesn't sound like a good design philosophy. Can't we use "spacers" in these rare cases instead (like Framer does)?

On the other hand padding is enclosed into the object. You know what you place inside and you can adjust padding  accordingly.

In short, I think `margin` is a bad solution to the "how to hack the spacing around a given element" problem. Grid should handle these cases very well when there is a fixed number of elements and you would like to craft a well-designed unique layout (e.g. a card with avatar + text + button). So IMO the use of grid can replace margin very well.  

Anyway I'm really open to discuss the need of margin I just had a little bit less popular opinion about it. :slightly_smiling_face:  I suggest approaching this from "what is the concrete problem we would like to solve" and find a solution to the problem (I.e. this design is hard to implement now, how to make it simpler?), instead of adding features which might be redundant or not really needed.

xiaoxiang781216:
@kisvegabor since margin is part of CSS spec, many designer and developer get used to develop UI layout with margin property. Actually, we are developing a JavaScript UI framework base on LVGL for smart watch. It's very important to make the transition as smooth as possible for the traditional web developer, so we decide to follow CSS spec as close as possible. Because LVGL layout lack the support of margin(and other) property, we have to disable all LVGL layout engine and layout(base on https://github.com/facebook/yoga) the widget out side LVGL. But this approach has some bad side effect in the production environment, so the better approach is to improve lv_flex to support the key CSS property.
Margin is one key issue which block we use lv_flex directly, so it's important to see margin included in LVGL layout system, @kisvegabor . 🙂

kisvegabor:
I understand your situation and I don't want to argue about your decision of following CSS. Probably it's a complex system and there are no simple answers. :slightly_smiling_face: 
I also don't want to repeat myself about IMO how bad CSS is. 

So let's approach it constructively! What is the problem and what how can we solve it?

**Problem**
LVGL is CSS-like so it sounds reasonable to assume there is `margin` too. It breaks portability between web-like designs and embedded. So the question is not really "how to implement UIs without margin" but "have margin because CSS has it too". 

**Options**
1. Support `margin` everywhere just like CSS. I'm definitely against it because it brings a lot of complication everywhere. E.g. a simple `lv_obj_set_pos()` shouldn't care about `margin` 
2. Support margin only in `flex` and `grid`. It makes more sense as it's a local complication.
4. Offer a straight forward `margin` replacement workaround.  Usually I wrap the widget into a transparent container and set the `padding` on the wrapper. It has the same effect as `margin`. It works with LVGL only but it messes up the hierarchy and it's hacky.

**Solution**
I could be ok with 2) if we really keep is as local as possible:
- only for flex and grid
- no margin collapse
- maybe make it optional? On flex now we need to get only `flex_grow` and margin means 4 extra property to get for each object. On layouts with 1000 items it might matter.
- shall we apply `margin` on `%` values too? E.g. with `width=100%` it might make sense to use margin.

W-Mai:
https://github.com/lvgl/lvgl/pull/3845

scripts/style_api_gen.py
```diff
@@ -99,6 +99,23 @@
  'style_type': 'num',   'var_type': 'lv_coord_t', 'default':0, 'inherited': 0, 'layout': 1, 'ext_draw': 0,
  'dsc': "Sets the padding between the columns. Used by the layouts."},
 
+{'section': 'Margin', 'dsc' : "Properties to describe spacing around an object. Very similar to the margin properties in HTML."},
+{'name': 'MARGIN_TOP',
+ 'style_type': 'num',   'var_type': 'lv_coord_t',  'default':0, 'inherited': 0, 'layout': 1, 'ext_draw': 0,
+ 'dsc': "Sets the margin on the top. It makes the object area bigger in this direction."},
```
kisvegabor:
```suggestion
 'dsc': "Sets the margin on the top. The object will keep this space from its siblings in layouts. "},
```

kisvegabor:
I've checked the changes and they look good in general. Thank you! 

A few things:
- With percentage size margin is still no considered. See [here](https://github.com/lvgl/lvgl/blob/master/src/core/lv_obj_pos.c#L99)
- Do you plan to add margin to grid too?
- It'd be really nice to have some unit tests for simple sizing/positioning, flex and grid. Are you interested in adding them? I think it'd be helpful during the development too and I assume you already have some test codes.

W-Mai:
> * With percentage size margin is still no considered. See [here](https://github.com/lvgl/lvgl/blob/master/src/core/lv_obj_pos.c#L99)

Aiming at the problem of `percentage` size, since the current padding of lvgl does not support percentage, I want to open a separate pr to support both `padding` and `margin`, but percentage may need some processing when it comes to problems such as `LV_SIZE_CONT` that need to calculate the size of the parent component.

> * Do you plan to add margin to grid too?

I have this plan, but I haven't looked at the code for the grid layout yet. I'll add margin to the grid layout later.

> * It'd be really nice to have some unit tests for simple sizing/positioning, flex and grid. Are you interested in adding them? I think it'd be helpful during the development too and I assume you already have some test codes.

I'm interested in it. I can try it.

kisvegabor:
> Aiming at the problem of percentage size, since the current padding of lvgl does not support percentage, I want to open a separate pr to support both padding and margin, but percentage may need some processing when it comes to problems such as LV_SIZE_CONT that need to calculate the size of the parent component.

I meant`margin=20px` on a child with `width = 100%` should make the child smaller. (Not `margin = 10%`, but it would be great too.)

kisvegabor:
I think only a few unit tests needs to be added and we are ready to go.

W-Mai:
I wrote a simple testcase and wondered if it would work

kisvegabor:
I don't know why but it seems the the new test file is not picked by CI. See https://github.com/lvgl/lvgl/actions/runs/3891743797/jobs/6642347842

Maybe `test_grid()`, `test_margin()` and `test_align()` should be in there own file.

W-Mai:
> I don't know why but it seems the the new test file is not picked by CI. See https://github.com/lvgl/lvgl/actions/runs/3891743797/jobs/6642347842
> 
> Maybe `test_grid()`, `test_margin()` and `test_align()` should be in there own file.

Does this result mean being picked by ci?
[398](https://github.com/lvgl/lvgl/actions/runs/3891743797/jobs/6642347842#step:5:399)
> [ 82%] Generating test_margin_Runner.c
[ 82%] Building C object CMakeFiles/test_line.dir/src/test_cases/test_line.c.o
[ 82%] Building C object CMakeFiles/test_margin.dir/src/test_cases/test_margin.c.o
[ 82%] Building C object CMakeFiles/test_line.dir/test_line_Runner.c.o
[ 82%] Building C object CMakeFiles/test_margin.dir/test_margin_Runner.c.o
[ 83%] Linking C executable test_line
[ 84%] Linking C executable test_margin
[ 84%] Built target test_line
[ 84%] Generating test_mem_Runner.c
[ 84%] Built target test_margin
[ 85%] Generating test_msg_Runner.c

> 18/37 Test #19: test_line ........................   Passed    0.07 sec
19/37 Test #20: test_margin ......................   Passed    0.40 sec

kisvegabor:
Ah, I see 2 problem:
1. Seemingly I'm blind
2. Searching with Ctrl+F in the CI log is not working well as only parts of the log is loaded. 

:smile: 

Regardless to the CI picked it up correctly, just for the sake of cleaner code, could you still separate the test into 3 files?

W-Mai:
> Ah, I see 2 problem:
> Seemingly I'm blind
> Searching with Ctrl+F in the CI log is not working well as only parts of the log is loaded.

hah, I had the same problem before😁

kisvegabor:
Just one last suggestion about the docs.  If you agree with it please update the docs and after I can merge this PR.

scripts/style_api_gen.py
W-Mai: Haha, I'm a Chinese with poll English.😁
That's great.

I should update all 4 description, right?

W-Mai:
> Just one last suggestion about the docs. If you agree with it please update the docs and after I can merge this PR.

Well done.

kisvegabor:
Great, thank you much! :tada: :tada: :tada: 
I was really against margin but if use it wisely it can really help :relaxed:

PR #3819, opened 2022-11-21
modify lv_obj_transform_point

PeterDengRX:

Try to modify the problem that the coordinates are calculated incorrectly when the object is zoomed out and dragged

### Checkpoints
 

src/core/lv_indev.c
```diff
@@ -816,6 +818,8 @@ static void indev_button_proc(lv_indev_t * i, lv_indev_data_t * data)
  */
 static void indev_proc_press(_lv_indev_proc_t * proc)
 {
+    uint8_t indev_flag = false;
```
kisvegabor: What about a more descriptive name?
```suggestion
    bool no_move = false;
```

src/core/lv_obj_pos.c
```diff
@@ -795,18 +795,18 @@ void lv_obj_move_children_by(lv_obj_t * obj, lv_coord_t x_diff, lv_coord_t y_dif
     }
 }
 
-void lv_obj_transform_point(const lv_obj_t * obj, lv_point_t * p, bool recursive, bool inv)
+void lv_obj_transform_point(const lv_obj_t * obj, lv_point_t * p, lv_point_t * p1, bool recursive, bool inv)
```
kisvegabor:
```suggestion
void lv_obj_transform_point(const lv_obj_t * obj, lv_point_t * p, lv_point_t * pref, bool recursive, bool inv)
```

src/core/lv_obj_pos.h
```diff
@@ -355,7 +355,7 @@ void lv_obj_move_children_by(struct _lv_obj_t * obj, lv_coord_t x_diff, lv_coord
  * @param recursive     consider the transformation properties of the parents too
  * @param inv           do the inverse of the transformation (-angle and 1/zoom)
  */
-void lv_obj_transform_point(const struct _lv_obj_t * obj, lv_point_t * p, bool recursive, bool inv);
+void lv_obj_transform_point(const struct _lv_obj_t * obj, lv_point_t * p, lv_point_t * p1, bool recursive, bool inv);
```
kisvegabor: Please add a detailed description about the new parameter in the comments above.

kisvegabor (commented):
Wow, you are genius :slightly_smiling_face: I've tested it and it works very well.
I've tested it like this
tr


In the end we should merge it to `master` but I'll take of rebasing.

src/core/lv_indev.c
PeterDengRX: I can’t think of a better name, so let's use this.

src/core/lv_obj_pos.c
PeterDengRX: I can’t think of a better name, so let's use this.

src/core/lv_obj_pos.h
PeterDengRX: @param p1             a point to the last untransformed coordinate
Is It okay to describe like this, I am not good at english, please point out any grammatical errors.

kisvegabor: I'm also not a native speaker but let's try this:
```
p1 a previous untransformed point.  If the object was moving and it was transformed the a previous point helps to fix incorrect calculations due to the transformed coordinate system. 
```

Is it technically correct?

src/core/lv_obj_pos.c
kisvegabor: "this" means "pref" or "p1"? :slightly_smiling_face:

src/core/lv_obj_pos.h
PeterDengRX: yes ,it is correct. p a current untransformed point, so (p - p1) can calculate the current moving distance, and pivot +  (p - p1) ensures taht the distance between the current pressing point and the starting point of the object is consistent with the previous one.

src/core/lv_obj_pos.c
PeterDengRX: "pref" means "preface". how about using "prev_p"? it means the previous point

kisvegabor: I mean "reference point" (should have been "p_ref") but "prev_p" is fine too.

src/core/lv_obj_pos.h
kisvegabor: >  p a current untransformed point, so (p - p1) can calculate the current moving distance, and pivot + (p - p1) ensures taht the distance between the current pressing point and the starting point of the object is consistent with the previous one.

Please add it as a comment in the function :slightly_smiling_face:

kisvegabor: I wonder how it works with nested transformation. E.g. if there is a 45 deg rotated rectangle in which there is a -70 deg rotated other rectangle. As transformation is recursive `p` will be transformed during the process.

PeterDengRX: Only focus on the most foreground object when pressing rhe screen. As long as the value of `proc->types.pointer.vect` is correct, no matter how nested transformation, it will work normally when moving the object. What do you think?

But I wonder if `lv_obj_transform_point` is not called in function `indev_proc_press`, will it cause any problems? Or what are your considerations for adding `lv_obj_transform_point` to function `indev_proc_press` at taht time? 

It seems that the value of `proc->types.pointer.vect` can be calculated correctly without calling `lv_obj_transform_point`.

kisvegabor:
I made some more tests with this code:
```c
lv_obj_t * panel = lv_obj_create(lv_scr_act());
lv_obj_set_size(panel, 280, 150);
lv_obj_set_flex_flow(panel, LV_FLEX_FLOW_ROW);
lv_obj_align(panel, LV_ALIGN_CENTER, 0, 20);

uint32_t i;
for(i = 0; i < 10; i++) {
    lv_obj_t * btn = lv_btn_create(panel);
    lv_obj_set_size(btn, 150, 100);
    lv_obj_set_style_transform_pivot_x(btn, 75, 0);
    lv_obj_set_style_transform_pivot_y(btn, 50, 0);
    lv_obj_set_style_transform_angle(btn, 1800, 0);

    lv_label_set_text_fmt(label, "Panel %"LV_PRIu32"\nno snap", i);
    lv_obj_clear_flag(btn, LV_OBJ_FLAG_SNAPPABLE | LV_OBJ_FLAG_SCROLL_ON_FOCUS);
    lv_obj_center(label);
}

lv_obj_update_snap(panel, LV_ANIM_ON);
```

And seems the algorithm gets confused sometimes:
Peek 2022-11-25 11-23

Do you have any ideas?

PeterDengRX:
> I made some more tests with this code:
> 
> ```c
> lv_obj_t * panel = lv_obj_create(lv_scr_act());
> lv_obj_set_size(panel, 280, 150);
> lv_obj_set_flex_flow(panel, LV_FLEX_FLOW_ROW);
> lv_obj_align(panel, LV_ALIGN_CENTER, 0, 20);
> 
> uint32_t i;
> for(i = 0; i < 10; i++) {
>     lv_obj_t * btn = lv_btn_create(panel);
>     lv_obj_set_size(btn, 150, 100);
>     lv_obj_set_style_transform_pivot_x(btn, 75, 0);
>     lv_obj_set_style_transform_pivot_y(btn, 50, 0);
>     lv_obj_set_style_transform_angle(btn, 1800, 0);
> 
>     lv_obj_t * label = lv_label_create(btn);
>     lv_label_set_text_fmt(label, "Panel %"LV_PRIu32"\nno snap", i);
>     lv_obj_clear_flag(btn, LV_OBJ_FLAG_SNAPPABLE | LV_OBJ_FLAG_SCROLL_ON_FOCUS);
>     lv_obj_center(label);
> }
> 
> lv_obj_update_snap(panel, LV_ANIM_ON);
> ```
> 
> And seems the algorithm gets confused sometimes: Peek 2022-11-25 11-23 [ Peek 2022-11-25 11-23 ](https://user-images.githubusercontent.com/7599318/203962535-7119fca7-f6cd-435a-97ed-565f13d810f5.gif) [ ](https://user-images.githubusercontent.com/7599318/203962535-7119fca7-f6cd-435a-97ed-565f13d810f5.gif)
> 
> Do you have any ideas?

The test here shows that the value of `obj->coords` remains unchanged after the pressing point is moved. But I don't know what causes it.

PeterDengRX:
The test here shows that the value of obj->coords remains unchanged after the pressing point is moved. But I don't know what causes it.

PeterDengRX:
Hi, kisvegabor, I tried to fix the problem above, it semms to work fine, but the algorithm still gets confused if the object has the `LV_OBJ_FLAG_SNAPPABLE` flag added. 
indev_pointer_proc:
图片

_lv_indev_scroll_handler:
图片

I would still recommend removing `lv_obj_transform_point(indev_obj_act, &proc->types.pointer.act_point, &proc->types.pointer.indev_point, true, true);` in `indev_proc_press`, it semms like an sasier solution.

kisvegabor:
Let's reiterate:  The problem is that, if an object is transformed, moved, and pressed the last point seems different as the points are depending on the widget's position.  But we need to transform the points as the vector needs to interpreted in the widget's coordinate system.

So the goal is to transform  the vector.  E.g. if the widget is
- rotated by 45 deg: we need to rotate the vector back by -45 deg to interpret the vector on the untransformed widget's coordinate system 
- zoomed to 200%: we need to assume 1/2 smaller vector (inverted transformation)

So the pivot doesn't matter, only the length and direction of the vector needs to be changed. Hence we can write a custom transform function which always transforms the vector around the (0;0) pivot.

I've tried it out and it worked well for me:
```diff
diff --git a/src/core/lv_indev.c b/src/core/lv_indev.c
index 064464256..9a1b5e4c3 100644
--- a/src/core/lv_indev.c
+++ b/src/core/lv_indev.c
@@ -340,9 +340,6 @@ lv_obj_t * lv_indev_search_obj(lv_obj_t * obj, lv_point_t * point)
 static void indev_pointer_proc(lv_indev_t * i, lv_indev_data_t * data)
 {
     lv_disp_t * disp = i->driver->disp;
-    /*Save the raw points so they can be used again in _lv_indev_read*/
-    i->proc.types.pointer.last_raw_point.x = data->point.x;
-    i->proc.types.pointer.last_raw_point.y = data->point.y;
 
     if(disp->driver->rotated == LV_DISP_ROT_180 || disp->driver->rotated == LV_DISP_ROT_270) {
         data->point.x = disp->driver->hor_res - data->point.x - 1;
@@ -384,6 +381,10 @@ static void indev_pointer_proc(lv_indev_t * i, lv_indev_data_t * data)
         indev_proc_release(&i->proc);
     }
 
+    /*Save the raw points so they can be used again in _lv_indev_read*/
+    i->proc.types.pointer.last_raw_point.x = data->point.x;
+    i->proc.types.pointer.last_raw_point.y = data->point.y;
+
     i->proc.types.pointer.last_point.x = i->proc.types.pointer.act_point.x;
     i->proc.types.pointer.last_point.y = i->proc.types.pointer.act_point.y;
 }
@@ -853,6 +854,7 @@ static void indev_proc_press(_lv_indev_proc_t * proc)
         if(indev_reset_check(proc)) return;
     }
 
+    lv_point_t act_raw_point = proc->types.pointer.act_point;
     lv_obj_transform_point(indev_obj_act, &proc->types.pointer.act_point, true, true);
 
     /*If a new object was found reset some variables and send a pressed event handler*/
@@ -900,8 +902,23 @@ static void indev_proc_press(_lv_indev_proc_t * proc)
     }
 
     /*Calculate the vector and apply a low pass filter: new value = 0.5 * old_value + 0.5 * new_value*/
-    proc->types.pointer.vect.x = proc->types.pointer.act_point.x - proc->types.pointer.last_point.x;
-    proc->types.pointer.vect.y = proc->types.pointer.act_point.y - proc->types.pointer.last_point.y;
+    proc->types.pointer.vect.x = act_raw_point.x - proc->types.pointer.last_raw_point.x;
+    proc->types.pointer.vect.y = act_raw_point.y - proc->types.pointer.last_raw_point.y;
+
+    /*Transform the vector*/
+    lv_obj_t * parent = indev_obj_act;
+    lv_point_t pivot = {0, 0};
+    while(parent) {
+        int16_t angle = lv_obj_get_style_transform_angle(parent, 0);
+        int16_t zoom = lv_obj_get_style_transform_zoom(parent, 0);
+
+        if(angle != 0 || zoom != LV_IMG_ZOOM_NONE) {
+            angle = -angle;
+            zoom = (256 * 256) / zoom;
+            lv_point_transform(&proc->types.pointer.vect, angle, zoom, &pivot);
+        }
+        parent = lv_obj_get_parent(parent);
+    }
 
     proc->types.pointer.scroll_throw_vect.x = (proc->types.pointer.scroll_throw_vect.x + proc->types.pointer.vect.x) / 2;
     proc->types.pointer.scroll_throw_vect.y = (proc->types.pointer.scroll_throw_vect.y + proc->types.pointer.vect.y) / 2;
```

The only issue is that if you horizontally scroll a button with 180 degree rotation the `vect.x` becomes `-vect_x` and if the scroll is propagated to the parent, the parent will be scrolled to the opposite direction. 

Anyway, the above fix should solve the major issue (shaking when pressing) and the scroll chaining problem can be solved locally during scroll chaining by transforming back the vector as we step back one parent.

PeterDengRX:
I try to remove this code and  it work prefectly fine.
`+    /*Transform the vector*/
+    lv_obj_t * parent = indev_obj_act;
+    lv_point_t pivot = {0, 0};
+    while(parent) {
+        int16_t angle = lv_obj_get_style_transform_angle(parent, 0);
+        int16_t zoom = lv_obj_get_style_transform_zoom(parent, 0);
+
+        if(angle != 0 || zoom != LV_IMG_ZOOM_NONE) {
+            angle = -angle;
+            zoom = (256 * 256) / zoom;
+            lv_point_transform(&proc->types.pointer.vect, angle, zoom, &pivot);
+        }
+        parent = lv_obj_get_parent(parent);
+    }`

aaa

PeterDengRX:
I try to remove this code and  it work prefectly fine.
图片

aaa

kisvegabor:
try with this example code:
```c
void lv_example_scroll_22(void)
{
    lv_obj_t * panel = lv_obj_create(lv_scr_act());
    lv_obj_set_size(panel, 300, 200);
    lv_obj_set_scroll_snap_x(panel, LV_SCROLL_SNAP_CENTER);
    lv_obj_set_flex_flow(panel, LV_FLEX_FLOW_ROW);
    lv_obj_set_style_flex_track_place(panel, LV_FLEX_ALIGN_CENTER, 0);
    lv_obj_align(panel, LV_ALIGN_CENTER, 0, 20);
    lv_obj_center(panel);

    uint32_t i;
    for(i = 0; i < 10; i++) {
        lv_obj_t * btn = lv_btn_create(panel);
        lv_obj_set_size(btn, 150, 100);
        lv_obj_add_flag(btn, LV_OBJ_FLAG_SCROLLABLE);
        lv_obj_set_style_transform_pivot_x(btn, 75, 0);
        lv_obj_set_style_transform_pivot_y(btn, 50, 0);
        lv_obj_set_style_transform_angle(btn, 900, 0);

        lv_label_set_text_fmt(label, "Panel %d\n\n\nAnd some very long text to be\nsure\nthe button\nis\nscrollable.", i);
    }

    lv_obj_update_snap(panel, LV_ANIM_ON);
}
```

With the transformation horizontal scroll "gesture" scrolls vertically.

a

PeterDengRX:
I try to change src/core/lv_indev_scroll.c

图片

and remove `lv_obj_transform_point`

    if(new_obj_searched && proc->types.pointer.last_obj) {
        proc->types.pointer.scroll_throw_vect.x = 0;
        proc->types.pointer.scroll_throw_vect.y = 0;
        _lv_indev_scroll_throw_handler(proc);
        if(indev_reset_check(proc)) return;
    }


    //lv_obj_transform_point(indev_obj_act, &proc->types.pointer.act_point, true, true);

    /*If a new object was found reset some variables and send a pressed event handler*/
    if(indev_obj_act != proc->types.pointer.act_obj) {
        proc->types.pointer.last_point.x = proc->types.pointer.act_point.x;
        proc->types.pointer.last_point.y = proc->types.pointer.act_point.y;

bbb

kisvegabor:
Sorry, I closed the PR by accident while messed up something with git...

I've added a few minor fixes and refactoring and merged it in  https://github.com/lvgl/lvgl/commit/84cf05d8b23b31e000db757a278545e58fcbcbe8

Please confirm if it's working well on your end too.

PR #3821, opened 2022-11-21
test(cmake): exclude test targets from build only configurations

some00:
Related to #2337

Possible solution for a build error. Suggested in: https://github.com/lvgl/lvgl/issues/2337#issuecomment-1322689179

### Checkpoints
 

some00:
I personally don't recommend this solution. It doesn't scale well, but the source of the problem I think is the usage of globbing for gathering source files. [CMake documentation](https://cmake.org/cmake/help/latest/command/file.html#filesystem) has a note about it.
In short, an already configured build won't detect a newly added file, which can confuse newcomers. A reconfiguration has to be triggered by calling cmake/ccmake or deleting the existing build folders.

I would recommend creating a function out of the current loop and listing the source files. This can be mentioned in the documentation as part of the process to create a new test case.
For the current problem this would mean a simple if statement around the call with `test_meter.c` argument.
The downside is possible conflicts.
@kisvegabor What do you think? Should I change to loop to a function? The `ifdef` in test source file(s) solution would scale better and won't conflict either.

kisvegabor:
Thank you for looking into this issue.

I think all we need in the end is having different source files for build tests (e.g. `OPTIONS_FULL_32BIT`) and different source files for unit tests (e.g. `OPTIONS_TEST_SYSHEAP`). 

- For build tests we need `lvgl/src` + `lvgl/demos` + `lvgl/examples`
- For unit tests we need `lvgl/src` + `lvgl/demos` + `lvgl/tests`

Do you think it's doable?

some00:
I'll update the PR.

some00:
I took the chance to try and reorganize the few hundred lines of compilation options. Hope you won't mind, but again happy to revert that part.

Unfortunately, test cases depend on the examples. One instance is in `test_animimg.c`. Global variable `animimg001` is used, which is defined in `examples/assets/animimg001.c`.
So test cases still depend on `lvgl_examples` target, but that shouldn't be an issue for `test_meter.c`.

kisvegabor:
> Unfortunately, test cases depend on the examples. One instance is in test_animimg.c. Global variable animimg001 is used, which is defined in examples/assets/animimg001.c.

It's not intentional. We have [`tests/src/test_assets`](https://github.com/lvgl/lvgl/tree/master/tests/src/test_assets) and `examples/assets/animimg*.c` should be copied there.

kisvegabor:
Sorry for the trouble, but due to a different issue I moved the `-DLV_...` configs from CMake to header files. See [here](https://github.com/lvgl/lvgl/commit/077cd2e15d103bc71cae8472cf8af9d7dbd1ae87). Could you mind rebasing on top of these changes?

some00:
Is there something I should do about the CI errors? They seem unrelated to my changes.
`Error response from daemon: login attempt to https://ghcr.io/v2/ failed with status: 503 Service Unavailable`

kisvegabor:
It happens time to time. I've restarted the tests.

tests/CMakeLists.txt
```diff
@@ -169,7 +181,7 @@ foreach( test_case_fname ${TEST_CASE_FILES} )
         ${test_case_fname}
         ${test_runner_fname}
     )
-    target_link_libraries(${test_name} test_common lvgl_examples lvgl_demos lvgl png m ${TEST_LIBS})
+    target_link_libraries(${test_name} test_common lvgl_demos lvgl png m ${TEST_LIBS})
```
kisvegabor, line 184: Why was the examples removed from here?

some00: Based on your [comment](https://github.com/lvgl/lvgl/pull/3821#issuecomment-1323422245) the unit tests shouldn't build and link to `lvgl_examples`. I must have misunderstood that.

kisvegabor: Ah, it's only for the unit test. Got it. Thank you

kisvegabor:
It looks good from my side. Can I merge it?

some00:
Yes, I'm finished.

kisvegabor:
Thank you!

PR #3833, opened 2022-11-25
Clear the warning when TRANSITION_TIME = 0.

liuyanhb:
The authors ignore that trans_props is defined but not used when TRANSITION_TIME = 0.


kisvegabor:
Thank you!

PR #3845, opened 2022-11-30
feat(lv_flex): added support for `lv_flex`

W-Mai:

> **Options**
> 
> 1. Support `margin` everywhere just like CSS. I'm definitely against it because it brings a lot of complication everywhere. E.g. a simple `lv_obj_set_pos()` shouldn't care about `margin`
> 2. Support margin only in `flex` and `grid`. It makes more sense as it's a local complication.
> 3. Offer a straight forward `margin` replacement workaround.  Usually I wrap the widget into a transparent container and set the `padding` on the wrapper. It has the same effect as `margin`. It works with LVGL only but it messes up the hierarchy and it's hacky.
> 
> **Solution** I could be ok with 2) if we really keep is as local as possible:

https://github.com/lvgl/lvgl/pull/3807#issuecomment-1329602031

I only added margin to the lv_flex according to option 2, and it seemed to work fine. 

But there are some problems with it: 

Under `LTR`, `margin bottom` and `margin right` cannot participate in the calculation of scrolling and content width and height. At the same time, the problems also exist in `margin bottom` and `margin left`. 

As shown in the figure:
image
It realized with `lv_scroll` and `lv_pos` modified 👆
image
It realized without `lv_scroll` and `lv_pos` modified 👆

Notice the last pink square. In the first picture, there is a certain amount of white space on the right, which is generated by margin-right. 
In the second picture, there is no space on the right, which is due to the fact that margin is not involved in the calculation.

### Discussion 

Whether we should add an interface, after the layout calculation is completed, let the layout affect the width and height of the parent component and affect the calculation of scroll.
Or do you have other methods to solve the above problems?

### Checkpoints
 
### DIFF
--- a/src/layouts/flex/lv_flex.c
+++ b/src/layouts/flex/lv_flex.c
@@ -61,6 +61,8 @@ static void children_repos(lv_obj_t * cont, flex_t * f, int32_t item_first_id, i
 static void place_content(lv_flex_align_t place, lv_coord_t max_size, lv_coord_t content_size, lv_coord_t item_cnt,
                           lv_coord_t * start_pos, lv_coord_t * gap);
 static lv_obj_t * get_next_item(lv_obj_t * cont, bool rev, int32_t * item_id);
+static lv_coord_t lv_obj_get_width_with_margin(const lv_obj_t * obj);
+static lv_coord_t lv_obj_get_height_with_margin(const lv_obj_t * obj);
 
 /**********************
  *  GLOBAL VARIABLES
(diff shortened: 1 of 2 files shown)


kisvegabor:
I meant adding "global" margin properties (e.g. `lv_style_set_margin_top`) just not using it during position simple position calculations. 

> Under LTR, margin bottom and margin right cannot participate in the calculation of scrolling and content width and height. At the same time, the problems also exist in margin bottom and margin left.


Good question. Let's collect what are the places and cases where margin should/could be considered besides flex and grid:
- scroll (as you just mentioned)
- parent's `LV_SIZE_CONTENT` size (the parent should be large enough to include the children with their margin)?
- margin collapse?
- % sizes (E.g. with 100% width and some margin the object's width  shouldn't be  `parent width - object margin`)?
- anything else?

W-Mai:
There should be no other cases to consider

### margin collapse

I think it's okay not to think about it for a while. This is a very troublesome and confusing feature.

So I thought it would be more intuitive for `margin` to leave it as adding a lasting "`offset`" to the element, it is more intuitive and shows the effect of what kind of code the programmer writes.

Before we used Facebook's yogalayout as a `substitute` for lv_flex, there is no margin collapse problem in it, and it is very intuitive to use it.

> * % sizes (E.g. with 100% width and some margin the object's width shouldn't be `parent width - object margin`)?

Yes, from a simple point of view, `parent width - object margin` is the most straightforward way. That's what `yogalayout` does.

### How to solve?
So are we going to add these features directly to `lv_obj_scroll` and relative files or add new interfaces?

kisvegabor:
> I think it's okay not to think about it for a while. This is a very troublesome and confusing feature.

I agree.

> Yes, from a simple point of view, parent width - object margin is the most straightforward way. That's what yogalayout does.

To be more precise: `parent_width - parent_padding - object margin`


----

So in summary:
- scroll (as you just mentioned):  Yes
- parent's LV_SIZE_CONTENT size (the parent should be large enough to include the children with their margin): Yes
- margin collapse: No
- % sizes (E.g. with 100% width and some margin the object's width shouldn't be parent width - object margin)? Yes

---

> So are we going to add these features directly to lv_obj_scroll and relative files or add new interfaces?

Please handle it similarly to padding. So with new global style properties.

kisvegabor:
Continue in #3807

Issue #3780, opened 2022-11-02
Improve scroll direction prediction


FASTSHIFT:
### Perform all steps below and tick them with [x]
- [x] Read the [FAQ](https://docs.lvgl.io/master/intro/index.html#faq)
- [x] Check the related part of the [Documentation](https://docs.lvgl.io/)
- [x] Update lvgl to the latest version
- [x] Reproduce the issue in a [Simulator](https://docs.lvgl.io/latest/en/html/get-started/pc-simulator.html)

### Describe the bug
The current lvgl will determine the direction of the responsive scrolling according to the scrolling axis of the component when it is clicked. When the scroll axes of two components are perpendicular to each other, lvgl will take the scrolling direction of the first clicked component as the scrolling direction of the response, not the actual scrolling direction of the user.

### To Reproduce

```c
static void scroll_dir_test()
{
    lv_obj_t * obj = lv_obj_create(lv_scr_act());
    lv_obj_set_size(obj, lv_pct(200), lv_pct(80));

    lv_obj_t * slider = lv_slider_create(obj);
    lv_obj_set_size(slider, 100, lv_pct(80));
    lv_obj_center(slider);
}
```

### Expected behavior
The first scrolling direction should be based on the user's scrolling direction, which is more in line with the user's intuition.

https://user-images.githubusercontent.com/26767803/199457093-1cea4b8b-83d5-4e1b-b3b2-062d7927a9c2.mp4

### Screenshots or video

https://user-images.githubusercontent.com/26767803/199457269-b8f525ce-f5f8-4dae-8322-cfdb0172e2c6.mp4

kisvegabor:
Just to be sure I understand correctly: e.g. in case of a horizontal scroll gesture we should check if there is an object (among the parents too) which can be scrolled horizontally.  If so, actually it works like this with containers. The slider is a special case as it "absorbs" the scrolls. It's because it has 
```c
    lv_obj_clear_flag(obj, LV_OBJ_FLAG_SCROLL_CHAIN); //Absorb the scroll
    lv_obj_clear_flag(obj, LV_OBJ_FLAG_SCROLLABLE); //Children shouldn't add scrollbars and make the slider scrollable 
```

Adding
```c
  lv_obj_add_flag(slider, LV_OBJ_FLAG_SCROLL_CHAIN_HOR); //Absorb only vertical scroll
```

to your example should result in the desired behavior. 

[...]

PR #3874, opened 2022-12-19
fix(slider): improved drag direction prediction

FASTSHIFT:

After this [discussion](https://github.com/lvgl/lvgl/issues/3780), I continued to try to optimize the slider dragging, mainly by transferring the initial judgment of the drag from `LV_EVENT_PRESSED` to` LV_EVENT_PRESSING`, and adding the drag threshold detection, the drag experience has been greatly improved, see the following video:

Before:

https://user-images.githubusercontent.com/26767803/208390740-563b4133-ac24-44cf-add1-8b0b08a9d44c.mp4

After:

https://user-images.githubusercontent.com/26767803/208390810-c74d1da6-502f-4ce4-899e-9b30d53195b7.mp4

### Checkpoints
 
### DIFF
--- a/src/widgets/slider/lv_slider.c
+++ b/src/widgets/slider/lv_slider.c
@@ -37,6 +37,7 @@ static void position_knob(lv_obj_t * obj, lv_area_t * knob_area, const lv_coord_
 static void draw_knob(lv_event_t * e);
 static bool is_slider_horizontal(lv_obj_t * obj);
 static void drag_start(lv_obj_t * obj);
+static void update_knob_pos(lv_obj_t * obj, bool check_drag);
 
 /**********************
  *  STATIC VARIABLES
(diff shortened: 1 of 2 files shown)


kisvegabor (commented):
If I understood correctly with this change if you keep pressing the slider 1 pixel next to the center of the knob nothing happens, right?

If so `lv_coord_t drag_range = indev->driver->scroll_limit;` wouldn't be enough? 
Or in the `LONG_PRESS` event if there is no scrolling we can still set the pressed value even if it's less than `drag_range`.

What do you think? 
`

W-Mai:
> If I understood correctly with this change if you keep pressing the slider 1 pixel next to the center of the knob nothing happens, right?
> 
> If so `lv_coord_t drag_range = indev->driver->scroll_limit;` wouldn't be enough? Or in the `LONG_PRESS` event if there is no scrolling we can still set the pressed value even if it's less than `drag_range`.
> 
> What do you think? `

I think `* 2` is equivalent to the effect of a dead zone. You can `*` any value, and different values will have different hand feeling.

kisvegabor:
> I think * 2 is equivalent to the effect of a dead zone. You can * any value, and different values will have different hand feeling.

My problem is that I feel `*2` an arbitrary value. Why `*2`? Why not `*3`? I suggest keeping the original `scroll_limit` is it works well too.

What about this?
> Or in the LONG_PRESS event if there is no scrolling we can still set the pressed value even if it's less than drag_range.

W-Mai:
`*2` is more comfortable than `*1`.
It's a little subjective in it.

> Or in the LONG_PRESS event if there is no scrolling we can still set the pressed value even if it's less than drag_range.

I think it's ok. How about @FASTSHIFT ?

FASTSHIFT:
> My problem is that I feel `*2` an arbitrary value. Why `*2`? Why not `*3`? I suggest keeping the original `scroll_limit` is it works well too.

OK.

>  Or in the LONG_PRESS event if there is no scrolling we can still set the pressed value even if it's less than drag_range.

I looked at how my phone would behave. If I clicked on an area outside of knob, the slider wouldn't update the new value until the finger was lifted, indicating that it was using the click event. Should we consider using the click event?

FASTSHIFT:
I tried to optimize the drag and click again, it is now exactly the same experience as on a smartphone :)


https://user-images.githubusercontent.com/26767803/209169485-e22236b9-214d-4d91-9f7e-3ef3090e4fff.mp4

kisvegabor:
Amazing! Looks great, merging! :slightly_smiling_face:

PR #3891, opened 2022-12-29
fix(core): when a gesture detected, a followed by click shouldn't trig

somebodyLi:

### Checkpoints
 
### DIFF
--- a/src/core/lv_indev.c
+++ b/src/core/lv_indev.c
@@ -973,7 +973,7 @@ static void indev_proc_release(_lv_indev_proc_t * proc)
 
         /*Send CLICK if no scrolling and no previous gesture detected*/
         if(scroll_obj == NULL && proc->types.pointer.gesture_sent == 0) {
-            if(proc->long_pr_sent == 0) {
+            if(proc->long_pr_sent == 0  && proc->types.pointer.gesture_sent == 0) {
                 lv_event_send(indev_obj_act, LV_EVENT_SHORT_CLICKED, indev_act);
                 if(indev_reset_check(proc)) return;
             }
--- a/src/core/lv_indev.c
+++ b/src/core/lv_indev.c
@@ -973,7 +973,7 @@ static void indev_proc_release(_lv_indev_proc_t * proc)
 
         /*Send CLICK if no scrolling and no previous gesture detected*/
         if(scroll_obj == NULL && proc->types.pointer.gesture_sent == 0) {
-            if(proc->long_pr_sent == 0  && proc->types.pointer.gesture_sent == 0) {
+            if(proc->long_pr_sent == 0) {
                 lv_event_send(indev_obj_act, LV_EVENT_SHORT_CLICKED, indev_act);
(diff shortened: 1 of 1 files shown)


src/core/lv_indev.c
```diff
@@ -971,8 +971,8 @@ static void indev_proc_release(_lv_indev_proc_t * proc)
         lv_event_send(indev_obj_act, LV_EVENT_RELEASED, indev_act);
         if(indev_reset_check(proc)) return;
 
-        /*Send CLICK if no scrolling*/
-        if(scroll_obj == NULL) {
+        /*Send CLICK if no scrolling and no previous gesture detected*/
+        if(scroll_obj == NULL && proc->types.pointer.gesture_sent == 0) {
             if(proc->long_pr_sent == 0) {
```
kisvegabor, line 976: The condition should be added one line below to process the `else` part regardless to the gesture.
```suggestion
            if(proc->long_pr_sent == 0  && proc->types.pointer.gesture_sent == 0) {
```

kisvegabor (commented):
Thank you, I was also about to add a similar fix.

Just a minor thing.

src/core/lv_indev.c
somebodyLi: only this line is not enough? the click should be trig also ?

kisvegabor:
Ah, you are right. I was lost in the `if`s :slightly_smiling_face: 

Thank you!

kisvegabor:
I was about pressing the merge button when something came to my mind. Let say there is a chart (or any other object) which is not scrollable. You want an event on CLICK but as the user clicks by accident he sliders his finger as if it was a gesture. As the chart is not interested in the gestures the user would expect to get CLICK event even if the finger was slid. 

In general the user should be able to freely slide his finger and get CLICK in the end if he is no interested in the gestures. 

To solve it we could change `LV_OBJ_FLAG_GESTURE_BUBBLE` to `LV_OBJ_FLAG_GESTURE_ENABLE`. It would work almost like `LV_OBJ_FLAG_GESTURE_BUBBLE` (if it's enabled on an object it absorbs the gesture, if not enabled it propagates it to the parent). However, it can be disabled by default on all widgets (including screens). And finally we should send a gesture only if it was enabled on on object or one of its parent. This way we could know that whether the gesture was really processed/used and whether we should send CLICK or not.

What do you think?

somebodyLi:
> get CLICK in the end if he is no interested in the gestures.

Sorry, I can't agree with you on this view .  `GESTURE` and `CLICK` anyway should be mutually exclusive

kisvegabor:
The problem is that we don't know if the gesture was consumed or not. So we need a solution to indicate whether the user is interested in the gesture event or not. 

That's why I've suggested  to have `LV_OBJ_FLAG_GESTURE_ENABLE`. And once it's set, we will know that the gesture is consumed.

However (it just cam to my mind), what if the user is interested only in the LEFT gesture but the gestures of the other directions should ignored and treated as CLICK/SCROLL/etc.

Issue #3714, opened 2022-09-30
Troubles with STM32 DMA2D acceleration


tdjastrzebski:
Yesterday I have tried enabling STM32 DMA2D acceleration with D-cache enabled and MPU disabled.  
The results were not great, lots of 'garbage' on the screen, typically resulting from D-cache problems. Problems were clearly visible only with `lv_demo_music`. I reviewed [lv_gpu_stm32_dma2d.c](https://github.com/lvgl/lvgl/blob/965552e4465772dd3c12892216412fc01e651f6e/src/draw/stm32_dma2d/lv_gpu_stm32_dma2d.c) file and several question marks popped up, but first some background. I think the problem to be solved is that DMA (including DMA2D) does not clean (flush) the source buffer before starting the transfer, so dirty reads may happen.  
1. Why do we clean AND **invalidate** cache before transfer? I think cleaning only should be enough.
2. Why do we clen the entire memory cache? Cleaning only source buffer area should be sufficient.
3. Why do we clean/invalidate cache before filling the buffer with color in `lv_draw_stm32_dma2d_blend_fill()` method? There is no source buffer to be cleaned/flushed.
4. I can be wrong here, but I think source buffer address should be 32b-aligned and data length must be multiple of cache line size (32b). If so, do we ensure that?  

What I suggest:
1. Using `CleanDCache_by_Addr_Aligned()` method (below) or similar.
2. Modifying `clean_dcache_cb` so it takes address and length args.  

After making the changes STM32 DMA2D acceleration on my STM32F769I-DISCO board works without visible glitches but I am still not certain about source buffer alignment.

```C
void CleanDCache_by_Addr_Aligned(uint32_t sourceAddress, uint32_t dsize) {
	uint32_t sourceAddressAligned = sourceAddress & ~(uint32_t)(32U - 1);
	uint32_t lengthAligned = dsize + (sourceAddress & (uint32_t)(32U - 1));
	lengthAligned += 32U - (lengthAligned % 32U);
	__DSB();
	while (lengthAligned > 0) {
		SCB->DCCMVAC = sourceAddressAligned;  // register accepts only 32byte aligned values, only bits 31..5 are valid
		sourceAddressAligned += 32U;
		lengthAligned -= 32U;
	};
	__DSB();
	__ISB();
}
```
![IMG_E6828B](https://user-images.githubusercontent.com/7325277/193229854-c447ecb4-4a00-44ad-a11a-29a24d043458.JPG)

[...]

PR #3899, opened 2023-01-03
STM32 DMA2D support rework (8.3)

tdjastrzebski:
This PR is a major rework of the current LVGL STM32 DMA2D support. It adds more hardware acceleration, especially for bitmap-blended solid colors used (e.g.) to draw anti-aliased fonts.
The feature has been fairly tested using LVGL demos with both 32 and 16 bit colors, but it still should be considered as experimental.
See https://github.com/lvgl/lvgl/issues/3714 for full discussion, especially [this](https://github.com/lvgl/lvgl/issues/3714#issuecomment-1365187036) post.


tdjastrzebski:
I will reissue this PR as I realized more code cleanup was needed.

PR #3904, opened 2023-01-04
stm32 dma2d rework (8.3)

### LINKED ISSUE #3714: Troubles with STM32 DMA2D acceleration

tdjastrzebski:
This PR is a major rework of the current LVGL STM32 DMA2D support. It adds more hardware acceleration, especially for bitmap-blended solid colors used (e.g.) to draw anti-aliased fonts. In 32-bit color mode all four types of rectangles are painted with dma2d. In 16-bit color three types are dma2d painted, including the type which is the most performance critical.
The feature has been fairly tested using LVGL demos with both 32 and 16 bit colors, but it should still be considered experimental.
Resolves #3714, see [this](https://github.com/lvgl/lvgl/issues/3714#issuecomment-1365187036) post for a summary.

### DIFF
--- a/src/draw/stm32_dma2d/lv_gpu_stm32_dma2d.c
+++ b/src/draw/stm32_dma2d/lv_gpu_stm32_dma2d.c
@@ -72,6 +72,7 @@ void lv_draw_stm32_dma2d_ctx_init(lv_disp_drv_t * drv, lv_draw_ctx_t * draw_ctx)
 
     dma2d_draw_ctx->blend = lv_draw_stm32_dma2d_blend;
     dma2d_draw_ctx->base_draw.draw_img_decoded = lv_draw_stm32_dma2d_img_decoded;
+    //dma2d_draw_ctx->base_draw.draw_img = lv_draw_stm32_dma2d_img;
     // Note: currently it does not make sense use lv_gpu_stm32_dma2d_wait_cb() since waiting starts right after the dma2d transfer
     //dma2d_draw_ctx->base_draw.wait_for_finish = lv_gpu_stm32_dma2d_wait_cb;
     dma2d_draw_ctx->base_draw.buffer_copy = lv_draw_stm32_dma2d_buffer_copy;
@@ -90,11 +91,11 @@ static void lv_draw_stm32_dma2d_blend(lv_draw_ctx_t * draw_ctx, const lv_draw_sw
         return;
     }
     // Note: x1 must be zero. Otherwise, there is no way to correctly calculate dest_stride.
-    //LV_ASSERT_MSG(draw_ctx->buf_area->x1 == 0); // critical
+    //LV_ASSERT_MSG(draw_ctx->buf_area->x1 == 0); // critical?
     // Both draw buffer start address and buffer size *must* be 32-byte aligned since draw buffer cache is being invalidated.
     //uint32_t drawBufferLength = lv_area_get_size(draw_ctx->buf_area) * sizeof(lv_color_t);
     //LV_ASSERT_MSG(drawBufferLength % CACHE_ROW_SIZE == 0); // critical, but this is not the way to test it
-    LV_ASSERT_MSG((uint32_t)draw_ctx->buf % CACHE_ROW_SIZE == 0, "draw_ctx.buf is not 32B aligned"); // critical
+    //LV_ASSERT_MSG((uint32_t)draw_ctx->buf % CACHE_ROW_SIZE == 0, "draw_ctx.buf is not 32B aligned"); // critical?
 
     if(dsc->src_buf) {
         // For performance reasons, both source buffer start address and buffer size *should* be 32-byte aligned since source buffer cache is being cleaned.
@@ -199,7 +200,7 @@ static void lv_draw_stm32_dma2d_buffer_copy(lv_draw_ctx_t * draw_ctx, void * des
     // Both draw buffer start address and buffer size *must* be 32-byte aligned since draw buffer cache is being invalidated.
     //uint32_t drawBufferLength = lv_area_get_size(draw_ctx->buf_area) * sizeof(lv_color_t);
     //LV_ASSERT_MSG(drawBufferLength % CACHE_ROW_SIZE == 0); // critical, but this is not the way to test it
-    LV_ASSERT_MSG((uint32_t)draw_ctx->buf % CACHE_ROW_SIZE == 0, "draw_ctx.buf is not 32B aligned"); // critical
+    //LV_ASSERT_MSG((uint32_t)draw_ctx->buf % CACHE_ROW_SIZE == 0, "draw_ctx.buf is not 32B aligned"); // critical?
     // FIXME:
     // 1. Both src_buf and dest_buf pixel size *must* be known to use DMA2D.
     // 2. Verify both buffers start addresses and lengths are 32-byte (cache row size) aligned.
@@ -235,6 +236,7 @@ static void lv_draw_stm32_dma2d_img_decoded(lv_draw_ctx_t * draw_ctx, const lv_d
                                        img_dsc->opa, bitmapColorFormat, ignoreBitmapAlpha);
     }
     else {
+        // all more complex cases which require additional image transformations
         lv_draw_sw_img_decoded(draw_ctx, img_dsc, coords, src_buf, color_format);
 
     }
@@ -295,7 +297,7 @@ static lv_res_t lv_draw_stm32_dma2d_img(lv_draw_ctx_t * draw_ctx, const lv_draw_
     // Both draw buffer start address and buffer size *must* be 32-byte aligned since draw buffer cache is being invalidated.
     //uint32_t drawBufferLength = lv_area_get_size(draw_ctx->buf_area) * sizeof(lv_color_t);
     //LV_ASSERT_MSG(drawBufferLength % CACHE_ROW_SIZE == 0); // critical, but this is not the way to test it
-    LV_ASSERT_MSG((uint32_t)draw_ctx->buf % CACHE_ROW_SIZE == 0, "draw_ctx.buf is not 32B aligned"); // critical
+    //LV_ASSERT_MSG((uint32_t)draw_ctx->buf % CACHE_ROW_SIZE == 0, "draw_ctx.buf is not 32B aligned"); // critical?
 
     // For performance reasons, both source buffer start address and buffer size *should* be 32-byte aligned since source buffer cache is being cleaned.
     //uint32_t srcBufferLength = lv_area_get_size(src_area) * sizeof(lv_color_t); // TODO: verify src pixel size = sizeof(lv_color_t)
--- a/src/draw/stm32_dma2d/lv_gpu_stm32_dma2d.c
+++ b/src/draw/stm32_dma2d/lv_gpu_stm32_dma2d.c
@@ -71,7 +71,7 @@ void lv_draw_stm32_dma2d_ctx_init(lv_disp_drv_t * drv, lv_draw_ctx_t * draw_ctx)
     lv_draw_stm32_dma2d_ctx_t * dma2d_draw_ctx = (lv_draw_sw_ctx_t *)draw_ctx;
 
     dma2d_draw_ctx->blend = lv_draw_stm32_dma2d_blend;
-    dma2d_draw_ctx->base_draw.draw_img = lv_draw_stm32_dma2d_img;
+    dma2d_draw_ctx->base_draw.draw_img_decoded = lv_draw_stm32_dma2d_img_decoded;
     // Note: currently it does not make sense use lv_gpu_stm32_dma2d_wait_cb() since waiting starts right after the dma2d transfer
     //dma2d_draw_ctx->base_draw.wait_for_finish = lv_gpu_stm32_dma2d_wait_cb;
     dma2d_draw_ctx->base_draw.buffer_copy = lv_draw_stm32_dma2d_buffer_copy;
@@ -83,7 +83,7 @@ void lv_draw_stm32_dma2d_ctx_deinit(lv_disp_drv_t * drv, lv_draw_ctx_t * draw_ct
     LV_UNUSED(draw_ctx);
 }
 
-void lv_draw_stm32_dma2d_blend(lv_draw_ctx_t * draw_ctx, const lv_draw_sw_blend_dsc_t * dsc)
+static void lv_draw_stm32_dma2d_blend(lv_draw_ctx_t * draw_ctx, const lv_draw_sw_blend_dsc_t * dsc)
 {
     if(dsc->blend_mode != LV_BLEND_MODE_NORMAL) {
         lv_draw_sw_blend_basic(draw_ctx, dsc);
@@ -116,7 +116,6 @@ void lv_draw_stm32_dma2d_blend(lv_draw_ctx_t * draw_ctx, const lv_draw_sw_blend_
     if(dsc->mask_buf && dsc->mask_res == LV_DRAW_MASK_RES_TRANSP) return;
     else if(dsc->mask_res == LV_DRAW_MASK_RES_FULL_COVER) mask = NULL;
 
-
     lv_coord_t dest_stride = lv_area_get_width(draw_ctx->buf_area);
     if(mask != NULL) {
         // For performance reasons, both mask buffer start address and buffer size *should* be 32-byte aligned since mask buffer cache is being cleaned.
@@ -125,9 +124,7 @@ void lv_draw_stm32_dma2d_blend(lv_draw_ctx_t * draw_ctx, const lv_draw_sw_blend_
         //LV_ASSERT_MSG((uint32_t)mask % CACHE_ROW_SIZE == 0); // FIXME: assert fails (performance, non-critical)
 
         lv_coord_t mask_stride = lv_area_get_width(dsc->mask_area);
-        lv_point_t mask_offset; // mask offset in relation to draw_area
-        mask_offset.x = draw_area.x1 - dsc->mask_area->x1;
-        mask_offset.y = draw_area.y1 - dsc->mask_area->y1;
+        lv_point_t mask_offset = lv_area_get_offset(dsc->mask_area, &draw_area); // mask offset in relation to draw_area
 
         if(dsc->src_buf == NULL) {  // 93.5%
             lv_area_move(&draw_area, -draw_ctx->buf_area->x1, -draw_ctx->buf_area->y1);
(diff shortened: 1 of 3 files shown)


src/draw/stm32_dma2d/lv_gpu_stm32_dma2d.c
```diff
@@ -113,153 +88,560 @@ void lv_draw_stm32_dma2d_ctx_deinit(lv_disp_drv_t * drv, lv_draw_ctx_t * draw_ct
...
+ * @param color color to be painted, note: alpha is ignored
+ */
+LV_STM32_DMA2D_STATIC void _lv_draw_stm32_dma2d_blend_fill(const lv_color_t * dest_buf, lv_coord_t dest_stride,
+                                                           const lv_area_t * draw_area, lv_color_t color, lv_opa_t opa)
 {
-    /*Simply fill an area*/
-    int32_t area_w = lv_area_get_width(fill_area);
-    int32_t area_h = lv_area_get_height(fill_area);
-    invalidate_cache();
+    assert_param(!isDma2dInProgess); // critical
```
kisvegabor: Why assert here?

src/draw/stm32_dma2d/lv_gpu_stm32_dma2d.c
```diff
@@ -11,100 +11,75 @@
 
 #if LV_USE_GPU_STM32_DMA2D
 
+#include "stdio.h"
+#include "main.h"
```
kisvegabor: It should be removed.

src/draw/stm32_dma2d/lv_gpu_stm32_dma2d.c
```diff
@@ -11,100 +11,75 @@
 
 #if LV_USE_GPU_STM32_DMA2D
 
+#include "stdio.h"
```
kisvegabor: Why is it needed?

kisvegabor (commented):
Thank you for sending this PR. 

I wanted to try it out bu due to `#include "main.h"` it wasn't working.

src/draw/stm32_dma2d/lv_gpu_stm32_dma2d.c
tdjastrzebski: That's right, both those includes (**stdio.h** and **main.h**) can be removed, but `LV_GPU_DMA2D_CMSIS_INCLUDE` needs to be set to (e.g.) **"stm32f7xx_hal.h"**, rather than **"stm32f769xx.h"**. This way constants like `DMA2D_INPUT_ARGB8888` can be included from HAL, rather than redefined in LVGL. Consequently, it would probably make sense to rename `LV_GPU_DMA2D_CMSIS_INCLUDE` to `LV_STM32_HAL_INCLUDE` or similar. Unless dependency on *_hal.h is not desired. There are pros and cons. 
What do you think?
I hate conditional includes as they often break intellisense, but probably there is no other way around if including **main.h** is not a good idea either.

src/draw/stm32_dma2d/lv_gpu_stm32_dma2d.c
tdjastrzebski: This purpose for this assert is to ensure that any previous dma2d transfer initiated by this driver was properly awaited for and finished.
This applies only to this driver, since dma2d can be used by the user for other purposes, e.g. in flush_cb().

src/draw/stm32_dma2d/lv_gpu_stm32_dma2d.c
tdjastrzebski: See the previous response above.

tdjastrzebski:
Probably there is one more minor change required. In two places I use `printf()` for debug trace. I think LVGL established own practice. What should I use instead?

src/draw/stm32_dma2d/lv_gpu_stm32_dma2d.c
kisvegabor: It worked thanks.

kisvegabor:
I've tested the performance of the new drivers but I got mixed results. 

### lv_demo_widgtes
I used `LV_USE_PERF_MONITOR` and got 2-3 more FPS **without** DMA2D. 

 ### lv_demo_benchmark
**Without DMA2D**
image

**With DMA2D**
image


So for simple fill software rendering seemed faster. However for will with opa DMA2D was faster.

---

I also found that the images was not loaded with DMA2D (I haven't debugged why). Only "No data" was displayed.

---

> Probably there is one more minor change required. In two places I use printf() for debug trace. I think LVGL established own practice. What should I use instead?

`LV_LOG_ERROR` can be used.

tdjastrzebski:
I made PR update. Missing images issue should be resolved. Probably something has changed in codebase in the meantime causing demo bitmap images to change their color format (`img.header.cf`) while SW render fallback did not work correctly. That inspired further image bitmap drawing improvements, more color formats are handled by dma2d.
I would not expect spectacular overall performance improvements. On STM32F769NI-DISCO board I see 15% performance gain (total `lv_demo_benchmark` time). This improvement is relatively small probably because copying/blending to/with draw buffer accounts only for a fraction of LVGL CPU time.
However, when new LVGL architecture is in place and dma2d 'transfer complete' event is used instead of wait loops, I expect dma2d to noticeably offload CPU. This is the very reason which prompted me to contribute to this project. I need CPU time primarily for DSP. Nice graphics is important, albeit not critical.
I completely disabled dest buffer L1 cache invalidation, leaving only src cache cleaning. This seems to be OK under two conditions: 1. dma2d is used within `flush_cb()`, 2. all draw buffer writes are done with dma2d. The later is not always the case. For that reason, especially in 16B color mode, `SCB_CleanDCache_by_Addr()` may need to be called within `flush_cb()` before dma2d buffer copy. That still does not guarantee correct substractions and I do not see a good solution within the current architecture BESIDES disabling draw buffer L1 cache with MPU. I may want to further explore this option when I have more time, but I am not certain this should be part of this stm32_dma2d driver.
**In general, combining dma2d and cached SW draw buffer writes should better be avoided.** It causes nothing but L1 cache maintenance headache.

tdjastrzebski:
@kisvegabor Please review

kisvegabor:
Thank you for the updates. I can try it out later today.

> However, when new LVGL architecture is in place and dma2d 'transfer complete' event is used instead of wait loops, I expect dma2d to noticeably offload CPU. This is the very reason which prompted me to contribute to this project. I need CPU time primarily for DSP. Nice graphics is important, albeit not critical.

So let's consider these changes as an experiment (we can still merge it) and we will do it better in v9's architecture, right?

> 2. all draw buffer writes are done with dma2d. 

SW render also writes the draw buffers. Do you mean frame buffer?

> In general, combining dma2d and cached SW draw buffer writes should better be avoided. It causes nothing but L1 cache maintenance headache.

You have mentioned that it's possible but there is any performance impact of using cache on internal RAM?

tdjastrzebski:
Agreed, I suggest to label those changes as experimental. Probably I will use them in prod with 32B color.
> all draw buffer writes are done with dma2d.

Exactly, because of SW draw buffer rights, cache maintenance is critical. In 32B color mode I managed to eliminate direct SW writes almost completely, probably only besides subtractions and images other than those dma2d-supported (RGB888, ARGB8888, RGB565). It is not possible to achieve it in 16B color for two reasons: (1) use of ARGB8565 color not supported by dma2d and (2) the need for extra buffer to combine mask with RGB565 bitmap before dma2d blending. It is not an issue with 32B color since (1) ARGB8565 is not used that much, (2) ARGB8888 is primarily used so mask byte can be simply copied to ARGB8888 bitmap before blending with BG by dma2d.

Cache on internal RAM could be disabled selectively for draw buffer region only. See [`MpuRamConfig()`](https://github.com/tdjastrzebski/STM32F769I-DISCO-LVGL/blob/8c0c6695d37807cc74ec64d6a9b145cf6fb96107/master.cpp#L233). It is a work in progress and does not seem to work yet.

Btw, I get way higher FPS results, it can be a bug - see #3933

tdjastrzebski:
At this point, there is at least one remaining issue: should `LV_GPU_DMA2D_CMSIS_INCLUDE` be renamed to `LV_STM32_HAL_INCLUDE`? HAL include is needed to avoid redefining in LVGL DMA2D masks and constants like e.g. `DMA2D_FGPFCCR_RBS_Pos`, which (unfortunately) are not defined in CMSIS but in device HAL. There are pros and cons, convenience vs. increased ST lib dependency. I would opt for convenience. If so, **lv_conf_template.h** should be modified too:

```C
#if LV_USE_GPU_STM32_DMA2D
    /*Must be defined to include path of CMSIS header of target processor
    e.g. "stm32f7xx_hal.h" or "stm32f429xx_hal.h"*/
    #define LV_GPU_STM32_HAL_INCLUDE
#endif
```

kisvegabor:
I've tried it out and it's working well. It became slightly faster indeed.

> HAL include is needed to avoid redefining in LVGL DMA2D masks and constants like e.g. DMA2D_FGPFCCR_RBS_Pos, which (unfortunately) are not defined in CMSIS but in device HAL.

Actually, we have intentionally removed the dependency to HAL for more felxibility. 

`DMA2D_FGPFCCR_RBS_Pos` seems to be in `stm32f769xx.h`:
image

Only `DMA2D_INPUT_A8` seems only missing which not that much and we can redefine them.

> Agreed, I suggest to label those changes as experimental. Probably I will use them in prod with 32B color.

We can't really label something as experimental if merge into a patch release. However, if it's not critical to merge it for you we can keep this PR open, and continue to work on it when v9 is ready for that.

tdjastrzebski:
That's right, `DMA2D_FGPFCCR_RBS_Pos` probably was not the best example. I removed HAL dependency.

I would definitely merge it.
1) It should not break anything since probably no one is/was using DMA2D other than in flush_cb().
2) I think, and even hope, once it makes it to the 8.3 release branch more people will be able to contribute.
3) I am not certain about 16B color, but I believe with 32B color this version can be used "in prod" as it is right now. I already use it and merging would help.

bhspyder:
Tested this with the audio mixer GUI example and for some reason it breaks the level sliders and the waveform slider. Everything works as expected running the same example with lvgl release/v8.3.4.

kisvegabor:
Hi Brandon,

Thanks for the feedback! Do you mean the Audio mixer example from SquareLine or `lv_demo_music()`?

If the former, which board do you use? I don't know any STM boards that has 480x800 portrait display.

bhspyder:
I'm using the audio mixer example from SquareLine. I have it running on custom hardware (480x854). I just resized the project in SquareLine to match. The behavior I'm seeing is the slider image showing at 100% for any value other than 0 until it drops to 0 then it shows 0.

tdjastrzebski:
@bhspyder It does not sound like a rendering issue, but I can take a look. Where do I find the source code?

bhspyder:
I can't share the whole project, but I created a quick repo with my lv_config.h and the ui files generated by SquareLine:
https://github.com/bhspyder/lvgl_audiomixer_stm32f746xx.git

tdjastrzebski:
@bhspyder, One more fix posted, please test, there is a good chance it fixes the issue.
@kisvegabor, After timer issue fix (#3933 ) I get correct lv_demo_benchmark() results.
Without DMA2D I get weighted 34fps, 88% opa speed. With DMA2D numbers are 42fps and 97%. I am nicely surprised :)
Tested with STM32F769NI-DISCO and 32B color, complete code with env setup [here](https://github.com/tdjastrzebski/STM32F769I-DISCO-LVGL/tree/DMA2D)

bhspyder:
@tdjastrzebski I tested the latest, but the issue is still present. It starts working again if the only change I make is to change the lvgl directory to point to the release/v8.3.4 tag in the main lvgl repo. If it helps, I am using direct-mode double buffered with framebuffers in external SDRAM and the sliders are using images. The sliders are using the image as an indicator so it would be needing to render only part of the image depending on the slider value.

tdjastrzebski:
@bhspyder As far as I can tell, audiomixer works fine on my STM32F769NI-DISCO board, faders move smoothly, external SDRAM used as well, working mixer_demo code [here](https://github.com/tdjastrzebski/STM32F769I-DISCO-LVGL/tree/mixer_demo)
See: https://drive.google.com/file/d/1OcXRmKgrVAkyBh3dXMm8wkvgde12I-8L/view
The problem may be within flush_cb() handler. Keep in mind that dma(2d) does not use L1 cache, try invalidating draw buffer cache (SCB_InvalidateDCache_by_Addr()) before copying screen bytes, unless you use dma(2d) to do that.
I do not think I can have any other clues without access to your test environment.

bhspyder:
@tdjastrzebski the MAX sliders you used in the video are not the issue. It's the red->green LEVEL indicators and the gray waveform at the top that are not working. I am using direct-mode with true double buffering. I am swapping active frame buffers in my flush_cb and syncing the changed areas to the inactive buffer using DMA2D.

tdjastrzebski:
@bhspyder, Does my demo replicate the problem? If so, please share some video for comparison.
However, if my test demo does not replicate the problem, than there is no way I can help/advise without a repro.
Btw, another pitfall I can think of: dma2d registry settings are not reinitialized properly. Keep in mind that my lvgl dma2d driver may modify them completely between flush_cb() calls. This problem can be identified using memcpy() instead of dma2d. Sample code [here](https://github.com/tdjastrzebski/STM32F769I-DISCO-LVGL/blob/d59655afee004c8f51b6ec1da48495aa66d745a0/master.cpp#L221)

bhspyder:
Sorry for the confusion. The video did not demonstrate the issue. See the attached image. The areas highlighted in red are draggable sliders and as you can see in the image, they should be able to be set to different levels between 0 - 100%. My issue is I am not seeing those change when I drag on them. They stay at 100% until I drag to where 0% would be and then they show the empty image.

lvgl_audio_mixer

tdjastrzebski:
@bhspyder Now I see what you mean. Could you create UI containing one problematic (level indicator) slider only?
I mean, repro as simple as possible.

bhspyder:
@tdjastrzebski  I updated the repo with a UI simplified to only having the affected sliders:
https://github.com/bhspyder/lvgl_audiomixer_stm32f746xx.git

tdjastrzebski:
@bhspyder thank you, I will look into this as it seems it is indeed rendering defect. I think that some extra clipping or mask should be applied when drawing the slider bitmap and it is not applied correctly.

tdjastrzebski:
@bhspyder Please check the latest version, I believe the problem is fixed.
@kisvegabor To make the longs story short, I think to draw bitmap image dma2d driver should use `lv_draw_stm32_dma2d_img_decoded()` function, rather than `lv_draw_stm32_dma2d_img()` and tap into the pipeline a bit later.

bhspyder:
@tdjastrzebski Yup that resolved the issue on my end 😄

tdjastrzebski:
@kisvegabor Unless someone else is willing to test it, as far as I am concerned, this PR is ready to be merged.
Further improvements would require changes in LVGL codebase - something to think about for the rev 9.

tdjastrzebski:
I checked 16B performance with DMA2D. I got weighted **50fps**, 96%opa vs 38fps, 89% opa without DMA2D. I think this 31% improvement is respectable, it is even higher than improvement in 32B color format. I used the same [STM32F769NI-DISCO demo](https://github.com/tdjastrzebski/STM32F769I-DISCO-LVGL/tree/DMA2D).

From the dma2d driver perspective, 5 things could be improved in rev 9:
1. Cleaner driver interface. Once we know what hardware drivers are capable of, better interface could be crafted. I would start with a reference/default SW driver which could be swapped and used exactly like a hardware driver, rather then use SW fallbacks all over the place, particularly inside the driver - that would also simplify testing. Such driver functions should be loosely coupled and, in particular, do NOT contain logic checking if request can be handled, e.g. if zoom=0. **Just buffer pointers and clear description of the operation to be performed on them should be enough.** That is for the blending, drawing (e.g. arcs), if HW supported, is a different story. I think drivers could also have some kind of manifest, simplifying the config. Also, clear, consistent terminology use (draw, blend, paint, etc.) would be helpful.
2. Implement awaitable pattern, instead of waiting for a hardware to complete in wait loops.
3. Implement better buffer memory alignment, **including buffer length**. Several extra reserved bytes would not cause a problem, but it would allow for safe output (dest) buffer L1 cache maintenance (invalidations). User can manage draw buffer alignment, but not internal buffers, which, I think, are being used as dma2d output buffers as well - e.g. subtractions.
4. Reduce the number of small areas being blended, e.g. small font characters. Why not to blend several small characters at once? This should substantially speed up HW blending by reducing the number of calls, thus total dma2d setup time/overhead.
5. Make sure source buffers content remains intact during blending to prevent ugly artefacts showing up.

kisvegabor:
Thank you guys. I'm merging it. 

@tdjastrzebski 
 > From the dma2d driver perspective, 4 things could be improved in rev 9: [...]

In general I agree with the points you have listed. We are heading into this direction. I hope in February I can get back to the new drawing architecture and make some progress. I'll tag you when I'll have something to discuss.

PR #3910, opened 2023-01-09
fix(grid): correctly account for side padding with disabled borders

arbrauns:

The border width cannot be naively added to the padding, since borders can be enabled and disabled per-side. The `lv_obj_get_style_space_*()` functions account for this.

### Checkpoints
- Rest: N/A, trivial change


kisvegabor:
Thank you!

PR #3911, opened 2023-01-09
lv_line: allow LV_PCT for point coordinates

arbrauns:

This allows specifying the coordinates of line points using a percentage of the line's overall bounding box.

### Checkpoints
 
- Rest: N/A

### DIFF
--- a/src/widgets/line/lv_line.c
+++ b/src/widgets/line/lv_line.c
@@ -121,6 +121,16 @@ static void lv_line_constructor(const lv_obj_class_t * class_p, lv_obj_t * obj)
     LV_TRACE_OBJ_CREATE("finished");
 }
 
+static inline lv_coord_t resolve_point_coord(lv_coord_t coord, lv_coord_t max)
+{
+    if(LV_COORD_IS_PCT(coord)) {
+        return LV_CLAMP(0, max * LV_COORD_GET_PCT(coord) / 100, max);
+    }
+    else {
+        return coord;
+    }
+}
+
 static void lv_line_event(const lv_obj_class_t * class_p, lv_event_t * e)
 {
     LV_UNUSED(class_p);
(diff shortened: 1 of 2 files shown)


src/widgets/line/lv_line.c
```diff
@@ -182,17 +186,25 @@ static void lv_line_event(const lv_obj_class_t * class_p, lv_event_t * e)
         for(i = 0; i < line->point_num - 1; i++) {
             lv_point_t p1;
             lv_point_t p2;
-            p1.x = line->point_array[i].x + x_ofs;
-            p2.x = line->point_array[i + 1].x + x_ofs;
+
+            lv_coord_t w = lv_obj_get_width(obj);
+            lv_coord_t h = lv_obj_get_height(obj);
+
+            p1.x = lv_clamp_width(line->point_array[i].x, 0, w, w) + x_ofs;
```
kisvegabor: Why `lv_clamp_width` is needed here?  `min/max-width` is rather related to `LV_EVENT_GET_SELF_SIZE` but min-max is handled by LVGL internally.

However I can't see where the percentage coordinates are converted to pixel . Am I missing something?

arbrauns: That happens inside `lv_clamp_width`, which is why I used it. I couldn't find any other utility function that does this.

kisvegabor:
```c
lv_coord_t lv_clamp_width(lv_coord_t width, lv_coord_t min_width, lv_coord_t max_width, lv_coord_t ref_width)
{
    if(LV_COORD_IS_PCT(min_width)) min_width = (ref_width * LV_COORD_GET_PCT(min_width)) / 100;
    if(LV_COORD_IS_PCT(max_width)) max_width = (ref_width * LV_COORD_GET_PCT(max_width)) / 100;
    return LV_CLAMP(min_width, width, max_width);
}
```

It converts only converts only `min_width` and `max_width` but keeps `width` as it is. (Just clamps it between min and max).

arbrauns: Huh, you're right, I misread that. It only happened to work for me because `LV_PCT(100)` is some huge number that then got clamped down to the width, which results in the same thing. Should be fixed now.

kisvegabor:
Looks good, thank you! :slightly_smiling_face:

PR #3920, opened 2023-01-12
arch/cmsis pack enhancement

bbkarim:

- LVGL 9.0.0-dev
- Refactor CMSIS PDSC file
  * Reworked components to get more granulated sub-groups
  * Renamed LVGL-Essential into LVGL-CORE Clean-up its files
  * Add missing hearder files
  * Include only selected headerfiles by CMSIS pack configuration
  * Clean-up LVGL-CORE files
  * Updated GPU STM32-DMA2D to support STM32U5xx
  * Remove compilation warnings
  * Add more widgets and demos

### DIFF
--- a/Kconfig
+++ b/Kconfig
@@ -787,73 +787,80 @@ menu "LVGL configuration"
 			default y if !LV_CONF_MINIMAL
 	endmenu
 
-	menu "Extra Widgets"
-		config LV_USE_ANIMIMG
-			bool "Anim image."
-			default y if !LV_CONF_MINIMAL
-		config LV_USE_CALENDAR
-			bool "Calendar."
-			default y if !LV_CONF_MINIMAL
-		config LV_CALENDAR_WEEK_STARTS_MONDAY
-			bool "Calendar week starts monday."
-			depends on LV_USE_CALENDAR
-		config LV_USE_CALENDAR_HEADER_ARROW
-			bool "Use calendar header arrow"
-			depends on LV_USE_CALENDAR
-			default y
-		config LV_USE_CALENDAR_HEADER_DROPDOWN
-			bool "Use calendar header dropdown"
-			depends on LV_USE_CALENDAR
-			default y
-		config LV_USE_CHART
-			bool "Chart."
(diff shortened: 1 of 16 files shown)


GorgonMeducer:
PLEASE do NOT mix patches about different topics into one PR.
This PR worths at least 7 separated small PRs. 


@kisvegabor

GorgonMeducer:
> Renamed LVGL-Essential into LVGL-CORE Clean-up its files

Are there any special reason that the name 'LVGL-CORE' is better than 'LVGL-Essential'? 

When I first created the LVGL cmsis-pack, I choose this name. I know it is no better than core, but the same there is no special reason that core is better than essential. 

Suppose you have contributed some code to an opensource project and someone later rename it without practice reason, will you be comfortable? 

@kisvegabor sorry for disturbing, do we have any suggestions about similar situations when doing contribution?

bbkarim:
> > Renamed LVGL-Essential into LVGL-CORE Clean-up its files
> 
> Are there any special reason that the name 'LVGL-CORE' is better than 'LVGL-Essential'?
> 
> When I first created the LVGL cmsis-pack, I choose this name. I know it is no better than core, but the same there is no special reason that core is better than essential.
> 
> Suppose you have contributed some code to an opensource project and someone later rename it without practice reason, will you be comfortable?
> 
@GorgonMeducer I'm sorry for that, I'll of course revert that if you're feeling uncomfortable.
I forgot to rise this change so we can discussed it together...

I renamed that into CORE because it is including the **src/core** files. But you're right, it is still including more files from **src/darw**, **src/font**, **src/hal** and **src/misc**
I'm wondering if we need to split that too into sub-groups so we can disable fonts for example...

> @kisvegabor sorry for disturbing, do we have any suggestions about similar situations when doing contribution?

bbkarim:
I'm closing this.

Everything is now implemented is separated PR as needed.

Below PR are covering the same here but after first review done here:

- #3923 
- #3924 
- #3925 
- #3926 
- #3927 
- #3928 
- #3929 

Review will continue inside these PR.

PR #3923, opened 2023-01-13
fix(scripts): in lv_conf_internal preserve newline value for updated …

bbkarim:

1. file type changed from `scripts/lv_conf_internal_gen.py: Python script, ASCII text executable` to `scripts/lv_conf_internal_gen.py: Python script, ASCII text executable, with CRLF line terminators`
2. Preserve the file type for the generated lv_conf_internal.h file


kisvegabor:
Thanks!

PR #3924, opened 2023-01-13
fix(hal): remove extra include of "../draw/sdl/lv_draw_sdl.h"

bbkarim:
* remove extra include of "../draw/sdl/lv_draw_sdl.h"
* remove compilation warnings


kisvegabor:
Thank you!

PR #3925, opened 2023-01-13
feat(gpu): add support for STM32U5xx devices

bbkarim:

* added STM32U5 compilation flag
* restrict Cache maintenance code to Cortex M7 devices


GorgonMeducer:
We can update current cmsis-pack pdsc for just adding this one(update the LVGL::lvgl just like other GPU supports) so the January monthly update can have this support.

kisvegabor:
@bbkarim I don't have a STM32U5xx dev board so I can't try it out. Please, confirm that you have tested that the GPU really work with STM32U5xx.

PR #3926, opened 2023-01-13
arch(lvgl): include only needed header files

bbkarim:

* include only needed header files as defined in lv_conf.h


kisvegabor:
For v9 (what we are developing now) I've removed the concept of "Extra Widgets". Now all widgets are in a simple list.
Could you do it this way in Kconfig file too?

PR #3946, opened 2023-01-26
fix(themes): show chart indicator in monochrome theme

boraozgen:

The chart indicator (point) was not displayed in the monochrome theme. This PR fixes it by setting the indicator style in the same manner as the default theme.

Mind that in order to backport this to v8, the `lv_style_set_size` call must be changed accordingly.

Checkpoints done.


kisvegabor:
Thank you!

PR #3965, opened 2023-02-02
fix(flush,tick): use atomics instead of volatile for synchronization

Sh3Rm4n:

As of the documentation[^1] `lv_tick_inc()` and `lv_disp_flush_ready()` should be thread-safe.

These functions alter global LVGL state to transfer information to the LVGL thread. Because of mis-optimizations these variables where previously already declared as `volatile`.

`volatile` however **is not enough**[^2] to ensure thread-safety and can still lead to subtle mis-optimizations.

To ensure correct synchronization, leverage the builtin, portable atomic implementation C11 provides.

To be backwards compatible, an opt in option `LV_USE_ATOMICS` was introduced, which is also only active, when the compiler supports it.

The atomics types replacing the volatile types are drop-in replacements with the same memory layout (e.g. `volatile int` => `atomic_int`).

Builtin increment and decrement and compound assignments on atomics are `memory_order_seq_cst`, so do provide even stricter ordering guarantees than volatile, while also being truly thread-safe for every CPU architecture supported by the compiler. [^3] This also reduces changes in the lines of code, because no `atomic_fetch_...` function is needed [^4]

This patch also removes volatile declarations of some bit-fields, because these fields are not shared between threads, used for the `lv_disp_flush_ready()` function. I assume these fields to be declared as volatile for historical reasons.

[^1]: https://docs.lvgl.io/8.3/porting/os.html?highlight=lv_tick_in#interrupts
[^2]: https://stackoverflow.com/questions/4557979/when-to-use-volatile-with-multi-threading/4558031#4558031
[^3]: https://en.cppreference.com/w/c/language/atomic
[^4]: https://en.cppreference.com/w/cpp/header/stdatomic.h

### Checkpoints

<details><summary>checkpoints</summary>

- [x] ~~Add [Examples](https://github.com/lvgl/lvgl/tree/master/examples) if relevant.~~
- [x] ~~Add [Tests](https://github.com/lvgl/lvgl/blob/master/tests/README.md) if applicable.~~
 
  </details>

### DIFF
--- a/src/hal/lv_hal_tick.c
+++ b/src/hal/lv_hal_tick.c
@@ -7,6 +7,7 @@
  *      INCLUDES
  *********************/
 #include "lv_hal_tick.h"
+#include "../misc/lv_types.h"
 #include <stddef.h>
 
 #if LV_TICK_CUSTOM == 1
@@ -29,8 +30,13 @@
  *  STATIC VARIABLES
  **********************/
 #if !LV_TICK_CUSTOM
-    static uint32_t sys_time = 0;
-    static volatile uint8_t tick_irq_flag;
+    #if LV_USE_ATOMICS == 1
+        static _Atomic(uint32_t) sys_time = 0;
+        static _Atomic(uint8_t) tick_irq_flag;
+    #else
+        static uint32_t sys_time = 0;
+        static volatile uint8_t tick_irq_flag;
+    #endif
 #endif
 
 /**********************
--- a/src/hal/lv_hal_tick.c
+++ b/src/hal/lv_hal_tick.c
@@ -8,6 +8,7 @@
  *********************/
 #include "lv_hal_tick.h"
 #include "../misc/lv_types.h"
+#include <stdatomic.h>
 #include <stddef.h>
 
 #if LV_TICK_CUSTOM == 1
@@ -32,10 +33,10 @@
 #if !LV_TICK_CUSTOM
     #if LV_USE_ATOMICS == 1
         static _Atomic(uint32_t) sys_time = 0;
-        static _Atomic(uint8_t) tick_irq_flag;
+        static atomic_int tick_irq_flag;
     #else
-        static uint32_t sys_time = 0;
-        static volatile uint8_t tick_irq_flag;
+        static volatile uint32_t sys_time = 0;
+        static volatile int tick_irq_flag;
     #endif
 #endif
 
--- a/src/hal/lv_hal_disp.h
+++ b/src/hal/lv_hal_disp.h
@@ -23,6 +23,11 @@ extern "C" {
 #include "../misc/lv_area.h"
 #include "../misc/lv_ll.h"
 #include "../misc/lv_timer.h"
+#include "../misc/lv_types.h"
+
+#if LV_USE_ATOMICS == 1
+#include <stdatomic.h>
+#endif
 
 /*********************
  *      DEFINES
@@ -47,6 +52,11 @@ struct _lv_disp_t;
 struct _lv_disp_drv_t;
 struct _lv_theme_t;
 
+#if LV_USE_ATOMICS == 1
+#define FLUSHING_TYPE atomic_int
+#else
+#define FLUSHING_TYPE volatile int
+#endif
 /**
  * Structure for holding display buffer information.
  */
@@ -57,13 +67,13 @@ typedef struct _lv_disp_draw_buf_t {
     /*Internal, used by the library*/
     void * buf_act;
     uint32_t size; /*In pixel count*/
-    /*1: flushing is in progress. (It can't be a bit field because when it's cleared from IRQ Read-Modify-Write issue might occur)*/
-    volatile int flushing;
-    /*1: It was the last chunk to flush. (It can't be a bit field because when it's cleared from IRQ Read-Modify-Write issue might occur)*/
-    volatile int flushing_last;
-    volatile uint32_t last_area         : 1; /*1: the last area is being rendered*/
-    volatile uint32_t last_part         : 1; /*1: the last part of the current area is being rendered*/
+    FLUSHING_TYPE flushing;
+    /*It was the last chunk to flush. (It can't be a bit field because when it's cleared from IRQ Read-Modify-Write issue might occur)*/
+    FLUSHING_TYPE flushing_last;
+    uint32_t last_area         : 1; /*1: the last area is being rendered*/
+    uint32_t last_part         : 1; /*1: the last part of the current area is being rendered*/
 } lv_disp_draw_buf_t;
+#undef FLUSHING_TYPE
 
 typedef enum {
     LV_DISP_ROT_NONE = 0,
--- a/src/misc/lv_types.h
+++ b/src/misc/lv_types.h
@@ -32,6 +32,15 @@ extern "C" {
 
 #endif
 
+/*Use atomics instead of volatile variables for state which is potentially shared between threads, as long as
+ *the compiler supports it.*/
+#if (defined(__cplusplus) && __cplusplus >= 201103L)\
+    || (__STDC_VERSION__ >= 201112L && !defined(__STDC_NO_ATOMICS__))
+#define LV_USE_ATOMICS 1
+#else
+#define LV_USE_ATOMICS 0
+#endif
+
 /**********************
  *      TYPEDEFS
  **********************/


kisvegabor:
Hi,

Thank you for the detailed explanation. I wasn't aware of the `atomic` features but it looks really good. 

I just wonder if there is a way to enable atomic automatically to avoid the use of `LV_USE_ATOMIC` in `lv_conf.h`.

E.g. add this in `lv_type.h`:
```c
#ifndef LV_USE_ATOMIC /*Just to have a way to force it if for some reason the below mechanism doesn't work.*/
  #if  (defined(__cplusplus) && __cplusplus >= 201103L) \
        || (__STDC_VERSION__ >= 201112L && !defined(__STDC_NO_ATOMICS__))
    #define LV_USE_ATOMICS 1
    #include <stdatomic.h>
  #else
    #define LV_USE_ATOMICS 0
  #endif
#endif /*#ifndef LV_USE_ATOMIC*/
```

Sh3Rm4n:
Yeah, that should be possible. I was very conservative with my approach, trying not to change anything by default, but if you prefer the automatic switch depending on the compiler support, I'm happy to change it. :slightly_smiling_face: 

Do you have in mind to totally remove the `LV_USE_ATOMICS` option from `lv_conf` and let that define only live in `lv_type.h`?

kisvegabor:
> I was very conservative with my approach, trying not to change anything by default, but if you prefer the automatic switch depending on the compiler support, I'm happy to change it.

I'm afraid if we don't make it automatic 99% of people will never use it. :slightly_smiling_face: 

> Do you have in mind to totally remove the LV_USE_ATOMICS option from lv_conf and let that define only live in lv_type.h?

Yes, I think we should remove `LV_USE_ATOMICS` from `lv_conf.h`, but we can define it in `lv_types.h`. This way any part of the code can use it later.

Sh3Rm4n:
I've moved the define to `lv_types.h` as you suggested and also changed the atomic types a little, to ensure that the memory layout is compatible to the volatile version.

Sh3Rm4n:
Alright, so pycparser does currently not support `stdatomics.h` and fails parsing it. So I've updated the dependencies: 

* https://github.com/lvgl/lv_binding_micropython/pull/255
* https://github.com/lvgl/lv_micropython/pull/62

src/hal/lv_hal_tick.c
```diff
@@ -29,8 +30,13 @@
  *  STATIC VARIABLES
  **********************/
 #if !LV_TICK_CUSTOM
-    static uint32_t sys_time = 0;
-    static volatile uint8_t tick_irq_flag;
+    #if LV_USE_ATOMICS == 1
+        static _Atomic(uint32_t) sys_time = 0;
```
kisvegabor, line 35: I wasn't atomic previously. Was it changed intentionally?

src/hal/lv_hal_tick.c
```diff
@@ -29,8 +30,13 @@
...
+        static _Atomic(uint8_t) tick_irq_flag;
```
kisvegabor: I'd use `atomic_int` and `volatile int` just to use the same pattern as with `flushing` and `flushing_last`.

src/hal/lv_hal_tick.c
Sh3Rm4n: Yes, because `lv_tick_inc`, which should be thread-safe, has a read-write access to the global `sys_time` variable. 
Maybe it should be declared volatile as well for the non-atomic fallback?

src/hal/lv_hal_tick.c
Sh3Rm4n: Changed that :)

src/hal/lv_hal_tick.c
kisvegabor: There is a trick to make `lv_tick_inc()` and `sys_time` "thread-safe". See [here](https://github.com/lvgl/lvgl/blob/d979967d38990ced3c4a4ac6d0e967a46909ba07/src/hal/lv_hal_tick.c#L64-L73). I think this trick is required because if atomic is not supported on a 16 bit system, an `uint32_t` variable acn be read/written only in 2 steps. So the `lv_tick_inc()` interrupt can come in the wrong time.

It's not the case with the flush related variables as they are `int` and we can assume that they are written by one instruction.

What do you think?

Sh3Rm4n: Yes, this might reduce the risk on 16-bit systems for a corrupt read. In case of  `stdatomic` support, we wouldn't need that work-around at all, so in that case, if `LV_USE_ATOMICS == 1` the function could just return `sys_time` in any case, even for 16-bit systems. 

But here are two misunderstandings. 1 assembly instruction does not necessarily mean atomic, it is important though that it "appears atomic" aka. the CPU is in the right mode. E.g. ARM is adding **additional** instructions in case of atomic operations: [[ldrex, strex]](https://developer.arm.com/documentation/dht0008/a/arm-synchronization-primitives/exclusive-accesses/ldrex-and-strex?lang=en)

And still `volatile` gives no synchronization guarantees between multiple threads, so it is technically not enough. E.g. on sophisticated systems with multiple CPUs `sys_time` or `tick_irq_flag` could be cached by the core cache, but the volatile access does not lead into synchronizing these caches. 

So it might work on systems with one CPU no caching and when `lv_tick_inc()` is called in an ISR. But with atomics, is safe to call in any circumstances. It would even be safe to call `lv_tick_inc()` from multiple threads or when `lv_tick_get()` could pre-empt `lv_tick_inc()` (assuming I'll remove the `tick_irq_flag` workaround in case of `LV_USE_ATOMICS == 1`).

Sh3Rm4n: Okay, so after some reading I'm not to sure, if  `_Atomic(uint32_t)` compiles on every 16-bit platform, as the CPU might not support atomics `> uintmax_t`. But this error might be only reported by the final linking script, when it tries to find a polyfill lock for those atomic? (I am really not sure)> 

So the safe bet is to revert the sys_time -> atomic change, and keep the `tick_irq_flag` while loop, but still make that flag atomic.

kisvegabor: > So the safe bet is to revert the sys_time -> atomic change, and keep the tick_irq_flag while loop, but still make that flag atomic

I think it's a good compromise. It will work on simple embedded system (1 CPU) and on larger system atomic will be supported.

kisvegabor:
Please confirm that, the current version is working as we have discussed and can be merged.

src/hal/lv_hal_tick.c
Sh3Rm4n: Okay, I dug into this a little further: 

So, the `LV_USE_ATOMICS` check ensures, that `stdatomic` and the `_Atomic` keyword exists. But this is only a check if the compiler supports atomic. 

Now it comes down to the architecture if it does support atomic instructions as well. E.g. ARM with the [[ldrex, strex]](https://developer.arm.com/documentation/dht0008/a/arm-synchronization-primitives/exclusive-accesses/ldrex-and-strex?lang=en) and memory barrier support. But even for these architectures it comes down to the size of the type to be atomic, which mostly depends on the address width of the architecture as well. 

This property can be checked by the [`atomic_is_lock_free`](https://en.cppreference.com/w/c/atomic/atomic_is_lock_free) check.

So for example for [`avr` (godbolt)](https://godbolt.org/#g:!((g:!((g:!((h:codeEditor,i:(filename:'1',fontScale:14,fontUsePx:'0',j:1,lang:___c,selection:(endColumn:7,endLineNumber:12,positionColumn:7,positionLineNumber:12,selectionStartColumn:7,selectionStartLineNumber:12,startColumn:7,startLineNumber:12),source:'%23include+%3Cstdint.h%3E%0A%23include+%3Cstdatomic.h%3E%0A%23include+%3Cstdio.h%3E%0A%0A%23if+HAVE_ATOMIC_FETCH_OP%0A%23error+bla%0A%23endif%0A%0A_Atomic(uint16_t)+num%3B%0A%0A/*+Type+your+code+here,+or+load+an+example.+*/%0Auint16_t+times2(void)+%7B%0A++++return+num+%2B%3D+num%3B%0A%7D%0A%0Aint+main()+%7B%0A++++printf(%22%25d%22,+atomic_is_lock_free(%26num))%3B%0A++++printf(%22%25lu%22,+times2())%3B%0A%7D%0A'),l:'5',n:'0',o:'C+source+%231',t:'0')),k:34.176891489942705,l:'4',m:100,n:'0',o:'',s:0,t:'0'),(g:!((h:compiler,i:(compiler:cavrg1220,deviceViewOpen:'1',filters:(b:'0',binary:'0',binaryObject:'1',commentOnly:'0',demangle:'0',directives:'0',execute:'0',intel:'1',libraryCode:'0',trim:'1'),flagsViewOpen:'1',fontScale:14,fontUsePx:'0',j:1,lang:___c,libs:!(),options:'-latomic+-O3',selection:(endColumn:1,endLineNumber:1,positionColumn:1,positionLineNumber:1,selectionStartColumn:1,selectionStartLineNumber:1,startColumn:1,startLineNumber:1),source:1),l:'5',n:'0',o:'+AVR+gcc+12.2.0+(Editor+%231)',t:'0')),header:(),k:35.922218641626074,l:'4',m:100,n:'0',o:'',s:0,t:'0'),(g:!((h:output,i:(compilerName:'x86-64+gcc+12.2',editorid:1,fontScale:14,fontUsePx:'0',j:1,wrap:'1'),l:'5',n:'0',o:'Output+of+AVR+gcc+12.2.0+(Compiler+%231)',t:'0')),k:29.900889868431236,l:'4',m:100,n:'0',o:'',s:0,t:'0')),l:'2',n:'0',o:'',t:'0')),version:4) there is no builtin atomic support so functions like `__atomic_fetch_add_2` are always called, independent on the size of the atomic variable. 

For [ARM (godbolt)](https://godbolt.org/#g:!((g:!((g:!((h:compiler,i:(compiler:cclang1500,deviceViewOpen:'1',filters:(b:'0',binary:'1',binaryObject:'1',commentOnly:'0',demangle:'0',directives:'0',execute:'1',intel:'0',libraryCode:'0',trim:'1'),flagsViewOpen:'1',fontScale:14,fontUsePx:'0',j:1,lang:___c,libs:!(),options:'--target%3Dthumbv7em-none-eabihf',selection:(endColumn:1,endLineNumber:1,positionColumn:1,positionLineNumber:1,selectionStartColumn:1,selectionStartLineNumber:1,startColumn:1,startLineNumber:1),source:1,wantOptInfo:'1'),l:'5',n:'0',o:'+x86-64+clang+15.0.0+(Editor+%231)',t:'0')),header:(),k:50,l:'4',m:100,n:'0',o:'',s:0,t:'0'),(g:!((h:codeEditor,i:(filename:'1',fontScale:14,fontUsePx:'0',j:1,lang:___c,selection:(endColumn:24,endLineNumber:14,positionColumn:24,positionLineNumber:14,selectionStartColumn:24,selectionStartLineNumber:14,startColumn:24,startLineNumber:14),source:'%23include+%3Cstdint.h%3E%0A%23include+%3Cstdatomic.h%3E%0A//+%23include+%3Cstdio.h%3E%0A%0A_Atomic(uint32_t)+num%3B%0A_Atomic(uint64_t)+num2%3B%0A%0A/*+Type+your+code+here,+or+load+an+example.+*/%0Auint32_t+times2(void)+%7B%0A++++return+num+%2B%3D+num%3B%0A%7D%0A%0Auint32_t+times2_64(void)+%7B%0A++++return+num2+%2B%3D+num2%3B%0A%7D%0A%0A'),l:'5',n:'0',o:'C+source+%231',t:'0')),header:(),k:50,l:'4',n:'0',o:'',s:0,t:'0')),l:'2',n:'0',o:'',t:'0')),version:4) however, it depends on the size of the variable. 

Anyway, in cases of calling the builtin `__atomic_*` functions it is not guaranteed to be lockless anymore, and `-latomic` aka. libatomic must be linked, so that the program can compile. AFAIK, avr provides no `libatomic` by default, and it must be user provided. Meaning, we can not assume, that this will compile for every architecture.

With that knowledge, I worry we can't just enable using atomics by default, as this might break user code, which worked upon that point, because they did not provide any `libatomic` or just did not add the `-latomic` linker flag. 

Ideally, such support should be tested by built tools like cmake before automatically enabling it, but this heavily on the user use-case. I'd propose to add the `LV_USE_ATOMICS` options back to `lv_conf.h`.
We can still default it to `1` aka enabled (if a breaking change is ok), but the user at least has a chance to disable it, if they did not implement libatomic.

Sh3Rm4n:
No, with the reasoning layed out in https://github.com/lvgl/lvgl/pull/3965#discussion_r1102516354, I'd like to propose to reintroduce the LV_USE_ATOMIC option inside the `lv_conf.h` again. 

Well technically for my use either version is working, but I'm only working on ARM 32-bit with atomic support. And CI also only checks for thse kinds of architectures.

kisvegabor:
Which AVR family does it mean? Note that the minimal requirement for LVGL is 16 bit MCUs. Anyway, I understand that AVR is just an example, and there can be issues with other vendors/tools too.

Let's summarize what we know so far:
- `volatile` might not be enough on advanced architectures with complex caching, multiple CPUs, etc
- `atomic` should solve this issue, but it might not be available or complicated to add/link in some cases
-  fortunately, in the cases where `atomic` is really needed it should be available too
- ideally `LV_USE_ATOMIC` should be enabled automatically when it's available AND needed.
- it's hard to enable it automatically so what can we do:
  - Disable it by default: we risk that it will cause hard to find issues for people  (although I don't remember issues like this)
  - Enable it by default: we risk having easy to understand compile issues.  

Now we are working on v9, where we can have API breaking changes. And the development will last long so I'm in the favor auto-detecting it automatically  as now, and let's see if people report issues. We can change it at any time during the development of v9.

What do you think?

Sh3Rm4n:
I think your comment summarizes it pretty accurately. :+1: 

Yes, AVR was just an example for an architecture, where 32-bit values are not atomic nor does the CPU provide atomic support. 

> Now we are working on v9, where we can have API breaking changes. And the development will last long so I'm in the favor auto-detecting it automatically as now, and let's see if people report issues. We can change it at any time during the development of v9.
> What do you think?

I like this idea. If v9 allows breaking changes, this might be a good opportunity to introduce atomic values. And I also think the risk of reports is pretty low, as 32-bit MCUs with atomic support will be the norm. Which means, that the `-latomic` library must not be linked, as long as the atomic values have an built-in instruction equivalent, which is true for now. 

So after I've tackled https://github.com/lvgl/lvgl/pull/3965#discussion_r1102505601 the PR can be used as is. :)

kisvegabor:
Okay, I'm merging it! I learned a lot from our talk. :slightly_smiling_face:

kisvegabor:
Please update update the Pull Request in the MicroPython repo.

PR #3980, opened 2023-02-10
fix(chart,ticks): change line chart ticks

CMoranMINTL:

Change amount of tick marks for the chart widget to be equal to amount of data points if chart is of type LV_CHART_TYPE_LINE. Reference discussion can be found in this forum [post](https://forum.lvgl.io/t/line-graph-x-axis-tick-labels/10821).

### Checkpoints
 
### DIFF
--- a/src/widgets/chart/lv_chart.c
+++ b/src/widgets/chart/lv_chart.c
@@ -1576,7 +1576,14 @@ static void draw_x_ticks(lv_obj_t * obj, lv_draw_ctx_t * draw_ctx, lv_chart_axis
     }
 
     p1.y = y_ofs;
-    uint32_t total_tick_num = (t->major_cnt - 1) * t->minor_cnt;
+    uint32_t total_tick_num;
+    if (chart->type == LV_CHART_TYPE_LINE && (((t->major_cnt - 1) * t->minor_cnt) == chart->point_cnt)) {
+        total_tick_num = chart->point_cnt - 1;
+    }
+    else {
+        total_tick_num = (t->major_cnt - 1) * t->minor_cnt;
+    }
+    if (!total_tick_num) return; /*Ensure total_tick_num is not equal to zero*/
     for(i = 0; i <= total_tick_num; i++) { /*one extra loop - it may not exist in the list, empty label*/
         bool major = false;
         if(i % t->minor_cnt == 0) major = true;
--- a/src/widgets/chart/lv_chart.c
+++ b/src/widgets/chart/lv_chart.c
@@ -1577,13 +1577,13 @@ static void draw_x_ticks(lv_obj_t * obj, lv_draw_ctx_t * draw_ctx, lv_chart_axis
 
     p1.y = y_ofs;
     uint32_t total_tick_num;
-    if (chart->type == LV_CHART_TYPE_LINE && (((t->major_cnt - 1) * t->minor_cnt) == chart->point_cnt)) {
+    if(chart->type == LV_CHART_TYPE_LINE && (((t->major_cnt - 1) * t->minor_cnt) == chart->point_cnt)) {
         total_tick_num = chart->point_cnt - 1;
     }
     else {
         total_tick_num = (t->major_cnt - 1) * t->minor_cnt;
     }
-    if (!total_tick_num) return; /*Ensure total_tick_num is not equal to zero*/
+    if(!total_tick_num) return; /*Ensure total_tick_num is not equal to zero*/
     for(i = 0; i <= total_tick_num; i++) { /*one extra loop - it may not exist in the list, empty label*/
         bool major = false;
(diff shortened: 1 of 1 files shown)


pete-pjb:
A note for @kisvegabor, hope you are well! Long time no speak. 😃  I have been working with @CMoranMINTL on this pull request and thought you might like to see if you want to back port it.

Thank you for your contribution @CMoranMINTL .

Kind Regards,

Pete

kisvegabor:
Hi @CMoranMINTL and @pete-pjb,

If I understood correctly, the problem is that the ticks always end on a major tick, but in some cases should end on minor tick. Is this correct?

In v9, I was planning to unify and reconsider how the "tick APIs".  Now we need to set the number of major ticks and the number of minors between them. This way it's not possible to create ticks like:
```
| , , , , | , , , , | , ,
```

I think a more flexible solution would be to specify the total number of ticks (12 in the above example) and a step count for majors (5 in the example). In the chart example from the forum it'd be `tick_count = 50`, `nth_major = 2`. 

What do you think? I know that it's a larger change though.

pete-pjb:
Hi @kisvegabor ,

I think that approach works well, it is probably simpler for the user too.

Kind Regards,

Pete

CMoranMINTL:
Hi @kisvegabor ,

You are understanding the problem correctly. I also agree the proposed v9 solution would be a more eloquent way to accomplish this.

kisvegabor:
>  I also agree the proposed v9 solution would be a more eloquent way to accomplish this.

Happy to hear that :slightly_smiling_face:

Are you interested in making some experiments with it?

CMoranMINTL:
> Are you interested in making some experiments with it?

I am interested for sure, but I'm currently only using LVGL for my day job and I'm unsure how much time I'd be able to dedicate to it. Perhaps, in the future, if I use LVGL for a personal project I may have more time for it, for now I will have to say no.

kisvegabor:
No problem, I perfectly understand. 

Anyway, I leave this issue open for while. In the hope that someone picks it up.

C47D:
Hi @kisvegabor , what would be a good api name for those operations?

> I think a more flexible solution would be to specify the total number of ticks (12 in the above example) and a step count for majors (5 in the example). In the chart example from the forum it'd be tick_count = 50, nth_major = 2.

C47D:
With less than 50 ticks it works as expected, e.g. 5
imagen

imagen
or 25
imagen

With 50 it does this:
imagen

Funny I can see the tick value string with the correct value:
imagen

The updates are in https://github.com/C47D/lvgl/tree/chart_ticks_test

C47D:
Fixed the issue by disabling the `lv_obj_add_event_cb(ui.chart, draw_tick_marks_cb, LV_EVENT_DRAW_PART_BEGIN, NULL);` code from the forum example code

imagen

C47D:
Here's the output with the following configuration code, the x axis is configured to set a major tick every 5 ticks.

```c
static void update_data_points(lv_event_t* e)
{
    lv_obj_t* obj = lv_event_get_target(e);

    if (code != LV_EVENT_VALUE_CHANGED)
    {
        return;
    }

    char dropdown_str[4] = { '\0' };
    lv_coord_t major_count = 0;
    lv_coord_t minor_count = 0;

    lv_dropdown_get_selected_str(obj, dropdown_str, 4);
    data_point_count = (uint16_t) strtod(dropdown_str, NULL);

    LV_LOG_USER("Data point count %d", data_point_count);

    lv_chart_set_point_count(ui.chart, data_point_count);
    lv_chart_set_axis_tick_count(ui.chart, LV_CHART_AXIS_PRIMARY_X, data_point_count);
    lv_chart_set_axis_tick_nth_major(ui.chart, LV_CHART_AXIS_PRIMARY_X, 5);
}
```

imagen

imagen

kisvegabor:
Hi Carlos,

I think ultimately we should go for a `scale` widget as I described [here](https://github.com/lvgl/lvgl/issues/4043#issuecomment-1464386802). It could be widely used for the existing widgets and for any custom purpose. What do you think about it?

C47D:
So instead of using ticks and labels in the chart we should use a scale instead? It would be really useful for re-usability, but I'm worried about time, I can work here in small batches of time.

I would do it like this, let me know if you agree:
- Land this update once is ready
- Create scale widget (this two can be done in parallel)
- Incrementally use it in other widgets

kisvegabor:
> Land this update once is ready

I find the currently proposed solution (in the PR now) not general enough, but if it will be replaced by the scale widget in v9, I'm ok with it as a temporal solution.

C47D:
Well, I think the original issue was solved with this PR (without my changes) and a custom draw callback. So maybe we can merge it and later use the new widget.

kisvegabor:
Okay, merging. Thanks!

PR #3995, opened 2023-02-16
Protect redefining LV_USE_ATOMICS

### LINKED ISSUE #3994: Atomic not supported on Raspberry PI (GCC 10.2.1 | aarch64)
### Introduce the problem
```
[build] /usr/lib/gcc/aarch64-linux-gnu/10/include/stdatomic.h:40:9: error: ‘_Atomic’ does not name a type`
[build]    40 | typedef _Atomic _Bool atomic_bool;
```

### System Info:
OS = Debian GNU/Linux 11 (bullseye)
gcc (Debian 10.2.1-6) 10.2.1 20210110

### Suggested solution
https://github.com/lvgl/lvgl/pull/3995

Chuvi-w:

Fix https://github.com/lvgl/lvgl/issues/3994

### DIFF
--- a/src/misc/lv_types.h
+++ b/src/misc/lv_types.h
@@ -34,11 +34,13 @@ extern "C" {
 
 /*Use atomics instead of volatile variables for state which is potentially shared between threads, as long as
  *the compiler supports it.*/
+#ifndef LV_USE_ATOMICS
 #if (defined(__cplusplus) && __cplusplus >= 201103L)\
     || (__STDC_VERSION__ >= 201112L && !defined(__STDC_NO_ATOMICS__))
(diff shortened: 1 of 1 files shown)


kisvegabor:
@Sh3Rm4n we were talking about that it should be added it seems both of missed it in the end. 

So do you also agree to merge it?

Sh3Rm4n:
I don't see this as a solution other than having a way to overwrite the LV_USE_ATOMICS flag. But I think this is better solved with a proper config option instead. 

But this PR and the corresponding issue reveals another problem with atomic and the atomic support, see https://github.com/lvgl/lvgl/issues/3994#issuecomment-1434162987.

PR #4023, opened 2023-02-27
test(main.py): add cli option to select test suite

### LINKED ISSUE #4019: [Unit tests] Running a single test file
### Introduce the problem
Lately when running the test suite the last test execution takes a long time and I had to Ctrl-C it.

imagen

You can see the 40th test is missing.

At work (and for personal projects) I'm used to run only the test file I'm working on, and do nightly runs of all the test files (like LVGL currently do), so I wondered how feasible it would be to run only one test file. I tried to write a Makefile to run only the lv_chart test runner but I found out it has a lot of dependencies on other source files so I gave up.

### Examples and cases
While working in a test file we should not be worried about breaking other test files.

### Suggested solution
Try to find out if it's possible to update, compile and run a single test runner of a given configuration and add an option to the main.py to run a selected test runner, this way hand made Makefiles would not be needed.

This can lead to running test files in less than a second, like this one that took 80ms
imagen

some00:
Fixes: #4019

### Checkpoints
 
### DIFF
--- a/tests/main.py
+++ b/tests/main.py
@@ -90,7 +90,7 @@ def build_tests(options_name, build_type, clean):
                            '--parallel', str(os.cpu_count())])
 
 
-def run_tests(options_name):
+def run_tests(options_name, test_suite):
     '''Run the tests for the given options name.'''
 
     print()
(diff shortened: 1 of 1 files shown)


kisvegabor:
It was fast! :slightly_smiling_face: Thank you!

@C47D could you take a look at it?

some00:
@C47D the new flag is `--test-suite` and it handles regex filter. For example:
```
./main.py test --test-suite test_msg
```
will only run one test suite.

C47D:
> It was fast! :slightly_smiling_face: Thank you!
> 
> @C47D could you take a look at it?

Sure, I'll take a look at it later today! 

**EDIT**
It's much faster, thanks @some00 . Do you think it's possible to only build the selected test suite? If not or it would require more work then this is great as it is.
imagen

some00:
@C47D It would make main.py too complex and limit how the tests can added. I would need to get a list of all the tests from ctest and assume that test names are also executable names.

C47D:
@some00 Then this is great as it is! It already made testing the current widget I'm working on a lot faster.

kisvegabor:
Amazing! I merge it.

Thank you @some00 for adding it so quickly!

PR #4042, opened 2023-03-07
scroll slider into view on focus

kutu:
Currently most focusable widgets are scrolling into view by default, but slider is not.


kisvegabor:
Makes sense, thank you.

The failing CI is unrelated.

PR #4078, opened 2023-03-21
Tests: malloc/realloc/free stub in unit tests

C47D:

Control `lv_malloc` behaviour (return value) inside unit tests, the stubbed malloc is not enabled by default, so it doesn't break the current behaviour. Naming is not proper, just a dirty implementation

I've tested it with the following test file:

```c
#if LV_BUILD_TEST
#include "../lvgl.h"

#include "lv_test_malloc.h"
#include "unity/unity.h"

static lv_obj_t * active_screen = NULL;
static lv_obj_t * meter;

void setUp(void)
{
    active_screen = lv_scr_act();
    meter = lv_meter_create(active_screen);
}

void tearDown(void)
{
    active_screen = NULL;
    FAKE_STOP_RETURN; // <- Disables malloc stub, this can be done here or at the end of each test FAKE_MALLOC_RETURN was used
}

void test_meter_add_needle_line_returns_null_on_oom(void)
{
    FAKE_MALLOC_RETURN(NULL); // <- Enables malloc stub and sets the return value of malloc

    TEST_ASSERT_NULL(lv_meter_add_needle_line(meter, 1, lv_palette_main(LV_PALETTE_RED), 1));
}

void test_meter_on_successfull_allocation(void)
{
    lv_meter_indicator_t * indicator = lv_meter_add_needle_line(meter, 1, lv_palette_main(LV_PALETTE_RED), 2);
    TEST_ASSERT_NOT_NULL(indicator);

    TEST_ASSERT_EQUAL(LV_METER_INDICATOR_TYPE_NEEDLE_LINE, indicator->type);
    TEST_ASSERT_EQUAL(1, indicator->type_data.needle_line.width);
    TEST_ASSERT_EQUAL(2, indicator->type_data.needle_line.r_mod);
}

#endif
```

And the output:
imagen

### Checkpoints
 
### DIFF
--- a/tests/src/lv_test_conf.h
+++ b/tests/src/lv_test_conf.h
@@ -23,7 +23,7 @@ extern "C" {
 #define LV_USE_BUILTIN_SNPRINTF 1
 #define LV_STDLIB_INCLUDE <stdlib.h>
 #define LV_MALLOC       lv_test_malloc
-#define LV_REALLOC      realloc
+#define LV_REALLOC      lv_test_realloc
 #define LV_FREE         free
 #define LV_MEMSET       memset
 #define LV_MEMCPY       memcpy
@@ -34,7 +34,7 @@ extern "C" {
 #define LV_USE_BUILTIN_MEMCPY   1
 #define LV_USE_BUILTIN_SNPRINTF 1
 #define LV_MALLOC       lv_test_malloc
-#define LV_REALLOC      lv_realloc_builtin
+#define LV_REALLOC      lv_test_realloc
 #define LV_FREE         lv_free_builtin
 #define LV_MEMSET       lv_memset_builtin
 #define LV_MEMCPY       lv_memcpy_builtin
--- a/tests/src/lv_test_conf.h
+++ b/tests/src/lv_test_conf.h
@@ -15,13 +15,14 @@ extern "C" {
 /***********************
  * PLATFORM CONFIGS
  ***********************/
+#include "lv_test_malloc.h"
 
 #ifdef LVGL_CI_USING_SYS_HEAP
 #define LV_USE_BUILTIN_MALLOC   0
 #define LV_USE_BUILTIN_MEMCPY   1
 #define LV_USE_BUILTIN_SNPRINTF 1
 #define LV_STDLIB_INCLUDE <stdlib.h>
-#define LV_MALLOC       malloc
+#define LV_MALLOC       lv_malloc_test_wrapper
 #define LV_REALLOC      realloc
 #define LV_FREE         free
 #define LV_MEMSET       memset
@@ -32,7 +33,7 @@ extern "C" {
 #define LV_USE_BUILTIN_MALLOC   1
 #define LV_USE_BUILTIN_MEMCPY   1
 #define LV_USE_BUILTIN_SNPRINTF 1
-#define LV_MALLOC       lv_malloc_builtin
+#define LV_MALLOC       lv_malloc_test_wrapper
 #define LV_REALLOC      lv_realloc_builtin
 #define LV_FREE         lv_free_builtin
 #define LV_MEMSET       lv_memset_builtin
--- a/tests/src/lv_test_conf.h
+++ b/tests/src/lv_test_conf.h
@@ -22,7 +22,7 @@ extern "C" {
 #define LV_USE_BUILTIN_MEMCPY   1
 #define LV_USE_BUILTIN_SNPRINTF 1
 #define LV_STDLIB_INCLUDE <stdlib.h>
-#define LV_MALLOC       lv_malloc_test_wrapper
+#define LV_MALLOC       lv_test_malloc
 #define LV_REALLOC      realloc
 #define LV_FREE         free
 #define LV_MEMSET       memset
@@ -33,7 +33,7 @@ extern "C" {
 #define LV_USE_BUILTIN_MALLOC   1
 #define LV_USE_BUILTIN_MEMCPY   1
 #define LV_USE_BUILTIN_SNPRINTF 1
-#define LV_MALLOC       lv_malloc_test_wrapper
+#define LV_MALLOC       lv_test_malloc
 #define LV_REALLOC      lv_realloc_builtin
 #define LV_FREE         lv_free_builtin
 #define LV_MEMSET       lv_memset_builtin
--- a/tests/src/lv_test_conf.h
+++ b/tests/src/lv_test_conf.h
@@ -24,7 +24,7 @@ extern "C" {
 #define LV_STDLIB_INCLUDE <stdlib.h>
 #define LV_MALLOC       lv_test_malloc
 #define LV_REALLOC      lv_test_realloc
-#define LV_FREE         free
+#define LV_FREE         lv_test_free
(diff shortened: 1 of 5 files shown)


C47D:
I need to figure out how to handle the malloc option on the failing CI checks tho

tests/src/lv_test_conf.h
```diff
@@ -32,7 +33,7 @@ extern "C" {
 #define LV_USE_BUILTIN_MALLOC   1
 #define LV_USE_BUILTIN_MEMCPY   1
 #define LV_USE_BUILTIN_SNPRINTF 1
-#define LV_MALLOC       lv_malloc_builtin
+#define LV_MALLOC       lv_malloc_test_wrapper
 #define LV_REALLOC      lv_realloc_builtin
```
kisvegabor: I'm not sure, but maybe we need this for realloc too. Or should realloc  return `lv_control_lv_malloc_stub` too?

kisvegabor (commented):
I think allowing to set a `malloc_cb` would be more flexible.

E.g. with the current technique we can test if there are 2 mallocs in a function call. We can make both of them fail (i.e. it will return after the first) or both of them pass.

However with a callback we can do this:
```c
void * my_cb(size_t s)
{
    if(cnt == 0) {
        cnt++;
        return lv_test_malloc_normal(s);
    } else {
        return NULL;
    }
}
...


lv_test_malloc_set_cb(my_cb);
cnt = 0;
lv_some_func_to_test();

lv_test_malloc_set_cb(NULL); //Normal operation
...
```

In `lv_test_malloc.c`
```c
void * lv_test_malloc_normal(size_t s)
{
#ifdef LVGL_CI_USING_SYS_HEAP
        return malloc(size);
#else // LVGL_CI_USING_DEF_HEAP and others
        return lv_malloc_builtin(size);
#endif
}

void * lv_test_malloc(size_t s)
{
    if(malloc_stub_cb) return malloc_stub_cb(s);
    else return lv_test_malloc_normal(s);
}
```

What do you think?

C47D:
> I think allowing to set a malloc_cb would be more flexible.
E.g. with the current technique we can test if there are 2 mallocs in a function call. We can make both of them fail (i.e. it will return after the first) or both of them pass.
However with a callback we can do this:

Yup, I like it better. The implementation is updated.

tests/src/lv_test_conf.h
C47D: Not really sure, but it's easy to add, let me do that

C47D: Done in [7cb4173](https://github.com/lvgl/lvgl/pull/4078/commits/7cb4173a5994c523908db70e3c982b02d934e4b4)

C47D:
I'm not sure how the checkbox should behave if mem allocation fails, should it keep the old text? It currently sets the txt to NULL.

I tried to use the realloc stub but with this test I'm getting a memleak report. If `p` is freed in `realloc_stub` then the memleak is no longer present and the assert passes.

```c
void * realloc_stub(void * p, size_t new_size)
{
    (void) new_size;
    (void) p; // Replace with free if you want to avoid a memleak :)

    return NULL;
}

void test_checkbox_should_return_dinamically_allocated_text_realloc_fails(void)
{
    const char * message = "Hello World!";
    const char * message2 = "LVGL Rocks";

    active_screen = lv_scr_act();
    checkbox = lv_checkbox_create(active_screen);

    lv_test_realloc_set_cb(realloc_stub);

    lv_checkbox_set_text(checkbox, message);
    lv_checkbox_set_text(checkbox, message2); // realloc failure

    // Text is NULL because of free in stub
    TEST_ASSERT_NULL(lv_checkbox_get_text(checkbox));

    lv_test_realloc_set_cb(NULL);
}

void * malloc_stub(size_t size)
{
    (void) size;

    return NULL;
}

void test_checkbox_has_previous_text_when_malloc_fails(void)
{
    const char * message = "Hello World!";

    active_screen = lv_scr_act();
    checkbox = lv_checkbox_create(active_screen);

    TEST_ASSERT_EQUAL_STRING("Check box", lv_checkbox_get_text(checkbox));

    lv_test_malloc_set_cb(malloc_stub);

    lv_checkbox_set_text(checkbox, message);

    // Previous text is default string
    // TEST_ASSERT_EQUAL_STRING("Check box", lv_checkbox_get_text(checkbox));

    lv_test_malloc_set_cb(NULL);
}
```

`lv_checkbox_set_text` needs to handle mem allocation failure so I added an early return
```c
{
    size_t len = _lv_txt_ap_calc_bytes_cnt(txt);
#else
    size_t len = strlen(txt);
#endif

    if(!cb->static_txt) cb->txt = lv_realloc(cb->txt, len + 1);
    else  cb->txt = lv_malloc(len + 1);

    if (NULL == cb->txt) return; // <- mem allocation failure check

#else
    strcpy(cb->txt, txt);
#endif

    cb->static_txt = 0;

    lv_obj_invalidate(obj);
}
```

tests/src/lv_test_conf.h
C47D: Maybe we also need to stub free?

tests/src/lv_test_conf.h
```diff
@@ -15,14 +15,15 @@ extern "C" {
...
 #ifdef LVGL_CI_USING_SYS_HEAP
 #define LV_USE_BUILTIN_MALLOC   0
 #define LV_USE_BUILTIN_MEMCPY   1
 #define LV_USE_BUILTIN_SNPRINTF 1
 #define LV_STDLIB_INCLUDE <stdlib.h>
-#define LV_MALLOC       malloc
-#define LV_REALLOC      realloc
+#define LV_MALLOC       lv_test_malloc
+#define LV_REALLOC      lv_test_realloc
 #define LV_FREE         free
```
kisvegabor: Yes, we should have `lv_test_free` too. Who know what magic will `test_malloc` and `test_realloc` will do that should not be directly fed into `free`.

kisvegabor (commented):
I really like it!

kisvegabor:
> lv_checkbox_set_text needs to handle mem allocation failure so I added an early return

To be honest I wasn't thinking much about what should happen if the UI is out of memory. It turned out that it's a controversy opinion, but I think if the UI is out of memory there isn't much you can do. We had long discussions about it on various places so wouldn't like to go into detail now. Anyway an early return is really something that we can do. So 
Maybe in case of the checkbox setting  a `" "` as static text could be an option too.

tests/src/lv_test_conf.h
C47D: Done in 5789edb

C47D:
> > lv_checkbox_set_text needs to handle mem allocation failure so I added an early return
> 
> To be honest I wasn't thinking much about what should happen if the UI is out of memory. It turned out that it's a controversy opinion, but I think if the UI is out of memory there isn't much you can do. We had long discussions about it on various places so wouldn't like to go into detail now. Anyway an early return is really something that we can do. So Maybe in case of the checkbox setting a `" "` as static text could be an option too.

Maybe we can approach that in another PR

kisvegabor:
LGTM! I merge it.

Thank you, Carlos!

PR #4089, opened 2023-03-25
checkbox: Check for mem allocation failure in set_text

C47D:

Handle memory allocation failure in `lv_checkbox_set_text` and add test.

### Checkpoints
 
### DIFF
--- a/src/widgets/checkbox/lv_checkbox.c
+++ b/src/widgets/checkbox/lv_checkbox.c
@@ -68,25 +68,30 @@ lv_obj_t * lv_checkbox_create(lv_obj_t * parent)
 void lv_checkbox_set_text(lv_obj_t * obj, const char * txt)
 {
     lv_checkbox_t * cb = (lv_checkbox_t *)obj;
-#if LV_USE_ARABIC_PERSIAN_CHARS
-    size_t len = _lv_txt_ap_calc_bytes_cnt(txt);
-#else
-    size_t len = strlen(txt);
-#endif
 
-    if(!cb->static_txt) cb->txt = lv_realloc(cb->txt, len + 1);
-    else  cb->txt = lv_malloc(len + 1);
-
-    LV_ASSERT_MALLOC(cb->txt);
-    if(NULL == cb->txt) return;
+    if(NULL != txt) {
+        size_t len = 0;
 
 #if LV_USE_ARABIC_PERSIAN_CHARS
-    _lv_txt_ap_proc(txt, cb->txt);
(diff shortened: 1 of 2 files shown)


C47D:
@kisvegabor is it safe to pass `NULL` to `strcpy` as `src` parameter? I've take a look at the [c reference](https://en.cppreference.com/w/c/string/byte/strcpy) and I think it isn't. For `strlen` when the `str` parameter is [NULL the return value is 0](https://en.cppreference.com/w/c/string/byte/strlen).

I think it would be safer like this:
```c
{

    if (NULL != txt)
    {
        size_t len = 0;

        len = _lv_txt_ap_calc_bytes_cnt(txt);
#else
        len = strlen(txt);
#endif

        if(!cb->static_txt) cb->txt = lv_realloc(cb->txt, len + 1);
        else cb->txt = lv_malloc(len + 1);

        LV_ASSERT_MALLOC(cb->txt);
        if(NULL == cb->txt) return;

#else
        strcpy(cb->txt, txt);
#endif

        cb->static_txt = 0;
    }

    lv_obj_invalidate(obj);
}
```

kisvegabor:
> is it safe to pass NULL to strcpy as src parameter?

I don't think it's safe either.

C47D:
I can include the above code snippet in this request, if you think it is ok.

kisvegabor:
Yes, please. But please use `if (txt == NULL) return;` to have less nesting.

BTW, sometimes I see people using `if(NULL != txt)` instead of `if (txt != NULL)`. Do you know what is the advantage of having the constant expression first?

pete-pjb:
Hi @kisvegabor ,

See [this](https://stackoverflow.com/questions/370366/why-put-the-constant-before-the-variable-in-a-comparison)


cc @C47D

kisvegabor:
I see, thank you Pete! :slightly_smiling_face:

C47D:
Yup, the link @pete-pjb provided is the reason I tent to use NULL (or any other constant for that matter) at the left side of the comparison.

@kisvegabor I've updated the behavior of the function and also added a test for it. There's another manual work I did to count the number of times malloc was called, which could be done automagically by having a mocking capability in the unit tests.

C47D:
> Yes, please. But please use if (txt == NULL) return; to have less nesting.

I think that's not possible because according to the docs when passing NULL the text should be refreshed, so it can't return early.

kisvegabor:
You are right, Let's no with your solution.  I merge it. 

Do we need mocking on other places too?

C47D:
Maybe not, let's find out writing more tests :)

PR #4116, opened 2023-04-05
feat(keyboard): add functions for adding/removing default event

### LINKED ISSUE #4112: Removing lv_keyboard_def_event_cb used to be easier before v9
### Introduce the problem

In v8 the default event callback of the keyboard widget could be removed with

```
lv_obj_remove_event_cb(keyboard, lv_keyboard_def_event_cb);
```

On `master` this now seems to require looping over all events to find the correct index for `lv_obj_remove_event`.

```
uint32_t num_keyboard_events = lv_obj_get_event_count(keyboard);
for (uint32_t i = 0; i < num_keyboard_events; ++i) {
    if (lv_event_dsc_get_cb(lv_obj_get_event_dsc(keyboard, i)) == lv_keyboard_def_event_cb) {
        lv_obj_remove_event(keyboard, i);
        break;
    }
}
```

### Examples and cases

This occurs in any application that wants to replace the default callback with a custom one, e.g. [unl0kr](https://gitlab.com/cherrypicker/unl0kr) or [buffyboard](https://gitlab.com/cherrypicker/buffyboard).

### Suggested solution

Maybe a convenience method could be added to remove an event callback using its function pointer?

Johennes:

This adds function for safely adding and removing the default event on keyboard widgets.

Fixes: #4112

### Checkpoints
 
### DIFF
--- a/src/widgets/keyboard/lv_keyboard.c
+++ b/src/widgets/keyboard/lv_keyboard.c
@@ -309,6 +309,40 @@ bool lv_btnmatrix_get_popovers(const lv_obj_t * obj)
  *====================*/
 
 /**
+ * Add the default keyboard event if it wasn't added previously.
+ * @param obj pointer to a keyboard object
+ */
+void lv_keyboard_add_default_event(lv_obj_t * obj)
+{
+    lv_keyboard_t * keyboard = (lv_keyboard_t *)obj;
+    if(keyboard->def_event_dsc) {
+        return;
+    }
+    lv_obj_add_event(obj, lv_keyboard_def_event_cb, LV_EVENT_VALUE_CHANGED, NULL);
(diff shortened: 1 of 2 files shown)


kisvegabor:
Saving the `event_dsc` is not a good idea. [Here](https://github.com/lvgl/lvgl/blob/master/src/misc/lv_event.c#L91) you an see that when a new event is added, the list of `event_dsc`s is reallocated. So the saved `event_dsc`s will be invalid. Actually this is one of the main  sources of all these extra complications around deleting events. 

So I suggest having only `lv_keyboard_remove_default_event` and let the user add it's event with `lv_obj_add_event` directly.

A minor note: please don't add [function comments](https://github.com/lvgl/lvgl/pull/4116/files#diff-9e71900b20013cab12ffa173382c71f288ea475c1c1726f1de6d1dbbfce7abeaR311-R314) in the C file.

PR #4121, opened 2023-04-07
feat(refr): implement the monitor component

FASTSHIFT:

- [x] Implemented a common monitor widget.
- [x] Rearranged the creation process and global variables of monitor in `lv_refr`.
- [x] The new frame rate statistics method displays the frame counter and rendering time separately (because in the case of using the VSYNC signal to trigger the rendering, it is not accurate to use the rendering time to calculate the frame rate).
- [x] Support to use LOG to print the frame rate (because rendering the monitor component also takes extra time).

### Checkpoints
 

kisvegabor:
I like the idea in general but I think it'd make more sense to add into the `others` folder,  and also move the `perf` and `mem` callbacks there to make the `lv_refr.c` cleaner.

It's slightly related, but I think showing the last WARNING and ERROR log messages would be also useful. Maybe we could use the same generic `monitor` to to display the last message. The problem is that it can't show the asserts, as LVGL will stop there. Maybe switching to a character mode in case of an assert ? I'm not sure... Anyway, it's an other topic. :slightly_smiling_face:

FASTSHIFT:
@kisvegabor 
All features have been completed. 
I implemented a frame rate printing function(enable the `LV_USE_PERF_MONITOR_LOG_MDOE`), what do you think?

kisvegabor:
It looks good! I merge it, thank you!

kisvegabor:
Just one thing: we should find a different name for this component. In a UI library the word "monitor" can be easily misunderstood and it conflicts with a former `LV_USE_MONITOR` (SDL window)  component from `lv_drivers`.

To be honest I don't have any good name ideas. Do you have any?

FASTSHIFT:
These are some of the names ChatGPT suggested:

1. PerformanceTracker
2. MetricsCollector
3. ResourceMonitor
4. EfficiencyObserver
5. Speedometer
6. HealthCheck
7. OptimizationInspector
8. CapacityWatcher
9. UtilizationAnalyzer
10. WorkloadProfiler

What about `lv_metrics`?

PR #4128, opened 2023-04-11
Add indev - set_feedback_cb() function

PGNetHun:

Since the display & input driver architecture change the `feedback_cb` field is private (in [lv_indev_private.h](https://github.com/lvgl/lvgl/blob/master/src/core/lv_indev_private.h#L34)), and it's not possible to define a callback function to listen to input device events. 

But this feature (to listen to input device events) is useful for example to have an auto-sleep feature: the callback function stores the time of last input event (for example touch), and a separate auto-sleep thread periodically checks this stored time, and after 10 minutes of inactivity (when there was no user input), the device/display is switched to sleep mode.

Other use case (as documentation mentions) is for example playing a sound for every input event / touch: https://docs.lvgl.io/master/porting/indev.html#feedback

### Checkpoints
- [X] Run `code-format.py` from the scripts folder. [astyle](http://astyle.sourceforge.net/install.html) needs to be installed.
- [X] Update the [Documentation](https://github.com/lvgl/lvgl/tree/master/docs) if needed
 
- [X] Follow the [Styling guide](https://github.com/lvgl/lvgl/blob/master/docs/CODING_STYLE.md)
- [X] Arguments must be named in H files too.

### DIFF
--- a/docs/porting/indev.md
+++ b/docs/porting/indev.md
@@ -170,9 +170,12 @@ The default value of the following parameters can be changed in `lv_indev_t`:
 
 ### Feedback
 
-Besides `read_cb` a `feedback_cb` callback can be also specified in `lv_indev_t`.
-`feedback_cb` is called when any type of event is sent by the input devices (independently of its type). This allows generating feedback for the user, e.g. to play a sound on `LV_EVENT_CLICKED`.
+Besides `read_cb` a `feedback_cb` callback can be also specified.
 
+`feedback_cb` is called when any type of event is sent by the input devices (independently of its type). This allows generating feedback for the user, e.g. to play a sound on `LV_EVENT_CLICKED`.
+```c
+lv_indev_set_feedback_cb(indev, feedback_cb);
+```
 
 ### Associating with a display
 Every input device is associated with a display. By default, a new input device is added to the last display created or explicitly selected (using `lv_disp_set_default()`).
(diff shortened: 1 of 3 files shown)


FASTSHIFT:
I think it is more appropriate to add a `LV_EVENT_FEEDBACK`, because it can support more event callback processing instead of only one. What do you think?

kisvegabor:
> But this feature (to listen to input device events) is useful for example to have an auto-sleep feature: the callback function stores the time of last input event (for example touch), and a separate auto-sleep thread periodically checks this stored time, and after 10 minutes of inactivity (when there was no user input), the device/display is switched to sleep mode.

For this purpose you can use `lv_disp_get_inactive_time(disp)` too. 

> I think it is more appropriate to add a LV_EVENT_FEEDBACK, because it can support more event callback processing instead of only one. What do you think?

Shall we send this event to the display and use the triggering `indev` as a parameter in the event?

PGNetHun:
> > But this feature (to listen to input device events) is useful for example to have an auto-sleep feature: the callback function stores the time of last input event (for example touch), and a separate auto-sleep thread periodically checks this stored time, and after 10 minutes of inactivity (when there was no user input), the device/display is switched to sleep mode.
> 
> For this purpose you can use `lv_disp_get_inactive_time(disp)` too.

Thanks, I did not notice that.

Should I close this PR, or do you want to do it?

kisvegabor:
I think we should reconsider the `feedback_cb` for v9, and it will be a longer discussion starting from understanding the real problems and goals.

So let's close it for now.

PR #4152, opened 2023-04-20
feat(object_tree): make lv_obj_get_index return error value if the object has no parent

sparkles43:

If lv_obj_get_index is called by passing an object that has no parent, it returns 0. This indicates that the object's non-existing parent has a child which is the passed object. I find this very confusing and have changed it to return the error value 0xFFFFFFFF which is also used if the passed object is not among the parent's children.

I haven't yet checked where this function is used, but I think we need to do something. If this code change is raising other issues, we might just improve the documentation of lv_obj_get_index  in the header by saying that ONLY "an object with a parent" or "an object that is a child for sure" may be passed.

Currently the documentation says "Get the index of a child" which implies that only children will work but I think it should be more striking.

### Checkpoints
- ~~Add [Examples](https://github.com/lvgl/lvgl/tree/master/examples) if relevant.~~
- ~~Add [Tests](https://github.com/lvgl/lvgl/blob/master/tests/README.md) if applicable.~~
- ~~If you added new options to `lv_conf_template.h` run [lv_conf_internal_gen.py](https://github.com/lvgl/lvgl/blob/release/v8.3/scripts/lv_conf_internal_gen.py) and update [Kconfig](https://github.com/lvgl/lvgl/blob/release/v8.3/Kconfig).~~
 
### DIFF
--- a/src/core/lv_obj_tree.c
+++ b/src/core/lv_obj_tree.c
@@ -336,7 +336,7 @@ uint32_t lv_obj_get_index(const lv_obj_t * obj)
     LV_ASSERT_OBJ(obj, MY_CLASS);
 
     lv_obj_t * parent = lv_obj_get_parent(obj);
-    if(parent == NULL) return 0;
+    if(parent == NULL) return 0xFFFFFFFF;
 
     uint32_t i = 0;
     for(i = 0; i < lv_obj_get_child_cnt(parent); i++) {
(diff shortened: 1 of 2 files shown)


kisvegabor:
It makes sense. This function shouldn't be used on screens.

kisvegabor:
The title says it's "Draft". So isn't it ready to merge?

sparkles43:
I just felt a bit unsure whether it might break some uses of the function. It is called at a few occasions and even used in the layers documentation.

kisvegabor:
Getting the child index of a screen is corner case and a little bit non-sense too. So wouldn't worry too much about breaking codes. 

So I merge it, thank you.

PR #4156, opened 2023-04-22
Fix SDL window size not changing on some platforms

embeddedt:

We had an issue with not being able to change the display size on the Micropython JS port anymore in v9: https://github.com/lvgl/lv_binding_micropython/issues/264#issuecomment-1513794078

For some reason Emscripten doesn't seem to accept the first window size and requires setting it again.

This fix was suggested by @kdschlosser and appears to fix the issue.


kdschlosser:
thanks for fixing this.

kisvegabor:
Thanks!

PR #4159, opened 2023-04-25
Update lv_gpu_stm32_dma2d.c

xuxeu:
If the hardware enables cache, it needs to be invalidated, otherwise it will cause a splash screen


kisvegabor:
Thank you! Please fix the formatting issue as [shown here](https://github.com/lvgl/lvgl/actions/runs/4793172029/jobs/8525381897?pr=4159).

kisvegabor:
Why was it closed?

PR #4161, opened 2023-04-25
Update lv_gpu_stm32_dma2d.c

xuxeu:
If the hardware enables cache, it needs to be invalidated, otherwise it will cause a splash screen


kisvegabor:
Thank you!

PR #4175, opened 2023-04-27
removes non ascii characters

kdschlosser:

removes non ascii characters from the documentation. Not all non ascii characters just things like the non ascii double quotes and single quotes and also the "NBSP" or non ascii space

### DIFF
--- a/docs/CHANGELOG.rst
+++ b/docs/CHANGELOG.rst
@@ -48,7 +48,7 @@ Changelog
 Performance
 ~~~~~~~~~~~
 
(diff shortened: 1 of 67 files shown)


kisvegabor:
Wow, it's interesting that we have so many `“` and `’`. I don't think they are used intentionally. Probably it's a character encoding issue. Sometimes I see here copy-pasted code where strings look like `“hello“`, which clearly wouldn't work.

Anyway, thank you!

kdschlosser:
it happened when I did the conversion from the markdown to the reST.

PR #4177, opened 2023-04-27
removes recommonmark from doc build

kdschlosser:
as the title says, removes recommonmark from the documentation build

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -26,7 +26,7 @@ sys.path.insert(0, os.path.abspath('./_ext'))
 # from subprocess import PIPE, Popen
 
 # import recommonmark
-from recommonmark.transform import AutoStructify
+# from recommonmark.transform import AutoStructify
 from sphinx.builders.html import StandaloneHTMLBuilder
 
 # -- General configuration ------------------------------------------------
@@ -272,10 +272,10 @@ _, repo_commit_hash = subprocess.getstatusoutput("git rev-parse HEAD")
 # Example configuration for intersphinx: refer to the Python standard library.
 
 def setup(app):
-    app.add_config_value('recommonmark_config', {
-            'enable_eval_rst': True,
-            'enable_auto_toc_tree': 'True',
-            }, True)
-    app.add_transform(AutoStructify)
+    # app.add_config_value('recommonmark_config', {
+    #         'enable_eval_rst': True,
+    #         'enable_auto_toc_tree': 'True',
+    #         }, True)
+    # app.add_transform(AutoStructify)
     app.add_css_file('css/custom.css')
     app.add_css_file('css/fontawesome.min.css')


kisvegabor:
We are getting closer:
```
...
sphinx-sitemap: sitemap.xml was generated for URL https://docs.lvgl.io/master/en/html/ in /home/runner/work/lvgl/lvgl/out_html/sitemap.xml
Sphinx parallel build error:
KeyError: 'anchorname'


****************
Building
****************
Generate the list of examples
Running doxygen

cd "/tmp/tmptz51k5h9.lvgl_docs" && doxygen Doxyfile
-------------------------------------
Reading Doxygen output

sphinx-build -b latex "/tmp/tmptz51k5h9.lvgl_docs" "/tmp/tmptz51k5h9.lvgl_docs/out_latex" -j 2
-------------------------------------

cd "/tmp/tmptz51k5h9.lvgl_docs/out_latex" && latexmk -pdf "LVGL.tex"
-------------------------------------

sphinx-build -b html "/tmp/tmptz51k5h9.lvgl_docs" "/home/runner/work/lvgl/lvgl/out_html" -D version="9.0.0-dev" -E -j 2
-------------------------------------
Exit build due to previous error
Error: Process completed with exit code 255.
```

And there many many warnings similar to:
```
Missing character: There is no 可 (U+53EF) in font [FreeSerif.otf]/OT:script=lat
n;language=dflt;!
```
and
```
Missing character: There is no ב (U+05D1) in font DejaVu Sans Mono/OT:script=la
tn;language=dflt;!
```

I think these were there in the MD version too.


Besides 
```
/tmp/tmptz51k5h9.lvgl_docs/index.rst:17: WARNING: toctree contains reference to document 'ROADMAP' that doesn't have a title: no link will be generated
```

kdschlosser:
The warnings are coming from  files like file_explorer.rst.

There is HTML code being injected using the document files that contain characters that are outside the range the given font supports.

```

.. raw:: html

   <details>

.. raw:: html

   <summary>

中文

.. raw:: html

   </summary>

.. raw:: html

   <p>

``lv_file_explorer``
提供API让我们可以浏览文件系统中的内容。\ ``lv_file_explorer``
只提供了文件浏览功能，并不提供实际的文件操作功能，也就是说，不能像PC那样点击一个图片文件就可以打开查看该图片。\ ``lv_file_explorer``
会告诉您当前点击的文件的完整路径和名称，文件操作功能需要用户自己实现。

``lv_file_explorer`` 中的文件列表基于 `lv_table </widgets/table>`__
实现，快速访问栏基于 `lv_list </widgets/list>`__
实现。因此，要注意确保使能了 ``lv_table`` 和 ``lv_list``\ 。

.. raw:: html

   </p>

.. raw:: html

   </details>
```

I am not sure why the documentation contains multiple languages. If the documentation is to be available in more than one language then there needs to be a different set of files for each of the languages that are available. This way a specific font that provides all of the characters for that language can be used.

To my knowledge there are no fonts that support every possible language. I think that some of the characters possible overlay each other and that is why it doesn't exist.


Now the actual error that is causing the build failure I am thinking is this.

```
KeyError: 'anchorname'
```

But that gives me no insight into where the actual problem is. You are going to have to give me some time to see if I can replicate the issue and to add some code that will give me more insight to where the problem is occurring.

The question is this. seeing as how a user is able to build the documentation and with how the internet is pretty much everywhere as well is there really a need to generate a PDF version of the documentation? You may have some kind of an agreement with a vendor where they want the documentation in PDF format. I could understand the necessity to have it if that is the case. If not then why are the docs being compiled into a PDF at all? They can be compiled and viewed locally as HTML or they can be viewed from the website..

Another solution is to use an HTML to PDF library instead of using sphinx to build the docs. a combination of pdfkit for Python and wkhtmltopdf for Linux and PDF's can be generated form the HTML that is output by sphinx.

IDK if that will solve the issue with the unicode characters, it might. I am sure it would solve the keyerror issue because that issue doesn't exist nor does the warnings of the documentation is generated without the PDF.

so for now if you want to generate only the HTML docs add skip_latex to the build.py command

```
python build.py skip_latex
```

that will bypass the PDF generation.

kdschlosser:
the roadmap warning I created a PR for. #4182 
I am not 100% done with the conversion from MD to reST. I thought I had hit all of the titles. I could have made a boo boo when fixing the merge conflicts and the contents of that file ended up getting swapped out for the contents of an MD file. I know I wasn't getting any warnings when I was compiling the docs (skipping the PDF build)

Not a big deal to with that file as it was only the headers that had to be corrected.

kisvegabor:
I was also surprised that the PDF version is needed by people, but it is. Some people wants to print it, other prefer searching in it. When it wasn't available temporarily earlier people started reporting it very quickly, so I guess it's being used. So we should add it back at some point. 

Anyway, I've added `skip_latex` in https://github.com/lvgl/lvgl/commit/a5d7d207239780dffd1152b14fea0a9f38ce57b4 for now.

For mixing other languages, seemingly there is no simple way to define a fallback front in Latex. Probably we can remove the Chinese part from the docs, but occasionally we can have non-ASCII parts in the examples.

kdschlosser:
what I am thinking is if I don't use latex and instead use  pdfkit for Python and wkhtmltopdf for Linux to generate the PDF I am hoping that the unicode characters will come out like they should. I am going to give that a try and see how it goes.

Another option is to give a link to the documentation presented all on a single page. Then the user can simply "print" the page to PDF from their browser. I think the way I mentioned above is going to do the trick. unicode can be a royal pain sometimes. I am going to check and see if there is a way to define a fallback font to use for latex.

kisvegabor:
Okay, let's ty it later.

kdschlosser:
ok so the code and simulator buttons not working in the docs. I have tracked down the issue to this

"git rev-parse HEAD" This gets run in a subprocess in the config file. I have to change the directory it is being run in. the problem stems from moving the compilation to a temporary directory. Not a big deal to fix and I should have it done by later today. I am going to look into the PDF issues today as well.

kdschlosser:
#4213 corrects some of the problems in the documentation. I still have to address the PDF problem. But the major ones like the example buttons and the changelog should be fixed.

kisvegabor:
I've tested #4213 and it works well! :rocket: 

The PDF is really not a high priority.

PR #4182, opened 2023-04-28
finishes conversion of ROADMAP.rst file

kdschlosser:
as the title says..

### DIFF
--- a/docs/ROADMAP.rst
+++ b/docs/ROADMAP.rst
@@ -1,8 +1,11 @@
-# Roadmap
+Roadmap
+=======
 
-## Planned for v9
+Planned for v9
+--------------
 
-### Naming and API
+Naming and API
+~~~~~~~~~~~~~~
 - [x] `lv_style_set_size()` should have separate width and height parameters
 - [x] Reconsider image color formats.
 - [ ] More consistent names:`remove/clear/delete/del`, `offset/ofs`, `add/create/register`, `id/idx/index`, `middle/mid/center`, `img/image`, `txt/text`, `opa/opacity/alpha`, `scr/screen`, `disp, display`, `finished/complete/completed/ready`, `buf/buffer`, `..._cb`, angle, rotation`, `zoom, scale`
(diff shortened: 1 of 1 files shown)


kisvegabor:
Thank you! :pray:

kisvegabor:
And the docs is built now! Amazing!

I've found only one issue: 
The [live examples](https://docs.lvgl.io/master/examples.html) are not working. Probably @embeddedt can help with it.

kdschlosser:
This is the url for the locations of the images.

https://docs.lvgl.io/9.0.0-dev/_static/built_lv_examples/index.html?example=lv_example_get_started_1&w=320&h=240

which is a problem. It should be
https://docs.lvgl.io/master/_static/built_lv_examples/index.html?example=lv_example_get_started_1&w=320&h=240

There is nothing in the actual documentation that does this. I am able to change the path to be master but that is going to cause problems in later versions of LVGL.

kdschlosser:
the documentation path should not be master, it should be the 9.0.0-dev

embeddedt:
I think the issue here is we have some conflicting logic for choosing the version.

* When a commit is pushed to a branch, the [GitHub action](https://github.com/lvgl/lvgl/blob/2549f23773daf5a1d1a2c4295f1fd6a50edae24f/.github/workflows/compile_docs.yml#L67-L93) deploys to the version-specific name given by `scripts/find_version.sh`. That script ignores the patch version, as the assumption is it don't matter for documentation. It also ignores the `dev` qualifier for simplicity, as if a minor version is unreleased, it's fine to keep deploying breaking documentation on top of the same URL. In the current scenario this deploys to the `9.0` folder.
* Additionally, pushes to the `master` branch will deploy to the `master` folder, regardless of version. (This is so that the absolute latest documentation can always be found at the same URL.) Note that the copy of the documentation in this folder still references the version name above for the live examples, which works fine since they are both pushed at the same time.
* The new `build.py` logic now [generates a more advanced version specifier](https://github.com/lvgl/lvgl/blob/2549f23773daf5a1d1a2c4295f1fd6a50edae24f/docs/build.py#L201-L221) that includes both the patch release and dev suffix, and passes this to Sphinx as the version to use. This causes the [example link generator](https://github.com/lvgl/lvgl/blob/2549f23773daf5a1d1a2c4295f1fd6a50edae24f/docs/_ext/lv_example.py#L90) to generate links that reference `9.0.0-dev`, rather than `9.0` as was previously done. This is what causes the issue as no docs are deployed to the `9.0.0-dev` folder.

I believe the issue can be solved simply by adjusting `build.py` to only consider major and minor version.

kdschlosser:
OK so what you need to have done is only the major.minor version number set for the documentation. 


I can do that easily.

kdschlosser:
#4189 fixes the issue

PR #4190, opened 2023-04-29
(fix) lv_spinbox doubling characters entered from the keyboard

strijar:
Duplicate characters occurred when entering them from the keyboard or sending them through events.


kisvegabor:
Thank you!

PR #4196, opened 2023-05-01
Adds lv_scale widget

C47D:

Adds lv_scale widget

### Checkpoints
 

C47D:
**NOTE**
I'm using `lv_port_win_codeblocks` for development, the lvgl submodule is a bit behind master over there.

kisvegabor:
Thank you for opening this PR. Let's specify the features of the widget. This is my idea:
- vertical left/right, horizontal top/bottom, round inner/outer forms (e.g. vertical left means that the major ticks are larger to the left and the labels are placed to the left )
- min, max values
- tick settings:
  - minor tick count
  - major count (how many minor to skip?)
  - minor and major tick style
  - tick recoloring? maybe it's enough to use custom events
- tick label settings
  - label to major 
  - use label texts from an array or automatically based on some formatting policies
  - rotate labels by a degree, or to the current degree in case of round scale (will be easy to do when `feat/parallel-rendering` is merged)

In `feat/parallel-rendering` I've already removed the `LV_EVENT_DRAW_PART_BEGIN/END` events and replaced them with something much simpler.  So these are not required.

What we need to do is figuring out a nice API and use the `lv_draw_...` functions to draw the ticks and lines.

What do you think?

src/widgets/scale/lv_scale.c
```diff
@@ -101,20 +139,106 @@ static void lv_scale_event(const lv_obj_class_t * class_p, lv_event_t * event)
...
+        part_draw_dsc.p2 = &point2;
+
+        part_draw_dsc.label_dsc = &label_dsc;
+        char text_buffer[20];
+        lv_snprintf(text_buffer, sizeof(text_buffer), "Hi %" LV_PRId32, tick_idx);
+        part_draw_dsc.label_dsc = &label_dsc;
+        part_draw_dsc.text = text_buffer;
+        part_draw_dsc.text_length = 10;
+
+        lv_event_send(obj, LV_EVENT_DRAW_PART_BEGIN, &part_draw_dsc);
```
kisvegabor: `LV_EVENT_DRAW_PART_*`  doesn't need to be added as they are already removed in `feat/parallel-rendering`.

src/widgets/scale/lv_scale.h
```diff
@@ -23,16 +23,51 @@ extern "C" {
...
+
 /**********************
  *      TYPEDEFS
  **********************/
 
+/**
+ * Scale mode
+ */
+enum {
+    LV_SCALE_MODE_HORIZONTAL    = 0x00,
```
kisvegabor: Maybe we can specify it as `LV_SCALE_MODE_HORIZONTAL_TOP` and `LV_SCALE_MODE_HORIZONTAL_BOTTOM` to specify in which direction the labels should be added.

src/widgets/scale/lv_scale.h
```diff
@@ -23,16 +23,51 @@ extern "C" {
...
 
 extern const lv_obj_class_t lv_scale_class;
 
+/**
+ * `type` field in `lv_obj_draw_part_dsc_t` if `class_p = lv_scale_class`
+ * Used in `LV_EVENT_DRAW_PART_BEGIN` and `LV_EVENT_DRAW_PART_END`
+ */
+typedef enum {
+    LV_SCALE_DRAW_PART_TICK_LABEL,       /**< Used on tick lines and labels*/
+} lv_scale_draw_part_type_t;
```
kisvegabor: It's not required either.

src/widgets/scale/lv_scale.c
C47D: I'll add a TODO comment to remove them, if I remove it now I won't be able to continue as I'm using a bit old commit of lvgl.

src/widgets/scale/lv_scale.h
C47D: Will do

src/widgets/scale/lv_scale.h
C47D: Will add a TODO comment to remove it

kisvegabor:
Hi Carlos, 

I've tried out the scale widget and added a skeleton for some examples.

First, I wanted to try out a simple horizontal scale (see `lv_example_scale_1.c`) but it seems its width is not picked correctly.

C47D:
Hi,

I've noticed there were some properties not set in the constructor (see 8728041) and an Off-by-one bug in the main line drawing (see dead5d8b).
I'm testing with this example:
```c
  /*Initialize LVGL*/
   lv_init();
  hal_init(800, 480);

  lv_obj_t * scale = lv_scale_create(lv_scr_act());
  lv_obj_set_size(scale, lv_disp_get_hor_res(NULL) / 2, lv_disp_get_ver_res(NULL) / 2);

  static lv_style_t indicator_style;
  lv_style_init(&indicator_style);
  /* Label style properties */
  lv_style_set_text_font(&indicator_style, &lv_font_montserrat_14);
  lv_style_set_text_color(&indicator_style, lv_color_hex(0xff00ff));
  /* Major tick properties */
  lv_style_set_line_color(&indicator_style, lv_color_hex(0x00ff00));
  lv_style_set_width(&indicator_style, 10U); // Tick length
  lv_style_set_line_width(&indicator_style, 2U); // Tick width
  lv_obj_add_style(scale, &indicator_style, LV_PART_INDICATOR);

  static lv_style_t minor_ticks_style;
  lv_style_init(&minor_ticks_style);
  lv_style_set_line_color(&minor_ticks_style, lv_color_hex(0xff0000));
  lv_style_set_width(&minor_ticks_style, 5U); // Tick length
  lv_style_set_line_width(&minor_ticks_style, 2U); // Tick width
  lv_obj_add_style(scale, &minor_ticks_style, LV_PART_ITEMS);

  static lv_style_t main_line_style;
  lv_style_init(&main_line_style);
  /* Main line properties */
  lv_style_set_line_color(&main_line_style, lv_color_hex(0x0000ff));
  lv_style_set_line_width(&main_line_style, 2U); // Tick width
  /* Used on round mode */
  lv_style_set_arc_color(&main_line_style, lv_color_hex(0x0000ff));
  lv_style_set_arc_width(&main_line_style, 2U);
  lv_obj_add_style(scale, &main_line_style, LV_PART_MAIN);

  lv_obj_center(scale);
```
And I get this result:
image

Now the style configuration for all three parts (main, items and indicators) is mandatory, which I think is not user friendly as you mentioned. Is there a way to get the application style and apply it to the scale parts so setting the styles is not needed?

C47D:
This is the output by just creating the scale with this code:

```c
lv_obj_t * scale = lv_scale_create(lv_scr_act());
lv_obj_set_size(scale, lv_disp_get_hor_res(NULL) / 2, lv_disp_get_ver_res(NULL) / 2);
lv_obj_center(scale);
```

image

src/widgets/scale/lv_scale.c
```diff
@@ -365,6 +369,10 @@ static void scale_draw_indicator(lv_obj_t * obj, lv_event_t * event)
     lv_draw_line_dsc_t line_dsc;
     lv_draw_line_dsc_init(&line_dsc);
     lv_obj_init_draw_line_dsc(obj, LV_PART_INDICATOR, &line_dsc);
+    /* Set tick line width to 2u when is not set */
+    if(0U == line_dsc.width) {
```
kisvegabor: I think it's not a good idea because there can be use cases when the main line is hidden.

Instead we should set the default line width from the theme. (I pushed a fix for it)

kisvegabor:
> With https://github.com/lvgl/lvgl/commit/b118e9c2c66d92706385f6763bec6a3f19833b70 the scale looks like this

With this code I think the scale should be centered both horizontally and vertically.

src/widgets/scale/lv_scale.c
C47D: Is this fix in master? I don't see it.

C47D:
> > With [b118e9c](https://github.com/lvgl/lvgl/commit/b118e9c2c66d92706385f6763bec6a3f19833b70) the scale looks like this
> 
> With this code I think the scale should be centered both horizontally and vertically.

Working in the fix, I was setting the scale based on the meter widget, which for the horizontal and vertical modes doesn't apply because they're lines, not an arc. I think this looks much better.

image

kisvegabor:
Looks great! I'm just thinking about how to align e.g. a horizontal bottom a scale below a bar. How to set its height properly?  

Maybe we can set the height to `LV_SIZE_CONTENT` and let the scale to set its height automatically to involve all the lines, ticks and labels.

The other (and simpler) option would be to align the scale to the edge where there are no labels. E.g. in case of horizontal bottom scale align the whole scale to the top. In case of vertical left, align to the left. 

I think to 2nd option would be much simpler and more intuitive too. What do you think?

src/widgets/scale/lv_scale.c
kisvegabor: Now I' can see it either :smile: It seems I haven't pushed it or messed up something. Anyway, you can just add it in `lv_theme_deafult.c` in `theme_apply` similarly to the other widgets.

C47D:
Hi

> Looks great! I'm just thinking about how to align e.g. a horizontal bottom a scale below a bar. How to set its height properly?

For this I was thinking on using the styles.

> The other (and simpler) option would be to align the scale to the edge where there are no labels. E.g. in case of horizontal bottom scale align the whole scale to the top. In case of vertical left, align to the left.

Do you have an sketch on this use? I kind of can't follow the idea sorry :smiley_cat:

kisvegabor:
> Do you have an sketch on this use? I kind of can't follow the idea sorry

Sure, but while I was playing with it I've noticed something. With this code:

```c
    lv_obj_t * scale = lv_scale_create(lv_scr_act());
    lv_obj_set_size(scale, lv_disp_get_hor_res(NULL) / 2, lv_disp_get_ver_res(NULL) / 2);
    lv_scale_set_label_show(scale, true);
    lv_scale_set_total_tick_count(scale, 10);
    lv_scale_set_major_tick_every(scale, 5);
    lv_obj_center(scale);

    lv_obj_set_style_bg_color(scale, lv_color_hex3(0xf00), 0);
    lv_obj_set_style_bg_opa(scale, LV_OPA_20, 0);
```
I get:
image

But after
```c
  //lv_obj_center(scale);
```

image


I found that, it's because `y_ofs` is not calculated correctly here: 
```c
        if(LV_SCALE_MODE_HORIZONTAL_BOTTOM == scale->mode) {
            x_ofs = obj->coords.x1 + pad_left - scroll_left;
            y_ofs = obj->coords.y2 - obj->coords.y1;
        }
```

I think you meant:
```c
y_ofs = obj->coords.y1 + (obj->coords.y2 - obj->coords.y1) / 2;
```

But what I recommend in case of `LV_SCALE_MODE_HORIZONTAL_BOTTOM` is:
```c
y_ofs = obj->coords.y1; //Update in all draw functions
```
Which results in:

image

This way we always know that the `LV_SCALE_MODE_HORIZONTAL_BOTTOM` is aligned to the top of the widget's bounding box, and it would be easy to align the scale precisely below a bar or a chart.

Note that we still need to add `main_line_width / 2` to the `y_ofs` to be sure the half line width won't be clipped.

C47D:
> Note that we still need to add main_line_width / 2 to the y_ofs to be sure the half line width won't be clipped.

Thanks for finding the issue, is this `main_line_width / 2` required on all modes, or only on horizontal bottom?

src/widgets/scale/lv_scale.c
C47D: Done in 4f000ae35 and 890c6c92f and this is the result, does it look ok?

image

kisvegabor:
On all modes. Let's say the we have a horizontal bottom scale on x1:100, y1:20, x2:300, y2:50. If you draw a line with width:4 px at y:20 the line will be drawn between y:19 and y:22. 

The same applies to all cases when the line is drawn on the very edge of the widget.

C47D:
> On all modes. Let's say the we have a horizontal bottom scale on x1:100, y1:20, x2:300, y2:50. If you draw a line with width:4 px at y:20 the line will be drawn between y:19 and y:22.
> 
> The same applies to all cases when the line is drawn on the very edge of the widget.

I've commited the diff above in e225d88

For horizontal mode this is the new positioning, see update in 6af3d7d
image

kisvegabor:
Thank you.

I've added a few examples but kept your example in `lv_example_scale_5()`. I've add some `//TODO` comments in the examples when I fond some issues. Please check them out. 

I've also fixed a few minor thing.

I also noticed this issue:
image

Here the main line should be aligned to the end of the tick.

C47D:
> 

Thanks, I'll check it later today, I was on holiday last weekend.

C47D:
> Thank you.
> 
> I've added a few examples but kept your example in `lv_example_scale_5()`. I've add some `//TODO` comments in the examples when I fond some issues. Please check them out.
> 
> I've also fixed a few minor thing.
> 
> I also noticed this issue: image
> 
> Here the main line should be aligned to the end of the tick.

Thanks for noticing it, does it look better now?
image

C47D:
Fixed comment _TODO min value should be on the bottom_ from example 2 with commit f3337bd
image
image

kisvegabor:
Hi Carlos,

Thank you!

image

Here the main line position should consider the section's tick width.

kisvegabor:
Thank you Carlos.

I've checked the example and `lv_example_scale_1` looks like this on a 800x480 screen:
image

Maybe some rounding error?

`lv_example_scale_4` has some issues too:
image


Could you take a look at these?

C47D:
> Thank you Carlos.
> 
> I've checked the example and `lv_example_scale_1` looks like this on a 800x480 screen: image
> 
> Maybe some rounding error?
> 
> `lv_example_scale_4` has some issues too: image
> 
> Could you take a look at these?

Hi, I've found a couple of missing updates in the horizontal tick calculation that fixes this, I'll commit the fix once I clean it up. Same applies for the minor ticks.

C47D:
Here are the updated examples, they look much better now (in my opinion)

---
Example 1
image

---
Example 2
image

---
Example 3
image

---
Example 4
image

---
Example 5

This one in particular has a return statement after creating the scale, so I'm not sure if this is as expected.

image

---
TODO

- [ ] Adjust main line in arc modes to cover custom tick widths
- [ ] Add setters for arc mode, missing setter for angle range and rotation
- [ ] Feedback from Gabor

C47D:
Hi @kisvegabor , the arc issues are in the TODO list I did in my previous comment, already working on it. The one I might need some help is adjusting the section main line to the width of the ticks, I don't know how to adjust the angles calculation taking the tick widths into consideration, do you have some pointers/suggestions to solve that?

kisvegabor:
It's really not obvious and I'm sure there will be some ugly rounding errors in the end. 

For the sake of simplicity what if make the tick start from the opposite side of the arc? Like this:
image

C47D:
I've worked in the horizontal and vertical modes, here are the examples

Example 2 Horizontal mode
image

---
Example 2 Vertical mode
image

C47D:
This is the render of example 3 with the proposed adjustment:

Round inner mode: Ticks 10 and 40 are a bit far from arc
image

Round outter mode
image

---
This is example 4
Here I still need to adjust the section arc to the minor tick width. I still don't imagine how to compensate for that.

image

kisvegabor:
Looks great! How far is it from merging?

C47D:
I only need to replace some magic numbers with defines, than it will be ready to be merged.

kisvegabor:
Great! Once it's there we can iron out the smaller issue in small patches.

C47D:
@kisvegabor Ready to be merged

kisvegabor:
Cool, thank you so much! I'll try to integrate it with the width demo!

PR #4236, opened 2023-05-20
feat(timer): implementation of adaptive call timer handler cycle method

FASTSHIFT:
The return value of `lv_timer_handler` can be used to automatically calculate the next start time, instead of using a user-specified time. This can reduce rendering and interaction delays.
I suggest renaming it to `lv_timer_periodic_handler`.

What do you think? @GorgonMeducer 

### Checkpoints
 
### DIFF
--- a/docs/porting/timer_handler.rst
+++ b/docs/porting/timer_handler.rst
@@ -35,6 +35,16 @@ the porting:
        ...
    }
 
+Or use the sleep time automatically calculated by LVGL:
+
+.. code:: c
+
+   while(1) {
+     ...
+     lv_timer_periodic_handler();
+     ...
+   }
+
 In an OS environment, you can use it together with the **delay** or
 **sleep** provided by OS to release CPU whenever possible:
 
--- a/src/misc/lv_timer.c
+++ b/src/misc/lv_timer.c
@@ -38,7 +38,7 @@ static bool lv_timer_run = false;
 static uint8_t idle_last = 0;
 static bool timer_deleted;
 static bool timer_created;
-static uint32_t timer_time_till_next;
+static uint32_t timer_time_untill_next;
 
 /**********************
  *      MACROS
@@ -123,13 +123,13 @@ LV_ATTRIBUTE_TIMER_HANDLER uint32_t lv_timer_handler(void)
         }
     } while(LV_GC_ROOT(_lv_timer_act));
 
-    uint32_t time_till_next = LV_NO_TIMER_READY;
+    uint32_t time_untill_next = LV_NO_TIMER_READY;
     next = _lv_ll_get_head(&LV_GC_ROOT(_lv_timer_ll));
     while(next) {
         if(!next->paused) {
             uint32_t delay = lv_timer_time_remaining(next);
-            if(delay < time_till_next)
-                time_till_next = delay;
+            if(delay < time_untill_next)
+                time_untill_next = delay;
         }
 
         next = _lv_ll_get_next(&LV_GC_ROOT(_lv_timer_ll), next); /*Find the next timer*/
@@ -144,19 +144,19 @@ LV_ATTRIBUTE_TIMER_HANDLER uint32_t lv_timer_handler(void)
         idle_period_start = lv_tick_get();
     }
 
-    timer_time_till_next = time_till_next;
+    timer_time_untill_next = time_untill_next;
     already_running = false; /*Release the mutex*/
 
-    TIMER_TRACE("finished (%" LV_PRIu32 " ms until the next timer call)", time_till_next);
+    TIMER_TRACE("finished (%" LV_PRIu32 " ms until the next timer call)", time_untill_next);
     LV_PROFILER_END;
-    return time_till_next;
+    return time_untill_next;
 }
 
 LV_ATTRIBUTE_TIMER_HANDLER void lv_timer_periodic_handler(void)
 {
     static uint32_t last_tick = 0;
 
-    if(lv_tick_elaps(last_tick) >= timer_time_till_next) {
+    if(lv_tick_elaps(last_tick) >= timer_time_untill_next) {
         lv_timer_handler();
         last_tick = lv_tick_get();
     }
@@ -277,6 +277,7 @@ void lv_timer_set_repeat_count(lv_timer_t * timer, int32_t repeat_count)
 void lv_timer_reset(lv_timer_t * timer)
 {
     timer->last_run = lv_tick_get();
+    lv_timer_handler_resume();
 }
 
 /**
@@ -296,6 +297,15 @@ void lv_timer_enable(bool en)
 uint8_t lv_timer_get_idle(void)
 {
     return idle_last;
+}
+
+/**
+ * Get idle period start tick
+ * @return the lv_timer idle period start tick
+ */
+uint32_t lv_timer_get_time_untill_next(void)
+{
+    return timer_time_untill_next;
 }
 
 /**
@@ -375,7 +385,7 @@ static uint32_t lv_timer_time_remaining(lv_timer_t * timer)
 static void lv_timer_handler_resume(void)
 {
     /*If there is a timer which is ready to run then resume the timer loop*/
-    if(timer_time_till_next == LV_NO_TIMER_READY) {
-        timer_time_till_next = 0;
+    if(timer_time_untill_next == LV_NO_TIMER_READY) {
+        timer_time_untill_next = 0;
     }
 }
--- a/src/misc/lv_timer.c
+++ b/src/misc/lv_timer.c
@@ -38,7 +38,7 @@ static bool lv_timer_run = false;
 static uint8_t idle_last = 0;
 static bool timer_deleted;
 static bool timer_created;
-static uint32_t timer_time_untill_next;
+static uint32_t timer_time_until_next;
 
 /**********************
  *      MACROS
@@ -123,13 +123,13 @@ LV_ATTRIBUTE_TIMER_HANDLER uint32_t lv_timer_handler(void)
         }
     } while(LV_GC_ROOT(_lv_timer_act));
 
-    uint32_t time_untill_next = LV_NO_TIMER_READY;
+    uint32_t time_until_next = LV_NO_TIMER_READY;
     next = _lv_ll_get_head(&LV_GC_ROOT(_lv_timer_ll));
     while(next) {
         if(!next->paused) {
             uint32_t delay = lv_timer_time_remaining(next);
-            if(delay < time_untill_next)
-                time_untill_next = delay;
+            if(delay < time_until_next)
+                time_until_next = delay;
         }
 
         next = _lv_ll_get_next(&LV_GC_ROOT(_lv_timer_ll), next); /*Find the next timer*/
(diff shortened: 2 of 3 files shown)


GorgonMeducer:
@FASTSHIFT  What you suggested is a good idea. I support you. Could you please add a dedicated implementation for the mentioned functionality and keep the original one, i.e. `lv_timer_handler_run_in_period()`?

I noticed that you already gave it a good name: `lv_timer_periodic_handler()`.

Why can we give users more options rather than replace my idea with yours?

FASTSHIFT:
> @FASTSHIFT What you suggested is a good idea. I support you. Could you please add a dedicated implementation for the mentioned functionality and keep the original one, i.e. ?`lv_timer_handler_run_in_period()`
> 
> I noticed that you already gave it a good name: .`lv_timer_periodic_handler()`
> 
> Why can we give users more options rather than replace my idea with yours?

Is the intention behind `lv_timer_handler_run_in_period` to force `lv_timer_handler` to run at the user-specified period, rather than the period calculated by lvgl?

GorgonMeducer:
Yes, exactly.

FASTSHIFT:
> Yes, exactly.

OK, I'll add another API.

kisvegabor:
I see a possible issue. Let's say there is no indevs, so we have only the animation and render timer. If there are no animations and there are no updates on the screen. `lv_timer_handler` will return `LV_NO_TIMER_READY` (0xffffff) (basically infinity).  If a the user presses a hardware button, which triggers an animation or render (e.g. `lv_scr_load()` is called) `time_till_next` won't be updated and `lv_timer_handle` won't be called. 

So we need a mechanism to  reset `time_till_next` in `lv_timer_resume`, `lv_timer_create`.  Or we can have an `lv_timer_get_time_until_next()` function. It returns an internal `time_till_next` variable which will be reset automatically and internally.

What do you think?

FASTSHIFT:
> I see a possible issue. Let's say there is no indevs, so we have only the animation and render timer. If there are no animations and there are no updates on the screen. `lv_timer_handler` will return `LV_NO_TIMER_READY` (0xffffff) (basically infinity). If a the user presses a hardware button, which triggers an animation or render (e.g. `lv_scr_load()` is called) `time_till_next` won't be updated and `lv_timer_handle` won't be called.
> 
> So we need a mechanism to reset `time_till_next` in `lv_timer_resume`, `lv_timer_create`. Or we can have an `lv_timer_get_time_until_next()` function. It returns an internal `time_till_next` variable which will be reset automatically and internally.
> 
> What do you think?

We may need to add a notification interface at the end of `_lv_inv_area` to notify the lvgl loop to start running again. In this notification interface we can reset `time_till_next`.

kisvegabor:
> We may need to add a notification interface at the end of _lv_inv_area to notify the lvgl loop to start running again. In this notification interface we can reset time_till_next.

Not necessarily. E.g. `_lv_inv_area` calls `lv_timer_resume`. So we can update `time_till_next` in `lv_timer_resume()`, `lv_timer_create()` and similar functions. 

In the end `time_till_next` can change only if a `lv_timer_...` function is called so we can handle all these in `lv_timer.c` locally.

FASTSHIFT:
But if my rendering action is triggered by the vsync signal, it is not controlled by the timer, how should we deal with this?

kisvegabor:
Do you mean if you decouple the refresh timer as described [here](https://docs.lvgl.io/master/porting/disp.html#decoupling-the-display-refresh-timer)? If so, if you don't use a timer for refreshing, it shouldn't count in `time_till_next` calculations either. Or am I misunderstanding something?

FASTSHIFT:
> Do you mean if you decouple the refresh timer as described [here](https://docs.lvgl.io/master/porting/disp.html#decoupling-the-display-refresh-timer)? If so, if you don't use a timer for refreshing, it shouldn't count in `time_till_next` calculations either. Or am I misunderstanding something?

Yes, although I did not use the timer to refresh the UI, but `lv_timer_handler` still needs to be called to handle some other affairs. When it returns `LV_NO_TIMER_READY`, so we need a notification mechanism to let it resume running.

kisvegabor:
As I started to write the example below it came to my mind that it can be even simpler. So this is what I mean:
```c
static inline LV_ATTRIBUTE_TIMER_HANDLER void lv_timer_periodic_handler(void)
{
    static uint32_t last_tick = 0;
 
    if(lv_tick_get() >= lv_timer_get_next_run_time()) {
        lv_timer_handler();
    }
}
```
`lv_timer_get_next_run_time` returns an absolute tick (e.g. 13478 ms) when the next task will run. 
- If a timer is created, deleted, paused, resumed etc, `lv_timer_get_next_run_time` will return a different value. 
- If the display refresh timer is removed, then it's not interested here as it's not a timer anymore. You can call it as you wish.

I can send a new PR to demonstrate what I mean in the practice.

FASTSHIFT:
> As I started to write the example below it came to my mind that it can be even simpler. So this is what I mean:
> 
> ```c
> static inline LV_ATTRIBUTE_TIMER_HANDLER void lv_timer_periodic_handler(void)
> {
>     static uint32_t last_tick = 0;
>  
>     if(lv_tick_get() >= lv_timer_get_next_run_time()) {
>         lv_timer_handler();
>     }
> }
> ```
> 
> `lv_timer_get_next_run_time` returns an absolute tick (e.g. 13478 ms) when the next task will run.
> 
> * If a timer is created, deleted, paused, resumed etc,  will return a different value.`lv_timer_get_next_run_time`
> * If the display refresh timer is removed, then it's not interested here as it's not a timer anymore. You can call it as you wish.
> 
> I can send a new PR to demonstrate what I mean in the practice.

What is the internal implementation of `lv_timer_get_next_run_time`, will it traverse the timer list every time?

kisvegabor:
> What is the internal implementation of lv_timer_get_next_run_time, will it traverse the timer list every time?

Not necessarily. We can have a single `static uint32_t next_run_time` that we update in `lv_timer_handler/resume/pause/etc`.

FASTSHIFT:
> > What is the internal implementation of lv_timer_get_next_run_time, will it traverse the timer list every time?
> 
> Not necessarily. We can have a single `static uint32_t next_run_time` that we update in `lv_timer_handler/resume/pause/etc`.

I implemented a `lv_timer_handler_resume` method to resume the timer loop operation, what do you think?

kisvegabor:
We might lose a little if `lv_timer_handler_resume` always sets  `timer_time_till_next` to zero. E.g. in `lv_timer_reset` `timer_time_till_next` could be `timer_time_till_next = LV_MIN(timer_time_till_next, timer->period)`). 

Could you also add an API function to get `timer_time_till_next`? By the way, as non native speaker I really don't know but would't it be cleaner to use `until` instead of `till`?

FASTSHIFT:
> We might lose a little if `lv_timer_handler_resume` always sets `timer_time_till_next` to zero. E.g. in `lv_timer_reset` `timer_time_till_next` could be `timer_time_till_next = LV_MIN(timer_time_till_next, timer->period)`).
> 
> Could you also add an API function to get `timer_time_till_next`? By the way, as non native speaker I really don't know but would't it be cleaner to use `until` instead of `till`?

I added a new interface: `lv_timer_get/set_time_untill_next`, does this meet your meaning?

FASTSHIFT:
> We might lose a little if `lv_timer_handler_resume` always sets `timer_time_till_next` to zero. E.g. in `lv_timer_reset` `timer_time_till_next` could be `timer_time_till_next = LV_MIN(timer_time_till_next, timer->period)`).

Set `timer_time_until_next` to 0, after waking up, `lv_timer_handler` will re-update `timer_time_until_next` to set a new sleep time to ensure that there will be no error. And it will only take effect when `timer_time_until_next == LV_NO_TIMER_READY`, it will not take effect when there is a resident timer.

> Could you also add an API function to get `timer_time_till_next`? By the way, as non native speaker I really don't know but would't it be cleaner to use `until` instead of `till`?

Done.

kisvegabor:
> it will not take effect when there is a resident timer.

What if there is a timer with 1000 ms period which currently running. Let's say the resident time is 900 ms. And you create a new timer with 1ms period time. This way we need to wait 900 ms for the first call of the new timer.

FASTSHIFT:
> > it will not take effect when there is a resident timer.
> 
> What if there is a timer with 1000 ms period which currently running. Let's say the resident time is 900 ms. And you create a new timer with 1ms period time. This way we need to wait 900 ms for the first call of the new timer.

You are right, so we should have `lv_timer_handler` resume running as soon as the timer parameter is changed:
```c
static void lv_timer_handler_resume(void)
{
    timer_time_until_next = 0;
}
```

Here's a test code I wrote that I think works as expected:

```c
static void timer_cb_1000ms(lv_timer_t* timer)
{
    LV_LOG_USER("ready");
}

static void timer_cb_10ms(lv_timer_t* timer)
{
    LV_LOG_USER("ready");
    lv_timer_del(timer);
}

int main(int argc, char** argv)
{
    uint32_t start_tick = lv_tick_get();

    /*Initialize LVGL*/
    lv_init();

    lv_timer_create(timer_cb_1000ms, 1000, NULL);

    while (1) {
        /* Create timer outside lv_timer_handler */
        if (lv_tick_elaps(start_tick) > 900) {
            static lv_timer_t* timer = NULL;
            if (!timer) {
                LV_LOG_USER("create timer");
                timer = lv_timer_create(timer_cb_10ms, 10, NULL);
            }
        }

        lv_timer_periodic_handler();
    }

    return 0;
}
```

```bash
[User]  (0.000, +0)      lv_timer_periodic_handler: calling lv_timer_handler()  (in lv_timer.c line #160)
[User]  (0.901, +901)    main: create timer     (in main.c line #100)
[User]  (0.901, +0)      lv_timer_periodic_handler: calling lv_timer_handler()  (in lv_timer.c line #160)
[User]  (0.911, +10)     lv_timer_periodic_handler: calling lv_timer_handler()  (in lv_timer.c line #160)
[User]  (0.911, +0)      timer_cb_10ms: ready   (in main.c line #70)
[User]  (1.000, +89)     lv_timer_periodic_handler: calling lv_timer_handler()  (in lv_timer.c line #160)
[User]  (1.000, +0)      timer_cb_1000ms: ready         (in main.c line #65)
[User]  (2.000, +1000)   lv_timer_periodic_handler: calling lv_timer_handler()  (in lv_timer.c line #160)
[User]  (2.000, +0)      timer_cb_1000ms: ready         (in main.c line #65)
[User]  (3.000, +1000)   lv_timer_periodic_handler: calling lv_timer_handler()  (in lv_timer.c line #160)
[User]  (3.000, +0)      timer_cb_1000ms: ready         (in main.c line #65)
[User]  (4.000, +1000)   lv_timer_periodic_handler: calling lv_timer_handler()  (in lv_timer.c line #160)
[User]  (4.000, +0)      timer_cb_1000ms: ready         (in main.c line #65)
[User]  (5.000, +1000)   lv_timer_periodic_handler: calling lv_timer_handler()  (in lv_timer.c line #160)
[User]  (5.000, +0)      timer_cb_1000ms: ready         (in main.c line #65)
[User]  (6.000, +1000)   lv_timer_periodic_handler: calling lv_timer_handler()  (in lv_timer.c line #160)
[User]  (6.000, +0)      timer_cb_1000ms: ready         (in main.c line #65)
[User]  (7.000, +1000)   lv_timer_periodic_handler: calling lv_timer_handler()  (in lv_timer.c line #160)
```

kisvegabor:
Looks good, thank you for the update.

Just one last thing: could you show any example about how it can be effectively combined with OS `sleep`?

FASTSHIFT:
> Looks good, thank you for the update.
> 
> Just one last thing: could you show any example about how it can be effectively combined with OS `sleep`?

If we use it in conjunction with OS, we may not be able to simply pass the return value of lv_timer_handler to os's sleep function. When `LV_NO_TIMER_READY` is returned, the semaphore is used to block the LVGL timer loop from running, and the semaphore is posted to enable the lvgl to resume running when an external event occurs.

FASTSHIFT:
Can we consider supporting patterns that are read by input devices triggered by events? Most touchpad chips support interrupt mode, reporting coordinates only when the user touches them. This is very meaningful for reducing power consumption, which can reduce the frequent wake up of the system.

kisvegabor:
In `feat/parallel-rendering` branch we have integrated an [OSAL](https://github.com/lvgl/lvgl/tree/feat/parallel-rendering/src/osal). In it we can add an abstraction for sleep too and other things too make `lv_timer` effective with OS.

For now I vote for merging it now and when `feat/parallel-rendering` is merged when can improve it further. What do you think?

FASTSHIFT:
> In `feat/parallel-rendering` branch we have integrated an [OSAL](https://github.com/lvgl/lvgl/tree/feat/parallel-rendering/src/osal). In it we can add an abstraction for sleep too and other things too make `lv_timer` effective with OS.
> 
> For now I vote for merging it now and when `feat/parallel-rendering` is merged when can improve it further. What do you think?

I agree, we can merge this PR first.

Issue #1858, opened 2020-10-21
Stride implementation


kisvegabor:
@jozba pointed out that the concept of "stride" is missing from LVGL but it's required for some GPUs.

I've opened this issue to discuss the requirements and need for this feature. 

@jozba could you summarize what we know so far?

jozba:
Usual approach to buffer processing is to use buffer width and height in pixels. With limited resources in embedded world, it makes sense to save every possible byte of memory. Also, most CPUs have byte access to any address. However, when HW accelerators are to be involved, new limitations arise here, which are the tax for offloading the CPU. 

1) Stride. HW accelerators (GPUs, VPUs, blitters, LCD controllers etc.) usually require buffers to meet some width requirements. This is called stride. For instance, HW can require a buffer stride to be multiple of 16 pixels. E.g. if buffer real data width is 100 px, allocated width has to be 112 px in order to be accessed correctly by HW. One may look at this as 12 pixels in each line is wasted in this case, but that's the cost for HW not being forced to access or process non-burst-sized chunks of memory. Some accelerators require even height stride (e.g. GPU working in tiled mode of 4x4 squares require height to be multiple of 4 px). 

If all SW algorithms use +stride to move "one line up", instead of +width, the buffer can be safely processed also by HW accelerators. In case of no stride requirements, stride can be configured to be "multiple of 1 px", meaning width = stride. 

This will affect also web image generator and other tools, that will need to generate c-arrays that meet the stride requirement, and also add information to image descriptor strucure, what is real data width, and what is actual stride. 

2) Alignment. Due to performance and design reasons, some HW blocks require the buffer starting address to be at some particular boundary. E.g, on 16 or 32 byte boundary. Even if there is no such requirement, aligned memory access is faster. 

For this reason, it would be good to have buffer allocator, that can platform specific. It could be used for allocating buffer that will be accessed by HW and will ensure that buffer will meet stride and alignment requirements.

kisvegabor:
Thanks for the explanation. For the display sizes, the width criteria usually meet by default. I.e. all comment widths (240, 320, 480, 800) are multiple of 16. 

However, it's a real issue with images, they can have any size. 

What if we simply want to fill an area with a color. E.g. the area is (7;1) (21;88). Is there some kind of requirements here as well?

jozba:
Buffer width of multiple 16px was just an example. It can be even 32px. But it doesn't matter if we have a generic approach to buffers that takes into account the stride and real width.

[...]

PR #1824, opened 2020-09-30
VG-Lite: Added support for fill and BLIT

jozba:
VG-Lite accelerated features:
  - fill (+ transparency)
  - BLIT (+ transparency)

Limitation:
  Image width must be aligned to 16 for VG-Lite to process the buffer.

Signed-off-by: Jozef Bastek <jozef.bastek@nxp.com>

PR #4241, opened 2023-05-24
feat(mem): add memalign API

FASTSHIFT:

Supplement `memalign` memory allocation API.

### Checkpoints
 
### DIFF
--- a/env_support/cmsis-pack/lv_conf_cmsis.h
+++ b/env_support/cmsis-pack/lv_conf_cmsis.h
@@ -56,6 +56,7 @@
 #define LV_STDIO_INCLUDE  <stdio.h>
 #define LV_STRING_INCLUDE <string.h>
 #define LV_MALLOC       lv_malloc_builtin
+#define LV_MEMALIGN     lv_memalign_builtin
 #define LV_REALLOC      lv_realloc_builtin
 #define LV_FREE         lv_free_builtin
 #define LV_MEMSET       lv_memset_builtin
--- a/lv_conf_template.h
+++ b/lv_conf_template.h
@@ -63,6 +63,7 @@
 #define LV_STDIO_INCLUDE  <stdint.h>
 #define LV_STRING_INCLUDE <stdint.h>
 #define LV_MALLOC       lv_malloc_builtin
+#define LV_MEMALIGN     lv_memalign_builtin
 #define LV_REALLOC      lv_realloc_builtin
 #define LV_FREE         lv_free_builtin
 #define LV_MEMSET       lv_memset_builtin
--- a/src/lv_conf_internal.h
+++ b/src/lv_conf_internal.h
@@ -203,6 +203,13 @@
         #define LV_MALLOC       lv_malloc_builtin
     #endif
 #endif
+#ifndef LV_MEMALIGN
+    #ifdef CONFIG_LV_MEMALIGN
+        #define LV_MEMALIGN CONFIG_LV_MEMALIGN
+    #else
+        #define LV_MEMALIGN     lv_memalign_builtin
+    #endif
+#endif
 #ifndef LV_REALLOC
     #ifdef CONFIG_LV_REALLOC
         #define LV_REALLOC CONFIG_LV_REALLOC
--- a/src/misc/lv_malloc_builtin.c
+++ b/src/misc/lv_malloc_builtin.c
@@ -164,6 +164,13 @@ void * lv_malloc_builtin(size_t size)
     return lv_tlsf_malloc(tlsf, size);
 }
 
+void * lv_memalign_builtin(size_t align, size_t size)
+{
+    cur_used += size;
+    max_used = LV_MAX(cur_used, max_used);
+    return lv_tlsf_memalign(tlsf, align, size);
+}
+
 void * lv_realloc_builtin(void * p, size_t new_size)
 {
     return lv_tlsf_realloc(tlsf, p, new_size);
--- a/src/misc/lv_malloc_builtin.h
+++ b/src/misc/lv_malloc_builtin.h
@@ -55,6 +55,7 @@ lv_mem_builtin_pool_t lv_mem_builtin_add_pool(void * mem, size_t bytes);
 void lv_mem_builtin_remove_pool(lv_mem_builtin_pool_t pool);
 
 void * lv_malloc_builtin(size_t size);
+void * lv_memalign_builtin(size_t align, size_t size);
 void * lv_realloc_builtin(void * p, size_t new_size);
 void lv_free_builtin(void * p);
 
--- a/src/misc/lv_mem.c
+++ b/src/misc/lv_mem.c
@@ -39,6 +39,9 @@
 /**********************
  *  STATIC PROTOTYPES
  **********************/
+#if LV_LOG_LEVEL <= LV_LOG_LEVEL_INFO
+    static void lv_mem_dump_info(void);
+#endif
 
 /**********************
  *  STATIC VARIABLES
@@ -75,11 +78,40 @@ void * lv_malloc(size_t size)
     if(alloc == NULL) {
         LV_LOG_INFO("couldn't allocate memory (%lu bytes)", (unsigned long)size);
 #if LV_LOG_LEVEL <= LV_LOG_LEVEL_INFO
-        lv_mem_monitor_t mon;
-        lv_mem_monitor(&mon);
-        LV_LOG_INFO("used: %6d (%3d %%), frag: %3d %%, biggest free: %6d",
-                    (int)(mon.total_size - mon.free_size), mon.used_pct, mon.frag_pct,
-                    (int)mon.free_biggest_size);
+        lv_mem_dump_info();
+        return NULL;
+#endif
+    }
+
+#if LV_MEM_ADD_JUNK
+    lv_memset(alloc, 0xaa, size);
+#endif
+
+    MEM_TRACE("allocated at %p", alloc);
+
+    return alloc;
+}
+
+/**
+ * Allocate a block of memory with a specified alignment.
+ * @param align The desired alignment of the memory block.
+ * @param size The size of the memory block to allocate.
+ * @return A pointer to the allocated memory block, or NULL if the allocation failed.
+ */
+void * lv_memalign(size_t align, size_t size)
+{
+    MEM_TRACE("align(%lu) allocating %lu bytes", (unsigned long)align, (unsigned long)size);
+    if(size == 0) {
+        MEM_TRACE("using zero_mem");
+        return &zero_mem;
+    }
+
+    void * alloc = LV_MEMALIGN(align, size);
+
+    if(alloc == NULL) {
+        LV_LOG_INFO("couldn't allocate memory (%lu bytes)", (unsigned long)size);
+#if LV_LOG_LEVEL <= LV_LOG_LEVEL_INFO
+        lv_mem_dump_info();
         return NULL;
 #endif
     }
@@ -185,3 +217,14 @@ void lv_mem_monitor(lv_mem_monitor_t * mon_p)
 /**********************
  *   STATIC FUNCTIONS
  **********************/
+
+#if LV_LOG_LEVEL <= LV_LOG_LEVEL_INFO
+static void lv_mem_dump_info(void)
+{
+    lv_mem_monitor_t mon;
+    lv_mem_monitor(&mon);
+    LV_LOG_INFO("used: %6d (%3d %%), frag: %3d %%, biggest free: %6d",
+                (int)(mon.total_size - mon.free_size), mon.used_pct, mon.frag_pct,
+                (int)mon.free_biggest_size);
+}
+#endif
--- a/src/misc/lv_mem.h
+++ b/src/misc/lv_mem.h
@@ -55,6 +55,14 @@ typedef struct {
 void * lv_malloc(size_t size);
 
 /**
+ * Allocate a block of memory with a specified alignment.
+ * @param align The desired alignment of the memory block.
+ * @param size The size of the memory block to allocate.
+ * @return A pointer to the allocated memory block, or NULL if the allocation failed.
+ */
+void * lv_memalign(size_t align, size_t size);
+
+/**
  * Free an allocated data
  * @param data pointer to an allocated memory
  */
--- a/tests/src/lv_test_conf.h
+++ b/tests/src/lv_test_conf.h
@@ -23,6 +23,7 @@ extern "C" {
 #define LV_USE_BUILTIN_SNPRINTF 1
 #define LV_STDLIB_INCLUDE <stdlib.h>
 #define LV_MALLOC       lv_test_malloc
+#define LV_MEMALIGN     lv_test_memalign
 #define LV_REALLOC      lv_test_realloc
 #define LV_FREE         lv_test_free
 #define LV_MEMSET       memset
@@ -34,6 +35,7 @@ extern "C" {
 #define LV_USE_BUILTIN_MEMCPY   1
 #define LV_USE_BUILTIN_SNPRINTF 1
 #define LV_MALLOC       lv_test_malloc
+#define LV_MEMALIGN     lv_test_memalign
 #define LV_REALLOC      lv_test_realloc
 #define LV_FREE         lv_test_free
 #define LV_MEMSET       lv_memset_builtin
@@ -47,6 +49,7 @@ extern "C" {
 #define LV_USE_BUILTIN_SNPRINTF 1
 #define LV_STDLIB_INCLUDE "include/lv_mp_mem_custom_include.h"
 #define LV_MALLOC       m_malloc
+#define LV_MEMALIGN     m_memalign
 #define LV_REALLOC      m_realloc
 #define LV_FREE         m_free
 #define LV_MEMSET       lv_memset_builtin
--- a/tests/src/lv_test_malloc.c
+++ b/tests/src/lv_test_malloc.c
@@ -5,12 +5,14 @@
 #endif
 
 static lv_malloc_stub_cb malloc_stub_cb;
+static lv_memalign_stub_cb memalign_stub_cb;
 static lv_realloc_stub_cb realloc_stub_cb;
 static lv_free_stub_cb free_stub_cb;
 
 void lv_test_malloc_init(void)
 {
     malloc_stub_cb = NULL;
+    memalign_stub_cb = NULL;
     realloc_stub_cb = NULL;
     free_stub_cb = NULL;
 }
@@ -29,6 +31,15 @@ void * lv_test_malloc_normal(size_t size)
 #endif
 }
 
+void * lv_test_memalign_normal(size_t align, size_t size)
+{
+#ifdef LVGL_CI_USING_SYS_HEAP
+    return memalign(align, size);
+#else // LVGL_CI_USING_DEF_HEAP and others
+    return lv_memalign_builtin(align, size);
+#endif
+}
+
 void * lv_test_realloc_normal(void * p, size_t new_size)
 {
 #ifdef LVGL_CI_USING_SYS_HEAP
@@ -42,6 +53,17 @@ void * lv_test_malloc(size_t s)
 {
     if(malloc_stub_cb) return malloc_stub_cb(s);
     else return lv_test_malloc_normal(s);
+}
+
+void lv_test_memalign_set_cb(lv_memalign_stub_cb stub_memalign)
+{
+    memalign_stub_cb = stub_memalign;
+}
+
+void * lv_test_memalign(size_t align, size_t size)
+{
+    if(memalign_stub_cb) return memalign_stub_cb(align, size);
+    else return lv_test_memalign_normal(align, size);
 }
 
 void lv_test_realloc_set_cb(lv_realloc_stub_cb stub_realloc)
--- a/tests/src/lv_test_malloc.h
+++ b/tests/src/lv_test_malloc.h
@@ -4,12 +4,16 @@
 #include <stdlib.h>
 
 typedef void * (* lv_malloc_stub_cb)(size_t);
+typedef void * (* lv_memalign_stub_cb)(size_t align, size_t size);
 typedef void * (* lv_realloc_stub_cb)(void * p, size_t new_size);
 typedef void (* lv_free_stub_cb)(void * p);
 
 void lv_test_malloc_init(void);
 void lv_test_malloc_set_cb(lv_malloc_stub_cb stub_malloc);
 void * lv_test_malloc(size_t s);
+
+void lv_test_memalign_set_cb(lv_memalign_stub_cb stub_memalign);
+void * lv_test_memalign(size_t align, size_t size);
 
 void lv_test_realloc_set_cb(lv_realloc_stub_cb stub_realloc);
 void * lv_test_realloc(void * p, size_t new_size);


FASTSHIFT:
`memalign` is not a standard C library function, is there any solution?

kisvegabor:
There were several discussions about  similar features so I'd like to ask why/in which scenario you need to allocate aligned memory. The main use case that I can imagine is getting a buffer for drawing, but for this we need a more complex and carefully designed API.

FASTSHIFT:
> There were several discussions about similar features so I'd like to ask why/in which scenario you need to allocate aligned memory. The main use case that I can imagine is getting a buffer for drawing, but for this we need a more complex and carefully designed API.

Yes, I am thinking of designing an API dedicated to get drawing buffer, which has the following functions:
1. Supports memory alignment, because many GPU/DMA hardware has strong requirements for address alignment.
2. It cooperates with img cache, when the memory The cache is automatically released when the allocation is insufficient.
3. Independent from the main heap to solve the problem of large-size buffer allocation failure caused by memory fragmentation.

FASTSHIFT:
In addition, when do we plan to add stride support? https://github.com/lvgl/lvgl/issues/1858

kisvegabor:
I'm thinking mostly about the same.  

I'm thinking if there are any extra use cases besides allocating a buffer for drawing. If not we can choose a more specific name, like `lv_malloc_draw_buf(size)`, where alignment is platform specific and comes from a `define`. 

cc @nicusorcitu

nicusorcitu:
More memory alignment discussion are [here](https://github.com/lvgl/lvgl/pull/1824#issuecomment-1501875731)

FASTSHIFT:
> I'm thinking mostly about the same.
> 
> I'm thinking if there are any extra use cases besides allocating a buffer for drawing. If not we can choose a more specific name, like `lv_malloc_draw_buf(size)`, where alignment is platform specific and comes from a `define`.
> 
> cc @nicusorcitu

I agree with this.

kisvegabor:
What do you think about the "flag" idea from [this comment](https://github.com/lvgl/lvgl/pull/1824#issuecomment-1508259006)? Do you think we need them?  I'm not sure. In the practice I can hardly image many type of draw buffers. 

----

Slightly related topic: I'm planning to update how the LVGL'S STDIO wrappers are implemented. Now we have defines like `LV_MALLOC  some_function`.  But I think it'd be cleaner and not worse if we had a single `LV_USE_STDIO_IMPLEMENTATION`
 define with values like
```
LV_STDIO_NONE       0 
LV_STDIO_C          1
LV_STDIO_LVGL       2
LV_STDIO_RT_THREAD  3
LV_STDIO_NUTTX      4
```

We could have files like `lv_stdio_c.c/h` with `#if LV_USE_STDIO == LV_STDIO_C` where we can have the implementations of `lv_malloc`, `lv_free`, `lv_strcpy`, `lv_snprintf`, etc.

This way we won't be able to easily mix e.g. LVGL's `malloc` with C `strcpy` but it could be important. So we can have:
```
LV_USE_MALLOC_IMPLEMENTATION
LV_USE_STR_IMPLEMENTATION
LV_USE_MEM_IMPLEMENTATION
```

Maybe instead of `LV_USE_` we can introduce `LV_SELECT_`

The extra advantage of this approach is that it's Kconfig compatible.


What do you think?

FASTSHIFT:
> What do you think about the "flag" idea from [this comment](https://github.com/lvgl/lvgl/pull/1824#issuecomment-1508259006)? Do you think we need them? I'm not sure. In the practice I can hardly image many type of draw buffers.
> 
> Slightly related topic: I'm planning to update how the LVGL'S STDIO wrappers are implemented. Now we have defines like `LV_MALLOC some_function`. But I think it'd be cleaner and not worse if we had a single `LV_USE_STDIO_IMPLEMENTATION` define with values like
> 
> ```
> LV_STDIO_NONE       0 
> LV_STDIO_C          1
> LV_STDIO_LVGL       2
> LV_STDIO_RT_THREAD  3
> LV_STDIO_NUTTX      4
> ```
> 
> We could have files like `lv_stdio_c.c/h` with `#if LV_USE_STDIO == LV_STDIO_C` where we can have the implementations of `lv_malloc`, `lv_free`, `lv_strcpy`, `lv_snprintf`, etc.
> 
> This way we won't be able to easily mix e.g. LVGL's `malloc` with C `strcpy` but it could be important. So we can have:
> 
> ```
> LV_USE_MALLOC_IMPLEMENTATION
> LV_USE_STR_IMPLEMENTATION
> LV_USE_MEM_IMPLEMENTATION
> ```
> 
> Maybe instead of `LV_USE_` we can introduce `LV_SELECT_`
> 
> The extra advantage of this approach is that it's Kconfig compatible.
> 
> What do you think?

I support adding FLAG parameters to memory allocation to enhance scalability, allowing for more precise control over memory. However, this may also increase the complexity of design and implementation.

Regarding STDIO wrapping, I believe that apart from dynamic memory allocation and file system interfaces, for other functions such as strcpy and snprintf, I prefer using the standard C interfaces provided by the toolchain. In a large project, most programs rely on the standard C interfaces, and I think LVGL can also benefit from them. Introducing too many STDIO wrappers would add to the complexity of the code.

kisvegabor:
> I support adding FLAG parameters to memory allocation to enhance scalability, allowing for more precise control over memory. However, this may also increase the complexity of design and implementation.

Let's collect the flags, and we will see. This is what I have in mind:
- fast but small: use internal SRAM
- slower but larger: external RAM
- for DMA
- for GPU

> I prefer using the standard C interfaces provided by the toolchain. 

It turned out the e.g. the toolchain's `memcpy` is significantly slower in some cases then our basic `lv_memcpy` implementation. That's why I like the idea of having a these functions implemented in LVGL to provide a platform independent solution.

kisvegabor:
Based on https://github.com/lvgl/lvgl/issues/4258 an `lv_draw_buf_invalidate_cache()` function might be required too.

FASTSHIFT:
> Let's collect the flags, and we will see. This is what I have in mind:
> 
> * fast but small: use internal SRAM
> * slower but larger: external RAM
> * for DMA
> * for GPU

Have you learned about which hardware has special requirements for buffers? We can first understand and summarize some hardware requirements for comprehensive consideration, because some functions may overlap.

These are some hardware requirements as far as I know:
VG-Lite:
* The first address is 64-byte aligned
* Image width 16px alignment

ESP32:
* DMA can only access a limited address area
* SRAM is small but fast
* PSRAM is large but slow

> It turned out the e.g. the toolchain's memcpy is significantly slower in some cases then our basic lv_memcpy implementation.  That's why I like the idea of having a these functions implemented in LVGL to provide a platform independent solution.

For functions that have a large impact on performance, such as memcpy/memset functions, I support the use of lvgl's internal implementation and provide a wrapper interface, but for small data volumes that are not sensitive to performance, such as strcpy and strlen, I still prefer to use toolchain implementations.

kisvegabor:
> We can first understand and summarize some hardware requirements for comprehensive consideration, because some functions may overlap.

So in summary
- align (platform specific, e.g. if we enable VGLite, we know it's 64 byte) 
- stride (also platform specific)
- small/large doesn't need to be flag as the allocator can decide what to do based on the `size`
- fast/slow: also shouldn't be a flag as we should always use the fastest possible option. (e.g. SRAM for small, ext RAM for large)
- DMA: I'm not sure we can decide it it advance whether the buffer will be used by DMA later or not. E.g. a decoded PNG might be blended by GPU or CPU too.

So finally I think flags cannot convey too many information.

E.g. if in `lv_conf.h` the user selects `LV_VENDOR_NXP` the content of `lv_malloc_buf_nxp.c` is enabled in which we have `lv_malloc_draw_buf()` and `lv_draw_buf_invalidate_cache()`, which work depending of the `LV_USE_NXP_VGLITE` and `LV_USE_NXP_PXP` affecting stride and alignment.


> For functions that have a large impact on performance, such as memcpy/memset functions, I support the use of lvgl's internal implementation and provide a wrapper interface, but for small data volumes that are not sensitive to performance, such as strcpy and strlen, I still prefer to use toolchain implementations.

Probably you are right. `memcpy`, `memset`, and `snprintf` + `malloc`, `free`, `realloc` should be enough. `str*` might be unnecessary. 

---

Some new thoughts from #4258:
>> Can you confirm that using a platform specific lv_malloc_draw_buf(size, flags) functions to allocate buffers for decoding can solve the issue of "we need to provide a private allocation function to hardware vendors during image decoding."

> A: yes, lv_malloc_draw_buf can solve this issue.

>> Implementing the event_cb on images would take some time, and there are a lot of open questions. Can we manage the problem of cache invalidation by always calling a lv_draw_buf_invalidate_cache(buf, buf_size) function after decoding?

> A: I agree with it.


cc @terry0012

FASTSHIFT:
> > We can first understand and summarize some hardware requirements for comprehensive consideration, because some functions may overlap.
> 
> So in summary
> 
> * align (platform specific, e.g. if we enable VGLite, we know it's 64 byte)
> * stride (also platform specific)
> * small/large doesn't need to be flag as the allocator can decide what to do based on the `size`
> * fast/slow: also shouldn't be a flag as we should always use the fastest possible option. (e.g. SRAM for small, ext RAM for large)
> * DMA: I'm not sure we can decide it it advance whether the buffer will be used by DMA later or not. E.g. a decoded PNG might be blended by GPU or CPU too.
> 
> So finally I think flags cannot convey too many information.
> 
> E.g. if in `lv_conf.h` the user selects `LV_VENDOR_NXP` the content of `lv_malloc_buf_nxp.c` is enabled in which we have `lv_malloc_draw_buf()` and `lv_draw_buf_invalidate_cache()`, which work depending of the `LV_USE_NXP_VGLITE` and `LV_USE_NXP_PXP` affecting stride and alignment.

I totally agree.
We only need to expose the two memory allocation interfaces of `lv_malloc` and `lv_malloc_draw_buf` to the upper layer application, which can ensure maximum portability.

> > For functions that have a large impact on performance, such as memcpy/memset functions, I support the use of lvgl's internal implementation and provide a wrapper interface, but for small data volumes that are not sensitive to performance, such as strcpy and strlen, I still prefer to use toolchain implementations.
> 
> Probably you are right. `memcpy`, `memset`, and `snprintf` + `malloc`, `free`, `realloc` should be enough. `str*` might be unnecessary.
> 

OK.

kisvegabor:
Great! @nicusorcitu, do you also agree with it?

nicusorcitu:
Hi Gabor,
I am out of office with limited internet access. I will get back monday, sorry for delay. Thanks.
Trimis: Monday, June 5, 2023 6:44:37 PM
Către: lvgl/lvgl ***@***.***>
Cc: Nicușor Cîțu ***@***.***>; Mention ***@***.***>
Subiect: [EXT] Re: [lvgl/lvgl] feat(mem): add memalign API (PR #4241)

Great! @nicusorcitu<https://github.com/nicusorcitu>, do you also agree with it?

nicusorcitu:
>E.g. if in lv_conf.h the user selects LV_VENDOR_NXP the content of lv_malloc_buf_nxp.c is enabled in which we have lv_malloc_draw_buf() and lv_draw_buf_invalidate_cache(), which work depending of the LV_USE_NXP_VGLITEand LV_USE_NXP_PXP affecting stride and alignment.
>
I am still by phone but i will try to respond.

  1.   Why do we need a platform specific implementation for "lv_malloc_draw_buf()"? It can be a generic one to use a define value for alignament in bytes. I suppose is the same implementation if we have alignament of 64 bytes, 32 bytes or 0 bytes.
  2.  How is the stride handled? I am not clear about it.
  3.  "Lv_draw_buf_invalidate_cache()" make sense to be platform specific and clear only a specific address range. I agree on that one.

Thanks.
De la: Nicușor Cîțu ***@***.***>
Trimis: Thursday, June 8, 2023 9:08:05 AM
Către: lvgl/lvgl ***@***.***>; lvgl/lvgl ***@***.***>
Cc: Mention ***@***.***>
Subiect: Re: [EXT] Re: [lvgl/lvgl] feat(mem): add memalign API (PR #4241)

Hi Gabor,
I am out of office with limited internet access. I will get back monday, sorry for delay. Thanks.
Trimis: Monday, June 5, 2023 6:44:37 PM
Către: lvgl/lvgl ***@***.***>
Cc: Nicușor Cîțu ***@***.***>; Mention ***@***.***>
Subiect: [EXT] Re: [lvgl/lvgl] feat(mem): add memalign API (PR #4241)

Great! @nicusorcitu<https://github.com/nicusorcitu>, do you also agree with it?

kisvegabor:
Thank you for responding, but enjoy your holiday :slightly_smiling_face: 

>   Why do we need a platform specific implementation for "lv_malloc_draw_buf()"? It can be a generic one to use a define value for alignament in bytes. I suppose is the same implementation if we have alignament of 64 bytes, 32 bytes or 0 bytes.

Actually it's not only vendor but board dependent. E.g. a board can have a large eternal RAM, a medium sized internal RAM, and a small tightly coupled RAM. We should provide a way to manage all these for a given board setup.

In the built-in solution we can provided some functionalities via `lv_conf.h` defines. E.g. if the buffer size is smaller than `LV_MALLOC_DRAW_BUF_SMALL_LIMIT` use `lv_malloc`, else allocate in the `LV_MALLOC_DRAW_BUF_LARGE_START / END_ADR` .

> How is the stride handled? I am not clear about it.

It can be set automatically depending on the selected vendor and GPUs, and considered in `lv_malloc_draw_buf`.

nicusorcitu:
> Actually it's not only vendor but board dependent. E.g. a board can have a large eternal RAM, a medium sized internal RAM, and a small tightly coupled RAM. We should provide a way to manage all these for a given board setup.

>In the built-in solution we can provided some functionalities via lv_conf.h defines. E.g. if the buffer size is smaller than LV_MALLOC_DRAW_BUF_SMALL_LIMIT use lv_malloc, else allocate in the LV_MALLOC_DRAW_BUF_LARGE_START / END_ADR
>
I do not understand why we can not satisfy both conditions with the builtin function?!
I think the same implementation is already available in the `lv_mem_init_builtin()`:
```
    /*Allocate a large array to store the dynamically allocated data*/
    static LV_ATTRIBUTE_LARGE_RAM_ARRAY MEM_UNIT work_mem_int[LV_MEM_SIZE / sizeof(MEM_UNIT)];
    tlsf = lv_tlsf_create_with_pool((void *)work_mem_int, LV_MEM_SIZE);
```
We can make use of that attribute: `LV_ATTRIBUTE_LARGE_RAM_ARRAY` and to be defined as using a region from external RAM.
We can define a new define: `LV_ATTRIBUTE_SMALL_RAM_ARRAY` and to use a region from internal RAM.
Based on a `LV_MALLOC_DRAW_BUF_SMALL_LIMIT` value, we can decide which region to be used. All those defines are platform specific and can be set different on each platform in lv_conf.h.

>> How is the stride handled? I am not clear about it.

>It can be set automatically depending on the selected vendor and GPUs, and considered in lv_malloc_draw_buf.
>
My understanding is that `lv_malloc_draw_buf()` will be used to allocate a buffer to draw (destination buffer).

On GPU, the liniar format images require stride alignment and memory address alignment **only** for the source buffer (e.g. image img_cogwheel_rgb565).
E.G. if the img_cogwheel_rgb_565 is 100x100 px, we need to hold it into a buffer of 112x100px (each line shall have an extra 12px of data padding). Width is 100 while stride is 112. How we can handle that?

Destination buffer require alignment **only** in case of tiled format (pixels grouped bye 4x4).

kisvegabor:
> We can make use of that attribute: LV_ATTRIBUTE_LARGE_RAM_ARRAY and to be defined as using a region from external RAM.
We can define a new define: LV_ATTRIBUTE_SMALL_RAM_ARRAY and to use a region from internal RAM.
Based on a LV_MALLOC_DRAW_BUF_SMALL_LIMIT value, we can decide which region to be used. All those defines are platform specific and can be set different on each platform in lv_conf.h.

We can really do something like this. But what I say, is only that we need to provide a way for users to implement all these themsleves too. 


> On GPU, the liniar format images require stride alignment and memory address alignment only for the source buffer (e.g. image img_cogwheel_rgb565).

If there is a global `LV_DRAW_BUF_STRIDE 16` config option `lv_malloc_draw_buf` will return the buffer by considering the stride. Anyone uses that buffer later shall know that it has `LV_DRAW_BUF_STRIDE` stride.

> Destination buffer require alignment only in case of tiled format (pixels grouped bye 4x4).

For the sake of simlicity I would treat both with the same (larger) stride.

terry0012:
> > We can make use of that attribute: LV_ATTRIBUTE_LARGE_RAM_ARRAY and to be defined as using a region from external RAM.
> > We can define a new define: LV_ATTRIBUTE_SMALL_RAM_ARRAY and to use a region from internal RAM.
> > Based on a LV_MALLOC_DRAW_BUF_SMALL_LIMIT value, we can decide which region to be used. All those defines are platform specific and can be set different on each platform in lv_conf.h.
> 
> We can really do something like this. But what I say, is only that we need to provide a way for users to implement all these themsleves too.
> 

Hi，Gabor，can we confirm that this function can be completed in this way （use lv_malloc_draw_buf）?

FASTSHIFT:
I think we can simply implement a set of interfaces first: 
`lv_malloc_draw_buf`
`lv_realloc_draw_buf`
`lv_aligned_alloc_draw_buf`
`lv_free_draw_buf`
Its biggest role is to separate from `lv_malloc` instead of considering stride, and it is no different from `lv_malloc` in use, so it can be easily connected with memory allocators.
The memory allocation interface with stride can be repackaged based on this set of interfaces.
Because the stride requirements may change dynamically due to the color format, directly using a fixed `LV_DRAW_BUF_STRIDE` value may not meet the requirements.

kisvegabor:
> Because the stride requirements may change dynamically due to the color format, directly using a fixed LV_DRAW_BUF_STRIDE value may not meet the requirements.

I think one of the most important parts of this is to use a constant stride for all buffers. Without it, it can happen that e.g. the PNG decoder uses a stride which won't be compatible with the GPU in the end.

Besides as far as I know the [stride is in bytes](https://learn.microsoft.com/en-us/windows/win32/medfound/image-stride), so it should be independent of the color format. Actually we were thinking about using pixel or byte as stride, but seems to answer the question. 

I'd treat the alignment the same way as stride, that is by using a global `DRAW_BUF_ALIGN` config.

I agree with the proposed set of functions, but let's add the arguments too: 
```c
typedef struct {
    void * buf;
    lv_coord_t width;
    lv_coord_t stride;
    lv_coord_t height;
}lv_draw_buf_t;

lv_draw_buf_t  * lv_malloc_draw_buf(lv_coord_t w, lv_coord_t h);
lv_draw_buf_t  * lv_realloc_draw_buf(lv_draw_buf_t  * buf, lv_coord_t w, lv_coord_t h);
void lv_free_draw_buf(lv_draw_buf_t  * buf);
```

What do you think?

FASTSHIFT:
> ```c
>  w, lv_coord_t h)
> ```

What about the compressed format? We cannot calculate the buffer size through the width and height, because the actual occupied size is related to the content of the image.

FASTSHIFT:
> > Because the stride requirements may change dynamically due to the color format, directly using a fixed LV_DRAW_BUF_STRIDE value may not meet the requirements.
> 
> I think one of the most important parts of this is to use a constant stride for all buffers. Without it, it can happen that e.g. the PNG decoder uses a stride which won't be compatible with the GPU in the end.
> 
> Besides as far as I know the [stride is in bytes](https://learn.microsoft.com/en-us/windows/win32/medfound/image-stride), so it should be independent of the color format. Actually we were thinking about using pixel or byte as stride, but seems to answer the question.
> 
> I'd treat the alignment the same way as stride, that is by using a global `DRAW_BUF_ALIGN` config.
> 
> I agree with the proposed set of functions, but let's add the arguments too:
> 
> ```c
> typedef struct {
>     void * buf;
>     lv_coord_t width;
>     lv_coord_t stride;
>     lv_coord_t height;
> }lv_draw_buf_t;
> 
> lv_draw_buf_t  * lv_malloc_draw_buf(lv_coord_t w, lv_coord_t h);
> lv_draw_buf_t  * lv_realloc_draw_buf(lv_draw_buf_t  * buf, lv_coord_t w, lv_coord_t h);
> void lv_free_draw_buf(lv_draw_buf_t  * buf);
> ```
> 
> What do you think?

What about the compressed format? We cannot calculate the buffer size through the width and height, because the actual occupied size is related to the content of the image.

kisvegabor:
During decoding we always know the real width and height. E.g. in case of png it's read [here](https://github.com/lvgl/lvgl/blob/9f901067025d302cda6598493ff789b6fc01ae67/src/libs/png/lv_png.c#L103-L133).

FASTSHIFT:
> During decoding we always know the real width and height. E.g. in case of png it's read [here](https://github.com/lvgl/lvgl/blob/9f901067025d302cda6598493ff789b6fc01ae67/src/libs/png/lv_png.c#L103-L133).

For the [ETC2 format](https://en.wikipedia.org/wiki/Ericsson_Texture_Compression), it can be directly read and drawn by the GPU (no need to decode into RGB data), the memory they occupy is not determined by the width, height and bpp, but by the size of the file.

nicusorcitu:
```
typedef struct {
    void * buf;
    lv_coord_t width;
    lv_coord_t stride;
    lv_coord_t height;
}lv_draw_buf_t;
```
If you want to use lv_coord_t as type, then you should consider stride in px same as width/height. When you say "coordinates" it refers to pixel coordinates.

>Because the stride requirements may change dynamically due to the color format, directly using a fixed LV_DRAW_BUF_STRIDE value may not meet the requirements.
>
Yes, it is true. This is why I think making the stride alignment in pixels will be simpler because it can meet the requirement of a single defined `LV_DRAW_BUF_STRIDE_PX 16`.
The dinamic stride in bytes can be calculated based on this formula:
`stride_bytes = stride_px * bits_per_pixels(format) / 8;`
But stride_bytes shall be *no less than 8 bytes* (for index1/2/4 formats the above formula does not apply).

kisvegabor:
> For the [ETC2 format](https://en.wikipedia.org/wiki/Ericsson_Texture_Compression), it can be directly read and drawn by the GPU (no need to decode into RGB data), the memory they occupy is not determined by the width, height and bpp, but by the size of the file.

In this case the GPU (OpenGL) will handle the the image drawing. It gets an image in any format and can decide what to do with it. If it can manage that image format on it's own, it doesn't have to call any decoder, just draw the image. So no LVGL draw buffer is involved. Or am I missing something?

> If you want to use lv_coord_t as type, then you should consider stride in px same as width/height. When you say "coordinates" it refers to pixel coordinates.

Okay, We can use `uint32_t stride`. Actually in v9 I1m considering to remove `lv_coord_t` and replace it with `plain int32_t`.


> Yes, it is true. This is why I think making the stride alignment in pixels will be simpler because it can meet the requirement of a single defined LV_DRAW_BUF_STRIDE_PX 16.
The dinamic stride in bytes can be calculated based on this formula:
stride_bytes = stride_px * bits_per_pixels(format) / 8;
But stride_bytes shall be no less than 8 bytes (for index1/2/4 formats the above formula does not apply).

I thought stride is independent from the color format and it's always e.g. 16 for A8 and ARGB888 too. So isn't it the case?

----

We should add a color format parameter too: 
```c
typedef struct {
    void * buf;
    lv_coord_t width;
    lv_coord_t height;
    uint32_t stride;
    lv_color_format_t color_format;
}lv_draw_buf_t;

lv_draw_buf_t  * lv_malloc_draw_buf(lv_coord_t w, lv_coord_t h, lv_color_format_t color_format);
```

terry0012:
> > For the [ETC2 format](https://en.wikipedia.org/wiki/Ericsson_Texture_Compression), it can be directly read and drawn by the GPU (no need to decode into RGB data), the memory they occupy is not determined by the width, height and bpp, but by the size of the file.
> 
> In this case the GPU (OpenGL) will handle the the image drawing. It gets an image in any format and can decide what to do with it. If it can manage that image format on it's own, it doesn't have to call any decoder, just draw the image. So no LVGL draw buffer is involved. Or am I missing something?

For pkm files with ETC2 format, it contains a 16-byte header information and a data section.  If we treat is as a image source, we still need to parse the pkm file and write the data to the draw buffer as input data for GPU rendering. In this case, the data size still needs to be passed by lv_malloc_draw_buf.

kisvegabor:
In the new parallel rendering architecture (will be merged in a few weeks) it works like this:
- we need to parse the header to get width and height and save a pointer to the image as an image source
- when the image needs to be drawn LVGL creates a `draw_task`
- GPUs will see that there is an image with a format that I can process and pick that `draw_task`
- GPU renders the special image as it wants

You can check out the idea of the new architecture here: https://github.com/lvgl/lvgl/issues/4016#issue-1599634031

FASTSHIFT:
Not stale.

kisvegabor:
Okkkay, finally the parallel rendering update is merged. As this feature would be very important for GPU integration too, let's continue with it.

So in summary I think this is what we need:
```c
typedef struct {
    void * buf;
    lv_coord_t width;
    lv_coord_t height;
    uint32_t stride;
    lv_color_format_t color_format;
}lv_draw_buf_t;

lv_draw_buf_t  * lv_draw_buf_malloc(lv_coord_t w, lv_coord_t h, lv_color_format_t color_format);
lv_draw_buf_t  * lv_draw_buf_realloc(lv_draw_buf_t  * buf, lv_coord_t w, lv_coord_t h, lv_color_format_t color_format);
void lv_draw_buf_free(lv_draw_buf_t  * buf);
void lv_draw_buf_invalidate_cache(lv_draw_buf_t  * buf)
```

These should be in a file like `lv_mem_buf_builtin.c` and there should be a common `lv_mem_buf.h`. 

Later we can add useful features to the built-in allocator, but in the first round let's use the the normal malloc internally. If we don't talk about images for now, we need to use these functions only during [layer allocation](https://github.com/lvgl/lvgl/blob/388a5d7f975a91eb964e419e8d3ab81fd8a306c5/src/draw/sw/lv_draw_sw.c#L120) and optionally draw buf allocation.

I've used the concept of stride in the rendering algorithms, but it's not fully consistent and needs to be tested of course.

Do you plan to work on the implementation of this feature?

nicusorcitu:
I wold say the stride shall be used for images too, or for small buffers as letters. Not only for layers.

kisvegabor:
Letters are also a good example for sure. The letter buffer is allocated [here](https://github.com/lvgl/lvgl/blob/master/src/draw/lv_draw_label.c#L399).

kisvegabor:
I started to work on the draw_buf API. It should be ready in Monday.

kisvegabor:
I've pushed the mem align and draw bug update here: https://github.com/lvgl/lvgl/tree/feat/draw_buf

Although we were talking about it, I think the most controversial part will be that the stride is not stored in the the images and draw buffers.

It's because the stride can be always calculated from the width of the image, its color format and `LV_DRAW_BUF_STRIDE_ALIGN`. Storing this value could cause inconsistency and hell of special cases. 

For C array images `header.w` should be set to the desired width and each line should be padded to match `LV_DRAW_BUF_STRIDE_ALIGN`. In the image converter we can do something like:
```c
//Let's say width in bytes is 10
#if LV_DRAW_BUF_STRIDE_ALIGN == 1
  #define STRIDE_PAD
#elif LV_DRAW_BUF_STRIDE_ALIGN == 4
  #define STRIDE_PAD  0x00, 0x00   //2 byte padding is needed (10 + 2 = 12)
#elif LV_DRAW_BUF_STRIDE_ALIGN == 8
  #define STRIDE_PAD  0x00, 0x00, 0x00, 0x00, 0x00, 0x00  //10 + 6 = 16
...
#elif LV_DRAW_BUF_STRIDE_ALIGN == 256
  #define STRIDE_PAD  0x00, 0x00, ...   //10 + 246 + 6 = 256
#endif


static uint8_t buf[] = {
  0x11, 0x22, 0x33, ... STRIDE_PAD,
  0xaa, 0xbb, 0xcc, ... STRIDE_PAD,
}
```

For images from files, we either:
1. update the decoders to consider stride but it's not always possible (e.g. in case of FFMPEG decoder)
2. Use images with larger width to fulfill the stride requirements and somehow tell what is the real width. 

Maybe  1) is better as we need to use `lv_draw_buf_t` anyway for mage decoders for unified cache invalidation.

nicusorcitu:
Why did you removed the buffer_copy callback?

kisvegabor:
> Why did you removed the buffer_copy callback?

It's replaced by `lv_draw_buf_copy()`. It didn't make sense to store the copy callback in each layer instance because it's specific to the given platform so it can be implemented in a central place.

kisvegabor:
Similarly to the OS API I'll update the draw buf implementation to have platform specific parts. E.g. `lv_draw_buf_builtin.c`, `lv_draw_buf_nxp.c`.

nicusorcitu:
> Similarly to the OS API I'll update the draw buf implementation to have platform specific parts.
>
Please let me know when it is ready (no pressure) so I will start looking on the nxp implementation.

It was under discussion also how to integrate the clean and invalidate data cache.
Till now that callback could have been set as a callback to the display. It it not related to the display but it was a very nice way to have it set from the SDK integration code and get called from the lvgl draw side. If we will move it to the lvgl draw side, there will be a lot of platform specific defines.

kisvegabor:
@nicusorcitu 
I'll work on it next week and we will see if something needs to be imporved.

kisvegabor:
@nicusorcitu According to your experience, is better to use pixel or bytes for stride? E.g. in the NXP GPU is strides defined is bytes or pixels?

nicusorcitu:
The stride is measured in bytes. Whenever I need to set it for the hardware (both pxp and vglite) it is set in bytes.
But on the upper layer, like on the lvgl framework, I do not think it really matter if it is bytes or pixels. You can use which one is more convenient and cleaner.
From the latest code I see you are using a translation from color format to bytes in 'lv_draw_buf_width_to_stride' -> `uint32_t width_byte =  w * lv_color_format_get_size(color_format);`
Keep in mind that not all of the formats are covered by `lv_color_format_get_size`. For example the format `LV_COLOR_FORMAT_I2` have 2 bits/px. So probably the function shall return the value in bits and not pixels.

**The general rule for vglite, the stride is defined as 16 pixels (in bytes) but no less than 8 bytes.**
E.G. `LV_COLOR_FORMAT_I2` have the stride = 8 bytes which represents actually 4*8=32 pixels.

image
image

kisvegabor:
Thank you, Nicu. Let's continue in #4414

nicusorcitu:
> I'll work on it next week and we will see if something needs to be imporved.
>
I still want to understand what shall I do for draw_buf for the GPU purpose? Shall I create a separate file for that?
As I looked right now only those APIs require "custom" implementation:
`lv_draw_buf_copy()`
`lv_draw_buf_invalidate_cache()` - which BTW shall have an extra argument with the size in bytes that need to be cleared.

PR #4249, opened 2023-05-26
Fixes few things with docs

kdschlosser:
Check the commit messages. It is explained in those what has been done.


kisvegabor:
Looks great! Special thanks for `check` / `uncheck` in the roadmap. Now we can sue RST for it too.

kdschlosser:
Yessir. If there is something that needs to be done  that has to be a way. Welp, that is the way to do it. It looks good too without being too hacky

kdschlosser:
You can now see some of the importance of changing over to RST. It really wakes up Sphinx and what it is able to do. I have been chatting with the author of the sphinx plugin that parses the doxygen code. Right now the plugin doesn't evaluate the incoming text from doxygen so using sphinx syntax doesn't currently work. That is going to change. It would be nice to have the clickable links for those things as well. I haven't dug my heels into it yet but I cannot imagine that it cannot be done and neither does the plugin author. At some point I will start messing about with it.

kisvegabor:
Ouch, GitHub doesn't show the checkbox from rst files: https://github.com/lvgl/lvgl/blob/master/docs/ROADMAP.rst

It seems it works only in md files.

kdschlosser:
I could write a script that would read the original ROADMAP.md file and generate a ROADMAP.rst file that can be used in the documentation. more work to do but not a huge deal to do.

Is there a script that automatically generates the roadmap file? If there isn't one one can be made if you used the milestones feature in github I can write a script that would be able to pull that information to be able to build both a github compatible roadmap and also a sphinx compatible one (at documentation compile time)

kisvegabor:
Now I maintain the roadmap by hand. It's the most flexible now.  We can consider other alternatives later but if you already have the scripts ready could you make it convert the md to rst on the fly?

kdschlosser:
I think you will like the PR I just submitted.  #4259

kdschlosser:
https://github.com/kdschlosser/lvgl/blob/roadmap/docs/ROADMAP.rst

Issue #4218, opened 2023-05-12
Map css cubic-bezier(x1, y1, x2, y2) to lv_bezier3 parameters


XuNeo:
### Introduce the problem

Normally in css, the animation path is specified via `cubic-bezier(x1, y1, x2, y2)` and the UI designer also tends to use it.
I cannot figure out how to map the x1, y1, x2, y2 parameter to u0,u1,u2,u3 in lv_bezier3 function. 

For example, `ease-in-out` is specified by `cubic-bezier(0.42, 0.0, 0.58, 1.0)`, and `lv_bezier3` uses equilent(shift 10bits) parameter (0, 0.0488, 0.9297, 1).

Is there an equation to map those parameters? 

Thanks.

kisvegabor:
According to this [comment](https://stackoverflow.com/a/21962457/1999969):
```
x(t) = (1.0 - t^2) * p0.x + 2.0 * (1 - t) * t * p1.x + t^2 * p2.x;
y(t) = (1.0 - t^2) * p0.y + 2.0 * (1 - t) * t * p1.y + t^2 * p2.y;
```

Here the problem is that we increment `t` gradually and we get `x` and `y`. And `x(t)` depends on only `p.x` and `y(t)` only on `p.y`. But what we need is `y(x)`. So you move on the x axis (time in case of an animation) and get the y coordinates (the value of the animation).  

![image](https://github.com/lvgl/lvgl/assets/7599318/beaa4706-e51f-40da-a10e-5a8edf62fca7)

However, it's not so simple to rearrange `x(t)` to get `t` for each gradually increasing `x` and than calculate `y(t)`. To be honest I don't know how and what CSS calculates but it seems to me the simple bezier equations can't be used for animations as they are.  When I implemented `lv_bezier3` I found that there is a version where the x cooridnate of the control points are fixed at 0.33 and 0.66. This way we really have only a `y(t)` function.

Having a real bezier curve animation would be really great, so if you have any ideas, I'd highly appreciate it :slightly_smiling_face:

kisvegabor:
[Useful reading](https://math.stackexchange.com/questions/841169/fastest-way-to-obtain-the-parametric-value-t-of-a-bezier-curve-for-a-given-set), but unfortunately the linked book is not available.

XuNeo:
Thanks for your great comments, I now can understand what the problem is and found an implementation lib for `js` [here](https://github.com/gre/bezier-easing).

[...]

PR #4254, opened 2023-05-28
feat(math): add cubic-bezier function

XuNeo:
### Add basic math support for cubic-bezier

The newly added function is for cubic-bezier function normally used in [css](https://www.cssportal.com/css-cubic-bezier-generator/). 

Discussed here: #4218 
References can be found here:
 - https://opensource.apple.com/source/WebCore/WebCore-955.66/platform/graphics/UnitBezier.h
 - https://github.com/gre/bezier-easing

It has not been integrated with lvgl's animation yet.

This PR firstly needs to discuss whether or not using float calculation instead of fix-point. Below test results on a cortex-m55 MCU shows that float is much faster.

image

Test code:
```c
    int32_t x, x1, y1, x2, y2, y;
    int i = 0;
    while(i++ < 10) {
        x1 = lv_rand(0, 1024);
        x2 = lv_rand(0, 1024);
        y1 = lv_rand(0, 1024);
        y2 = lv_rand(0, 1024);
        double fx1 = x1 / 1024.0;
        double fx2 = x2 / 1024.0;
        double fy1 = y1 / 1024.0;
        double fy2 = y2 / 1024.0;
        double fx, fy;

        int j = 0;
        while(j++ < 10) {
            x = lv_rand(0, 1024);
            uint32_t t_int = lv_tick_get();
            int k = 10000;
            while(k --)
              y = lv_cubic_bezier(x, x1, y1, x2, y2);
            t_int = lv_tick_elaps(t_int);

            fx = x / 1024.0;
            uint32_t t_float = lv_tick_get();
            k = 10000;
            while(k --)
              fy = lv_cubic_bezier_f(fx, fx1, fy1, fx2, fy2);
            t_float = lv_tick_elaps(t_float);

            printf("t_int: %dms, t_float: %dms\n", (int)t_int, (int)t_float);
        }
    }
```

### Checkpoints
 
### DIFF
--- a/src/misc/lv_anim.c
+++ b/src/misc/lv_anim.c
@@ -25,12 +25,16 @@
  *      TYPEDEFS
  **********************/
 
+#define _path_cubic_bezier(a, x1, x2, y1, y2) lv_anim_path_cubic_bezier(a, (x1) * 1024, (y1) * 1024, (x2) * 1024, (y2) * 1024)
+
 /**********************
  *  STATIC PROTOTYPES
  **********************/
 static void anim_timer(lv_timer_t * param);
 static void anim_mark_list_change(void);
 static void anim_ready_handler(lv_anim_t * a);
+static int32_t lv_anim_path_cubic_bezier(const lv_anim_t * a, int32_t x1,
+                                         int32_t y1, int32_t x2, int32_t y2);
 
 /**********************
  *  STATIC VARIABLES
@@ -221,58 +225,22 @@ int32_t lv_anim_path_linear(const lv_anim_t * a)
 
 int32_t lv_anim_path_ease_in(const lv_anim_t * a)
 {
-    /*Calculate the current step*/
-    uint32_t t = lv_map(a->act_time, 0, a->time, 0, LV_BEZIER_VAL_MAX);
-    int32_t step = lv_bezier3(t, 0, 50, 100, LV_BEZIER_VAL_MAX);
-
-    int32_t new_value;
-    new_value = step * (a->end_value - a->start_value);
-    new_value = new_value >> LV_BEZIER_VAL_SHIFT;
-    new_value += a->start_value;
-
-    return new_value;
+    return _path_cubic_bezier(a, 0.42, 0.0, 1.0, 1.0);
 }
 
 int32_t lv_anim_path_ease_out(const lv_anim_t * a)
 {
-    /*Calculate the current step*/
-    uint32_t t = lv_map(a->act_time, 0, a->time, 0, LV_BEZIER_VAL_MAX);
-    int32_t step = lv_bezier3(t, 0, 900, 950, LV_BEZIER_VAL_MAX);
-
-    int32_t new_value;
-    new_value = step * (a->end_value - a->start_value);
-    new_value = new_value >> LV_BEZIER_VAL_SHIFT;
-    new_value += a->start_value;
-
-    return new_value;
+    return _path_cubic_bezier(a, 0.0, 0.0, 0.58, 1.0);
 }
 
 int32_t lv_anim_path_ease_in_out(const lv_anim_t * a)
 {
-    /*Calculate the current step*/
-    uint32_t t = lv_map(a->act_time, 0, a->time, 0, LV_BEZIER_VAL_MAX);
-    int32_t step = lv_bezier3(t, 0, 50, 952, LV_BEZIER_VAL_MAX);
-
-    int32_t new_value;
-    new_value = step * (a->end_value - a->start_value);
-    new_value = new_value >> LV_BEZIER_VAL_SHIFT;
-    new_value += a->start_value;
-
-    return new_value;
+    return _path_cubic_bezier(a, 0.42, 0.0, 0.58, 1.0);
 }
 
 int32_t lv_anim_path_overshoot(const lv_anim_t * a)
 {
-    /*Calculate the current step*/
-    uint32_t t = lv_map(a->act_time, 0, a->time, 0, LV_BEZIER_VAL_MAX);
-    int32_t step = lv_bezier3(t, 0, 1000, 1300, LV_BEZIER_VAL_MAX);
-
-    int32_t new_value;
-    new_value = step * (a->end_value - a->start_value);
-    new_value = new_value >> LV_BEZIER_VAL_SHIFT;
-    new_value += a->start_value;
-
-    return new_value;
+    return lv_anim_path_cubic_bezier(a, 341, 0, 683, 1300);
 }
 
 int32_t lv_anim_path_bounce(const lv_anim_t * a)
@@ -462,4 +430,18 @@ static void anim_mark_list_change(void)
         lv_timer_pause(_lv_anim_tmr);
     else
         lv_timer_resume(_lv_anim_tmr);
+}
+
+static int32_t lv_anim_path_cubic_bezier(const lv_anim_t * a, int32_t x1, int32_t y1, int32_t x2, int32_t y2)
+{
+    /*Calculate the current step*/
+    uint32_t t = lv_map(a->act_time, 0, a->time, 0, LV_BEZIER_VAL_MAX);
+    int32_t step = lv_cubic_bezier(t, x1, y1, x2, y2);
+
+    int32_t new_value;
+    new_value = step * (a->end_value - a->start_value);
+    new_value = new_value >> LV_BEZIER_VAL_SHIFT;
+    new_value += a->start_value;
+
+    return new_value;
 }
--- a/src/misc/lv_math.c
+++ b/src/misc/lv_math.c
@@ -16,8 +16,8 @@
  *      TYPEDEFS
  **********************/
 
-#define NEWTON_ITERATIONS       8
-#define CUBIC_PRECISION_BITS    20 /* 10 or 20 bits recommended, int64_t calculation is used for 20bit precision */
+#define CUBIC_NEWTON_ITERATIONS     8
+#define CUBIC_PRECISION_BITS        10 /* 10 or 14 bits recommended, int64_t calculation is used for >14bit precision */
 
 #if CUBIC_PRECISION_BITS < 10 || CUBIC_PRECISION_BITS > 20
     #error "cubic precision bits should be in range of [10, 20] for 32bit/64bit calculations."
@@ -103,12 +103,6 @@ uint32_t lv_bezier3(uint32_t t, uint32_t u0, uint32_t u1, uint32_t u2, uint32_t 
     return v1 + v2 + v3 + v4;
 }
 
-static float do_cubic_bezier_f(float t, float a, float b, float c)
-{
-    /*a*t^3 + b*t^2 + c*t*/
-    return ((a * t + b) * t + c) * t;
-}
-
 /**
  * cubic-bezier Reference:
  *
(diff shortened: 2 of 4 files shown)


kisvegabor:
Wow, you did it! Very nice!

What is the difference in performance if `CUBIC_PRECISION_BITS` is `10` or `20`?

XuNeo:
If `CUBIC_PRECISION_BITS ` is set to 10, there's chance result error exceeds 30%. I written a test case which randoms all input parameter, and compare results with float version. 

In terms of speed, `float` version looks still faster.

```
t_int: 24ms, t_float: 13ms
t_int: 11ms, t_float: 10ms
t_int: 17ms, t_float: 13ms
t_int: 10ms, t_float: 10ms
t_int: 10ms, t_float: 10ms
t_int: 11ms, t_float: 9ms
t_int: 10ms, t_float: 10ms
t_int: 17ms, t_float: 13ms
t_int: 17ms, t_float: 13ms
t_int: 24ms, t_float: 15ms
t_int: 11ms, t_float: 10ms
t_int: 17ms, t_float: 13ms
t_int: 17ms, t_float: 12ms
t_int: 17ms, t_float: 13ms
t_int: 17ms, t_float: 12ms
t_int: 17ms, t_float: 13ms
t_int: 17ms, t_float: 13ms
t_int: 11ms, t_float: 9ms
t_int: 17ms, t_float: 12ms
t_int: 17ms, t_float: 13ms
t_int: 10ms, t_float: 13ms
t_int: 11ms, t_float: 9ms
t_int: 10ms, t_float: 10ms
t_int: 17ms, t_float: 13ms
t_int: 10ms, t_float: 12ms
t_int: 34ms, t_float: 13ms
t_int: 24ms, t_float: 15ms
t_int: 41ms, t_float: 16ms
t_int: 17ms, t_float: 12ms
t_int: 10ms, t_float: 10ms
t_int: 11ms, t_float: 9ms
t_int: 11ms, t_float: 13ms
t_int: 10ms, t_float: 10ms
t_int: 4ms, t_float: 7ms
t_int: 4ms, t_float: 10ms
t_int: 11ms, t_float: 10ms
t_int: 11ms, t_float: 10ms
t_int: 11ms, t_float: 9ms
t_int: 11ms, t_float: 10ms
t_int: 4ms, t_float: 7ms
t_int: 11ms, t_float: 9ms
t_int: 11ms, t_float: 9ms
t_int: 17ms, t_float: 13ms
t_int: 11ms, t_float: 9ms
t_int: 17ms, t_float: 13ms
t_int: 11ms, t_float: 9ms
t_int: 18ms, t_float: 12ms
t_int: 5ms, t_float: 9ms
t_int: 11ms, t_float: 9ms
t_int: 17ms, t_float: 12ms
t_int: 24ms, t_float: 13ms
t_int: 10ms, t_float: 10ms
t_int: 17ms, t_float: 13ms
t_int: 10ms, t_float: 10ms
t_int: 11ms, t_float: 10ms
t_int: 17ms, t_float: 12ms
t_int: 11ms, t_float: 10ms
t_int: 17ms, t_float: 13ms
t_int: 17ms, t_float: 12ms
t_int: 11ms, t_float: 10ms
t_int: 17ms, t_float: 13ms
t_int: 10ms, t_float: 10ms
t_int: 17ms, t_float: 13ms
t_int: 17ms, t_float: 13ms
t_int: 11ms, t_float: 12ms
t_int: 17ms, t_float: 12ms
t_int: 17ms, t_float: 12ms
t_int: 17ms, t_float: 13ms
t_int: 10ms, t_float: 10ms
t_int: 17ms, t_float: 13ms
t_int: 17ms, t_float: 13ms
t_int: 29ms, t_float: 18ms
t_int: 17ms, t_float: 13ms
t_int: 30ms, t_float: 17ms
t_int: 30ms, t_float: 18ms
t_int: 30ms, t_float: 17ms
t_int: 17ms, t_float: 13ms
t_int: 11ms, t_float: 10ms
t_int: 23ms, t_float: 13ms
t_int: 17ms, t_float: 12ms
t_int: 23ms, t_float: 16ms
t_int: 37ms, t_float: 18ms
```

kisvegabor:
It seems with 10 bit precision the performance is quite the same. I've slightly modified your test code to check the error and print the sum time with int and float:
```c

  int32_t t_int_sum = 0;
  int32_t t_float_sum = 0;
  int32_t x, x1, y1, x2, y2, y;
  int i = 0;
  while(i++ < 10) {
      x1 = lv_rand(0, 1024);
      x2 = lv_rand(0, 1024);
      y1 = lv_rand(0, 1024);
      y2 = lv_rand(0, 1024);
      double fx1 = x1 / 1024.0;
      double fx2 = x2 / 1024.0;
      double fy1 = y1 / 1024.0;
      double fy2 = y2 / 1024.0;
      double fx, fy;

      int j = 0;
      while(j++ < 100) {
          x = lv_rand(0, 1024);
          uint32_t t_int = lv_tick_get();
          int k = 100000;
          while(k --)
              y = lv_cubic_bezier(x, x1, y1, x2, y2);
          t_int = lv_tick_elaps(t_int);
          t_int_sum += t_int;

          fx = x / 1024.0;
          uint32_t t_float = lv_tick_get();
          k = 100000;
          while(k --)
              fy = lv_cubic_bezier_f(fx, fx1, fy1, fx2, fy2);
          t_float = lv_tick_elaps(t_float);
          t_float_sum += t_float;
          
          //Added by kisvegabor
          int32_t err = (int32_t)(fy*1024.0) - y;
          if(LV_ABS(err) > 0) {
              printf("err: %d (y:%d)\n", err, y);
              exit(0);
          }

          printf("t_int: %dms, t_float: %dms\n", (int)t_int, (int)t_float);
      }
  }

  printf("t_int_sum: %dms, t_float_slum: %dms\n", (int)t_int_sum, (int)t_float_sum);
``` 

I saw only very little error very very rarely. Can you mention a case for the 30% error? Or is it rather theretical and hard to find in the practice?

I wonder if we can increase the precision by a few bits. It seem we can find the worst case scenarios where we have the largest numbers and set the precision to fit into 32 bits. 

I'd like to avoid float if possible because not all MCUs support it and even is they support it, they might be very slow with that (e.g. ESP, see [here](https://blog.classycode.com/esp32-floating-point-performance-6e9f6f567a69)).

XuNeo:
Iv_rand is not an actual rand so the simple test cannot reveal some cases until we increase the loop count.

```
x, x1, y1, x2, y2: 1021,1013,393,1023,65
/home/neo/projects/luavgl/deps/lvgl/tests/src/test_cases/test_math.c:35:test_math_cubic_bezier_result_should_be_precise:FAIL: Expected 303 to be less than or equal to 300
```

I agree avoiding float benefits low end MCUs. However I also suspect if math used in animation is the actual bottle neck.

Either way, we can still add a configuration like `LV_ENABLE_FLOAT` or `LV_ARCH_SUPPORT_FLOAT` etc. to indicate that it's OK or better to use float for internal calculation, while keep the API to use fixed-point parameter.

kisvegabor:
> Either way, we can still add a configuration like LV_ENABLE_FLOAT or LV_ARCH_SUPPORT_FLOAT etc. to indicate that it's OK or better to use float for internal calculation, while keep the API to use fixed-point parameter.

It's really an option, but I'd like to add if only if it's really required.

>  Expected 303 to be less than or equal to 300

It's 1% error which could be ok. Am I right to assume that, if we could increment the precision to 12 bit the error would be 0?

XuNeo:
> It's really an option, but I'd like to add if only if it's really required.
> 
Sure, let me check the error under normal ease functions instead of those random data.


> It's 1% error which could be ok. Am I right to assume that, if we could increment the precision to 12 bit the error would be 0?

The result is normalized to 10bit, so 300 / 1024 would mean 30% error.  The error won't be 0 but would be lower than 10bit. Actually we can safely increase to 14bit precison without need of 64bit calculation for intermediates.

XuNeo:
300 in my result is the error value, which is set as the threshold for the tests only.

kisvegabor:
I can reproduce that 300 error:
```c
  int32_t x = 1021;
  int32_t x1 = 1013;
  int32_t y1 = 393;
  int32_t x2 = 1023;
  int32_t y2 = 65;

  int32_t y_int = lv_cubic_bezier(x, x1, y1, x2, y2);
  float y_float = lv_cubic_bezier_f(x / 1024.0, x1 / 1024.0, y1 / 1024.0, x2 / 1024.0, y2 / 1024.0);

  printf("int: %d\n", y_int);
  printf("float: %f\n", y_float * 1024);
```

```
int: 711
float: 711.796021
```

Am I doing something wrong?

XuNeo:
You need to change CUBIC_PRECISION_BITS from 20 to 10 in order to see the error.

kisvegabor:
:man_facepalming: Ahhh, sorry. I was sure that I left it on 10 bits :smile: 

Now
```
int: 1015
float: 711.796021
```

With 14 bit precision would the error be 300 / 16 = ~18 or there is a more complicated connection?

kisvegabor:
I've just realized that your current implementation work with any precision bits not only 10 and 20. Splendid!

With 14 I got:
```c
int: 715
float: 711.796021
```

I haven't fully examined the implementation but the random test worked well with 19 bit too even after removing the 64 bit cast from here:
```c
        d = ((int64_t)xs * (1L << CUBIC_PRECISION_BITS)) / d;

```

kisvegabor:
Please don't let this stale 🙂  It would be very sad if haven't finished it. I suggest using the max possible precision which still fits to int32 and add a config option to lv_conf.h to enable floats in general. Eg LV_USE_FLOAT.

What do you think?

XuNeo:
I agree on max precision within 32bit, which is to set CUBIC_PRECISION_BITS to 14. 
The reason I haven't done it is I need to check real word anim path and make sure that 14bit precision is enough.

Sorry for the late update, busy days. I promise check it on this weekend.

kisvegabor:
Okay, thank you. Don't forget to rest on the weekend. :slightly_smiling_face:

src/misc/lv_math.c
```diff
@@ -16,6 +16,13 @@
  *      TYPEDEFS
  **********************/
 
+#define CUBIC_NEWTON_ITERATIONS     8
+#define CUBIC_PRECISION_BITS        10 /* 10 or 14 bits recommended, int64_t calculation is used for >14bit precision */
```
kisvegabor, line 20: Why can't we use 14 bye default?

XuNeo: The provided test in test_math.c shows that for normally used ease function(ease, ease-in, ease-out, ease-in-out), 10bits is still good enough for 0.5% relative error. While 14bit may provides more accuracy, it requires some additional shift operation. Do you recommend it to 14bits?

src/misc/lv_math.c
```diff
@@ -96,6 +103,144 @@ uint32_t lv_bezier3(uint32_t t, uint32_t u0, uint32_t u1, uint32_t u2, uint32_t
...
+    int64_t d; /*slope value at specified t*/
+#else
+    int32_t d;
+#endif
+
+    if(x == 0 || x == 1024) return x;
+
+    /* input is always 10bit precision */
+
+#if CUBIC_PRECISION_BITS != 10
```
XuNeo: Additional shift operation here and end of function. @kisvegabor

src/misc/lv_math.c
kisvegabor: In this case let's go with 10 bits.

kisvegabor:
Just one last thing: could you update the built in easing functions to use the new bezier functions and remove the old implementation? 

I think we can use the cubic versions from [here](https://easings.net/) as a reference.

XuNeo:
> Just one last thing: could you update the built in easing functions to use the new bezier functions and remove the old implementation?
> 
> I think we can use the cubic versions from [here](https://easings.net/) as a reference.

Sure thing.

src/misc/lv_anim.c
```diff
@@ -221,58 +225,22 @@ int32_t lv_anim_path_linear(const lv_anim_t * a)
...
-    uint32_t t = lv_map(a->act_time, 0, a->time, 0, LV_BEZIER_VAL_MAX);
-    int32_t step = lv_bezier3(t, 0, 50, 100, LV_BEZIER_VAL_MAX);
-
-    int32_t new_value;
-    new_value = step * (a->end_value - a->start_value);
-    new_value = new_value >> LV_BEZIER_VAL_SHIFT;
-    new_value += a->start_value;
-
-    return new_value;
+    return _path_cubic_bezier(a, 0.42, 0.0, 1.0, 1.0);
```
kisvegabor, line 228: I see that these floats are resolved in compile time, but for sake of clarity would you mind using
```c
return _path_cubic_bezier(a, 430, 0, 1024, 1024);
```

kisvegabor (commented):
Thank you, just a minor comment.

Also thank you for adding a test! :pray:

src/misc/lv_anim.c
XuNeo: Do you prefer using a macro to convert the float values? 

This should be a macro since `LV_BEZIER_VAL_MAX` can be changed theoretically.

```
#define LV_BEZIER_VAL(f) ((int32_t)((f) * LV_BEZIER_VAL_MAX))


int32_t lv_anim_path_ease_in(const lv_anim_t * a)
{
    return lv_anim_path_cubic_bezier(a, LV_BEZIER_VAL(0.42), LV_BEZIER_VAL(0), LV_BEZIER_VAL(1.0), LV_BEZIER_VAL(1));
}

```

kisvegabor: > This should be a macro since LV_BEZIER_VAL_MAX can be changed theoretically.

I see. I think we can safely assume that these are really evaluated in compile time and they won't make the `float` lib added to the binary.

kisvegabor:
I messed up something wit merging, but it should be ok now. I've fixed the failing spinner tests too.

Thank you very much for this amazing update! :rocket:

Issue #4221, opened 2023-05-14
performance monitor frames per second is very inaccurate.


kdschlosser:
### Perform all steps below and tick them with [x]
- [x] Read the [FAQ](https://docs.lvgl.io/master/intro/index.html#faq)
- [x] Check the related part of the [Documentation](https://docs.lvgl.io/)
- [x] Update lvgl to the latest version
- [] Reproduce the issue in a [Simulator](https://docs.lvgl.io/latest/en/html/get-started/pc-simulator.html)

### Describe the bug
the frames per second calculation in the performance monitor produces an incorrect/inaccurate number.

### To Reproduce

I have to add the code to do the FPS calculations to a repo I have for a CPython binding. Once I do that you would have to clone the repo I have for a cpython binding of LVGL compile the binding using the included LVGL and install wxPython then run the wx_python script that is included. This would show you the FPS that LVGL calculates and it would print out to the terminal the real FPS.

### Expected behavior
For the FPS to be correct

### Screenshots or video
This is after I had already made some changes to improve how LVGL is doing the calculations.

![image](https://github.com/lvgl/lvgl/assets/10932297/01d62420-f6a5-40ff-ae4f-cac44c3d5f11)

It was much further off then what you see in the above image.

![image](https://github.com/lvgl/lvgl/assets/10932297/248152cc-09b2-43d1-90bb-39a92e50f7f9)

The printed numbers are the same but LVGL was reporting a way higher number.

The problem is multifold. There is not a singular cause. I will list the issues here.

* assumption that exactly 1000 milliseconds (1 second) has passed between when calculations are done.
* doing elapsed time calculation each time a frame does or does not get rendered.
* frame count gets incremented even when nothing gets drawn to the display

The calculation should be

`fps = 1 / ((stop_time - start_time) / frame count / time_resolution)`

the start time gets set to the current ticks when the calculation is done. The frame count gets reset to zero. The only thing that is updated when the display buffer gets written is the frame count. The frame count should ONLY be updated if there is data that needs to be written to the display. When the timer for the performance monitor expires it doesn't mean that is when the calculation actually gets done. It is when `lv_task_handler` gets called and that is completely dependent on the user. 

[...]

PR #4274, opened 2023-06-11
fix(sysmon): increase FPS sampling rate

FASTSHIFT:

Reduced FPS statistics period from 1000ms to 300ms.

### Checkpoints
 
### DIFF
--- a/src/core/lv_disp_private.h
+++ b/src/core/lv_disp_private.h
@@ -147,8 +147,6 @@ struct _lv_disp_t {
     /*Miscellaneous data*/
     uint32_t last_activity_time;        /**< Last time when there was activity on this display*/
 
-    uint32_t last_render_start_time;
-
     /** OPTIONAL: Called periodically while lvgl waits for operation to be completed.
      * For example flushing or GPU
      * User can execute very simple tasks here or yield the task*/
--- a/src/core/lv_refr.c
+++ b/src/core/lv_refr.c
@@ -235,9 +235,6 @@ void _lv_disp_refr_timer(lv_timer_t * tmr)
     LV_PROFILER_BEGIN;
     REFR_TRACE("begin");
 
-    uint32_t start = lv_tick_get();
-
-
     if(tmr) {
         disp_refr = tmr->user_data;
 #if LV_USE_PERF_MONITOR == 0 && LV_USE_MEM_MONITOR == 0
@@ -268,8 +265,6 @@ void _lv_disp_refr_timer(lv_timer_t * tmr)
     }
 
     lv_disp_send_event(disp_refr, LV_EVENT_REFR_START, NULL);
-
-    disp_refr->last_render_start_time = start;
 
     /*Refresh the screen's layout if required*/
     lv_obj_update_layout(disp_refr->act_scr);
@@ -1089,6 +1084,9 @@ static void call_flush_cb(lv_disp_t * disp, const lv_area_t * area, lv_color_t *
 
     if(disp->draw_ctx->buffer_convert) disp->draw_ctx->buffer_convert(disp->draw_ctx);
 
+    lv_disp_send_event(disp, LV_EVENT_FLUSH_START, &offset_area);
     disp->flush_cb(disp, &offset_area, color_p);
+    lv_disp_send_event(disp, LV_EVENT_FLUSH_FINISH, &offset_area);
+
     LV_PROFILER_END;
 }
--- a/src/misc/lv_event.h
+++ b/src/misc/lv_event.h
@@ -99,6 +99,8 @@ typedef enum {
     LV_EVENT_RESOLUTION_CHANGED,
     LV_EVENT_REFR_START,
     LV_EVENT_REFR_FINISH,
+    LV_EVENT_FLUSH_START,
+    LV_EVENT_FLUSH_FINISH,
 
     _LV_EVENT_LAST,               /** Number of default events*/
 
--- a/src/others/sysmon/lv_sysmon.c
+++ b/src/others/sysmon/lv_sysmon.c
@@ -22,11 +22,14 @@
  *      TYPEDEFS
  **********************/
 typedef struct {
-    lv_disp_t * disp;
     uint32_t    refr_elaps_sum;
     uint32_t    refr_cnt;
+    uint32_t    render_start_time;
     uint32_t    render_elaps_sum;
     uint32_t    render_cnt;
+    uint32_t    flush_start_time;
+    uint32_t    flush_elaps_sum;
+    uint32_t    flush_cnt;
 } perf_info_t;
 
 /**********************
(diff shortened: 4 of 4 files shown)


kisvegabor:
[Here](https://github.com/lvgl/lvgl/issues/4221#issuecomment-1558929192) we were talking about adding some more info to sysmon.

What do you think about adding these?
```
CPU 38%, 30 FPS (12 redraws)
Render: 2ms, Flush: 1ms
```

FASTSHIFT:
> [Here](https://github.com/lvgl/lvgl/issues/4221#issuecomment-1558929192) we were talking about adding some more info to sysmon.
> 
> What do you think about adding these?
> 
> ```
> CPU 38%, 30 FPS (12 redraws)
> Render: 2ms, Flush: 1ms
> ```

I agree, how can we measure the flush time, add a timestamp to the call_flush_cb or throw an event externally? Note that when using DMA+double draw buffer mode, the screen refresh and lvgl rendering are parallel, is this something to consider?

FASTSHIFT:
I added count printing for redraw and refr in frame rate printing:

```bash
sysmon: 59 FPS (redraw: 2 / refr: 18), render 0 ms, CPU 1%
sysmon: 59 FPS (redraw: 0 / refr: 17), render 0 ms, CPU 1%
sysmon: 58 FPS (redraw: 1 / refr: 18), render 0 ms, CPU 2%
sysmon: 60 FPS (redraw: 1 / refr: 18), render 0 ms, CPU 3%
sysmon: 59 FPS (redraw: 1 / refr: 18), render 0 ms, CPU 3%
sysmon: 60 FPS (redraw: 1 / refr: 18), render 0 ms, CPU 2%
sysmon: 59 FPS (redraw: 1 / refr: 18), render 0 ms, CPU 2%
sysmon: 59 FPS (redraw: 1 / refr: 18), render 0 ms, CPU 1%
sysmon: 60 FPS (redraw: 1 / refr: 18), render 0 ms, CPU 1%
```

The reason I didn't display the label on the screen was that displaying too much data would be crowded and ugly on a small resolution device.

kisvegabor:
> I agree, how can we measure the flush time, add a timestamp to the call_flush_cb or throw an event externally?

We could send `FLUSH_START` and `FLUSH_READY` events too where we can measure the flush time.

> Note that when using DMA+double draw buffer mode, the screen refresh and lvgl rendering are parallel, is this something to consider?

Yes, we should consider it too.

----

> The reason I didn't display the label on the screen was that displaying too much data would be crowded and ugly on a small resolution device.

Let reiterate. I think these are the things we should display:
- CPU busy
- render count: can be dropped IMO
- FPS:  avg time between `REFR_START`s (max is limited by the call period the refr timer)
- render time: avg time between `RENDER_START` and `RENDER_END`
- flush time: avg time between `FLUSH_START` and `FLUSH_END`
- total time: avg time between `RENDER_START` and `FLUSH_END` (it might be difficult to know which render start refers to which flush end with 2 draw buffers)


```
CPU 38%, 30 FPS
7 ms (4 | 3)
```

Where the last line means:
```
total time ms (render time | flush time)
```

I think it's not to long.

---

BTW, I think we can remove `disp->last_render_start_time` as it can measured if required in the `RENDER_START` event.

FASTSHIFT:
> ```
> CPU 38%, 30 FPS
> 7 ms (4 | 3)
> ```
> 
> Where the last line means:
> 
> ```
> total time ms (render time | flush time)
> ```
> 
> I think it's not to long.

240x240 resolution, when both `LV_USE_PERF_MONITOR` and `LV_USE_MEM_MONITOR` are enabled, the display will overlap:
image

FASTSHIFT:
> BTW, I think we can remove disp->last_render_start_time as it can measured if required in the RENDER_START event.

Done.

> flush time: avg time between FLUSH_START and FLUSH_END

Done.

I still recommend using LV_LOG output in detail, which is convenient for recording and observation, and now the LOG is like this:
```bash
sysmon: 61 FPS (redraw: 1 / refr: 18), render 1ms (flush 1ms), CPU 12%
sysmon: 61 FPS (redraw: 1 / refr: 19), render 1ms (flush 1ms), CPU 19%
sysmon: 61 FPS (redraw: 1 / refr: 18), render 1ms (flush 1ms), CPU 21%
sysmon: 61 FPS (redraw: 1 / refr: 19), render 1ms (flush 1ms), CPU 21%
sysmon: 61 FPS (redraw: 1 / refr: 18), render 1ms (flush 1ms), CPU 22%
sysmon: 61 FPS (redraw: 1 / refr: 19), render 1ms (flush 1ms), CPU 22%
sysmon: 61 FPS (redraw: 1 / refr: 18), render 1ms (flush 1ms), CPU 24%
sysmon: 61 FPS (redraw: 1 / refr: 19), render 1ms (flush 1ms), CPU 20%
sysmon: 62 FPS (redraw: 1 / refr: 18), render 1ms (flush 1ms), CPU 20%
sysmon: 61 FPS (redraw: 1 / refr: 19), render 1ms (flush 1ms), CPU 24%
```

kisvegabor:
I agree to print a detailed log. It's also helpful to draw a chart from it later.

However, I'd like have simple real time data as well. Let's consider the 240 width as a minimum target where both mem and perf monitor should fit. Actually we need only a little bit of extra space. 

What about removing the "used" word from the mem monitor? 
```
12.3kB, 28%
19% frag.
```

kdschlosser:
@FASTSHIFT 

Would it not be better to capture the current ticks and store that and then increment the frame count when the buffer gets drawn. Keep the existing mechanism that uses the timer and when that callback gets called to do the math to calculate the FPS. Doing it this way would mean it would make no difference when the function gets called to do the calculations becaause the calculation would assume thing.

```
fps = 1000 / ((stop_tick - start_tick) / frame_count)
```

example is...
```
start_tick = 100
stop_tick = 1350
frame_count = 65

fps = 1000 / ((1350 - 100) / 65)
fps = 1000 / (1250 / 65)
fps = 1000 / 19.23076923076923
fps = 52
```
    uint32_t fps = info->refr_elaps_sum ? (1000 * info->frame_cnt / info->refr_elaps_sum) : 0;

Once the calculation has been done then the start time gets set to the current tick and the frame count gets zeroed out. 
can also easily make this integer math right up to the point where the label gets set for the FPS.

doing this would take care of any inconsistencies in when lv_task_handler gets called. With the way it is done currently is it assumed that lv_task_hander is being called in a multiple of exactly 1 second, if it's not that is going to cause the calculation to be skewed.

FASTSHIFT:
> I agree to print a detailed log. It's also helpful to draw a chart from it later.
> 
> However, I'd like have simple real time data as well. Let's consider the 240 width as a minimum target where both mem and perf monitor should fit. Actually we need only a little bit of extra space.
> 
> What about removing the "used" word from the mem monitor?
> 
> ```
> 12.3kB, 28%
> 19% frag.
> ```

All features are done, I think it's time to consider merging.

FASTSHIFT:
LOG printing effect:
```bash
sysmon: 64 FPS (refr: 19 | redraw: 1 | flush: 1), refr 0ms (render 0ms | flush 1ms), CPU 0%
sysmon: 62 FPS (refr: 18 | redraw: 1 | flush: 1), refr 0ms (render 0ms | flush 1ms), CPU 0%
sysmon: 64 FPS (refr: 18 | redraw: 1 | flush: 1), refr 0ms (render 1ms | flush 0ms), CPU 0%
sysmon: 64 FPS (refr: 18 | redraw: 1 | flush: 1), refr 0ms (render 0ms | flush 1ms), CPU 0%
sysmon: 64 FPS (refr: 18 | redraw: 1 | flush: 1), refr 0ms (render 0ms | flush 0ms), CPU 0%
sysmon: 65 FPS (refr: 19 | redraw: 1 | flush: 1), refr 0ms (render 0ms | flush 0ms), CPU 0%
sysmon: 64 FPS (refr: 18 | redraw: 1 | flush: 1), refr 0ms (render 0ms | flush 2ms), CPU 0%
```

UI display effect:
image

kisvegabor:
Cool :sunglasses: 
Thank you so much! Merging!

PR #4276, opened 2023-06-11
fix(stm32): Static function prototypes moved from .h to .c

jacquesdmernok:

The current draw functions have prototype declarations in the .h file. These are not used globally, and resulted in unused function warnings. The static prototypes have been moved to the .c file. 

Two other unused functions which used the "static" keyword, were also changed to use the "LV_STM32_DMA2D_STATIC" define. These are:
-lv_gpu_stm32_dma2d_wait_cb
-lv_draw_stm32_dma2d_img


kisvegabor:
LGTM! Thank you, Jacques.

PR #4280, opened 2023-06-12
fix: del only pinyin that in the textarea

Yangnaan:

presssing a pinyin candidate shall not delete the chinese chars that already in textarea.


kisvegabor:
Thank you!

Issue #4271, opened 2023-06-07
calender focus fails when input group is changed after creation


pm-mic:
### Perform all steps below and tick them with [x]
- [x] Read the [FAQ](https://docs.lvgl.io/master/intro/index.html#faq)
- [x] Check the related part of the [Documentation](https://docs.lvgl.io/)
- [x] Update lvgl to the latest version
- [x] Reproduce the issue in a [Simulator](https://docs.lvgl.io/latest/en/html/get-started/pc-simulator.html)

### Describe the bug
When changing the input group of the calendar after it has been created, the calendar will fail to focus when the associated input group is active.

Because the documentation mentions that the buttonmatrix is the base of the calendar, I also checked with the buttonmatrix. However, the buttonmatrix shows the expected behaviour.

This bug is present both in v8.3 and v9.0.

### To Reproduce
```c
void lv_example_calendar_1(void)
{
    lv_group_t *group = lv_group_create();
    //lv_group_set_default(group);                                                           /* <= this works */
    lv_obj_t  * calendar = lv_calendar_create(lv_scr_act());

    lv_group_add_obj(group, calendar);                                                   /* <= this doesn't work */
    lv_win32_add_all_input_devices_to_group(group);

    lv_obj_set_size(calendar, 185, 185);
    lv_obj_align(calendar, LV_ALIGN_CENTER, 0, 27);
    lv_obj_add_event_cb(calendar, event_handler, LV_EVENT_ALL, NULL);

    lv_calendar_set_today_date(calendar, 2021, 02, 23);
    lv_calendar_set_showed_date(calendar, 2021, 02);

    /*Highlight a few days*/
    static lv_calendar_date_t highlighted_days[3];     
    highlighted_days[0].year = 2021;
    highlighted_days[0].month = 02;
    highlighted_days[0].day = 6;

    highlighted_days[1].year = 2021;
    highlighted_days[1].month = 02;
    highlighted_days[1].day = 11;

    highlighted_days[2].year = 2022;
    highlighted_days[2].month = 02;
    highlighted_days[2].day = 22;

    lv_calendar_set_highlighted_dates(calendar, highlighted_days, 3);

#if LV_USE_CALENDAR_HEADER_DROPDOWN
    lv_calendar_header_dropdown_create(calendar);
#elif LV_USE_CALENDAR_HEADER_ARROW
    lv_calendar_header_arrow_create(calendar);
#endif
    lv_calendar_set_showed_date(calendar, 2021, 10);
[...]

PR #4281, opened 2023-06-13
fix to get button objects from arrow header

pm-mic:

When trying to change the input group of the calendar component we need to get the btnmatrix component with lv_calendar_get_btnmatrix(). 

This PR adds the equivalent calls for the lv_calendar_header_arrow_t to get the next and previous buttons, which are also needed when trying to change the input group of the calendar.


pm-mic:
See https://github.com/lvgl/lvgl/issues/4271#issuecomment-1589823984 for a much easier solution to this issue.

PR #4323, opened 2023-06-27
perf(label): add adaptive redraw text to extend redraw area

FASTSHIFT:

For Italic or other non-typical text, we can add` ext_draw` field in `lv_font_t` to identify whether to expand the redraw area.
For most standard fonts, the extended redraw area is not necessary, and setting `ext_draw` to 0 by default ensures a minimum rendering area to maintain performance.

Before:
S}ADLNA_MH5KFHDF51A2MTF

After:
5%D_`QV~6 $M168EZ3~KHBX

### Checkpoints
 
### DIFF
--- a/src/font/lv_font.h
+++ b/src/font/lv_font.h
@@ -76,6 +76,7 @@ typedef struct _lv_font_t {
     lv_coord_t line_height;         /**< The real line height where any text fits*/
     lv_coord_t base_line;           /**< Base line measured from the top of the line_height*/
     uint8_t subpx  : 2;             /**< An element of `lv_font_subpx_t`*/
+    uint8_t ext_draw : 1;           /**< Italic or other non-typical letter needs to expand the drawing area*/
 
     int8_t underline_position;      /**< Distance between the top of the underline and base line (< 0 means below the base line)*/
     int8_t underline_thickness;     /**< Thickness of the underline*/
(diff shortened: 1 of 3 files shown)


kisvegabor:
Ideally the font converter or font engines could set this value explicitly instead of making LVGL to guess a hopefully good value. In [`lv_font_conv`](https://github.com/lvgl/lv_font_conv/) we could set it, but I'm not sure if it's possible in FreeType. What do you think?

We could add a special `AUTO` value too if the engine can't figure out the exact value explicitly.

FASTSHIFT:
> Ideally the font converter or font engines could set this value explicitly instead of making LVGL to guess a hopefully good value. In [`lv_font_conv`](https://github.com/lvgl/lv_font_conv/) we could set it, but I'm not sure if it's possible in FreeType. What do you think?
> 
> We could add a special value too if the engine can't figure out the exact value explicitly.`AUTO`

I very much agree with letting the font converter give an exact value to put in `ext_draw`, it can use the current empirical value as a default when the custom font renderer cannot provide an exact value.

kisvegabor:
It would be great if `0`, the default value would mean `AUTO` but `0` is also a valid value. 

I think the simplest thing that we can do now is adding a new field in the font converter with `LV_FONT_GYLPH_EXT_DRAW_SIZE_AUTO` value. We can add the actual calculation as next step.

This is [the related part](https://github.com/lvgl/lv_font_conv/blob/master/lib/writers/lvgl/lv_table_head.js) of the font converter.

Could wend a PR there too? Please don't forget to add an `LV_VERSION_CHECK(9,0,0)` guard.

FASTSHIFT:
cc @W-Mai

FASTSHIFT:
Not stale

PR #4326, opened 2023-06-29
fix: Explicit casts to avoid -Wdiscarded-qualifiers warnings

nebkat:

Explicitly cast `const char *`s to `char *`s where needed in order to prevent `-Wdiscarded-qualifiers` warnings.

### DIFF
--- a/src/core/lv_obj.c
+++ b/src/core/lv_obj.c
@@ -149,7 +149,7 @@ void lv_init(void)
     lv_img_cache_set_size(LV_IMG_CACHE_DEF_SIZE);
 #endif
     /*Test if the IDE has UTF-8 encoding*/
-    char * txt = "Á";
+    const char * txt = "Á";
 
     uint8_t * txt_u8 = (uint8_t *)txt;
     if(txt_u8[0] != 0xc3 || txt_u8[1] != 0x81 || txt_u8[2] != 0x00) {
--- a/src/extra/widgets/tabview/lv_tabview.c
+++ b/src/extra/widgets/tabview/lv_tabview.c
@@ -77,15 +77,17 @@ lv_obj_t * lv_tabview_add_tab(lv_obj_t * obj, const char * name)
 
     lv_obj_t * btns = lv_tabview_get_tab_btns(obj);
 
-    const char ** old_map = (const char **)tabview->map;
+    char * tmp = lv_mem_alloc(strlen(name) + 1);
+    strcpy(tmp, name);
+
+    char ** old_map = tabview->map;
     const char ** new_map;
 
     /*top or bottom dir*/
     if(tabview->tab_pos & LV_DIR_VER) {
         new_map = lv_mem_alloc((tab_id + 1) * sizeof(const char *));
         lv_memcpy_small(new_map, old_map, sizeof(const char *) * (tab_id - 1));
-        new_map[tab_id - 1] = lv_mem_alloc(strlen(name) + 1);
-        strcpy((char *)new_map[tab_id - 1], name);
+        new_map[tab_id - 1] = tmp;
         new_map[tab_id] = "";
     }
     /*left or right dir*/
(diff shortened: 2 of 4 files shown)


kisvegabor:
Thank you!

I can't see why but now the MicroPython CI is failing. These changes should matter much in the practice. @amirgon do you have any idea?

amirgon:
> @amirgon do you have any idea?

@kisvegabor The problem was intorduced on https://github.com/lvgl/lvgl/commit/3de61c76af0aed1665e9d61e47b4dd50b9e8bdf1, not in this PR.  
The test already fails on that commit, did you notice?

It is related to casting of the result of `get_param()` to `anim` struct:

https://github.com/lvgl/lvgl/blob/082244b46701d86faec02b2aa90744bd71fd9256/examples/widgets/tabview/lv_example_tabview_2.py#L1-L7

After your change in 3de61c76af0aed1665e9d61e47b4dd50b9e8bdf1, the param is passed as `NULL`, and the casting logic did not handle nulls correctly.
I just fixed the casting logic in the Micropython bindings to handle NULLs, please try to re-run the action and see if it passes now.

---

Regardless of this problem, this PR looks wrong to me.  
Could you explain why it makes sense to remove the constness of string literals?  
String literals are always `const char*`, and when casting string literals to `char*` some compilers will complain: `deprecated conversion from string constant to 'char*'`.   
A better solution is to change the left hand side to be `const char*` so that casting is not needed.

kisvegabor:
> The problem was intorduced on https://github.com/lvgl/lvgl/commit/3de61c76af0aed1665e9d61e47b4dd50b9e8bdf1, not in this PR.
The test already fails on that commit, did you notice?

Actually after 3de61c7 wasn't required so I removed it in https://github.com/lvgl/lvgl/commit/390c880f894cc963542b14e73166a35480d3f48d. So now all test are passing in master.

But I've just realized that this PR is against release/v8.3 and the fix wasn't cherry-picked there. Now release/v8.3 should be fixed too.

> A better solution is to change the left hand side to be const char* so that casting is not needed.

Take look at this as an example:
```c
new_map[tab_id * 2 - 3] = "\n";
new_map[tab_id * 2 - 2] = lv_mem_alloc(strlen(name) + 1);
new_map[tab_id * 2 - 1] = "";
strcpy((char *)new_map[(tab_id * 2) - 2], name);
```
Here we need to store both `const` and `non const` data in `new_map`. If we made it `const char ** new_map` `strcpy` would fail. To make all compilers happy, what we could do is:
```c
const char ** new_map;
...

char * tmp = lv_mem_alloc(strlen(name) + 1);
strcpy(tmp, name);
new_map[tab_id * 2 - 3] = "\n";
new_map[tab_id * 2 - 2] = tmp;
new_map[tab_id * 2 - 1] = "";
```

What do you think?

amirgon:
> ```c
> const char ** new_map;
> ...
> 
> char * tmp = lv_mem_alloc(strlen(name) + 1);
> strcpy(tmp, name);
> new_map[tab_id * 2 - 3] = "\n";
> new_map[tab_id * 2 - 2] = tmp;
> new_map[tab_id * 2 - 1] = "";
> ```
> 
> What do you think?

Yes, this sounds good because it's legal to assign `char*` to `const char*` variable, but not the other way around.  
And no casting is needed.

nebkat:
`lv_checkbox` can be fixed by using a `const char *static_txt` instead of a boolean flag. I didn't go through with that change because I think there are a few other locations where the boolean flag is used that might not be so simple.

As for `lv_tabview` I agree that the `(char *)` cast isn't great, but if the map is made `const` it will still need a cast for the `lv_mem_free` calls so nothing is really gained:

https://github.com/lvgl/lvgl/blob/9491c3ff6d2f8e56b13d8fb493d4b3ee98ef1a4b/src/extra/widgets/tabview/lv_tabview.c#L272-L283
> Passing 'const char *' to parameter of type 'void *' discards qualifiers

Ideally there would be no mixing of constant and allocated strings at all, but I doubt that is possible without breaking changes.

kisvegabor:
> lv_checkbox can be fixed by using a const char *static_txt instead of a boolean flag. I didn't go through with that change because I think there are a few other locations where the boolean flag is used that might not be so simple.

It will require 32 bits instead of 1 bit just to save a cast. I would like to keep the flag is possible.

> As for lv_tabview I agree that the (char *) cast isn't great, but if the map is made const it will still need a cast for the lv_mem_free calls so nothing is really gained:

In v9 the tabview will be reworked and use buttons in the header instead of a button matrix. So we need a solution only in v8.

nebkat:
Ah yes I missed the `:1` on `static_txt`. 

In that case I think this is still the correct solution to remove the warnings for now - we are knowingly relying on unsafe behavior so the cast is the best we can do.

kisvegabor:
> we are knowingly relying on unsafe behavior so the cast is the best we can do.

Yeah, that's what we can do know. Could you update the PR accordingly?

nebkat:
If `tabview->map` is `const char **`, then we get the warnings at the `lv_mem_free(tabview->map[i]);` so I've made some compromises if you wanted to avoid the `(char *)`.

That said, I believe this was incorrect:

> String literals are always const char*, and when casting string literals to char* some compilers will complain: deprecated conversion from string constant to 'char*'.

The warning is only when they are _not_ explicitly cast - so I still think the original commit may have been a cleaner solution.

The `(char *)` cast for `"Check box"` still remains also.

kisvegabor:
>> String literals are always const char*, and when casting string literals to char* some compilers will complain: deprecated conversion from string constant to 'char*'.

> The warning is only when they are not explicitly cast - so I still think the original commit may have been a cleaner solution.

If it really works well with explicit castint then I agree to keep the original solution.

@amirgon, do you know a specific compiler which shows a warning for this?

amirgon:
> @amirgon, do you know a specific compiler which shows a warning for this?

Explicit casting probably prevents this warning.

Still, assigning a string literal to non-const `char*` is unsafe. If anyone ever tries to modify this `char *` the result is an undefined behavior according to the C standard.  
I think there is a design problem in the code. Casting away constness looks like an easy solution, but not the right solution.

kisvegabor:
The real problem here is that we use the button matrix as in an unortodox way in the tabview. In v9, we will use normal buttons in the tabview header, but for v8 I vote for simple casting.  Writing a `const` memory by accident results in immediate crash on the first test, so it's easy to spot.

PR #4330, opened 2023-06-30
fix(example): Remove struct cast

amirgon:
Remove null cast for struct.
Related: https://github.com/lvgl/lv_binding_micropython/commit/53232d3a9febf4c6a9be3dc2f4825a1a2bd0426f


kisvegabor:
Thanks!

PR #4331, opened 2023-07-02
fix(meter): fix mid draw, fix(switch): fix get object area interface

zhbi98:
- Modularize the `meter` drawing `mid` section
- The plot `switch` indicator area using a uniform coordinate acquisition interface `lv_obj_get_content_coords`


kisvegabor:
Hi,

Thank you!
The `switch` part look good, however `meter` will be updated in v9 to be more modular using the still under development [`scale`](https://github.com/lvgl/lvgl/pull/4196) widget.  So please remove the `meter` part  from the PR as the meter widget will be completely reworked or removed in its current form.

zhbi98:
Thank you for your advice and I will do it

PR #4336, opened 2023-07-03
fix(switch): fix get object area interface

zhbi98:
- Draw switch indicator area using a uniform coordinate acquisition interface `lv_obj_get_content_coords`.
- Simplify `knob`  coordinate calculations.


kisvegabor:
Thank you!

PR #4338, opened 2023-07-04
fix: update Kconfig and lv_conf_kconfig.h

PeterBee97:

Fix some build errors found within a Kconfig system.

### Checkpoints
 
### DIFF
--- a/src/font/lv_font_fmt_txt.c
+++ b/src/font/lv_font_fmt_txt.c
@@ -387,6 +387,9 @@ static void decompress(const uint8_t * in, uint8_t * out, lv_coord_t w, lv_coord
         case 4:
             opa_table = opa4_table;
             break;
+        default:
+            LV_LOG_WARN("%d bpp is not handled", bpp);
+            return;
     }
 
     rle_init(in, bpp);
--- a/Kconfig
+++ b/Kconfig
@@ -21,6 +21,8 @@ menu "LVGL configuration"
 
 			config LV_COLOR_DEPTH_32
 				bool "32: ARGB8888"
+			config LV_COLOR_DEPTH_24
+				bool "24: RGB888"
 			config LV_COLOR_DEPTH_16
 				bool "16: RGB565"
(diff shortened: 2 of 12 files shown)


kisvegabor:
Hi,

Thank you so much! I've just merged a huge update to master to support run time rendering color format change, a multi core rendering. As part of it `lv_conf.h` was updated too. Now it's more Kconfig friendly as you can select the STD lib wrapper from a list. See [here](https://github.com/lvgl/lvgl/blob/master/lv_conf_template.h#L33-L40).

Do you have time an liking to update this PR according to the new `lv_conf.h`?

PeterBee97:
Definitely. CL updated.

lv_conf_template.h
```diff
@@ -439,6 +439,8 @@
 
 #define LV_USE_CHECKBOX   1
 
+#define LV_USE_COLORWHEEL 1
```
kisvegabor: The color wheel was removed.

src/font/lv_font_fmt_txt.c
```diff
@@ -385,6 +385,7 @@ static void decompress(const uint8_t * in, uint8_t * out, lv_coord_t w, lv_coord
             opa_table = opa3_table;
             break;
         case 4:
+        default:
```
kisvegabor: As default let's use
```c
LV_LOG_WARN("%d bpp is not handled", bpp);
return;
```

kisvegabor (commented):
Thank you very much! I had only a few minor comments.

src/font/lv_font_fmt_txt.c
PeterBee97: Done

lv_conf_template.h
PeterBee97: Sure, removing other mentions of COLORWHEEL

kisvegabor:
Awesome! Thank you very much!

PR #4345, opened 2023-07-06
Fix: check pinyin_ime->cand_str existd or not.

SwordofMorning:
If pinyin_ime->cand_str does not exist, the buffer will not be filled to avoid unexpected situations.


src/others/ime/lv_ime_pinyin.c
```diff
@@ -836,6 +836,15 @@ static void pinyin_page_proc(lv_obj_t * obj, uint16_t dir)
     uint16_t page_num = pinyin_ime->cand_num / LV_IME_PINYIN_CAND_TEXT_NUM;
     uint16_t sur = pinyin_ime->cand_num % LV_IME_PINYIN_CAND_TEXT_NUM;
 
+    /**
+     * @brief if pinyin_ime->cand_str not existed in dict, then do not fill the buffer.
+     * 
+     * @note If I input `hu`, it is a legal pinyin (registered in the dictionary);
+     * @note if I input `hup`, it is not a legal pinyin (not registered in the dictionary), 
+     * @note then not fill the buffer, but returned directly.
+     */
```
kisvegabor:
```suggestion
    /*if pinyin_ime->cand_str not existed in dict, then do not fill the buffer.
     * - If I input `hu`, it is a legal pinyin (registered in the dictionary);
     * - if I input `hup`, it is not a legal pinyin (not registered in the dictionary), 
     *   then not fill the buffer, but returned directly.
     */
```

kisvegabor:
Hi,

Thank you! I think it'd be great to have a few simple [unit tests](https://github.com/lvgl/lvgl/tree/master/tests) for Pinyin_ime. Would you be interested in adding it?

SwordofMorning:
Hi kisvegabor, I'm not familiar with unit testing for lvgl, but I've used Gtest on C++ programs before. So I need to spend some time to get familiar with it, do I just need to create a .c file containing unit tests for Pinyin_ime under `/tests/src/test_cases/`?

kisvegabor:
> do I just need to create a .c file containing unit tests for Pinyin_ime under /tests/src/test_cases/

Yes, that's it. You can check out the other unit tests as reference but there is no magic. Just add the test files an call the `TEST_ASSERT_`s. 

You can find the list of ASSERTs [here](https://github.com/ThrowTheSwitch/Unity) .

If you are on Linux, you can just call `main.py test` in `lvgl/tests` to run the tests.

SwordofMorning:
I now roughly understand the approach to lvgl unit testing. Now I execute the unit test through `main.py build test` and the following happens:
```
xjt@u16:~/Gogs/OpenSourceProgram/lvgl/tests$ python3 main.py build test


==================================================== ===================
Building: test_sysheap: Test config, system heap, 32 bit color depth
==================================================== ===================



================================
Running tests for test_sysheap
================================
Test project /home/xjt/Gogs/OpenSourceProgram/lvgl/tests/build_test_sysheap
No tests were found!!!


==================================================== =================
Building: test_defheap: Test config, LVGL heap, 32 bit color depth
==================================================== =================



================================
Running tests for test_defheap
================================
Test project /home/xjt/Gogs/OpenSourceProgram/lvgl/tests/build_test_defheap
No tests were found!!!


==================================================== ===
Building: full_32bit: Full config, 32 bit color depth
==================================================== ===



====================================================
Building: 24bit: Normal config, 24 bit color depth
====================================================



==================================================== =
Building: 16bit: Minimal config, 16 bit color depth
==================================================== =
```
I checked the issue log, you mentioned that `LV_BUILD_TEST` is set in the Makefile, how can I enable it?

kisvegabor:
Wao, that's strange. Using only `main.py build test` should work. This [`CMakeList.txt`](https://github.com/lvgl/lvgl/blob/master/tests/CMakeLists.txt) file should collect all the tests.

Which OS do you use?

SwordofMorning:
Ubuntu 16.04

SwordofMorning:
And I set the version of CMake to 3.5 instead of 3.12.

kisvegabor:
Hm, I have no idea. I cc a few people, maybe they can tell more.

cc @embeddedt @pete-pjb  @FASTSHIFT

embeddedt:
The problem might simply be that 16.04 is too old. I don't like to be the one to prematurely deprecate software, but the packages in it are unfortunately 7 years old at this point. Out of interest I tested in a 16.04 Docker container and kept running into error after error while trying to get the CMake script going. I gave up after fixing about 5.

SwordofMorning:
Okay, I can't update my Ubuntu for two reasons. One is that the computer (Intel Nuc) I use only supports system version 16.04; the other is that I am developing embedded Linux, and the compilation environment of this chip is also Used 16.04. So I can't help you add unit tests for ime_pinyin.

kisvegabor:
I see. You can still push the tests and let GitHub's CI run them.

SwordofMorning:
Hello kisvegabor, I made the following changes in this commit (8c893bcdd74699cf5313c0ac1e162037cf15d40d)：
1. I added new dictionary for ime_pinyin, which contains all Chinese characters of GB2312 (it is the most commonly used Chinese characters with more than 6000 characters). But I don't seem to modify lv_conf correctly, if you consider adding this dictionary to lvgl, please help me to modify lv_conf.
2. I added a basic creation unit test for ime_pinyin.

SwordofMorning:
Thank you for your help with CI issues. Below is my take on the default dictionary and cjk_font.

1. As you said, `lv_font_simsun_16_cjk` has only a few Chinese characters, but it may be too radical to replace `lv_font_simsun_16_cjk` with GB2312 characters, because lvgl fonts use bitmaps instead of vector graphics. At the same time, `lv_font_simsun_16_cjk` uses "internal fonts", so packaging fonts into the program may greatly increase the size of the program.
2. It may be good to replace the default dictionary with my newly uploaded dictionary, so that the input method retains as many candidate fonts as possible.

Regarding fonts, I can help write `doc.lvgl`. I will write how to package fonts into binary files (aka "external fonts") and be called by the application, rather than as part of the program. In this way, developers can make the fonts they want to use according to their needs. Not everyone wants to use as much Chinese as possible. If there is no need to input, a small font library is enough for developers to use, while minimizing the size of the program.

kisvegabor:
> It may be good to replace the default dictionary with my newly uploaded dictionary, so that the input method retains as many candidate fonts as possible.

Sounds good. We just need to be sure that the suggestions which doesn't exist in the font, should be skipped. 
With this code we can check if a character exists in the font:
```c
    lv_font_glyph_dsc_t dsc;
    if(lv_font_get_glyph_dsc(font, &dsc, unicode_letter, 0)) ...exists...
```


> Regarding fonts, I can help write doc.lvgl. I will write how to package fonts into binary files (aka "external fonts") and be called by the application, rather than as part of the program. 

We have docs for it [here](https://docs.lvgl.io/8.3/overview/font.html#load-a-font-at-run-time). Is there something missing from here?

SwordofMorning:
> We have docs for it [here](https://docs.lvgl.io/8.3/overview/font.html#load-a-font-at-run-time). Is there something missing from here?

Unlike what doc describes, I usually use the following form to create custom fonts:

```c
/*
*-------------------------------------------------- --------------
* Lvgl Font Tool
*
* Note: use unicode encoding
* Note: This font file is generated by Lvgl Font Tool V0.3
* Author: Ali (qq:617622104)
*-------------------------------------------------- --------------
*/


#include "../../lvgl.h"
#include <stdio.h>
#include <stdlib.h>
#include <string.h>

typedef struct{
     uint16_t min;
     uint16_t max;
     uint8_t bpp;
     uint8_t reserved[3];
}x_header_t;
typedef struct{
     uint32_t pos;
} x_table_t;
typedef struct{
     uint8_t adv_w;
     uint8_t box_w;
     uint8_t box_h;
     int8_t ofs_y;
}glyph_dsc_t;


static x_header_t __g_xbf_hd = {
     .min = 0,
     .max = 0,
     .bpp = 0,
};
static uint8_t __g_font_buf[1488];//If the bin file exists in SPI FLASH, this buff can be used


static uint8_t *__user_font_getdata(int offset, int size){
     //If the font is saved in SPI FLASH, SPIFLASH_Read(__g_font_buf,offset,size);
     //If the font has been loaded into SDRAM, just return the offset address directly, such as: return (uint8_t*)(sdram_fontddr+offset);
     uint32_t br;
     FILE *fp = NULL;
     fp = fopen("/root/Font/xFont_36.bin", "rb");
     if(!fp){
         fprintf(stderr, "can't open file xFont_36.bin");
         exit(1);
     }
     fseek(fp, offset, SEEK_SET);
     fread(__g_font_buf, (uint32_t) size, 1, fp);
     fclose(fp);
     return __g_font_buf;
}


static const uint8_t * __user_font_get_bitmap(const lv_font_t * font, uint32_t unicode_letter) {
     if(__g_xbf_hd.max==0) {
         uint8_t *p = __user_font_getdata(0, sizeof(x_header_t));
         memcpy(&__g_xbf_hd, p, sizeof(x_header_t));
     }
     if( unicode_letter>__g_xbf_hd.max || unicode_letter<__g_xbf_hd.min ) {
         return NULL;
     }
     uint32_t unicode_offset = sizeof(x_header_t)+(unicode_letter-__g_xbf_hd.min)*4;
     uint32_t *p_pos = (uint32_t *)__user_font_getdata(unicode_offset, 4);
     if( p_pos[0] != 0 ) {
         uint32_t pos = p_pos[0];
         glyph_dsc_t * gdsc = (glyph_dsc_t*)__user_font_getdata(pos, sizeof(glyph_dsc_t));
         return __user_font_getdata(pos+sizeof(glyph_dsc_t), gdsc->box_w*gdsc->box_h*__g_xbf_hd.bpp/8);
     }
     return NULL;
}


static bool __user_font_get_glyph_dsc(const lv_font_t * font, lv_font_glyph_dsc_t * dsc_out, uint32_t unicode_letter, uint32_t unicode_letter_next) {
     if(__g_xbf_hd.max==0) {
         uint8_t *p = __user_font_getdata(0, sizeof(x_header_t));
         memcpy(&__g_xbf_hd, p, sizeof(x_header_t));
     }
     if( unicode_letter>__g_xbf_hd.max || unicode_letter<__g_xbf_hd.min ) {
         return NULL;
     }
     uint32_t unicode_offset = sizeof(x_header_t)+(unicode_letter-__g_xbf_hd.min)*4;
     uint32_t *p_pos = (uint32_t *)__user_font_getdata(unicode_offset, 4);
     if( p_pos[0] != 0 ) {
         glyph_dsc_t * gdsc = (glyph_dsc_t*)__user_font_getdata(p_pos[0], sizeof(glyph_dsc_t));
         dsc_out->adv_w = gdsc->adv_w;
         dsc_out->box_h = gdsc->box_h;
         dsc_out->box_w = gdsc->box_w;
         dsc_out->ofs_x = 0;
         dsc_out->ofs_y = gdsc->ofs_y;
         dsc_out->bpp = __g_xbf_hd.bpp;
         return true;
     }
     return false;
}


//Siyuan Blackbody CN Medium,Medium,36
// font height: 70
//XBF font, external bin file
const lv_font_t xFont_36 = {
     .get_glyph_bitmap = __user_font_get_bitmap,
     .get_glyph_dsc = __user_font_get_glyph_dsc,
     .line_height = 70,
     .base_line = 0,
};

//end of file
```

This is similar to what is described in [Add a new font engine](https://docs.lvgl.io/8.3/overview/font.html#add-a-new-font-engine). Do you think it is necessary to update the documentation? Maybe we can give an example to put in `examples.html`?

kisvegabor:
Ah, I see. Yes, a custom font engine example would be amazing! But let's handle it separately. Could you send a new pull request with it?

SwordofMorning:
Sure. Before closing this PR, I need to confirm the work content of my next PR. Which method do you think is more suitable?

1. Add a part of description for [Font Engine](https://github.com/lvgl/lvgl/blob/master/docs/overview/font.rst#add-a-new-font-engine).
2. Create a new "Chinese localization" page in the `/others` directory, which contains two parts: `ime_pinyin` and `Chinese font`. Also add a hyperlink from  [`/overview/font` page](https://github.com/lvgl/lvgl/blob/master/docs/overview/font.rst#special-fonts)  pointing to `/others/chinese_localization`.

kisvegabor:
I thin kin this PR we should do the followings:
- Keep only your dictionary
- Show only those candidates which exists in the font.

In the new PR:
- Add a part of description for Font Engine. (I'm not sure info you are planning to add there, but we will see :slightly_smiling_face: )
- Add an example with your font engine to `examples/font/lv_example_font_1.c` (There is no font folder yet)

kisvegabor:
> [Replace dict and font check.](https://github.com/lvgl/lvgl/pull/4345/commits/46dfb73dea253d1cf2d0924aaf0845d1afcc50be)

Probably I'm missing something but where is the  "font check" part?

SwordofMorning:
Yes, my description may not be clear, it should be called "syllable check". The original purpose of my submission of this pr is to avoid the following situation: when the pinyin syllable does not exist in the dictionary (map's key), `pinyin_page_proc` will not try to refresh the content of `cand`, which is what line 922 does. If you still try to load the character in `cand`, it will cause an exception because the key of the dict does not have a corresponding value.

If I want to enter a Chinese character, for example: "a -> 啊"; but the Chinese character does not exist in the `font` which I specified, then a placeholder will be displayed. Whether the Chinese characters exist in the `font` is part of the function, which I have not completed, that is, the "font check" part.

I'm sorry for my mistake.

kisvegabor:
I see now, thank you.

> Whether the Chinese characters exist in the font is part of the function, which I have not completed

Once this "if the character exists in the font" part is added we can merge this PR. For reference see [this comment](https://github.com/lvgl/lvgl/pull/4345#issuecomment-1649866963).

SwordofMorning:
Hey kisvegabor,  I'm a little confused about `lv_font_get_glyph_dsc`. I use it to check letter exist in font or not :

```c
lv_font_t * pinyin_ime_font = lv_obj_get_style_text_font(pinyin_ime, LV_PART_MAIN);
lv_font_glyph_dsc_t pinyin_ime_glyph_dsc;

char font_check[4];
font_check[0] = 0x00;
for(uint8_t j = 0; j < 3; j++)
    font_check[j] = pinyin_ime->cand_str[offset + (i * 3) + j];

printf("font check = %x %x %x %x\n", 
    font_check[0], font_check[1], font_check[2], font_check[3]);
if (lv_font_get_glyph_dsc(pinyin_ime_font, &pinyin_ime_glyph_dsc, font_check, '\0'))
{
    printf("exist\n");
}
else
{
    printf("not exist\n");
}
```

For example I want to check the utf8 character of `啊` which is `e9 98 bf`. Whether I pass in `00 e9 98 bf` or `e9 98 bf 00`, `lv_font_get_glyph_dsc` returns false. Is there something I'm missing?

SwordofMorning:
By the way, if I use `啊`'s Unicode `0x0000963F`, it still return false. I use "external bin font", will there be a problem with this?

kisvegabor:
I suggest this to get the Unicode characters from and UTF-8 string.
```c
uint32_t i = 0;
while(txt[i]) {
    uint32_t unicode_letter = _lv_txt_encoded_next(txt, &i);
    if (lv_font_get_glyph_dsc(pinyin_ime_font, &pinyin_ime_glyph_dsc, unicode_letter, '\0')) ...
}
```

kisvegabor:
Hi,

Are there any news here? :slightly_smiling_face:

PR #4348, opened 2023-07-07
changes a couple of things to get the CPython binding working.

kdschlosser:
These are pretty minor changes. One is removing the declaration for the `_lv_font_clean_up_fmt_txt` function. The function wasn't defined which caused linking problems.

I also moved a couple of structures that were defined in source files and put them into the header files. If this is an issue let me know, I will see if I can manage to work around them not being exposed. I am not sure if ctypes will play nice with me if it is unable to determine the size of the structures. Because ctypes is disconnected so to speak from the structures in C code The only way for me to show their existence is if they can be seen in the header files. That includes the fields in the structures. This is if the structure is used aas a type for aa field in a public structure or as a type in a public function.

They are small changes, I hope it's not a big deal to make them. I managed to code around the rest of the issues I was having before.

The code generation script is too specific to LVGL for my liking, so it will need maintenance due to it having specific parts of it coded specifically to handle LVGL, if those things change it could break. Not a huge deal.

### DIFF
--- a/src/misc/lv_anim_timeline.h
+++ b/src/misc/lv_anim_timeline.h
@@ -23,7 +23,7 @@ extern "C" {
  *      TYPEDEFS
  **********************/
 
- /*Data of anim_timeline_dsc*/
+/*Data of anim_timeline_dsc*/
 typedef struct {
     lv_anim_t anim;
     uint32_t start_time;
--- a/src/misc/lv_anim_timeline.h
+++ b/src/misc/lv_anim_timeline.h
@@ -23,9 +23,18 @@ extern "C" {
  *      TYPEDEFS
  **********************/
 
-struct _lv_anim_timeline_t;
+ /*Data of anim_timeline_dsc*/
+typedef struct {
+    lv_anim_t anim;
+    uint32_t start_time;
+} lv_anim_timeline_dsc_t;
 
-typedef struct _lv_anim_timeline_t lv_anim_timeline_t;
+/*Data of anim_timeline*/
+typedef struct {
+    lv_anim_timeline_dsc_t * anim_dsc;  /**< Dynamically allocated anim dsc array*/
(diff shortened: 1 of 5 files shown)


src/misc/lv_event.h
```diff
@@ -25,9 +25,15 @@ extern "C" {
 /**********************
  *      TYPEDEFS
  **********************/
+typedef void (*lv_event_cb_t)(struct _lv_event_t * e);
```
kisvegabor, line 30: To solve the error:
```suggestion
struct _lv_event_t;

typedef void (*lv_event_cb_t)(struct _lv_event_t * e);
```

src/misc/lv_anim_timeline.h
```diff
@@ -23,9 +23,18 @@ extern "C" {
  *      TYPEDEFS
  **********************/
 
-struct _lv_anim_timeline_t;
-
-typedef struct _lv_anim_timeline_t lv_anim_timeline_t;
+ /*Data of anim_timeline_dsc*/
```
kisvegabor:
```suggestion
/*Data of anim_timeline_dsc*/
```

kdschlosser:
OK so I am not sure what the "resolved" means. LOL

kisvegabor:
Thanks, all looks good. I've fixed a simple error, that's why you see it resolved. :slightly_smiling_face:

kdschlosser:
Funny tho it works without declaring the struct ahead of time like that.

Is that the reason why the compiler warning existed?

kisvegabor:
It's really strange that it worked for you. I thought it should fail to compile. Sometimes miracles happen :slightly_smiling_face:

kdschlosser:
Depends on of you have compiler warning set to become errors I guess. with MSVC it might be a warning. who knows. It did compile tho. MSVC is so far off from C** compliance it's crazy. The hoops you have to jump through just to get functions exported where gcc exports all function right out of the box.

PR #4352, opened 2023-07-09
Specify color format value in enum

XuNeo:
### Specify color format value in enum

Color format's value is used for binary image file generation. It's important to keep it consistent. 
- Use enum makes it hard to track its real value and tends to change easily. 
- `enum` size is platform dependent, but color format is supposed to be 8bit exact.  

~This PR changes it to macro,~ makes the CF value compatible with v8's definition, so the bin file generated and the tool can still work in V9, at least for now.

I have changed the macro back to enum but specify enum value to make micropython binding pass.

### Checkpoints
 
### DIFF
--- a/src/draw/lv_img_buf.h
+++ b/src/draw/lv_img_buf.h
@@ -61,13 +61,11 @@ typedef struct {
 } lv_img_header_t;
 #else
 typedef struct {
-
     uint32_t cf : 5;          /*Color format: See `lv_color_format_t`*/
     uint32_t always_zero : 3; /*It the upper bits of the first byte. Always zero to look like a
                                  non-printable character*/
 
-    uint32_t reserved : 1; /*Reserved to be used later*/
-    uint32_t chroma_keyed : 2; /*Reserved to be used later*/
+    uint32_t reserved : 2; /*Reserved to be used later*/
 
     uint32_t w : 11; /*Width of the image map*/
     uint32_t h : 11; /*Height of the image map*/
--- a/src/misc/lv_color.h
+++ b/src/misc/lv_color.h
@@ -99,46 +99,59 @@ typedef struct {
 typedef uint8_t lv_opa_t;
 //! @endcond
 
+enum _lv_color_format_t {
+    LV_COLOR_FORMAT_UNKNOWN           = 0,
 
-typedef enum {
-    LV_COLOR_FORMAT_UNKNOWN,
+    LV_COLOR_FORMAT_RAW               = 0x01,
+    LV_COLOR_FORMAT_RAW_ALPHA         = 0x02,
 
     /*<=1 byte (+alpha) formats*/
-    LV_COLOR_FORMAT_L8,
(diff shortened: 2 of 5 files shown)


src/draw/lv_img_buf.h
```diff
@@ -61,13 +61,12 @@ typedef struct {
...
 #else
 typedef struct {
-
     uint32_t cf : 5;          /*Color format: See `lv_color_format_t`*/
     uint32_t always_zero : 3; /*It the upper bits of the first byte. Always zero to look like a
                                  non-printable character*/
 
     uint32_t reserved : 1; /*Reserved to be used later*/
-    uint32_t chroma_keyed : 2; /*Reserved to be used later*/
+    uint32_t chroma_keyed : 1; /*Reserved to be used later*/
```
kisvegabor: v9 won't support chroma keyed images so this part can be removed.

src/misc/lv_color.h
```diff
@@ -99,46 +99,57 @@ typedef struct {
...
-    LV_COLOR_FORMAT_I4,
-    LV_COLOR_FORMAT_I8,
+    LV_COLOR_FORMAT_L8                = 0x06,
+    LV_COLOR_FORMAT_I1                = 0x07,
+    LV_COLOR_FORMAT_I2                = 0x08,
+    LV_COLOR_FORMAT_I4                = 0x09,
+    LV_COLOR_FORMAT_I8                = 0x0A,
+    LV_COLOR_FORMAT_A1                = 0x0B,
+    LV_COLOR_FORMAT_A2                = 0x0C,
+    LV_COLOR_FORMAT_A4                = 0x0D,
```
kisvegabor: I was planning to remove the A1/A2/A4 image formats as the very same thing can be achieved by INDEXED1/2/4 formats with only a little bit of overhead. 

So could you remove A1/2/4?

src/draw/lv_img_buf.h
XuNeo: Done

src/misc/lv_color.h
XuNeo: I have removed A1/2/4.

So in v9 sw renderer won't support A1/2/4, but what about GPUs? Maybe we could define the common CFs in lvgl and let GPUs to handle the custom CF only. 

Would it be better if we add comment to A1/2/4 that they are not supported by sw renderer?

kisvegabor: > So in v9 sw renderer won't support A1/2/4, but what about GPUs? Maybe we could define the common CFs in lvgl and let GPUs to handle the custom CF only.

I was thinking about it as well. In this case we need to clearly indicate with comments that which color formats are support by SW and what require GPU.

So, please add A1/2/4 back again, but add comment about it's not support by the software renderer.

XuNeo: Done

kisvegabor:
Looks good, thank you!

PR #4362, opened 2023-07-11
Feat/freertos task notify

nicusorcitu:

Freertos updates.

### Checkpoints
 
### DIFF
--- a/src/draw/sw/lv_draw_sw.c
+++ b/src/draw/sw/lv_draw_sw.c
@@ -71,7 +71,6 @@ void lv_draw_sw_init(void)
         draw_sw_unit->idx = i;
 
 #if LV_USE_OS
-        lv_thread_sync_init(&draw_sw_unit->sync);
         lv_thread_init(&draw_sw_unit->thread, LV_THREAD_PRIO_HIGH, render_thread_cb, 8 * 1024, draw_sw_unit);
 #endif
     }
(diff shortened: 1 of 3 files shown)


kisvegabor:
Did the performance change due to this update?

nicusorcitu:
I would leave the two implementations in place so that further tests can clarify which one is better. I can say that task notification is much cleaner and faster. But maybe its speed is not so significant to really be seen in the FPS value. (which btw, the FPS value is calculated at the ms unit time). E.G. if a frame took 12,7 ms with semaphore but only 12,2 ms with task notify, the FPS result will be the same.

kisvegabor:
> E.G. if a frame took 12,7 ms with semaphore but only 12,2 ms with task notify, the FPS result will be the same.

True, we measure only milliseconds here, so both will be 12 ms, resulting in 1000 / 12 = 83 FPS.

---

So can I merge it now?

nicusorcitu:
Yes, if you agree with the changes from lv_draw_sw.c

I need to run the sync initialization from the thread itself because I need its own handle.

nicusorcitu:
I just ran some tests. I let CPU onlu, 3 draw threads.
1st run with semaphores and the 2nd run with task notification. The difference is significant. Please see attached results.

[3sw_draw_semaphores.txt](https://github.com/lvgl/lvgl/files/12038511/3sw_draw_semaphores.txt)
[3sw_draw_task_notify.txt](https://github.com/lvgl/lvgl/files/12038512/3sw_draw_task_notify.txt)

nicusorcitu:
BTW, I would not squash the commits but merge them as are in the PR.

kisvegabor:
It's good to know that the OS layer implementation has such a large impact.

PR #4363, opened 2023-07-12
Refresh test code. ***DO NOT MERGE***

kdschlosser:
I changed the organization of the code so double buffering will actually work. I also made some changes to the performance monitor so a true FPS fill be given, it will not count frames when there is no work to be done.

I have only tested this on my desktop PC with the default refresh timer set to 1 millisecond Test version and also baaseline version has the changed made to the performance monitor so they would not count frames where no work has been done. This gives a true indication of what kind of difference there would be performance wise between the original refresh setup and the one I modified. Using a slider and having it change value every 1 millisecond using a frame buffer that is the size of the display (480 x 320 x 32) the code change I made gave a boost of 20% which is a pretty large increase.

I would like someone to test it in a read world setup on an MCU using DMA memory and double buffering. I believe there is going to be an even larger increase in performance when this is done. The default refresh timer value for this test should be decreased to 1 millisecond to ensure the timer value is not limiting the speed in which an update is able to occur and the slider value should only be changed once a millisecond.

DO NOT MERGE this PR, it is only to be used as a proof of concept.

### DIFF
--- a/src/core/lv_refr.c
+++ b/src/core/lv_refr.c
@@ -890,22 +890,32 @@ static void draw_buf_flush(lv_disp_t * disp)
 
     bool flushing_last = disp->flushing_last;
 
+    // This code was moved from _lv_disp_refr_timer as it seem wrong to have a
+    // stall of the program directly after the fluck callback has been called.
+
     if(lv_disp_is_double_buffered(disp)) {
         while(disp->flushing);
         /*If there are 2 buffers swap them. With direct mode swap only on the last area*/
 
-        void * buf_on_screen = disp_refr->draw_buf_act;
-        void * buf_off_screen = disp_refr->draw_buf_act == disp_refr->draw_buf_1 ? disp_refr->draw_buf_2 : disp_refr->draw_buf_1;
+        // This might be wrong and some adjustments might need to be made. I
+        // understaand the buffer is being coppied and I don't think it matters if
+        // it happens before the flush function is called or after because after
+        // all it is a copy being made so long as all rendering to the buffer has
+        // been completed before the copy gets made.
+        if (disp_refr->render_mode == LV_DISP_RENDER_MODE_DIRECT) {
+            void * buf_on_screen = disp_refr->draw_buf_act;
+            void * buf_off_screen = disp_refr->draw_buf_act == disp_refr->draw_buf_1 ? disp_refr->draw_buf_2 : disp_refr->draw_buf_1;
 
-        lv_coord_t stride = lv_disp_get_hor_res(disp_refr);
-        uint32_t i;
-        for(i = 0; i < disp_refr->inv_p; i++) {
-            if(disp_refr->inv_area_joined[i]) continue;
-            disp_refr->layer_head->buffer_copy(
-                disp_refr->layer_head,
-                buf_off_screen, stride, &disp_refr->inv_areas[i],
-                buf_on_screen, stride, &disp_refr->inv_areas[i]
-            );
+            lv_coord_t stride = lv_disp_get_hor_res(disp_refr);
+            uint32_t i;
+            for(i = 0; i < disp_refr->inv_p; i++) {
+                if(disp_refr->inv_area_joined[i]) continue;
+                disp_refr->layer_head->buffer_copy(
+                    disp_refr->layer_head,
+                    buf_off_screen, stride, &disp_refr->inv_areas[i],
+                    buf_on_screen, stride, &disp_refr->inv_areas[i]
+                );
+            }
         }
         if(disp->render_mode != LV_DISP_RENDER_MODE_DIRECT || flushing_last) {
             disp->draw_buf_act = disp_refr->draw_buf_act == disp_refr->draw_buf_1 ? disp_refr->draw_buf_2 : disp_refr->draw_buf_1;
--- a/src/core/lv_refr.c
+++ b/src/core/lv_refr.c
@@ -346,6 +346,21 @@ void _lv_disp_refr_timer(lv_timer_t * tmr)
         return;
     }
 
+    /*Do nothing if there is no active screen*/
+    if(disp_refr->act_scr == NULL) {
+        disp_refr->inv_p = 0;
+        LV_LOG_WARN("there is no active screen");
+        return;
+    }
+
+    if (
+        (disp_refr->act_scr->scr_layout_inv == 0) &&
+        (disp_refr->inv_p == 0) &&
+        ((disp_refr->prev_scr == NULL) || (disp_refr->prev_scr->scr_layout_inv))
+    ) {
+        return;
+    }
+
     lv_disp_send_event(disp_refr, LV_EVENT_REFR_START, NULL);
 
     /*Refresh the screen's layout if required*/
@@ -356,15 +371,7 @@ void _lv_disp_refr_timer(lv_timer_t * tmr)
     lv_obj_update_layout(disp_refr->top_layer);
     lv_obj_update_layout(disp_refr->sys_layer);
 
-    /*Do nothing if there is no active screen*/
-    if(disp_refr->act_scr == NULL) {
-        disp_refr->inv_p = 0;
-        LV_LOG_WARN("there is no active screen");
-        goto refr_finish;
-    }
-
     lv_refr_join_area();
-
     refr_invalid_areas();
 
     if(disp_refr->inv_p == 0) goto refr_finish;
@@ -372,30 +379,6 @@ void _lv_disp_refr_timer(lv_timer_t * tmr)
     /*If refresh happened ...*/
     /*Call monitor cb if present*/
     lv_disp_send_event(disp_refr, LV_EVENT_RENDER_READY, NULL);
-
-    if(!lv_disp_is_double_buffered(disp_refr) || disp_refr->render_mode != LV_DISP_RENDER_MODE_DIRECT) goto refr_clean_up;
-
-    /*With double buffered direct mode synchronize the rendered areas to the other buffer*/
-    /*We need to wait for ready here to not mess up the active screen*/
-    while(disp_refr->flushing);
-
-    /*The buffers are already swapped.
-     *So the active buffer is the off screen buffer where LVGL will render*/
-    void * buf_off_screen = disp_refr->draw_buf_act;
-    void * buf_on_screen = disp_refr->draw_buf_act == disp_refr->draw_buf_1
-                           ? disp_refr->draw_buf_2
-                           : disp_refr->draw_buf_1;
-
-    lv_coord_t stride = lv_disp_get_hor_res(disp_refr);
-    uint32_t i;
-    for(i = 0; i < disp_refr->inv_p; i++) {
-        if(disp_refr->inv_area_joined[i]) continue;
-        disp_refr->layer_head->buffer_copy(
-            disp_refr->layer_head,
-            buf_off_screen, stride, &disp_refr->inv_areas[i],
-            buf_on_screen, stride, &disp_refr->inv_areas[i]
-        );
-    }
 
 refr_clean_up:
     lv_memzero(disp_refr->inv_areas, sizeof(disp_refr->inv_areas));
@@ -901,16 +884,35 @@ static void draw_buf_flush(lv_disp_t * disp)
      * and driver is ready to receive the new buffer.
      * If we need to wait here it means that the content of one buffer is being sent to display
      * and other buffer already contains the new rendered image. */
-    if(lv_disp_is_double_buffered(disp)) {
-        while(disp->flushing);
-    }
-
-    disp->flushing = 1;
 
     if(disp->last_area && disp->last_part) disp->flushing_last = 1;
     else disp->flushing_last = 0;
 
     bool flushing_last = disp->flushing_last;
+
+    if(lv_disp_is_double_buffered(disp)) {
+        while(disp->flushing);
+        /*If there are 2 buffers swap them. With direct mode swap only on the last area*/
+
+        void * buf_on_screen = disp_refr->draw_buf_act;
+        void * buf_off_screen = disp_refr->draw_buf_act == disp_refr->draw_buf_1 ? disp_refr->draw_buf_2 : disp_refr->draw_buf_1;
+
+        lv_coord_t stride = lv_disp_get_hor_res(disp_refr);
+        uint32_t i;
+        for(i = 0; i < disp_refr->inv_p; i++) {
+            if(disp_refr->inv_area_joined[i]) continue;
+            disp_refr->layer_head->buffer_copy(
+                disp_refr->layer_head,
+                buf_off_screen, stride, &disp_refr->inv_areas[i],
+                buf_on_screen, stride, &disp_refr->inv_areas[i]
+            );
+        }
+        if(disp->render_mode != LV_DISP_RENDER_MODE_DIRECT || flushing_last) {
+            disp->draw_buf_act = disp_refr->draw_buf_act == disp_refr->draw_buf_1 ? disp_refr->draw_buf_2 : disp_refr->draw_buf_1;
+        }
+    }
+
+    disp->flushing = 1;
 
     if(disp->flush_cb) {
         /*Rotate the buffer to the display's native orientation if necessary*/
@@ -920,15 +922,6 @@ static void draw_buf_flush(lv_disp_t * disp)
         }
         else {
             call_flush_cb(disp, &disp->refreshed_area, layer->buf);
-        }
-    }
-    /*If there are 2 buffers swap them. With direct mode swap only on the last area*/
-    if(lv_disp_is_double_buffered(disp) && (disp->render_mode != LV_DISP_RENDER_MODE_DIRECT || flushing_last)) {
-        if(disp->draw_buf_act == disp->draw_buf_1) {
-            disp->draw_buf_act = disp->draw_buf_2;
-        }
-        else {
-            disp->draw_buf_act = disp->draw_buf_1;
         }
     }
 }
--- a/src/others/sysmon/lv_sysmon.c
+++ b/src/others/sysmon/lv_sysmon.c
@@ -32,6 +32,8 @@ typedef struct {
     uint32_t    flush_start;
     uint32_t    flush_elaps_sum;
     uint32_t    flush_cnt;
+    uint8_t     delay_zero_update_cnt;
+    uint16_t    last_fps;
 } perf_info_t;
 
 /**********************
@@ -131,8 +133,11 @@ static void perf_monitor_disp_event_cb(lv_event_t * e)
             info->refr_start = lv_tick_get();
             break;
         case LV_EVENT_REFR_FINISH:
-            info->refr_elaps_sum += lv_tick_elaps(info->refr_start);
-            info->refr_cnt++;
+            uint32_t elapsed_time = lv_tick_elaps(info->refr_start);
+            if (elapsed_time > 0) {
+                info->refr_elaps_sum += elapsed_time;
+                info->refr_cnt++;
+            }
             break;
         case LV_EVENT_RENDER_START:
             info->render_start = lv_tick_get();
@@ -157,7 +162,6 @@ static void perf_monitor_event_cb(lv_event_t * e)
 {
     lv_obj_t * sysmon = lv_event_get_current_target_obj(e);
     perf_info_t * info = lv_obj_get_user_data(sysmon);
-
     uint32_t fps = info->refr_interval_sum ? (1000 * info->refr_cnt / info->refr_interval_sum) : 0;
     uint32_t cpu = 100 - lv_timer_get_idle();
     uint32_t refr_avg_time = info->refr_cnt ? (info->refr_elaps_sum / info->refr_cnt) : 0;
@@ -181,6 +185,19 @@ static void perf_monitor_event_cb(lv_event_t * e)
            refr_avg_time, render_real_avg_time, flush_avg_time,
            cpu);
 #else
+    if (fps == 0) {
+        info->delay_zero_update_cnt += 1;
+        if (info->delay_zero_update_cnt == 10) {
+            info->delay_zero_update_cnt = 0;
+            info->last_fps = 0;
+        } else {
+            fps = (uint32_t) info->last_fps;
+        }
+    } else {
+        info->delay_zero_update_cnt = 0;
+        info->last_fps = (uint16_t) fps;
+    }
+
     lv_label_set_text_fmt(
         sysmon,
         "%" LV_PRIu32" FPS, %" LV_PRIu32 "%% CPU\n"
--- a/src/others/sysmon/lv_sysmon.c
+++ b/src/others/sysmon/lv_sysmon.c
@@ -133,11 +133,8 @@ static void perf_monitor_disp_event_cb(lv_event_t * e)
             info->refr_start = lv_tick_get();
             break;
         case LV_EVENT_REFR_FINISH:
-            uint32_t elapsed_time = lv_tick_elaps(info->refr_start);
-            if (elapsed_time > 0) {
-                info->refr_elaps_sum += elapsed_time;
-                info->refr_cnt++;
-            }
+            info->refr_elaps_sum += lv_tick_elaps(info->refr_start);
+            info->refr_cnt++;
             break;
         case LV_EVENT_RENDER_START:
             info->render_start = lv_tick_get();
@@ -185,6 +182,14 @@ static void perf_monitor_event_cb(lv_event_t * e)
            refr_avg_time, render_real_avg_time, flush_avg_time,
            cpu);
 #else
+
+    /* **************************************************
+     * This causes the displayed FPS to persist if the FPS drops down to a zero.
+     * It makes it a little bit easier to see insteaad of flipping to a
+     * zero right away. If the FPS is >0 the new FPS will be displayed
+     * and the mechanics below will get reset.
+     * **************************************************/
+
     if (fps == 0) {
         info->delay_zero_update_cnt += 1;
         if (info->delay_zero_update_cnt == 10) {
@@ -197,6 +202,8 @@ static void perf_monitor_event_cb(lv_event_t * e)
         info->delay_zero_update_cnt = 0;
         info->last_fps = (uint16_t) fps;
     }
+
+    //  ***************************************************
 
     lv_label_set_text_fmt(
         sysmon,
--- a/src/draw/sw/lv_draw_sw.c
+++ b/src/draw/sw/lv_draw_sw.c
@@ -132,20 +132,15 @@ static int32_t lv_draw_sw_dispatch(lv_draw_unit_t * draw_unit, lv_layer_t * laye
         }
     }
 
-#if LV_USE_OS
     t->state = LV_DRAW_TASK_STATE_IN_PROGRESS;
     draw_sw_unit->base_unit.target_layer = layer;
     draw_sw_unit->base_unit.clip_area = &t->clip_area;
     draw_sw_unit->task_act = t;
 
+#if LV_USE_OS
     /*Let the render thread work*/
     lv_thread_sync_signal(&draw_sw_unit->sync);
 #else
-    t->state = LV_DRAW_TASK_STATE_IN_PROGRESS;
-    draw_sw_unit->base_unit.target_layer = layer;
-    draw_sw_unit->base_unit.clip_area = &t->clip_area;
-    draw_sw_unit->task_act = t;
-
     execute_drawing(draw_sw_unit);
 
     draw_sw_unit->task_act->state = LV_DRAW_TASK_STATE_READY;


src/core/lv_refr.c
```diff
@@ -922,15 +924,6 @@ static void draw_buf_flush(lv_disp_t * disp)
...
     }
-    /*If there are 2 buffers swap them. With direct mode swap only on the last area*/
-    if(lv_disp_is_double_buffered(disp) && (disp->render_mode != LV_DISP_RENDER_MODE_DIRECT || flushing_last)) {
-        if(disp->draw_buf_act == disp->draw_buf_1) {
-            disp->draw_buf_act = disp->draw_buf_2;
-        }
-        else {
-            disp->draw_buf_act = disp->draw_buf_1;
-        }
-    }
```
kisvegabor, line 933: It seems this part is just removed without adding it back anywhere.

src/others/sysmon/lv_sysmon.c
```diff
@@ -131,8 +133,11 @@ static void perf_monitor_disp_event_cb(lv_event_t * e)
             info->refr_start = lv_tick_get();
             break;
         case LV_EVENT_REFR_FINISH:
-            info->refr_elaps_sum += lv_tick_elaps(info->refr_start);
-            info->refr_cnt++;
+            uint32_t elapsed_time = lv_tick_elaps(info->refr_start);
+            if (elapsed_time > 0) {
```
kisvegabor: With your modifications in `lv_refr.c` `LV_EVENT_REFR_FINISH` is sent only if there was something to draw. < 1ms rendering time can actually happen e.g. when only the the cursor of a text area blinks. 

So I'd keep it as it is.

src/others/sysmon/lv_sysmon.c
```diff
@@ -181,6 +185,19 @@ static void perf_monitor_event_cb(lv_event_t * e)
...
+            info->delay_zero_update_cnt = 0;
+            info->last_fps = 0;
+        } else {
+            fps = (uint32_t) info->last_fps;
+        }
+    } else {
+        info->delay_zero_update_cnt = 0;
+        info->last_fps = (uint16_t) fps;
+    }
+
```
kisvegabor, line 205: What's the point of this section?

kisvegabor:
cc @FASTSHIFT

src/others/sysmon/lv_sysmon.c
kdschlosser: What I did here was a quick and dirty to remove the counts from being made for frames that nothing is being rendered. It was only for testing purposes and nothing more.

src/others/sysmon/lv_sysmon.c
kdschlosser: This was to cause the last non zero frame count to drag out for a bit of time so it can be seen instead of flipping to a zero immediately. Again This was only for the purposes of getting an accurate frame count during my test. The tests were really about using the while loop directly after a buffer being passed to the flush function and it defeating the purpose of even having double buffering if that is being done. Any changes I made to the performance monitor portion was only for the purposes of getting an accurate frame count of actual work being done while testing the changed made to the refresh to see if there is any performance gain.

src/core/lv_refr.c
kdschlosser: Nope it's there. it doesn't look exactly like this anymore. It's actually before the flush function is called. 

```c
if(lv_disp_is_double_buffered(disp)) {
        while(disp->flushing);
        /*If there are 2 buffers swap them. With direct mode swap only on the last area*/

        void * buf_on_screen = disp_refr->draw_buf_act;
        void * buf_off_screen = disp_refr->draw_buf_act == disp_refr->draw_buf_1 ? disp_refr->draw_buf_2 : disp_refr->draw_buf_1;

        lv_coord_t stride = lv_disp_get_hor_res(disp_refr);
        uint32_t i;
        for(i = 0; i < disp_refr->inv_p; i++) {
            if(disp_refr->inv_area_joined[i]) continue;
            disp_refr->layer_head->buffer_copy(
                disp_refr->layer_head,
                buf_off_screen, stride, &disp_refr->inv_areas[i],
                buf_on_screen, stride, &disp_refr->inv_areas[i]
            );
        }
        if(disp->render_mode != LV_DISP_RENDER_MODE_DIRECT || flushing_last) {
            disp->draw_buf_act = disp_refr->draw_buf_act == disp_refr->draw_buf_1 ? disp_refr->draw_buf_2 : disp_refr->draw_buf_1;    <<----- RIGHT HERE
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
                 right here
        }
    }
```

the logic there is 

```c
if(lv_disp_is_double_buffered(disp)) {
    if(disp->render_mode != LV_DISP_RENDER_MODE_DIRECT || flushing_last) {
            disp->draw_buf_act = disp_refr->draw_buf_act == disp_refr->draw_buf_1 ? disp_refr->draw_buf_2 : disp_refr->draw_buf_1;
        }
}
```

which is the same as 

```c
if(lv_disp_is_double_buffered(disp) && (disp->render_mode != LV_DISP_RENDER_MODE_DIRECT || flushing_last)) {
        if(disp->draw_buf_act == disp->draw_buf_1) {
            disp->draw_buf_act = disp->draw_buf_2;
        }
        else {
            disp->draw_buf_act = disp->draw_buf_1;
        }
    }
```

src/others/sysmon/lv_sysmon.c
kdschlosser: I think what I did was I added the omitting anything that was a zero time before I changed the code the handles the validation of work being performed and I forgot to remove the checking for a zero. I didn't have a blinking carat in my test and I was doing a double buffer full frame buffers for an 800 x 600 x 32 display.  which is quite a bit of work to do so it would take longer than 1 millisecond so the displayed FPS would not have been skewed at all by that being there.

kdschlosser:
This PR is only for a proof of concept, I just moved some things around and there is probably a better way to go about doing what I have done. That is the reason why I marked it "DO NOT MERGE"

This is the original code block that caused me to move things about.

```
    refr_invalid_areas();

    if(disp_refr->inv_p == 0) goto refr_finish;

    /*Call monitor cb if present*/
    lv_disp_send_event(disp_refr, LV_EVENT_RENDER_READY, NULL);

    if(!lv_disp_is_double_buffered(disp_refr) || disp_refr->render_mode != LV_DISP_RENDER_MODE_DIRECT) goto refr_clean_up;

    /*With double buffered direct mode synchronize the rendered areas to the other buffer*/
    /*We need to wait for ready here to not mess up the active screen*/
    while(disp_refr->flushing);
```

If a user has double buffering and is using DMA memory the mechanism that resets `flushing` field in `lv_disp_t` is when `lv_flush_ready` gets called. Now with DMA this is going to happen while LVGL is still able to execute code. when `    refr_invalid_areas();` gets ran that is when the flush callback gets called. That function call is at the top of the code block above. If you look at the bottom of the code block above you see `while(disp_refr->flushing);`. This is going to completely negate any performance gains gotten when using double buffering and DMA memory because that field is not going to be reset until after the buffer has emptied. so LVGL is not going to be able to go off and fill the second buffer while the first is being emptied. 


by moving that stall to right before the flush callback gets called ensures that 2 buffers and not being written to the display at the same time and also to ensure that there is an empty buffer that can be written to while the other is being emptied. And this is ONLY going to happen if there is actual work that actually needs to be done because if the rest of the changes made upstream where I added in the code that checks before calling  `refr_invalid_areas();`

kdschlosser:
I did make a boo boo tho which I am correcting now.

kdschlosser:
I updated the code. I added some comments, fixed the thing that needed fixing and I also removed the checking for a zero time seeing as how that doesn't need to be there because of the validation of work being performed.

src/core/lv_refr.c
```diff
@@ -901,17 +884,46 @@ static void draw_buf_flush(lv_disp_t * disp)
...
-
-    disp->flushing = 1;
 
     if(disp->last_area && disp->last_part) disp->flushing_last = 1;
     else disp->flushing_last = 0;
 
     bool flushing_last = disp->flushing_last;
 
+    // This code was moved from _lv_disp_refr_timer as it seem wrong to have a
```
kisvegabor, line 893: Please use `/*  */` style comments.

src/core/lv_refr.c
```diff
@@ -901,17 +884,46 @@ static void draw_buf_flush(lv_disp_t * disp)
...
+    if(lv_disp_is_double_buffered(disp)) {
+        while(disp->flushing);
+        /*If there are 2 buffers swap them. With direct mode swap only on the last area*/
+
+        // This might be wrong and some adjustments might need to be made. I
+        // understaand the buffer is being coppied and I don't think it matters if
+        // it happens before the flush function is called or after because after
+        // all it is a copy being made so long as all rendering to the buffer has
+        // been completed before the copy gets made.
+        if (disp_refr->render_mode == LV_DISP_RENDER_MODE_DIRECT) {
```
kisvegabor, line 905: In direct mode, we needed copy the previously rendered areas to the new buffer to synchronize the content. There a few things to consider:
- We need to copy the invalidated areas of the previous frame, so we can't wait much as the invalidated areas will be clear at the end of `disp_refr_timer`.
- We need to wait until the new frame is flushed, else we will overwrite the being displayed image with the copying. 

It means we need to copy before exiting `disp_refr_timer` and flushing needs to be ready before that. 

We can also save the invalidated areas. This way we can copy at latest before starting the rendering. So before rendering a new frame the other buffer needs to flushed and after that we need to copy previous invalidated areas. Saving the areas will require only a few hundred bytes, so probably we can do this to gain more performance. 

Note that all these copying is required only in direct mode. in full refresh mode we can wait for flushing right before sending a new flush.

kdschlosser: ok so say you have buffer 1 and buffer 2. there would be no difference if buffer 1 gets copied just before it gets flushed or just after it gets flushed because the only thing that is happening if the buffer data is getting written to the display so the data would not be changing. The only big difference is if it is done after the flush and DMA Memory is being used the flush function is going to exit before the data has been written This is why there was the while loop after the flush exited and before the buffer data was copied. By copying the data before the flush function gets called or expands the amount of time that is being given for the DMA transfer to be done before copying the buffer data again. That time is now being used to render new information to the already copied buffer..  There is still the while loop there in the event it takes longer for the DMA of the data then it does for LVGL to render. But once that buffer has been written to the display the data gets copied to the buffer that just finished being drawn the buffer gets swapped so the buffer that now have the newly copied data is set to the active buffer.

Use this as a rough idea of what is happening.
image

This is what I changed it to.

image


You can see by placing the buffer copy before the flush function being called will reduce the amount of time the program has to stall or it might remove the need to stall the program entirely depending on the size of the buffer and how the data is being sent. Where as the way it is set up now 100% of the time the program is going to have to stop and wait.

src/core/lv_refr.c
kdschlosser: not a problem. But remember this is not intended to be merged. It is to be used as a proof of concept and it can possibly be done differently or not at all if there is a problem caused by it.

kdschlosser: Just to let you know there are over 2000 occurrences of `//` being used in the current code 😜

src/core/lv_refr.c
kisvegabor: >  By copying the data before the flush function gets called or expands the amount of time that is being given for the DMA transfer to be done before copying the buffer data again.

The problem is that you can't do this. Let's say `buf1` is being displayed and we have the next frame to be flushed in `buf2`. The MCU's LCD periphery continuously reads `buf1`. So if we modify it, it will be visible on the screen too.  

That's why we can't copy anything to `buf1` before flushing `buf2`. The best we can do is
1. start flushing `buf2`
2. save the coordinates of areas to copy into `buf1`
3. start copying when the flush is ready but at latest before rendering.

kdschlosser: > The MCU's LCD periphery continuously reads buf1. So if we modify it, it will be visible on the screen too.

why would it be continuously be reading buf1? isn't that the point of having double buffering? so while one buffer is being written to the display the other can be rendered to?? If that is the case then double buffering when using direct mode would cause a slow down due to the stall and the need to copy the contents of the second buffer onto the first buffer.

You are going to have to elaborate on this a little bit more, I am not understanding what you are saying.

This is the part I am not understanding

> The MCU's LCD periphery continuously reads buf1. So if we modify it, it will be visible on the screen too.

when using double buffering buf1 wouldn't be getting altered while it was being emptied, buf2 would be and then it stalls until the flush function has finished writing buf1. then the data from buf2 gets copied to buf1 and the flush function then gets called to write the data from buf2  and LVGL goes off and renders new data to buf1 while buf2 is being written. Then it circles back around and gets to the stall, this time the program waits until buf2 is written and then the buf1 data gets copied to buf2 and the flush function gets called to write the data in buf1 and LVGL goes off and fills buf2

kisvegabor: This is how it works in the microcontroller world (assuming a typical double buffering scenario with LCD controller in the MCU):
- The MCU has 2 frame buffers typically in external SRAM
- Even if you do nothing on the screen, the LCD peripthery reads the  active frame buffer with 30-60 FPS. It's because TFTs has a very bad memory, and you need to keep telling them what to display.
- Flushing the new buffer can happen in two ways:
  1. In the `flush_cb` you just change the address from which the LCD periphery should read the data. In this case flushing happens immediately, however if you switch in the middle of "frame scanning" you will see tearing  (the upper part of image is coming from the previous frame, and the bottom part is coming from the next frame)
  2. Wait for VSYNC in the `flush_cb` and set the new address when the LCD periphery is not reading the frame buffer. (Usually there is short relax period between scans). In this case changing the frame buffer address if also just 1 instruction, but you might need to wait even 20-30 ms for the VSYNC to do that 1 instruction.

kdschlosser: OK and I understand that. which is why there is still the while loop that is in place. so no rendering takes place on the same buffer as what is being displayed and also so the buffer doesn't get changed out while one of them is still being transferred. The exact same mechanics that were there originally is still there it has simply been moved to a different location that allows LVGL to update the second buffer while the first is being emptied where as the way it is set up now is as soon as the flush function exits there is a stall which waits for the buffer to get written. If DMA memory is used the flush function will exit before the data has been written to the display. Only when the buffer has been emptied is when that while loop will exit. This defeats the purpose of using double buffering because the second buffer is not able to be updated while the first is being written to the display.

I can tell you that I do not experience any tearing or anything like that. I have run this test using wxPython as the framework that handles the rendering to the display I am able to simulate what DMA memory transfer would be using what is called a  paint event that Microsoft Windows does when a window needs to be redrawn. This paint event gets created due to the flush function for LVGL telling the window to refresh. once the paint event has drawn the frame buffer data the function gets called letting LVGL know the buffer data has been emptied.

kisvegabor:
> so no rendering takes place on the same buffer as what is being displayed and also so the buffer doesn't get changed out while one of them is still being transferred. The exact same mechanics that were there originally is still there it has simply been moved to a different location that allows LVGL to update the second buffer while the first is being emptied

These are 2 things here:
1. In `FULL_REFRESH` mode what you say would work. We can wait for the flush ready right before rendering. 
2. However in `DIRECT_MODE` we need to copy the rendered areas from the new buffer to the other buffer. But the "other buffer" is being displayed, so we need to wait until it's not displayed, which is signaled by `flush_ready`.

kdschlosser:
the copy is still taking place. it is happening before the flush function gets called.

so look at it this way.... data gets rendered to a frame buffer and just before flush gets called which is after the data was rendered that data gets copied to to the second frame buffer. so both frame buffers have identical data. then flush gets called passing the frame buffer that was just rendered to the flush function. The other frame buffer can now get used to render to and the render is able to take place without having to stall the program prior to the rendering taking place. Once the program comes full circle once again and after rendering has taken place if the buffer that is being flushed has not finished yet then there is a stall until it is finished so the data from the just rendered buffer can get copied to the frame buffer that was just emptied. This gives you 2 identical frame buffers once again... As I said, all that is being changed is when the data gets copied this allows LVGL to do other things while the buffer is being emptied instead of sitting there waiting for the flush function to finish so it can make the copy. It's a way to reduce the wait time but also a way to do something else while waiting.

kdschlosser:
It will be easier to understand if you see in written out in code.

if you run the script below in Python it will show exactly what I am talking about. You are going to notice the buffer data does get copied like it should be.
```python
import threading
import time
import random

buf1 = []
buf2 = []

active_buf = buf1

flushing = False

# I have to use a thread lock on all print statements so the data will stay organized when it is printed out.
# by doing the copy prior to the flush function the printed data gets all jumbled up
# this is a good thing to actually see happening because it means that more than one thing is being done at the same time
# but for the purposes of this test where I want to show the data being copied properly it makes it hard to see that happening.
print_lock = threading.Lock()


def flush_cb(buf):
    global flushing
    global start_time

    start_time = time.time()
    with print_lock:
        print('flushing buf:', buf)
    time.sleep(1)
    # buffer getting flushed
    flushing = False


def render():
    # adds data to the buffer
    active_buf.append(random.randrange(0, 100))
    with print_lock:
        print('rendering new data to the buffer:', active_buf)

    stop_time = time.time()
    with print_lock:
        print('time it takes from when flushing begins until rendering ends:', (stop_time - start_time) * 1000)


def task_handler():
    global buf1
    global buf2
    global active_buf
    global flushing

    render()

    while flushing:
        pass

    # copying the data
    if active_buf == buf1:
        buf2 = buf1[:]
        active_buf = buf2
        onscreen_buf = buf1
    else:
        buf1 = buf2[:]
        active_buf = buf1
        onscreen_buf = buf2

    flushing = True
    t = threading.Thread(target=flush_cb, args=(onscreen_buf,))
    t.start()


print('copy before flush')
print()

start_time = time.time()
start_run_time = start_time

loop_count = 0
# program loop code
while True:
    loop_count += 1
    task_handler()
    if loop_count == 15:
        break

stop_run_time = time.time()

with print_lock:
    print('\nLoop exited')
    print('loop execution time:', (stop_run_time - start_run_time) * 1000)


# this is here to make sure the last flushing is finished, the loop has exited already
while flushing:
    pass

print()
print()
print('The amount of time it takes from when the flush function starts until the next frame is rendered is 0.0 in all cases.')
print('That tells us that while the buffer is being emptied new data is able to be rendered.')

# reset variables
buf1 = []
buf2 = []

active_buf = buf1
flushing = False


def task_handler():
    global buf1
    global buf2
    global active_buf
    global flushing

    render()

    flushing = True
    t = threading.Thread(target=flush_cb, args=(active_buf,))
    t.start()

    while flushing:
        pass

    # copying the data
    if active_buf == buf1:
        buf2 = buf1[:]
        active_buf = buf2
    else:
        buf1 = buf2[:]
        active_buf = buf1


loop_count = 0

print()
print()
print('copy after flush')
print()

start_time = time.time()
start_run_time = start_time

# program loop code
while True:
    loop_count += 1
    task_handler()
    if loop_count == 15:
        break

stop_run_time = time.time()

with print_lock:
    print('\nLoop exited')
    print('loop execution time:', (stop_run_time - start_run_time) * 1000)

print()
print()
print('The amount of time it takes from when the flush function starts until the next frame is always > 1.0.')
print('This tells us that rendering is not able to take place while the buffer is being flushed.')

print('Look at the total loop execution times between the 2 runs.')
print('My way is over a second faster.')
```


The output from the above code is what you see below.. Notice the loop execution time and also the data being copied properly.

```shell
copy before flush

rendering new data to the buffer: [54]
time it takes from when flushing begins until rendering ends: 0.0
flushing buf: [54]
rendering new data to the buffer: [54, 90]
time it takes from when flushing begins until rendering ends: 0.0
flushing buf: [54, 90]
rendering new data to the buffer: [54, 90, 20]
time it takes from when flushing begins until rendering ends: 0.0
flushing buf: [54, 90, 20]
rendering new data to the buffer: [54, 90, 20, 31]
time it takes from when flushing begins until rendering ends: 0.0
flushing buf: [54, 90, 20, 31]
rendering new data to the buffer: [54, 90, 20, 31, 46]
time it takes from when flushing begins until rendering ends: 0.0
flushing buf: [54, 90, 20, 31, 46]
rendering new data to the buffer: [54, 90, 20, 31, 46, 66]
time it takes from when flushing begins until rendering ends: 0.0
flushing buf: [54, 90, 20, 31, 46, 66]
rendering new data to the buffer: [54, 90, 20, 31, 46, 66, 56]
time it takes from when flushing begins until rendering ends: 0.0
flushing buf: [54, 90, 20, 31, 46, 66, 56]
rendering new data to the buffer: [54, 90, 20, 31, 46, 66, 56, 19]
time it takes from when flushing begins until rendering ends: 0.0
flushing buf: [54, 90, 20, 31, 46, 66, 56, 19]
rendering new data to the buffer: [54, 90, 20, 31, 46, 66, 56, 19, 58]
time it takes from when flushing begins until rendering ends: 0.0
flushing buf: [54, 90, 20, 31, 46, 66, 56, 19, 58]
rendering new data to the buffer: [54, 90, 20, 31, 46, 66, 56, 19, 58, 98]
time it takes from when flushing begins until rendering ends: 0.0
flushing buf: [54, 90, 20, 31, 46, 66, 56, 19, 58, 98]
rendering new data to the buffer: [54, 90, 20, 31, 46, 66, 56, 19, 58, 98, 7]
time it takes from when flushing begins until rendering ends: 0.0
flushing buf: [54, 90, 20, 31, 46, 66, 56, 19, 58, 98, 7]
rendering new data to the buffer: [54, 90, 20, 31, 46, 66, 56, 19, 58, 98, 7, 21]
time it takes from when flushing begins until rendering ends: 0.0
flushing buf: [54, 90, 20, 31, 46, 66, 56, 19, 58, 98, 7, 21]
rendering new data to the buffer: [54, 90, 20, 31, 46, 66, 56, 19, 58, 98, 7, 21, 42]
time it takes from when flushing begins until rendering ends: 0.0
flushing buf: [54, 90, 20, 31, 46, 66, 56, 19, 58, 98, 7, 21, 42]
rendering new data to the buffer: [54, 90, 20, 31, 46, 66, 56, 19, 58, 98, 7, 21, 42, 54]
time it takes from when flushing begins until rendering ends: 0.0
flushing buf: [54, 90, 20, 31, 46, 66, 56, 19, 58, 98, 7, 21, 42, 54]
rendering new data to the buffer: [54, 90, 20, 31, 46, 66, 56, 19, 58, 98, 7, 21, 42, 54, 50]
time it takes from when flushing begins until rendering ends: 0.0
flushing buf: [54, 90, 20, 31, 46, 66, 56, 19, 58, 98, 7, 21, 42, 54, 50]

Loop exited
loop execution time: 14364.708185195923


The amount of time it takes from when the flush function starts until the next frame is rendered is 0.0 in all cases.
That tells us that while the buffer is being emptied new data is able to be rendered.


copy after flush

rendering new data to the buffer: [86]
time it takes from when flushing begins until rendering ends: 0.0
flushing buf: [86]
rendering new data to the buffer: [86, 13]
time it takes from when flushing begins until rendering ends: 1056.6959381103516
flushing buf: [86, 13]
rendering new data to the buffer: [86, 13, 8]
time it takes from when flushing begins until rendering ends: 1057.633876800537
flushing buf: [86, 13, 8]
rendering new data to the buffer: [86, 13, 8, 96]
time it takes from when flushing begins until rendering ends: 1043.107271194458
flushing buf: [86, 13, 8, 96]
rendering new data to the buffer: [86, 13, 8, 96, 78]
time it takes from when flushing begins until rendering ends: 1047.7535724639893
flushing buf: [86, 13, 8, 96, 78]
rendering new data to the buffer: [86, 13, 8, 96, 78, 97]
time it takes from when flushing begins until rendering ends: 1048.2254028320312
flushing buf: [86, 13, 8, 96, 78, 97]
rendering new data to the buffer: [86, 13, 8, 96, 78, 97, 91]
time it takes from when flushing begins until rendering ends: 1029.1481018066406
flushing buf: [86, 13, 8, 96, 78, 97, 91]
rendering new data to the buffer: [86, 13, 8, 96, 78, 97, 91, 70]
time it takes from when flushing begins until rendering ends: 1030.0531387329102
flushing buf: [86, 13, 8, 96, 78, 97, 91, 70]
rendering new data to the buffer: [86, 13, 8, 96, 78, 97, 91, 70, 84]
time it takes from when flushing begins until rendering ends: 1047.1739768981934
flushing buf: [86, 13, 8, 96, 78, 97, 91, 70, 84]
rendering new data to the buffer: [86, 13, 8, 96, 78, 97, 91, 70, 84, 98]
time it takes from when flushing begins until rendering ends: 1029.7582149505615
flushing buf: [86, 13, 8, 96, 78, 97, 91, 70, 84, 98]
rendering new data to the buffer: [86, 13, 8, 96, 78, 97, 91, 70, 84, 98, 33]
time it takes from when flushing begins until rendering ends: 1073.6732482910156
flushing buf: [86, 13, 8, 96, 78, 97, 91, 70, 84, 98, 33]
rendering new data to the buffer: [86, 13, 8, 96, 78, 97, 91, 70, 84, 98, 33, 13]
time it takes from when flushing begins until rendering ends: 1047.7385520935059
flushing buf: [86, 13, 8, 96, 78, 97, 91, 70, 84, 98, 33, 13]
rendering new data to the buffer: [86, 13, 8, 96, 78, 97, 91, 70, 84, 98, 33, 13, 38]
time it takes from when flushing begins until rendering ends: 1058.164119720459
flushing buf: [86, 13, 8, 96, 78, 97, 91, 70, 84, 98, 33, 13, 38]
rendering new data to the buffer: [86, 13, 8, 96, 78, 97, 91, 70, 84, 98, 33, 13, 38, 10]
time it takes from when flushing begins until rendering ends: 1029.1619300842285
flushing buf: [86, 13, 8, 96, 78, 97, 91, 70, 84, 98, 33, 13, 38, 10]
rendering new data to the buffer: [86, 13, 8, 96, 78, 97, 91, 70, 84, 98, 33, 13, 38, 10, 12]
time it takes from when flushing begins until rendering ends: 1040.663719177246
flushing buf: [86, 13, 8, 96, 78, 97, 91, 70, 84, 98, 33, 13, 38, 10, 12]

Loop exited
loop execution time: 15733.45422744751


The amount of time it takes from when the flush function starts until the next frame is always > 1.0.
This tells us that rendering is not able to take place while the buffer is being flushed.
Look at the total loop execution times between the 2 runs.
My way is over a second faster.
```

kisvegabor:
I don't know how wxWidgets handles frame buffers, but for embedded case what I described [here](https://github.com/lvgl/lvgl/pull/4363#discussion_r1265410714) and  [here](https://github.com/lvgl/lvgl/pull/4363#issuecomment-1642091507) should be true. 

Or do you see any flaw in the logic I described there?

kdschlosser:
I want to remove wxWidgets from the test case. I was requesting that someone test this on an actual MCU that uses direct mode. I wanted to know if this in fact would work properly. It should work with the code changes I have made or the changes might need some tweaks but it would allow for double buffering with DMA to work and get the benefits of using DMA memory. which in it's present state those benefits are not going to be gotten when using direct mode.


I do understand what you are saying with respect to how direct mode works and what needs to be done. The only flaw I am seeing is if DMA memory is used. The idea is not not have the program wait while the buffer is being sent to the display. That while loop after flush gets called is causing the program to wait until the buffer data gets sent. 

In the current code you are copying the buffer data after flush gets called so the invalidated areas are still usable to copy the data. In this code the buffer data is being copied before the flush call and the invalidated areas are still good to use in the copy. 

I don't know of a better way to explain it other than what I have done in the code. It's one of those things that is hard to visualize what is happening because of how complex the process is. I don't have the hardware needed to test it which is another issue. C is not my strong area either.

kisvegabor:
> In the current code you are copying the buffer data after flush gets called so the invalidated areas are still usable to copy the data.

Correct. We know the coordinates (as no new invalidation happened) and we have the data as well in the buffer which was just sent out.


> In this code the buffer data is being copied before the flush call and the invalidated areas are still good to use in the copy.

True, both the coordinates and data is there. And there can scenarios when you are right, and we can copy the data at any time. For example in case of the SDL driver we also use the Direct mode (with a single buffer though) and call `SDL_UpdateTexture` which gets the frame buffer and saves into a texture. 

But in general, where an MCUs LCD periphery is continuously reading the frame buffer we cannot do any modification on the active buffer as the changes will be visible on the screen as well. Therefore we need to make the other buffer active first.

A good place to do the copy could be the `lv_disp_flush_ready()`. But in this case we need to save the coordinate of the invalidated areas as the current list will be cleared at the end of the `disp_refr_timer`.

kdschlosser:
> True, both the coordinates and data is there. And there can scenarios when you are right, and we can copy the data at any time. For example in case of the SDL driver we also use the Direct mode (with a single buffer though) and call SDL_UpdateTexture which gets the frame buffer and saves into a texture.


What I have done in my tests is I have also added a second buffer and set the display size to 1920 x 1080 which makes it take longer for the buffer to get emptied. when the flush function gets called I added in code to the flush function that start a thread to do the rendering which allows the program to run while the buffer is being emptied. This simulates a DMA memory transfer. The program runs without any issues when doing this. No tearing occurs and there is no missing data that is seen.

If you take and copy the data from the buffer that was just rendered to to the buffer that just finished flushing (this is done before the flush) and then set the buffer that just finished flushing as the active buffer the buffer that just finished rendering is going to get passed to the flush function call. that is seen in this line in lv_refr.c

```c
call_flush_cb(disp, &disp->refreshed_area, layer->buf);
```

Notice that the buffer being passed is not disp->draw_buf_act.

The mechanics that set the `layer->buf` are still in place and that occurs after the pointer has been passed to the flush function.

so the routine goes
```
render to buf1
wait for buf2 to finish flushing
copy invalid areas from buf1 to buf2
set buf2 as active buffer
pass buf1 to flush function

render to buf2
wait for buf1 to finish flushing
copy invalid areas from buf2 to buf1
set buf1 as active buffer
pass buf2 to flush function

render to buf1
wait for buf2 to finish flushing
copy invalid areas from buf1 to buf2
set buf2 as active buffer
pass buf1 to flush function
```

where as the current routine is
```
render to buf1
pass buf1 to flush function
wait for buf1 to finish flushing
copy invalid areas from buf1 to buf2
set buf2 as active buffer

render to buf2
pass buf2 to flush function
wait for buf2 to finish flushing
copy invalid areas from buf2 to buf1
set buf1 as active buffer

render to buf1
pass buf1 to flush function
wait for buf1 to finish flushing
copy invalid areas from buf1 to buf2
set buf2 as active buffer
```

with the DMA memory in use with double buffering and direct mode you end up with this.


```
render to buf1
wait for buf2 to finish flushing
copy invalid areas from buf1 to buf2
set buf2 as active buffer
pass buf1 to flush function
---------------------------------------------------------------------------------
                                                                               This is the important area. Because there is no stall after the flush function
render to buf2                                                         other things like rendering can take place. 
---------------------------------------------------------------------------------
wait for buf1 to finish flushing
copy invalid areas from buf2 to buf1
set buf1 as active buffer
pass buf2 to flush function
```

where as in the current code

```
render to buf1
pass buf1 to flush function
---------------------------------------------------------------------------------------------
                                                                               unable to do any work because of the stall right after the flush function.
---------------------------------------------------------------------------------------------
wait for buf1 to finish flushing
copy invalid areas from buf1 to buf2
set buf2 as active buffer

render to buf2
pass buf2 to flush function
wait for buf2 to finish flushing
copy invalid areas from buf2 to buf1
set buf1 as active buffer
```

kisvegabor:
I think the point that you miss in MCUs the current frame buffers content is continuously read the LCD periphery. So `flush` doesn't send out the content anywhere, just tells the LCD periphery to read the content to display from this buffer from now on. 

Therefore:
```
render to buf1                         -> Ok
wait for buf2 to finish flushing       -> Ok, after that the content of buf2 is being displayed and read periodically by LCD periphery
copy invalid areas from buf1 to buf2   -> PROBLEM: buf2 is being dislayed, if we copy anything there it will be shown on the display
set buf2 as active buffer
pass buf1 to flush function
```

kdschlosser:
I understand that it is memory pointers being used. The data in memory gets sent to the display. Once that is done LVGL is told the process has finished. So no data is getting copied from one buffer to the next while data is being sent to the display. The same thing holds true in both the original and also this PR.

The ONLY difference is that code to stall the program to make sure that the writing of data has completed and the copy code has simply be relocated to before flush gets called instead of after.


> copy invalid areas from buf1 to buf2   -> PROBLEM: buf2 is being dislayed, if we copy anything there it will be shown on the display

How is this a problem? If the function has been called to let LVGL know that the data has been written to the display then it would be safe to modify the buffer.

If That is not the case then writing new data to the buffer while it is being sent to the display would be no different than changing the pointer to the buffer while it is in the middle of writing data to the display. 

 So question. When using single buffer in direct mode is the buffer being updated while that same buffer is being sent to the display? From what you have stated that would be a true statement. No ill effects happen in single buffer mode. Actually there would be zero purpose to using double buffering if data is being sent to the display at the same time the buffer is being updated by LVGL. Instead of having code to handle the use case of double buffer in direct mode and copying data from one buffer to the next LVGL should error informing the user that double buffering in direct mode is not supported and it would only slow things down.

kisvegabor:
>  If the function has been called to let LVGL know that the data has been written to the display then it would be safe to modify the buffer.

Nope. In case of double buffer FULL and DIRECT mode, when `lv_disp_féish_ready()` is called it means the new buffer in the display. Ideally no copy should be involved. The driver just needs to tell to the LCD periphery that from now, use this memory address for the display's content. So just an address swap. 

That's why I say if you modify anything on the flushed buffer it will still be visible on the display. 

I need to mention that the scenario you have described is also valid. That is the content of the buffer is copied somewhere and after that you can update the buffer. However there are difference drivers too where there is no copy.

kdschlosser:
so the time it takes for the flush function to set the new pointer should be really small is what you are saying?

kisvegabor:
>  so the time it takes for the flush function to set the new pointer should be really small is what you are saying?

Can be quick, but can take very long too. There are cases when the MCU (or rather the driver) is waiting for a VSYNC (blank period of the display) to set the new address. It avoids the issues of switching in the middle of a display refresh and having the previous frame on the top of the display and the new frame on the bottom (tearing).

PR #4371, opened 2023-07-13
chore: add config set LV_WIDGETS_HAS_DEAFULT_VALUE

yushuailong:

add config set label default text

### Checkpoints
 

fuxiujun:
Closes #4372

lv_conf_template.h
```diff
@@ -494,6 +494,8 @@
 
 #define LV_USE_WIN        1
 
+#define LV_WIDGETS_HAS_DEAFULT_VALUE  1
```
kisvegabor: Please move it before the first widget.

src/widgets/bar/lv_bar.c
```diff
@@ -231,7 +231,9 @@ static void lv_bar_constructor(const lv_obj_class_t * class_p, lv_obj_t * obj)
 
     lv_obj_clear_flag(obj, LV_OBJ_FLAG_CHECKABLE);
     lv_obj_clear_flag(obj, LV_OBJ_FLAG_SCROLLABLE);
+#if LV_WIDGETS_HAS_DEAFULT_VALUE
     lv_bar_set_value(obj, 0, LV_ANIM_OFF);
```
kisvegabor: I think we should keep. It just ensures that all internal variables are sete correctly.

src/widgets/calendar/lv_calendar.c
```diff
@@ -282,8 +282,10 @@ static void lv_calendar_constructor(const lv_obj_class_t * class_p, lv_obj_t * o
     lv_obj_set_flex_flow(obj, LV_FLEX_FLOW_COLUMN);
     lv_obj_set_flex_grow(calendar->btnm, 1);
 
+#if LV_WIDGETS_HAS_DEAFULT_VALUE
     lv_calendar_set_showed_date(obj, calendar->showed_date.year, calendar->showed_date.month);
```
kisvegabor, line 289: Here we just apply what was set in line 246-252. So we should either remove those or keep these lines too.  I vote for removing line 246-252 too.

src/widgets/checkbox/lv_checkbox.c
```diff
@@ -130,10 +130,12 @@ static void lv_checkbox_constructor(const lv_obj_class_t * class_p, lv_obj_t * o
     LV_UNUSED(class_p);
     LV_TRACE_OBJ_CREATE("begin");
 
+#if LV_WIDGETS_HAS_DEAFULT_VALUE
     lv_checkbox_t * cb = (lv_checkbox_t *)obj;
-
     cb->txt = (char *)"Check box";
     cb->static_txt = 1;
+#endif
```
kisvegabor, line 141: Please add 
```
#else
    cb->txt = (char *)"";
    cb->static_txt = 1;
#endif
```

src/widgets/keyboard/lv_keyboard.c
```diff
@@ -438,7 +438,9 @@ static void lv_keyboard_constructor(const lv_obj_class_t * class_p, lv_obj_t * o
     lv_obj_add_event(obj, lv_keyboard_def_event_cb, LV_EVENT_VALUE_CHANGED, NULL);
     lv_obj_set_style_base_dir(obj, LV_BASE_DIR_LTR, 0);
 
+#if LV_WIDGETS_HAS_DEAFULT_VALUE
     lv_keyboard_update_map(obj);
```
kisvegabor: Here we just apply `keyboard->mode`. I think we should keep this line ones the "maps" are there.

src/widgets/label/lv_label.c
```diff
@@ -660,9 +660,10 @@ static void lv_label_constructor(const lv_obj_class_t * class_p, lv_obj_t * obj)
     label->dot_tmp_alloc = 0;
 
     lv_obj_clear_flag(obj, LV_OBJ_FLAG_CLICKABLE);
+#if LV_WIDGETS_HAS_DEAFULT_VALUE
     lv_label_set_long_mode(obj, LV_LABEL_LONG_WRAP);
```
kisvegabor, line 663: The long mode should be set in all case to have a sane default state.

src/widgets/label/lv_label.c
```diff
@@ -660,9 +660,10 @@ static void lv_label_constructor(const lv_obj_class_t * class_p, lv_obj_t * obj)
...
     lv_label_set_text(obj, LV_LABEL_DEFAULT_TEXT);
```
kisvegabor, line 664: Let's add the if around the `LV_LABEL_DEFAULT_TEXT` like:
```c
#if LV_WIDGETS_HAS_DEAFULT_VALUE
   #define LV_LABEL_DEFAULT_TEXT "Text"
#else
   #define LV_LABEL_DEFAULT_TEXT ""
#endif
```

src/widgets/spinbox/lv_spinbox.c
```diff
@@ -354,7 +354,9 @@ static void lv_spinbox_constructor(const lv_obj_class_t * class_p, lv_obj_t * ob
     lv_textarea_set_one_line(obj, true);
     lv_textarea_set_cursor_click_pos(obj, true);
 
+#if LV_WIDGETS_HAS_DEAFULT_VALUE
     lv_spinbox_updatevalue(obj);
```
kisvegabor: As we need to set some defaults anyway, here I would keep it for consistency to apply the previously set values.

src/widgets/textarea/lv_textarea.c
```diff
@@ -848,7 +848,9 @@ static void lv_textarea_constructor(const lv_obj_class_t * class_p, lv_obj_t * o
     lv_obj_add_event(ta->label, label_event_cb, LV_EVENT_ALL, NULL);
     lv_obj_add_flag(obj, LV_OBJ_FLAG_SCROLL_ON_FOCUS);
     lv_obj_clear_flag(obj, LV_OBJ_FLAG_SCROLL_WITH_ARROW);
+#if LV_WIDGETS_HAS_DEAFULT_VALUE
     lv_textarea_set_cursor_pos(obj, 0);
```
kisvegabor, line 852: I think it also makes sense to explicitly set the cursor to the beginning.

kisvegabor, line 852: I think it also makes sense to explicitly set the cursor to the beginning.

kisvegabor (commented):
Thank you! I've added a few comments.

PR #4403, opened 2023-07-24
fix(config): fix typo in LV_USE_PERF_MONITOR and LV_USE_MEM_MONITOR

fronders:

lv_conf_template.h checks for PERF and MEM monitors align top mid were incorrect resulting in the monitor always displaying bottom right as in default


kisvegabor:
Thank you!

Issue #4382, opened 2023-07-17
[Feature] Seperate Decoder List for Every GPU


XuNeo:
### Introduce the problem
Current decoders are organized in a global list including bin, png, gif etc. 
For GPU, not only the decoded image needs to meet some requirements(stride, address alignment etc.), but GPU also needs to register special decoders(ETC etc.).

### Suggested solution

@terry0012 suggests to add a decoder list to every GPU(sw renderer included), so when GPU needs to draw an image, it firstly decodes it from supported decoders.
@FASTSHIFT implemented this idea and code looks cleaner.

kisvegabor:
I'm trying to understand what the problem is with using the global decoder list:
- Let's say you register an other PNG decoder and want to be sure the GPU uses you PNG decoder? It can be a valid use case as the output of the 2 PNG decoders might not be compatible. 
- If it's a special image format there is no problem as only the new decider will match and try to decode it.  

Note that #4241 will unify the memory alignment and stride of all buffers related to drawing.


In your suggestion the the decoder list are stored in the `lv_draw_unit_t`?

XuNeo:
>  It can be a valid use case as the output of the 2 PNG decoders might not be compatible.

I thinks that's the use case. #4241 is a solution, that we take all reasonable limitations of GPU into consideration. Strides width, memory alignement, and memory flags.

@FASTSHIFT @terry0012 Could you comment more? Will #4241 be enough?

XuNeo:
There are one limitation still not mentioned, but looks like we don't need seperated decoder list.

Premultiply RGB with alpha after decoding. `vglite` for example, needs it. One possible solution is: 
 - a. Add an accelerated premultiply callback from driver. 
 - b. Add config to indicate all images decoded should be multiplied. 
 - c. For image array locates in flash, there's should be a bit to indicate if image already been premultiplied.
 
Current software rendere does't support premultiplied image. We can simply disable it just like `tiled` image and treat it as GPU only format.

Do you think it's doable? Does premultiplied image benefit software rendere as well?

kisvegabor:
It can be useful in SW decoder as well, as we can save a multiplication during blending. I need to check if we can add it without too much code duplication. 

To indicate if an image premultipled I'd would add 
```c
LV_COLOR_FORMAT_RGB565A8_PREMULTIPLIED,
LV_COLOR_FORMAT_ARGB8888_PREMULTIPLIED
```

What do you think?

kisvegabor:
[...]

PR #4414, opened 2023-07-27
feat(tool): PNG to lvgl image in python

XuNeo:

Usage:
```
usage: LVGLImage.py [-h] [--ofmt {C,BIN,PNG}]
                    [--cf {L8,I1,I2,I4,I8,A1,A2,A4,A8,ARGB8888,XRGB8888,RGB565,RGB565A8,RGB888,TRUECOLOR,TRUECOLOR_ALPHA,AUTO}]
                    [-o OUTPUT] [-v]
                    input

LVGL PNG to bin image tool.

positional arguments:
  input                 the folder or filename to be recursively converted

optional arguments:
  -h, --help            show this help message and exit
  --ofmt {C,BIN,PNG}    output filename format, C or BIN
  --cf {L8,I1,I2,I4,I8,A1,A2,A4,A8,ARGB8888,XRGB8888,RGB565,RGB565A8,RGB888,TRUECOLOR,TRUECOLOR_ALPHA,AUTO}
                        bin image color format, use AUTO for automatically
                        choose from I1/2/4/8
  -o OUTPUT, --output OUTPUT
                        Select the output folder, default to ./output
  -v, --verbose
```
### Checkpoints
 
### DIFF
--- a/scripts/LVGLImage.py
+++ b/scripts/LVGLImage.py
@@ -33,8 +33,13 @@ def uint32_t(val) -> bytes:
         raise ParameterError(f"overflow: {hex(val)}")
 
 
-def color_premultiply(r, g, b, a):
-    return ((r * a) >> 8, (g * a) >> 8, (b * a) >> 8, a)
+def color_pre_multiply(r, g, b, a, background):
+    bb = background & 0xff
+    bg = (background >> 8) & 0xff
+    br = (background >> 16) & 0xff
+
+    return ((r * a + (255 - a) * br) >> 8, (g * a + (255 - a) * bg) >> 8,
+            (b * a + (255 - a) * bb) >> 8, a)
 
 
 class Error(Exception):
@@ -385,8 +390,9 @@ class LVGLImage:
             data_out = []  # stride adjusted new data
             if new_stride < current_stride:  # remove padding byte
                 for i in range(h):
-                    data_out.append(data_in[i * current_stride:(i + 1) *
-                                            new_stride])
+                    start = i * current_stride
+                    end = start + new_stride
+                    data_out.append(data_in[start:end])
             else:  # adding more padding bytes
                 padding = b'\x00' * (new_stride - current_stride)
                 for i in range(h):
@@ -536,8 +542,10 @@ class LVGLImage:
 #define LV_ATTRIBUTE_IMG_DUST
 #endif
 
-const LV_ATTRIBUTE_MEM_ALIGN LV_ATTRIBUTE_LARGE_CONST LV_ATTRIBUTE_IMG_DUST uint8_t {varname}_map[] = {{
+const LV_ATTRIBUTE_MEM_ALIGN LV_ATTRIBUTE_LARGE_CONST LV_ATTRIBUTE_IMG_DUST
+uint8_t {varname}_map[] = {{
 '''
+
         ending = f'''
 }};
 
@@ -576,6 +584,8 @@ const lv_img_dsc_t {varname} = {{
         self._check_ext(filename, ".png")
         self._check_dir(filename)
 
+        old_stride = self.stride
+        self.adjust_stride(align=1)
         if self.cf.is_indexed:
             data = self.data
             # Separate lvgl bin image data to palette and bitmap
@@ -620,14 +630,22 @@ const lv_img_dsc_t {varname} = {{
         with open(filename, "wb") as f:
             encoder.write_array(f, data)
 
-    def from_png(self, filename: str, cf: ColorFormat = None):
+        self.adjust_stride(stride=self.stride)
+
+    def from_png(self,
+                 filename: str,
+                 cf: ColorFormat = None,
+                 background: int = 0x00_00_00):
         '''
         Create lvgl image from png file.
         If cf is none, used I1/2/4/8 based on palette size
         '''
 
+        self.background = background
+
         if cf is None:  # guess cf from filename
-            # split filename string and match with ColorFormat to check which cf to use
+            # split filename string and match with ColorFormat to check
+            # which cf to use
             names = str(path.basename(filename)).split(".")
             for c in names[1:-1]:
                 if c in ColorFormat.__members__:
@@ -670,8 +688,8 @@ const lv_img_dsc_t {varname} = {{
         if palette_len != cf.ncolors:
             if not auto_cf:
                 logging.warning(
-                    f"{path.basename(filename)} palette: {palette_len}, extended to: {cf.ncolors}"
-                )
+                    f"{path.basename(filename)} palette: {palette_len}, "
+                    f"extended to: {cf.ncolors}")
             palette += [(255, 255, 255, 0)] * (cf.ncolors - palette_len)
 
         # Assemble lvgl image palette from PNG palette.
@@ -722,7 +740,7 @@ const lv_img_dsc_t {varname} = {{
             B = row[2::4]
             A = row[3::4]
             for r, g, b, a in zip(R, G, B, A):
-                r, g, b, a = color_premultiply(r, g, b, a)
+                r, g, b, a = color_pre_multiply(r, g, b, a, self.background)
                 luma = 0.2126 * r + 0.7152 * g + 0.0722 * b
                 rawdata += uint8_t(int(luma))
 
@@ -737,17 +755,17 @@ const lv_img_dsc_t {varname} = {{
         elif cf in (ColorFormat.XRGB8888, ColorFormat.TRUECOLOR):
 
             def pack(r, g, b, a):
-                r, g, b, a = color_premultiply(r, g, b, a)
+                r, g, b, a = color_pre_multiply(r, g, b, a, self.background)
                 return uint32_t((0xff << 24) | (r << 16) | (g << 8) | (b << 0))
         elif cf == ColorFormat.RGB888:
 
             def pack(r, g, b, a):
-                r, g, b, a = color_premultiply(r, g, b, a)
+                r, g, b, a = color_pre_multiply(r, g, b, a, self.background)
                 return uint24_t((r << 16) | (g << 8) | (b << 0))
         elif cf == ColorFormat.RGB565:
 
             def pack(r, g, b, a):
-                r, g, b, a = color_premultiply(r, g, b, a)
+                r, g, b, a = color_pre_multiply(r, g, b, a, self.background)
                 color = (r >> 3) << 11
                 color |= (g >> 2) << 5
                 color |= (b >> 3) << 0
@@ -922,6 +940,7 @@ class PNGConverter:
                  cf: ColorFormat,
                  ofmt: OutputFormat,
                  odir: str,
+                 background: int = 0x00,
                  align: int = 1,
                  keep_folder=True) -> None:
         self.files = files
@@ -931,6 +950,7 @@ class PNGConverter:
         self.pngquant = None
         self.keep_folder = keep_folder
         self.align = align
+        self.background = background
 
     def _replace_ext(self, input, ext):
         if self.keep_folder:
@@ -946,12 +966,16 @@ class PNGConverter:
         for f in self.files:
 
             if self.ofmt == OutputFormat.RLE_FILE:
-                rle = RLEImage().from_png(f, self.cf)
+                rle = RLEImage().from_png(f,
+                                          self.cf,
+                                          background=self.background)
                 rle.adjust_stride(align=self.align)
                 output.append((f, rle))
                 rle.to_rle(self._replace_ext(f, ".rle"))
             else:
-                img = LVGLImage().from_png(f, self.cf)
+                img = LVGLImage().from_png(f,
+                                           self.cf,
+                                           background=self.background)
                 img.adjust_stride(align=self.align)
                 output.append((f, img))
                 if self.ofmt == OutputFormat.BIN_FILE:
@@ -986,6 +1010,12 @@ def main():
                         type=int,
                         metavar='byte',
                         nargs='?')
+    parser.add_argument('--background',
+                        help="Background color for formats without alpha",
+                        default=0x00_00_00,
+                        type=lambda x: int(x, 0),
+                        metavar='color',
+                        nargs='?')
     parser.add_argument('-o',
                         '--output',
                         default="./output",
@@ -1019,6 +1049,7 @@ def main():
                              cf,
                              ofmt,
                              args.output,
+                             background=args.background,
                              align=args.align,
                              keep_folder=False)
     output = converter.convert()
@@ -1031,9 +1062,10 @@ def main():
 def test():
     logging.basicConfig(level=logging.INFO)
     f = "pngs/cogwheel.RGB565A8.png"
-    img = LVGLImage().from_png(f)
+    img = LVGLImage().from_png(f, cf=ColorFormat.RGB565, background=0xFF_FF_00)
     img.adjust_stride(align=16)
     img.to_bin(f + ".bin")
+    img.to_png(f + ".png")  # convert back to png
 
 
 if __name__ == "__main__":
--- a/scripts/LVGLImage.py
+++ b/scripts/LVGLImage.py
@@ -1,0 +1,1041 @@
+#!/usr/bin/env python3
+import os
+import logging
+import argparse
+import subprocess
+from os import path
+from enum import Enum
+from typing import List
+from pathlib import Path
+
+try:
+    import png
+except ImportError:
+    raise ImportError("Need pypng package, do `pip3 install pypng`")
+
+
+def uint8_t(val) -> bytes:
+    return val.to_bytes(1, byteorder='little')
+
+
+def uint16_t(val) -> bytes:
+    return val.to_bytes(2, byteorder='little')
+
+
+def uint24_t(val) -> bytes:
+    return val.to_bytes(3, byteorder='little')
+
+
+def uint32_t(val) -> bytes:
+    try:
+        return val.to_bytes(4, byteorder='little')
+    except OverflowError:
+        raise ParameterError(f"overflow: {hex(val)}")
+
+
+def color_premultiply(r, g, b, a):
+    return ((r * a) >> 8, (g * a) >> 8, (b * a) >> 8, a)
+
+
+class Error(Exception):
+
+    def __str__(self):
+        return self.__class__.__name__ + ': ' + ' '.join(self.args)
+
+
+class FormatError(Error):
+    '''
+    Problem with input filename format.
+    BIN filename does not conform to standard lvgl bin image format
+    '''
+
+
+class ParameterError(Error):
+    '''
+    Parameter for LVGL image not correct
+    '''
+
+
+class PngQuant:
+    '''
+    Compress PNG file to 8bit mode using `pngquant`
+    '''
+
+    def __init__(self, ncolors=256, dither=True, exec_path="") -> None:
+        executable = path.join(exec_path, "pngquant")
+        self.cmd = (f"{executable} {'--nofs' if not dither else ''} "
+                    f"{ncolors}  --force - < ")
+
+    def convert(self, filename) -> bytes:
+        if not os.path.isfile(filename):
+            raise BaseException(f"file not found: {filename}")
+
+        try:
+            compressed = subprocess.check_output(
+                f'{self.cmd} "{str(filename)}"',
+                stderr=subprocess.STDOUT,
+                shell=True)
+        except subprocess.CalledProcessError:
+            raise BaseException(
+                "cannot find pngquant tool, install it via "
+                "`sudo apt install pngquant` for debian "
+                "or `brew install pngquant` for macintosh "
+                "For windows, you may need to download pngquant.exe from "
+                "https://pngquant.org/, and put it in your PATH.")
+
+        return compressed
+
+
+class ColorFormat(Enum):
+    UNKNOWN = 0x00
+    TRUECOLOR = 0x04
+    TRUECOLOR_ALPHA = 0x05
+    L8 = 0x06
+    I1 = 0x07
+    I2 = 0x08
+    I4 = 0x09
+    I8 = 0x0A
+    A1 = 0x0B
+    A2 = 0x0C
+    A4 = 0x0D
+    A8 = 0x0E
+    ARGB8888 = 0x10
+    XRGB8888 = 0x11
+    RGB565 = 0x12
+    RGB565A8 = 0x14
+    RGB888 = 0x0F
+
+    @property
+    def bpp(self) -> int:
+        '''
+        Return bit per pixel for this cf
+        '''
+        cf_map = {
+            ColorFormat.UNKNOWN: 0x00,
+            ColorFormat.L8: 8,
+            ColorFormat.I1: 1,
+            ColorFormat.I2: 2,
+            ColorFormat.I4: 4,
+            ColorFormat.I8: 8,
+            ColorFormat.A1: 1,
+            ColorFormat.A2: 2,
+            ColorFormat.A4: 4,
+            ColorFormat.A8: 8,
+            ColorFormat.ARGB8888: 32,
+            ColorFormat.XRGB8888: 32,
+            ColorFormat.RGB565: 16,
+            ColorFormat.RGB565A8: 16,  # 16bpp + a8 map
+            ColorFormat.RGB888: 24,
+            ColorFormat.TRUECOLOR: 32,
+            ColorFormat.TRUECOLOR_ALPHA: 32,
+        }
+
+        return cf_map[self]
+
+    @property
+    def ncolors(self) -> int:
+        '''
+        Return number of colors in palette if cf is indexed1/2/4/8.
+        Return zero if cf is not indexed format        '''
+
+        cf_map = {
+            ColorFormat.I1: 2,
+            ColorFormat.I2: 4,
+            ColorFormat.I4: 16,
+            ColorFormat.I8: 256,
+        }
+        return cf_map.get(self, 0)
+
+    @property
+    def is_indexed(self) -> bool:
+        '''
+        Return if cf is indexed color format
+        '''
+        return self.ncolors != 0
+
+    @property
+    def is_alpha_only(self) -> bool:
+        return ColorFormat.A1.value <= self.value <= ColorFormat.A8.value
+
+    @property
+    def has_alpha(self) -> bool:
+        return self.is_alpha_only or self in (
+            ColorFormat.ARGB8888,
+            ColorFormat.XRGB8888,  # const alpha: 0xff
+            ColorFormat.TRUECOLOR,  # const alpha: 0xff
+            ColorFormat.TRUECOLOR_ALPHA,
+            ColorFormat.RGB565A8)
+
+    @property
+    def is_colormap(self) -> bool:
+        return self in (ColorFormat.ARGB8888, ColorFormat.RGB888,
+                        ColorFormat.XRGB8888, ColorFormat.RGB565A8,
+                        ColorFormat.RGB565, ColorFormat.TRUECOLOR_ALPHA,
+                        ColorFormat.TRUECOLOR)
+
+    @property
+    def is_luma_only(self) -> bool:
+        return self in (ColorFormat.L8, )
+
+
+def unpack_colors(data: bytes, cf: ColorFormat, w) -> List:
+    '''
+    Unpack lvgl 1/2/4/8/16/32 bpp color to png color: alpha map, grey scale,
+    or R,G,B,(A) map
+    '''
+    ret = []
+    bpp = cf.bpp
+    if bpp == 8:
+        ret = data
+    elif bpp == 4:
+        if cf == ColorFormat.A4:
+            values = [x * 17 for x in range(16)]
+        else:
+            values = [x for x in range(16)]
+
+        for p in data:
+            for i in range(2):
+                ret.append(values[(p >> (4 - i * 4)) & 0x0f])
+                if len(ret) % w == 0:
+                    break
+
+    elif bpp == 2:
+        if cf == ColorFormat.A2:
+            values = [x * 85 for x in range(4)]
+        else:  # must be ColorFormat.I2
+            values = [x for x in range(4)]
+        for p in data:
+            for i in range(4):
+                ret.append(values[(p >> (6 - i * 2)) & 0x03])
+                if len(ret) % w == 0:
+                    break
+    elif bpp == 1:
+        if cf == ColorFormat.A1:
+            values = [0, 255]
+        else:
+            values = [0, 1]
+        for p in data:
+            for i in range(8):
+                ret.append(values[(p >> (7 - i)) & 0x01])
+                if len(ret) % w == 0:
+                    break
+    elif bpp == 16:
+        #  This is RGB565
+        pixels = [(data[2 * i + 1] << 8) | data[2 * i]
+                  for i in range(len(data) // 2)]
+
+        values_5bit = [x * 8 for x in range(32)]
+        values_5bit[-1] = 255
+        values_6bit = [x * 4 for x in range(64)]
+        values_6bit[-1] = 255
+        for p in pixels:
+            ret.append(values_5bit[(p >> 11) & 0x1f])  # R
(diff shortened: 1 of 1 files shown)


kisvegabor:
Do you know a way to run this script in a browser for the [online version of the image converter](https://lvgl.io/tools/imageconverter)?

XuNeo:
Hi Gabor,

I don't know how to run it in a browser. The python tool relies on tool `pngquant`.
I think it's better to keep the `js` version tool for browser.
The reason I make this tool is JS version is too slow, and not good for CI.

kisvegabor:
I wanted to try out the script but it failed with:
```
  File "/home/kisvegabor/projects/lvgl/./img_conv_v9.py", line 3, in <module>
    import png
ModuleNotFoundError: No module named 'png'
```

Could you add a short guide about installing the prerequisites?

XuNeo:
I added guide to corresponding error message. 

It's tested on ubuntu, macos and windows.

For windows, I have to execute `python LVGLImage.py` instead of `./LVGLImage.py`

XuNeo:
Hi Gabor,
I have finished changes of `0x0 -> 0x00`.
Do you plan to update lv_img_header_t?

>  Especially if we add padding for stride on each line end.
It's more compilicated than I thought. Need to process bits by bits for CF's like indexed1/2/4. I'll add this feature later.

kisvegabor:
> Do you plan to update lv_img_header_t?

Do you mean [this](https://github.com/lvgl/lvgl/issues/4382#issuecomment-1667414748)?



> It's more compilicated than I thought. Need to process bits by bits for CF's like indexed1/2/4. I'll add this feature later.

Note that, stride is in bytes. So an image with 15 bits width, will be 2 bytes with stride = 1 and 4 bytes with stride = 4.

XuNeo:
> > Do you plan to update lv_img_header_t?
> 
> Do you mean [this](https://github.com/lvgl/lvgl/issues/4382#issuecomment-1667414748)?
> 
> >

Yes.

> Note that, stride is in bytes. So an image with 15 bits width, will be 2 bytes with stride = 1 and 4 bytes with stride = 4.

Let's check with NXP about stride alignement requirement on VGLITE. Attached is from NXP.

image

kisvegabor:
It seems the stride is 16 pixel for non indexed formats. So using pixel unit might make more sense.  For indexed image we can apply a trade of by saying all indexed formats are 1 byte/pixel, so the stride is 16 byte.

cc @nicusorcitu

XuNeo:
Hi Gabor,

I want to bring up the discussion of adding stride parameter to every image, so that any stride requirements on specific hardware can be fulfilled(by custom tool for sure).

Could you explain again of you concern?

Thanks

nicusorcitu:
I just comment on [this thread](https://github.com/lvgl/lvgl/pull/4241#issuecomment-1680085449) before seeing this comment.
The only formats (so far) that does not apply for `stride =16px` are index2 (stride = 32px) and index1 (stride = 64px).

kisvegabor:
> I want to bring up the discussion of adding stride parameter to every image, so that any stride requirements on specific hardware can be fulfilled(by custom tool for sure).
Could you explain again of you concern?

To be honest, I wasn't 100% sure about this decision, so let's discuss it. 

If the stride can be anything, and stored per image, how can e.g. the PNG decoder know which stride to use? Similarly, when LVGL needs to allocate a buffer for a layer, how can it decide which stride to use? 
If we set the stride alignment requirement by a define we can be sure that all internal and external modules of LVGL will use the same settings and everything will be compatible with everything. If we give freedom, there is room for error and incompatibility. 

---


The stride in pixel for indexed formats is really tricky, but I things we can find some special rules for these cases. E.g. always use at least one byte per pixel for stride alignment. So 
- stride alignment:  16 px but at least 16 bytes
- color format INDEXED_2
- image width: 70 px -> 17.5 bytes
- image width with stride: 32 bytes

This way we loose a few bytes for INDEXED_1, 2, 4, but I think it's acceptable.

XuNeo:
The stride align requirement for different color formats are different. If it's true for `vglite`, then we better assume it's also true for other GPUs.

I think the only way to solve it from `lvgl`'s point of view is to store the image width together with stride width in image header or data.

XuNeo:
> If the stride can be anything, and stored per image, how can e.g. the PNG decoder know which stride to use? Similarly, when LVGL needs to allocate a buffer for a layer, how can it decide which stride to use?

If I understand correctly, the new draw task mechanism requires every draw unit to decode the image by itself, thus knows how to meet alignment requirement. 

Or, the default alignment configuration should be used and decoder stores the stride width to decoded image so that when draw unit tries to draw this image, it detects if the alignment fits or not.



> This way we loose a few bytes for INDEXED_1, 2, 4, but I think it's acceptable.

For `vglite`, there's stride parameter for certain image, so adding more padding bytes won't be an issue. I'm not sure if other GPUs support this feature(stride). 
```c
    typedef struct vg_lite_buffer {
        int32_t width;                  /*! Width of the buffer in pixels. */
        int32_t height;                 /*! Height of the buffer in pixels. */
        int32_t stride;                 /*! The number of bytes to move from one line in the buffer to the next line. */
        vg_lite_buffer_layout_t tiled;  /*! Indicating the buffer memory layout is linear or tiled. */
        vg_lite_buffer_format_t format; /*! The pixel format of the buffer. */
        void * handle;                  /*! The memory handle of the buffer's memory as allocated by the VGLite kernel. */
        void * memory;                  /*! The logical pointer to the buffer's memory for the CPU. */
        /* ... */
    } vg_lite_buffer_t;
```

kisvegabor:
> If I understand correctly, the new draw task mechanism requires every draw unit to decode the image by itself, thus knows how to meet alignment requirement.

If the draw unit decodes the image for itself it's ok, but what to do with the general decoders, e.g. PNG. Let's say SW decodes a PNG image, it will be cached, and later (from the cache) it will be used by VG-Lite too.

In this case the SW might use `stride alignment = 1`, but VG-Lite  needs `stride alignment = 16`. So will we cache the image twice?  Or how can we manage this?

What about making the stride dependent on the color format in `lv_draw_buf_width_to_stride(width, color_format)`. You can have your custom implementation of this function and it can return any value based on the color format. This way `LV_DRAW_BUF_STRIDE_ALIGN` doesn't needs to be defined externally, but we still have singe source of truth to map any width and color format to stride in a consistent way.

Actually, we can still have `LV_DRAW_BUF_STRIDE_ALIGN` just make the demo and example images work with GPUs as well. But the define can have different meaning, e.g. it can be in pixels, or can have a formula inside, etc. It doesn't matter much, as you can even set it to 64 bytes just to test the GPUs. 

Finally, I think keeping byte unit for the stride would be better just because it's more flexible. E.g. in case of RGB888 format 1 pixel is 3 byte. With pixel unit stride you can't describe 16 byte or 32 byte stride.

kisvegabor:
@XuNeo, @nicusorcitu 
FYI, I updated draw_buf to allow selecting a custom implementation. See 1f807ae71

nicusorcitu:
Cool, i will look for it next week. Thank you!
Trimis: Saturday, August 19, 2023 12:04:38 AM
Către: lvgl/lvgl ***@***.***>
Cc: Nicușor Cîțu ***@***.***>; Mention ***@***.***>
Subiect: [EXT] Re: [lvgl/lvgl] feat(tool): PNG to lvgl image in python (PR #4414)

@XuNeo<https://github.com/XuNeo>, @nicusorcitu<https://github.com/nicusorcitu>
FYI, I updated draw_buf to allow selecting a custom implementation. See 1f807ae<https://github.com/lvgl/lvgl/commit/1f807ae7140784cf2ec28859156b22636024176c>

XuNeo:
> For VGLite I can use all the implementation from `LV_DRAW_BUF_BASIC` except the APIs `lv_draw_buf_copy` and `lv_draw_buf_invalidate_cache`. How do I do that without copy-paste the basic code?

For `vglite`, the `LV_DRAW_BUF_BASIC` mode won't be enough since the `indexed/1/2/4/8` have different alignment requirement. We can use the BASIC mode and set alignment to larger align bytes, but supporting `LV_DRAW_BUF_CUSTOM` mode can meet any alignment requirement GPUs could have.

nicusorcitu:
> For vglite, the LV_DRAW_BUF_BASIC mode won't be enough since the indexed/1/2/4/8 have different alignment requirement. We can use the BASIC mode and set alignment to larger align bytes, but supporting LV_DRAW_BUF_CUSTOM mode can meet any alignment requirement GPUs could have.
>
Only index 1/2/4 does not meet the alignment - e.g. there is no implementation for that in lv_color_format_get_size().
But it can be updated so that it will return 1 byte as @kisvegabor suggested and so that the alignment will be set to a larger bytes. But otherwise the implementation is the same! So how can we handle this without duplicating the code?

kisvegabor:
> BTW, `lv_draw_buf_invalidate_cache` need a second parameter, the size to be cleared starting with that address.

As you know the buffer address (start address), width and height, the end address can be calculated too.  However, in DIRECT_MODE it can be really required to invalidate the cache only on a smaller area.  So similarly to `lv_draw_buf_clear` we can add an `area` parameter.

> But otherwise the implementation is the same! So how can we handle this without duplicating the code?

Okay, maybe here replacing the whole draw_buf implementation is really overkill.
We have these functions now:
```c
//*******************
//Common function
//*******************

//Nothing special, just set the parameters. Can be the same for all platforms.
void lv_draw_buf_init(lv_draw_buf_t * draw_buf, lv_coord_t w, lv_coord_t h, lv_color_format_t color_format);
//Simple, just some address calcualtions
void * lv_draw_buf_go_to_xy(lv_draw_buf_t * draw_buf, lv_coord_t x, lv_coord_t y);
//Just wrapper around lv_draw_buf_width_to_stride
uint32_t lv_draw_buf_get_stride(const lv_draw_buf_t * draw_buf);


//**************************
//Board and project dependent
//**************************
//you can allocate anywhere and anyhow
void lv_draw_buf_malloc(lv_draw_buf_t * draw_buf);
void lv_draw_buf_realloc(lv_draw_buf_t  * draw_buf, lv_coord_t w, lv_coord_t h, lv_color_format_t color_format);
void lv_draw_buf_free(lv_draw_buf_t  * draw_buf);


//**************************
//MCU and Vendor dependent
//**************************
 
//There should be e.g. an NXP version from these and 
//these should do different things if VGLite or PXP is enabled
void lv_draw_buf_invalidate_cache(lv_draw_buf_t  * draw_buf);
uint32_t lv_draw_buf_width_to_stride(uint32_t w, lv_color_format_t color_format); //For stride calculation
void * lv_draw_buf_get_buf(lv_draw_buf_t * draw_buf);   //For buffer alignment
void lv_draw_buf_clear(lv_draw_buf_t * draw_buf, const lv_area_t * a);
void lv_draw_buf_copy(void * dest_buf, uint32_t dest_stride, const lv_area_t * dest_area,
                      void * src_buf, uint32_t src_stride, const lv_area_t * src_area, lv_color_format_t color_format);
//Probably we will have lv_draw_buf_rotate here as well
```

Do you agree with this grouping?




---

@nicusorcitu Can you confirm that both PXP and VGLite has a stride parameter which can be set freely and the stride is **not** calculated from the width and color format automatically?

nicusorcitu:
>  So similarly to lv_draw_buf_clear we can add an area parameter.
>
Yes, exactly what I wanted to say. We need a way to cleanup only a certain area (or size) not the whole buffer.
Is the area all the time starting from the buf address? e.g. is the area all the time at the beginning of the buffer or the coordinates can be anywhere is the buffer?

> //**************************
> //Board and project dependent
> //**************************
> //you can allocate anywhere and anyhow
>
Yes, but allow me to have the same place if I don't have to do anything different than basic implementation.

> //**************************
> //MCU and Vendor dependent
> //**************************
>  
> //There should be e.g. an NXP version from these and 
> //these should do different things if VGLite or PXP is enabled
>
Again, is not necessarily to have everything implemented by force. For example the PXP can simply use the same BASIC implementation for
```
uint32_t lv_draw_buf_width_to_stride(uint32_t w, lv_color_format_t color_format); //For stride calculation
void * lv_draw_buf_get_buf(lv_draw_buf_t * draw_buf);   //For buffer alignment
```
Why should I duplicate the code if I don't have to. I am only trying to simplify things.
This is why I suggested to have the same implementation way as before:
```
lv_layer_t * lv_draw_pxp_layer_init(lv_disp_t * disp)
{
    lv_pxp_layer_t * pxp_layer = (lv_pxp_layer_t *)lv_draw_sw_layer_init(disp);

    pxp_layer->buffer_copy = lv_draw_pxp_buffer_copy;

    return pxp_layer;
}
```
In this example only buffer_copy is different. Ok, you can rename the `layer` keyword to `draw_buffer`. Just my thought.

> Can you confirm that both PXP and VGLite has a stride parameter which can be set freely and the stride is not calculated from the width and color format automatically?
>
Yes, I confirm that both PXP and VGLite drivers use the stride as a value given in bytes.
```
typedef struct _pxp_as_buffer_config
{
    pxp_as_pixel_format_t pixelFormat; /*!< AS buffer pixel format. */
    uint32_t bufferAddr;               /*!< Input buffer address. */
    uint16_t pitchBytes;               /*!< Number of bytes between two vertically adjacent pixels. */
} pxp_as_buffer_config_t;
```
```
    typedef struct vg_lite_buffer {
        int32_t width;                  /*! Width of the buffer in pixels. */
        int32_t height;                 /*! Height of the buffer in pixels. */
        int32_t stride;                 /*! The number of bytes to move from one line in the buffer to the next line. */
        .....
    } vg_lite_buffer_t;
````

**Comment by kisvegabor** [comment]
> Is the area all the time starting from the buf address? e.g. is the area all the time at the beginning of the buffer or the coordinates can be anywhere is the buffer?

Can be anywhere.


> Again, is not necessarily to have everything implemented by force.

Okay, we can use the callback way, but I don't know where to attach these callbacks. These are related to the whole system not just a display or a layer. What about a single global `lv_draw_buf_dsc_t` structure to store the callback, and the `lv_draw_buf_...()` functions are just wrappers to call e.g.  `lv_draw_buf_dsc.malloc_cb()`?

**Comment by nicusorcitu** [comment]
> Can be anywhere.
>
OK, in this case we really need an extra argument with the area. Or, the below structure shall have an area, with coordinates instead of width and height:
```
typedef struct {
    void * buf;
    lv_coord_t width;
    lv_coord_t height;
    lv_color_format_t color_format;
    uint32_t usage_cnt;
    uint32_t flags;
} lv_draw_buf_t;
```

> What about a single global lv_draw_buf_dsc_t structure to store the callback, and the lv_draw_buf_...() functions are just wrappers to call e.g. lv_draw_buf_dsc.malloc_cb()?
>
It looks good to me. My point here is to avoid duplicating the code. The basic implementation covers the stride and memory alignment by simply changing the defines:
```
/*Align the stride of all layers and images to this bytes*/
#define LV_DRAW_BUF_STRIDE_ALIGN                1

/*Align the start address of draw_buf addresses to this bytes*/
#define LV_DRAW_BUF_ALIGN                       4
```

**Comment by nicusorcitu** [comment]
BTW, please put this piece of code to a separate function so that we can call it from every unit dispatcher:
https://github.com/lvgl/lvgl/blob/af14c3b324fbbc07622eec5ea55222eb12056dcc/src/draw/sw/lv_draw_sw.c#L86

**Comment by kisvegabor** [comment]
@nicusorcitu 
Draw buf is updated here: 45ff2a663
`lv_draw_layer_alloc_buf` is added here 7899f77ee

**Comment by nicusorcitu** [comment]
Thank you!

**Comment by nicusorcitu** [comment]
Hi @kisvegabor 
In acceleration code, every time I want to draw to specific area, I translate the absolute coordinates to buffer relative coordinates.
How do I do that since in the new draw_buf, there are no coordinates anylonger, just width or hight? Or there is no need anymore for this kind of translation?

https://github.com/lvgl/lvgl/blob/1c5df6c665b0d5f6f346d45b13109ff1f6ea78a6/src/draw/nxp/vglite/lv_draw_vglite.c#L177
OR
https://github.com/lvgl/lvgl/blob/1c5df6c665b0d5f6f346d45b13109ff1f6ea78a6/src/draw/nxp/vglite/lv_draw_vglite.c#L336

**Comment by kisvegabor** [comment]
> Is it still need to do this kind of translations? If yes, do I have to use draw_buf_ofs since there is no more blend_area associated to draw_buf?

Yes it's still required and you need to use `draw_buf_ofs`. Here is an example just to be sure we mean the same thing:
the `draw_buf_ofs` is `100; 200`, it's size is 300x400, and you would like to draw an 50x100 rectangle to `150; 250`.

So rectangle's absolute coordinates are `150;250   200;350`. To know where to draw it on the `draw_buf` you need to make this area relative to the draw buf, by calling `lv_area_move(&rect_area, -draw_buf_ofs.x, -draw_buf_ofs.y)` which will result in `50;50  100;150`.

**Comment by nicusorcitu** [comment]
Ok, this is exactly what I mean. I will use `draw_buf_ofs`. Thanks.

**Comment by nicusorcitu** [comment]
When processing LV_DRAW_TASK_TYPE_IMAGE for example, how do I check if the source image is aligned to my stride?
There is no stride field on `lv_img_header_t`.

**Comment by XuNeo** [comment]
> There is no stride field on `lv_img_header_t`

The decoded image must be already aligned. Thus no need to add stride to image header.

But I suggest adding `stride` to `lv_img_header_t` which will make debug easier, we have faced lots of align issue when debuging on v8. If memory is a concern, then we can add a config to remove it.

**Comment by nicusorcitu** [comment]
`LV_DRAW_TASK_TYPE_IMAGE`:
> The decoded image must be already aligned. Thus no need to add stride to image header.
>
But I need to double check that, it can not be aligned.

Same shall happen on below tasks:
`LV_DRAW_TASK_TYPE_BG_IMG` - there shall be a stride argument on image header too.
`LV_DRAW_TASK_TYPE_LABEL` - the buffer for bitmap letter from [here](https://github.com/lvgl/lvgl/blob/b92cfe801d64bcd28346984b8c083ea9ff1fbf06/src/draw/lv_draw_label.c#L404) need to be also stride aligned.

**Comment by kisvegabor** [comment]
We still need to update the image decoders to decompress images in a buffer which considers stride. As a proof of concept I'll do it with the PNG decoder.

It needs to be done for label drawing too. I'll update it this week.

> But I need to double check that, it might not be aligned.

Right now we have single source for knowing the stride: `lv_draw_buf_width_to_stride`. I really like it because it's consistent. If we add stride to the image as well, we need to decide if we shall use the `image.header.stride` or `lv_draw_buf_width_to_stride()`.

We had discussion with @XuNeo and had this idea for the C array images:
```c
//color format: RGB565
//width: 10px (20 bytes)

//This header is added by the image converter
//In the image converter the requirements defined in lv_draw_buf_width_to_stride are hardcoded, 
//so it knows the strie requirements of all supported vendors' for all color formats. 
#if PLATFORM == NXP  //16 byte stride alignement 
  #define PAD    0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00,  // 20  + 12 = 32byte  
#elif PLATFORM == VENDOR_X  //8 byte stride alignement 
  #define PAD    0x00, 0x00, 0x00, 0x00 // 20  + 4 = 24byte  
#elif PLATFORM == VENDOR_Y  //4 byte stride alignement 
  #define PAD    // 20  + 0 = 20byte  
#elif PLATFORM == CUSTOM  //8 byte stride alignement (set in the image converter)
  #define PAD    0x00, 0x00, 0x00, 0x00 // 20  + 4 = 24byte  
#endif

const uint8_t buf[] = {
    0xaa, 0xaa,  ... 10 px in total... ,0xaa, 0xaa, PAD,  //PAD ensures the stride alignment
    0xaa, 0xaa,  ... 10 px in total... ,0xaa, 0xaa, PAD,
    0xaa, 0xaa,  ... 10 px in total... ,0xaa, 0xaa, PAD,
    0xaa, 0xaa,  ... 10 px in total... ,0xaa, 0xaa, PAD,
    0xaa, 0xaa,  ... 10 px in total... ,0xaa, 0xaa, PAD,
} 
```

So if anyone converts any image to C array, the stride will be good for NXP automatically. 

However I understand it it would be even more foolproof if the images had a stride parameter and the above `#if` magic is rather hacky.

Actually the important thing is for the buffers created at runtime we should always use `lv_draw_buf_width_to_stride` because we don't know which draw unit will use that buffer so it needs to be compatible with all of them.

However if an image's stride is not compatible with a draw unit, simple it won't take it. So I'm ok with adding stride into image header and always use it instead of `lv_draw_buf_width_to_stride`. In this case the stride should be set by the user when the image is converted and there is no `#if ... #define PAD` in the converted images.

Do you agree with that?

nicusorcitu:
> Right now we have single source for knowing the stride: lv_draw_buf_width_to_stride. I really like it because it's consistent. If we add stride to the image as well, we need to decide if we shall use the image.header.stride or lv_draw_buf_width_to_stride().
>
My understanding is that `img_dsc->header.stride` is the image stride while the `lv_draw_buf_width_to_stride()` is a method (platform dependent) to calculate the expected stride based on image width and cf.

> However if an image's stride is not compatible with a draw unit, simple it won't take it. 
>
I will give one example:
While taking the `LV_DRAW_TASK_TYPE_IMAGE` I would like to check if the image has stride alignment:
E.G. 
```
if (img_dsc->header.stride == lv_draw_buf_width_to_stride(img_dsc->header.w, img_dsc->header.cf))
    return stride_ok;
```
The C array is generated by an external tool, independent of the library.

Same stride parameter need to exist also on the header for `LV_DRAW_TASK_TYPE_BG_IMG' to be checked before taking the task.

And for the `LV_DRAW_TASK_TYPE_LABEL` - the buffer for bitmap letter from [here](https://github.com/lvgl/lvgl/blob/b92cfe801d64bcd28346984b8c083ea9ff1fbf06/src/draw/lv_draw_label.c#L404) need to be also stride aligned to stride, e.g. to have some padding otherwise the GPU will fail. For this case I can not check the padding prior of taking the LABEL task.

XuNeo:
I vote for adding stride to lv_img_header_t. As mentioned above, add a config to remove it if memory is a concern.

For using GPU to draw letter, our test result shows that GPU is worse than CPU when there's need to deal with stride realignment. Is that same for NXP?

For font, is it suitable to change the lv_font_glyph_dsc_t ->bpp to same as image format `cf` by extending L8 to include L1/2/4? And if it's possible, we can have colored font.

kisvegabor:
Ok, I'll add the stride to the image header and update the SW render accordingly.

Similarly to image decoders the font engines should also create a buffer considering  `lv_draw_buf_width_to_stride`. I'll update the built in font engine accordingly today or tomorrow.

 > For font, is it suitable to change the lv_font_glyph_dsc_t ->bpp to same as image format cf by extending L8 to include L1/2/4? And if it's possible, we can have colored font.

I've kept the A8 format only to keep things simpler. I wanted to avoid dealing with A1/2/4 blending support. In the built in engine I convert everything to A8 [here](https://github.com/lvgl/lvgl/blob/master/src/font/lv_font_fmt_txt.c#L89).

nicusorcitu:
> For using GPU to draw letter, our test result shows that GPU is worse than CPU when there's need to deal with stride realignment. Is that same for NXP?
>
I do no want to rework each bitmap buffer and padding it to match the stride alignment. If multiple letters can be processed by he GPU then the result might be satisfying. Yet I do not have the results right now because currently one one letter is blit correctly in my test (the one that luckily have 16 bytes width).

kisvegabor:
FYI, I added the image header to `lv_draw_img_dsc_t` so all image data will be available for the draw_units to decide if they are interested in an image or not. 19c9afb57

kisvegabor:
> If multiple letters can be processed by he GPU then the result might be satisfying. Yet I do not have the results right now because currently one one letter is blit correctly in my test (the one that luckily have 16 bytes width).

I scheduled to update it tomorrow.

nicusorcitu:
[The commits](https://github.com/lvgl/lvgl/pull/4414#issuecomment-1693922049) have an issue somewhere. I am testing CPU only and I see the black screen, I am still debugging. It has to do with the display buffers.

nicusorcitu:
I believe so, It happens after I updated to this head commit: 538bae5de8dc5d012188d48421168be634c9024b

kisvegabor:
Which `LV_DISP_RENDER_MODE_*` do you use?

nicusorcitu:
```
    /*------------------------------------
     * Create a display and set a flush_cb
     * -----------------------------------*/
    lv_disp_t * disp = lv_disp_create(DEMO_BUFFER_WIDTH, DEMO_BUFFER_HEIGHT);

    /*Set up the functions to access to your display*/
    lv_disp_set_flush_cb(disp, DEMO_FlushDisplay);

    /*
     * Two buffers screen sized buffer for double buffering.
     * Both LV_DISP_RENDER_MODE_DIRECT and LV_DISP_RENDER_MODE_FULL works, see their comments.
     */
    memset(s_frameBuffer, 0, sizeof(s_frameBuffer));
    lv_disp_set_draw_buffers(disp, s_frameBuffer[0], s_frameBuffer[1], DEMO_FB_SIZE, LV_DISP_RENDER_MODE_FULL);
```

nicusorcitu:
I have a unit test that draws a simple rectangle, it seems the very first task to fill the screen with white color does not work, even it gets executed. The next task to draw the rectangle works, but the background stays black.
Initially I figured out this with the GPU that get stuck at exactly the same first task - to fill the screen with white. Then I test with CPU only and saw the problem is persistent.

kisvegabor:
FULL mode is working for me. Does it crash, or running but black screen? If you find the issue, I'll add a test for that (if applicable) .

kisvegabor:
We can have call now to debug it together is you wish.

nicusorcitu:
Yes, I am on the desk. Please send me an invite in teams. nicusor.citu@nxp.com

kisvegabor:
I'm using Google Meet, I hope it also works for you. I sent the invite.

kisvegabor:
I added `img.header.stride` in c3e212061.

For backward compatibility if it's `0` LVGL sets it to `header.w * px_size` in the decoder `info_cb`.

kisvegabor:
Glyph stride handling is added here: 0738e0472

I couldn't test it much yet, but it should work. :slightly_smiling_face:

kisvegabor:
@XuNeo I'v just realized that the binary images needs to be updated as the `img.header` was changed.

nicusorcitu:
Hi @kisvegabor 
There is still an issue somewhere upstream with the black screen. I am running the benchmark app CPU only and the objects are moving on the black screen. Only the last results are displayed on light gray background.

kisvegabor:
On our meeting we have found that if you have only a empty screen the color is correct.  Can you try gradually adding the parts of teh benchmark to see which one causes the problem?

kisvegabor:
@nicusorcitu I found the issue of black background in the benchmark and fixed it in 7e923d2a2
So it was on our side, sorry. :slightly_smiling_face:

nicusorcitu:
Thank you!
Trimis: Friday, September 1, 2023 3:11:30 PM
Către: lvgl/lvgl ***@***.***>
Cc: Nicușor Cîțu ***@***.***>; Mention ***@***.***>
Subiect: [EXT] Re: [lvgl/lvgl] feat(tool): PNG to lvgl image in python (PR #4414)

@nicusorcitu<https://github.com/nicusorcitu> I found the issue of black background in the benchmark and fixed it in 7e923d2<https://github.com/lvgl/lvgl/commit/7e923d2a26deffc6a85fdcc5d76b43cb291a1872>
So it was on our side, sorry. 🙂

kisvegabor:
@nicusorcitu Both should be fixed 3e4cad655 and af959cb94. Note that the `draw_buf` API was slightly updated.

nicusorcitu:
We need the color format as argument in `buf_malloc` and `aligne_buf`.
In my case, `LV_DRAW_BUF_ALIGN` is not a fixed value, it depends of the color format.

https://github.com/lvgl/lvgl/blob/378ad2492ceb03679e98b51bd5e2bdfcc8df8588/src/draw/lv_draw_buf.c#L156
https://github.com/lvgl/lvgl/blob/378ad2492ceb03679e98b51bd5e2bdfcc8df8588/src/draw/lv_draw_buf.c#L169

kisvegabor:
@nicusorcitu added here: 4068c079a

nicusorcitu:
And it works, text can be done by vglite using bitmap blits. Thank you for your support on adding the stride buffer.

nicusorcitu:
I am almost done with the support for PXP and VGLite (beta version - even I see the features are working).
I will prepare a PR upstream in the next weeks.

nicusorcitu:
https://github.com/lvgl/lvgl/blob/140369c5c0b1ee7e0997ea261b71daec7fde88fd/src/draw/sw/lv_draw_sw_img.c#L294
Above you need to use the draw_buf allocator, even is the CPU code. Later on the usage of `lv_draw_buf_width_to_stride` will wrongly shift the buffer and will crash.
The idea is that if you enable both CPU and GPU, there will be a single version of draw_buf callbacks. So CPU will still have the buffers aligned same as the GPU.

W-Mai:
compile any compilable languages codes to wasm-unkown-unknown. target file likes `test.wasm`

this codes contains two functions like `func0` and `func1`

used in web, import css and js like

```html
    <!-- PyScript CSS -->
    <link rel="stylesheet" href="https://pyscript.net/snapshots/2023.09.1.RC2/core.css">
    <!-- This script tag bootstraps PyScript -->
    <script type="module" src="https://pyscript.net/snapshots/2023.09.1.RC2/core.js"></script>
```

then add scripts like this:

```html
<script>
    WebAssembly.instantiateStreaming(fetch('test.wasm'))
        .then(obj => {
                wasm_exports = {
                    func0: obj.instance.exports.func0,
                    func1: obj.instance.exports.func1,
                };
            }
        );
</script>
```

then, you can use python codes like this:

```python
<py-script>
    import sys
    from pyscript import display
    from js import wasm_exports

    display(sys.version)

    wasm_exports.func0()
	wasm_exports.func1()

</py-script>
```

total codes be like:

```python
<!doctype html>
<html>
<head>
    <!-- Recommended meta tags -->
    <meta charset="UTF-8">
    <meta name="viewport" content="width=device-width,initial-scale=1.0">
    <!-- PyScript CSS -->
    <link rel="stylesheet" href="https://pyscript.net/snapshots/2023.09.1.RC2/core.css">
    <!-- This script tag bootstraps PyScript -->
    <script type="module" src="https://pyscript.net/snapshots/2023.09.1.RC2/core.js"></script>
</head>
<body>

<!-- your code goes here... -->

<script>
    WebAssembly.instantiateStreaming(fetch('test.wasm'))
        .then(obj => {
                wasm_exports = {
                    func0: obj.instance.exports.func0,
                    func1: obj.instance.exports.func1,
                };
            }
        );
</script>

<py-script>
    import sys
    from pyscript import display
    from js import wasm_exports

    display(sys.version)

    wasm_exports.func0()
    wasm_exports.func1()

</py-script>

</body>
</html>
```

kisvegabor:
Thank you! As we have discussed on our meeting, when you will have some time for it, please update the online test with
- uploading an image and setting it's name
- sending the uploaded image to the python script
- start the conversation
- get back the converted image by JS
- download the result

XuNeo:
Hi Gabor,
The python tool is now considered finished. Stride align and RGB565A8 special stride process are added in last commit.
Could you check if it's ready to merge?

For the online version tool, we can create a standalone issue to track it.

kisvegabor:
I have tested it a lot e few days ago and it worked very well. :+1: 

I merge it now, thank you!

XuNeo:
Added `background` option for color formats that have no alpha channel. Default background is black, it will erase text inside below test image.

test-image

kisvegabor:
It's a great coincidence that I forgot to push the merge button :slightly_smiling_face:

espzav:
@XuNeo Please, is there similar script in LVGL8 release package? I cannot find it.

W-Mai:
> @XuNeo Please, is there similar script in LVGL8 release package? I cannot find it.
> 
> I would like to make script for auto-generating C array images during build. I have it for LVGL9 and I found only JS script for LVGL8 and it is not good for cmake.

Here is a python script for lvgl image converter :

https://github.com/W-Mai/lvgl_image_converter

espzav:
@W-Mai Thank you very much!

PR #4418, opened 2023-07-29
feat(global): support multi-instance feature using a single global variable lv_global

bjsylvia:

### Checkpoints
 
### DIFF
--- a/lv_conf_template.h
+++ b/lv_conf_template.h
@@ -138,12 +138,6 @@
     #define LV_OS_CUSTOM_INCLUDE <stdint.h>
 #endif
 
-/* Declare the keyword for thread-local storage. Possible options:
- * -
- * - __thread
- */
-#define LV_THREAD_LOCAL     
-
 /*=======================
  * FEATURE CONFIGURATION
  *=======================*/
@@ -253,6 +247,9 @@
 #if LV_ENABLE_GC != 0
     #define LV_GC_INCLUDE "gc.h"                           /*Include Garbage Collector related things*/
 #endif /*LV_ENABLE_GC*/
+
+/*For custom `lv_global_default()` implementation set to 1*/
+#define LV_GLOBAL_CUSTOM 0
 
 /*Default image cache size. Image caching keeps some images opened.
  *If only the built-in image formats are used there is no real advantage of caching.
--- a/lv_conf_template.h
+++ b/lv_conf_template.h
@@ -68,17 +68,6 @@
 /*Default display refresh, input device read and animation step period.*/
 #define LV_DEF_REFR_PERIOD  33      /*[ms]*/
 
-/*Use a custom tick source that tells the elapsed time in milliseconds.
- *It removes the need to manually update the tick with `lv_tick_inc()`)*/
-#define LV_TICK_CUSTOM 0
-#if LV_TICK_CUSTOM
-    #define LV_TICK_CUSTOM_INCLUDE "Arduino.h"         /*Header for the system time function*/
-    #define LV_TICK_CUSTOM_SYS_TIME_EXPR (millis())    /*Expression evaluating to current system time in ms*/
-    /*If using lvgl as ESP32 component*/
-    // #define LV_TICK_CUSTOM_INCLUDE "esp_timer.h"
-    // #define LV_TICK_CUSTOM_SYS_TIME_EXPR ((esp_timer_get_time() / 1000LL))
-#endif   /*LV_TICK_CUSTOM*/
-
 /*Default Dot Per Inch. Used to initialize default sizes such as widgets sized, style paddings.
  *(Not so important, you can adjust it to modify default sizes and spaces)*/
 #define LV_DPI_DEF 130     /*[px/inch]*/
--- a/lv_conf_template.h
+++ b/lv_conf_template.h
@@ -138,6 +138,11 @@
     #define LV_OS_CUSTOM_INCLUDE <stdint.h>
 #endif
 
+/* Declare the keyword for thread-local storage. Possible options:
+ * -
+ * - __thread
+ */
+#define LV_THREAD_LOCAL     
 
 /*=======================
  * FEATURE CONFIGURATION
--- a/src/lv_init.c
+++ b/src/lv_init.c
@@ -6,8 +6,6 @@
 /*********************
  *      INCLUDES
  *********************/
-#include <stdlib.h>
-
 #include "core/lv_global.h"
 #include "core/lv_obj.h"
 #include "disp/lv_disp_private.h"
@@ -55,14 +53,14 @@ static inline void lv_global_init(lv_global_t * global)
     LV_ASSERT_NULL(global);
 
     if(global == NULL) {
+        LV_LOG_ERROR("lv_global cannot be null");
         return;
     }
 
-    memset(global, 0, sizeof(lv_global_t));
+    lv_memset(global, 0, sizeof(lv_global_t));
 
     global->style_refresh = true;
     global->layout_count = _LV_LAYOUT_LAST;
-    global->memory_zero = ZERO_MEM_SENTINEL;
     global->style_last_custom_prop_id = (uint16_t)_LV_STYLE_LAST_BUILT_IN_PROP;
     global->area_trans_cache.angle_prev = INT32_MIN;
     global->event_last_register_id = _LV_EVENT_LAST;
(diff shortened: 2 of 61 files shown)


lv_conf_template.h
```diff
@@ -138,6 +138,11 @@
     #define LV_OS_CUSTOM_INCLUDE <stdint.h>
 #endif
 
+/* Declare the keyword for thread-local storage. Possible options:
```
```suggestion
/* Declare the keyword for thread-local storage. For example:   __thread
```

src/lv_init.c
```diff
@@ -52,6 +55,83 @@ static bool lv_initialized = false;
...
+
+    if(global == NULL) {
+        return;
+    }
+
+    memset(global, 0, sizeof(lv_global_t));
+
+    global->style_refresh = true;
+    global->layout_count = _LV_LAYOUT_LAST;
+    global->memory_zero = ZERO_MEM_SENTINEL;
```
kisvegabor: We can initialize it `lv_mem_init()`

src/lv_init.c
```diff
@@ -52,6 +55,83 @@ static bool lv_initialized = false;
...
  **********************/
+static inline void lv_global_init(lv_global_t * global)
+{
+    LV_ASSERT_NULL(global);
+
+    if(global == NULL) {
+        return;
+    }
+
+    memset(global, 0, sizeof(lv_global_t));
```
kisvegabor:
```suggestion
    lv_memzero(global, sizeof(lv_global_t));
```

src/lv_init.c
```diff
@@ -52,6 +55,83 @@ static bool lv_initialized = false;
...
+    lv_global_t * data;
+
+    if((data = (lv_global_t *)pthread_getspecific(lv_global_key)) == NULL) {
+        data = lv_global_create();
+        LV_ASSERT_NULL(data);
+
+        (void) pthread_setspecific(lv_global_key, data);
+    }
+    return data;
+}
```
kisvegabor: I wonder if we can move this part out of LVGL and have
```c
void lv_init(void); //Allocate an lv_global_t and save its pointer.  
```
and 
```c
//Need a better name for this function
void lv_init_instance(LV_THREAD_LOCAL lv_global_t * global); //Just save the provided pointer
```

src/lv_init.c
bjsylvia: only remove memory_zero to lv_mem_init, Right?

src/lv_init.c
bjsylvia: We can move this part out of LVGL. But I am not quite clear on how lv_init_instance supports multiple instances.

My idea is to have LVGL provide a pair of  new and delete `lv_global_t`  method. For example:
```
lv_global_t * lv_global_new(void);  // Allocate an lv_global_t for custom porting layer
void lv_global_delete(lv_global_t *);  // Delete an lv_global_t
```
When `LV_THREAD_LOCAL` is not defined in LVGL, it means that the porting layer needs to implement the `lv_global_default` method to support multiple instances using `lv_global_new` and `lv_global_delete`.

src/lv_init.c
kisvegabor: Yes!

src/lv_init.c
kisvegabor: Can we do this?
```c
lv_global_t * g = malloc(sizeof(lv_global_t));
//Any magic with `g`
lv_init_instance(g);

...

lv_deinit();
lv_free(g);
//Extra clean up if needed
```

This way LVGL downs't have to know anything about `LV_THREAD_LOCAL`.

bjsylvia: From your example, I understand that the `malloc` and `free` functions of the `lv_global_t` structure do not necessarily have to be provided by LVGL, but the initialization of the structure is necessary. Besides initialization, how does `lv_init_instance` save the pointer to achieve binding with the current thread? 

Do you mean saving to a variable with the `__thread` keyword? AFAIK, in some systems, this keyword may not be supported. It may be necessary to use some platform-specific APIs or libraries to achieve thread-local storage. This means that the implementation of `lv_global_default` may vary depending on the system. we need to support custom implementations and cannot define a standard implementation approach in LVGL.

kisvegabor: I see, we really need `LV_THREAD_LOCAL`, but can this be enough?
```c
static LV_THREAD_LOCAL lv_global_t * lv_global;
void lv_init_instance(lv_global_t * g) //Normal pointer, not thread specific
{ 
   lv_global = g; //Save in LV_THREAD_LOCAL prefixed pointer
   lv_init();
}

void lv_init(void)
{
  if(lv_global == NULL) lv_global = lv_malloc(sizeof(lv_global_t));
  
   ...
}
```

What I don't really understand is the need for the pthread related part in Line 91-132 when LV_THREAD_LOCAL is not defined. Can you tell why is it required inside LVGL?

bjsylvia: The following description is based on the absence of `__thread `support.

In your example, a static `lv_global_t*` is defined, which behaves as follows in systems without the concept of processes. When thread1 is initialized, `lv_global `is assigned the value addr1, but when thread2 is initialized, `lv_global `is assigned the value addr2. At this point, when thread1 tries to get `lv_global`, it will return addr2 instead of addr1. This is not the expected behavior.

The codes related part in Line 91-132 implements the ability to obtain an `lv_global_t` instance that is associated with the current thread in different threads. We don't have to implement this code in LVGL. LVGL just needs to allow the use of an externally implemented `lv_global_default `method to get lv_global_t*.

for example:
```
//demo.c

lv_global_t * lv_global_default(void) {
...
}

int demo_entry()
{
    lv_global_t * g = lv_global_default();
    //Any magic with `g`
    lv_init_instance(g);

    ...

    lv_deinit();
    lv_free(g);
}
```

```
//lvgl
lv_global_t * lv_global_default(void);

void lv_init_instance(lv_global_t * g) 
{ 
   // init g

   lv_init();
}

```

kisvegabor: My assumption was that if we add the `LV_THREAD_LOCAL` prefix to `lv_global_t * lv_global` it will mean the address saved by its thread (`addr1` or `addr2` based on which thread asks it). If it's not the case we can really 
- implement a simple non threaded `lv_global_default()` is `LV_THREAD_LOCAL` is no defined or `0` 
- let the user implement  `lv_global_default()` is `LV_THREAD_LOCAL` defined.

If we are on the same page now, could you update the PR? :slightly_smiling_face:

bjsylvia: I suggest using `LV_GLOBAL_CUSTOM` to indicate whether the user wants to customize the `lv_global_default()` implementation. `LV_THREAD_LOCAL` is only used to identify the `lv_global` variable in the default `lv_global_default`.
- implement a simple `lv_global_default()` if `LV_GLOBAL_CUSTOM` is no defined.
- let the user implement `lv_global_default()` if `LV_GLOBAL_CUSTOM` defined.

If the user system supports `__thread` in this scenario, then `#define LV_THREAD_LOCAL __thread` allows for the direct use of the default implementation of `lv_global_default`.

What do you think? If possible, I will update the PR later.

kisvegabor: `LV_GLOBAL_CUSTOM` sounds good!

> If the user system supports __thread in this scenario, then #define LV_THREAD_LOCAL __thread allows for the direct use of the default implementation of lv_global_default.

If `__thread` is not required internally, I'd leave it to the user and mention some use cases options in the documentation instead. Actually I think it'd be enough to document these two use cases with `__thread` and `pthread_key`. If I understand correctly these are independent solutions. If `__thread` is supported, it's great as it's simpler, but if not there is a workaround using `pthread_key`. Can you confirm it?

bjsylvia: Yes. So you mean you prefer to leave both options for the user to implement, right? LVGL only keeps the following implementation internally.
```
lv_global_t * lv_global_default(void)
{
    static lv_global_t lv_global;
    return &lv_global;
}
```

kisvegabor: Yes, but to be more precise:
```
#ifndef LV_GLOBAL_CUSTOM
lv_global_t * lv_global_default(void)
{
    static lv_global_t lv_global;
    return &lv_global;
}
#endif
```

bjsylvia: Do we need to add config for `LV_GLOBAL_CUSTOM` in Kconfig?

bjsylvia: > Yes, but to be more precise:
> 
> ```
> #ifndef LV_GLOBAL_CUSTOM
> lv_global_t * lv_global_default(void)
> {
>     static lv_global_t lv_global;
>     return &lv_global;
> }
> #endif
> ```

PR has updated.

kisvegabor:
Thank you, I've just merged it to `feat/multi-instance` branch. I'll fix the CI issues tomorrow and merge to `master`.

PR #4420, opened 2023-07-31
feat(qrcode): add QR code scan support

lile7:

### Checkpoints
 
### DIFF
--- a/Kconfig
+++ b/Kconfig
@@ -1014,6 +1014,9 @@ menu "LVGL configuration"
 		config LV_USE_QRCODE
 			bool "QR code library"
 
+		config LV_USE_QRSCAN
+			bool "QR scan library"
+
 		config LV_USE_BARCODE
 			bool "Barcode library"
 
--- a/lv_conf_template.h
+++ b/lv_conf_template.h
@@ -583,6 +583,9 @@
 /*QR code library*/
 #define LV_USE_QRCODE 0
 
+/*QR scan library*/
+#define LV_USE_QRSCAN 0
+
 /*Barcode code library*/
 #define LV_USE_BARCODE 0
 
--- a/src/libs/qrscan/lv_qrscan.c
+++ b/src/libs/qrscan/lv_qrscan.c
@@ -1,0 +1,187 @@
+/*
+ * Copyright (C) 2023 Xiaomi Corporation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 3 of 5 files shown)


kisvegabor:
Thank you.

I'm thinking about if such thing should be part of a graphics library. Displaying a QR code widget is ok, but IMO processing a QR code has no connection to a UI.

Why do you think we should add it?

xiaoxiang781216:
> Thank you.
> 
> I'm thinking about if such thing should be part of a graphics library. Displaying a QR code widget is ok, but IMO processing a QR code has no connection to a UI.
> 
> Why do you think we should add it?

LVGL may add camera/video support in the future, do you think QR code scan is useful in this case?

PGNetHun:
> Thank you.
> 
> I'm thinking about if such thing should be part of a graphics library. Displaying a QR code widget is ok, but IMO processing a QR code has no connection to a UI.
> 
> Why do you think we should add it?

I agree, there is no place of QR scanner code in a graphics library. 
It should be "shipped" as a separate library, and not part of LVGL.

kisvegabor:
I think it's useful, but I still can't see why in a UI library. If we add a QR code scanner we could add a face recognition module, people counter, object recognition, image quality enhancer, etc too. However these are related to the data processing and the user interface.

xiaoxiang781216:
> I think it's useful, but I still can't see why in a UI library. If we add a QR code scanner we could add a face recognition module, people counter, object recognition, image quality enhancer, etc too. However these are related to the data processing and the user interface.

The key point is whether LVGL wants to support camera feature(e.g. preview widget) in the future. If it's yes, QR scanner is a very good camera demo.

kisvegabor:
I agree that it would be a good demo, but if we add it we need to document, maintain, support it, handle feature requests, etc. And we have enough things to maintain :slightly_smiling_face:

xiaoxiang781216:
here is a lvgl camera: https://www.youtube.com/watch?v=g9gcwbCFhIs. It will be more useful with qrscan.

kisvegabor:
I agree, but I still don't understand why should it part of LVGL? 100ask could add the QR scan feature very easily anyway.

These are my main conerns:
>  If we add a QR code scanner we could add a face recognition module, people counter, object recognition, image quality enhancer, etc too. However these are related to the data processing and not the user interface.
> if we add it we need to document, maintain, support it, handle feature requests, etc. And we have enough things to maintain

xiaoxiang781216:
If so, I think https://github.com/lvgl/lvgl/tree/master/src/libs/barcode and https://github.com/lvgl/lvgl/tree/master/src/libs/qrcode should remove from lvgl repo since both aren't related to lvgl directly too.

PGNetHun:
> Clean and lean.

Sorry, but I think not so many LVGL-based projects use QR scanning & camera, maybe a small percentage. 

But I (and hopefully others also) use JPEG, PNG, fonts, QR code **for displaying**, and these became the "core" part of this GUI library, so we can use these "out-of-the-box" without adding 3rd party libraries.

**I think LVGL should be as clean and lean as it is possible, and focus only on GUI.**

Every other non-GUI related feature should be implemented by using a 3rd party library.

PR #4434, opened 2023-08-03
build(cmake): pass freetype build flags to parent

amsobr:
If using freetype, the parent CMake project will need to know the freetype include and link dirs as well as the list of freetype libraries to link to.

This commit defines an option in the CMake project - LV_USE_FREETYPE -, which, if set, makes LVGL look for the freetype library using pkgconfig, and add the include dirs, link dirs and libraries to the cmake library, so that these are automatically passed along to dependants through the PUBLIC interface.

In the example below:
```cmake
add_subdirectory( path/to/lvgl )
add_executable( myapp ... )
target_link_libraries( myapp PRIVATE lvgl:lvgl )
```

If lv_conf.h has:
```C
#define LV_USE_FREETYPE 1
```

The build will fail because the compiler will need to know about the includes and freetype libraries.

This commit fixes that and spares parent projects from having to add the freetype library by hand.

This is mostly about developer-experience. Thanks for having a look and feel free to consider including it in LVGL if that's useful.

Cheers,
António

### DIFF
--- a/env_support/cmake/custom.cmake
+++ b/env_support/cmake/custom.cmake
@@ -12,6 +12,16 @@ if( LV_CONF_PATH )
     get_filename_component(LV_CONF_DIR ${LV_CONF_PATH} DIRECTORY)
 endif( LV_CONF_PATH )
 
+find_package(PkgConfig REQUIRED)
+
+# Option LV_USE_FREETYPE
+# If set, libfreetype2 is added to the dependencies of lvgl so that dependants
+# have the compilation and linking flags automatically set.
+if( LV_USE_FREETYPE )
+    pkg_check_modules( FREETYPE freetype2>=2.0 REQUIRED )
+endif( LV_USE_FREETYPE )
+
+
 # Option to build shared libraries (as opposed to static), default: OFF
 option(BUILD_SHARED_LIBS "Build shared libraries" OFF)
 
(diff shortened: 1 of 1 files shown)


amsobr:
I see one of the failed tests happened in an environment that doesn't have pkgconfig.

In hindsight I can see this chance potentially restricting the range en environments where lvgl would be usable.

If you think that this change has potential I can improve this to keep the workflow unchanged for those cases where either pkgconfig or freetype2.pc are missing:
* look for pkgconfig only if LV_USE_FREETYPE is set and alleviate the `REQUIRED`
  *  if `PkgConfig_FOUND` is false, just print some warning and let it be.
* look for freetype2 only if pkgconfig is available and, again, don't make it required (just warn about it)

Please met me know if this would be useful. If yes I can add these tweaks to the PR.

Thanks!
António

kisvegabor:
Hi,

Thank you for the clear and detailed explanation of the issue.

I'm not a  CMake expert, so it might not make any sense, but can we do something lile (pseudo code):
```
if find_package(PkgConfig REQUIRED) == FOUND {
  if  LV_USE_FREETYPE {
    pkg_check_modules( FREETYPE freetype2>=2.0 REQUIRED )
    target_compile_options( lvgl PUBLIC ${FREETYPE_CFLAGS})
    target_link_libraries(lvgl PUBLIC ${FREETYPE_LIBRARIES})
    target_link_options(lvgl PUBLIC ${FREETYPE_LDFLAGS})
  }
}
```

amsobr:
Hi!

Also not an expert here, but yes, that's the idea I was referring to. That pseudo-code can indeed be implemented in CMake. With the detail that looking for PkgConfig is done if USE_FREETYPE is enabled to avoid build failures when someone using the library doesn't care about PkgConfig at all (the armv6 test is an example)

I'll make that change in the project and update the PR as soon.

kisvegabor:
Great, thank you! :slightly_smiling_face:

PR #4442, opened 2023-08-10
fix(fragment): catch wrong instance_size

FASTSHIFT:

fragment instance_size cannot be lower than `sizeof(lv_fragment_t)`, otherwise it will cause memory corruption.

### Checkpoints
 
### DIFF
--- a/src/others/fragment/lv_fragment.c
+++ b/src/others/fragment/lv_fragment.c
@@ -26,7 +26,7 @@ lv_fragment_t * lv_fragment_create(const lv_fragment_class_t * cls, void * args)
 {
     LV_ASSERT_NULL(cls);
     LV_ASSERT_NULL(cls->create_obj_cb);
-    LV_ASSERT(cls->instance_size > 0);
+    LV_ASSERT(cls->instance_size >= sizeof(lv_fragment_t));
     lv_fragment_t * instance = lv_malloc(cls->instance_size);
     lv_memzero(instance, cls->instance_size);
     instance->cls = cls;


kisvegabor:
Thanks!

kisvegabor:
Do you use `lv_fragment` in your projects?

FASTSHIFT:
> Do you use `lv_fragment` in your projects?

Yes, I am trying to implement a sports band demo and this is my first time using `lv_fragment`.

kisvegabor:
I'm asking because I'm thinking for long about  different approach to handle this problem. It would be composed of 3 independent things:
1. variable binding: No [`lv_msg`](https://docs.lvgl.io/8.3/others/msg.html) can be used for that, but it's quite cumbersome.  In v9 we should have functions for widget like `lv_label_bind_string(label, some_buf);`, or lv_dropdown_bind_int(dropdown, &selected); if the bound variables changes the widgets should change as well, and vice versa. It can be used like:
```c
selected = 4;
lv_update_var(&selected); //Notify all widget who are interested in it
```
We can bind the `LV_FLAG`s and `LV_STATE` as well. E.g. `lv_obj_bind_flag(obj, LV_FLAG_HIDDEN, &to_show);` It would be very helpful to manage the data between the application and the UI. (Similar to what React and other web UI frameworks do, but much simpler)
 
2. Scene (or view, I'm nor sure which is the better name): It's very simple, just store the current "scene" of any widget as `obj->scene_act = 10`; When you change scene, `LV_EVENT_SCENE_CLOSE` and `LV_EVENT_SCENE_OPEN` will be sent, so that you can decide how to leave the current scene and load the new one. You an play animations, hind or reposition things, etc. It can apply to a single widget (e.g. slider with normal, warning or error scene) or complex widgets, like a list. 

3. Back button: I'm still not sure how it should work, but we need to store a history of some kind in a generic way

I think it can replace `lv_fragment` and can be more flexible. What do you think?

FASTSHIFT:
> I'm asking because I'm thinking for long about different approach to handle this problem. It would be composed of 3 independent things:
> 
> 1. variable binding: No [`lv_msg`](https://docs.lvgl.io/8.3/others/msg.html) can be used for that, but it's quite cumbersome.  In v9 we should have functions for widget like `lv_label_bind_string(label, some_buf);`, or lv_dropdown_bind_int(dropdown, &selected); if the bound variables changes the widgets should change as well, and vice versa. It can be used like:
> 
> ```c
> selected = 4;
> lv_update_var(&selected); //Notify all widget who are interested in it
> ```
> 
> We can bind the `LV_FLAG`s and `LV_STATE` as well. E.g. `lv_obj_bind_flag(obj, LV_FLAG_HIDDEN, &to_show);` It would be very helpful to manage the data between the application and the UI. (Similar to what React and other web UI frameworks do, but much simpler)
> 
> 2. Scene (or view, I'm nor sure which is the better name): It's very simple, just store the current "scene" of any widget as `obj->scene_act = 10`; When you change scene, `LV_EVENT_SCENE_CLOSE` and `LV_EVENT_SCENE_OPEN` will be sent, so that you can decide how to leave the current scene and load the new one. You an play animations, hind or reposition things, etc. It can apply to a single widget (e.g. slider with normal, warning or error scene) or complex widgets, like a list.
> 3. Back button: I'm still not sure how it should work, but we need to store a history of some kind in a generic way
> 
> I think it can replace `lv_fragment` and can be more flexible. What do you think?

Very advanced ideas, but I think scene variable binding and fragment are responsible for different work content:
- Fragment is mainly responsible for providing a framework for controlling the UI life cycle, and standardizing the creation and deletion process of UI programs. It is concerned with the switching between pages rather than the rendering effect;
- Variable binding mainly implements data-driven UI update, and it is more related to the rendering effect on the single page.

kisvegabor:
> Fragment is mainly responsible for providing a framework for controlling the UI life cycle, and standardizing the creation and deletion process of UI programs. It is concerned with the switching between pages rather than the rendering effect;

Can't we make somehow a generic widget property/behavior? I'm mean maybe we can find a few new event codes that can be used in a flexible way to achieve something similar.

PR #4446, opened 2023-08-11
Removes unused macro

kdschlosser:
removes the following macro from lv_refr.h

```c
#define LV_REFR_TASK_PRIO LV_TASK_PRIO_MID
```

both `LV_REFR_TASK_PRIO` and  `LV_TASK_PRIO_MID` are not used in LVGL at all.

### DIFF
--- a/src/core/lv_refr.h
+++ b/src/core/lv_refr.h
@@ -21,8 +21,6 @@ extern "C" {
  *      DEFINES
  *********************/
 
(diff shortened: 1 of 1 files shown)


kisvegabor:
True, thanks!

kdschlosser:
no worries m8. if it doesn't need to be there then it should be removed yes?

PR #4449, opened 2023-08-13
Removes unused function

kdschlosser:
Function `void lv_draw_unit_sw_create(lv_disp_t * disp, uint32_t cnt);` causes linking errors because it has no body. So It should be removed.


kisvegabor:
It's already fixed in `feat/` branch, but's let's fix it on `master` too now.
 
Thank you!

PR #4454, opened 2023-08-14
lv_slider: Add API for knob_inside functionality

### LINKED ISSUE #4412: [slider] Limit knob movement inside indicator
### Introduce the problem
See [Slider Knob out of the track](https://forum.lvgl.io/t/slider-knob-out-of-the-track/11956) forum post for more information.

### Examples and cases
A feature to limit the knob in the slider is desired.

### Suggested solution
See this [proposal](https://forum.lvgl.io/t/slider-knob-out-of-the-track/11956/5)

image

- At the left we need to align the value with the left side of the knob.
- At the left we need to align the value with the right side of the knob.
- At the center we need to align the value with the center of the knob.

@kisvegabor 
Would you like me to add a new slider mode for this feature? Should the knob size be always smaller than the indicator in this mode?

C47D:

Closes #4412 

### Checkpoints
 
### DIFF
--- a/src/widgets/slider/lv_slider.c
+++ b/src/widgets/slider/lv_slider.c
@@ -77,6 +77,16 @@ bool lv_slider_is_dragged(const lv_obj_t * obj)
     return slider->dragging ? true : false;
 }
 
+void lv_slider_set_knob_inside(lv_obj_t * obj, uint8_t enable)
+{
+    LV_ASSERT_OBJ(obj, MY_CLASS);
+    lv_slider_t * slider = (lv_slider_t *)obj;
+
+    slider->knob_inside = enable;
+
+    lv_obj_invalidate(obj);
+}
+
 /**********************
  *   STATIC FUNCTIONS
  **********************/
@@ -90,6 +100,7 @@ static void lv_slider_constructor(const lv_obj_class_t * class_p, lv_obj_t * obj
     slider->value_to_set = NULL;
     slider->dragging = 0U;
     slider->left_knob_focus = 0U;
+    slider->knob_inside = 0U;
 
     lv_obj_clear_flag(obj, LV_OBJ_FLAG_SCROLL_CHAIN_HOR);
     lv_obj_clear_flag(obj, LV_OBJ_FLAG_SCROLLABLE);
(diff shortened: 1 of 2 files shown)


C47D:
After some thoughts I think this proposal is the easier to implement and use.
image

Instead of dynamically changing the point where the value is taken from.

kisvegabor:
My only concert with that is it's quite complicated to create consistently looking slider with different ranges.

C47D:
I was trying to get some examples of other UI libraries and I haven't found one with the indicator bigger than the indicator. https://m2.material.io/components/sliders#usage

kisvegabor:
How do you mean:
>  the indicator bigger than the indicator.
?

An other concern with using the slider value as a limit is when the slider's range is very small, e.g. only 0-3. In this case you can't control the end position precisely.

C47D:
Still trying to figure out how to make this work, I've found [this](https://uxplanet.org/top-15-seekbar-and-slider-github-ui-libraries-and-components-java-swift-kotlin-d0f6a9658be3) page with a bunch of examples of other UI libraries.

It kind of makes something funny when the knob (or handle) is in the top center.
1 amcDmx_WQWy12xVIvAP5jA

kisvegabor:
It really applies to arcs as well. I still believe we need explicit control in pixels. E.g. use `margin` on the KNOB to tell what distance to keep.

C47D:
Not stale

kdschlosser:
I don't remember what the exact issue was with this. If a screen shot can be provided I can shed some light on how to solve the issue.

kdschlosser:
OK I found the issue in the forum that shows the problem again. This is an easy thing to handle. The ends of the slider are half circles so there has to be a center that is specified for each of the end arcs. That is your stop. That is where the max value and the min values get attached to. It only does this is the knob is a larger radius than the end arcs and the opa level is not set to transparent for the knob. You simply have the knob center range bound to the end arc centers.

kisvegabor:
I've discussed it with our designer and concluded that we need an exact pixel based setting for it. Using the margin properties on knob sounds like a good fit for this.

kdschlosser:
pixel based is the way to go with it. The trick is this only needs to be done if the knob is visible. it has any kind of an alpha level other than 255 then it cannot be done as the indicator will need to go the full distance across. Unless the indicator is aligned at the edge of the knob instead of the center. You will have to use math to move the alignment of the indicator to knob as the knob is being moved so the alignment will shift to the other side so at the 50% mark on the slider the knob's alignment would be the center

kisvegabor:
If think we can leave it fully to the users. If they set not fully covering opa for the knob they can set the know margin accordingly too.

C47D:
I'll close this PR as I have to think about the implementation first

kisvegabor:
It could be a good candidate for the [new shaping process](https://docs.lvgl.io/master/CONTRIBUTING.html#feature-development-workflow).

Issue #4448, opened 2023-08-11
Simple new feature: load font from buffer


kisvegabor:
I think it's a local and quite straightforward feature, so if you would like contribute to LVGL, this one is a perfect candidate. If you are interested in it, just drop a line below, and we can discuss the details.

------

Continue https://github.com/lvgl/lv_binding_js/issues/15#

### Introduce the problem

Now we have `lv_font_load()` to load a font from a binary file but in some cases it would be great to have an option to load the font from a buffer as well.

### Examples and cases
- Would be useful for the [JavaScript binding](https://github.com/lvgl/lv_binding_js)
- It's faster than loading from file

### Suggested solution

Add an lv_font_load_from_buffer() function.  We could use some function pointers like `read_cb` which could point to `lv_fs_read` or a `buf_read` function. This way we don't need to add a bunch of duplication.

cc @derekstavis

rik43:
"Javascript binding" - bad link

kisvegabor:
> https://github.com/lvgl/lv_binding_cpp/issues/8#issuecomment-1675827276

Fixed, thank you :slightly_smiling_face:

zjanosy:
@derekstavis
After discussing with Gabor I am going to implement this feature.
We have identified two possible use cases:
1. Read the whole structure into RAM. This is how the current lv_load_font() function works. Adding support for this is simple.
2. Read only the small structures, but use pointers to the larger structures in the buffer. This may make sense if the font is in (not too slow) external flash, and RAM is a scarce resource. This is also possible, though it would require more substantial changes.
Which one is your use case?

derekstavis:
Hmm, I can see two main use-cases for loading from a buffer:

1. The font is compiled into code (like with `XPM` files), and passed to `lv_load_font_from_buffer` as a pointer
2. The font has been loaded by a binding into RAM and passed to `lv_load_font_from_buffer` as a pointer

The use-case (2) is the one I'm planning to use on the JS binding

kisvegabor:
FYI, a PR is sent: #4462

kdschlosser:
Being able to load a font from a buffer object would be a great idea if it allows loading the font without needing to use the fsdriver. For the Python binding it would be a lot easier to open the file and dump the contents of the file into a single function call.

kdschlosser:
@kisvegabor 
[...]

PR #4462, opened 2023-08-16
feat(fs): add support for memory-mapped files

zjanosy:
Added File System Interface driver for memory-mapped files
 
This driver allows using a memory area as a file that can be read by normal file operations. It can be used, e.g., to store font files in slow flash memory, and load them into RAM on demand.

You can enable it in lv_conf.h:
```
#define LV_USE_FS_MEMFS 1
```
The actual implementation uses the built-in cache mechanism of the file system interface.

Since this is not an actual file system, file write and directories are not supported.

The default drive letter is 'M', but this can be changed in lv_conf.h:

```
#define LV_FS_MEMFS_LETTER 'M'
```

To use it seamlessly with the file system interface a new extended path object has been introduced:

```
lv_fs_path_ex_t mempath;
```

This structure can be initialized with the helper function:

```
lv_fs_make_path_ex(&mempath, (const uint8_t *) & my_mem_buffer, sizeof(my_mem_buffer));
```

Then the "file" can be opened with:

```
lv_fs_file_t file;
lv_fs_res_t res = lv_fs_open(&file, (const char *) & mempath, LV_FS_MODE_RD);
```
The path object can be used at any place where a file path is required, e.g.:

```
lv_font_t* my_font = lv_font_load((const char *) & mempath);
```


src/libs/fsdrv/lv_fs_memfs.c
```diff
@@ -0,0 +1,233 @@
...
+static lv_fs_res_t fs_close(lv_fs_drv_t * drv, void * file_p);
+static lv_fs_res_t fs_read(lv_fs_drv_t * drv, void * file_p, void * buf, uint32_t btr, uint32_t * br);
+static lv_fs_res_t fs_seek(lv_fs_drv_t * drv, void * file_p, uint32_t pos, lv_fs_whence_t whence);
+static lv_fs_res_t fs_tell(lv_fs_drv_t * drv, void * file_p, uint32_t * pos_p);
+#if 0
+static lv_fs_res_t fs_write(lv_fs_drv_t* drv, void* file_p, const void* buf, uint32_t btw, uint32_t* bw);
+static void * fs_dir_open(lv_fs_drv_t * drv, const char * path);
+static lv_fs_res_t fs_dir_read(lv_fs_drv_t * drv, void * dir_p, char * fn);
+static lv_fs_res_t fs_dir_close(lv_fs_drv_t * drv, void * dir_p);
+#endif
```
kisvegabor: Please remove these if these as they are not supported.

src/libs/fsdrv/lv_fs_memfs.c
```diff
@@ -0,0 +1,233 @@
...
+ *
+ * #define LV_FS_MEMFS_LETTER 'M'
+ *
+ * To use it seamlessly with the file system interface a new extended path object has been introduced:
+ *
+ * lv_fs_path_ex_t mempath;
+ *
+ * This structure can be initialized with the helper function:
+ *
+ * lv_fs_make_path_ex(&mempath, (const uint8_t *) & my_mem_buffer, sizeof(my_mem_buffer));
```
kisvegabor, line 27: Do we really need the `&` before `my_mem_buffer`?

src/libs/fsdrv/lv_fs_memfs.c
```diff
@@ -0,0 +1,233 @@
...
+    fs_drv.tell_cb = fs_tell;
+
+    fs_drv.dir_close_cb = NULL; // fs_dir_close;
+    fs_drv.dir_open_cb = NULL; // fs_dir_open;
+    fs_drv.dir_read_cb = NULL; // fs_dir_read;
+
+    lv_fs_drv_register(&fs_drv);
+}
+
+void lv_fs_make_path_ex(lv_fs_path_ex_t* path, const uint8_t* buf, uint32_t size)
```
kisvegabor: By using `const void* buf` we could avoid the cast when this function is called.

src/libs/fsdrv/lv_fsdrv.h
```diff
@@ -43,6 +43,10 @@ void lv_fs_posix_init(void);
 void lv_fs_win32_init(void);
 #endif
 
+#if LV_USE_FS_MEMFS != '\0'
```
kisvegabor: Actually it's mistake with the other drivers as well. It should be:

```suggestion
#if LV_USE_FS_MEMFS
```

src/libs/fsdrv/lv_fs_memfs.c
```diff
@@ -0,0 +1,233 @@
...
+    /*---------------------------------------------------
+     * Register the file system interface in LVGL
+     *--------------------------------------------------*/
+
+    /*Add a simple drive to open images*/
+    lv_fs_drv_init(&fs_drv);
+
+    /*Set up fields...*/
+    fs_drv.letter = LV_FS_MEMFS_LETTER;
+    fs_drv.cache_size = UINT32_MAX; //LV_FS_MEMFS_CACHE_SIZE;
```
kisvegabor: Let's add a define like `#define LV_FS_CACHE_FROM_BUFFER   UINT32_MAX` or so.

src/misc/lv_fs.c
```diff
@@ -81,22 +81,35 @@ lv_fs_res_t lv_fs_open(lv_fs_file_t * file_p, const char * path, lv_fs_mode_t mo
...
-    file_p->drv = drv;
-    file_p->file_d = file_d;
 
     if(drv->cache_size) {
         file_p->cache = lv_malloc(sizeof(lv_fs_file_cache_t));
         LV_ASSERT_MALLOC(file_p->cache);
         lv_memzero(file_p->cache, sizeof(lv_fs_file_cache_t));
-        file_p->cache->start = UINT32_MAX;  /*Set an invalid range by default*/
-        file_p->cache->end = UINT32_MAX - 1;
+        if (drv->cache_size == UINT32_MAX) {    // this is a memory-mapped file: set "cache" to the memory buffer
```
kisvegabor: Please use `/* */` style comments

src/misc/lv_fs.h
```diff
@@ -110,6 +110,12 @@ typedef struct {
     lv_fs_file_cache_t * cache;
 } lv_fs_file_t;
 
+typedef struct {
+    char            path[4];   // always "M:x"
```
kisvegabor: Why 4 bytes and not 3? 
```suggestion
    char            path[3];   /* always "M:" */
```

kisvegabor (commented):
Hi,

I really like to idea of using the cache like this. I've added few comment, but nothing extraordinary. 

Maybe we can still add an `lv_font_load_from_buffer` function to make it easier to use. 

Just for the record: I thought if we used callbacks in `lv_font_load` we could save  some `memcpy` but it's not the case, as we need to copy the data from the buffer to RAM anyway. It  doesn't matter if we do it by `memcpy` or in `lv_fs_read`.

src/libs/fsdrv/lv_fs_memfs.c
zjanosy: Done

src/libs/fsdrv/lv_fs_memfs.c
zjanosy: Apparently not.

src/libs/fsdrv/lv_fs_memfs.c
zjanosy: Changed.

src/libs/fsdrv/lv_fsdrv.h
zjanosy: Shall I fix it, or do you want to fix it for all drivers?

src/libs/fsdrv/lv_fs_memfs.c
zjanosy: Done

src/misc/lv_fs.c
zjanosy: Changed

src/misc/lv_fs.h
zjanosy: For dword alignment of the other fields. Probably the compiler would add one byte automatically anyway, but this is clearer.

src/libs/fsdrv/lv_fsdrv.h
zjanosy: I've fixed it.

src/libs/fsdrv/lv_fs_memfs.c
```diff
@@ -97,23 +90,23 @@ void lv_fs_memfs_init(void)
...
-    fs_drv.dir_read_cb = NULL; // fs_dir_read;
+    fs_drv.dir_close_cb = NULL;
+    fs_drv.dir_open_cb = NULL;
+    fs_drv.dir_read_cb = NULL;
 
     lv_fs_drv_register(&fs_drv);
 }
 
-void lv_fs_make_path_ex(lv_fs_path_ex_t* path, const uint8_t* buf, uint32_t size)
+void lv_fs_make_path_ex(lv_fs_path_ex_t* path, void* buf, uint32_t size)
```
kisvegabor: I think it makes sens to use `const void *` here, as we don't write anything into the buffer.

kisvegabor (commented):
Looks good, thank you for the update. Just a minor comment.

Could you add some unit tests as well?

kisvegabor:
Thank you! 

@derekstavis, @kdschlosser can you try it out this PR with the JS and Python binding?

kisvegabor:
@zjanosy 
Please add the new config options (e.g. `LV_USE_FS_MEMFS`) in `lv_conf_template.h` and run `lv_conf_internal_gen.py ` in the script folder.

kdschlosser:
I guess I am a little bit confused by this PR.

How does the file system know where "files" are located in memory? 
How does the memory location get passed to the file system driver?
Is the user responsible for creating callbacks to retrieve the data form memory?

If the user has to pass the memory location of the data to the file system driver using the callback functions what would make that any different then how it operates now? That is the part that I am confused on.


As an example.

Say I have some data stored in program memory and I want that data to be accessible by the file system in LVGL. The location of that data in program memory can be identified via an array the user creates with the pointers to where that data is located. The user would pass a string of the index to that location as the "file name" and when the callback function registered to the file system driver gets called the data could then be loaded from program memory and put into RAM.. 

That is where my confusion is coming from.

kisvegabor:
@kdschlosser See @zjanosy's comment [here](https://github.com/lvgl/lvgl/issues/4448#issuecomment-1683967188).

derekstavis:
> can you try it out this PR with the JS and Python binding?

The JS binding is using v8. I spent some time migrating to master and should hopefully be able to test this sometime this week.

src/libs/fsdrv/lv_fs_memfs.c
derekstavis:
```suggestion
void lv_fs_make_path_ex(lv_fs_path_ex_t* path, const void* buf, uint32_t size)
```

Compilation is failing here with this message:

```
lvgl/src/libs/fsdrv/lv_fs_memfs.c:109:6: error: conflicting types for 'lv_fs_make_path_ex'
void lv_fs_make_path_ex(lv_fs_path_ex_t* path, void* buf, uint32_t size)
     ^
lvgl/src/core/../draw/../misc/lv_fs.h:182:6: note: previous declaration is here
void lv_fs_make_path_ex(lv_fs_path_ex_t* path, const void* buf, uint32_t size);
```

derekstavis:
Update: Preliminarily, it seems like it's working!

Code:

```
(async () => {
  try {
    const font = await Font.load("./assets/output.font");
    console.log(font);
  } catch (e) {
    console.error(e);
  }
})();
```

Logs:

```
[User]  (0.299, +0)      NativeCompSetStyle: View 1 setStyle type 0     (in view_wrap.cpp line #5)
[User]  (0.650, +351)    FontConstructor: Creating Font instance        (in font.cpp line #13)
[User]  (0.650, +0)      FontConstructor: New Font instance created     (in font.cpp line #36)
[User]  (0.650, +0)      FontConstructor: Font created  (in font.cpp line #48)
{}
[User]  (0.650, +0)      FontFinalizer: Font release    (in font.cpp line #57)
```

Next I will look into modifying the core API to accept a Font object and see if it renders as expected!

derekstavis:
It works but I likely got the font conversion wrong:

image

kisvegabor:
Could you try with [this font](https://github.com/lvgl/lvgl/blob/master/tests/src/test_assets/font_1.fnt)? 

image

Although I'm not a JS/React guy, but it really doesn't feel intuitive to me. Please open a new issue to discuss it.

derekstavis:
It's tiny but yaay!! It worked:

image

zjanosy:
@kisvegabor I've fixed a couple of warnings, and added a unit test.

kisvegabor:
Thank you, it looks good! There are some conflicts as your not using the latest master, but I understand that because it's not available in Visual Studio yet.

kisvegabor:
@derekstavis 
Looks cool! :sunglasses: Thank you for testing it!

kisvegabor:
I've fixed the conflicts and merged master. 

@zjanosy Only one last thing is missing: we should add a brief documentation of this new feature. 
1. Just please mention MEMFS [here](https://github.com/lvgl/lvgl/blob/master/docs/libs/fs.rst)
2. and `lv_font_load_from_buffer` [here](https://github.com/lvgl/lvgl/blob/master/docs/overview/font.rst#load-a-font-at-run-time).

kisvegabor:
All looks good! Thank you, Zoltan!

tan1435014421:
Here's the call implementation:
lv_font_t * ArialFont16;
ArialFont16 = lv_font_load(“0:fn/ArialFont16.bin”);
lv_obj_set_style_text_font(Scanlab, ArialFont16, 0);
lv_label_set_text(Scanlab, “SCAN”);

tan1435014421:
> @tan1435014421 What is the exact problem?
> 
> `"0:fn/ArialFont16.bin"` should be something like `"A:fn/ArialFont16.bin"`. See [here](https://docs.lvgl.io/8.3/overview/file-system.html).

Sorry means that I generated the wrong .bin file, which has been solved.
Now I tried to get lvgl8.3.10 version to use FreeRTOS memory allocation failed, because my single-chip microcomputer memory is not very sufficient, before I used lvgl7.11 version to implement the memory allocation using FreeRTOS, can this lvgl8.3.10 not be able to use FreeRTOS memory allocation?

kisvegabor:
v8.3 should work with FreeRTOS memory allocation too. Just update this part of the config file: https://github.com/lvgl/lvgl/blob/release/v8.3/lv_conf_template.h#L48-L67

tan1435014421:
> v8.3 should work with FreeRTOS memory allocation too. Just update this part of the config file: https://github.com/lvgl/lvgl/blob/release/v8.3/lv_conf_template.h#L48-L67

Yes, I ported the file and it failed at runtime, and my judgment was that FreeRTOS failed to add support for pvPortRealloc, because FreeRTOS memory allocation does not have this feature. Thanks for your reply.

kisvegabor:
> and my judgment was that FreeRTOS failed to add support for pvPortRealloc,

I don't know about these details :slightly_frowning_face:

PR #4473, opened 2023-08-20
fix(decoder): Fix JPEG decoder color format issue, and add image decoder (PNG, JPEG) tests

PGNetHun:

Fix JPEG decoder rendering color format issue, which caused a black box rendered instead of the decoded image.
This PR still does not fix other JPEG rendering issue, that since the parallel rendering architecture commit (https://github.com/lvgl/lvgl/commit/f753265a799bdd910881238cb36f8d731f6d8727) the JPEG image is not decoded.
Two tests are added to test PNG & JPEG decoders, but at the moment JPEG decoder test is "ignored".
Test image is taken from Wikipedia: https://en.wikipedia.org/wiki/Test_card  (_PM5544: common PAL test pattern_)

### Checkpoints
 

PGNetHun:
The JPEG test pass with the color format fix (using `LV_COLOR_FORMAT_NATIVE` in `lv_sjpg.c`) applied to commit `de97dba`, which is referenced by `lv_micropython`, and is before parallel rendering commit.
With the old color format used in `lv_sjpg.c` (`LV_COLOR_FORMAT_RAW`) the test fails.
So we just need to decode the JPEG in the new parallel rendering architecture, and enable JPEG test (in `test_image.jpg.c`)

kisvegabor:
Thank you, @PGNetHun.

In the [feat/img-decoder-get-area](https://github.com/lvgl/lvgl/tree/feat/img-decoder-get-area) branch we have added something which is important for the JPEG decoder too. 

Instead of having a `read_line_cb` in the decoder we have a `get_area_cb` callback, where the decoder can tell which area it decoded and it can do the decoding in any order. In case of JPG it means that it can decode the image in 8x8 chunks too. 

See the [BMP decoder here](https://github.com/lvgl/lvgl/blob/feat/img-decoder-get-area/src/libs/bmp/lv_bmp.c#L182) as an example, and this is how it's used in the [SW image draw](https://github.com/lvgl/lvgl/blob/feat/img-decoder-get-area/src/draw/sw/lv_draw_sw_img.c#L184-L209) function. 

I wonder if you have time an interest to clean up the JPG decoder by using the this new get area function. Ideally the custom SJPG format shouldn't be required anymore, so I hope the implementation of the JPG decoder can be much cleaner. 

Finally, it'd be great to have any option somewhere/somehow decide if the image should be decoded in chunks or as the whole image which will be cached. But it's not critical now. 

---

I see in your profile that you are also from Budapest. Could you drop me a line to gabor@lvgl.io?  Just to get to know each other :slightly_smiling_face:

PGNetHun:
Cancel PR due to `feat/img-decoder-get-area` changes.

PR #4476, opened 2023-08-21
Adds additional SDL support features

kdschlosser:
Adds functions for setting position, brightness and other various things for an SDL window. It also exposes flags for controlling various elements of the the window creation, like resizing and displaying a title bar and border.

This is an API Breaking change. There is a new parameter added to the create function. If this parameter is supplied a zero it will behave exactly like it did prior. If a -1 is supplied to the additional parameter that would be specifying no flags at all. The flags can be or'ed together.

The flags are in a new enumeration named  `lv_sdl_window_flags_t`. There is a flag that is added to center the window to the desktop and that is used as a default.

There is also a flag for high dpi support if the OS and video driver support it.

I still have to complete the documentation for the added features. I wanted to see if this is of interest before I spend the time to do that.

### DIFF
--- a/lv_conf_template.h
+++ b/lv_conf_template.h
@@ -726,6 +726,9 @@
     #define LV_SDL_BUF_COUNT       1   /*1 or 2*/
     #define LV_SDL_FULLSCREEN      0    /*1: Make the window full screen by default*/
     #define LV_SDL_DIRECT_EXIT     1    /*1: Exit the application when all SDL widows are closed*/
+    #ifndef LV_SDL_TICK_THREAD
+        #define LV_SDL_TICK_THREAD 1
+    #endif
 #endif
 
 /*Driver for /dev/fb*/
--- a/src/dev/sdl/lv_sdl_window.c
+++ b/src/dev/sdl/lv_sdl_window.c
@@ -87,69 +87,69 @@ void lv_sdl_window_set_pos(lv_disp_t * disp, lv_coord_t x, lv_coord_t y)
 lv_coord_t lv_sdl_window_get_x(lv_disp_t * disp)
 {
     lv_sdl_window_t * dsc = lv_disp_get_driver_data(disp);
-    int *x = 0;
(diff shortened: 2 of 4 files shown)


kisvegabor:
Do we really need to repeat the SDL API?

E.g.

```c
void lv_sdl_window_set_maximum_size(lv_disp_t * disp, lv_coord_t width, lv_coord_t height)
{
    lv_sdl_window_t * dsc = lv_disp_get_driver_data(disp);
    SDL_SetWindowMaximumSize(dsc->window, (int) width, (int) height);
}
```

I think we should rather create an `lv_sdl_window_get_window_dsc(disp)` (with a better name maybe) on which any SDL function can be called. We just need document this option and show some examples somewhere. 

What do you think?

kdschlosser:
wow. I didn't even know you commented on this. How odd that I didn't get a notification for it.

kdschlosser:
The reason why  I did it the way it's done is to keep with the current API for SDL functions that are already in  LVGL. I also thought it would be best to keep the API the same as other LVGL functions....

as an example..

```c

lv_coord_t lv_sdl_window_get_maximum_width(lv_disp_t * disp)
{
    lv_sdl_window_t * dsc = lv_disp_get_driver_data(disp);
    int *w = 0;
    int *h = 0;

    SDL_GetWindowMaximumSize(dsc->window, w, h);
    return (lv_coord_t) *w;
}
```

LVGL's API there is no get size function. There is a set size function and then there is get width and get height. So to keep with the same API there would need to be wrapper functions. There is also type casts that are handled int he wrapper functions as well.


I do need to add the documentation for the new functions. Before I got involved in keying out a bunch of documentation aI wanted to know if this is something that would be added.

kisvegabor:
Thank you for the reply :slightly_smiling_face: 

My concert is that we really just repeat a bigger partion of the SDL API, which:
- increases LVGL API even further
- not all functions are added here, so the users might still miss something

PR #4477, opened 2023-08-22
Add back bin image builtin decoder support

XuNeo:

Add back bin image decoder support.

Currently only indexed1/2/4/8 are supported.

Minor fixes:
 - Decoder memory leak. The memory allocated during decoding belongs to decoder itself, make it clearly by put the pointer to decoder descriptor `user_data` field and free it when decoder closes. 
 - Image cf. Do not change the original image cf during decoding. Ideally, we could de-quantizing palette just before drawing

### Checkpoints
 

kisvegabor:
Thanks, but I think it should target the [`feat/get-area`](https://github.com/lvgl/lvgl/tree/feat/img-decoder-get-area) branch.

XuNeo:
`mpy` is failing on message
```
build-standard/lvgl/lv_mpy.c:357:57: error: ‘lv_global_t’ undeclared (first use in this function); did you mean ‘lv_label_t’?
```

kisvegabor:
I've updated the get_area branch from master. It should work now. Note that the JPG decoder is not updated yet and its test in MP is still failing. See the related discussion [here](https://github.com/lvgl/lvgl/pull/4473).

kisvegabor:
Cool! 

Could you add a few test for the binary decoder? Ideally we should have one test for each format (It's ok to have only one indexed image)

I'll also add it from the BMP decoder.

src/draw/lv_img_decoder.c
```diff
@@ -403,18 +447,75 @@ lv_res_t lv_img_decoder_built_in_open(lv_img_decoder_t * decoder, lv_img_decoder
...
+            decoded_area->y1++;
+            decoded_area->y2++;
+        }
+
+        if(decoded_area->y1 > full_area->y2) {
+            return LV_RES_INV;
+        }
+        else {
+            lv_coord_t bpp = lv_color_format_get_bpp(cf);
+            lv_coord_t x_fraction = decoded_area->x1 % (8 / bpp);
```
W-Mai: `8 / bpp`
if bpp is greater than 8, it will cause a runtime problem.

XuNeo: The guard of `if(LV_COLOR_FORMAT_IS_INDEXED(cf)) {` makes sure bpp is valid

tests/src/test_cases/test_image_formats.c
```diff
@@ -92,6 +117,19 @@ void test_image_built_in_decode_recolor(void)
     img_create("XRGB8888", &test_img_cogwheel_xrgb8888, false, true);
     img_create("ARGB8888", &test_img_cogwheel_argb8888, false, true);
 
+#if 0 /*Recolor has issues on draw for now */
```
XuNeo: Hi @kisvegabor ,
There's something wrong with img_draw_core function when recolor is enabled. Could you help to check the code? I cannot figure out how to address it for now. 

Thanks!

kisvegabor: Sure, I'll check it tomorrow.

kisvegabor:
It should work now. There was minor issue in `lv_draw_sw_img.c`

XuNeo:
Thanks Gabor. 

Do you mind I force push `fix-bin-decoder` again with rebase to master, rather than merge master to `fix-bin-decode` branch? So there's less noisy irrelevant commits.

XuNeo:
> I will squash it into one commit anyway on merge. 

But when we revisiting this PR, the changes in `File Changes` tab are messy, and are difficult to find out what's really changed in this PR.

Squash it into one commit also loses the commit details, and makes it difficult to revisit if it really has bugs.

I can cleanup the commit message like `fix warnings` to make it suitable as a standalone commit.

kisvegabor:
Okay, please rebase it. It's really useful to keep the PR clean for later inspection. 

However I can not really see added value in keeping multiple commits from a PR on merge. I see that it's more "honest", but what can you use it for 3 moth later? Knowing that the 3rd commit from 10 introduced a bug in a merged feature is not really useful IMO. Having a single commit at least makes the commit history easier to read. 

I know it's not an exact topic and there is no really good or bad approach, only preference. So it's just my view, I'd be really curious about yours.

XuNeo:
Thank you for your honest. I agree with you. On certain stage squashing commits could make code history looks cleaner. That's especially useful when developing a new feature.

And this PR is actually kind of suitable for squashing as the title says "Add back bin decoder". I'm totally fine with your decision.

There are several reasons for not squashing the commit, but the most important one from my point of view is the impact on collaboration. 

So when merging code back to our own development branches, there are always conflicts since we have local changes not upstreamed too(but eventually will upstreamed once verified and discussed). 

Keep commits smaller could make it easier to addres conflicts, because it's easier to understand what's the change, and which one should be kept.

I believe @xiaoxiang781216 has done lots of work when mainting NuttX. Could you explain your options? I thinks it's a balance needs to keep.

src/draw/sw/lv_draw_sw_img.c
```diff
@@ -204,9 +204,13 @@ LV_ATTRIBUTE_FAST_MEM void lv_draw_sw_img(lv_draw_unit_t * draw_unit, const lv_d
             lv_area_t absolute_decoded_area = relative_decoded_area;
             lv_area_move(&absolute_decoded_area, coords->x1, coords->y1);
             if(res == LV_RES_OK) {
-                img_draw_core(draw_unit, draw_dsc, &draw_area, decoder_dsc.img_data,
-                              LV_COLOR_FORMAT_IS_INDEXED(cf) ? LV_COLOR_FORMAT_ARGB8888 : cf,
-                              &sup, &absolute_decoded_area);
+                lv_area_t draw_area_sub = draw_area;
```
XuNeo: No need to init it since it's going to be overwrite in _lv_area_intersect. I'll update the commit message together with this change.

kisvegabor:
Merging :slightly_smiling_face:

PR #4494, opened 2023-08-27
feat(indev): Add 'lv_indev_read' to allow read events from specified indev

bjsylvia:
Add 'lv_indev_read' to allow read events from specified input device actively without relying on timer-driven

In many operating systems, there is the ability to listen to file descriptor (fd) using poll/epoll/select. The active listening approach to input fd is much more efficient than the timer loop method, and it also consumes much less power. 

### Checkpoints
 
### DIFF
--- a/src/indev/lv_indev.c
+++ b/src/indev/lv_indev.c
@@ -70,6 +70,7 @@ static void indev_proc_reset_query_handler(lv_indev_t * indev);
 static void indev_click_focus(lv_indev_t * indev);
 static void indev_gesture(lv_indev_t * indev);
 static bool indev_reset_check(lv_indev_t * indev);
+void indev_read(lv_indev_t * indev, lv_indev_data_t * data);
 
 /**********************
  *  STATIC VARIABLES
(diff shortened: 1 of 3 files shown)


src/indev/lv_indev.c
```diff
@@ -70,6 +70,7 @@ static void indev_proc_reset_query_handler(lv_indev_t * indev);
 static void indev_click_focus(lv_indev_t * indev);
 static void indev_gesture(lv_indev_t * indev);
 static bool indev_reset_check(lv_indev_t * indev);
+void indev_read(lv_indev_t * indev, lv_indev_data_t * data);
```
kisvegabor: If it's used only in this file, can we make it `static`? It would be nice to have new name for it which more different from `lv_indev_read()`. For example `read_core()` or so.

kisvegabor (commented):
Do I understand correctly that it's the same idea as decoupling the [display refresh time](https://docs.lvgl.io/8.3/porting/display.html#decoupling-the-display-refresh-timer)?

If you, don't forget that you need to guarantee that `lv_indev_read`, `lv_timer_handler`, and `_lv_disp_refr_timer` are not called at the same time.

bjsylvia:
> Do I understand correctly that it's the same idea as decoupling the [display refresh time](https://docs.lvgl.io/8.3/porting/display.html#decoupling-the-display-refresh-timer)?
> 
> If you, don't forget that you need to guarantee that `lv_indev_read`, `lv_timer_handler`, and `_lv_disp_refr_timer` are not called at the same time.

Yes. I will update docs later.

src/indev/lv_indev.c
bjsylvia: Done. I have renamed `indev_read` to `indev_read_core`.

bjsylvia:
I have added 'Decoupling the input device read timer' section in docs.

kisvegabor:
LGTM, thank you!

PR #4505, opened 2023-08-31
Update screen object opacity function documentation

ONLYA:

The function call to set the screen opacity is not valid. Update it to `lv_obj_set_style_bg_opa`.


kisvegabor:
Thank you!

Issue #4463, opened 2023-08-17
Support SIMD instructions for rendering acceleration


terry0012:
### Introduce the problem

Some chip architectures support SIMD acceleration instructions, such as MVE on Cortex-M55 and NEON on Cortex-A7.

To use SIMD acceleration, a common usage pattern is to embed SIMD code into the software draw unit using some macros.

Another way is to follow the usage of GPU and implement LVGL draw unit with SIMD, but when GPU and SIMD acceleration instructions exist at the same time, the current draw unit does not have a fallback mechanism, and a custom fallback mechanism  and another draw unit with SIMD need to be implemented in the GPU draw unit.

Generally speaking, when a single chip supports SIMD acceleration instructions, SIMD can completely replace the implementation function of software draw unit. Therefore, I think it is more appropriate to embed SIMD code into the software draw unit. 

### Examples and cases

this is the example in sw blend if we emed SIMD code into it. [refer to the libyuv repository](https://github.com/lemenkov/libyuv)

```

void argb8888_fill_row_c(uint32_t color32,  uint32_t * dest_buf, int width) {
            int x;
            for(x = 0; x < w - 16; x += 16) {
                dest_buf[x + 0] = color32;
                dest_buf[x + 1] = color32;
                dest_buf[x + 2] = color32;
                dest_buf[x + 3] = color32;

                dest_buf[x + 4] = color32;
                dest_buf[x + 5] = color32;
                dest_buf[x + 6] = color32;
                dest_buf[x + 7] = color32;

                dest_buf[x + 8] = color32;
                dest_buf[x + 9] = color32;
                dest_buf[x + 10] = color32;
                dest_buf[x + 11] = color32;

                dest_buf[x + 12] = color32;
                dest_buf[x + 13] = color32;
                dest_buf[x + 14] = color32;
                dest_buf[x + 15] = color32;
            }
            for(; x < w; x ++) {
                dest_buf[x] = color32;
            }
}

void argb8888_fill_row_neon(uint32_t color32,  uint32_t * dest_buf, int width) {
    asm volatile(
    ...
    ...
   );
}

void argb8888_fill_row_mve(uint32_t color32,  uint32_t * dest_buf, int width) {
    asm volatile(
    ...
    ...
   );
}

LV_ATTRIBUTE_FAST_MEM void lv_draw_sw_blend_color_to_argb8888(_lv_draw_sw_blend_fill_dsc_t * dsc)
{
    int32_t w = dsc->dest_w;
[...]

PR #4507, opened 2023-08-31
Add ESP-LCD support

zjanosy:

Implemented ESP LCD driver for the new driver architecture. It is mainly working, only the include paths need to be modified (currently they are hardwired to my working folder). The main code is in the src/dev/esp/ folder. Here is an example of how to use it with the ESP ST7789 driver, connected via SPI:

```c
#include "esp_lcd_panel_io.h"
#include "esp_lcd_panel_vendor.h"
#include "esp_lcd_panel_ops.h"

#include "lvgl/src/dev/esp/lv_esp_lcd.h"
#include "lvgl/src/dev/esp/lv_esp_tick.h"

/*Using SPI2 in the example*/
#define LCD_HOST  SPI2_HOST

#define LCD_PIXEL_CLOCK_HZ     (15 * 1000 * 1000)
#define LCD_BK_LIGHT_ON_LEVEL  1
#define LCD_BK_LIGHT_OFF_LEVEL !LCD_BK_LIGHT_ON_LEVEL
#define PIN_NUM_SCLK           7
#define PIN_NUM_MOSI           15
#define PIN_NUM_MISO           -1
#define PIN_NUM_LCD_DC         5
#define PIN_NUM_LCD_RST        6
#define PIN_NUM_LCD_CS         4
#define PIN_NUM_BK_LIGHT       -1
#define PIN_NUM_TOUCH_CS       -1

/*The pixel number in horizontal and vertical*/
#define LCD_H_RES              320
#define LCD_V_RES              240

/*Bit number used to represent command and parameter*/
#define LCD_CMD_BITS           8
#define LCD_PARAM_BITS         8

extern void ui_init(void);

void app_main(void) {
    static lv_disp_t * disp;

    ESP_LOGI(TAG, "Initialize LVGL 9 library");
    lv_init();

    ESP_LOGI(TAG, "Initialize SPI bus");
    spi_bus_config_t buscfg = {
        .sclk_io_num = PIN_NUM_SCLK,
        .mosi_io_num = PIN_NUM_MOSI,
        .miso_io_num = PIN_NUM_MISO,
        .quadwp_io_num = -1,
        .quadhd_io_num = -1,
        .max_transfer_sz = LCD_H_RES * 80 * sizeof(uint16_t),
    };
    ESP_ERROR_CHECK(spi_bus_initialize(LCD_HOST, &buscfg, SPI_DMA_CH_AUTO));

    ESP_LOGI(TAG, "Install panel IO");
    esp_lcd_panel_io_spi_config_t io_config = {
        .dc_gpio_num = PIN_NUM_LCD_DC,
        .cs_gpio_num = PIN_NUM_LCD_CS,
        .pclk_hz = LCD_PIXEL_CLOCK_HZ,
        .lcd_cmd_bits = LCD_CMD_BITS,
        .lcd_param_bits = LCD_PARAM_BITS,
        .spi_mode = 0,
        .trans_queue_depth = 10,
        /*The following two lines are important*/
        .on_color_trans_done = lv_esp_lcd_notify_color_trans_done,
        .user_ctx = &disp,
    };

    /*Attach the LCD to the SPI bus*/
    ESP_ERROR_CHECK(esp_lcd_new_panel_io_spi((esp_lcd_spi_bus_handle_t)LCD_HOST, &io_config, &io_handle));

    esp_lcd_panel_handle_t panel_handle = NULL;
    esp_lcd_panel_dev_config_t panel_config = {
        .reset_gpio_num = PIN_NUM_LCD_RST,
        .rgb_endian = LCD_RGB_ENDIAN_RGB,
        .bits_per_pixel = 16,
    };
    ESP_LOGI(TAG, "Install ST7789 panel driver");
    ESP_ERROR_CHECK(esp_lcd_new_panel_st7789(io_handle, &panel_config, &panel_handle));

    ESP_ERROR_CHECK(esp_lcd_panel_reset(panel_handle));
    ESP_ERROR_CHECK(esp_lcd_panel_init(panel_handle));

    esp_lcd_panel_swap_xy(panel_handle, true);
    esp_lcd_panel_mirror(panel_handle, true, false);
    esp_lcd_panel_invert_color(panel_handle, true);

    /*User can flush pre-defined pattern to the screen before we turn on the screen or backlight*/
    ESP_ERROR_CHECK(esp_lcd_panel_disp_on_off(panel_handle, true));

    /*Create display object*/
    disp = lv_esp_lcd_create(LCD_H_RES, LCD_V_RES, panel_handle);

    /*Initialize tick timer*/
    lv_esp_tick_init();

    /*Initialize UI*/
    ui_init();

    while (1) {
        /*The task running lv_timer_handler should have lower priority than that running `lv_tick_inc`*/
        lv_timer_handler();
        /*Raise the task priority of LVGL and/or reduce the handler period can improve the performance*/
        vTaskDelay(pdMS_TO_TICKS(10));
    }
}
```

### Checkpoints
 

kisvegabor:
cc @igrr @espzav 

With this update the user can integrate LVGL with esp_lcd with a single line of code. It's not working in this PR yet, but they will be able to use the generic `lv_disp_set_rotation()`,  `lv_disp_set_res()`, and `lv_disp_set_color_format()` functions to adjust the display.

What do you think?

There are two things which are not clear to us:
1. What is the intended way to add ESP include path to LVGL? We already have a [esp.cmake](https://github.com/lvgl/lvgl/blob/master/env_support/cmake/esp.cmake) file in LVGL, so I beilve we need to extend it somehow.
2. To be perfectly honest it's not a one liner integration, because you need to set `.on_color_trans_done = lv_esp_lcd_notify_color_trans_done`  as well. Can we figure it somehow in `lv_ecp_lcd_create` by seeing only the `panel_handle`?

espzav:
Hi @kisvegabor, we have something simillar in a component here: https://github.com/espressif/esp-bsp/tree/master/components/esp_lvgl_port We still adding new features into it. We plan add RGB LCD support and more low power support soon (low power waiting for new LVGL9 callbacks `LV_DISP_EVENT_INVALIDATE_AREA`  ). 
1. For using `esp_lcd`, you should put it into `REQUIRES` like this: https://github.com/espressif/esp-bsp/blob/master/components/esp_lvgl_port/CMakeLists.txt
2. You need `io_handle` too (same as in `esp_lvgl_port` component)

espzav:
> Great! Do you think t's worth maintaining and ESP and and LVGL version of it? I wonder if we can join our forces somehow? 🙂

I remember, when I started in Espressif, there was a lot of mess in LCD drivers for LVGL (there was a public repository). Our graphic team did a lot of work to clean the ESP LCD drivers and we made a lot of components for easier use of the LVGL on Espressif chips. Now, we maintain these components. 
The question is. Do we need a new "esp_lvgl_port" inside LVGL, or could we take `esp_lvgl_port` and use it inside LVGL? 
Second question: Our `esp_lvgl_port` is not only for LCD drivers. It can handle LCD touch, extarnal buttons and we have PR for USB HID support: https://github.com/espressif/esp-bsp/pull/155

One note: The `esp_lcd` doesn't contain all LCD driver, lot of drivers are as components here: https://github.com/espressif/esp-bsp/tree/master/components/lcd (same for LCD touch: https://github.com/espressif/esp-bsp/tree/master/components/lcd_touch). When you want to use some of these drivers, you should add it as a component (into file `idf_component.yml` in the project). 

My conclusion: I would like to see only one component for this purpose. It doesn't matter if it will be inside LVGL or on our side. The question is, what is missing and what is easier for end users. We can share our component with you, but you should do maintenance and I would like to see all the features that we have :-)

kisvegabor:
We are planning to add vendor independent support for the popular display and touch controllers. This way the user only needs to initialize the SPI or parallel port and write some simple callbacks for data and color transfer. 

In March on Embedded World I've discussed it with Ivan and he said, Espressif would like to keep its own display controller support and not rely on LVGL's display controllers because the users might want to use other libraries then LVGL. 

> The esp_lcd doesn't contain all LCD driver, lot of drivers are as components here:

There are 6 display controllers [here](https://github.com/espressif/esp-bsp/tree/master/components/lcd). I don't want to push you just would like to understand why don't you merge these drivers to [esp-idf / esp_lcd](https://github.com/espressif/esp-idf/tree/master/components/esp_lcd/src). 

> Our esp_lvgl_port is not only for LCD drivers. It can handle LCD touch, extarnal buttons and we have PR for USB HID support:

Adding support for more input devices is planned on our side as well. I wonder if it would be possible to hide e.g. the HID mouse devices under `esp_lcd_touch` so LVGL would see only a simple `read_cb` from which it would be easy to create and input device. 

> Do we need a new "esp_lvgl_port" inside LVGL, or could we take esp_lvgl_port and use it inside LVGL?

I don't see the whole picture, road map and potential issues in ESP-IDF but it seems to me the ideal case would be if all drivers (lcd, touch, button, kaypad, etc) where hidden by the `esp_lcd_*` API and LVGL could build a layer to configure these devices via LVGL's generic display and indev API.

What do you think?

tore-espressif:
Hello @kisvegabor @zjanosy @espzav

There are quite a few topics being discussed, I'll try to explain them one by one, starting with the simple ones:

**1. What is the intended way to add ESP include path to LVGL?**

If you build LVGL in ESP-IDF environment, the ESP include path is available to you. Let me know if you have any specific include problem, we can help with updating the esp.cmake

**2. There are 6 display controllers here. I don't want to push you just would like to understand why don't you merge these drivers to esp-idf / esp_lcd.**

Since the development of esp-idf started, it has become too bulky; containing loosely coupled components, that were not the 'core' functions. We want the esp-idf to contain only the core components. LCD controller drivers are external to esp32 chips, so their drivers are uploaded to [IDF component registry](https://components.espressif.com/components?q=lcd), same as LVGL. The fact, that there are 6 drivers natively in `esp_lcd` can be deemed as anti-pattern; we wanted to make it easier for new users with typical displays...

**3. I wonder if it would be possible to hide e.g. the HID mouse devices under esp_lcd_touch so LVGL would see only a simple read_cb from which it would be easy to create and input device.**

This sounds as a good idea, as long as the API is flexible enough to support all types of devices (touch controllers, HW buttons, HW rotary encoders, USB keyboards, mice, gamepads...)

**4. As we are adding drivers to LVGL, I would be happy to keep it on our side.**

Of course, understandable. The drivers that we already have are released under permissive license, so feel free to have a look and/or use it. As you already know, we must have LCD drivers decoupled from LVGL for potential other graphical libraries.

Porting LVGL to ESP32 based system is often a lot more than just implementing simple LCD driver `lcd_init` and `lcd_flush`. This is where `esp_lvgl_port` comes into the picture; it aims to solve the following:

* Support for new ESP chips (we do not expect you to do this for us :D )
* LVGL tick (e.g. tickless modes)
* FreeRTOS integration
* Backlight control (on/off or dimming)
* DSP optimizations for newer chips
* External RAM applications
* Multi display applications
* Touch controllers
* Display rotation (synced with touch rotation)
* Other input devices (Buttons, rotary encoders, USB devices)
* Low-power and sleep optimizations

We already ship `esp_lvgl_port` with all SquareLineStudio examples and **we hope it can get more attention from users in the future, by using it in your official examples**. After all, the points mentioned above are not trivial to solve

**5. it seems to me the ideal case would be if all drivers (lcd, touch, button, kaypad, etc) where hidden by the esp_lcd_\* API and LVGL could build a layer to configure these devices via LVGL's generic display and indev API.**

`esp_lvgl_port` aims to fulfill just this role! The fact that it is decoupled from `esp-idf` release cycle, makes it a lot easier for us to incorporate new (even breaking) changes. At the same time, we won't have to update the 'big-old' esp-idf, in case something changes in LVGL -> we can solve these things in esp_lvgl_port

kisvegabor:
Thanks for your comments Tomas!

> Porting LVGL to ESP32 based system is often a lot more than just implementing simple LCD driver lcd_init and lcd_flush. This is where esp_lvgl_port comes into the picture; it aims to solve the following:

I think this part is most complicated. If  we could see how to manage these points probably we will understand the other issues better as well.

> Support for new ESP chips (we do not expect you to do this for us :D )

What does it mean from LVGL integration's point of view? What do we need to do related to LVGL to support new chips?

> LVGL tick (e.g. tickless modes)

What do you exactly mean  by tickless mode?

> FreeRTOS integration

In v9 we have an [OS layer](https://github.com/lvgl/lvgl/tree/master/src/osal) built into LVGL where FreeRTOS is already supported.  We use it during rendering, but it might be useful for the ESP-LVGL integration as well. 
What parts of FreeRTOS do you need for ESP-LVGL integration.

> Backlight control (on/off or dimming)

It's really hardware specific if we need to control a pin or PWM. 

> DSP optimizations for newer chips

Do you mean SIMD instruction during rendering? If so we have an [ongoing discussion](https://github.com/lvgl/lvgl/issues/4463) about it. In short, @igrr described that you would like to host it on our side, but LVGL will provide an interface to add the ASM acceleration. 

> External RAM applications

In practice what do we need to do to support them?

> Multi display applications

The driver in this PR already support multiple displays. 

> Touch controllers

Similarly to `esp_lcd` we can build a layer on top of `esp_lcd_touch` and the other types as well.

> Display rotation (synced with touch rotation)

LVGL handles it internally. If you call `lv_disp_set_rotation` the touch coordinates will be rotated automatically.

> Other input devices (Buttons, rotary encoders, USB devices)

I hope they can we warpped in to esp_lcd_*`. :slightly_smiling_face: 

> Low-power and sleep optimizations

I guess it mean suspending the call of `lv_timer_handler()` in sleep and restart it on an event. If so we can figure out user/developer friendly was to handle this.

---

Sorry if I seem to be pit picking. :slightly_smiling_face: I just want to clearly see what are the exact issues to solve so that we can figure out solutions accordingly.

tore-espressif:
Thank you for your thorough comments!

> What does it mean from LVGL integration's point of view? What do we need to do related to LVGL to support new chips?

Ideally, nothing is needed from your side. But my past experiences showed that sometimes unexpected things can go wrong. Especially with very small MCUs (such as ESP32-C3/C2), with limited RAM and other HW resources.

> What do you exactly mean by tickless mode?

In LVGL v8, there was a `LV_TICK_CUSTOM` option. If it was enabled we did not have call `lv_tick_inc()` periodically. But it did not compile for ESP32.

I can see that this was changed in v9. I'm sorry, I must have missed it. We just wanted to get rid of one interrupt only for the tick timer...

> What parts of FreeRTOS do you need for ESP-LVGL integration.

Nothing complicated; something close to you idea of 'one-line-initialization' together with thread safety. We want to provide 
* function that create a task that handles LVGL events and 
* locking mechanism that would protect the user from race conditions

Not a rocket science, but it can help novice users a lot.

> Do you mean SIMD instruction during rendering?

Yes, that's it :) Let's discuss it in the separate thread

> In practice what do we need to do to support them (external RAM support)?

Unfortunately yes. There are many users who just can't sacrifice internal RAM for (relatively) huge frame buffers. This is even more obvious for RGB panels (displays without LCD controller), where we must have full frame buffer in RAM, (thus partial frame buffers cannot be used).

> LVGL handles it internally. If you call lv_disp_set_rotation the touch coordinates will be rotated automatically.

That is what we are doing in our drivers.
I meant, that all LCD drivers that we implement, support HW rotation.
(but you're right, this is probably an off-topic)

> The driver in this PR already support multiple displays.

Yes, this issue is probably solved. But it wasn't easy to have multiple LCDs on ESP before the `esp_lcd` component was released. So I listed it here for completeness :)

> I guess it mean suspending the call of lv_timer_handler() in sleep and restart it on an event. If so we can figure out user/developer friendly was to handle this.

Yes, this is related to the RTOS integration really. We must know when can we suspend the LVGL task (that periodically calls `lv_timer_handler()`)

> I hope they (buttons, rotary encoder, USB input) can we warpped in to esp_lcd_*`. 🙂

IMO, this will never happen. The maintainers of `esp_lcd` follow Single-responsibility principle, so it manages **only LCDs** (even the backlight control is outside of esp_lcd). That is the main reason why we keep talking about `esp_lvgl_port`: The single responsibility of `esp_lvgl_port` is to provide a component that will connect all esp-idf features and components to LVGL.

kisvegabor:
> I can see that this was changed in v9. I'm sorry, I must have missed it. We just wanted to get rid of one interrupt only for the tick timer...

Yes, just for the record, [`lv_tick_set_cb`](https://github.com/lvgl/lvgl/blob/master/src/tick/lv_tick.h#L66) can be used to set a tick source at run time. 

> Unfortunately yes. There are many users who just can't sacrifice internal RAM for (relatively) huge frame buffers. This is even more obvious for RGB panels (displays without LCD controller), where we must have full frame buffer in RAM, (thus partial frame buffers cannot be used).

We can move `lv_disp_set_draw_buffers(disp, draw_buf, draw_buf_2, buf_size, render_mode);` out from `lv_esp_lcd_create` so the user can freely configure the buffers. Of course we need some examples and good documentation. 

> Yes, this is related to the RTOS integration really. We must know when can we suspend the LVGL task (that periodically calls lv_timer_handler())

The return value of `lv_timer_handler()` is the number of milliseconds to the call of the next timer. `lv_timer_get_time_until_next()` also can be used to get this time. They return `LV_NO_TIMER_READY` when there is no active timer. Note that, LVGL automatically pauses the render and animation timers if there nothing to render or animate. If the user also disables the input device read timer manually before sleep, no internal LVGL timer will run.

> IMO, this will never happen. The maintainers of esp_lcd follow Single-responsibility principle, so it manages only LCDs (even the backlight control is outside of esp_lcd). That is the main reason why we keep talking about esp_lvgl_port: The single responsibility of esp_lvgl_port is to provide a component that will connect all esp-idf features and components to LVGL.

By `esp_lcd_*` I meant e.g. [`esp_lcd_touch`](https://github.com/espressif/esp-bsp/blob/master/components/lcd_touch/esp_lcd_touch/esp_lcd_touch.c). If we had `esp_lcd_touch` and `esp_lcd_button` we could easily build the LVGL wrapper on top of it.

---

It seems to me all the mentioned issues are manageable on LVGL side too. So the question is if there is a principal objection on Espressif's  side to maintain the ESP-LVGL layer in LVGL. Probably we will have such a layer/wrapper anyway  because I think it's useful for the users to get it with LVGL automatically and it's also a good reference and example of how we image the driver support in the `lvgl` repo. I hope we can join our forces in some way :slightly_smiling_face:

kisvegabor:
I'm considering to merge this PR to see how people will react and how it can be used in the practice. 

@zjanosy A few things are still missing:
- In the meantime `disp` was renamed to `display` in LVGL
- `esp_lcd_touch`should be added too
- I'll prepare a place for it, but we should have a minimal docs page for it too.

kisvegabor:
Not stale

kisvegabor:
Not stale, shall be finalized before January 15.

kisvegabor:
Not stale

kisvegabor:
Let's close it for now, and get back to this when there is a real need.

sukesh-ak:
@kisvegabor here is a working sample using `esp_lvgl_port` using `ST7796 ` and `FT5x06`
https://github.com/sukesh-ak/IDF-ESP_LCD-LVGL

Once `esp_lvgl_port` with `lvgl 9.x` support is published to the Espressif Component registry, I can send a sample for integration docs.

Btw, @zjanosy the code above does not take care of multi-tasking which needs a mutex. 
And this is taken care of in the sample while using `esp_lvgl_port`.

kisvegabor:
Hi,

By `esp_lvgl_port` do you mean the official ESP support package for LVGL?

sukesh-ak:
> Hi,
> 
> By `esp_lvgl_port` do you mean the official ESP support package for LVGL?

Yes

kisvegabor:
Got it!

> Once esp_lvgl_port with lvgl 9.x support is published to the Espressif Component registry, I can send a sample for integration docs.

That would be great! Actually we wasn't sure if this driver/interface is needed in LVGL. If Espressif maintains its own LVGL integration layer, we can advertise and support it.

sukesh-ak:
> Got it!
> 
> > Once esp_lvgl_port with lvgl 9.x support is published to the Espressif Component registry, I can send a sample for integration docs.
> 
> That would be great! Actually we wasn't sure if this driver/interface is needed in LVGL. If Espressif maintains its own LVGL integration layer, we can advertise and support it.

I think if Espressif and other vendors can keep the drivers upto-date, it makes more sense. 
So `LVGL` can focus on the graphics, widget, designer etc. because most of these MCU/CPU vendors might want to be open to any other UI frameworks too.

Also `espressif ` component repo and new `esp-bsp` has made life easier.

Personally, I am now comfortable with LVGL and I can write UI using notepad too :)

PR #4526, opened 2023-09-06
fix: do not copy invalid areas if not double buffered mode

lukasz-skalski:
Sync areas should not be copied if we are using direct mode with single buffer - otherwise we have memleak.


kisvegabor:
Nice catch! Thank you!

PR #4533, opened 2023-09-08
feat/dashed-chart-lines

Zer0-bit:
Hi - thx for this awesome lib first of all!

So i have a project where i was needing to have dashed chart lines and sadly discovered they aren't an oob option so instead of following the online sources of custom drawing the lines outside the chart widget i thought introducing this as an option for the chart itself will be a much nicer way :)
image

It's not in a fully ready state yet because i hardcoded some styling values like dash width and gap but those can also be handled gracefully by adding their own options  in a separate or existing struct. 

So what i did is added a new enum to be able to specify the line type:
```cpp
LV_CHART_LINE_SOLID,  // Solid line (default)
LV_CHART_LINE_DASHED, // Dashed line
```
Defined a new property in the **lv_chart_series_t** struct:
```cpp
lv_chart_line_style_t line_style;
```

And added a function to set the line type:
```cpp
void lv_chart_set_series_line_style(lv_chart_series_t * ser, lv_chart_line_style_t line_style);
```

Then it can be used in the main code just byt setting it's style after a series is initialised, ex:

```cpp
lv_chart_series_t *example_series = lv_chart_add_series(chart, lv_palette_main(LV_PALETTE_RED), LV_CHART_AXIS_SECONDARY_Y);
lv_chart_set_series_line_style(example_series, LV_CHART_LINE_DASHED);
```

P.S. I'm going to make it more customisable as i'm working on making it draw dashed in all the directions but wanted to raise this PR to see what's the general consensus on adding this to the official codebase ?

### DIFF
--- a/src/extra/widgets/chart/lv_chart.c
+++ b/src/extra/widgets/chart/lv_chart.c
@@ -988,6 +988,7 @@ static void draw_series_line(lv_obj_t * obj, lv_draw_ctx_t * draw_ctx)
                                 lv_draw_line_dsc_init(&draw_dashed_line);
                                 draw_dashed_line.dash_gap = lv_dpx(10);
                                 draw_dashed_line.dash_width = lv_dpx(10);
+                                draw_dashed_line.width = lv_dpx(3);
                                 draw_dashed_line.color = ser->color;
                                 lv_draw_line(draw_ctx, &draw_dashed_line, &p1, &p2);
                             } else {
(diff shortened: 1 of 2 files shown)


kisvegabor:
Hey,

Thank you! 

I think the root of the problem is that the software render doesn't support skewed dashed lines. Therefore line dashing is very limited.

However there are 2 good news:
1. In v9 LVGL will get more GPU support so with a capable GPU dashing can be solved.
2. I have some ideas to handle dashed line support with software render as well.

So I would rather wait for a more complete solution.

Zer0-bit:
> Hey,
> 
> Thank you!
> 
> I think the root of the problem is that the software render doesn't support skewed dashed lines. Therefore line dashing is very limited.
> 
> However there are 2 good news:
> 
> 1. In v9 LVGL will get more GPU support so with a capable GPU dashing can be solved.
> 2. I have some ideas to handle dashed line support with software render as well.
> 
> So I would rather wait for a more complete solution.

Thanks for your reply and explanation on the current plans, it's nice the lvgl v9 is fixing many of these scenarios, i'll most definitely be upgrading to v9 when it ships (hopefully not too many api changes lol ), but in the meantime i'll just use a fork to get the ball rolling with my needs from the chart widget :)

Thanks again for an awesome project 👍

kisvegabor:
Thank you for your understanding :slightly_smiling_face:

PR #4535, opened 2023-09-08
fix(docs): dropdown: fix function name in description of static options

kolliemo:
### Description 

Docs say that adding a static option list is done via `lv_dropdown_set_static_options`, correct is `lv_dropdown_set_options_static` (as in API below in the page)


kisvegabor:
Thank you!

Issue #4092, opened 2023-03-27
Suggestion: add a unique identifier for lv_obj


bjsylvia:
### Introduce the problem

During the LVGL development process, we cannot access an lv_obj instance without saving it at creation time.

### Suggested solution

If lv_obj has a unique identifier (string name or integer), we can get the object by the root object and its ID. 

For example:
    
    /*
     1. add 'uint16_t id' in lv_obj_t struct
     2. new apis:
         void lv_obj_set_id(lv_obj_t* obj, int16_t id);
         lv_obj_t* lv_obj_get_obj_by_id(lv_obj_t* root, int16_t id);
    */

    // test

    // uid
    int16_t imgid = 201;

    // creation
    {
      lv_obj_t * img = lv_img_create(lv_scr_act());
      lv_obj_set_id((lv_obj_t*)img, imgid); 
      lv_img_set_src(img, LV_SYMBOL_OK "Ok");
      ...
    }
    // ...

    //operation 
    {
      lv_obj_t* img = (lv_img_t*)lv_obj_get_obj_by_id(lv_scr_act(), imgid);
      if (img) {
        lv_img_set_src(img, LV_SYMBOL_CLOSE "Close");
      }
    }

kisvegabor:
I was thinking about it too but for a different reason. In your example now you need to save `imgid` into a global variable. With the same effort you could save the image object too into an `lv_obj_t * img1` variable. 


I was thinking about this feature to make debuging easier. If you see only a lot of `lv_obj_t*`s it's hard to say which one is which. Adding a name or ID would help a lot with it.

bjsylvia:
Yes，I agree that the ID of lv_obj can be useful for debugging.  It also makes it possible to delay getting an object instance until it is actually used.

kisvegabor:
Okay, we can easily by adding a new field to `lv_obj_t`. It should be optional though. 

What do you think, shall we let the user `typedef` the type of the ID or is it ok if we hard code it to `int` or `void*`?

bjsylvia:
It seems that 'let the user typedef the type of the ID' can add flexibility and customization to the system. But void* maybe more convenient. Which idea do you like more?

kisvegabor:
I'd prefer `void *` because it's more unified. If we typedef-ed it the UIs would be less portable as the widgets would rely on the parent projects config.

fvanroie:
> During the LVGL development process, we cannot access an lv_obj instance without saving it at creation time.

[...]

PR #4536, opened 2023-09-08
Add obj id and dump support

XuNeo:

See discussion here #4092 

Add field `id` to `lv_obj_t`. Assign `id` during obj's constructor callback.

`id` is generated by couting how many objects for this class has been created.
Also added class name field.

### Checkpoints
 
### DIFF
--- a/src/core/lv_obj.c
+++ b/src/core/lv_obj.c
@@ -49,6 +49,7 @@ static void lv_obj_set_state(lv_obj_t * obj, lv_state_t new_state);
 /**********************
  *  STATIC VARIABLES
  **********************/
+
 const lv_obj_class_t lv_obj_class = {
     .constructor_cb = lv_obj_constructor,
     .destructor_cb = lv_obj_destructor,
@@ -59,6 +60,7 @@ const lv_obj_class_t lv_obj_class = {
     .group_def = LV_OBJ_CLASS_GROUP_DEF_FALSE,
     .instance_size = (sizeof(lv_obj_t)),
     .base_class = NULL,
+    .name = "obj",
 };
 
 /**********************
@@ -306,6 +308,10 @@ static void lv_obj_constructor(const lv_obj_class_t * class_p, lv_obj_t * obj)
     obj->flags |= LV_OBJ_FLAG_SCROLL_WITH_ARROW;
     if(parent) obj->flags |= LV_OBJ_FLAG_GESTURE_BUBBLE;
 
+#if LV_USE_OBJ_ID
+    lv_obj_assign_id(class_p, obj);
+#endif
+
     LV_TRACE_OBJ_CREATE("finished");
 }
 
@@ -341,6 +347,10 @@ static void lv_obj_destructor(const lv_obj_class_t * class_p, lv_obj_t * obj)
         lv_free(obj->spec_attr);
         obj->spec_attr = NULL;
     }
+
+#if LV_USE_OBJ_ID
+    lv_obj_free_id(obj);
+#endif
 }
 
 static void lv_obj_draw(lv_event_t * e)
--- a/src/core/lv_obj.h
+++ b/src/core/lv_obj.h
@@ -185,6 +185,9 @@ typedef struct _lv_obj_t {
     uint32_t style_other_prop_is_set;
 #endif
     void * user_data;
+#if LV_USE_OBJ_ID
+    void * id;
+#endif
     lv_area_t coords;
     lv_obj_flag_t flags;
     lv_state_t state;
@@ -346,6 +349,42 @@ const lv_obj_class_t * lv_obj_get_class(const lv_obj_t * obj);
  * @return          true: valid
  */
 bool lv_obj_is_valid(const lv_obj_t * obj);
+
+#if LV_USE_OBJ_ID
+
+/**
+ * Assign an id to an object if not previously assigned
+ * Set `LV_USE_OBJ_ID_BUILTIN` to 1 to use builtin method to generate object ID.
+ * Otherwise, these functions including `lv_obj_[assign|free|stringify]_id` should be implemented externally.
+ *
+ * @param class_p   the class this obj belongs to. Note obj->class_p is the class currently being constructed.
+ * @param obj   pointer to an object
+ */
+void lv_obj_assign_id(const lv_obj_class_t * class_p, lv_obj_t * obj);
+
+/**
+ * Free resources allocated by `lv_obj_assign_id`
+ * @param obj   pointer to an object
+ */
+void lv_obj_free_id(lv_obj_t * obj);
+
+/**
+ * Format an object's id into a string.
+ * @param obj   pointer to an object
+ * @param buf   buffer to write the string into
+ * @param len   length of the buffer
+ */
+const char * lv_obj_stringify_id(lv_obj_t * obj, char * buf, uint32_t len);
+
+#if LV_USE_OBJ_ID_BUILTIN
+/**
+ * Free resources used by builtin ID generator.
+ */
+void lv_objid_builtin_destroy(void);
+
+#endif
+
+#endif /*LV_USE_OBJ_ID*/
 
 /**********************
  *      MACROS
--- a/src/core/lv_obj_class.h
+++ b/src/core/lv_obj_class.h
@@ -62,6 +62,7 @@ typedef struct _lv_obj_class_t {
     void (*event_cb)(const struct _lv_obj_class_t * class_p,
                      struct _lv_event_t * e);  /**< Widget type specific event function*/
     void * user_data;
+    const char * name;
     lv_coord_t width_def;
     lv_coord_t height_def;
     uint32_t editable : 2;             /**< Value from ::lv_obj_class_editable_t*/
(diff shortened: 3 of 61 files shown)


bjsylvia (commented):
should avoid to use static global variable, it will break multi-instance feature.

XuNeo:
> should avoid to use static global variable, it will break multi-instance feature.

I think we only need to protect `lv_obj_assign_id` where a read-modify-write happen on global variable `obj_count`. As an ID, it's OK to share the same counter for multiple instance right?

src/core/lv_obj_id_builtin.c
```diff
@@ -0,0 +1,73 @@
...
+void lv_obj_assign_id(lv_obj_t * obj)
+{
+    LV_ASSERT(obj && obj->class_p);
+    LV_ASSERT(obj->id == NULL);
+    if(obj->class_p->obj_count == NULL) return;
+
+    uint32_t count = *obj->class_p->obj_count;
+    count += 1;
+    obj->id = (void *)(lv_uintptr_t)count;
+    *obj->class_p->obj_count = count;
```
XuNeo: Add protection for multiple thread environment.

src/core/lv_obj_class.h
```diff
@@ -59,6 +59,12 @@ typedef struct _lv_obj_class_t {
     void (*event_cb)(const struct _lv_obj_class_t * class_p,
                      struct _lv_event_t * e);  /**< Widget type specific event function*/
     void * user_data;
+#if LV_USE_CLASS_NAME
+    const char * name;
+#endif
+#if LV_USE_OBJID_BUILTIN
+    uint32_t * obj_count; /* Points to a static variable for logging object counts */
```
kisvegabor: Why does it need to be pointer? Can't we use a simple `uint32_t`?

src/core/lv_obj_tree.h
```diff
@@ -164,6 +164,12 @@ uint32_t lv_obj_get_index(const struct _lv_obj_t * obj);
  */
 void lv_obj_tree_walk(struct _lv_obj_t * start_obj, lv_obj_tree_walk_cb_t cb, void * user_data);
 
+/**
+ * Iterate through all children of any object and print its ID.
```
kisvegabor:
```suggestion
 * Iterate through all children of any object and print their ID.
```

src/core/lv_obj.h
```diff
@@ -347,6 +350,26 @@ const lv_obj_class_t * lv_obj_get_class(const lv_obj_t * obj);
  */
 bool lv_obj_is_valid(const lv_obj_t * obj);
 
+/**
+ * Assign an id to an object if not previously assigned
+ * @param obj   pointer to an object
+ */
```
kisvegabor: Mention that these functions needs to implemented externally if `LV_USE_OBJID_BUILTIN` is not enabled.

kisvegabor (commented):
It looks good in general, just a few comments.

src/core/lv_obj.h
XuNeo: Done.

src/core/lv_obj_class.h
XuNeo: class like `lv_obj_class` is always defined as `const`, so we have to use pointer to another variable.

```c
#if LV_USE_OBJID_BUILTIN
    static uint32_t obj_count = 0;
#endif

const lv_obj_class_t lv_obj_class = {
#if LV_USE_OBJID_BUILTIN
    .obj_count = &obj_count,
#endif

#if LV_USE_CLASS_NAME
    .name = "obj",
#endif
};

```

src/core/lv_obj.c
```diff
@@ -59,6 +63,13 @@ const lv_obj_class_t lv_obj_class = {
     .group_def = LV_OBJ_CLASS_GROUP_DEF_FALSE,
     .instance_size = (sizeof(lv_obj_t)),
     .base_class = NULL,
+#if LV_USE_OBJID_BUILTIN
+    .obj_count = &obj_count,
+#endif
+
+#if LV_USE_CLASS_NAME
```
XuNeo: Should we remove config `LV_USE_CLASS_NAME`, make it alwasy enabled? It only consumes little flash size.

src/core/lv_obj.c
```diff
@@ -59,6 +63,13 @@ const lv_obj_class_t lv_obj_class = {
     .group_def = LV_OBJ_CLASS_GROUP_DEF_FALSE,
     .instance_size = (sizeof(lv_obj_t)),
     .base_class = NULL,
+#if LV_USE_OBJID_BUILTIN
+    .obj_count = &obj_count,
```
XuNeo: `obj_count` could be useful for custom `lv_obj_assign_id`, let's the config to `#if LV_USE_OBJID` ?

src/core/lv_obj.c
kisvegabor: I agree!

src/core/lv_obj_class.h
kisvegabor: Doh, that's true. I still wonder if we can figure out something cleaner. 

A simple solution wold be to have a `static uint32_t id;` in the assign_id function, but this way we couldn't count the number of widgets by type, only a global ID. 

Do you think it wold be suitable as well?

XuNeo: We can use dynamic array to track each type of widget and resize the array when new class type is used.

kisvegabor: It can work, but how will we convert `obj->class_p` to an array index?

XuNeo: I have pushed the patch to remove urgly `obj_count`.

src/core/lv_obj_class.c
```diff
@@ -165,22 +165,16 @@ bool lv_obj_is_group_def(lv_obj_t * obj)
...
         /*Construct the base first*/
-        lv_obj_construct(obj);
+        lv_obj_construct(obj, class_p->base_class);
     }
 
-    /*Restore the original class*/
-    obj->class_p = original_class_p;
-
-    if(obj->class_p->constructor_cb) obj->class_p->constructor_cb(obj->class_p, obj);
+    if(class_p->constructor_cb) class_p->constructor_cb(class_p, obj);
```
XuNeo: I understand the obj->class_p needed to be changed during construction now. But we need to find a better way to call `lv_obj_assign_id`. Currently it's inside of basic `obj->constructor_cb`, where all widgets will go through. But withouth this change, obj->class_p is always `obj`.

src/core/lv_obj_id_builtin.c
```diff
@@ -0,0 +1,126 @@
...
+    uint32_t i;
+    uint32_t id = 0;
+    lv_global_t * global = LV_GLOBAL_DEFAULT();
+    class_info_t * info = NULL;
+
+    if (obj == NULL || obj->class_p == NULL) return;
+    if (global == NULL) return;
+
+#if LV_USE_OS
+    lv_mutex_lock(&global->objid_mutex);
```
kisvegabor: Why do we need a mutex here? 

LVGL is not thread safe anyway, so if two objects are created at the same time in two threads, the user needs to use a mutex around the `create` functions.

XuNeo: True, I'll remove it later.

XuNeo:
Hi Gabor, 
I tried to solve mpy CI issue by enabling `LV_USE_OBJID_BUILTIN` if `LV_USE_OBJID` enabled. However, it seems the python script doest like this way.  
>Verify that lv_conf_internal.h matches repository state / verify-conf-internal (pull_request) 

https://github.com/lvgl/lvgl/blob/a53b6811e4f36f99f7dd6ffd65a811aa261c11a5/src/lv_conf_internal.h#L744-L750

And mpy is still failing, the error is weird, if `LV_USE_OBJID` is not enabled, why it's complaing no definition to `lv_obj_assign_id`

kisvegabor:
In `lv_obj.h` we have the `obj_id` related functions defined, so when the binding generator parses the headers it assumes that these functions are also implemented somewhere. Therefore a binding will be generated for these functions. 

However the functions are disabled with `LV_USE_OBJID_BUILTIN` in [lv_obj_id_builtin.c](https://github.com/lvgl/lvgl/pull/4536/files#diff-d7f6c8e04ee6e3d6c271cdf16685762dd6daa77f5ddb11f8df5070c4db3ec9be). Therefor when the binding is being compiled, it will fail, as e.g.  `lv_obj_assign_id` is not implemented anywhere. 

To resolve it, all we need to do is adding `#if LV_USE_OBJID` around all 4 functions in `lv_obj.h`

kisvegabor:
Looks good, the CI is also passing. Could you add a documentation page and a simple test as well?

XuNeo:
Thank you Gabor. I will add the documenation and test next day.

kisvegabor:
It looks good. Would you like to add anything else?

XuNeo:
I think that's all I have. We have used this feature internally to debug UI crashes and it's usefull.
Could you please review and merge it?

kisvegabor:
Looks good to me! :clap:  :clap:  :clap:  Merging, thank you!

kisvegabor:
Oh, I've just noticed that `obj_id.rst` is not added to [index.rst](https://github.com/lvgl/lvgl/blob/master/docs/others/index.rst).

kisvegabor:
The failing CI is unrelated and already fixed in master so I merge it. 

Thank you for this very useful feature!

PR #4538, opened 2023-09-08
fix: build on Windows (MinGW environment)

defogprog:
### Fix build on Windows (MinGW environment)

On Windows environment edited lines (in files `lv_examples.mk` and `lv_extra.mk`) don't work properly, so edited them to work both on Linux and MinGW, like it's done in `lv_demos.mk`.


kisvegabor:
Thank you!

PR #4539, opened 2023-09-08
feat(indev):Add an indev reset event

xaowang96:
### Add an indev reset event

When someone calls lv_indev_reset, the active obj will always stay in the pressed state or other state. We need to send it a indev reset event to restore its default state.

### Checkpoints
 
### DIFF
--- a/src/core/lv_obj.c
+++ b/src/core/lv_obj.c
@@ -649,6 +649,9 @@ static void lv_obj_event(const lv_obj_class_t * class_p, lv_event_t * e)
     else if(code == LV_EVENT_DRAW_MAIN || code == LV_EVENT_DRAW_POST || code == LV_EVENT_COVER_CHECK) {
         lv_obj_draw(e);
     }
+    else if(code == LV_EVENT_INDEV_RESET) {
+        lv_obj_set_state(obj, LV_STATE_DEFAULT);
+    }
 }
 
 /**
--- a/src/core/lv_obj.c
+++ b/src/core/lv_obj.c
@@ -650,7 +650,8 @@ static void lv_obj_event(const lv_obj_class_t * class_p, lv_event_t * e)
         lv_obj_draw(e);
     }
     else if(code == LV_EVENT_INDEV_RESET) {
-        lv_obj_set_state(obj, LV_STATE_DEFAULT);
+        lv_obj_clear_state(obj, LV_STATE_PRESSED);
+        lv_obj_clear_state(obj, LV_STATE_SCROLLED);
     }
 }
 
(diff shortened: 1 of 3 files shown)


src/core/lv_obj.c
```diff
@@ -649,6 +649,9 @@ static void lv_obj_event(const lv_obj_class_t * class_p, lv_event_t * e)
     else if(code == LV_EVENT_DRAW_MAIN || code == LV_EVENT_DRAW_POST || code == LV_EVENT_COVER_CHECK) {
         lv_obj_draw(e);
     }
+    else if(code == LV_EVENT_INDEV_RESET) {
+        lv_obj_set_state(obj, LV_STATE_DEFAULT);
```
kisvegabor: It will clean all states including `DISABLED` or `CHECKED` which is clearly not the intention. 

It seems to me the whole issue applies only to the `PRESSED` state. So I wonder why can't we just remove the `PRESSED` state in `lv_indev_reset()`?

xaowang96:
@kisvegabor Hi, gabor, How do you think about we send an event to act_obj to let it reset its own state when indev is reseted?

kisvegabor:
I'm sorry, I just saw that I've written my review but haven't sent it. :sweat: (Sent now)

So my questions in short: are there any real life use case when we need to reset anything else than the `PRESSED` state.

src/core/lv_obj.c
xaowang96: You are right, I modify the code to only clear two states, `PRESSED` and `SCROLLED`

xaowang96:
> So my questions in short: are there any real life use case when we need to reset anything else than the `PRESSED` state?

Yes, we also need to consider the obj being scrolled. If we reset indev and do nothing for this, the scroll obj will stay in a wrong state. So, we can send this event to the scrolling obj to let it do something to make its own reset to normal state.

src/indev/lv_indev.c
```diff
@@ -344,11 +347,26 @@ void lv_indev_reset(lv_indev_t * indev, lv_obj_t * obj)
...
+            if(obj == NULL || indev->pointer.scroll_obj == obj) {
+                if(indev->pointer.scroll_obj) {
+                    /* Avoid recursive calls */
+                    scroll_obj = indev->pointer.scroll_obj;
+                    indev->pointer.scroll_obj = NULL;
+                    lv_obj_send_event(scroll_obj, LV_EVENT_INDEV_RESET, indev);
+                    scroll_obj = NULL;
+                }
+            }
         }
```
kisvegabor: The code in `if(indev) {...}` is repeated in the `else {...}` branch as well. Could you move these sections into a function like `indev_reset_core()`?

kisvegabor (commented):
You are right, `SCROLL` also needs to be cleared.

src/indev/lv_indev.c
xaowang96: Okey, I add a new function `indev_reset_core ` to reset indev.

kisvegabor:
Looks good, thank you!

PR #4541, opened 2023-09-08
Implemented variable binding (src/others/observable/lv_observable.*

zjanosy:
Implemented variable binding (src/others/observable/lv_observable.*)


src/others/observable/lv_observable.h
```diff
@@ -0,0 +1,263 @@
...
+ * @param obj the LVGL object what will be affected by the change
+ */
+void* lv_bind_formatted_int_to_label_text(lv_subject_t* subject, lv_subject_t* fmt, lv_obj_t* obj);
+
+/**
+ * Generic method to subscribe to an int value with a user-defined callback
+ * @param subject address of the subject
+ * @param cb the callback called when the subject changes
+ * @param obj the LVGL object what will be affected by the change
+ * @param data1 optional additional parameter stored in the observer object
```
kisvegabor: For the MicroPython binding we will need a `user_data` field in `lv_subject_t`. I wonder if we can use only a single `user_data` instead of 2 `data`.

src/others/observable/lv_observable.h
```diff
@@ -0,0 +1,263 @@
...
+ * @param fmt the format string
+ */
+void* lv_bind_int_to_label_text_fmt(lv_subject_t* subject, lv_obj_t* obj, const char* fmt);
+
+/**
+ * Bind and int value to a label text using a variable format string
+ * @param subject address of the subject
+ * @param fmt the address of the format string (an observable string value)
+ * @param obj the LVGL object what will be affected by the change
+ */
```
kisvegabor: Add
```c
@note if any of the subjects changes the label's text will be updated.
```

src/others/observable/lv_observable.h
```diff
@@ -0,0 +1,263 @@
...
+void* lv_bind_bitmask_to_obj_state(lv_subject_t* subject, lv_obj_t* obj, lv_state_t flag, lv_bitmask_pattern_t pattern);
+
+/**
+ * Bind int value to a local style property
+ * @param subject address of the subject
+ * @param obj the LVGL object what will be affected by the change
+ * @param prop the style property
+ * @param selector the part/state selector
+ */
+void* lv_bind_int_to_obj_local_style_prop(lv_subject_t* subject, lv_obj_t* obj, lv_style_prop_t prop, lv_style_selector_t selector);
```
kisvegabor: I find it too special to have only `int`s. I would remove it in the first pass, a reconsider it in a new PR.

src/others/observable/lv_observable.h
```diff
@@ -0,0 +1,263 @@
...
+ */
+void lv_subject_copy_string(lv_subject_t* subject, char* buf);
+
+/**
+ * Set bitmask and value comparison pattern of the observer
+ * @param observer address of the observer
+ * @param pattern pattern used to mask out interested bits from the bitmask's value and compare it with a constant value
+ * @note The value
+ */
+void lv_observer_bitmask_set_cond(lv_observer_t* observer, lv_bitmask_pattern_t pattern);
```
kisvegabor: It seems it's not used anymore.

src/others/observable/lv_observable.h
```diff
@@ -0,0 +1,263 @@
...
+ * @param data2 optional additional parameter stored in the observer object
+ * @return the observer object
+ * @note This is a generic interface, which is normally not called directly. For specific purposes there are specific wrappers.
+ * @note The observer object is allocated on the heap. It will be freed automatically when the object is deleted or when it is unsubscibed from the subject.
+ */
+void* lv_subject_subscribe_obj(lv_subject_t* subject, lv_observer_cb_t cb, lv_obj_t* obj, void* data1, void* data2);
+
+/**
+ * Remove observer from the list of subscribers
+ * @param subject address of the observer object returned when a widget was subscribed to the subject
```
kisvegabor:
```suggestion
 * @param observer address of the observer object returned when a widget was subscribed to the subject
```

src/others/observable/lv_observable.h
```diff
@@ -0,0 +1,263 @@
...
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+/*********************
+ *      INCLUDES
+ *********************/
+
+#include "lvgl/lvgl.h"
```
kisvegabor: Include only the required header from a relative path. E.g.
```c
#include "../../misc/lv_ll.h"
```

src/others/observable/lv_observable.h
```diff
@@ -0,0 +1,263 @@
...
+
+/*********************
+ *      DEFINES
+ *********************/
+
+#define LV_SUBJECT_TYPE_INT     0 /**< an int32_t*/
+#define LV_SUBJECT_TYPE_POINTER 1 /**< a void pointer*/
+#define LV_SUBJECT_TYPE_COLOR   2 /**< an lv_color_t*/
+#define LV_SUBJECT_TYPE_FLOAT   3 /**< a float*/
+#define LV_SUBJECT_TYPE_BITMASK 4 /**< an array of up to 15 binary flags*/
```
kisvegabor: The comment seems outdated.

kisvegabor (commented):
Thank you for the updated.

src/others/observable/lv_observable.h
zjanosy: Could you be more specific? What type should user_data be? How can you replace the two variables with one without dynamic allocation?

src/others/observable/lv_observable.h
zjanosy: Added:
@note If either the value or the formwat string changes the label's text will be updated.

src/others/observable/lv_observable.h
zjanosy: We could rename it to 'lv_bind_subject_to_obj_local_style_prop', since the subject value can contain other types as well, and there is no 'int' specific in this function.

src/others/observable/lv_observable.h
zjanosy: There was a mismatch between the header and the implementation. Now I have renamed it to 'lv_observer_bitmask_set_pattern'.

src/others/observable/lv_observable.h
zjanosy: Good catch. Actually, the implementation was using a wrong name as well.

src/others/observable/lv_observable.h
zjanosy: Fixed.

src/others/observable/lv_observable.h
kisvegabor: It should be `void * user_data`.

We can do dynamic allocation :)

src/others/observable/lv_observable.h
kisvegabor: How can we decide which value to pass to the style?

zjanosy: I believe we can simply cast the lv_subject_value_t to lv_style_value_t.

src/others/observable/lv_observable.h
zjanosy: Yes, we can, but shall we? In most cases there is only a single parameter. We only have two parameters (so far) for style properties. Do we want to allocate memory for a single int as well? It seems like an overkill to me.
We could also implement some kind of inheritance, so that we could derive new "classes" with additional data if needed. I don't know how this would work with MicroPython though.

src/others/observable/lv_observable.h
kisvegabor: `float` needs to removed anyway, because LVGL is integer only. Without it the  `lv_subject_value_t` and `lv_style_value_t` will be quite the same.
 
In v9 
- there will be a `LV_USE_FLOAT` option to enable then on some places. 
- we can use a single `lv_value_t` on both places. 

However in v8 we should align this feature to what we already have there.

kisvegabor:
Let's prepare it to merged to `master` first and later back port it to v8.

kisvegabor:
I merged it into `master`. Probably the docs still needs to be updated when we will see how it looks like on the web page.

PR #4559, opened 2023-09-14
fix(docs): Fix small mistakes in grid layout docs

niklasf:


kisvegabor:
Both are correct, thank you!

PR #4568, opened 2023-09-19
Release/beta/pxp vglite acceleration v9

nicusorcitu:

### Checkpoints
 

kisvegabor:
Thank you Nicu! It looks good in general. Instead of discussing minor things I will like to merge it to avoid conflicts.

We can send smaller PRs later if needed.

nicusorcitu:
thank you. please let me know what are your concerns.

kisvegabor:
Have you already made any benchmarking?

nicusorcitu:
Not yet. This is the next step.
Trimis: Thursday, September 21, 2023 10:58:28 AM
Către: lvgl/lvgl ***@***.***>
Cc: Nicușor Cîțu ***@***.***>; Author ***@***.***>
Subiect: [EXT] Re: [lvgl/lvgl] Release/beta/pxp vglite acceleration v9 (PR #4568)

Have you already made any benchmarking?

kisvegabor:
Okay, very exciting :slightly_smiling_face:

nicusorcitu:
I am trying to run the lvgl tests on NXP target.
I did run:
`scripts/install-prerequisites.sh`
`./tests/main.py --clean --report build test`
But when I inspected the `tests/report/index.html` I found only the SW draw files have been used.

Can you please tell me how do I enable the GPU draw and how can I run the tests the target?
Thanks.

kisvegabor:
I can't imagine a good way to run the unit tests on a hardware. Each unit test is an executable, so to run them on a target you needed to flush the device hundreds of times. :slightly_frowning_face:

XuNeo:
> I can't imagine a good way to run the unit tests on a hardware. Each unit test is an executable, so to run them on a target you needed to flush the device hundreds of times. 🙁

Is it possible to make the test run on target? We have mentioned it internally that it would be great to run through test cases on hardware whenever making a big change.

I think we can open a new issue to track it. No hurry yet.

kisvegabor:
Right now, each unit test is a separate "unit" and goes to a separate executable. For example example almost all test cases have a "setUp" and a "tearDown" functions. See [here](https://github.com/lvgl/lvgl/blob/master/tests/src/test_cases/test_msg.c). Therefore it's not possible to compile all tests into a single binary as it is.

Although we could call `setUp` and `tearDown` manually at beginning and end of each test case. After that we could simply call all the `test_*` functions one by one.

anaGrad:
I tried to just call one test, and I got stuck at the fact that `unity_support.c` includes `png.h`. Do you know what version of libpng is needed and which files should we keep from the library?

nicusorcitu:
Hi @kisvegabor 
1. We tried to build libpng as well as the dependency zlib but no much success.
2. The alternative would be to make use of the single header png libs:
https://github.com/nothings/stb/blob/master/stb_image.h
https://github.com/nothings/stb/blob/master/stb_image_write.h
That would require to rewrite `read_png_file` `write_png_file` and `png_release` from
https://github.com/lvgl/lvgl/blob/9b069721041ea24ec39c21b2d8c17fadec2efffd/tests/unity/unity_support.c#L50
to use the APIs from the above headers.

In this way we might be able to make the tests to run on any target. What do you think?
Thanks.

kisvegabor:
@nicusorcitu we can do that, but how will you upload all the reference PNG files to the target board?

kisvegabor:
Exactly what's the point of running the tests on an MCU? Note that screenshot compare won't work, as VG-Lite rotates the images slightly differently for sure. Even letter blending won't be exactly the same. 

If the goal is to test a lot of rendering cases, I would rather create a demo in which a lot scenarios can be inspected. Similar to the benchmark demo, but by showing various drawings on pages and have previous/next buttons so that a human can check the result one by one.

What do you think?

nicusorcitu:
1. We can do otherwise: run the tests and dump the FB (screenshot) as png and transfer them over the serial on the host in a folder having the name of the test. The comparation with the reference images can be done on the host after the tests runs completely. We don;t have to make the png diff on target.
2. I know there is no 1:1 comparation between CPU output and GPU output. We might have some scripts to evaluate the PSNR (peak signal-to-noise ratio) or SSIM (structural similarity index measure) in order to measure the similarity between images.

nicusorcitu:
The goal is to compare the similarity between the images on various scenarios. And to have a kind of acceptance test suite.
When there is any degradation we know it.

kisvegabor:
I have 2 concerns with using the unit tests on MCUs:
- The unit test system is not designed to support these kind of scenarios, so it might require significant amount of works to improve it this way
- Running the screenshot compare this way requires a lot of work from the chip vendor's side as well. It's amazing if we have a SSIM system in place for NXP, but I'm not sure if we can ask it from all the chip vendors.

I would suggest starting with something simpler that we can improve later on. 

What I have in mind is to have the following pages (each has a "with opacity" or "non-normal blend mode" variant as well):
1. fill (radius + gradient too)
2. border (each sides + radius)
3. outline (radius) 
4. box shadow (offset, size, spread)
5. text (normal text + underline/strike through, placeholder)
6. triangle (just some rectangles)
7. image (various formats + transformation)
9. line (various angles + line caps)
10. arc (some arcs + caps)
11. vector (later)
8. layer (blend mode, transformation)
12. mask bitmap (not implemented SW render yet)
13. mask rectangle

E.g. for the border draw task we can have these border types: 
1. full 
2. left 
3. right
4. top
5.  bottom
6.  top-bottom
7.  left-right
8.  top-left
9.  bottom-right

with 
  - no radius, 1 px width
  - no radius, 5 px width: 
  - 5 px radius, 1 px width
  - 5 px radius, 5 px width
  - 50 px radius, 1 px width
  - 50 px radius, 5 px width
 
So 10 columns (title + 9 border side options) and 7 rows (6 width/radius + 1 for control and info). On a 480x272 screen we can have a 48 x 38 -> 40 x 30 rectangles. I think it's ok.

As control on the last row:
- next/previous
- full refresh/partial refresh
- render times and CPU load

What do you think?

nicusorcitu:
> It's amazing if we have a SSIM system in place for NXP, but I'm not sure if we can ask it from all the chip vendors.
>
It is just a short pythin script we have to use once we have the images to compare. We have done it for other project but I intend to make something similar for lvgl accelerator too.

We do already have a unit test list that covers all the NXP draw tasks. You already had a chance to see some of the screenshots.
For example - the rotation and recolor. Or draw a simple line, or draw a dashed line with rounded edges.

The intend is to see how those unit tests can be made automatically - e.g. to run CPU only and then to compare with GPU.

So I saw the automated system upstream and I thought that making it to run on target will simplify a lot.

BTW, introspecting the size of the test images I see:
```
nxf68195@lw15016 ~/bitbucket/lvgl_mcu/lvgl/tests (feature/nxp_v9)$ du -h
220K    ./src/test_files/binimages
228K    ./src/test_files
4,6M    ./src/test_assets
28K     ./src/test_cases/libs
196K    ./src/test_cases/widgets
68K     ./src/test_cases/draw
420K    ./src/test_cases
5,3M    ./src
44K     ./ref_imgs/libs
1,9M    ./ref_imgs/draw
2,7M    ./ref_imgs
288K    ./unity
12K     ./makefile
21M     ./report
29M     .
nxf68195@lw15016 ~/bitbucket/lvgl_mcu/lvgl/tests (feature/nxp_v9)$
```
The images are in `2,7M    ./ref_imgs` and `4,6M    ./src/test_assets`, they could fit on target too in SDRAM or FLASH.

kisvegabor:
I've opened #4640 to discuss it.  I've added your suggestions to the initial comment.

PR #4579, opened 2023-09-20
feat(obj): add unified obj property set/get API

XuNeo:

Add unified obj set/get property API

Example:
```c

static void obj_property_example(void)
{
#define MY_PROPERTIES_COUNT 7

  LV_IMG_DECLARE(img_demo_widgets_avatar);

  /*Pointer type of property value*/
  const lv_point_t pivot_50 = {
    LV_PCT(50),
    LV_PCT(50),
  };

  /*Const properties*/
  const struct _obj_info_t {
    const lv_obj_class_t *clz;
    const int properties_count;

    lv_property_t properties[MY_PROPERTIES_COUNT];
  } my_ui[] = {
    {
      .clz = &lv_image_class,
      .properties_count = MY_PROPERTIES_COUNT,
      .properties = {
        { .id = LV_PROPERTY_IMAGE_SRC, .ptr = &img_demo_widgets_avatar, },
        { .id = LV_PROPERTY_IMAGE_PIVOT, .ptr = &pivot_50, },
        { .id = LV_PROPERTY_IMAGE_ROTATION, .num = 900, },
        { .id = LV_PROPERTY_IMAGE_SCALE, .num = 128, },
        { .id = LV_PROPERTY_OBJ_HIDDEN, .num = 0, },
        { .id = LV_PROPERTY_OBJ_CLICKABLE, .num = 1, },
        { .id = LV_STYLE_IMAGE_OPA, .num = 128, },
      }
    }
  };

  lv_obj_t *root = lv_obj_create(lv_scr_act());
  lv_obj_center(root);
  lv_obj_set_size(root, 200, 200);

  /*Runtime generated properties*/
  lv_property_t prop_parent = {
    .id = LV_PROPERTY_OBJ_PARENT,
    .ptr = root,
  };

  for (int i = 0; i < sizeof(my_ui)/sizeof(my_ui[0]); i++) {
    const struct _obj_info_t *info = &my_ui[i];
    lv_obj_t * obj = lv_obj_class_create_obj(info->clz, lv_scr_act());
    lv_obj_class_init_obj(obj);

    for (int j = 0; j < info->properties_count; j++) {
      lv_obj_property_set(obj, &info->properties[j]);
    }

    lv_obj_property_set(obj, &prop_parent);

    printf("obj %d: %p\n", i, obj);
    printf("clickable: %d\n", lv_obj_has_flag(obj, LV_OBJ_FLAG_CLICKABLE));
    printf("hidden: %d\n", lv_obj_has_flag(obj, LV_OBJ_FLAG_HIDDEN));
    lv_property_t prop;
    prop = lv_obj_property_get(obj, LV_PROPERTY_OBJ_CLICKABLE);
    printf("clickable2: %d\n", lv_obj_property_get(obj, LV_PROPERTY_OBJ_CLICKABLE).num);
    printf("hidden: %d\n", lv_obj_property_get(obj, LV_PROPERTY_OBJ_HIDDEN).num);
  }
}
```
### Checkpoints
 
Edit: update code example.


XuNeo:
Inside `lvgl/src/misc/lv_obj_property.h`, I have wrapped my code with `LV_USE_OBJ_PROPERTY`. But CI is still complaining.

src/misc/lv_obj_property.h
```diff
@@ -0,0 +1,134 @@
...
+ */
+
+#ifndef LV_OBJ_PROPERTY_H
+#define LV_OBJ_PROPERTY_H
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+#if LV_USE_OBJ_PROPERTY
```
kisvegabor: You should include `lv_conf_internal.h` first (or an other header file which indirectly includes it) to see if `LV_USE_OBJ_PROPERTY` is set or not.

kisvegabor:
Getting closer. The rp2 issue can be resolved by rebasing on master, however I have no idea why only the ESP32 build complains. Especially as you enabled `LV_USE_OBJ_PROPERTY` by default.

kisvegabor:
Finally what solved the ESP MicroPython build issue?

I think later we should extend this feature to 
- animations
- display
- indev

to have unified handling of each major component of LVGL (styles are already working like this, but it might make sense to unify them with this concept too)

If you agree, at this stage I think what we need to is just avoid `OBJ` specific names and keep it as generic as possible.

XuNeo:
> Finally what solved the ESP MicroPython build issue?

Ah, I have disaled `LV_USE_OBJ_PROPERTY ` by default. We should check it again why only ESP32 failed.

For widget, I haven't think about how to unify style to this. It should be unified since style is also a property. I'll take a look next week.

I think `lv_obj_set_property` is specially for `obj`. For other modules like `animation`, I haven't seen a need for unified API to set parameters. 

The main reason it's needed for `obj` is that we could write many custom `obj`s that have different properties. Module like `animation` has  fixed parameters by `lvgl` thus no such need.

kisvegabor:
> I think lv_obj_set_property is specially for obj. For other modules like animation, I haven't seen a need for unified API to set parameters.

Probably you are right. It would be overkill for those.

XuNeo:
The main part is finished. I'll add a simple doc and proper test late today.

kisvegabor:
Awesome 👍

XuNeo:
Hi Gabor,
CI has passed, could you help to review the code? I have commited all I have for now. Thanks.

src/core/lv_obj.h
```diff
@@ -138,6 +142,48 @@ typedef _lv_obj_flag_t lv_obj_flag_t;
 typedef uint32_t lv_obj_flag_t;
 #endif /*DOXYGEN*/
 
+#if LV_USE_OBJ_PROPERTY
+enum {
+    /* obj property of flags */
+    LV_PROPERTY_ID(OBJ, FLAGS_START,            LV_PROPERTY_TYPE_INT,       1),
```
kisvegabor: I wonder if we can make it easier to maintain and automatic to follow the flag changes. 

Can we say that flag start is `0x8000 0000` (1 << 31) and the values are e.g. `0x80000000 | LV_OBJ_FLAG_HIDDEN`?

An other option is to add a unit test for each flag so that we can notice if there are any inconsistencies. Adding unit test seems more elegant.

What do you think?

kisvegabor (commented):
Looks great, just one comment.

src/core/lv_obj.h
XuNeo: I think I have overlooked it. We can simply regard `flags` as one property.
Set/get will modify/read obj->flags directly with property->value. User should read/modify/write the flags.

I'll update the commit to see if that works more clearly.

bjsylvia (commented):
Are we also considering float type? The subsequent support for scaling will expand the float type support for widgets.

XuNeo:
No problem, I can add it.

src/core/lv_obj.h
XuNeo: I have added it in https://github.com/lvgl/lvgl/pull/4655
The property `FLAGS` can be used to remove or add flags based on bit31 in value, using the way you suggested.

src/misc/lv_obj_property.c
```diff
@@ -0,0 +1,179 @@
...
+
+/**********************
+ *      MACROS
+ **********************/
+
+/**********************
+ *   GLOBAL FUNCTIONS
+ **********************/
+
+lv_result_t lv_obj_property_set(lv_obj_t * obj, const lv_property_t * value)
```
kisvegabor: What about this?
```c
lv_result_t lv_obj_set_property(lv_obj_t * obj, uint32_t id, const lv_value_group_t value)
```

`lv_value_group_t` could be used here, in styles and in [lv_observer](https://github.com/lvgl/lvgl/blob/master/src/others/observer/lv_observer.c) as well.

XuNeo: Let's add `lv_value_group_t` in another PR.

XuNeo:
Added back obj flag properties, added test case for it.

kisvegabor:
Thank you. Do you agree with it as well or just added back because I asked? :smile:

XuNeo:
Hi Gabor, I agree with the solution. It's the original method.
It's not ideal to define flags twice but I think it's acceptable for now. Until we find a better method, let's go with it.

src/misc/lv_obj_property.c
```diff
@@ -0,0 +1,179 @@
+/**
```
kisvegabor: As these files into the `core` folder.

src/misc/lv_obj_property.c
```diff
@@ -0,0 +1,179 @@
...
+
+/**********************
+ *      MACROS
+ **********************/
+
+/**********************
+ *   GLOBAL FUNCTIONS
+ **********************/
+
+lv_result_t lv_obj_property_set(lv_obj_t * obj, const lv_property_t * value)
```
kisvegabor: I wonder if we can have a function later like:
```c
#define lv_obj_set(obj, prop, v)    lv_obj_propery_set_direct(obj, prop, (lv_value_group_t) v);
```
(The names are just examples)

kisvegabor: For the sake of consistency I suggest using the `lv_obj_set_property` word order.

kisvegabor (commented):
I've just tried it out and it looks and works great. Allowing to use an array of properties is huge! :+1: 

I've added a few simple comments.

src/misc/lv_obj_property.c
XuNeo: We can simply rename `lv_obj_set_property` to `lv_obj_set`? 
Let's firstly see how `lv_obj_set_property` can be used. If it can be widely used, then I suppose it's better to add `lv_obj_set`.

kisvegabor: Sounds good to me. We can easily refactor it at any time.

kisvegabor:
LGTM! Let me know when I can hit that merge button! :slightly_smiling_face:

XuNeo:
Hello Gabor, I'm good to go.

PR #4621, opened 2023-10-02
fix(dev): do not wait for vsync between partial linux drm flushes

niklasf:
The driver currently waits for vsync between each potentially partial flush, killing performance. Instead, check `lv_display_flush_is_last()`. Seing 5x FPS improvement for `lv_demo_benchmark()`.

Based on style-adjustments in individually reviewable commits.


kisvegabor:
Amazing, thank you for the update!

PR #4636, opened 2023-10-09
feat(image_decoder): add post-process image method

FASTSHIFT:

Related discussion: https://github.com/lvgl/lvgl/issues/4382

### Checkpoints
 

kisvegabor:
Thank you!

For what purpose do you use the `user_data`s?

FASTSHIFT:
> Thank you!
> 
> For what purpose do you use the `user_data`s?

Use `user_data` to pass the context pointer, which is the case for most callbacks of lvgl.

src/draw/lv_image_decoder.h
```diff
@@ -44,6 +44,19 @@ typedef _lv_image_src_t lv_image_src_t;
 typedef uint8_t lv_image_src_t;
 #endif /*DOXYGEN*/
 
+enum _lv_image_decoder_process_flag_t {
+    LV_IMAGE_DECODER_PROCESS_FLAG_NONE = 0,
```
kisvegabor: As the pre/post process callbacks are application specific, I wonder if we need to add a predefined flags. 

I think having `process_state` (instead of `process flags` as `process state is more generic`) is enough.

kisvegabor (commented):
Looks good in general! Just one comment. 

Please also add a few sentences about this feature in the docs, mention it's purpose, some use-cases, etc.

src/draw/lv_image_decoder.h
FASTSHIFT: Done

FASTSHIFT:
@kisvegabor 
I noticed that the `process_state` in `lv_image_decoder_dsc_t` does not seem to be able to maintain state with the cache life cycle, which will lead to multiple post-processings because dsc is stored in local variables instead of being managed by the cache. Is there any good way?

kisvegabor:
Ouch, that's true. 

So the changes done by the pre/post process callbacks (setting of `process_state`) should be saved into the cache as well, but the cache is not used at this point yet. 

What if we save `dsc->process_state` [here](https://github.com/lvgl/lvgl/blob/7506b61527195fdc96803fe94f5ba141b6e5a9e8/src/libs/lodepng/lv_lodepng.c#L199) in the cache? `cache_entry->user_data` seems like a good place for that. 

Once it's saved in the cache, it can be retrieved too in `<format>_decoder_open()`.

What do you think?

FASTSHIFT:
> Ouch, that's true.
> 
> So the changes done by the pre/post process callbacks (setting of `process_state`) should be saved into the cache as well, but the cache is not used at this point yet.
> 
> What if we save `dsc->process_state` [here](https://github.com/lvgl/lvgl/blob/7506b61527195fdc96803fe94f5ba141b6e5a9e8/src/libs/lodepng/lv_lodepng.c#L199) in the cache? `cache_entry->user_data` seems like a good place for that.
> 
> Once it's saved in the cache, it can be retrieved too in `<format>_decoder_open()`.
> 
> What do you think?

I moved the `process_state` to `lv_cache_entry_t` and it seems to be working fine.
This is the processing code for premultiplication after PNG decoding:
```c
static lv_result_t my_image_post_process(lv_image_decoder_dsc_t* dsc, void* user_data)
{
    if (dsc->header.cf == LV_COLOR_FORMAT_ARGB8888) {
        lv_cache_entry_t * entry = dsc->user_data;

        if (!(entry->process_state & LV_IMAGE_DECODER_PROCESS_STATE_PREMULTIPLIED_ALPHA)) {
            LV_LOG_USER("premultiplied dsc = %p", dsc);
            lv_color32_t* image = (lv_color32_t*)dsc->img_data;
            uint32_t px_cnt = dsc->header.w * dsc->header.h;

            while (px_cnt--) {
                image->red = LV_UDIV255(image->red * image->alpha);
                image->green = LV_UDIV255(image->green * image->alpha);
                image->blue = LV_UDIV255(image->blue * image->alpha);
            }

            entry->process_state |= LV_IMAGE_DECODER_PROCESS_STATE_PREMULTIPLIED_ALPHA;
        }
    }

    return LV_RESULT_OK;
}
```

FASTSHIFT:
I think the decoder preprocessing callback may have no effect, because it does not directly affect the behavior of the decoder, nor can it obtain the `process_state` status of `lv_cache_entry_t`. So just keep the post-processing callback interface.

FASTSHIFT:
Can we specifically add a pointer of type `lv_cache_entry_t` in `lv_image_decoder_dsc_t`? Because the user may not necessarily put the `cache_entry` pointer in `user_data`, we cannot assume in post-processing that `user_data` must point to `cache_entry`.

kisvegabor:
I'm ok with removing the pre-processing.

Having cache entry seems doable too, just please update the current decoders accordingly.

kisvegabor:
Something just came to my mind. Imagine this situation:
- We have ARGB8888 image to draw
- When the draw task is created it's assigned to a very simple GPU, called GPU1
- When the image is decoded in the actual draw function of GPU1 its format will be changed ARGB8888-premultiplied
- And let's say GPU1 can't handle premultiplied images

To solve it we need to have a callback in the decoder info callback as well, to modify the format according the final format after post processing.

FASTSHIFT:
> Something just came to my mind. Imagine this situation:
> 
> * We have ARGB8888 image to draw
> * When the draw task is created it's assigned to a very simple GPU, called GPU1
> * When the image is decoded in the actual draw function of GPU1 its format will be changed ARGB8888-premultiplied
> * And let's say GPU1 can't handle premultiplied images
> 
> To solve it we need to have a callback in the decoder info callback as well, to modify the format according the final format after post processing.

I think that if GPU1 cannot support drawing the corresponding image, it should not modify any image content. The GPU should be responsible for its own post-processed images and should not leave it to other GPUs.

FASTSHIFT:
How about putting the post-processing callback into the draw_unit structure? We can pass in the draw_unit pointer in lv_image_decoder_open and let draw_unit decide how to post-process.

kisvegabor:
Hm, that's interesting. So let's say all ARGB8888 are assigned to GPU1. So it can safely do any magic on them for itself. 

However, why is good for the GPU to convert the image pre-multiplied? Is it faster to blend even for a GPU?

FASTSHIFT:
> Hm, that's interesting. So let's say all ARGB8888 are assigned to GPU1. So it can safely do any magic on them for itself.
> 
> However, why is good for the GPU to convert the image pre-multiplied? Is it faster to blend even for a GPU?

Blend Normal: S * Sa + (1 - Sa) * D
Blend Src Over:  S + (1 - Sa) * D

Because the GPU hardware implementation of SRC OVER's blend mode requires fewer logic gates, more hardware cost savings, the disadvantage is that you need to provide pre-multiplied images.

kisvegabor:
Clear, there might be some power saving as well, but I guess it's negligible.

What if the `draw_unit` just calls a function after decoding? E.g.
```c
lv_image_decoder_open(...)
post_process(decoder_dsc->img_data, ...);
```

`lv_image_decoder_open` set save `lv_cache_entry_t *` in `decoder_dsc` so `post_process` can update the flags in it.

It seems cleaner and more flexible to me.

XuNeo:
Since lvgl already takes image stride alignement, start address alignment into consideration, is it reasonable we pass parameters to decoder to describe how to decode the image. And put decoded image(draw buffer ideally) to cache together with its parameter. 

Every draw unit may call decoder_open with therir preferred parameters.

If pre-multiply is a must feature, then let's support it in lvgl layer and avoid so many custom callbacks and post processing.

FASTSHIFT:
> Clear, there might be some power saving as well, but I guess it's negligible.
> 
> What if the `draw_unit` just calls a function after decoding? E.g.
> 
> ```c
> lv_image_decoder_open(...)
> post_process(decoder_dsc->img_data, ...);
> ```
> 
> `lv_image_decoder_open` set save `lv_cache_entry_t *` in `decoder_dsc` so `post_process` can update the flags in it.
> 
> It seems cleaner and more flexible to me.

Do you mean to let `draw_unit` call `post_process` API manually? I think it is also a good way to handle it, so there is no need to add additional callback functions.

FASTSHIFT:
I added the lv_image_decoder_post_process function and it works like this:
```c
void gpu_draw_image(lv_draw_unit_t * draw_unit, const lv_draw_image_dsc_t * draw_dsc, const lv_area_t * coords)
{
    ...
    lv_image_decoder_dsc_t decoder_dsc;
    lv_result_t res = lv_image_decoder_open(&decoder_dsc, draw_dsc->src, draw_dsc->recolor, -1);
    if(res != LV_RESULT_OK) {
        LV_LOG_ERROR("Failed to open image");
        return;
    }

    res = lv_image_decoder_post_process(&decoder_dsc);
    if(res != LV_RESULT_OK) {
        LV_LOG_ERROR("Failed to post process image");
        return;
    }
    ...
}
```
I think it can meet the current needs :)

kisvegabor:
Could you handle caching of the post processed image as well?

FASTSHIFT:
> Could you handle caching of the post processed image as well?

Yes, you can get the `dsc->cache_entry` pointer in the `post_process` callback function for processing.

FASTSHIFT:
What do you think of this design? I will update the document later.

FASTSHIFT:
Document update completed ;)

kisvegabor:
I'm sorry but it seems there is a slight misunderstanding. If we can call any custom `post_process_cb` after `decoder_open` what's the purpose of registering a callback? I think the user can just call any functions he wants without any changes in LVGL. (We should still document this approach though)

FASTSHIFT:
> ```c
> process_state 
> ```

If this is the case, adding a callback function does not make much sense. We can only keep` cache_entry->process_state` in this PR to record the post-processing status, which is more concise.

FASTSHIFT:
> I agree. 👍 (And a little bit of docs 🙂 )

Done🙂.

kisvegabor (commented):
Looks cool, merging :clap: :clap: :clap:

FASTSHIFT:
> Looks cool, merging 👏 👏 👏

Can we merge?

kisvegabor:
Sorry 😅

Issue #4347, opened 2023-07-06
Reconsider the Micropython an other binding examples


kisvegabor:
Now the Micropython examples are one-to-one copy of the C examples. As a few years have been passed now I see a few issues with this approach:
- If anyone changes or adds a new C example they need to add and test the related Micropython and all the other binding  example too. It requires a lot of background knowledge from all the bindings and their environment setup. 
- Now we have 100+ examples. If someone wants to add a new binding  with all the examples it's an unreasonably huge work
- What is possible and straight forward in C is not necessary possible and typical in a binding. Therefore if the C example is the base, the binding examples might feel unnatural. Or if we follow the C way of thinking there might be binding features that might not be showcased.

My suggestions is to decouple the binding examples form the C examples. For example we can have 5 C examples for a `lv_btn` to show all the things LVGL can do with button. Next to that we can have 3 MicroPyhon examples show casing e.g. a simple button; a styled button; and some Pythonic magic which has no C counterpart. And let's say we can also have only 1 JavaScript example.

What do you think?

cc @amirgon @uraich

pete-pjb:
I agree with this way of thinking @kisvegabor .

kdschlosser:
If the API of the bindings followed as close as possible to the C API this would actually be not so much of an issue. Provided that the examples are not overly complex, Writing a script that would convert the examples to their specific language would not be a big thing to do.

I personally feel that LVGL should only have examples in C and the MicroPython binding should have examples with it and so on and so forth. The mixing of LVGL C and LVGL MicroPython in documentation and also in code should not be done. It makes things more difficult to maintain because You are always having to think about a binding and what it is going to effect. Make whatever changes to LVGL needs to be made without any thought for any bindings this includes examples and let the maintainers of the binding change what needs to be changed in the binding to make it work or submit a PR to LVGL to make a change to fix an issue in a binding so long as the change isn't specific to a single binding. What I mean by that is if a change is made and that change breaks 3 bindings but fixes one and the only purpose to the change was to fix the one, probably not a good change to make.

kdschlosser:
same thing goes for tests. and compilation. no binding should get compiled at the same time as LVGL. the test cases should not be mixed together.

I am a fan of visual help when trying to explain things.

![image](https://github.com/lvgl/lvgl/assets/10932297/1b177612-5351-40de-9062-867824e11911)

[...]

PR #4645, opened 2023-10-10
MicroPython test

ghost:

Created a small LVGL MicroPython-binding testing framework and test-cases.
Please read the short README.md for further details... and of course don't hesitate to ask...

### Checkpoints
 

tests/micropython/test_cases/test_animation.py
```diff
@@ -0,0 +1,54 @@
+#! ../../../../../../../lv_micropython/ports/unix/build-standard/micropython ../run_testcase.py
+#LVGL MicroPython binding tester script: Animation to change widget-position, start & ready callbacks
+
+LV_TESTANIM_FRAMERATE = 10 #ms
+LV_TESTANIM_LOOPCOUNT = 30
+
+
+#test-objects:
+
+lv_testbutton = lv.button( lv.scr_act() )
```
kisvegabor, line 10: The `lv_` prefix is not required for variables like these.  Having the `lv_` prefix suggests that it is related to LVGL and not the user's code.

Feel free to call it just `testbtn`.

tests/micropython/test_cases/test_animation.py
```diff
@@ -0,0 +1,54 @@
...
+lv_testanim.set_path_cb(lv.anim_t.path_overshoot)
+lv_testanim.set_custom_exec_cb( lambda a,v: lv_testbutton.set_x(v) )
+
+lv_testsignal_start = False
+lv_testsignal_ready = False
+
+
+#tests:
+
+def lv_testanim_start_cb (o):
```
kisvegabor, line 26: [Same](https://github.com/lvgl/lvgl/pull/4645#discussion_r1353262601) for the functions.

tests/micropython/test_cases/test_widgets.py
```diff
@@ -0,0 +1,69 @@
+#! ../../../../../../../lv_micropython/ports/unix/build-standard/micropython ../run_testcase.py
+#LVGL MicroPython binding tester script: Widget-creation/deletion and basic flag/property-settings
+
+
```
kisvegabor, line 4: The purpose of this test would be to see
- if we can create widgets
- deleting widgets doesn't cause crash
- various data types can be passed. For this is recommend setting and getting: 
    - label text (test string)
    - an external array for a chart (integer array)
    - points for a line widget (point array)
    - the map of a button matrix (string array) 
    - options of the dropdown (multi line string) 
    - passing flags and states (enum)

I think having test only for these widgets with a lot of `gc.collect()` calls should be suitable.

kisvegabor (commented):
Thank you! I've added a few comments. I'll ask others to take look as well.

tests/micropython/test_cases/test_filesystem.py
```diff
@@ -0,0 +1,62 @@
...
+lv_filestatus = lv_testfile.open( "S:" + LV_TESTFILE_NAME, lv.FS_MODE.WR )
+
+lv_test.assert_equal( lv.FS_RES.OK, lv_filestatus, "File-creation (with open)" )
+
+
+lv_written_bytes = [0]
+lv_filestatus = lv_testfile.write( lv_testarray , LV_TESTLIST_SIZE, bytearray(lv_written_bytes) )
+
+lv_test.assert_equal( lv.FS_RES.OK, lv_filestatus, "Writing into file (status)" )
+#lv_test.assert_equal( LV_TESTLIST_SIZE, bytearray(lv_written_bytes)[0], "Writing into file (count)" )
```
kisvegabor, line 39: TODO

tests/micropython/test_cases/test_filesystem.py
```diff
@@ -0,0 +1,62 @@
...
+lv_filestatus = lv_testfile.open( "S:" + LV_TESTFILE_NAME, lv.FS_MODE.RD )
+
+lv_test.assert_equal( lv.FS_RES.OK, lv_filestatus, "Opening file for reading" )
+
+
+lv_read_bytes = 0
+lv_filestatus = lv_testfile.read( lv_readback_array, LV_TESTLIST_SIZE, bytearray(lv_read_bytes) )
+
+lv_test.assert_equal( lv.FS_RES.OK, lv_filestatus, "Reading back data from file (status)" )
+#lv_test.assert_equal( LV_TESTLIST_SIZE, lv_read_bytes, "Reading back data from file (count)" )
```
kisvegabor, line 53: TODO

tests/micropython/run_testcase.py
```diff
@@ -0,0 +1,121 @@
+#! ../../../../../../lv_micropython/ports/unix/build-standard/micropython
```
PGNetHun, line 1: I think we could delete this line, the path will be different for every developer

tests/micropython/run_all_tests.py
```diff
@@ -0,0 +1,41 @@
...
+import sys
+import os
+import glob
+
+sys.dont_write_bytecode = True
+import test_constants as lv_const
+
+
+class Var:
+    MICROPYTHON_COMMAND = "./"  #"../../../../../../lv_micropython/ports/unix/build-standard/micropython "
```
PGNetHun, line 13: The path to MicroPython (unix) executable should come from environment variable or as an argument of this script. But not hard-coded path.
Please check [argparse Python module](https://docs.python.org/3/library/argparse.html) for easy usage of arguments.

Example:
```
import argparse

parser = argparse.ArgumentParser(prog="run_all_tests", description="Run all MicroPython tests")
parser.add_argument("-m", "--micropython", help="Path to MicroPython unix executable")
args = parser.parse_args()

# Access argument value:
print("Path: ", args.micropython)
```

tests/micropython/test_cases/test_animation.py
```diff
@@ -0,0 +1,54 @@
+#! ../../../../../../../lv_micropython/ports/unix/build-standard/micropython ../run_testcase.py
```
PGNetHun, line 1: Can you please shorten this path to not include `../lv_micropython/`?
The correct relative path is `../../../../../ports/unix/build-standard/micropython` 
Same for other python files.

PGNetHun (commented):
The idea is very good, and nice work.

PGNetHun:
Discussion related to this PR: https://github.com/lvgl/lvgl/issues/4347#issuecomment-1744308286

tests/micropython/run_testcase.py
kisvegabor: As LVGL is a submodule in `lv_binding_micropython` it has a fixed place and this path shouldn't really change. Or am I missing something?

PGNetHun: If we run the tests from `lv_micropython`, then the path can be relative, but 1 level shorter:
`../../../../../ports/unix/build-standard/micropython`     (without `../lv_micropython/` )

PGNetHun:
@kisvegabor 
Is it decided, that binding tests should be stored in `lvgl` repository, and not in the binding repository (in our case `lv_micropython`)?

tests/micropython/run_all_tests.py
PGNetHun: Using default value:
```
parser.add_argument(
  "-m", 
  "--micropython", 
  default="../../../../../ports/unix/build-standard/micropython",
  help="Path to MicroPython unix executable"
)
```

tests/micropython/run_testcase.py
kisvegabor: True!

tests/micropython/run_all_tests.py
kisvegabor: If we run the test from `lv_micropython`, I think the same argument applies for [here](https://github.com/lvgl/lvgl/pull/4645#discussion_r1355806689).

kisvegabor:
> Is it decided, that binding tests should be stored in lvgl repository, and not in the binding repository (in our case lv_micropython)?

Good question! What is important for me is to limit to what extent we maintain the bindings during the development of the core LVGL. We should deal only with a few core tests, which can be easily updated by anyone when LVGL changes. As these tests needs to in sync with both LVGL and the bindings none of the places is ideal.

I still vote for having them in LVGL because this way if we rename or change something in a PR, we can easily update all the tests as well. If we had the tests in multiple repos, updating the tests on ~5 places would be quite inconvenient.

kisvegabor:
I'll take care of the other suggestion early next week. 

By the way, couldn't make the  [LVGL File system tests](https://github.com/lvgl/lvgl/pull/4645/files#diff-6a782b12609b64eceffd659c9143b1a2d277d36ace673f0100ab09685f36d8a9) work due to parameter type mismatch. Could you check it out and point us into the right direction?

kdschlosser:
I have a question on this. Where are the tests running? is it a script that you are running when launching MicroPython????

If that is what is being done I believe there is a better way to go about doing it. When launching MicroPython the repl takes over stdin. What can be done is you use CPython as the testing platform and it runs MicroPython in a subprocess and the CPython scrips passes code over stdin for the subprocess and it monitors stdout for the subprocess for errors. This would allow the testing to continue to run even if there is an error with one of the tests. So instead of having to run the tests, fix a big, run the tests, fix another bug, run the test etc... it would be run the tests and get a list of all of the things that are not working properly and go and fix them. 


Doing that would also allow you to use the unittest framework to handle the testing. As a bonus you can also run multiple instances of MicroPython and the tests can run in parallel which will speed things up.. To check for API naming changes that is done in a single test with doing a recursive iteration over lvgl.\_\_dict\_\_ and the objects contained in it to pull the names of all of the different bits and pieces. Get the JSON output from gen_mpy up to snuff so it will output all of the names and types and that can be used to make sure that functions and enumerations are correct and where they are supposed to be and it will also check values of enumerations to make sure they are correct as well.

Only if the API was the same... That would make things a whole lot easier to deal with... :thinking:

kdschlosser:
on another note. You should consonantly be monitoring the amount of used and available memory to see if there is a memory leak at all. This is something that can happen very easily and due to the limited memory that is available on an MCU every single byte that is not properly garbage collected could cause an issue. It might be best to limit the amount of available RAM when the binding is compiled. This way it would be more like an MCU environment. You could see how bad memory fragmentation gets when running the tests.

kdschlosser:
Here is some test code that allows for injecting lines of code directly into the repl of MicroPython. 

This would allow you to run one test at a time and see what the results are for each test instead of having a bunch of tests in a single source file that micropython runs using exit codes as the way to identify that a failure has occurred. This also allows you to capture the actual traceback information so you would know exactly what is happening when a failure occurs. There are simply too many things that can go wrong and trying to cover every single error that can occur for each test to send an appropriate exit code would make it a bear to write and also to maintain. Going this route eliminates the need to do that. check the output for `"Traceback"` and if it is there then the output from MicroPython can be sent to stderr in CPython.

You can also transfer the frame buffer data using this mechanism as well. You would need to base64 encode the frame buffer data and the decode it in CPython, not a big deal to do. 

What I would do is have a folder named, micropy_tests and in that folder you would have a file for each test that is to be run that is written to run in MicroPython. file extension would be .py one for the purposes of making it easier when writing the code in an IDE. When a test case runs the name of the test case would align with the file so a single function would handle reading the source file and sending it to micropython.


Another benefit.... Running tests in parallel to each other. You can launch more than one instance of a micropython subprocess. you would have a single thread made to handle each of those subprocesses and the main would be responsible for sending data to each thread and collecting the results from each thread. You could test every piece of the binding and not have it take forever to do it. 

```python
import subprocess
import sys
import os


ctrl_c = b'\x03'  # 2 times to exit any running code
ctrl_d = b'\x04'  # exit paste mode committing what has been pasted
ctrl_e = b'\x05'  # enter paste mode

paste_mode_prompt = b'==='
repl_prompt = b'>>>'

code = b'''\
for i in range(1000):
    print(i)
'''

os.environ['MICROPYINSPECT'] = '1'

p = subprocess.Popen(
    [os.path.expandvars('$HOME/lv_micropython/ports/unix/build-standard/micropython')],
    stdout=subprocess.PIPE,
    stderr=subprocess.PIPE,
    stdin=subprocess.PIPE,
    env=os.environ
)

entered_repl = False
entered_paste_mode = False
exit_count = 0
output = b''
while p.poll() is None:
    char = p.stdout.read(1)
    if char != b'':
        output += char

        if b'999' in output:
            sys.stdout.write(output.decode('utf-8'))
            sys.stdout.flush()
            p.stdin.close()

        elif repl_prompt in output:
            sys.stdout.write(output.decode('utf-8'))
            sys.stdout.flush()
            output = output.split(repl_prompt, 1)[-1]

            if not entered_repl:
                p.stdin.write(ctrl_e)
                p.stdin.flush()
                entered_repl = True

        elif paste_mode_prompt in output:
            sys.stdout.write(output.decode('utf-8'))
            sys.stdout.flush()
            output = output.split(paste_mode_prompt, 1)[-1]
            if not entered_paste_mode:
                p.stdin.write(code)
                p.stdin.write(ctrl_d)
                p.stdin.flush()
                entered_paste_mode = True

if not p.stdout.closed:
    p.stdout.close()

if not p.stderr.closed:
    p.stderr.close()

sys.stdout.flush()
```

kdschlosser:
The above is just a suggestion, I think it would make writing tests and also maintaining them easier to do.

kdschlosser:
so here is a more involved example I made.

[tests.zip](https://github.com/lvgl/lvgl/files/13197009/tests.zip)

All that would need to be done is a new folder gets made under the micropy_tests folder. That new folder's name would be the name you want to give for the test. no special characters in the name. the name has to be something that can be used as a variable name. This is what is called the test case. a test case is a group of tests that you want to run. so if you are testing a widget then give it the same name as the widget name. Inside that folder you need to make 2 files for each thing you want to test. so say you are doing the slider and you want to test setting the knob color, then you would name the test files "knob_color_micropy.py" and "knob_color_results.py". I supplied an example so you can see what needs to be in those 2 files. The file ending in "_micropy.py" is the script that gets sent to MicroPython and the results is well. the result data if any. If there are no results you are passing back you can either leave the file empty or don't make a results file at all.


That's about as simple as you can make it for creating tests. It has a small code footprint to make the test system work properly and it will also keep the amount of code needed for each test to as small as possible.

kisvegabor:
I would like to keep it as simple as possible for 2 reasons:
- Make the test simple to write for binding maintainers
- Make the tests simple to maintain. If someone changes LVGL, it should be straight forward to see what to update in a binding test without setting the up the whole environment.

By having only a few simple tests running them in parallel to save time, or a whole unit test framework is overkill and we can really keep things very simple. So the goal is that any random C developer/contributor should see and know how it works and what is happening. 

Whether to have a screenshot tests is big decision. The issues I see here:
- We need to set up a screenshot compare test engine
- We need to have more complex tests
- We need to maintain the screenshots (the C screenshot compare test work like [this](https://github.com/lvgl/lvgl/tree/master/tests#asserts))


And these test should run on every LVGL commit and PR to see that LVGL is still compatible with the binding. I see that e.g. MicroPython is very complex, but I think we can safely assume that if e.g. a few functions `lv_label` is working the other widgets are working as well. What we need to tests is that whether we can pass the required data types to LVGL and get them back. E.g. label text, line point array, chart series coordinates, etc.

kdschlosser:
> We need to set up a screenshot compare test engine

easy to do

> We need to have more complex tests

using the framework I have designed above makes that easy to do as well

> We need to maintain the screenshots (the C screenshot compare test work like [this](https://github.com/lvgl/lvgl/tree/master/tests#asserts))

since there is already screen shots being used in the C tests those same screen shots can be used. no reason to have to maintain 2 sets of screen shots.


The reason why the image comparison engine is easy to do is all you need to do is set up a frame buffer callback that takes the buffer data b64 encodes it and dimps it to sdtout in micropython. the CPython script would then read the encoded buffer data, decode it and do a very simple comparison .

```python
if 'buffer_data' in results:
    buffer_data = base64.decode(results['buffer_data'])
    self.assertEqual(buffer_data, image_data, 'data is not the same')
```

That is really about the extent of the engine.  super easy to do.

I do not believe there would be anything easier to use that the framework I have set up. No need to have to know a list of special exit codes or to ensure all errors that can happen are covered in every single test. Code it like you would run it normally without any error testing or checking and if there is an error it would be captured by the CPython script which is what is handling the tests.


The parallel thing is something that can optionally be done. It is not something that has to be done right away. In the example I gave for a framework it doesn't run in parallel. But it can be made to if the testing process starts to take a long time to run. 

Here is a cool thing. You already have a bucket load of code already written for MicroPython. All the examples can be run as tests. Not only does this ensure that the binding is running properly it also ensures that all of the examples are coded properly for any API changes made. You don't have to change a single thing in any of the examples to be able to test them either. That's the best part about it. The modules that get imported in the examples can be dynamically created in code at the beginning of the run of example tests. This would take care of what needs to be done in order to provide everything that is needed for the tests to run. 

The examples would cover all of the operating API kinds of things and then you would have to check returned values. 3/4rs of the work is pretty much already done. maybe 2 hours to code up the dynamic module code to handle the imports that are specific to the simulator. nothing that is exceedingly hard to do. So an example test runs and the next test after that would be to check input and output of different data types for that example. It would then move onto the next example which would be a new test case so the test would be in a new micropython subprocess so a clean sandbox environment. The last test case would have it's micropython session closed.

All of the above is done without needing to make any coding changes to any of the examples to make it work..

Think it could be made any easier than that?

kdschlosser:
as far as the screen shot engine. PIL is an already well known Python Image Library (PIL) and it has the ability to take care of creating the output png file so it can be saved as an artifact and inspected. PIL is written mostly in C so it would be pretty fast at taking care of the error image creation. PIL is not available in MicroPython bit it is available in CPython and by doing the base64 encoding of the buffer data to export the data from MicroPython to CPython it makes PIL available to use.

kisvegabor:
Okay, let's discuss the screenshot compare way. Decoding the PNG reference images can be challenging in the binding, but for the binding related tests we can have som raw bin images too. 

It's important to have only a few number of tests for the binding becasue
- We need the be fair with all bindings by having clear requirements to the acknowledged as an official binding. We can't ask for 100 of tests from everyone. 
- If something changes in LVGL we (I or any contributor) cannot maintain 100 tests/examples for 5 bindings.
- The tests should be very simple, targeted and not redundant to help finding issues.

So with screenshot compare we can have these very simple targeted examples to test various parts of LVGL:
1. Create some basic widgets and set their properties (they test passing various data types too).  Set the coordinates manually
2. Create a lot of list elements, style them and use a layout to arrange them
3. Create +/- buttons and add an events to increment/decrement a label
4. Load a PNG image and a font + add button and animate them on click
5. [Observer](https://docs.lvgl.io/master/others/observer.html) with slider + label

I can create the C version of these tests/examples. Form this point one of the criteria for bindings would be to make the screenshot coompare tests pass. These can serve as code examples in the [README](https://github.com/lvgl/lvgl#robot-examples) too. 

What do you think?

kdschlosser:
whatever you would like to have done as far as the tests go. If you can write the tests in C I can port them to MicroPython fairly quickly. 

The reason why I had mentioned the examples is because they already have to be maintained and there is a snot load of them. They can either all be used or only some of them that doesn't matter. It reduces the amount of maintenance work that would need to be done if those examples are able to be used, which is what the ultimate goal is I had thought.

It's a simple way to reduce work. 

The peasant doesn't decide what gets planted in the field. LOL..  It is your decision and I will do what you want done, I will toss ideas out there on things and if they aren't used I am not one of those people that gets their panties in a twist over it.

I tossed out the CPython idea because it is a way to make the tests easier to write and maintain. There is a whole lot less boiler plate code that needs to be used and would have to be repeated in each test if the tests are entirely contained in MicroPython. 

It is good to check the buffer output in bindings separately because of things like #4716. There are things in bindings that can cause visual issues that only show up in the binding. Keep that in mind.

kisvegabor:
The big question is who maintains the binding examples. If we would like to have several stable and officially recommended bindings we cannot maintain them in the core LVGL as it would be too much work to update the hundreds of examples. Instead the bindings should have their own maintainers. If we don't have enough people in binding to maintaining 100 examples probably there should be only a 10-20 carefully selected examples. 

Thank you for help with this. I'll create these examples next week.

kdschlosser:
Now you see where I am coming from. There are already a large number of examples for both the micropython binding as well as the  LVGL itself. would be nice to use them in some manner for the testing as well if at all possible. This way it is less code to worry about maintaining.

kisvegabor:
> This way it is less code to worry about maintaining.

The questions is who worries about what. At main LVGL repo we should focus on LVGL, and the binding repos should focus on bindings. So these too needs to be separated. However this way, we can't relay on binding examples as much as we are doing now. 


I'll add the C examples next week.

kisvegabor:
@kdschlosser 
Finally I added a single complex test in 3bb649db7

I think if a binding is capable of reproducing this UI using the same features (e.g. layouts) than we can say it's working.

What do you think?

kdschlosser:
Remember in the tests there is no limit to the display size so you can literally go all out.

Is there a way to programmatically open thing like dropdowns and programmatically press the button. in a manner that would replicate an indev driver. 


OK well that is just a dumb question and I answered it myself when I typed indev. It can be done using the indev callback.

kisvegabor:
We can click the widgets in the tests. See for example [here](https://github.com/lvgl/lvgl/blob/master/tests/src/test_cases/widgets/test_dropdown.c#L143-L151).

However I was trying to avoid it to save the hassle of integrating a fake indev driver in each binding tests. However probably it's not a big deal. 

Keep in mind that we are looking for the simplest yet suitable solution now in order to have as simple binding test cases as possible to make their maintenance simple as well. 

So I wonder what is added value of clicking on a lot of widgets. If we can click one (e.g. a button) we can assume that the other widgets can be clicked as well.

kdschlosser:
I get ya on keeping it simple for the time being. But i wouldn't be me if I didn't also think about what might be done in the future and making sure that whatever the design is the work done now would not need to be redone. There is nothing I hate more than having to do something over because of not planning for the future... Because the tests are going to run on something that is not going to actually be visible there is no size constraint on how large the "display" is. For performance purposes I would not go overboard but there is no reason why you cannot use a 1920 x 1080 frame buffer size and render all of the widgets to it at a single time. Good way to test not only the widgets but also check for performance related issues as well...

kisvegabor:
If we need more tests we can just create `test_bindings_2.c` and have an other reference images too. So I feel like it's quite salable and future proof.

1920 x 1080 is too large as on a full HD monitor due to the start menu and window border some parts of the window will be clipped.

kdschlosser:
> 1920 x 1080 is too large as on a full HD monitor due to the start menu and window border some parts of the window will be clipped.


Good point. didn't think about that. 1600 x 900??

kdschlosser:
As an idea there should be a way to inject indev input into a single widget instead of having to use absolute coordinates from the screen. So for sat the slider there would be a function to move the slider as if it was touched from point a to point be without a user or a test program having to know coordinates.


In other UI frameworks I have used the "window" is the only object that receives any kind of input event. Those events are then issued  to any active objects or objects that the event occurred in coordinate wise, like a click event. Because of this kind of a setup events can be programmatically created and sent into a widget so the widget would respond accordingly as if it was actual input. Does LVGL work this way?

kisvegabor:
> Good point. didn't think about that. 1600 x 900??

All tests are using 800x480 now, so I wouldn't change it without a really good reason. 

> As an idea there should be a way to inject indev input into a single widget instead of having to use absolute coordinates from the screen.

It's easy, we can always get get the coordinates of the widget and click on it. 

---

I've just realized that an important thing is missing from the current test: deleting widgets. I'll add these:
- Start a looping animation on a widget and delete the widget
- Keep pressing a widget with an input device, delete  the widget, and release the input device
- Delete all widget from the screen

It will add 3 more screen shot images.

kisvegabor:
@kdschlosser We are getting close to end of v9's development cycle and I would like to have this test in place for MicroPython before v9 is released.

My binding test is in place (I will add something for delete test) but I wonder if you can help by adding the MicroPython version of this test.

kdschlosser:
yeah I can do that I just need to know what tests you want to run on MicroPython

kisvegabor:
Amazing!

Finally it can be very simple:
1. Run the MPY version of [`test_binding.c`](https://github.com/lvgl/lvgl/blob/master/tests/src/test_cases/test_bindings.c)
2. Compare the [screenshot](https://github.com/lvgl/lvgl/blob/master/tests/ref_imgs/binding.png)

kdschlosser:
not so fast. The point to the tests is to test the ability of the code generator. if you make an MPY if the test using the code generator that would be self defeating as it is using the thing to generate the test of what is supposed to be tested.

The code should be manually written to make sure it is going to work as it should. It should also be written in Python and not in C code. If it is written in C code the test is not going to be using the python API but instead will be using the C API. and hat is needing to be tested is the Python API.

kisvegabor:
Ah, I also meant writing the test in Micropython too. Can you do it shall I give it a try?

kdschlosser:
I can do it if you tell me what tests need to be done.

kisvegabor:
To be more precise:
1. Write test_binding.c in MicroPython
2. Do a [screenshot](https://github.com/lvgl/lvgl/blob/master/tests/ref_imgs/binding.png) compare. We might need the a plain bin version of the screenshot for easier comparison. 
3. Update [this CI](https://github.com/lvgl/lvgl/blob/master/.github/workflows/build_micropython.yml#L72-L74) to run the `test_binding.py` + the screenshot compare

kisvegabor:
Implemented in #4997

PR #4655, opened 2023-10-14
feat(obj): rename lv_obj_set_flag to lv_obj_update_flag

XuNeo:

~~Add API to update/get obj flags.
`Update`  means either add or remove flags depending on `LV_OBJ_FLAG_ACTION_REMOVE` is cleared or not.~~

Rename `lv_obj_set_flag()` to `lv_obj_update_flag` since `lv_obj_set_flag(obj, f, false)` is actually clear flag but not `set` which is confusing. 

### Checkpoints
 
### DIFF
--- a/src/core/lv_obj.c
+++ b/src/core/lv_obj.c
@@ -152,8 +152,7 @@ void lv_obj_remove_flag(lv_obj_t * obj, lv_obj_flag_t f)
 
 }
 
-
-void lv_obj_set_flag(lv_obj_t * obj, lv_obj_flag_t f, bool v)
+void lv_obj_update_flag(lv_obj_t * obj, lv_obj_flag_t f, bool v)
 {
     if(v) lv_obj_add_flag(obj, f);
     else lv_obj_remove_flag(obj, f);
--- a/src/core/lv_obj.h
+++ b/src/core/lv_obj.h
@@ -223,7 +223,7 @@ lv_obj_t * lv_obj_create(lv_obj_t * parent);
 void lv_obj_add_flag(lv_obj_t * obj, lv_obj_flag_t f);
 
 /**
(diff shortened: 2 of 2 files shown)


kisvegabor:
In the set/get API will it be intuitive enough? After thinking about it, having the flags are dedicated properties feels better to me.

> Remove lv_obj_set_flag() since lv_obj_set_flag(obj, f, false) is actually clear flag but not set which is confusing.

True, the naming is really not so good. What about `lv_obj_update_flag(obj, flag, true/false)`?

XuNeo:
> After thinking about it, having the flags are dedicated properties feels better to me.
> 

Do you mean add property ID for every flag or only one `FLAG` property, with flags to be changed in `property->value`?
I'm using the latter one but it requies bit31 to mark add or remove flags. Is it acceptable?

kisvegabor:
I'ó mean having dedicated properties for each flag. It feel more "flat" and simpler to me.

kisvegabor:
Do you think keeping it as `lv_obj_update_flag()` is not a good idea?

XuNeo:
Hi Gabor,
I haven't seen usage of `lv_obj_update_flag `, I thought it's for `obj property`. 
Existing `lv_obj_add/remove_flag` is more friendly to use.  Do you see potential usage of `lv_obj_update_flag`?

kisvegabor:
Oh, actually I added it as a helper function because I found that in a lot of cases I'm having 
```
if(something) add_flag
else clear_flag
```

I do this a lot during UI development, so I guess it would be useful to have a helper.

XuNeo:
That makes sense. I added them back now. 
Thank you for the explanation.  :)

kisvegabor:
Looks good, merging!

PR #4659, opened 2023-10-15
fix(dev): drm driver awaits vsync before drm_wait_vsync

niklasf:
Typical timings before:

    main loop
        other work                   10ms
        drm_flush
            drm_wait_vsync            0ms (drm_dmabuf_set_plane already waited)
            drm_dmabuf_set_plane     15ms (blocking wait for vsync!)

Typical timings after:

    main loop
        other work                   10ms
        drm_flush
            drm_wait_vsync            5ms (15ms - time spent on other work)
            drm_dmabuf_set_plane      0ms (now non-blocking)

PR #4661, opened 2023-10-15
perf(dev): use triple buffering for linux drm

niklasf:
Based on #4659. Flushing into a third buffer allows us to more efficiently use the time that is spent waiting for the previous commit/vsync.

`lv_demo_benchmark(LV_DEMO_BENCHMARK_MODE_REAL)` on Raspberry Pi 1: 10 -> 17 weighted fps.

### DIFF
--- a/src/dev/display/drm/lv_linux_drm.c
+++ b/src/dev/display/drm/lv_linux_drm.c
@@ -65,9 +65,10 @@ typedef struct {
     drmModePropertyPtr plane_props[128];
     drmModePropertyPtr crtc_props[128];
     drmModePropertyPtr conn_props[128];
-    drm_buffer_t drm_bufs[2]; /* DUMB buffers */
-    uint8_t active_drm_buf_idx; /* double buffering handling */
-    bool inactive_drm_buf_dirty;
+    drm_buffer_t drm_bufs[2]; /*DUMB buffers*/
+    uint8_t active_drm_buf_idx; /*Double buffering handling*/
+    uint8_t * intermediate_buffer;
+    unsigned long int intermediate_buffer_size;
 } drm_dev_t;
 
 /**********************
@@ -155,10 +156,8 @@ void lv_linux_drm_set_file(lv_display_t * disp, const char * file, int64_t conne
     lv_coord_t ver_res = drm_dev->height;
     lv_coord_t width = drm_dev->mmWidth;
 
-    uint32_t draw_buf_size = hor_res * ver_res / 4; /*1/4 screen sized buffer has the same performance */
-    lv_color_t * draw_buf = malloc(draw_buf_size * sizeof(lv_color_t));
-    LV_ASSERT_MALLOC(draw_buf);
-    lv_display_set_draw_buffers(disp, draw_buf, NULL, draw_buf_size, LV_DISPLAY_RENDER_MODE_PARTIAL);
+    lv_display_set_draw_buffers(disp, drm_dev->intermediate_buffer, NULL, drm_dev->intermediate_buffer_size,
+                                LV_DISPLAY_RENDER_MODE_DIRECT);
     lv_display_set_resolution(disp, hor_res, ver_res);
 
     if(width) {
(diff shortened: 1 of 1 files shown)


kisvegabor:
Maybe it wasn't appropriate, but I merged the other PR first. Could you take a look and fix the conflicts?

I'm thinking if we can eliminate the `memcpy` of the rendered image by using `LV_DISPLAY_RENDER_MODE_DIRECT` instead of `LV_DISPLAY_RENDER_MODE_PARTIAL` can be an option. See here: https://docs.lvgl.io/master/porting/disp.html#draw-buffers

niklasf:
No worries, that was the intent, because the other change is much more clear-cut.

Direct rendering into the third buffer could be interesting. Will try and report back.

Direct rendering into the background buffer doesn't work well, because then we'd have to block waiting for vsync to ensure it really is no longer in the foreground.

kisvegabor:
I mean `DIRECT_MODE` with 2 buffers. This is what happens:
1. LVGL renders into `buf1` and sends it for flushing
2. You set it's address on VSYNC in the background
3. E.g. 30 ms later before LVGL is about to render the next frame it will wait for flushing to be finished and the buffer will be synchronized. See [here](https://github.com/lvgl/lvgl/blob/master/src/core/lv_refr.c#L459).
4. LVGL renders into `buf2` and sends it for flushing 
5. And so on.

The difference compared to the current `MODE_PARTIAL` is that 
- We needs only 2 buffers
- There is no copy at all , or just a minimal during syncing the buffers. This way the rendered image can be used directly. With screen sized updates it could really make a difference.

niklasf:
Looks like that should work great. Hitting two issues with it:
* In order to receive the page flip event there needs to be a thread polling the file descriptor for it. Ideally it would be the same instead of the busy wait for `!flushing`. Could there be a way to hook into it?
* That aside, I am seeing something strange when trying this on my old Raspberry. Copying *from* the mmapped drm buffer appears to be very slow, an order of magnitude slower than writing, so that syncing a bit from the foreground buffer is slower than copying the entire buffer from RAM. If it really does depend on hardware, it might have to be configurable :(

kisvegabor:
> In order to receive the page flip event there needs to be a thread polling the file descriptor for it. Ideally it would be the same thread instead of the busy wait for !flushing. Could there be a way to hook into it?

How would you poll the file descriptor? I mean exactly what kind of code do use for that? A simple file `read` which yields the thread while the file is empty?

niklasf:
> A simple file read which yields the thread while the file is empty?

Internally yes. Basically exactly what [`drm_wait_vsync()`](
https://github.com/lvgl/lvgl/blob/4a7f31017e0f891e43cd3ecf2d543a54aa51f8f8/src/dev/display/drm/lv_linux_drm.c#L792) does, i.e., `poll` on a file descriptor that yields the thread followed by `drmHandleEvent` which is then guarnateed to be non-blocking.

kisvegabor:
Hmm, so it seems we should have a callback or some other mechanism to let the user use its own wait loop. In an RTOS scenario it should a wait for mutex or something like that.  

What I see so far:
- It's a per display configuration and not a global one
- It's tightly related to how the `flush_cb` is implemented

So it seems having a callback similar to `flush_cb`, e.g. `flush_wait_cb` would do the job. 

What do you think?

niklasf:
I agree it makes sense.

Maybe we can still have this patch as an intermediate step? It could be backported to v8, while a new `flush_wait_cb` probably can't. Updated it to use direct rendering at least for the third buffer.

kisvegabor:
Sounds good to me. So please confirm that it's ready to merge, and I'll add the `flush_wait_cb` hopefully next week.

kisvegabor:
Thank you as always :+1:

niklasf:
> and I'll add the flush_wait_cb hopefully next week

Cool. Please `@` me on the pull request. Happy to test or implement the drm driver integration.

kisvegabor:
Sure! Thank you in advance! 

In the weekend I updated the structure of the docs and now we have place for [drivers and libraries](https://docs.lvgl.io/master/integration/index.html). 

There isn't many thing there yet. I just copied the exsisting related docs to their new place. So I'm still not sure what would be the best structure to document  drm, evdev and similar libs/drivers. 

If you have liking to help establishing the structure of the docs and add a page for evdev and drm, please open a new issue to brain storm about it.

PR #4664, opened 2023-10-17
fix(style): make img recolor inherited

BlackHawkC7:


kisvegabor:
Hi,

I'm thinking about it as well, but we can't do it v8 as it might break some people's existing UIs.

Could you add the same change for the `master` branch? Please also update [this script](https://github.com/lvgl/lvgl/blob/master/scripts/style_api_gen.py#L241) and run it as well. (Just set `Inherited: 1`)

BlackHawkC7:
Thanks for the info. I will try it.

Sorry for the empty description. I was using this feature in my project with version 7. I have recently migrated to version 8 and then I was struggling with fixing my buttons until I figured out that inheritance was the problem.

kisvegabor:
In this case I suggest keeping this modification local, as we can't add these changes to v8. :slightly_frowning_face:

PR #4680, opened 2023-10-20
feat(timer): support lv_timer_handler_set_resume_cb

bjsylvia:

This feature allows events to be customized when the timer handler is awakened.

### Checkpoints
 
### DIFF
--- a/docs/overview/timer.rst
+++ b/docs/overview/timer.rst
@@ -69,6 +69,18 @@ You can make a timer repeat only a given number of times with
 automatically be deleted after it's called the defined number of times.
 Set the count to ``-1`` to repeat indefinitely.
 
+Enable and Disable
+******************
+
+You can enable or disable a timer with :cpp:expr:`lv_timer_enable(en)`.
+
+Pause and Resume
+****************
+
+:cpp:expr:`lv_timer_pause(timer)` pauses the specified timer.
+
+:cpp:expr:`lv_timer_resume(timer)` resumes the specified timer.
+
 Measure idle time
 *****************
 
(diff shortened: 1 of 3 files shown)


kisvegabor:
What is a real life use case for it?

bjsylvia:
> What is a real life use case for it?

In the nuttx porting layer based on the implementation of libuv, when we bind the `lv_timer_handler`  to a custom `uv_timer`, the `uv_timer` needs to be awakened simultaneously when the `lv_timer_handler`  switches from stop to resume state. This enhancement has been applied in the nuttx porting layer.

bjsylvia:
you can find use case in `lv_nuttx_uv_timer_init` which implements in https://github.com/lvgl/lvgl/pull/4683.

kisvegabor:
Got it now, however it feels a little bit too specific to me. If we have a "resume" why don't we have callbacks for other changes like "pause", "period change", "reset", "delete", etc. 

Maybe we could add events to timers as well? Similarly to widgets and  displays. It might be overkill, but we should find a more generic solution as the current suggestion solves only a specific problem.

bjsylvia:
> Got it now, however it feels a little bit too specific to me. If we have a "resume" why don't we have callbacks for other changes like "pause", "period change", "reset", "delete", etc.
> 
> Maybe we could add events to timers as well? Similarly to widgets and displays. It might be overkill, but we should find a more generic solution as the current suggestion solves only a specific problem.

Let me describe our actual needs: for example, when using libuv as the main event loop of the application, we bind `lv_timer_handler` to a certain `uv_timer`. For performance or other reasons, `uv_timer` calls will be stopped when there is no active LVGL timer present. So when do we need to resume this `uv_timer`? It is when `lv_timer_handler_resume` is called. We don't need to pay attention to the different state transitions of a specific timer, which is why it is named `lv_timer_handler_set_resume_cb` instead of `lv_timer_set_resume_cb`.

This is a situation that will be encountered when binding LVGL with various event loop mechanisms already adopted by the system. 

What do you think?

kisvegabor:
Now I understand it better, thank you! 

And I guess you know when to stop the `uv_timer` from the return value `lv_timer_handler()`.

Okay, I have no better and simpler idea either. Just please add 1-2 sentences about it and its purpose in the docs.

bjsylvia:
> And I guess you know when to stop the `uv_timer` from the return value `lv_timer_handler()`.

Yes, I see.

> 
> Okay, I have no better and simpler idea either. Just please add 1-2 sentences about it and its purpose in the docs.

I have added some description in `docs/overview/timer.rst`. Please review it.

kisvegabor:
Merged, thank you!

PR #4528, opened 2023-09-07
feat(draw): add basic vector APIs

PeterBee97:
### Description of the feature or fix
Vector APIs and basic draw task distribution following #4388 . Some matrix helper functions are provided.
### Checkpoints
- [x] Run `code-format.py` from the scripts folder. [astyle](http://astyle.sourceforge.net/install.html) needs to be installed.
- [ ] Update the [Documentation](https://github.com/lvgl/lvgl/tree/master/docs) if needed
- [ ] Add [Examples](https://github.com/lvgl/lvgl/tree/master/examples) if relevant. 
- [ ] Add [Tests](https://github.com/lvgl/lvgl/blob/master/tests/README.md) if applicable.
- [ ] If you added new options to `lv_conf_template.h` run [lv_conf_internal_gen.py](https://github.com/lvgl/lvgl/blob/release/v8.3/scripts/lv_conf_internal_gen.py) and update [Kconfig](https://github.com/lvgl/lvgl/blob/release/v8.3/Kconfig).
 
Be sure the following conventions are followed:
- [ ] Follow the [Styling guide](https://github.com/lvgl/lvgl/blob/master/docs/CODING_STYLE.md)
- [ ] Prefer `enum`s instead of macros. If inevitable to use `define`s export them with `LV_EXPORT_CONST_INT(defined_value)` right after the `define`.
- [ ] In function arguments prefer `type name[]` declaration for array parameters instead of `type * name`
- [ ] Use typed pointers instead of `void *` pointers
- [ ] Do not `malloc` into a static or global variables. Instead declare the variable in `lv_global_t` structure in [`lv_global.h`](https://github.com/lvgl/lvgl/blob/master/src/core/lv_global.h) and mark the variable with `(LV_GLOBAL_DEFAULT()->variable)` when it's used. See a detailed description [here](https://docs.lvgl.io/master/get-started/bindings/micropython.html#memory-management).
- [ ] Widget constructor must follow the `lv_<widget_name>_create(lv_obj_t * parent)` pattern.
- [ ] Widget members function must start with `lv_<modul_name>` and should receive `lv_obj_t *` as first argument which is a pointer to widget object itself.  
- [ ] `struct`s should be used via an API and not modified directly via their elements.
- [ ] `struct` APIs should follow the widgets' conventions. That is to receive a pointer to the `struct` as the first argument, and the prefix of the `struct` name should be used as the prefix of the function name too (e.g.  `lv_disp_set_default(lv_disp_t * disp)`)
[...]

PR #4691, opened 2023-10-23
feat(draw): add implements vector graphic APIs (#4528)

onecoolx:

### Checkpoints
 

kisvegabor:
Awesome! :tada: :tada: :tada: 

Just 2 comments: 
1. The demo looks more like ~3 examples. So could you convert it into examples? Maybe add to the other canvas examples. 
2. Could you give me some instructions about how to try it out with ThorVG on Linux?

onecoolx:
> Awesome! 🎉 🎉 🎉
> 
> Just 2 comments:
> 
> 1. The demo looks more like ~3 examples. So could you convert it into examples? Maybe add to the other canvas examples.
> 2. Could you give me some instructions about how to try it out with ThorVG on Linux?

You can download the latest version of ThorVG code, and then use the following command to compile and install in the code directory：
```
mkdir builddir
meson --prefix /usr builddir -Dbindings=capi
meson compile -C builddir
sudo meson install -C builddir
```
use the following patche in the **lv_port_pc_eclipse** project:

```
diff --git a/Makefile b/Makefile
index cac463c..7763e07 100644
--- a/Makefile
+++ b/Makefile
@@ -14,8 +14,8 @@ WARNINGS ?= -Wall -Werror=float-conversion -Wextra \
 						-Wempty-body -Wshift-negative-value -Wstack-usage=2048 \
             -Wtype-limits -Wsizeof-pointer-memaccess -Wpointer-arith
 
-CFLAGS ?= -O3 -I$(LVGL_DIR)/ $(WARNINGS) `pkg-config --cflags wayland-client` `pkg-config --cflags xkbcommon`
-LDFLAGS ?= -lSDL2 -lm `pkg-config --libs wayland-client` `pkg-config --libs xkbcommon` -lpthread
+CFLAGS ?= -O3 -I$(LVGL_DIR)/ $(WARNINGS) `pkg-config --cflags wayland-client` `pkg-config --cflags xkbcommon` `pkg-config --cflags thorvg`
+LDFLAGS ?= -lSDL2 -lm `pkg-config --libs wayland-client` `pkg-config --libs xkbcommon` `pkg-config --libs thorvg` -lpthread
 BIN = demo
 
 
diff --git a/lv_conf.h b/lv_conf.h
index 5e2152c..37b6868 100644
--- a/lv_conf.h
+++ b/lv_conf.h
@@ -829,6 +829,10 @@
 
 /*Demonstrate scroll settings*/
 #define LV_USE_DEMO_SCROLL          1
+
+#define LV_USE_THORVG 1
+#define LV_USE_VECTOR_GRAPHIC   1
+#define LV_USE_DEMO_VECTOR_GRAPHIC  1
 /*--END OF LV_CONF_H--*/
 
 #endif /*LV_CONF_H*/
diff --git a/main.c b/main.c
index 8f47c55..a5305dc 100644
--- a/main.c
+++ b/main.c
@@ -72,7 +72,8 @@ int main(int argc, char **argv)
 
   //lv_example_chart_3();
 
-  lv_demo_widgets();
+ // lv_demo_widgets();
+    lv_demo_vector_graphic();
 
   while(1) {
       /* Periodically call the lv_task handler.

```

kisvegabor:
:drum: :drum: :drum: 
Aaaand the magic happened! 

image


It's really nice! 

There is only one thing which is bothering me: Do you think it's possible to compile ThorVG with Eclipse as a simple managed build? It would be great to add a very simple version of ThorVG directly to LVGL. We can keep only the very core functionality and the SVG loader, and make it possible to just enable a config to use it. Of course if someone wants to use the full features ThorVG he can use it as well.

What do you think?

onecoolx:
> 🥁 🥁 🥁 Aaaand the magic happened!
> 
> image
> 
> It's really nice!
> 
> There is only one thing which is bothering me: Do you think it's possible to compile ThorVG with Eclipse as a simple managed build? It would be great to add a very simple version of ThorVG directly to LVGL. We can keep only the very core functionality and the SVG loader, and make it possible to just enable a config to use it. Of course if someone wants to use the full features ThorVG he can use it as well.
> 
> What do you think?

I’m confused. Do you want to create an external link to thorvg in lv_port_pc_eclipse? Or do you want to maintain a private thorvg branch for LVGL? Can you give me some guide?
Does anyone  have some ideas? @XuNeo @bjsylvia

kisvegabor:
I mean both options should be supported:
1. `LV_USE_THORVG_INTERNAL`: use the simple internal ThorVG. (It will ensure a great get started experience and simple porting)
1. `LV_USE_THORVG_EXTERNAL`: assume ThorVG is installed. (It will ensure having all the features and customization)

onecoolx:
> I mean both options should be supported:
> 
> 1. `LV_USE_THORVG_INTERNAL`: use the simple internal ThorVG. (It will ensure a great get started experience and simple porting)
> 2. `LV_USE_THORVG_EXTERNAL`: assume ThorVG is installed. (It will ensure having all the features and customization)

So in `LV_USE_THORVG_INTERNAL` mode, you want to put ThorVG's c++ code into lvgl and compile them together? right？

kisvegabor:
Yes, but only the really essential parts. E.g. we don't need all the loaders. SVG should be enough.

onecoolx:
> Yes, but only the really essential parts. E.g. we don't need all the loaders. SVG should be enough.

I try it and found that it needs to include most .cpp and .h files.  There are many problems when compiling ThorVG's c++ files with lvgl.

kisvegabor:
E.g. in [this meson file](https://github.com/thorvg/thorvg/blob/main/src/loaders/meson.build) I found that it's quite modular. Even if we need to add the whole lib, what were those many problems?

In the meantime I opened this issue at ThorVG: https://github.com/thorvg/thorvg/issues/1742

onecoolx:
> E.g. in [this meson file](https://github.com/thorvg/thorvg/blob/main/src/loaders/meson.build) I found that it's quite modular. Even if we need to add the whole lib, what were those many problems?
> 
> In the meantime I opened this issue at ThorVG: [thorvg/thorvg#1742](https://github.com/thorvg/thorvg/issues/1742)

Most of error are related to the c++ compilation option '-Werror=float-conversion'.  It needs to be fixed in thorvg's source code or turn off the compilation option.

kisvegabor:
How does it work with Meson? Do this turn off this compiler option?

Could you  link a line from the source code where this error is reported?

onecoolx:
> How does it work with Meson? Do this turn off this compiler option?
> 
> Could you link a line from the source code where this error is reported?

I've fixed all the compilation issues and added the ThorVG‘s code  into LVGL. Most build error like:
```
//code:
 angle = angle * M_PI / 180.0;

//build error:
 error: conversion from ‘double’ to ‘float’ may change value [-Werror=float-conversion]

// fix code:
 angle = angle * (float)M_PI / 180.0f;
```

I also updated the code of the [lv_port_lvgl_eclipse ](https://github.com/lvgl/lv_port_pc_eclipse/pull/132)project.

src/libs/thorvg/tvgAnimation.cpp
```diff
@@ -0,0 +1,119 @@
...
+
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+
```
kisvegabor, line 22: For the simplest integration we should have this in the first empty line
```c
#include "../../lv_conf_internal.h"
#if LV_USE_THORVG_INTERNAL
```

And `#endif /*LV_USE_THORVG_INTERNAL*/` at the end of the file. 

I thing it can be add added by a script in a fairly simple way. We should save that script in the `scripts` folder or here for later use.

kisvegabor (commented):
Man, you rock!!!

src/libs/thorvg/tvgAnimation.cpp
onecoolx: I think we should modify the thorvg‘s code as little as possible, because we may need to frequently update the latest community version of thorvg in the future. So I use the build options. What do you think?
```
file(GLOB_RECURSE THORVG_SOURCES ${LVGL_ROOT_DIR}/src/libs/thorvg/*.cpp)
...
if(NOT LV_CONF_BUILD_DISABLE_THORVG_INTERNAL)
endif()
```

kisvegabor:
I'm trying it out and found a minor issue:
```
/bin/ld: ./lvgl/src/libs/thorvg/tvgCapi.o: in function `tvg_animation_set_frame':
/home/kisvegabor/projects/lvgl/eclipse-workspace/lvgl_cpp/Debug/../lvgl/src/libs/thorvg/tvgCapi.cpp:733: undefined reference to `tvg::Animation::frame(float)'
```

```cpp
TVG_API Tvg_Result tvg_animation_set_frame(Tvg_Animation* animation, uint32_t no)
{
    if (!animation) return TVG_RESULT_INVALID_ARGUMENT;
    return (Tvg_Result) reinterpret_cast<Animation*>(animation)->frame(no);
}
```

kisvegabor:
I also found that `lv_example_canvas_8` is not working. 
image

And I added the `#if LV_USE_THORVG_INTERNAL` guard to all files

onecoolx:
> I also found that `lv_example_canvas_8` is not working. image
> 
> And I added the `#if LV_USE_THORVG_INTERNAL` guard to all files

I added some test cases and fixed some issues. Well, I will adjust the code according to your ideas.

kisvegabor:
I started to do the same, but I stop it then :slightly_smiling_face: 

Note that, I've removed `LV_USE_THORVG` as enabling either `LV_USE_THORVG_INTERNAL/EXTERNAL` should be enough as well.

onecoolx:
> I started to do the same, but I stop it then :slightly_smiling_face: 
> 
> 
> 
> Note that, I've removed `LV_USE_THORVG` as enabling either `LV_USE_THORVG_INTERNAL/EXTERNAL` should be enough as well.

ok，I see

kisvegabor:
I've removed lv_coord_t and replaced by plain int32_t. There is compatibility typedef in lv_api_map.h. A simple search and replace should do the job. 

I don't want to bother you with following this changes, so let me fix them later today

onecoolx:
> I've removed lv_coord_t and replaced by plain int32_t. There is compatibility typedef in lv_api_map.h. A simple search and replace should do the job.
> 
> I don't want to bother you with following this changes, so let me fix them later today

Okay, I fixed the problem in the build script and added taste cases. So far, it works well.  :)

<div>

</div>

<div>

</div>

kisvegabor:
The test are passing locally, but the masked image has some [artifacts](https://github.com/lvgl/lvgl/pull/4691/files#diff-c337073a8ab0f8e75a0294b6961ff212096cbe1a090793c56acb4ae0133a5692). 

Two final things to do before merging it:
- Please convert `lv_demo_vector_graphics` to `lv_example_canvas_9` with 320x240 resolution. Probably we can show less things, e.g. only a skewed rounded rectangle instead of many shapes. 
- Please add API comments to `lv_draw_vector.h`. I know it's not the most fun thing to do, but without it we highly limit the adaption of vector graphics.

kisvegabor:
The CI failed when running on ARMv6. It's hard to see what could be the problem, but I think it's some off by one pixels issues due to float precision. 

When I just tried out the vector API I found a few minor issues:
- The image blending as I mentioned
- I've tried this:
```c
    lv_area_t rect = {5, 5, WIDTH - 5, HEIGHT - 5};
    lv_vector_dsc_set_fill_color(ctx, lv_color_lighten(lv_color_black(), 50));
    lv_clear_rect(ctx, &rect);
```
but rectangle overflowed from the left to right
image


Maybe these are related to the failing CI.

onecoolx:
> The CI failed when running on ARMv6. It's hard to see what could be the problem, but I think it's some off by one pixels issues due to float precision.
> 
> When I just tried out the vector API I found a few minor issues:
> 
> * The image blending as I mentioned
> * I've tried this:
> 
> ```c
>     lv_area_t rect = {5, 5, WIDTH - 5, HEIGHT - 5};
>     lv_vector_dsc_set_fill_color(ctx, lv_color_lighten(lv_color_black(), 50));
>     lv_clear_rect(ctx, &rect);
> ```
> 
> but rectangle overflowed from the left to right image
> 
> Maybe these are related to the failing CI.

Ok, let me check it.

kisvegabor:
Sorry, the shifted image was just local issue on my side. 

However, [these test](https://github.com/lvgl/lvgl/actions/runs/6732436482/job/18299135279?pr=4691) really fail on ARMv6:
image

onecoolx:
> Sorry, the shifted image was just local issue on my side.
> 
> However, [these test](https://github.com/lvgl/lvgl/actions/runs/6732436482/job/18299135279?pr=4691) really fail on ARMv6: image

I found that ThorVG's raw image can only handle `alpha-premultiplied` type,  but the decoder result of lvgl is `un-alpha-premultiplied`, which cause to draw errors when drawing semi-transparent pictures.  I fixed it temporarily, but I'm a bit confused, is it wise to modify the ThorVG‘s code? @kisvegabor @XuNeo

XuNeo:
> I found that ThorVG's raw image can only handle `alpha-premultiplied` type,

Hi Gabor,
How about we add `premultiplied` flag to `lv_image_header_t`? It's also required by `vglite` thus makes sense to use a common bit in header. `sw` renderer can simply ignore those images as not supported.

tests/unity/unity_support.c
```diff
@@ -103,11 +103,12 @@ bool lv_test_assert_image_eq(const char * fn_ref)
             uint32_t act_px = 0;
             memcpy(&ref_px, ptr_ref, 3);
             memcpy(&act_px, ptr_act, 3);
-            //printf("0xFF%06x, ", act_px);
 
             uint8_t act_swap[3] = {ptr_act[2], ptr_act[1], ptr_act[0]};
 
             if(memcmp(act_swap, ptr_ref, 3) != 0) {
+                TEST_PRINTF("Erron on x:%d, y:%d. Expected %08X, Actual %08X", x, y, ptr_ref, act_swap);
```
kisvegabor:
```suggestion
                TEST_PRINTF("Error on x:%d, y:%d. Expected %X, Actual %X", x, y, ref_px, act_px);
```

onecoolx:
> Thanks for the update, I'll review the comments today.
> 
> Finally it seems the tests will pass too. I was fighting a little bit with the environment variables in GitHub CI.
> 
> In the meantime I saw that some warnings were disabled [here](https://github.com/lvgl/lvgl/pull/4691/files#diff-4461c617ceae0c7e0206622aacdf555aedd62eb129c2efb74c84fa1567bcbe0dR90-R110). It guess these were disabled to make the ThorVG compile without warnings but it'd be important to keep all these warnings for other parts of LVGL. Could you add an option in Cmake to specify separate compile options for ThorVG?

sure, I will fix it.

src/draw/lv_draw_vector.h
```diff
@@ -0,0 +1,541 @@
...
+ * @param path          pointer to a path
+ */
+void lv_vector_dsc_add_path(lv_vector_dsc_t * dsc, const lv_vector_path_t * path);
+
+/**
+ * Clear a rectangle area use current fill color
+ * @param dsc           pointer to a vector graphic descriptor
+ * @param rect          the area to clear in the buffer
+ */
+void lv_clear_rect(lv_vector_dsc_t * dsc, const lv_area_t * rect);
```
kisvegabor: This name is too generic. What about `lv_draw_vector_clear_rect()`

What will be the color of the rectangle after clearing?

src/draw/lv_draw_vector.h
```diff
@@ -0,0 +1,541 @@
...
+ * @param subpath           pointer to another path which will be added
+ */
+void lv_vector_path_append_path(lv_vector_path_t * path, const lv_vector_path_t * subpath);
+
+/**
+ * Create a vector graphic descriptor
+ * @param layer         pointer to a layer
+ * @return              pointer to the created descriptor
+ */
+lv_vector_dsc_t * lv_vector_dsc_create(lv_layer_t * layer);
```
kisvegabor, line 307: The other draw related function has names like this: `lv_draw_rect_dsc_init`

So I wonder if we can
1. Use it as an `init` function: `lv_vector_dsc_init(&dsc)`
2. Set the layer in `lv_draw_vector(&dsc, layer)`. See [label draw](https://github.com/lvgl/lvgl/blob/master/src/draw/lv_draw_label.h#L115-L116)
3. For consistency with other draw functions use `lv_draw_vector_dsc_*` instead of `lv_vector_dsc_*` (should be just a search and replace)

kisvegabor (commented):
Thank you for the update. 

It looks really great, I just found a few minor inconsistency in the API compared to other draw functions. 

What do you think about them?

src/draw/lv_draw_vector.h
onecoolx: Clear uses the current fill color, which is a separate operation so that the gpu's quick fill feature can be used later. As for the name, the original lv_vector_clear seems more reasonable to me, as it has the same prefix as the other functions

src/draw/lv_draw_vector.h
onecoolx: I don't think that's a good idea.

1. The struct lv_vector_dsc_t is too complex, which contains data structures such as list Pointer and dynamic array，and is not suitable for _init initialization.

2. Passing in the layer when creating dsc is to get scissor_area for clipping, I don't think it's a bad idea to bind ahead of time.

3. lv_draw_vector as a prefix causes the function name to be too long. I think lv_vector is enough to distinguish it from other drawing functions. By the way, I still like lv_vector_draw not lv_draw_vector ：）

@XuNeo @bjsylvia What do you think?

kisvegabor: I don't have a strong opinion on these. We should add these updates only if they doesn't cause too much trouble to keep up consistency.

src/draw/lv_draw_vector.h
kisvegabor: What about `lv_vector_clear_rect()` or `lv_vector_clear_area` (to avoid abbreviation)?

onecoolx: > What about `lv_vector_clear_rect()` or `lv_vector_clear_area` (to avoid abbreviation)?

 Putting it this way,  `lv_vector_clear_area`  maybe be better.

kisvegabor: Okay, please update it to `lv_vector_clear_area` and let's merge the feature.

We can add smaller PRs later if needed.

onecoolx: > Okay, please update it to `lv_vector_clear_area` and let's merge the feature.
> 
> We can add smaller PRs later if needed.

ok

kisvegabor:
And all looks good! Merging! 

Thank you for this very valuable contribution!

zhuxingang:
Is there any example using SVG feature added to the example directory?

Finomnis:
> Is there any example using SVG feature added to the examples directory? Does this new SVG draw feature has good performance on vg_lite GPU platform?

@zhuxingang did you ever find answers to those questions?

PR #4695, opened 2023-10-23
Fix execute permissions and shebangs

szepeviktor:

Ran [the Integrity workflow](https://github.com/szepeviktor/byte-level-care/blob/master/.github/workflows/reusable-integrity.yml) and found chaotic situation in execute permissions.

Now only files with shebang are 0755.


szepeviktor:
🤯 When GitHub says "Empty file." in the diff it really mean "No change."

kisvegabor:
Awesome, thank you! :+1:

szepeviktor:
Yes, it can be added.
See `integrity.yml` that "calls" it.

I run workflows for every byte in the repository, even for insignificant files.

kisvegabor:
I've taken a deeper look into it and found this:
```yml
 # Move TODO-s into GitHub issues!
            name: "Search for TODO-s and FIXME-s 🐌"
            run: |
                ! git grep --extended-regexp --ignore-case -I -e '\b(TODO|FIXME)\b' \
                    -- ':!:.github/workflows/reusable-integrity.yml' ':!:.github/workflows/back-end.yml'
```

Could you tell a few works about it? We have a couple of TODOs and I'm not ready to convert all of them to issues :smile:

szepeviktor:
> Could you tell a few works about it? We have a couple of TODOs

This workflow assumes that viktor is onboard.
So say: _Hey, Viktor! Send a PR removing those TODO-s and open an issue for them!_

You could remove this TODO checking but it will keep TODO-s buried deep in the source code!

PR #4703, opened 2023-10-24
fix issue:LVGL V9 music compiling error

lion2tomato:
fix issue:LVGL V9 music compiling error when setting LV_DEMO_MUSIC_LARGE =y


kisvegabor:
Thanks!

PR #4704, opened 2023-10-24
MSVC build failure due to missing decl specs

1000len-4959:
Although CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS is already used in lvgl it does not work for global data (according to cmake docs).
Therefore we need to decorate global data with some decl specs to make build work with MSVC.
This PR addresses this issue.

### Checkpoints
 

1000len-4959:
@kisvegabor 
Did you mean I should add this code snippet to __lv_conf.h__ in .devcontainer?

#ifdef WIN32
  #ifdef INSIDE_DLL
    #define DLLIMPEXP __declspec(dllexport) extern
  #else
    #define DLLIMPEXP __declspec(dllimport) extern
  #endif
#else
  #define DLLIMPEXP extern
#endif

kdschlosser:
YES!!! This was a HUGE annoyance for me with the CPython Binding. I ended up using pycparser to read over all of the header files to build a header file to decorate the functions properly. It was quite the headache. By having the functions properly decorated LVGL would be able to be compiled into a DLL and could be accessed easily by any other program. 

GCC by default will export without the need to do anything special, not so much with MSVC, 

Best way to go about this is like what has been decided to use a macro to handle it. There is one addition that needs to be made to the macro.

```
#ifdef _MSC_VER
    #ifdef _DLL
        #define DLLIMPEXP extern "C" __declspec(dllexport)
    #else
        #define DLLIMPEXP extern "C" __declspec(dllimport)
    #endif
#else
    #define DLLIMPEXP
#endif

```

This is a better solution. Reason being is the _MSC_VER macro gets set when using the MSVC compiler. WIN32 may get set if compiling using MinGW but the compiler that is being used is not going to be MSVC. so best to use a macro that gets set by the MSVC compiler. The _DLL macro is also something that gets set by the MSVC compiler when compiling a DLL which is when we want to export the functions.

There is a catch to this tho. There is an issue because of quite a few functions being static. you cannot simply export those functions. wrapper functions need to be made in order to make those functions accessible. 


`extern "C"` tells the compiler to undecorate the function names. This is important to use so ordinals do not have to be used to access the functions. 


MSVC is also a fickle beast when it comes to defining functions in header files that are being exported. This is kind of a no no. A function prototype need to be declared in the header file and the function definition needs to be in a source file. This actually makes for cleaner code when the header files are only used for declarations and not actual  running code. Doing this would also make it a lot easier to use pycparser for creating bindings because no source code get processed which complicates things by a very large amount. There is a ripple effect that will make things a lot easier downstream if this is done right.

kdschlosser:
I also do not think that should not be in the lv_conf.h file because it is technically speaking not something that is user changeable. This is something that would be automatically done and it might be better suited to be located in an internal config file instead of where users make changes in order to add and remove features from LVGL.

Ju1He1:
This issue bothers me as well :)

so which location would you guys propose to add these macros? The lv_conf_internal.h or the lv_conf.h

kdschlosser:
using `set(CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS ON)` is only valid if you are using cmake. I am able to compile LVGL in about 5 seconds if I don't use cmake. I wrote a python script that sets up a proper build environment for using MSVC and I compile the source files in parallel. I am able to compile 32 source files in parallel and then link the object files when all of the source files are done compiling. It is always nice for a user to be able to compile using any build front end they want. Unless each function and global is declared as `extern "C" __declspec(dllexport)` I am not able to do this. Currently I have to use MSVC create a preprocessor output file and parse the file locating the things that are needing to be exported and generating a header file and source file with function wrappers that are exported. It's an ugly process and I am not able to keep the LVGL function names. So now I have to also generate a wrapper in Python code that wraps the functions yet again only for the purpose of making the names right. It's a crazy hack that I would love to have gone.

Talk about a large amount of work that has to be done and also the entry for potential bugs is HUGE because of it. Now yes I can compile using cmake. But when developing, making a change and compiling over and over again having the compile time be a few seconds vs close to a minute makes it just a tad more palatable.

The other thing is by specifically setting that macro to functions you can control what functions are to be public or not. I believe there is also a way to do this for GCC I would have to take a look at what needs to be done there.. This will reduce the number of functions that are seen by the user when compiling a shared library.

kdschlosser:
you can see what I am doing in terms of an MSVC build system here..

https://github.com/kdschlosser/lv_cpython/tree/develop

kisvegabor:
If it doesn't has to be so customizsble but we can add it [here](https://github.com/lvgl/lvgl/blob/master/src/lv_conf_internal.h#L2800). I mean something like this:
```c
/*If not defined yet try to define automatically*/
#ifndef LV_ATTRIBUTE_EXTERN
    #if defined(_MSC_VER)
        #ifdef _DLL
            #define DLLIMPEXP extern "C" __declspec(dllexport)
        #else
            #define DLLIMPEXP extern "C" __declspec(dllimport)
        #endif
    #elif defined(OTHER_PLATFORM)
        ...
    #else 
        #define LV_ATTRIBUTE_EXTERN  extern  /*Default*/
   #endif
#endif
```

Finally do we need to use `LV_ATTRIBUTE_EXTERN` only for those global variables?

kdschlosser:
If you want to allow LVGL to be compiled using the MSVC compiler without having to use cmake then that macro would need to be added to every single public function and global. 

There is another way to go about doing it using a def file but I have not gotten this to actually work. I don't remember what it's issue was. I believe it had something to do with the static functions.

kdschlosser:
and the code you have is not going to work. It needs to be


```c
/*If not defined yet try to define automatically*/
#ifndef DLLIMPEXP 
    #if defined(_MSC_VER)
        #ifdef _DLL
            #define DLLIMPEXP extern "C" __declspec(dllexport)
        #else
            #define DLLIMPEXP extern "C" __declspec(dllimport)
        #endif
    #elif defined(OTHER_PLATFORM)
        ...
    #else 
        #define DLLIMPEXP extern  /*Default*/
   #endif
#endif
```

Ju1He1:
> Finally do we need to use LV_ATTRIBUTE_EXTERN only for those global variables?

@kisvegabor  if we add CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS and use cmake to build lvgl than yes. This is the only place where we need to add it.
If somebody (like @kdschlosser ) does NOT use cmake to build lvgl, we need to decorate ALL global functions and ALL global data with this stuff.

IMHO since there are already some cmake files in the LVGL repo we can simply add CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS to the root CmakeLists.txt

Ju1He1:
> and the code you have is not going to work. It needs to be
> 
> ```c
> /*If not defined yet try to define automatically*/
> #ifndef DLLIMPEXP 
>     #ifdef _MSC_VER
>         #ifdef _DLL
>             #define DLLIMPEXP extern "C" __declspec(dllexport)
>         #else
>             #define DLLIMPEXP extern "C" __declspec(dllimport)
>         #endif
>     #elif defined(OTHER_PLATFORM)
>         ...
>     #else 
>         #define DLLIMPEXP extern  /*Default*/
>    #endif
> #endif
> ```

why do you think we need and extra `extern "C"`. All header functions are already within an `extern C` block. Isn't this a duplicate.
I pulled the OP branch and build it locally using latest MSVC 17 and it worked fine

kdschlosser:
I was going by what the MSVC documentation stated.

It looks like if this is used

```c
#ifdef __cplusplus
extern "C" { 
#endif

exported code here

#ifdef __cplusplus
}
#endif
```

then extern "c" doesn't need to be used in the macro. so long as the exported functions/globals are wrapped declaring them all "c" code

kisvegabor:
> then extern "c" doesn't need to be used in the macro. so long as the exported functions/globals are wrapped declaring them all "c" code

We do this is all header files.

kdschlosser:
> We do this is all header files.
I did not go and check so if that is done so long as the declaration is inside of that then you shouldn't need the extern "c"

I am going to mess about with the def file again  to see exactly what is causing that to fail compilation. I believe there is a combination of things that is causing the def file not to work. One of them is something to do with the static functions in some of the widgets and the other is function not being declared and instead they are being defined in the header files. 

MSVC has some tweaky issues like that with some things. I do not know how cmake is handling the exports to make it all work. Probably should check into what it is dong to get it all to work.

Ju1He1:
> > We do this is all header files.
> > I did not go and check so if that is done so long as the declaration is inside of that then you shouldn't need the extern "c"
> 
> I am going to mess about with the def file again to see exactly what is causing that to fail compilation. I believe there is a combination of things that is causing the def file not to work. One of them is something to do with the static functions in some of the widgets and the other is function not being declared and instead they are being defined in the header files.
> 
> MSVC has some tweaky issues like that with some things. I do not know how cmake is handling the exports to make it all work. Probably should check into what it is dong to get it all to work.

Taking the CMake docs
https://cmake.org/cmake/help/latest/prop_tgt/WINDOWS_EXPORT_ALL_SYMBOLS.html
Basically all global symbols are automatically exported into the .def files. If you omit the WINDOWS_EXPORT_ALL_SYMBOLS part only global symbols which are decorated with dllexport are exported into the .def file. CMake also clearly says that this will not work for global data symbols, they still need a__declspec(dllimport)



The design question here is: How would you like to proceed:

1. Decorate ALL functions and ALL global data with the decls specs. This will work too, but is much more work
2. use the (relatively new) WINDOWS_EXPORT_ALL_SYMBOLS to get this done at least for all global functions, and decorate global data with the decls specs

In our company we use 2) to build our multi platform code, and this works nicely ;) 


I'm using MSVC as well, so this kind of bothers me. Therefore I'd pretty much appreciate it if we could get this PR on the way ;)

kdschlosser:
I just gave it a go and it does look like it works properly.

kisvegabor:
> use the (relatively new) WINDOWS_EXPORT_ALL_SYMBOLS to get this done at least for all global functions, and decorate global data with the decls specs

I'm definitely for this! Option 1) would be a hell to maintain. :slightly_smiling_face: 

Could you update the PR accordingly. Just please use the more generic `LV_ATTRIBUTE_EXTERN_DATA` define name.

kdschlosser:
I say go for it. I have already opened PR's for fixing a build issue with the demos and also correcting all of the compiler warnings. I know the demos error fix has been merged so make sure to update your branch with the latest code changes. I'll check on the compiler warnings to se what is happening with that PR. I could have broken a build on another OS when I made the changes. It might still need some work done. The changes in that PR shouldn't effect what you need to do tho.

1000len-4959:
@kisvegabor 
so there are two things to adjust in Pull Request,

1. adding  in root cmakelists

>  set(CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS ON) 

2. adding the following Code to lv_conf_internal.h

> #ifdef _MSC_VER
>   #ifdef LV_ATTRIBUTE_EXTERN_DATA
>     #define LV_ATTRIBUTE_EXTERN  __declspec(dllexport) extern
>   #else
>     #define LV_ATTRIBUTE_EXTERN  __declspec(dllimport) extern
>   #endif //LV_INSIDE_DLL
> #else
>   #define LV_ATTRIBUTE_EXTERN extern
> #endif //_MSC_VER
>             #define LV_ATTRIBUTE_EXTERN extern "C" __declspec(dllimport)
>         #endif
>     #else 
>         #define LV_ATTRIBUTE_EXTERN  extern  /*Default*/
>    #endif
> #endif

kdschlosser:
as a question. any way to have everything that gets compiled using cmake all put into a single folder named build instead of things placed into the root of the project and other things placed into the lib folder...

1000len-4959:
@kdschlosser 
I think your suggestion is good, I will adapt the root cmakelists

image

kisvegabor:
I like the idea of using Cmake as much as possible. Just 2 notes:
1. I meant using `LV_ATTRIBUTE_EXTERN_DATA` in the `.h` files as we might have different prefix for functions. so adding `_DATA` expresses the intent in clearer way.
2. If we use CMake like this we need to add back `LV_ATTRIBUTE_EXTERN_DATA` to `lv_conf_template/internal.h` to define it to empty. Else will have undefined identifier errors.

kisvegabor:
Thank you! I fixed a few minor things (e.g. `lv_conf_internal.h` is generated from `lv_conf_template.h`).

Let's see what the CI says.

kdschlosser:
The problem with creating the macro in the cmake file like that is others locating the macro if needs be at some later date. The last place I would think to look for the definition of a C macro would be in a cmake file. I would start at the primary header and look backwards. It's kind of an odd place for it.

kisvegabor:
I've just fixed a conflict. Not all looks good to me. Let me know if you think it's read to merge.

1000len-4959:
@kisvegabor it looks good, please merge it

1000len-4959:
I think this solution is not really good, what if we want to build lvgl statically
```
if (MSVC)
  if (BUILD_SHARED_LIBS)
    target_compile_definitions(lvgl PUBLIC LV_ATTRIBUTE_EXTERN_DATA=__declspec\(dllimport\))
  else()
    target_compile_definitions(lvgl PUBLIC LV_ATTRIBUTE_EXTERN_DATA=__declspec\(dllexport\))
  endif()
endif()
```
I would change the code to the following code
```
if (MSVC)
  # Check if LVGL is being built as a Shared Library.
  if (BUILD_SHARED_LIBS)
    # Defines LV_ATTRIBUTE_EXTERN_DATA for importing from a DLL when lvgl is used as a DLL.
    target_compile_definitions(lvgl PUBLIC LV_ATTRIBUTE_EXTERN_DATA=__declspec\(dllimport\))
  elseif(LVGL_STATIC_BUILD)
    # If LVGL_STATIC_BUILD is defined, then no __declspec declarations are needed.
    target_compile_definitions(lvgl PUBLIC LV_ATTRIBUTE_EXTERN_DATA=)
  else()
    # Defines LV_ATTRIBUTE_EXTERN_DATA for exporting to a DLL when lvgl is built as a DLL.
    target_compile_definitions(lvgl PUBLIC LV_ATTRIBUTE_EXTERN_DATA=__declspec\(dllexport\))
  endif()
endif()
```

kisvegabor:
Merged, thank you for your contribution!

PR #4711, opened 2023-10-26
feat(tiny_ttf): add option to disable kerning

niklasf:
For some fonts, or in some cases for performance, it can be useful to disable kerning. Added `lv_tiny_ttf_set_use_kerning()` to configure this. On by default, as before.


kisvegabor:
I think it could be a useful feature in general. What do you think about adding a flag to `lv_font_t` to indicate whether to use kerning or not?

niklasf:
Makes sense. Closing in favor of #4725.

PR #4714, opened 2023-10-26
fix(tiny_ttf): fix font metrics confusion

niklasf:
Fix underline position. It is relative to the baseline.

Change `lv_tiny_ttf_create_xxx()` and `lv_tiny_ttf_set_size()` to accept font size rather than line height. Font size is the more conventional way to specify font scale. It is also more consistent with the naming of the function, and the names of the built-in fonts in lvgl.


kisvegabor:
`font_size` is really more common. I believe the [online converter](https://lvgl.io/tools/fontconverter)  (and [lv_font_conv](https://github.com/lvgl/lv_font_conv)) uses that as well.

PR #4725, opened 2023-10-29
feat(font): add option to disable kerning

niklasf:
Alternative approach to #4711. Now the option is independent of the font engine. That is `LV_FONT_KERNING_NONE` will disable kerning, `LV_FONT_KERNING_NORMAL` will use kerning *if the engine and font supports it*. Naming based on [font-kerning in CSS](https://developer.mozilla.org/en-US/docs/Web/CSS/font-kerning).

Not sure about these:

* `LV_FONT_KERNING_NORMAL` is zero, so that it is the default in zeroed `lv_font_t`, but semantically it's the truthy value.
* Other than `user_data`, this would be the first property of `lv_font_t` that is not controlled by the font engine.

It would also be completely fair to reject this feature. It's always possible to just generate the font itself without kerning.


src/font/lv_font.c
```diff
@@ -62,7 +62,7 @@ bool lv_font_get_glyph_dsc(const lv_font_t * font_p, lv_font_glyph_dsc_t * dsc_o
     dsc_out->resolved_font = NULL;
 
     while(f) {
-        bool found = f->get_glyph_dsc(f, dsc_out, letter, letter_next);
+        bool found = f->get_glyph_dsc(f, dsc_out, letter, f->kerning == LV_FONT_KERNING_NONE ? 0 : letter_next);
```
kisvegabor, line 65: It's so elegant to handle it like this! 👍

kisvegabor:
Looks really good! 

To finalize it we need a little bit of background work:
- Add an API function like `lv_font_set_kerning(lv_font_t * font, lv_font_kerning_t kerning)`
- Add a paragraph about it in the docs: https://github.com/lvgl/lvgl/blob/master/docs/overview/font.rst
- Add screen shot compare tests, similar to:
   - https://github.com/lvgl/lvgl/blob/master/tests/src/test_cases/draw/test_draw_label.c#L59-L66
   - https://github.com/lvgl/lvgl/blob/master/tests/src/test_cases/libs/test_tiny_ttf.c#L33-L39

niklasf:
Added docs and tests.

kisvegabor:
Thank you! It looks exactly as I've imagined :slightly_smiling_face: 

The MicroPython test is not failing locally, so I merged it. Let's see what happens aster the merge.

PR #4726, opened 2023-10-30
chore(demo): fix -Wenum-int-mismatch

niklasf:
To match signature at https://github.com/lvgl/lvgl/blob/829198c2353faf480f3302bfec097e7e11bb1d09/demos/render/lv_demo_render.h#L51


kisvegabor:
Thanks!

PR #4727, opened 2023-10-30
feat(tiny_ttf): backport Tiny TTF to lvgl 8

niklasf:
Unlike FreeType, the API offered by sbtt supports obtaining glyph metrics without rendering a bitmap, which is much more efficient with lvgl's typical usage patterns. It also supports kerning.

It would be nice to have this in lvgl 8, too.

Backported in two commits, to make this reviewable. The first is a verbatim copy from lvgl 9. The second are small changes to make it work under lvgl 8:

* Some renamed APIs
* Cache API not available under lvgl 8. Use lv_lru, instead
* Not rendering at a specific stride


kisvegabor:
Hey,

Thanks! Looks great, just please add a basic screenshot compare test too.

niklasf:
Done. The test suite actually uncovered a memory leak (I didn't realize the LRU cache doesn't take ownership of the keys, like it does for the values).

kisvegabor:
Cool, thank you!

PR #4728, opened 2023-10-30
feat(stdlib): add and use lv_malloc_zeroed

niklasf:
`lv_malloc()` followed by `lv_memzero()` is a very common pattern. We could introduce `lv_calloc()` for brevity and to use:

* `calloc()` for `LV_STDLIB_CLIB`
* `m_malloc0()` for MicroPython

### DIFF
--- a/src/stdlib/lv_mem.c
+++ b/src/stdlib/lv_mem.c
@@ -83,8 +83,8 @@ void * lv_malloc(size_t size)
         LV_LOG_INFO("used: %6d (%3d %%), frag: %3d %%, biggest free: %6d",
                     (int)(mon.total_size - mon.free_size), mon.used_pct, mon.frag_pct,
                     (int)mon.free_biggest_size);
-        return NULL;
 #endif
+        return NULL;
     }
 
 #if LV_MEM_ADD_JUNK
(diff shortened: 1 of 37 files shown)


src/stdlib/lv_mem.c
```diff
@@ -83,8 +83,8 @@ void * lv_malloc(size_t size)
         LV_LOG_INFO("used: %6d (%3d %%), frag: %3d %%, biggest free: %6d",
                     (int)(mon.total_size - mon.free_size), mon.used_pct, mon.frag_pct,
                     (int)mon.free_biggest_size);
-        return NULL;
 #endif
+        return NULL;
```
niklasf, line 87: Notable change

kisvegabor:
Do you know if `calloc` internally can be more efficient than `malloc` + `memset`?

niklasf:
It depends. Compilers generally optimize `malloc` + `memset` to `calloc`, already. So we would only see an effect for previously missed optimizations. And then it only helps if the allocator has prepared some zeroed memory.

It's definitely convenient, though, especially if `size` is computed.

kisvegabor:
Sorry, but I don't feel it strong enough to have new API for this. Maybe just a helper like 
```c
void * lv_malloc_zero(size_t size)
{
     void * p =lv_malloc(size);
     if(p) lv_memzro(p, size);
    return p;
}
```

I think it's better because
1. Its API is the same as malloc
2. It doesn't require a new function in all stdlib wrappers

niklasf:
Sure. Seems like a useful utility to add even in this way.

kisvegabor:
Great, thank you!

PR #4735, opened 2023-11-01
Try to fix a label bug.

Polarix:
### Feature
- Add a interface for label, used for set a formated text with argument list.
 
### Fix
- Fix label text paint will overflow the border when long mode set to LV_LABEL_LONG_CLIP.

### DIFF
--- a/src/widgets/lv_label.c
+++ b/src/widgets/lv_label.c
@@ -172,6 +172,31 @@ void lv_label_set_text_fmt(lv_obj_t * obj, const char * fmt, ...)
     lv_label_refr_text(obj);
 }
 
+void lv_label_set_text_vfmt(lv_obj_t * obj, const char* fmt, va_list args)
+{
+    LV_ASSERT_OBJ(obj, MY_CLASS);
+    LV_ASSERT_NULL(fmt);
+
+    lv_obj_invalidate(obj);
+    lv_label_t * label = (lv_label_t *)obj;
+
+    /*If text is NULL then refresh*/
+    if(fmt == NULL) {
+        lv_label_refr_text(obj);
+        return;
+    }
+
+    if(label->text != NULL && label->static_txt == 0) {
+        lv_mem_free(label->text);
+        label->text = NULL;
+    }
+
+    label->text = _lv_txt_set_text_vfmt(fmt, args);
+    label->static_txt = 0; /*Now the text is dynamically allocated*/
+
+    lv_label_refr_text(obj);
+}
+
 void lv_label_set_text_static(lv_obj_t * obj, const char * text)
 {
     LV_ASSERT_OBJ(obj, MY_CLASS);
--- a/src/widgets/lv_label.c
+++ b/src/widgets/lv_label.c
@@ -849,18 +849,10 @@ static void draw_main(lv_event_t * e)
         lv_area_move(&txt_coords, 0, -s);
         txt_coords.y2 = obj->coords.y2;
     }
-    if(label->long_mode == LV_LABEL_LONG_SCROLL || label->long_mode == LV_LABEL_LONG_SCROLL_CIRCULAR) {
-        const lv_area_t * clip_area_ori = draw_ctx->clip_area;
-        draw_ctx->clip_area = &txt_clip;
-        lv_draw_label(draw_ctx, &label_draw_dsc, &txt_coords, label->text, hint);
-        draw_ctx->clip_area = clip_area_ori;
-    }
-    else {
-        lv_draw_label(draw_ctx, &label_draw_dsc, &txt_coords, label->text, hint);
-    }
 
-    const lv_area_t * clip_area_ori = draw_ctx->clip_area;
+    const lv_area_t* clip_area_ori = draw_ctx->clip_area;
     draw_ctx->clip_area = &txt_clip;
+    lv_draw_label(draw_ctx, &label_draw_dsc, &txt_coords, label->text, hint);
 
     if(label->long_mode == LV_LABEL_LONG_SCROLL_CIRCULAR) {
         lv_point_t size;
--- a/src/widgets/lv_label.h
+++ b/src/widgets/lv_label.h
@@ -109,6 +109,15 @@ void lv_label_set_text(lv_obj_t * obj, const char * text);
 void lv_label_set_text_fmt(lv_obj_t * obj, const char * fmt, ...) LV_FORMAT_ATTRIBUTE(2, 3);
 
 /**
+ * Set a new formatted text for a label. Memory will be allocated to store the text by the label.
+ * It's also used for c++ binding when package a new format function.
+ * @param obj           pointer to a label object
+ * @param fmt           `printf`-like format
+ * @param args          argument list.
+ */
+void lv_label_set_text_vfmt(lv_obj_t * obj, const char* fmt, va_list args);
+
+/**
  * Set a static text. It will not be saved by the label so the 'text' variable
  * has to be 'alive' while the label exists.
  * @param obj           pointer to a label object


kisvegabor:
> Fix label text paint will overflow the border when long mode set to LV_LABEL_LONG_CLIP.

Could you show a code snippet to reproduce this issue?

Polarix:
> > Fix label text paint will overflow the border when long mode set to LV_LABEL_LONG_CLIP.
> 
> Could you show a code snippet to reproduce this issue?

OK，following is demo code.

```c
            lv_obj_t* label_obj = lv_label_create(s_container_panel);
            lv_obj_set_style_text_font(label_obj, &regular_24px, LV_STATE_DEFAULT);
            lv_obj_set_style_text_color(label_obj, lv_color_black(), LV_STATE_DEFAULT);
            lv_label_set_text(label_obj, "Hello world!Hello world!");
            lv_obj_set_pos(label_obj, 5, 5);
            lv_obj_set_size(label_obj, 100, LV_SIZE_CONTENT);
            lv_label_set_long_mode(label_obj, LV_LABEL_LONG_CLIP);
            lv_obj_set_style_border_width(label_obj, 1, LV_STATE_DEFAULT);
            lv_obj_set_style_border_color(label_obj, lv_color_black(), LV_STATE_DEFAULT);
            lv_obj_set_style_border_opa(label_obj, LV_OPA_100, LV_STATE_DEFAULT);
```
Without borders, this is really difficult to detect。

kisvegabor:
Thank you!

I think I could fix the clipping issue like this:
```c
//lv_label.c, draw_main()
    if(label->long_mode == LV_LABEL_LONG_WRAP) {
        int32_t s = lv_obj_get_scroll_top(obj);
        lv_area_move(&txt_coords, 0, -s);
        txt_coords.y2 = obj->coords.y2;
    }
    /*Check LV_LABEL_LONG_CLIP too*/
    else if(label->long_mode == LV_LABEL_LONG_SCROLL || label->long_mode == LV_LABEL_LONG_SCROLL_CIRCULAR || label->long_mode == LV_LABEL_LONG_CLIP) {
        const lv_area_t clip_area_ori = layer->clip_area;
        layer->clip_area = txt_clip;
        lv_draw_label(layer, &label_draw_dsc, &txt_coords);
        layer->clip_area = clip_area_ori;
    }
    else {
        lv_draw_label(layer, &label_draw_dsc, &txt_coords);
    }

```

If you agree, could you modify the PR accordingly?

lhdjply:
> Thank you!
> 
> I think I could fix the clipping issue like this:
> 
> ```c
> //lv_label.c, draw_main()
>     if(label->long_mode == LV_LABEL_LONG_WRAP) {
>         int32_t s = lv_obj_get_scroll_top(obj);
>         lv_area_move(&txt_coords, 0, -s);
>         txt_coords.y2 = obj->coords.y2;
>     }
>     /*Check LV_LABEL_LONG_CLIP too*/
>     else if(label->long_mode == LV_LABEL_LONG_SCROLL || label->long_mode == LV_LABEL_LONG_SCROLL_CIRCULAR || label->long_mode == LV_LABEL_LONG_CLIP) {
>         const lv_area_t clip_area_ori = layer->clip_area;
>         layer->clip_area = txt_clip;
>         lv_draw_label(layer, &label_draw_dsc, &txt_coords);
>         layer->clip_area = clip_area_ori;
>     }
>     else {
>         lv_draw_label(layer, &label_draw_dsc, &txt_coords);
>     }
> ```
> 
> If you agree, could you modify the PR accordingly?

this is version 8.3 ，is not master。
`lv_draw_label(layer,` &label_draw_dsc, &txt_coords);
The function is missing two parameters

kisvegabor:
Yes, I really tested with `master`. However the idea is the same, just add the CLIP mode here:
```c
    else if(label->long_mode == LV_LABEL_LONG_SCROLL || label->long_mode == LV_LABEL_LONG_SCROLL_CIRCULAR || label->long_mode == LV_LABEL_LONG_CLIP) {

```

Polarix:
> Yes, I really tested with `master`. However the idea is the same, just add the CLIP mode here:
> 
> ```c
>     if(label->long_mode == LV_LABEL_LONG_SCROLL || label->long_mode == LV_LABEL_LONG_SCROLL_CIRCULAR || label->long_mode == LV_LABEL_LONG_CLIP) {
> ```

And there is also the bug of overflow of the top and bottom border in LV_LABEL_LONG_WRAP mode.
Like this.
lvgl_bug-2

kisvegabor:
Ahh, true! Just to see why it happens: In some cases the letters of italic fonts are out of their box. And if we set the `lv_label`'s size according the box sizes of the letters, parts of the last letter might be clipped. 

See 
image


So now we have a few pixels of extra draw area round the label. However it should be applied only horizontally in case of `WARP` mode. In other modes we should clip the text to the real widget size.  

What do you think?

Polarix:
> Ahh, true! Just to see why it happens: In some cases the letters of italic fonts are out of their box. And if we set the `lv_label`'s size according the box sizes of the letters, parts of the last letter might be clipped.
> 
> See image
> 
> So now we have a few pixels of extra draw area round the label. However it should be applied only horizontally in case of `WARP` mode. In other modes we should clip the text to the real widget size.
> 
> What do you think?

But in my opinion，what ever, all of text paint should not overflow the border，so the clip area should not greater then border of the control。
According to my previous win32 development experience，MFC & wxWidgets both are like this.

Demo

Polarix:
Append a shot with different bg color.
Demo2

kisvegabor:
And do you solve the issue of clipped letters  in wxWidgets or Win32?

In LVGL we could say that the user should add some padding to the labels.

Polarix:
> And do you solve the issue of clipped letters in wxWidgets or Win32?
> 
> In LVGL we could say that the user should add some padding to the labels.

No no no, I mean that's paint content cannot overflow the border in others GUI like QT or wxWidgets.
and in LVGL, set pading style cannot reduce the clip area。

I think that if text paint is allowed to overflow the control border, then overlap might be happening. like this.
lvgl_bug-3
But if overlaped control repaint, out of label control area content is not repaint, like this, after clicked, button will be repaint. but out of label content will be covered.
lvgl_bug-4

so, in my opinion, in any case, it makes no sense to draw control content outside of the control.

kisvegabor:
I see, so what we can do is:
- set the text's clip are to the widget's coordinates 
- if the user needs a larger area around the text he can use padding. E.g.
image
```c
  lv_obj_t * label = lv_label_create(lv_scr_act());
  lv_obj_center(label);
  lv_obj_set_style_border_width(label, 1, 0);
  lv_obj_set_style_pad_all(label, 20, 0);
```

Polarix:
> I see, so what we can do is:
> 
> * set the text's clip are to the widget's coordinates
> * if the user needs a larger area around the text he can use padding. E.g.
>   image
> 
> ```c
>   lv_obj_t * label = lv_label_create(lv_scr_act());
>   lv_obj_center(label);
>   lv_obj_set_style_border_width(label, 1, 0);
>   lv_obj_set_style_pad_all(label, 20, 0);
> ```

Yes, pad value might be useful, but it's not a panacea, I found this issue because I was try to make such a widget:
lvgl_bug-5
In this widget, multiple strings will be stored, when current string scrolling end, next string will scroll up to replace current string.
I try to set pad value to set text vertical center, but it shown like this.
so, I think the pad value might no improve when text-height/widget+pad is greater then label control height/widget.

kisvegabor:
I mean applying padding on the label widget.

kisvegabor:
I'll take care of this before releasing v9

kisvegabor:
Not stale

kisvegabor:
Closing this PR now. See https://github.com/lvgl/lvgl/pull/5381

PR #299, opened 2018-07-06
fix compile errors related to USE_LV_ANIMATION, format code

aenchevich:
Fixing compile errors related to USE_LV_ANIMATION and formatting code to project coding style.

PR #4738, opened 2023-11-01
perf(dev): use flush_wait_cb for direct double-buffered drm

niklasf:
Follow up to #299, using the new flush wait mechanism.

As expected, the average time to flush on my old Raspberry Pi dropped from ~10ms to ~0ms.

I see you kept both mechanisms. The new mechanism is technically strictly more powerful than `lv_display_flush_ready()`, and drivers could always wait for their own volatile flag, but I guess it's so common that it's worth keeping around for convenience?

### DIFF
--- a/src/dev/display/drm/lv_linux_drm.c
+++ b/src/dev/display/drm/lv_linux_drm.c
@@ -66,9 +66,6 @@ typedef struct {
     drmModePropertyPtr crtc_props[128];
     drmModePropertyPtr conn_props[128];
     drm_buffer_t drm_bufs[2]; /*DUMB buffers*/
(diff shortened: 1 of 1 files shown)


kisvegabor:
Wow, 10ms to 0ms is huge. What FPS do you see now?

Yes, I kept it just for backward compatibility.

niklasf:
In my current test setup it doesn't move FPS much, because the weak device spends too much time in rendering for a way too large display. It shoud help in general, though.

kisvegabor:
Anyway, 10 ms is 10 ms. :slightly_smiling_face:

PR #4745, opened 2023-11-02
Kconfig: Add guard checking for LVGL module

faxe1008:
Adds a guard to the Kconfig file to prevent generation of LV_* symbols if the module is not used.

It was noticed that due to how the module is set up in the zephyr ecosystem a lot of `LV_` kconfig symbols end up in the generated header files even if the application is build without usage of the LVGL module. In our fork the change already has a PR: [pull/48](https://github.com/zephyrproject-rtos/lvgl/pull/48). I thought we might as well upstream the changes.

### Checkpoints
 
### DIFF
--- a/Kconfig
+++ b/Kconfig
@@ -1,5 +1,4 @@
-# Kconfig file for LVGL v8.0
-
+if LVGL
 menu "LVGL configuration"
 
 	# Define CONFIG_LV_CONF_SKIP so we can use LVGL
@@ -1410,3 +1409,4 @@ menu "LVGL configuration"
 	endmenu
 
 endmenu
+endif


kisvegabor:
Hey,

If for example Espressif (ESP32) uses LVGL's Kconfig they need to add the LVGL config option somehow right?

faxe1008:
Hi, yes that is true. If its not feasable for esp-idf to add this I'm happy to have this only in our fork.

bjsylvia:
@faxe1008 @kisvegabor Why assume that a project containing the LVGL source code library must have a configuration item named `LVGL`? Avoiding the inclusion of the overall LVGL configuration can be achieved by enabling it through a custom CONFIG and then including `lvgl/Kconfig` in your project. Such as:

```
menuconfig USE_LVGL
    bool "Enable LVGL"
    default n

if USE_LVGL
source "$YOUR_LVGL_PATH/Kconfig"
endif
```

faxe1008:
Its because of the build ecosystem that zephyr uses. Essentially modules are pulled (unless allow/deny listed) and in the case of lvgl we are using the kconfig the project provides. We could in theory add another kconfig in between as you suggested, but at that point we can simply put the guard directly into the kconfig of our fork. I was not aware that esp-idf is also using kconfig for this. So if the hassle to add this is too big I'm also happy with just having this in our fork and closing this PR

kisvegabor:
It really seems like specific to Zephyr, and if we add this we will break other systems. So unless there is an option to keep the original simlicity I would recommend having the extra guard in your fork.

PR #4764, opened 2023-11-08
fix(render): make parallel rendering work based on Nuttx Simulator

bjsylvia:

Fix the white screen display issue caused by a busy loop and potential use of uninitialized condition on Nuttx simulator.

### Checkpoints
 
### DIFF
--- a/src/draw/lv_draw.h
+++ b/src/draw/lv_draw.h
@@ -119,8 +119,10 @@ typedef struct _lv_draw_unit_t {
      * @return              >=0:    The number of taken draw task:
      *                                  0 means the task has not yet been completed.
      *                                  1 means a new task has been accepted.
-     *                      -1:     There are no available draw tasks at all.
-     *                              Also means to no call the dispatcher of the other draw units as there is no draw task to take
+     *                      -1:     The draw unit wanted to work on a task but couldn't do that
+     *                              due to some errors (e.g. out of memory).
+     *                              It signals that LVGL should call the dispatcher later again
+     *                              to let draw unit try to start the rendering again.
      */
     int32_t (*dispatch_cb)(struct _lv_draw_unit_t * draw_unit, struct _lv_layer_t * layer);
 
--- a/src/draw/lv_draw.h
+++ b/src/draw/lv_draw.h
@@ -116,8 +116,10 @@ typedef struct _lv_draw_unit_t {
      * A draw task should be assign only if the draw unit can draw it too
      * @param draw_unit     pointer to the draw unit
      * @param layer         pointer to a layer on which the draw task should be drawn
-     * @return              >=0:    The number of taken draw task
-     *                      -1:     There where no available draw tasks at all.
+     * @return              >=0:    The number of taken draw task:
+     *                                  0 means the task has not yet been completed.
+     *                                  1 means a new task has been accepted.
+     *                      -1:     There are no available draw tasks at all.
      *                              Also means to no call the dispatcher of the other draw units as there is no draw task to take
      */
     int32_t (*dispatch_cb)(struct _lv_draw_unit_t * draw_unit, struct _lv_layer_t * layer);
--- a/src/draw/sw/lv_draw_sw.c
+++ b/src/draw/sw/lv_draw_sw.c
@@ -113,7 +113,7 @@ static int32_t lv_draw_sw_dispatch(lv_draw_unit_t * draw_unit, lv_layer_t * laye
 
 #if LV_USE_OS
     /*Let the render thread work*/
-    lv_thread_sync_signal(&draw_sw_unit->sync);
+    if(draw_sw_unit->inited) lv_thread_sync_signal(&draw_sw_unit->sync);
 #else
     execute_drawing(draw_sw_unit);
 
@@ -133,6 +133,7 @@ static void render_thread_cb(void * ptr)
     lv_draw_sw_unit_t * u = ptr;
 
     lv_thread_sync_init(&u->sync);
+    u->inited = true;
 
     while(1) {
         while(u->task_act == NULL) {
@@ -148,6 +149,8 @@ static void render_thread_cb(void * ptr)
         /*The draw unit is free now. Request a new dispatching as it can get a new task*/
         lv_draw_dispatch_request();
     }
+    u->inited = false;
+    lv_thread_sync_delete(&u->sync);
 }
 #endif
 
--- a/src/draw/sw/lv_draw_sw.c
+++ b/src/draw/sw/lv_draw_sw.c
@@ -80,6 +80,14 @@ static int32_t lv_draw_sw_delete(lv_draw_unit_t * draw_unit)
 {
 #if LV_USE_OS
     lv_draw_sw_unit_t * draw_sw_unit = (lv_draw_sw_unit_t *) draw_unit;
+
+    LV_LOG_INFO("cancel software rendering thread");
+    draw_sw_unit->exit_status = true;
+
+    if(draw_sw_unit->inited) {
+        lv_thread_sync_signal(&draw_sw_unit->sync);
+    }
+
     return lv_thread_delete(&draw_sw_unit->thread);
 #else
     LV_UNUSED(draw_unit);
@@ -137,7 +145,15 @@ static void render_thread_cb(void * ptr)
 
     while(1) {
         while(u->task_act == NULL) {
+            if(u->exit_status) {
+                break;
+            }
             lv_thread_sync_wait(&u->sync);
+        }
+
+        if(u->exit_status) {
+            LV_LOG_INFO("ready to exit software rendering thread");
+            break;
         }
 
         execute_drawing(u);
@@ -149,8 +165,10 @@ static void render_thread_cb(void * ptr)
         /*The draw unit is free now. Request a new dispatching as it can get a new task*/
         lv_draw_dispatch_request();
     }
+
     u->inited = false;
     lv_thread_sync_delete(&u->sync);
+    LV_LOG_INFO("exit software rendering thread");
 }
 #endif
 
--- a/src/draw/sw/lv_draw_sw.h
+++ b/src/draw/sw/lv_draw_sw.h
@@ -36,6 +36,7 @@ typedef struct {
     lv_thread_sync_t sync;
     lv_thread_t thread;
     volatile bool inited;
+    volatile bool exit_status;
 #endif
     uint32_t idx;
 } lv_draw_sw_unit_t;
--- a/src/draw/sw/lv_draw_sw.h
+++ b/src/draw/sw/lv_draw_sw.h
@@ -35,6 +35,7 @@ typedef struct {
 #if LV_USE_OS
     lv_thread_sync_t sync;
     lv_thread_t thread;
+    volatile bool inited;
 #endif
     uint32_t idx;
 } lv_draw_sw_unit_t;
--- a/src/draw/lv_draw.c
+++ b/src/draw/lv_draw.c
@@ -261,9 +261,6 @@ bool lv_draw_dispatch_layer(struct _lv_display_t * disp, lv_layer_t * layer)
             lv_draw_unit_t * u = _draw_info.unit_head;
             while(u) {
                 int32_t taken_cnt = u->dispatch_cb(u, layer);
-                if(taken_cnt < 0) {
-                    break;
-                }
                 if(taken_cnt >= 0) render_running = true;
                 u = u->next;
             }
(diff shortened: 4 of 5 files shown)


kisvegabor:
I've just realized something. Now that we have introduced the `evaluate_cb` (so a `draw_unit` can get only the tasks assigned to it, see [here](https://github.com/lvgl/lvgl/blob/master/src/draw/lv_draw.c#L316)) it can happen that there are no available tasks for a GPU, but there are available tasks for an other one. Therefore 
```
 if(taken_cnt < 0) {
   break;
 }
```
is wrong. `dispatch_cb` should return only >= 0 (uint32_t) as it  doesn't know if there are tasks for the other draw units and we need to check all the draw units without breaking. 

As I'm looking into again I think `need_waiting` should `render_running` which means: at least one rendering is in progress which will trigger a dispatch request when it's ready so no need to send a dispatch request manually.

kisvegabor:
cc @nicusorcitu

nicusorcitu:
This is what I was trying to explain also :)

kisvegabor:
Oh, sorry :sweat_smile: Sometimes I feel that I figured out something that people are explaining to me for days (or years) :smile:

bjsylvia:
> I've just realized something. Now that we have introduced the `evaluate_cb` (so a `draw_unit` can get only the tasks assigned to it, see [here](https://github.com/lvgl/lvgl/blob/master/src/draw/lv_draw.c#L316)) it can happen that there are no available tasks for a GPU, but there are available tasks for an other one. Therefore
> 
> ```
>  int32_t taken_cnt = u->dispatch_cb(u, layer);
>  if(taken_cnt < 0) {
>    break;
>  }
> ```
> 
> is wrong. `dispatch_cb` should return only >= 0 (uint32_t) as it doesn't know if there are tasks for the other draw units and we need to check all the draw units without breaking.

I think it is necessary for `dispatch_cb` to return a negative value when there are no tasks to handle or when memory allocation fails. This situation indicates that there are no rendering tasks being executed. 
what we need to handle is to remove the `break` statement you mentioned above, allowing all draw units to be traversed completely.
 
> As I'm looking into again I think `need_waiting` should `render_running` which means: at least one rendering is in progress which will trigger a dispatch request when it's ready so no need to send a dispatch request manually.

Done: replace  `need_waiting` with `render_running`

src/draw/sw/lv_draw_sw.h
```diff
@@ -35,6 +35,7 @@ typedef struct {
 #if LV_USE_OS
     lv_thread_sync_t sync;
     lv_thread_t thread;
+    volatile bool inited;
```
nicusorcitu, line 38: The correct word is "initialized"

src/draw/lv_draw.h
```diff
@@ -116,8 +116,10 @@ typedef struct _lv_draw_unit_t {
      * A draw task should be assign only if the draw unit can draw it too
      * @param draw_unit     pointer to the draw unit
      * @param layer         pointer to a layer on which the draw task should be drawn
-     * @return              >=0:    The number of taken draw task
-     *                      -1:     There where no available draw tasks at all.
+     * @return              >=0:    The number of taken draw task:
+     *                                  0 means the task has not yet been completed.
+     *                                  1 means a new task has been accepted.
+     *                      -1:     There are no available draw tasks at all.
```
kisvegabor:
```suggestion
     *                      -1:     The draw unit wanted to work on on a task but couldn't do that due to some errors (e.g. out of memory). 
     *                              It signals that LVGL should call the dispatcher later again to let draw unit try to start the rendering again.  
```

src/draw/sw/lv_draw_sw.h
kisvegabor: We are really abut avoiding abbreviation in the API, however "inited" is used in many other places as well. So let's use it now, and review all occurrences separately if required.

kisvegabor:
@bjsylvia I agree with your comments

src/draw/sw/lv_draw_sw.c
```diff
@@ -113,7 +113,7 @@ static int32_t lv_draw_sw_dispatch(lv_draw_unit_t * draw_unit, lv_layer_t * laye
 
 #if LV_USE_OS
     /*Let the render thread work*/
-    lv_thread_sync_signal(&draw_sw_unit->sync);
+    if(draw_sw_unit->inited) lv_thread_sync_signal(&draw_sw_unit->sync);
```
nicusorcitu, line 124: You are trying to avoid a case where the signal could be send from the lvg_main thread to the draw_thread unit that is not yet running, hence not waiting for the signal yet.
This would happen only in case the lvgl_main thread have a bigger priority than the draw_thread.
Am I right?

src/draw/lv_draw.h
nicusorcitu: I believe the return code of any unit dispatcher shall be like this:
```
 0 - unit is busy with other task or unit does not support any new task from the list
 1 - new task has been taken
-1 - out of memory during layer allocation or there are no more tasks in the list.
```

src/draw/sw/lv_draw_sw.c
bjsylvia: Not only the situation you mentioned, but also the thread scheduling strategies of various RTOSes are different, and there may be different performance behaviors even with the same priority.

nicusorcitu:
```
else
    return 0;
```
You can not return 1 if the condition is not true.

src/draw/lv_draw.h
kisvegabor: >  0 - unit is busy with other task or unit does not support any new task from the list

Clear: in both cases a dispatch request will be sent when the current draw task or an other draw task in an other unit is ready, so this unit can try render a task. I would add this longer description in the comments because these parts are really complicated.

>  1 - new task has been taken

Also clear. Maybe we can return the number of taken tasks, but I don't know for what we can use that information

> -1 - out of memory during layer allocation or there are no more tasks in the list.

1. Out of memory: it's fine, but let's describe that LVGL will try to dispatach later
2. No more tasks: it's a different case as no more dispatching is required. We can see it from `layer->draw_task_head == NULL`

src/draw/sw/lv_draw_sw.c
kisvegabor: Or `return -1` to ask LVGL to dispatch again.

nicusorcitu: I was thinking maybe another unit is available to take it. -1 shall work too.

src/draw/lv_draw.h
bjsylvia: I have updated description for -1. Please check it.

src/draw/sw/lv_draw_sw.c
bjsylvia: Still returning 1 because we have configured `task_act` for draw unit, which means once the thread starts successfully, `execute_drawing` in `render_thread_cb` will be triggered without waiting for any condition." 

This is consistent with our definition of `dispatch_cb` returning 1, which means a new task has been accepted.

nicusorcitu: No, because how would you know next time if your unit is busy with other task in order to return 0?

bjsylvia: When `draw_act` is not consumed and then dispatched again, it will return 0.

kisvegabor: `if(draw_sw_unit->inited == false)`  the unit shouldn't take any tasks, as it's not initialized. So the `inited` flag should be checked very early.

bjsylvia: Please see Line 139. So we can dispatch it before draw unit thread's ready.

bjsylvia: Assuming that returning -1 directly maybe cause the main thread is in a busy loop. If returning 0, the current processing logic is the same as returning 1, but semantically it is not consistent with the case where the task is first taken and 1 is returned.

kisvegabor: I was writing a long comment about why I think I'm right, and in the meantime I realized that you are right. :slightly_smiling_face: 

I agree to return 1. 

Just to be sure we are on the same page this is what happens:
1. The render thread is not initialized yet, but `u->task_act` is set in the dispatcher.
2. the dispatcher returns 1 to indicate it taken a task
3. When the render thread will run, it will initialize the sync and due to the code below it will process it immediately:
```c
        while(u->task_act == NULL) {
            lv_thread_sync_wait(&u->sync);
        }
```

nicusorcitu: @bjsylvia how does it look your dispatcher code? Can you please post it here?

bjsylvia: > I was writing a long comment about why I think I'm right, and in the meantime I realized that you are right. 🙂
> 
> I agree to return 1.
> 
> Just to be sure we are on the same page this is what happens:
> 
> 1. The render thread is not initialized yet, but `u->task_act` is set in the dispatcher.
> 2. the dispatcher returns 1 to indicate it taken a task
> 3. When the render thread will run, it will initialize the sync and due to the code below it will process it immediately:
> 
> ```c
>         while(u->task_act == NULL) {
>             lv_thread_sync_wait(&u->sync);
>         }
> ```

Yes. We are on the same page.

bjsylvia: > @bjsylvia how does it look your dispatcher code? Can you please post it here?

Now I am currently only adapting to the operation of software rendering.

bjsylvia:
Is there anything else that needs to be discussed or modified?

src/draw/lv_draw.h
```diff
@@ -119,8 +119,10 @@ typedef struct _lv_draw_unit_t {
      * @return              >=0:    The number of taken draw task:
      *                                  0 means the task has not yet been completed.
```
nicusorcitu, line 120: I do not understand this comment " 0 means the task has not yet been completed."
Which task?

It shall be "0 means the unit is busy with other task" and I will complete with "... or the unit does not support any new task from the list"  - see the explanation below.

nicusorcitu: In the SW dispatcher it it not the case as it supports all the tasks.

But in the GPU dispatcher you will have something like this:
```
    /* Try to get an ready to draw. */
    lv_draw_task_t * t = lv_draw_get_next_available_task(layer, NULL, DRAW_UNIT_ID_GPU);

    /* Return 0 is no selection, some tasks can be supported by SW unit only. */
    if(t == NULL)
        return 0;
```
Returning 0 will allow the SW dispatcher to run, and:
- if task taken it will return 1
- if no more tasks available, it will return -1 and will break the while loop.

bjsylvia: > In the SW dispatcher it it not the case as it supports all the tasks.
> 
> But in the GPU dispatcher you will have something like this:
> 
> ```
>     /* Try to get an ready to draw. */
>     lv_draw_task_t * t = lv_draw_get_next_available_task(layer, NULL, DRAW_UNIT_ID_GPU);
> 
>     /* Return 0 is no selection, some tasks can be supported by SW unit only. */
>     if(t == NULL)
>         return 0;
> ```
> 
> Returning 0 will allow the SW dispatcher to run, and:
> 
> * if task taken it will return 1
> * if no more tasks available, it will return -1 and will break the while loop.

I have deleted error `break` sentence. Please just return -1 when there are no tasks to be processed. You can check it again.

nicusorcitu: Oh, my bad about the -1. Sorry.

But still, what shall the GPU dispatcher return in the above code?

bjsylvia: return -1

src/draw/sw/lv_draw_sw.c
```diff
@@ -148,6 +149,8 @@ static void render_thread_cb(void * ptr)
         /*The draw unit is free now. Request a new dispatching as it can get a new task*/
         lv_draw_dispatch_request();
     }
+    u->inited = false;
+    lv_thread_sync_delete(&u->sync);
```
lhdjply, line 170: I think these two lines of code should be written in 'lv_draw_sw_delete'

bjsylvia:
@kisvegabor I have added the ability for rendering threads to exit, avoiding crashes or memory leaks when the instance exits. Please review it.

src/draw/sw/lv_draw_sw.c
bjsylvia: Let's init and delete `sync` in the same method and thread.

nicusorcitu (approved):

lhdjply (approved):

kisvegabor:
Looks good to me too! I hope it will fix some strange errors with NXP GPUs as well.

PR #4767, opened 2023-11-08
Added custom data to table cells

strijar:
Implemented adding user data to table cells


kisvegabor:
Great feature! Thank you!

kisvegabor:
I've just ported it `master` and noticed that `user_data` is freed automatically. So it assumes that it was malloced by the user. It can be ok, but it should be mentioned in the API comments of `lv_table_set_user_data`.

I've also renamed the `lv_table_set/get_user_data` to `lv_table_set/get_cell_user_data`. If you agree with this, could you send a follow-up PR? (I should notice it while reviewing this PR, sorry)

PR #4768, opened 2023-11-08
Integrity

szepeviktor:
Run "Integrity"
https://github.com/szepeviktor/byte-level-care/blob/master/.github/workflows/reusable-integrity.yml


szepeviktor:
In CI

> Setting up astyle (3.1-2build1) ...

My astyle version is 3.4.10

kisvegabor:
Can we force using a newer astyle in CI?

szepeviktor:
Not from Ubuntu repos
https://packages.ubuntu.com/search?keywords=astyle&searchon=names&exact=1&suite=all&section=all
https://tracker.debian.org/pkg/astyle

I develop a workflow that compiles it from source.

szepeviktor:
This is not a real PR.
I redo it to contain only pre-commit fixes.

szepeviktor:
@kisvegabor We have a problem.
pre-commit config runs astyle this way `entry: astyle --options=scripts/code-format.cfg --ignore-exclude-errors`
this contradicts `scripts/code-format.py`

Please decide what to do.

kisvegabor:
Are there any problems with using `--ignore-exclude-errors` on both places?

szepeviktor:
> Are there any problems with using `--ignore-exclude-errors` on both places?

That is present at both places.
The difference is in paths:
https://github.com/lvgl/lvgl/blob/8328c579c6cdddd25e4bc3bb39d957a364205249/scripts/code-format.py#L5-L15

kisvegabor:
Let me try to better understand it. So now when I commit something `pre-commit` runs. It uses and older version of `astyle`. `code-format-py` uses `astyle` from source, so sometimes I see formatting issues only when the changes are pushed. (Except is I run `code-format.py` locally)

Is there a way to make `pre-commit` use `code-format.py` directly? I think it would be the most unified solution.

szepeviktor:
I was also thinking about this problem.
My conclusion is: when pre-commit is configured to check changed files only it should be able to pick files, when astyle checks all files it picks files on its own.

My suggestions
1. always check all files
2. configure astyle (currently `scripts/code-format.cfg`) to pick files without a python script

This ways astyle can run stand-alone
- from pre-commit
- from CI
- from the command line

kisvegabor:
Let's try running code-format.py on all files. `tests/builds_*` also needs to be excluded and ideally we should hide astlye's log when it was run by pre-commit.

kisvegabor:
I close this issue because there was no activity here for a while.

If you have remarks about this topic feel free to comment here and reopen the issue.

PR #4770, opened 2023-11-09
fix(rtthread):too few arguments to function 'rt_sem_create'

lhdjply:
When using OS, due to looping in the following code .

In lv_draw.c line258
```c
        if(layer_ok) {
            /*Find a draw unit which is not busy and can take at least one task*/
            /*Let all draw units to pick draw tasks*/
            lv_draw_unit_t * u = _draw_info.unit_head;
            while(u) {
                if(taken_cnt < 0) {
                    break;
                }
                if(taken_cnt > 0) one_taken = true;
                u = u->next;
            }
        }
```
When lvgl is created through threads with a priority higher than "rended_thread_cb", it will result in sync not being received.

### DIFF
--- a/src/draw/sw/lv_draw_sw.c
+++ b/src/draw/sw/lv_draw_sw.c
@@ -80,6 +80,10 @@ static int32_t lv_draw_sw_delete(lv_draw_unit_t * draw_unit)
 {
 #if LV_USE_OS
     lv_draw_sw_unit_t * draw_sw_unit = (lv_draw_sw_unit_t *) draw_unit;
+
+    lv_thread_sync_delete(&draw_sw_unit->sync);
+    lv_thread_sync_delete(&draw_sw_unit->sync_ok);
+
     return lv_thread_delete(&draw_sw_unit->thread);
 #else
     LV_UNUSED(draw_unit);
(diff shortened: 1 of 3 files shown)


kisvegabor:
I don't perfectly understand how it solves the priority issue. 

Could you add some comment in the code to explain why wait and signal and what scenarios could happen?

cc @nicusorcitu @bjsylvia

nicusorcitu:
I do not see the reason of using another signal.

lhdjply:
Because sending sync is used in a loop, if the priority of sending sync is higher than receiving sync, it will result in entering the sending state again after sending sync. So after sending sync, entering a waiting state can ensure that sync reception is normal.

nicusorcitu:
This commit shall fix your problem:
https://github.com/lvgl/lvgl/pull/4764/commits/36a071356ac3272c9885186828ec02d8e554de99

lhdjply:
I tried it, and his method can also solve my problem. It's just a different usage method. Sorry, I didn't see anyone mention this PR already

lhdjply:
Since @ bjsylvia has already submitted this PR, let's use his method. Then I'll fix the missing parameters in rtthread

kisvegabor:
As the updated PR contains only RT-Thread related fixes, I merge it.

Let's continue  the sync related discussion in #4764

PR #4795, opened 2023-11-12
Added x11 driver (display/input) support

gcopoix:
Added X11 support (by XLib) for simulating the display content in a Linux desktop window.

This is an alternative for SDL library on a linux system.
Supports
- window resizing
- close event handling (can set close callback to cleanup application properly)
- easy kb/mouse inputs creation (only 1 call to lv_x11_inputs_create() required)
- internal lv_tick_inc handling

to enable the simulation as easy as possible.

Tested in my VSCode CMake project: https://github.com/gcopoix/lv_port_pc_vscode/tree/feature/v9

### Checkpoints

### DIFF
--- a/docs/integration/driver/display/X11.rst
+++ b/docs/integration/driver/display/X11.rst
@@ -1,0 +1,113 @@
+=======
+X11
+=======
+
+Overview
+-------------
+
+The `X11 display/input driver <https://github.com/lvgl/lvgl/src/dev/x11>`__ offers support for simulating the LVGL display
+in a X11 desktop window. It is an alternative to Wayland, XCB, SDL or Qt.
+
+The main purpose for this display driver is for testing/debugging the LVGL application in a Linux simulation window.
+
+
+Prerequisites
+-------------
+
+The X11 display driver uses XLib to access the linux window manager.
+
+1. Install XLib: ``sudo apt-get install libx11-6`` (should be installed already)
+2. Install XLib development package: ``sudo apt-get install libx11-dev``
+
+
+Configure X11 display driver
+----------------------------
+
+1. Enable the X11 display driver support in lv_conf.h, by cmake compiler define or by KConfig
+    .. code:: c
+
+        #define LV_USE_X11  1
+
+2. Optional configuration options:
+    -  Double buffering
+        .. code:: c
+
+            #define LV_X11_DOUBLE_BUFFER  1 /*preferred default*/
+            // or
+            #define LV_X11_DOUBLE_BUFFER  0 /*not recommended*/
+
+    - Render mode
+        .. code:: c
+
+            #define LV_X11_RENDER_MODE     LV_DISPLAY_RENDER_MODE_PARTIAL /*preferred default*/
+            // or
+            #define LV_X11_RENDER_MODE     LV_DISPLAY_RENDER_MODE_DIRECT  /*not recommended*/
+            // or
+            #define LV_X11_RENDER_MODE     LV_DISPLAY_RENDER_MODE_FULL    /*not recommended*/
+
+Usage
+-----
+
+The minimal initialisation (e.g. in main.c):
+
+.. code:: c
+
+    int main(int argc, char ** argv)
+    {
+        ...
+
+        /* initialize X11 display driver */
+        lv_disp_t * disp = lv_x11_window_create("LVGL X11 Simulation", monitor_hor_res, monitor_ver_res);
+
+        /* initialize X11 input drivers (for keyboard, mouse & mousewheel) */
+        lv_x11_inputs_create(disp, NULL);
+
+        ...
+
+        while(true)
+        {
+            ...
+
+            /* Periodically call the lv_task handler */
+            lv_task_handler();
+        }
+    }
+
+Full initialisation with mouse pointer symbol and application close callback
+
+.. code:: c
+
+    bool terminated = false;
+
+    static void on_close_cb(lv_event_t * e)
+    {
+        ...
+
+        terminate = true;
+    }
+
+    int main(int argc, char ** argv)
+    {
+        ...
+
+        /* initialize X11 display driver */
+        lv_disp_t * disp = lv_x11_window_create("LVGL X11 Simulation", monitor_hor_res, monitor_ver_res);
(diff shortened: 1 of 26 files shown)


src/dev/x11/lv_x11.h
```diff
@@ -0,0 +1,95 @@
...
+ * @return             pointer to the display object
+ */
+lv_display_t * lv_x11_window_create(char const * title, int32_t hor_res, int32_t ver_res);
+
+/**
+ * set optional application callback to get informed on X11 window close event (to cleanup application)
+ * @param[in] disp      the created X11 display object from @lv_x11_window_create
+ * @param[in] close_cb  callback funtion to be called
+ */
+void lv_x11_window_set_close_cb(lv_display_t * disp, lv_x11_close_cb close_cb, void * user_data);
```
kisvegabor: Can we use an `LV_EVENT_DELETED` event instead of a dedicated callback?

kisvegabor (commented):
Very nice, just a minor comment above. 

At some point we need to get started with the documentation of these drivers as well. 

So I wonder if you would like to be the pioneer of driver docs by adding these sections to an X11 page [here](https://docs.lvgl.io/master/integration/driver/index.html):
- Overview: What X11 is, target platform, installation
- Usage: What function we have
- Example: A basic example
Could you add

src/dev/x11/lv_x11.h
gcopoix: LV_EVENT_DELETED as I've understood is to delete objects (e.g. widgets?).
lv_x11_window_set_close_cb() callback is [triggered](https://github.com/gcopoix/lvgl/blob/feature/x11_driver/src/dev/x11/lv_x11_display.c#L256-L263) from the window manager close event. The idea is to give the application the possibility to close the window (and its objects) properly (instead of "terminated" by the display driver). In the [example referenced](https://github.com/gcopoix/lv_port_pc_vscode/blob/feature/v9/main/src/main.c#L119-L130) I leave the main loop after this [event](https://github.com/gcopoix/lv_port_pc_vscode/blob/feature/v9/main/src/main.c#L53) and call [lv_display_remove()](https://github.com/gcopoix/lv_port_pc_vscode/blob/feature/v9/main/src/main.c#L129) to properly shutdown the display object, so the application can exit properly.

So it is basically a question of order: The display drivers knows first if the window has been shutdown, but the main application should be responsible for handling the display shutdown (and other things to properly end the application).
I haven't found any usage of LV_EVENT_DELETED, did you mean LV_EVENT_DELETE (without 'D')?
Here the same question: If the display driver fires/handles this event, the main would still be the responsible anchor to handle the shutdown.

What do you think?

kisvegabor: This is how I handled it in the SDL driver:
1. I registered an [`LV_EVENT_DELETE` callback](https://github.com/lvgl/lvgl/blob/master/src/dev/sdl/lv_sdl_window.c#L89) on the display for clean up. So when `lv_display_remove` is called either by the user or when the window is closed the display will clean up properly.  
2. Remove the display in the Window close event. See [here](https://github.com/lvgl/lvgl/blob/master/src/dev/sdl/lv_sdl_window.c#L227-L228).
 
 I think the same can be applied for X11 too. 
 
 If you would like to terminate the main loop as well, you can register an callback for the display when you have created, and set a global variable in it to tell main loop to stop.
 
 What do you think?
 
 > I haven't found any usage of LV_EVENT_DELETED, did you mean LV_EVENT_DELETE (without 'D')?
 
 Sorry, yes, I  meant `LV_EVENT_DELETE`.

gcopoix: 1. rebased on current master
2. added initial documentation in docs/integration/driver/display tree
3. replaced own close_cb with LV_EVENT_DELETE callback

What's wrong with the pipeline (Micropython CI / Build unix port (pull_request)) ?

docs/integration/driver/display/X11.rst
```diff
@@ -0,0 +1,113 @@
...
+        /* initialize X11 display driver */
+        lv_disp_t * disp = lv_x11_window_create("LVGL X11 Simulation", monitor_hor_res, monitor_ver_res);
+        lv_display_add_event(disp, on_close_cb, LV_EVENT_DELETE, disp);
+
+        /* initialize X11 input drivers (for keyboard, mouse & mousewheel) */
+        LV_IMG_DECLARE(my_mouse_cursor_icon);
+        lv_x11_inputs_create(disp, &my_mouse_cursor_icon);
+
+        /* set window close callback to terminate application properly */
+        lv_x11_window_set_close_cb(disp, on_close_cb, disp);
```
kisvegabor: It shouldn't be required as we have ` lv_display_add_event(disp, on_close_cb, LV_EVENT_DELETE, disp);`.

docs/integration/driver/display/index.rst
```diff
@@ -6,3 +6,4 @@ Display
     :maxdepth: 2
 
     ili9341
+    X11
```
kisvegabor: As it's not only a display, let's move it one level up ([here](https://github.com/lvgl/lvgl/tree/master/docs/integration/driver)). Just add the `X11.rst` next to the display and indev folders.

src/dev/x11/lv_x11.h
kisvegabor: > What's wrong with the pipeline (Micropython CI / Build unix port (pull_request)) ?

It's unrelated, I hope it will be resolved soon.

kisvegabor (commented):
Thank you! It looks great! Just a few minor comments.

docs/integration/driver/display/X11.rst
gcopoix: I've added LV_X11_DIRECT_EXIT option, so the exit from the display driver can be used as option.
Usage described in [X11.rst](https://github.com/gcopoix/lvgl/blob/feature/x11_driver/docs/integration/driver/X11.rst).
 I would like to keep this option, so the application can have the possibility to cleanup on exit (and not always  terminated by display driver)

docs/integration/driver/display/index.rst
gcopoix: done

kisvegabor:
Looks great! I think this documentation will be a great reference!

Thank you!

PR #4796, opened 2023-11-13
chore(image): remove explicit initialization to always_zero

XuNeo:

`lv_image_dsc_t` is used both as decoded image and raw C-array image. It's not easy to separate them. This PR only now only modify several minor issues.

- for const lv_image_dst_t, no need to clear always_zero
- decoder info_cb already clears header to zero.
- obj constructor will clear all memory to zero.
- manually init struct variable to { 0 }
    > lv_image_dsc_t img_dsc = { 0 };
- add const  for C-array images.

~~Bin image header may include field that a general image not needed.
For example, BIN image data could be further compressed using RLE or zLib etc. that needs additional flags to indicate compression method.~~

~~A general image header however doesn't need those information and can be reused by lv_draw_buf_t.~~
```c

typedef struct {
    lv_color_format_t cf;
    uint16_t flag;
    uint32_t w;
    uint32_t h;
    uint32_t stride;
} lv_image_header_t;

typedef struct {
    lv_image_header_t header;
    uint32_t flag;
    uint8_t align;
    void * buf;
} lv_draw_buf_t;


typedef struct {
    uint32_t cf : 5;          
    uint32_t always_zero : 3; 
    uint32_t reserved_1 : 4;
    uint32_t compression: 4; /* 0: none, 1: lvgl RLE, 2: zlib, 3: lz4, others: reserved. */
    uint32_t flag: 16;

    uint32_t w: 16;
    uint32_t h: 16;
    uint32_t stride: 16;    
    uint32_t reserved_2: 16; 
} lv_bin_image_header_t;
```

### Checkpoints
 

XuNeo:
`always_zero` has been taken into consideration. 
However I realized it's not possible to do such separation because `lv_image_dsc_t` can also be used as input image to decoder, like lv_canvas.

Ideally we could distinguish between a non-decoded image and decoded image. But looks like it involves too much modules.

I have updated the PR to include only minor naming issues found.
- for const lv_image_dst_t, no need to clear `always_zero`
- decoder_info_cb already clears `header` to zero.
- manually init struct variable to { 0 }
   > lv_image_dsc_t img_dsc = { 0 };
- add `const ` for C-array images.

docs/overview/img.rst
```diff
@@ -134,7 +134,6 @@ variable to display it using LVGL. For example:
    uint8_t my_img_data[] = {0x00, 0x01, 0x02, ...};
 
    static lv_image_dsc_t my_img_dsc = {
-       .header.always_zero = 0,
```
kisvegabor: I would keep it here.

kisvegabor (commented):
It could be really complicated so let's proceed step by step. The current changes look good, just a minor note.

kisvegabor:
Cool, thank you!

PR #4809, opened 2023-11-15
fix(draw): remove LV_LAYER_MAX_MEMORY_USAGE

bjsylvia:
Fix busy loop that maybe caused by the layer task and smaller `LV_LAYER_MAX_MEMORY_USAGE` than needed.

You can reproduce this exception by keeping `LV_LAYER_MAX_MEMORY_USAGE` at the default setting of 150 (kb) and running the transform demo with the window size 640x480.

### Checkpoints
 

kisvegabor:
So it will ignore `LV_LAYER_MAX_MEMORY_USAGE` for the first layer. However there are cases when it doesn't help. Imagine that
1. `LV_LAYER_MAX_MEMORY_USAGE` is 150 kB 
1. We have a 150x200 widget with (ARGB) = 150x200x4 = 120kB, and it's rotated (120kB is used)
1. This widget has a 100x100 child which is also rotated requiring 100x100x4 = 40kB.
1. We have allocated 120kB for the first widget, but in order to merge that layer we need to finalize the 100x100 child too, so in total we need 160kB.
1. And infinite loop.

You suggested solution really helps in a lot of cases, however the proper solution would be to have some kind of dependency discovery an  always allocate memory when it's necessary. 

Or detect infinite loops somehow and discard some layers or assert.

bjsylvia:
> So it will ignore `LV_LAYER_MAX_MEMORY_USAGE` for the first layer. However there are cases when it doesn't help. Imagine that
> 
> 1. `LV_LAYER_MAX_MEMORY_USAGE` is 150 kB
> 2. We have a 150x200 widget with (ARGB) = 150x200x4 = 120kB, and it's rotated (120kB is used)
> 3. This widget has a 100x100 child which is also rotated requiring 100x100x4 = 40kB.
> 4. We have allocated 120kB for the first widget, but in order to merge that layer we need to finalize the 100x100 child too, so in total we need 160kB.
> 5. And infinite loop.
> 

Yes. You are right. I have a question, how do we suggest users configure the size of `LV_LAYER_MAX_MEMORY_USAGE`?

> You suggested solution really helps in a lot of cases, however the proper solution would be to have some kind of dependency discovery an always allocate memory when it's necessary.
> 

I agree it, but currently I don't have a good idea. Do you have any suggestions?

> Or detect infinite loops somehow and discard some layers or assert.

This will result in abnormal display and a poor user experience.

kisvegabor:
All these are really good  questions and I don't have good answers either. 

So there are 2 main topic:
**How to set the initial value of `LV_LAYER_MAX_MEMORY_USAGE`?**
1. Set it to a 0xffffffff for no limit
2. Set it to LV_MEM_SIZE / 2 if the builtin malloc is used

I vote for 1) as normal out of memory on buffer allocation is handled, and LVGL will keep trying to dispatch. 

**What to do on infinite loops?**
Spinning in the loop without any indications is a bad idea for sure. What we can do is:
1. Disregard `LV_LAYER_MAX_MEMORY_USAGE` and try allocating memory. It can still fail due to out of memory
2. Break the loop after some time/attempts, which will result in incomplete rendering
3. `ASSERT`

Actually it's not that hard to calculate how much memory we need in total. Just iterate through all the layers, if a layer has a `layer->parent`, save the memory required for the parent and grandparent, and so on. So we check all the chains and use the size for the greater. If it's greater than `LV_LAYER_MAX_MEMORY_USAGE` it cannot be rendered. 

So in this scenario:
```
screen -> layer1 (100kB) -> layer2 (500kB)
             -> layer3 (300kB) -> layer4 (400 kB)
```

We need 700 kB. 

Except when `layer1` and `layer3` are already allocated we need 800 kB. So it's gets really complex. Probably it would be better detect is by time. E.g. is there is no progress for 1000 ms, break the loop and report an error.

bjsylvia:
It seems that specifying `LV_LAYER_MAX_MEMORY_USAGE` is not a good idea, as the business UI complexity has too much impact on it. Therefore, limiting the value may cause UI rendering interruptions and errors, which would result in a very poor user experience.

bjsylvia:
Although it is possible to evaluate the usage of the layer buffer during this drawing, is it too costly to introduce such evaluation for every drawing?

bjsylvia:
It seems that specifying `LV_LAYER_MAX_MEMORY_USAGE` is not a good idea, as the business UI complexity has too much impact on it. Therefore, limiting the value may cause UI rendering interruptions and errors, which would result in a very poor user experience. So I also vote for "1) Set it to a 0xffffffff for no limit".

kisvegabor:
There is one case when having a limit is really reasonable. Imagine a widget with 200x300 size, requiring 200x300x4 = 240kB memory. We set `style_opa_layered` to `50%`. However for non transformed widgets LVGL can slice the layer into `LV_DRAW_SW_LAYER_SIMPLE_BUF_SIZE` chunks. Let's say it's set to 24 kB, so the widget can be drawn in 10 chunks (smaller layers). These layers are independent and we can really respect the `LV_LAYER_MAX_MEMORY_USAGE` limit. 


Let use this example:
```
screen -> layer10 (transformed) -> layer12 (chunks with opa)
                                -> layer13 (chunks with opa)
                                -> layer14 (chunks with opa)
                                -> layer15 (transformed) -> layer16 (chunks with opa)
                                                         -> layer17 (chunks with opa) -> layer18 (tranformed)
       -> layer 20 (transformed)
       -> layer 30 (chunks with opa)
       -> layer 31 (chunks with opa)
       -> layer 32 (chunks with opa)
       -> layer 40 (transformed) -> layer 41 (transformed) -> layer 42 (transformed)
```

What we know for sure?
1. Layer10 needs to be allocated, else Layer 12..14 cannot be merged
1. Layer15 needs to be allocated, else Layer 16..17 cannot be merged
1. Layer17 needs to be allocated, else Layer 18 cannot be merged
2. From Layer12..14 at least one need to be allocated to  consume them one by one
3. Layer20 can be done at any time
4. Layer30..32 can be done at any time 
5. Layer 40 needs to be allocated only when Layer 41 is ready
6. Both Layer 41 and 42 needs to be allocated
7. At least on chain from Layer 10, Layer 20 Layer 30, or Layer 40 need to started, but chains can be rendered sequentially. 

I'll use this terminology:
- branch: E.g: Layer 10 -> 15 -> 16, or 10 -> 15 -> 17
- parent layer: E.g. Layer 10 is the parent of 15

So the root of the problem is that there are parent layers. We can say that parent layers need to be allocated no matter what, else the child layer cannot be merged.

However the branches can be handled sequentially. 

What about these rules?
- Try to finish single branch first. (Depth first search)
- Allocate buffers from backwards.   Layer 10 is not required when merging Layer 16 is ready and we need to merge it to to 15.
- If there is free memory try adding layers from the next branch by considering `LV_LAYER_MAX_MEMORY_USAGE`.

kisvegabor:
cc @nicusorcitu

XuNeo:
Hi Gabor,
Just to confirm, we are OK to remove `LV_LAYER_MAX_MEMORY_USAGE ` and rely on `LV_DRAW_SW_LAYER_SIMPLE_BUF_SIZE` right?

kisvegabor:
Yes, but simple layers (opa or blend_mode) shall depend on each other to not fill the whole memory with small chunks. I think it's simple and okay-ish solution that we can improve later.

bjsylvia:
@kisvegabor I have removed `LV_LAYER_MAX_MEMORY_USAGE`. Please review it.

kisvegabor:
Thanks! 
I'll add the the "simple layers (opa or blend_mode) shall depend on each other" later sometime.

Issue #4624, opened 2023-10-04
More life-like benchmark


kisvegabor:
### Introduce the problem

From the current benchmark it's hard to tell the  system's performance with a real UI. Besides it contains to many cases and hard to see the bottlenecks.

### Suggested solution

I suggest renaming the current `lv_demo_benchmark` to `lv_demo_draw_test` and create a new shorter and more targeted benchmark. I can imagine these steps:
 
### Basic tests
1. **Empty screen, change background color**: measure fill rate and also acts as a reference FPS 
2. **Moving a screen sized image** (can be tiled to save flash): it shows the memory read performance too
3. **A single small moving rectangle**: show partial refreshing performance
4. **Multiple small moving rectangles** parallel rendering performance and v8 benchmark compare
5. **Multiple moving RGB565 images** parallel rendering performance and v8 benchmark compare
6. **Multiple moving ARGB8888 images** parallel rendering performance and v8 benchmark compare
7. **A single smaller rotated ARGB8888 image** image transformation performance 
8. **Multiple text moving** text rendering + parallel rendering
9. **A lot of text in a screen size container scrolled** large text rendering performance

### Complex tests
1.  **Some  containers moving with icons and multiple texts**: more real life UI and parallel rendering test
2. **1) + a semi transparent overlay changing color**: show screen size semi-transparent rendering for modals
3. **1) + the widgets have animated opa**: opacity performance
4. **1) + the widgets have animated opa_layer**: layer rendering performance 
5. **1) + widget is rotated** show layer performance with layer rotation
6. **1) but several cards below each other and scroll them**  test scrolling and parallel rendering

### Real life UIs
1. **Watch face** Wallpaper, 3 clock hands, label in the background, moving arcs, animate elements in/out
2. **Widgets demo**

As a result, for each step we should display:
- Time between two render starts (in includes flushing, layout recalculation, etc)
- The same as FPS as this is the perceived performance
- Pure rendering time without flushing
- Time of flushing
- CPU usage from OS, or if not OS from `lv_timer_handler` as now

What do you think?

cc @GorgonMeducer @nicusorcitu

bjsylvia:
Can you also provide some test cases for measuring the performance of parallel rendering feature?

nicusorcitu:
1. That would be indeed an benchmark closer to the real life application. It will be very good to evaluate the performance overall.
[...]

PR #4812, opened 2023-11-15
fix(draw): use byte as stride unit instead of pixel

XuNeo:

For image source that stride is aligned to certain bytes, use pixel as draw unit will fail.

### Checkpoints
 
### DIFF
--- a/src/draw/sw/blend/lv_draw_sw_blend_to_argb8888.c
+++ b/src/draw/sw/blend/lv_draw_sw_blend_to_argb8888.c
@@ -49,6 +49,8 @@ static void lv_color_mix_with_alpha_cache_init(lv_color_mix_alpha_cache_t * cach
 
 LV_ATTRIBUTE_FAST_MEM static inline void blend_non_normal_pixel(lv_color32_t * dest, lv_color32_t src,
                                                                 lv_blend_mode_t mode, lv_color_mix_alpha_cache_t * cache);
+LV_ATTRIBUTE_FAST_MEM static inline void * drawbuf_next_row(const void * buf, uint32_t stride);
+
 /**********************
  *  STATIC VARIABLES
  **********************/
@@ -68,7 +70,7 @@ LV_ATTRIBUTE_FAST_MEM void lv_draw_sw_blend_color_to_argb8888(_lv_draw_sw_blend_
     lv_opa_t opa = dsc->opa;
     const lv_opa_t * mask = dsc->mask_buf;
     int32_t mask_stride = dsc->mask_stride;
-    int32_t dest_stride_px = dsc->dest_stride / 4;
+    int32_t dest_stride = dsc->dest_stride;
 
     lv_color_mix_alpha_cache_t cache;
     lv_color_mix_with_alpha_cache_init(&cache);
@@ -106,7 +108,7 @@ LV_ATTRIBUTE_FAST_MEM void lv_draw_sw_blend_color_to_argb8888(_lv_draw_sw_blend_
                 dest_buf[x] = color32;
             }
 
-            dest_buf += dest_stride_px;
+            dest_buf = drawbuf_next_row(dest_buf, dest_stride);
         }
     }
     /*Opacity only*/
@@ -118,7 +120,7 @@ LV_ATTRIBUTE_FAST_MEM void lv_draw_sw_blend_color_to_argb8888(_lv_draw_sw_blend_
             for(x = 0; x < w; x++) {
                 dest_buf[x] = lv_color_32_32_mix(color_argb, dest_buf[x], &cache);
             }
-            dest_buf += dest_stride_px;
+            dest_buf = drawbuf_next_row(dest_buf, dest_stride);
         }
     }
     /*Masked with full opacity*/
@@ -131,7 +133,7 @@ LV_ATTRIBUTE_FAST_MEM void lv_draw_sw_blend_color_to_argb8888(_lv_draw_sw_blend_
                 dest_buf[x] = lv_color_32_32_mix(color_argb, dest_buf[x], &cache);
             }
 
-            dest_buf += dest_stride_px;
+            dest_buf = drawbuf_next_row(dest_buf, dest_stride);
             mask += mask_stride;
         }
     }
@@ -144,7 +146,7 @@ LV_ATTRIBUTE_FAST_MEM void lv_draw_sw_blend_color_to_argb8888(_lv_draw_sw_blend_
                 color_argb.alpha = LV_OPA_MIX2(mask[x], opa);
                 dest_buf[x] = lv_color_32_32_mix(color_argb, dest_buf[x], &cache);
             }
-            dest_buf += dest_stride_px;
+            dest_buf = drawbuf_next_row(dest_buf, dest_stride);
             mask += mask_stride;
         }
     }
@@ -181,9 +183,9 @@ LV_ATTRIBUTE_FAST_MEM static void rgb565_image_blend(_lv_draw_sw_blend_image_dsc
     int32_t h = dsc->dest_h;
     lv_opa_t opa = dsc->opa;
     lv_color32_t * dest_buf_c32 = dsc->dest_buf;
-    int32_t dest_stride_px = dsc->dest_stride / 4;
+    int32_t dest_stride = dsc->dest_stride;
     const lv_color16_t * src_buf_c16 = (const lv_color16_t *) dsc->src_buf;
-    int32_t src_stride_px = dsc->src_stride / 2;
+    int32_t src_stride = dsc->src_stride;
     const lv_opa_t * mask_buf = dsc->mask_buf;
     int32_t mask_stride = dsc->mask_stride;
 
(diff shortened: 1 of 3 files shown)


kisvegabor:
I was working on the same here: 2a5f9ecb2

Does it solve your issue too?

kisvegabor:
For the sake of simplicity I divide the stride by 2 (for RGB565) or 4  (for XRGB8888) but we can do that as the stride should be always 2^N aligned.

XuNeo:
> Does it solve your issue too?

No, it's failing on image with manually aligned stride. 
I have rebased the code. The principle is that for both src, and dest, treat them all as draw buf and using stride to goto next row.

XuNeo:
Emmm, there's still conflict. Do you want me to hold on until you finish it?

kisvegabor:
Finished :slightly_smiling_face: 
 
> No, it's failing on image with manually aligned stride.

Can you mention an image width, and stride combination which doesn't work? Or event better, could you send an image that I can test width?

XuNeo:
I'm testing with `cogwheel.RGB888.bin`.
It can be generated with commands below. I'm testing with stride aligned to 16bytes. Indexed image won't work for now. I'm checking it.

`./scripts/LVGLImage.py -v --ofmt BIN --cf AUTO  --align 16 ../pngs`
[pngs.zip](https://github.com/lvgl/lvgl/files/13364391/pngs.zip)

kisvegabor:
For RGB888 I used the stride in byte during blending. 

I think the best would be to add all image format tests to `LV_DEMO_RENDER_IMAGE_NORMAL`. Only 4 images can fit on a page so we can create `LV_DEMO_RENDER_IMAGE_1_NORMAL`, `LV_DEMO_RENDER_IMAGE_2_NORMAL`, etc.

I can do this on the weekend but if you have time to deal with it earlier it's would be great. Of course we should use the stride=64 images as it was discussed in #4624. 

FYI, I [added render tests](https://github.com/lvgl/lvgl/tree/master/tests/ref_imgs/draw/render) for RGB565, RGB888, XRGB888, and ARGB8888 targets. So by adding more image format tests to `lv_demo_render` we should have a pretty good coverage.

XuNeo:
Agree to add more render test. I'm currently looking the issue of stride process in bin image decoder, and will prepare another PR when finished.

Once it's done, then another PR for more renderer test. 
I still have other promises on our meeting like adding compression support to bin etc. Please do add the test if you have time.

kisvegabor:
I started to work on the image format tests but realized that, indexed, alpha only, RLE, etc are related to image decoding and not purely image rendering. This way they are the same for all GPUs and renderers.

With this in mind keeping them separated from the rendering tests makes more sense to me.  I tink the current tests with the cogwheel are fine, however probably we should replace the cogwheel with something else which has more colors and less symmetrical. We could use a smaller image too, in order to have more tests on a screen.

While working on it I've noticed that you treat the stride of RGB565A8 differently than I imagined. It seems you use 3 bytes pixel size and calculate the stride from it. However using 3 bytes we cannot navigate in this image format as we have 2 bytes/pixels in the RGB565 part and 1 byte/pixel in he A8 part. 

It's a little bit hacky but what I did in SW render is setting the stride for the RGB565 part, and assume half of it for the A8 part.

XuNeo:
> With this in mind keeping them separated from the rendering tests makes more sense to me. I tink the current tests with the cogwheel are fine, however probably we should replace the cogwheel with something else which has more colors and less symmetrical. We could use a smaller image too, in order to have more tests on a screen.

Agree. Need also add different width such as 127x127 that none-aligned.

How would you provide different configuration combinations in CI test? For example, `LV_BIN_DECODER_RAM_LOAD` or not.


> While working on it I've noticed that you treat the stride of RGB565A8 differently than I imagined. It seems you use 3 bytes pixel size and calculate the stride from it. However using 3 bytes we cannot navigate in this image format as we have 2 bytes/pixels in the RGB565 part and 1 byte/pixel in he A8 part.

I'm not sure about it. I tend to specify stride separately. Because once we introduce YUV format, it will meet same issue. See below struct, where`uv_stride`, `v_stride`, `alpha_stride` could be different.

Instead of converting stride unit from byte to pixel (2 bytes/pixels for map, 1 byte/pixel for alpha), it may be a good idea to say alpha stride is equal to map stride in bytes for now. The python converter script can take care of it. In the end when separate stride parameters are supported, then it's a easy migration.

```
    typedef struct vg_lite_yuvinfo {
        vg_lite_swizzle_t   swizzle;        /* UV swizzle. */
        vg_lite_yuv2rgb_t   yuv2rgb;        /* 601 or 709 conversion standard. */
        uint32_t            uv_planar;      /* UV(U) planar address. */
        uint32_t            v_planar;       /* V planar address. */
        uint32_t            alpha_planar;   /* Alpha planar address. */
        uint32_t            uv_stride;      /* UV(U) stride. */
        uint32_t            v_stride;       /* V stride. */
        uint32_t            alpha_stride;   /* Alpha stride. */
        uint32_t            uv_height;      /* UV(U) height. */
        uint32_t            v_height;       /* V height. */
        void *              uv_memory;      /* The logical pointer to the UV(U) planar memory. */
        void *              v_memory;       /* The logical pointer to the V planar memory. */
        void *              uv_handle;      /* The memory handle of the UV(U) planar. */
        void *              v_handle;       /* The memory handle of the V planar. */
    } vg_lite_yuvinfo_t;
```

kisvegabor:
> Agree. Need also add different width such as 127x127 that none-aligned.

I vote for something smaller and more arbitrary, like, 37x41

> How would you provide different configuration combinations in CI test? For example, LV_BIN_DECODER_RAM_LOAD or not.

Now we are running the unit tests with 2 configurations. The main difference is that one uses stdlib malloc, the other uses tlsf `malloc`, however I also wee that we need more config options for:
- `LV_BIN_DECODER_RAM_LOAD`
- Parallel rendering
- Float support
- Anything else?

We need to check which one of these needs to have their own tests and which can be combined with the others. I thinking about this:
1. no parallel, no os, no float, TLSF: test TLSF based malloc, stdlib malloc is superior (because it support ASAN) so we should use it is the other tests
2. no parallel, pthread, float, stdlib malloc: 
3. 8 threads, pthread,  some other configs, stdlib malloc
4. ? threads, pthread,  some other configs, stdlib malloc

> Instead of converting stride unit from byte to pixel (2 bytes/pixels for map, 1 byte/pixel for alpha), it may be a good idea to say alpha stride is equal to map stride in bytes for now. The python converter script can take care of it. In the end when separate stride parameters are supported, then it's a easy migration.

So e.g. for a 30x20 image, with stride align = 32 the stride for RGB would be 30x2 = 60 -> 64, and we would use the  64 for the A8 part as well, although 32 would be enough?  If so I think it's too wasteful even for a temporary solution. It looks like having 4 bytes per pixel instead of 3, so 33% increase. On a 200x200 image, it's 120kB vs 160kB.

XuNeo:
> Now we are running the unit tests with 2 configurations. The main difference is that one uses stdlib malloc, the other uses tlsf `malloc`, however I also wee that we need more config options for:

I agree with your config combination. We also need to include different stride alignment. Probably `LV_DRAW_BUF_STRIDE_ALIGN` and `LV_DRAW_BUF_ALIGN`.

1. base: no parallel, no os, no float, stdlib malloc, LV_DRAW_BUF_STRIDE_ALIGN = 0
2. base + LV_DRAW_BUF_STRIDE_ALIGN = 64 + float + TLSF
3. base + LV_DRAW_BUF_STRIDE_ALIGN = 64 + float + `LV_BIN_DECODER_RAM_LOAD=1`( need ASAN to verify 2)
4. base + 8 threads


> I vote for something smaller and more arbitrary, like, 37x41
Do you have an candidate to replace current `cogwheel`?



> So e.g. for a 30x20 image, with stride align = 32 the stride for RGB would be 30x2 = 60 -> 64, and we would use the 64 for the A8 part as well, although 32 would be enough? If so I think it's too wasteful even for a temporary solution. It looks like having 4 bytes per pixel instead of 3, so 33% increase. On a 200x200 image, it's 120kB vs 160kB.

RGB565A8 seems hard to follow a consistent definition for stride and bit-per-pixel. Let's see if we can have better options like eliminating this format by converting it to RGB565 + alpha map mask after decoding. So draw unit can process it in a more standard color format, unless GPU specifically wants RGB565A8 via decoder open argument.

As for now, I can firstly make it work with the calculated alpha stride.


For this PR, I tested ARGB8888 binary image with master branch and it seems working now. Is this PR still good to merge?

kisvegabor:
> I agree with your config combination. We also need to include different stride alignment. Probably LV_DRAW_BUF_STRIDE_ALIGN and LV_DRAW_BUF_ALIGN.

Sounds good. Let's discuss it in a separate issue.

> Do you have a candidate test image to replace current cogwheel? Please update it and then new test case can use it.

I will search something tomorrow. 


> RGB565A8 seems hard to follow a consistent definition for stride and bit-per-pixel.

I think we can define our own rules for it, like a custom image format. I believe using RGB stride / 2 for A8 stride is safe for GPUs as well.

> For this PR, I tested ARGB8888 binary image with master branch and it seems working now. Is this PR still good to merge? The idea is to unify all usage of stride and make sure unit is byte to avoid converting from pixel to byte.

Your solution is really more unified and less hacky, so I merge it. :blush:

PR #4815, opened 2023-11-16
fixed lvgl.pc source dir in install step if using FetchContent

gcopoix:
If using LVGL by FetchContent in an own project the binary directory is selected by cmake (prefixed in _deps normally).
This fixes the install step respecting the current binary dir and not the global one.

No checkpoint items applied as it is only a minor change in cmake.

### DIFF
--- a/docs/integration/driver/X11.rst
+++ b/docs/integration/driver/X11.rst
@@ -1,29 +1,29 @@
-=======
-X11
-=======
+=========================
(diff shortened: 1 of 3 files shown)


kisvegabor:
Thank you! 

As these are small changes, I can merge it now, but next time, please open a separate PR for unrelated changes. (i.e. in this case one for the docs update and an other for  lvgl.pc).

PR #4827, opened 2023-11-19
feat(indev): start scroll_throw animation using lv_anim_start

bjsylvia:

### Checkpoints
 
### DIFF
--- a/src/indev/lv_indev.c
+++ b/src/indev/lv_indev.c
@@ -1501,7 +1501,6 @@ static void indev_scroll_throw_anim_cb(void * var, int32_t v)
 {
     LV_ASSERT_NULL(var);
     LV_UNUSED(v);
-
     lv_indev_t * indev = (lv_indev_t *)var;
 
     _lv_indev_scroll_throw_handler(indev);
@@ -1529,11 +1528,12 @@ static void indev_scroll_throw_anim_start(lv_indev_t * indev)
     lv_anim_t a;
     lv_anim_init(&a);
     lv_anim_set_var(&a, indev);
-    lv_anim_set_time(&a, 5000);
-    lv_anim_set_values(&a, 0, 100);
+    lv_anim_set_time(&a, 1024);
+    lv_anim_set_values(&a, 0, 1024);
     lv_anim_set_exec_cb(&a, indev_scroll_throw_anim_cb);
     lv_anim_set_ready_cb(&a, indev_scroll_throw_anim_ready_cb);
     lv_anim_set_deleted_cb(&a, indev_scroll_throw_anim_ready_cb);
+    lv_anim_set_repeat_count(&a, LV_ANIM_REPEAT_INFINITE);
 
     indev->scroll_throw_anim = lv_anim_start(&a);
 }
--- a/src/indev/lv_indev.c
+++ b/src/indev/lv_indev.c
@@ -73,6 +73,18 @@ static void indev_read_core(lv_indev_t * indev, lv_indev_data_t * data);
 static void indev_reset_core(lv_indev_t * indev, lv_obj_t * obj);
 static lv_result_t send_event(lv_event_code_t code, void * param);
 
+static void indev_scroll_throw_anim_start(lv_indev_t * indev);
+static void indev_scroll_throw_anim_cb(void * var, int32_t v);
+static void indev_scroll_throw_anim_ready_cb(lv_anim_t * anim);
+static inline void indev_scroll_throw_anim_reset(lv_indev_t * indev)
+{
+    if(indev) {
+        indev->pointer.scroll_throw_vect.x = 0;
+        indev->pointer.scroll_throw_vect.y = 0;
+        indev->scroll_throw_anim = NULL;
+    }
+}
+
 /**********************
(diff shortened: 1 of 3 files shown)


bjsylvia:
@kisvegabor About `scroll throw` animation, please review it.

src/indev/lv_indev.c
```diff
@@ -1477,3 +1496,44 @@ static lv_result_t send_event(lv_event_code_t code, void * param)
...
+}
+
+static void indev_scroll_throw_anim_start(lv_indev_t * indev)
+{
+    LV_ASSERT_NULL(indev);
+
+    lv_anim_t a;
+    lv_anim_init(&a);
+    lv_anim_set_var(&a, indev);
+    lv_anim_set_time(&a, 5000);
```
kisvegabor: Just to confirm, it's just a large enough value and doesn't have any meaning. If so, I think se should use something larger, like 1,000,000.

kisvegabor (commented):
I haven't tested it yet, but I was thinking about the same concept just with a timer instead of an animation.

The animation might be better as can use the `delete_cb`.

So looks good in general. I'll try it out tomorrow.

src/indev/lv_indev.c
bjsylvia: If the duration is too large, under the current implementation mechanism of the animation, the callback will not be triggered when the value is the same. This will cause a large time interval between two frames of the animation, resulting in a longer animation time and reduced smoothness.

https://github.com/lvgl/lvgl/blob/423c288070ccfb90977529aae425b72419733c77/src/misc/lv_anim.c#L392-L396

We should ensure that the value of "value" changes with each frame during calculation, so that the scrolling animation is smooth. I updated the "value" configuration to better match a duration of 5000.

kisvegabor: We can make it the same as duration. This way it will change on each frame for sure. That is use 1 million for both time and value.

kisvegabor:
Tested and it seems to be working very well! If you agree, please update the animation time and value and I will merge it.

bjsylvia:
https://github.com/lvgl/lvgl/blob/224fb4370333a928088e83657fac32bce969c933/src/misc/lv_anim.c#L230C8-L233

bjsylvia:
Currently, different curves in `lv_anim` use `lv_map` to round the value to the `LV_ANIM_RESOLUTION` range when calculating the value. Therefore, if the value is scaled beyond this range, it cannot be guaranteed that the value of each frame will be different from the previous value. Please see the relevant code below.

https://github.com/lvgl/lvgl/blob/224fb4370333a928088e83657fac32bce969c933/src/misc/lv_anim.c#L230-L233

https://github.com/lvgl/lvgl/blob/224fb4370333a928088e83657fac32bce969c933/src/misc/lv_anim.c#L22

Therefore, the solution of setting the value configuration to 1000000 is not feasible. In order to meet the demand for the maximum duration, I suggest configuring the value as the maximum value of `LV_ANIM_RESOLUTION`, and configuring the duration as the result of value*60. What do you think?

src/indev/lv_indev.c
bjsylvia: Currently, different curves in `lv_anim` use `lv_map` to round the value to the `LV_ANIM_RESOLUTION` range when calculating the value. Therefore, if the value is scaled beyond this range, it cannot be guaranteed that the value of each frame will be different from the previous value. Please see the relevant code below.

https://github.com/lvgl/lvgl/blob/224fb4370333a928088e83657fac32bce969c933/src/misc/lv_anim.c#L230-L233

https://github.com/lvgl/lvgl/blob/224fb4370333a928088e83657fac32bce969c933/src/misc/lv_anim.c#L22

Therefore, the solution of setting the value configuration to 1000000 is not feasible. In order to meet the demand for the maximum duration, I suggest configuring the value as the maximum value of `LV_ANIM_RESOLUTION`, and configuring the duration as the result of value*60. What do you think?

bjsylvia: Or maybe we should change the limit `LV_ANIM_RESOLUTION`.

bjsylvia: @kisvegabor What's your opinion?

kisvegabor: I think we are safe to set `LV_ANIM_RESOLUTION 65536`. And instead of 1000000 use 65535  in scroll animation.

bjsylvia: Ok, I have updated it.

bjsylvia:
@kisvegabor Does CI failure have to do with modifying the `LV_ANIM_RESOLUTION`? I have no idea.

kisvegabor:
Could it be an integer overflow in lv_map?

bjsylvia:
Yes. `LV_ANIM_RESOLUTION` can't larger than 1024...

kisvegabor:
Ah, I see. What if we are using an infinite looping animation with value and time set to 1024?

bjsylvia:
> Ah, I see. What if we are using an infinite looping animation with value and time set to 1024?

Done.

kisvegabor:
Thank you, merged!

PR #4833, opened 2023-11-20
chore(draw_buf): add draw buf struct definition

XuNeo:

Initial proposal of adding `draw_buf` struct.

### Checkpoints
 
### DIFF
--- a/src/draw/lv_draw_buf.h
+++ b/src/draw/lv_draw_buf.h
@@ -15,6 +15,7 @@ extern "C" {
  *********************/
 #include "../misc/lv_area.h"
 #include "../misc/lv_color.h"
+#include "lv_image_buf.h"
 
 /*********************
  *      DEFINES
@@ -23,6 +24,13 @@ extern "C" {
 /**********************
  *      TYPEDEFS
  **********************/
+
+typedef struct {
+    lv_image_header_t header;
+    uint32_t data_size;     /*Total buf size in bytes*/
+    void * data;
+    void * _unaligned;      /*Unaligned address of data*/
+} lv_draw_buf_t;
 
 typedef void * (*lv_draw_buf_malloc_cb)(size_t size, lv_color_format_t color_format);
 
@@ -152,6 +160,49 @@ void lv_draw_buf_clear(void * buf, uint32_t w, uint32_t h, lv_color_format_t col
 void lv_draw_buf_copy(void * dest_buf, uint32_t dest_w, uint32_t dest_h, const lv_area_t * dest_area_to_copy,
                       void * src_buf,  uint32_t src_w, uint32_t src_h, const lv_area_t * src_area_to_copy,
                       lv_color_format_t color_format);
+
+/**
+ * Note: Eventually, lv_draw_buf_malloc/free will be kept as private.
+ *       For now, we use `create` to distinguish with malloc.
+ *
+ * Create an draw buf by allocating struct for `lv_draw_buf_t` and allocating a buffer for it
+ * that meets specified requirements.
+ *
+ * @param w         the buffer width in pixels
+ * @param h
+ * @param cf        the color format for image
+ * @param stride    the stride in bytes for image. Use 0 for automatic calculation based on
+ *                  w, cf, and global stride alignment configuration.
+ */
+lv_draw_buf_t * lv_draw_buf_create(uint32_t w, uint32_t h, lv_color_format_t cf, uint32_t stride);
+
+/**
+ * Destroy a draw buf by free the actual buffer if it's marked as LV_IMAGE_FLAGS_MODIFIABLE in header.
+ * Then free the lv_draw_buf_t struct.
+ */
+void lv_draw_buf_destroy(lv_draw_buf_t * buf);
+
+/**
+ * @todo, need to replace lv_draw_buf_go_to_xy.
+ * Return pointer to the buffer at the given coordinates
+ */
+void * lv_draw_buf_goto_xy(lv_draw_buf_t * buf, uint32_t x, uint32_t y);
+
+/**
+ * As of now, draw buf share same definition as `lv_image_dsc_t`.
+ * And is interchangeable with `lv_image_dsc_t`.
+ */
+
+static inline void lv_draw_buf_from_image(lv_draw_buf_t * buf, const lv_image_dsc_t * img)
+{
+    lv_memcpy(buf, img, sizeof(lv_image_dsc_t));
+}
+
+static inline void lv_draw_buf_to_image(const lv_draw_buf_t * buf, lv_image_dsc_t * img)
+{
+    lv_memcpy(img, buf, sizeof(lv_image_dsc_t));
+}
+
 /**********************
  *      MACROS
  **********************/
--- a/src/draw/lv_draw_buf.h
+++ b/src/draw/lv_draw_buf.h
@@ -189,6 +189,11 @@ void lv_draw_buf_destroy(lv_draw_buf_t * buf);
 void * lv_draw_buf_goto_xy(lv_draw_buf_t * buf, uint32_t x, uint32_t y);
 
 /**
+ * Adjust the stride of a draw buf.
+ */
+lv_draw_buf_t * lv_draw_buf_adjust_stride(const lv_draw_buf_t * src, uint32_t stride);
+
+/**
  * As of now, draw buf share same definition as `lv_image_dsc_t`.
  * And is interchangeable with `lv_image_dsc_t`.
  */
--- a/src/draw/lv_draw_buf.c
+++ b/src/draw/lv_draw_buf.c
@@ -115,6 +115,70 @@ void lv_draw_buf_copy(void * dest_buf, uint32_t dest_w, uint32_t dest_h, const l
                                                       color_format);
 }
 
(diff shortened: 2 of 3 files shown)


XuNeo:
CI depends on #4832. Let ignore the CI failure for now.

src/draw/lv_draw_buf.h
```diff
@@ -24,6 +25,12 @@ extern "C" {
  *      TYPEDEFS
  **********************/
 
+typedef struct {
+    lv_image_header_t header;
+    uint32_t data_size;     /*Total buf size in bytes*/
```
kisvegabor, line 30: Why do we need this?

src/draw/lv_draw_buf.h
```diff
@@ -152,6 +159,70 @@ void lv_draw_buf_clear(void * buf, uint32_t w, uint32_t h, lv_color_format_t col
...
+ * Note: Eventually, lv_draw_buf_malloc/free will be kept as static.
+ *       For now, we use `create` to distinguish with malloc.
+ *
+ * Create an draw buf by allocating struct for `lv_draw_buf_t` and allocating a buffer for it
+ * that meets specified requirements.
+ *
+ * @param w         the buffer width in pixels
+ * @param h
+ * @param cf        the color format for image
+ * @param buf_align the start address alignment for the buffer
```
kisvegabor: I think it  `LV_DRAW_BUF_ALIGN` should be used internally instead of having a dedicated parameter. 

Or are there any advantages of allocating a not align or differently aligned buffer?

kisvegabor (commented):
Feels quite elegant, however I can't tell how complicated will it be to use `lv_draw_buf_t` on all places, instead of using some simpler functions directly. 

Anyway, let's go with it. I definitely see the added value of storing the stride and other data with the buffer, assuming it won't introduce too much boiler plate.

src/draw/lv_draw_buf.h
XuNeo: `data_size` is important when we using the buffer.
E.g. save an image to file system, make an copy of `draw buf`. 
Or, it can be used as double-check before accessing memory.
Or as a debug measure to check crash reasons.

src/draw/lv_draw_buf.h
XuNeo: Two draw units may have different requirements on stride like CPU(1) and GPU(64).
Include `buf_align` provides us best flexibility, and we can also provide API using default `buf_align`.

XuNeo:
> however I can't tell how complicated will it be to use `lv_draw_buf_t` on all places, instead of using some simpler functions directly.

I'm planning to update some modules with latest `draw buf` interface these days. Let's see if more helpers can save us.

src/draw/lv_draw_buf.h
XuNeo: That also reminds me default buf_align should be bound with every draw unit, instead of a global config.

kisvegabor: (I've just noticed that this comment wasn't sent...)

If stride and buffer align are bound to a draw unit, it impossible for a general purpose image decoder to create an image with a stride and buffer align fitting to any draw unit. 

Therefore I think we should use smallest common stride and align in a global way.

src/draw/lv_draw_buf.h
```diff
@@ -152,6 +160,79 @@ void lv_draw_buf_clear(void * buf, uint32_t w, uint32_t h, lv_color_format_t col
...
+ * @param buf      the output of assembled draw buffer
+ * @param w        buffer width in pixels
+ * @param h        buffer height in pixels
+ * @param cf       color format for image
+ * @param flags    flags for image
+ * @param stride   stride in bytes for image. Must match real stride of `buf`.
+ * @param data     data buffer used for image
+ * @return         LV_RESULT_OK if succeeded, otherwise LV_RESULT_INVALID
+ */
+lv_result_t lv_draw_buf_assemble(lv_draw_buf_t * buf,
```
kisvegabor:
```suggestion
lv_result_t lv_draw_buf_init(lv_draw_buf_t * buf,
```

We don't have "assemble" keyword yet.

src/draw/lv_draw_buf.h
XuNeo: It's now possible via decoder args, where each draw unit fill in their own requirement, in this case stride and buffer align.

https://github.com/lvgl/lvgl/blob/3e1d7524dab5da5019382788e7aaa9d26fe4c9f4/src/draw/lv_image_decoder.h#L61-L65

It's probably won't be used. Align them both to say 64byte for all draw unit may suits most cases.
I'm just thinking keeping the parameter but provides another simplified version to create draw buffer.
Let me update the PR later.

FASTSHIFT:
During the process of debugging freetype, I found that the image cache provided by freetype cannot guarantee that the output bitmap meets the width alignment and address alignment requirements. We may need to manage the cache ourselves and use draw buf to uniformly manage the bitmap output by freetype.
The advantage of using the draw buf structure for management is that the image information is created together with the image, and no information will be lost in the context transfer, making it easier to track problems such as misalignment.

src/draw/lv_draw_buf.h
kisvegabor: My approach it not open the door for the user to over complicate things. If we believe a global common align and stride is suitable  let's go with it and solve this problem if really there is a use case.

kisvegabor:
@FASTSHIFT 
The idea is that the `get_bitmap` should return an address and stride aligned bitmap. I think it's even better to use our cache for FreeType, because we have more control over it.

FASTSHIFT:
> @FASTSHIFT The idea is that the `get_bitmap` should return an address and stride aligned bitmap. I think it's even better to use our cache for FreeType, because we have more control over it.

If `get_bitmap` returns the `draw_buf` structure, what do you think is its shortcoming? Is it taking up more memory and introducing construction overhead?

kisvegabor:
Yes, what we loose is the extra memory and the construction overhead for sure. 
What we win is unified stride and address align handling and unified cache usage.

src/draw/lv_draw_buf.h
XuNeo: I have remove it. From current usage in decoder, I have seen no need for it. Let's add when it's really needed.

src/draw/lv_draw_buf.h
XuNeo: Agreed. The `buf_align` has been removed now.

XuNeo:
Hi Gabor, 
I have updated the draw_buf related APIs, please help to review it. 
Thank you.

kisvegabor:
I'm thinking about the `stride` parameter in `lv_draw_buf_create`. I understand that images can have any stride and those stride settings might not be supported by a draw unit, but do we really need to allow manually crafting draw_bufs with any stride value? 

What is the advantage of providing full control over stride, instead of setting the stride automatically from the width and color format.

XuNeo:
> The stride alignment of SW render, VG-Lite, PXP, other? That's I believe having a global stride is cleaner.

SW render, VG-Lite, PXP, other have different alignment requirement, so it's not correct `having a global stride is cleaner`?

Aside from that, the stride parameter is a basic para for draw buf just like w, h. That's why I mentioned the PNG decoder case.

> For PNG decoders, e.g. 100x100 image, the draw buffer alloced for directly decoded data is **400Byte**. If GPU requires 64byte aligned stride, then a 448Byte stride draw buffer is alloced and lv_draw_buf_copy can process possibly by DMA2D.

XuNeo:
> * During drawing it will be decoded to a not stride aligned buffer.

I hope the code here can explain it better.

https://github.com/lvgl/lvgl/pull/4968/files#diff-e78cbe701293b586a1376e5917219514c197244669c224744ee13af8f7055e8fR144-R157

kisvegabor:
> For PNG decoders, e.g. 100x100 image, the draw buffer alloced for directly decoded data is 400Byte. If GPU requires 64byte aligned stride, then a 448Byte stride draw buffer is alloced and lv_draw_buf_copy can process possibly by DMA2D.

I understand that the buffers can be copied with on-the-fly stride alignment, but
1. it require an other buffer -> extra memory. If it was a wallpaper then it could mean MBs of RAM
2. it requires time, especially important as usually the buffers are in external RAM

It would be nice to have solution by design, and keep the buffer copy as a fallback for very special situation. 

In my head I still have the "no room for failure" policy. If we choose a common stride which works for all it can't be messed up. 

There are 2 places where it really matter and not trivial:
1. image decoders: it might be hard to make e.g. libpng return a stride aligned image. However `decoder_info_cb` can tell in advance what the stride will be so the draw_units can decide if it's ok for them or not. Maybe a draw_unit will accept a not stride aligned image too because it knows that will use a buffer_copy internally. But it shouldn't be the default assumption because of the above mention memory and time reasons.  
2. font bitmaps: we can either 
    1. return the bitmap as it is and let the draw_unit make the stride alignment
    2. returned stride aligned bitmap
    
Regarding fonts, now we have `ii.` implemented, but if we have a simple `buf_copy` to align stride, `i.` would be better as we have smaller bitmaps here and some draw units (e.g. SW and PXP) can use the bitmaps without any alignment.

For all other things, like snapshot buffer, canvas buffer, layer buffers, etc, we can use a stride which works for all to not put any limitation on the compatible draw_units.

PR #4834, opened 2023-11-20
feat(decoder): add decoder open args

XuNeo:

Image decoder can use args to determine how to decode the image, like stride alignment requirements, premultiply alpha value to RGB value etc.

### Checkpoints
 
### DIFF
--- a/src/draw/lv_image_decoder.h
+++ b/src/draw/lv_image_decoder.h
@@ -50,6 +50,22 @@ struct _lv_image_decoder_t;
 struct _lv_cache_entry_t;
 
 /**
+ * Image decoder args.
+ * It determines how to decoder an image, e.g. whether to premultiply the alpha or not.
+ * It should be passed to lv_img_decoder_open() function. If NULL is provided, default
+ * args are used.
+ *
+ * Default args:
+ * all field are zero or false.
+ */
+typedef struct _lv_image_decoder_args_t {
+    int32_t frame_id;       /*A not use args*/
+    bool stride_align;      /*Whether stride should be aligned*/
+    bool premultiply;       /*Whether image should be premultiplied or not after decoding*/
+    bool no_cache;          /*Whether this image should be kept out of cache*/
+} lv_image_decoder_args_t;
+
+/**
  * Get info from an image and store in the `header`
  * @param src the image source. Can be a pointer to a C array or a file name (Use
  * `lv_image_src_get_type` to determine the type)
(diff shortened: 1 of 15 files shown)


src/draw/lv_image_decoder.h
```diff
@@ -49,6 +49,22 @@ struct _lv_image_decoder_dsc_t;
...
+/**
+ * Image decoder args.
+ * It determines how to decoder an image, e.g. whether to premultiply the alpha or not.
+ * It should be passed to lv_img_decoder_open() function. If NULL is provided, default
+ * args are used.
+ *
+ * Default args:
+ * all field are zero or false.
+ */
+typedef struct _lv_image_decoder_args_t {
```
kisvegabor, line 61: It should be highlighted that it's not sure that the request will be full filled. E.g. a decoder might not support pre multiplied or stride aligned decoding.

src/draw/lv_image_decoder.h
```diff
@@ -49,6 +49,22 @@ struct _lv_image_decoder_dsc_t;
...
+    int32_t frame_id;       /*A not use args*/
+    bool stride_align;      /*Whether stride should be aligned*/
+    bool premultiply;       /*Whether image should be premultiplied or not after decoding*/
+    bool no_cache;          /*Whether this image should be kept out of cache*/
```
kisvegabor, line 65: What does it mean in the practice e.g. in case of a PNG image?

kisvegabor, line 65: 1. It can be added later too, but for vector image we can specify the target size here as well. 
2. What about a target color format option?

kisvegabor (commented):
I think it's an important update and we will rely on it a lot in the future.

src/draw/lv_image_decoder.h
XuNeo: For PNG decoders support output stride(assume so), it can use the `args` directly. Otherwise, a global helper function may be used to convert stride by 2d-copying to a new buffer.

Similar for `premultiply`.

`no_cache` is a use case for frame animation where every image could only use once. Or a RAM image from video, that updates every frame.

XuNeo: For vector image, I'm afraid more `args` are needed. Let's add them when get there.

If I understand correctly, you mean a `prefered color format`? I'm not sure of how to use it for now. It could be a future feature too.

kisvegabor: Got it.

kisvegabor: `prefered color format` really a better name as it's just a request that might or might not be fulfilled. 

By this option I mean for an RGB565 display the user might want RGB565 decoded image too.

But it's not critical for sure.

kisvegabor:
In general I agree with this update and we can extend it later if needed.

Merging.

PR #4858, opened 2023-11-24
fix(draw_line): use lv_point_precise_t to describe coordinates

FASTSHIFT:

`lv_point_precise_t` should be used to describe the position of the point instead of `lv_value_precise_t`.

### Checkpoints
 
### DIFF
--- a/demos/widgets/lv_demo_widgets.c
+++ b/demos/widgets/lv_demo_widgets.c
@@ -1419,17 +1419,17 @@ static void chart_event_cb(lv_event_t * e)
             lv_draw_triangle_dsc_t tri_dsc;
 
             lv_draw_triangle_dsc_init(&tri_dsc);
-            tri_dsc.p[0].x = (int32_t)draw_line_dsc->p1_x;
-            tri_dsc.p[0].y = (int32_t)draw_line_dsc->p1_y;
-            tri_dsc.p[1].x = (int32_t)draw_line_dsc->p2_x;
-            tri_dsc.p[1].y = (int32_t)draw_line_dsc->p2_y;
-            tri_dsc.p[2].x = (int32_t)(draw_line_dsc->p1_y < draw_line_dsc->p2_y ? draw_line_dsc->p1_x : draw_line_dsc->p2_x);
-            tri_dsc.p[2].y = (int32_t)LV_MAX(draw_line_dsc->p1_y, draw_line_dsc->p2_y);
+            tri_dsc.p[0].x = (int32_t)draw_line_dsc->p1.x;
+            tri_dsc.p[0].y = (int32_t)draw_line_dsc->p1.y;
+            tri_dsc.p[1].x = (int32_t)draw_line_dsc->p2.x;
+            tri_dsc.p[1].y = (int32_t)draw_line_dsc->p2.y;
+            tri_dsc.p[2].x = (int32_t)(draw_line_dsc->p1.y < draw_line_dsc->p2.y ? draw_line_dsc->p1.x : draw_line_dsc->p2.x);
+            tri_dsc.p[2].y = (int32_t)LV_MAX(draw_line_dsc->p1.y, draw_line_dsc->p2.y);
             tri_dsc.bg_grad.dir = LV_GRAD_DIR_VER;
 
             int32_t full_h = lv_obj_get_height(obj);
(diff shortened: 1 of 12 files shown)


demos/widgets/lv_demo_widgets.c
```diff
@@ -1419,17 +1419,17 @@ static void chart_event_cb(lv_event_t * e)
             lv_draw_triangle_dsc_t tri_dsc;
 
             lv_draw_triangle_dsc_init(&tri_dsc);
-            tri_dsc.p[0].x = (int32_t)draw_line_dsc->p1_x;
-            tri_dsc.p[0].y = (int32_t)draw_line_dsc->p1_y;
-            tri_dsc.p[1].x = (int32_t)draw_line_dsc->p2_x;
-            tri_dsc.p[1].y = (int32_t)draw_line_dsc->p2_y;
-            tri_dsc.p[2].x = (int32_t)(draw_line_dsc->p1_y < draw_line_dsc->p2_y ? draw_line_dsc->p1_x : draw_line_dsc->p2_x);
-            tri_dsc.p[2].y = (int32_t)LV_MAX(draw_line_dsc->p1_y, draw_line_dsc->p2_y);
+            tri_dsc.p[0].x = (int32_t)draw_line_dsc->p1.x;
```
XuNeo, line 1422: Let's add an API to deal with this potential type conversion.
lv_point_copy(lv_point_t, lv_point_t);
lv_point_precise_copy(lv_point_precise_t, lv_point_precise_t);

lv_point_copy_from_precise(lv_point_t, lv_point_precise_t);
lv_point_copy_to_precise(lv_point_precise_t, lv_point_t);

XuNeo: Ideally we can hide the `lv_point_t` details and force user to use API to access `x`,`y`.

examples/widgets/canvas/lv_example_canvas_7.c
```diff
@@ -27,10 +27,10 @@ void lv_example_canvas_7(void)
     dsc.width = 4;
     dsc.round_end = 1;
     dsc.round_start = 1;
-    dsc.p1_x = 15;
-    dsc.p1_y = 15;
-    dsc.p2_x = 35;
-    dsc.p2_y = 10;
+    dsc.p1.x = 15;
```
XuNeo, line 30: Add API of `lv_point_set`, `lv_point_precise_set`.

demos/widgets/lv_demo_widgets.c
kisvegabor: > lv_point_copy(lv_point_t, lv_point_t);
lv_point_precise_copy(lv_point_precise_t, lv_point_precise_t);

Do we really need these? `p1 = p2` can do the same.

> lv_point_copy_from_precise(lv_point_t, lv_point_precise_t);
lv_point_copy_to_precise(lv_point_precise_t, lv_point_t);

I agree but  I suggest this API:
```c
lv_point_t lv_point_from_precise(lv_point_precise_t * );
lv_point_precise_t lv_point_to_precise(lv_point_t * );
```

XuNeo: OK. Let's go with adding API only for mismatched types.

XuNeo: My only concern is that `lv_value_precise_t` could be same as `int32_t`. Thus code written for `!LV_USE_FLOAT` could potentially misuse p1 = p2 between `lv_value_precise_t` `lv_value_t`.

And I'm certain this will happen because we have just fixed bugs when deal with `lv_coord_t` where people misuse it with int16_t.

kisvegabor: > My only concern is that lv_value_precise_t could be same as int32_t. Thus code written for !LV_USE_FLOAT could potentially misuse p1 = p2 between lv_value_precise_t lv_value_t.

Can we assume that issues like this will be filtered out by our CI which compiles and tests LVGL with `LV_USE_FLOAT` on and off too?

> And I'm certain this will happen because we have just fixed bugs when deal with lv_coord_t where people misuse it with int16_t.

I know it's just an example, but as you probably know `lv_coord_t` is removed and `int32_t` is used instead.

XuNeo: > Can we assume that issues like this will be filtered out by our CI which compiles and tests LVGL with `LV_USE_FLOAT` on and off too?
> 
> >

Unfortunately no. The faulty code is from application code written by inexperienced engineer.
They have used int16_t instead of lv_coord_t in v8, and causes incompatibility when we removed lv_coord_t and replaced with int32_t in v9.

PR #4860, opened 2023-11-24
feat(draw): add unified NEON acceleration

PeterBee97:
Adding lv_blend_neon.S which handles rgb565, rgb888, xrgb8888 and argb8888 blending to each other with optional mask and opacity. 80 functions are exported in <700 lines thanks to flexible assembly macros. Speed is not measured yet, but should produce results similar to C code with small quantization errors (due to inaccuracy of vrecpe).

Running the lvgl test produces err output like this:
blend_to_argb8888_err
The correct reference image is:
blend_to_argb8888
Also the biggest errors lie within rgb565, which I use the naive method of expanding to 888 before blending instead of the [fast method](https://stackoverflow.com/a/50012418/1999969) C code uses.
### Checkpoints
 

PeterBee97:
Also there is an issue with building. Currently including .S in lvgl.mk will cause it to only compile on arm platform because .S files are compiled unconditionally (lvgl CI is not harmed because it doesn't use lvgl.mk). But there is no assembly-friendly header file that exposes the LV_DRAW_SW_ASM macro for asm file to include. What is the best way to handle this?

kisvegabor:
Fantastic! 
Probably we need an other set of reference images for this, once we will have a CI for NEON.

> Also there is an issue with building. Currently including .S in lvgl.mk will cause it to only compile on arm platform because .S files are compiled unconditionally (lvgl CI is not harmed because it doesn't use lvgl.mk). But there is no assembly-friendly header file that exposes the LV_DRAW_SW_ASM macro for asm file to include. What is the best way to handle this?

Can we just `#include lv_conf_internal.h` and use `#if` around the whole content of the .S file?

PeterBee97:
> Can we just `#include lv_conf_internal.h` and use `#if` around the whole content of the .S file?
True. The problem was with the enums and typedefs in lv_conf_internal.h. Fixed with adding #ifndef __ASSEMBLY__ around those definitions.

src/misc/lv_types.h
```diff
@@ -55,6 +58,8 @@ extern "C" {
  *      TYPEDEFS
  **********************/
 
+#ifndef __ASSEMBLY__
```
kisvegabor, line 62: For future reference please add comment here why it is required and what problem it solves.

src/misc/lv_types.h
```diff
@@ -96,6 +103,8 @@ typedef int32_t lv_intptr_t;
  *      MACROS
  **********************/
 
+#ifndef __ASSEMBLY__
```
kisvegabor: Same here

kisvegabor (commented):
Thanks!

src/misc/lv_types.h
PeterBee97: Removed since this is not necessary

src/misc/lv_types.h
PeterBee97: Done

src/misc/lv_types.h
```diff
@@ -55,6 +58,9 @@ extern "C" {
  *      TYPEDEFS
  **********************/
 
+// Exclude C enum and struct definitions when included by assembly code
```
kisvegabor:
```suggestion
/* Exclude C enum and struct definitions when included by assembly code */
```

GorgonMeducer:
@kisvegabor will start the work in December

kisvegabor:
Amazing! 

@PeterBee97 
I've tested the latest code with the Eclipse project but lv_belnd_neon.S is being compiled regardless of 
`#if LV_USE_DRAW_SW_ASM == LV_DRAW_SW_ASM_NEON`

Probably the value of `LV_DRAW_SW_ASM_NEON` is not recognized in `lv_types.h`.

Do you have any idea?

kisvegabor:
It's compiled even if I do this:
```c

#include "lv_blend_neon.h"
#define LV_DRAW_SW_ASM_NEON 1
#define LV_USE_DRAW_SW_ASM 0

#if LV_USE_DRAW_SW_ASM == LV_DRAW_SW_ASM_NEON
```

PeterBee97:
> Amazing!
> 
> @PeterBee97 I've tested the latest code with the Eclipse project but lv_belnd_neon.S is being compiled regardless of `#if LV_USE_DRAW_SW_ASM == LV_DRAW_SW_ASM_NEON`
> 
> Probably the value of `LV_DRAW_SW_ASM_NEON` is not recognized in `lv_types.h`.
> 
> Do you have any idea?

That's weird...to my knowledge .s file ignores preprocessor macros but .S should not. It works on NuttX with gcc and makefile.

kisvegabor:
Seems like an eclipse issue or bug
- https://www.eclipse.org/forums/index.php/t/210430/
- https://mcuoneclipse.com/2016/05/05/assembly-files-in-eclipse-cdt-projects/

I'll try to fix it by associating *.S as C source file.

As a lot of vendor IDEs are Eclipse based, we need to verify it in a thoroughly. 

Would be possible to put everything in the h file or have asm in a c file?

PeterBee97:
> I just made some research and wonder if we can wrap all the asm code into an `__asm__ ()` block and have C extension.

I just downloaded eclipse and managed to compile it. See #4884

PeterBee97:
> > I just made some research and wonder if we can wrap all the asm code into an `__asm__ ()` block and have C extension.
> 
> I just downloaded eclipse and managed to compile it. See #4884

Well it's not passing the CI...so it may not be a problem with CMake but with toolchain? Both CMake and Makefile projects work on my side. I'm using gcc version 11.4.0 (Ubuntu 11.4.0-1ubuntu1~22.04)

kisvegabor:
Do you mean the code as it is in master was working with Eclipse for you?

PeterBee97:
> Do you mean the code as it is in master was working with Eclipse for you?
> 
> If so was it a cmake project or a managed build?

Yes as a cmake project. With #4884 it will be compiled by cc and added the assembly flags correctly. The compile command is then like this:
```
/usr/bin/cc -DLV_CONF_INCLUDE_SIMPLE -DLV_LVGL_H_INCLUDE_SIMPLE -D__ASSEMBLY__ -isystem /home/peter/lvgl  -o CMakeFiles/lvgl.dir/src/draw/sw/blend/neon/lv_blend_neon.S.o -c /home/peter/lvgl/src/draw/sw/blend/neon/lv_blend_neon.S
```

And preprocessor directives work properly. I'm using `gcc version 11.4.0 (Ubuntu 11.4.0-1ubuntu1~22.04)`

kisvegabor:
It turned out that problem was that my `lv_conf.h` included `<stdint.h>` without `#ifndef __ASSEMBLY__`. It can be safely removed as it's already included in `lv_conf_internal.h`. 

I just needed to add this to the beginning of the `lv_belnd_neon.S`
```c
#ifndef __ASSEMBLY__
#define __ASSEMBLY__
#endif

#include "lv_blend_neon.h"
#if LV_USE_DRAW_SW_ASM == LV_DRAW_SW_ASM_NEON
```

It was required in both Desktop and STM32 build. Anyway, I think it's fine to have these. If you agree could you update https://github.com/lvgl/lvgl/pull/4884 accordingly?

PeterBee97:
> ```c
> #ifndef __ASSEMBLY__
> #define __ASSEMBLY__
> #endif
> ```

Done.

PeterBee97:
> It turned out that problem was that my `lv_conf.h` included `<stdint.h>` without `#ifndef __ASSEMBLY__`. It can be safely removed as it's already included in `lv_conf_internal.h`.
> 
> I just needed to add this to the beginning of the `lv_belnd_neon.S`
> 
> ```c
> #ifndef __ASSEMBLY__
> #define __ASSEMBLY__
> #endif
> 
> #include "lv_blend_neon.h"
> #if LV_USE_DRAW_SW_ASM == LV_DRAW_SW_ASM_NEON
> ```
> 
> It was required in both Desktop and STM32 build. Anyway, I think it's fine to have these. If you agree could you update #4884 accordingly?

Added.

PR #4865, opened 2023-11-25
fix(text): do not draw letters treated as marker

### LINKED ISSUE #4856: Invisible unicode characters will be displayed
- [x] Read the [FAQ](https://docs.lvgl.io/master/intro/index.html#faq)
- [x] Check the related part of the [Documentation](https://docs.lvgl.io/)
- [x] Update lvgl to the latest version
- [x] Reproduce the issue in a [Simulator](https://docs.lvgl.io/latest/en/html/get-started/pc-simulator.html)

### Describe the bug

Some characters in the unicode code are invisible, They will have some special functions. [example](https://invisible-characters.com)
On mainstream operating systems they will be parsed correctly and not displayed. But LVGL will render them. This will cause strings copied from other operating systems to not be displayed correctly.
### To Reproduce

```c
  lv_obj_t* label = lv_label_create(lv_screen_active());
  lv_obj_set_style_text_font(label, &lv_font_montserrat_48, 0);
  lv_label_set_text(label, "\u200E\u202AHELLO WORLD!\u202C\u200E");
  lv_obj_center(label);
```
### Expected behavior

image
VScode can preview them correctly

### Screenshots or video

image

XuNeo:

Fix #4856

Refer to https://invisible-characters.com/ and double checked with https://www.compart.com/en/unicode/

Definitely not all unicodes are included because it's tricky to say which unicode is `invisible`.

This check could potentially affect rendering speed in case of long text to show.

### Checkpoints
 
### DIFF
--- a/src/draw/lv_draw_label.c
+++ b/src/draw/lv_draw_label.c
@@ -100,6 +100,8 @@ LV_ATTRIBUTE_FAST_MEM void lv_draw_letter(lv_layer_t * layer, lv_draw_label_dsc_
         return;
     }
 
+    if(_lv_text_is_marker(unicode_letter)) return;
+
     lv_font_glyph_dsc_t g;
     lv_font_get_glyph_dsc(dsc->font, &g, unicode_letter, 0);
 
@@ -273,8 +275,6 @@ void lv_draw_label_iterate_letters(lv_draw_unit_t * draw_unit, const lv_draw_lab
             uint32_t letter;
             uint32_t letter_next;
             _lv_text_encoded_letter_next_2(bidi_txt, &letter, &letter_next, &i);
-            if(_lv_text_is_marker(letter)) /*Markers are valid letters but should not be rendered.*/
-                continue;
 
             letter_w = lv_font_get_glyph_width(font, letter, letter_next);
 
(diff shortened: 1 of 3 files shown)


kisvegabor:
Am I correct to assume that these invisible characters should look like as they weren't there at all? It means if there is a text like `a_b` (`_` is an invisible character) and the letter space is 10 px, only 10px space should be added between `a` and `b`, right? So it's different from a normal "space" character, where letter space should be drawn of both sides. 

Therefore I think invisible is not a best term as all kind of "space" and "tab" are also invisible.  I think "marker" characters or so, might be a more descriptive term.

Shouldn't `< 0x20` characters belong to this category too?

XuNeo:
`marker` is better.
I have included letters look like these. There's no real representation of the letter, and I think they are `markers`.
image


> Shouldn't `< 0x20` characters belong to this category too?

Yes, I haven't included yet, because it's wrong to include them in text. They are different with other unicodes that are valid `marker`s.

kisvegabor:
For consistency I wonder if the same shall be applied [here](https://github.com/lvgl/lvgl/blob/master/src/misc/lv_text.c#L353) as well.

Or we might rely on  `width==0` to detect markers in general, which is more generic and faster.

XuNeo:
> Or we might rely on `width==0` to detect markers in general, which is more generic and faster.

For those markers, they are not available in font. Will it return `width == 0`?

kisvegabor:
Good point, no necessarily, however we can ensure it on a central place, [here](https://github.com/lvgl/lvgl/blob/master/src/font/lv_font.c#L89). This block could be moved to the top of the function too. 

What do you think?

XuNeo:
Hi Gabor, 

Put the logic of check letter type to `lv_font` doesn't look right. If letter is not found in font, it show a box placeholder instead. 
So the PR has been updated to make sure the `letter` to `draw` is not marker instead.

kisvegabor:
Makes sense, looks good, merging, thanks. :slightly_smiling_face:

Issue #4869, opened 2023-11-26
`lv_image_dsc_t->data` could not meet buf align requirement


XuNeo:
### Introduce the problem

Current `lv_image_dsc_t` definition includes a simple data pointer for all kinds of images.
For `Indexed2` image, `palette` takes `4 x sizeof(lv_color32_t) = 16Bytes`.
It makes the image data following palette cannot meet requirement of `64byte` alignment.

It also makes decoding 8bit PNG a little more complicated. We have to alloc a new contiguous memory for pallete + img_data.

For RGB565A8 image, we cannot describe the stride for alpha map inside.

```c
typedef struct {
    lv_image_header_t header; /**< A header describing the basics of the image*/
    uint32_t data_size;     /**< Size of the image in bytes*/
    const uint8_t * data;   /**< Pointer to the data of the image*/
} lv_image_dsc_t;
```

### Suggested solution

Depending on the color format of the image, using different struct to describe data.
For example, an indexed image should use `indexed` field.

```c

typedef struct {
    lv_image_header_t header;           /*A header describing the basics of the image*/
    uint32_t data_size;                 /*Size of the data in bytes*/
    union {
        const uint8_t * data;           /*Uncompressed normal image data*/

        struct {
            uint32_t compress_method: 4;
            uint32_t flags: 4;
            uint32_t uncompress_size: 24;
            const uint8_t * compressed; /*Compressed raw data*/
        };

        struct {                        /*Special image of indexed format*/
            const uint8_t * palette;
            const uint8_t * data;
        } indexed;

        struct {                        /*Special image of RGB565A8 format*/
            const uint8_t * rgb;
            const uint8_t * alpha;      /*Alpha map of RGB565A8*/
            uint32_t alpha_stride;      /*Stride for the alpha map in bytes*/
        } rgb565a8;
    };
} lv_image_dsc_t;

```

kisvegabor:
I would approach it in a more generic way. What about
```c
typedef struct {
    lv_image_header_t header;           /*A header describing the basics of the image*/
    uint32_t data_size;                 /*Size of the data in bytes*/
    const void * buf;
    const void * buf_sup;           /*Optional supplementary buffer*/
}
```

Regarding `compressed`
```c
            uint32_t compress_method: 4;   /*Can we have it in lv_image_header_t*/
            uint32_t flags: 4;                            /*Can we have it in lv_image_header_t*/
[...]

PR #4870, opened 2023-11-26
Add RLE compressed binary image support

XuNeo:

Add flag to image header to indicate the data is compressed.
Separate definition of `data`, using union to include compressed image data struct.

Added test for RLE compressed image.

Update python script to support RLE compressed image.
Usage:
```
  --compress {NONE,RLE}
                        Binary data compress method, default to NONE
```

### Checkpoints
 

XuNeo:
Hi Gabor,
The CI reports error but I cannot reproduce locally. Could you help me on how to reproduce it?

kisvegabor:
It fails for me locally too. This is how image_format_rle_compressed_err.png looks like:  
image_format_rle_compressed_err

It seems get_info is not working as the size not set.

src/draw/lv_image_buf.h
```diff
@@ -113,9 +126,20 @@ typedef struct {
...
-    const uint8_t * data;   /**< Pointer to the data of the image*/
+    union {
+        const uint8_t * data;   /**< Pointer to the data of the image*/
+        struct _lv_image_compressed_t {
+            uint32_t method: 4; /*Compression method, see `lv_image_compress_t`*/
+            uint32_t reserved : 28;  /*Reserved to be used later*/
+            uint32_t compressed_size;  /*Compressed data size in byte*/
+            uint32_t decompressed_size;  /*Decompressed data size in byte*/
+            const uint8_t * data; /*Compressed data*/
+        } compressed;
```
kisvegabor: See my comment [here](https://github.com/lvgl/lvgl/issues/4869#issuecomment-1829425792).

XuNeo: > In an other issue we were talking about having an lv_image_dsc_t for the compressed/original image, and an other lv_draw_buf_t for the decoded image. Can we assume that the decoder can set the uncompressed data size during decoding? If it's required for the decoder I would rather store in the compressed image's header. I feel like it's a custom data required only for custom formats, so it should not be exposed as public.

> @XuNeo One more note: nameless unions are C11 standard we are on C99. We can consider jumping to C11, but it requires serious consideration and research.

You are correct. Let's move the compressed data format inside lv_bin_decoder.c and use C99.

XuNeo: > In an other issue we were talking about having an lv_image_dsc_t for the compressed/original image, and an other lv_draw_buf_t for the decoded image. Can we assume that the decoder can set the uncompressed data size during decoding? If it's required for the decoder I would rather store in the compressed image's header. I feel like it's a custom data required only for custom formats, so it should not be exposed as public.

> @XuNeo One more note: nameless unions are C11 standard we are on C99. We can consider jumping to C11, but it requires serious consideration and research.

You are correct. Let's move the compressed data format inside lv_bin_decoder.c and use C99.

XuNeo:
Hi @kisvegabor,
This PR is ready for review. Could you check it? Thank you.

kisvegabor (commented):
Looks good, thank you!

Issue #4339, opened 2023-07-04
PNG (lodepng) forces you to supply an entire out buffer


codewitch-honey-crisis:
### Perform all steps below and tick them with [x]
- [X] Read the [FAQ](https://docs.lvgl.io/master/intro/index.html#faq)
- [X] Check the related part of the [Documentation](https://docs.lvgl.io/)
- [X] Update lvgl to the latest version
- [ X] (n/a) Reproduce the issue in a [Simulator](https://docs.lvgl.io/latest/en/html/get-started/pc-simulator.html) 

### Describe the bug

It's not really a bug per se, but a performance/resource usage issue, wherein you must - unless I dramatically misunderstand the API - supply an in memory output buffer to hold the decompressed PNG image. This is what I would consider to be undesirable on IoT devices. 

### To Reproduce
Use this code
[https://github.com/lvgl/lvgl/blob/master/examples/libs/png/lv_example_png_1.c](https://github.com/lvgl/lvgl/blob/master/examples/libs/png/lv_example_png_1.c)

### Expected behavior

Well, it would be nice if it would call you back with pixel information in order to draw, but I'm not sure how compatible that would be with the way LVGL works. Pngle would be a good option for this instead of lodepng

kisvegabor:
It would be really great, but our current PNG decoder ([lodepng](https://github.com/lvandeve/lodepng)) doesn't support it. :slightly_frowning_face:

codewitch-honey-crisis:
Pngle does, and would be easy to integrate if I could figure out a way to get LVGL to draw on demand without a backing bitmap. I'm a bit of a noob with LVGL. Edit: I'd be willing to integrate it as a contribution. Just need a bit of guidance regarding the drawing as above.

kisvegabor:
I'm afraid that it will be very slow as Pngle [calls a callback](https://github.com/kikuchan/pngle#generic-c) on each pixel to draw. It should draw at least a line in a callback call.

codewitch-honey-crisis:
That could be easily done. I just don't know how to coerce LVGL to call me back to do a draw. When doing a canvas, it draws to a bitmap. In my own UI lib (which i was sort of forced to write for my graphics lib to support ESP32s) i get an on_paint() callback when a control needs to redraw. I would need something similar in LVGL.

kisvegabor:
[...]

PR #4872, opened 2023-11-26
feat(lib): add lz4 lib support

XuNeo:

Add lz4 lib, so we can compress bin images with lz4.

### Checkpoints
 
### DIFF
--- a/env_support/cmake/custom.cmake
+++ b/env_support/cmake/custom.cmake
@@ -20,6 +20,7 @@ file(GLOB_RECURSE SOURCES ${LVGL_ROOT_DIR}/src/*.c)
 file(GLOB_RECURSE EXAMPLE_SOURCES ${LVGL_ROOT_DIR}/examples/*.c)
 file(GLOB_RECURSE DEMO_SOURCES ${LVGL_ROOT_DIR}/demos/*.c)
 file(GLOB_RECURSE THORVG_SOURCES ${LVGL_ROOT_DIR}/src/libs/thorvg/*.cpp)
+file(GLOB_RECURSE LZ4_SOURCES ${LVGL_ROOT_DIR}/src/libs/lz4/*.c)
 
 # Build LVGL library
 add_library(lvgl ${SOURCES})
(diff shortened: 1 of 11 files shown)


XuNeo:
CI fails because micropython includes all source files under `src/` to compile. But `lz4` doesn't have control of `LV_USE_LZ4`. 
It's a pain to modify all `lz4` sources.

kisvegabor:
It seems this PR and #4873 is about the same feature. Can we close on of them?

kisvegabor:
What is the advantage of LZ4 over PNG?

XuNeo:
> It seems this PR and #4873 is about the same feature. Can we close onw of them?

One is to add the lz4 lib, another is to add lz4 compressed image support.



> What is the advantage of LZ4 over PNG?

I have no benchmark yet. 
The whole idea is to fill the gap between good compression ratio of PNG and good decode speed of BIN. 
PNG is too slow, and BIN is too big. Nothing between until RLE and possibly LZ4.

The compression of BIN image could also perform on line by line, thus could potentially solve issues like #4339 for IOT devices that require high compression ratio and small memory in the same time.

kisvegabor:
I just wanted to point out updating `tests/CMakeLists.txt` but you have already figured it put :slightly_smiling_face: 

Before moving forward:
1. Please add only the really required files. E.g. we probably don't need anything from `dll/examples` or the `md`, `Makefike`, `in` files.
2. Let's do a basic benchmark just to have an idea about the impact of this change. It would be nice to see the size and speed of a 16x16, 100x100, 400x400 image with bin, PNG and LZ4.

kisvegabor:
Cool! 

> Make sure all source files are in a plain folder structure. So a easy add_lvgl_if.sh script can process it.

We can make `add_lvgl_if.sh` work recursively so it's not required to have a flat structure. It would be better to move `add_lvgl_if.sh` it to the `scrips` folder to easily run it for any 3rd part lib without duplicating the script.

env_support/cmake/custom.cmake
```diff
@@ -47,6 +48,14 @@ if(NOT LV_CONF_BUILD_DISABLE_THORVG_INTERNAL)
     endif()
 endif()
 
+if(NOT LV_CONF_BUILD_DISABLE_LZ4_INTERNAL)
```
kisvegabor: This block is about when INTERNAL is not disabled, so internal is enabled. Why do we need any kind of CMake update in this case? Those 2 lz4 files will be `glob`ed  with all the other LVGL files.

kisvegabor (commented):
Only 2 files, very nice!

kisvegabor:
As we have discussed before moving forward, it would be great to see the real advantage of LZ4 over PNG, RLE and plain bin.

env_support/cmake/custom.cmake
XuNeo: You are correct. I have fixed them now.

kisvegabor:
All looks good, merging. Thanks you!

PR #4883, opened 2023-11-28
feat(draw): add global matrix drawing modes

FASTSHIFT:

1. Using a matrix to realize the transform effect of obj can fully utilize the hardware performance that supports 2.5D GPU without going through the layer's draw image operation, which greatly reduces the memory copy overhead.
2. Using the global alpha capability of the GPU, you can achieve drawing with global transparency to bypass the secondary color mixing operation of the layer. Completely solve the performance problems caused by `style_opa_layered`.
3. The matrix can be used to achieve pseudo-3D effects, and menus with 3D effects are no longer a dream ;)

cc @onecoolx 

Demo video:

https://github.com/lvgl/lvgl/assets/26767803/fdb8eab3-876b-4ad2-b70e-a02fa65d910e

### Checkpoints
 

Kconfig
```diff
@@ -174,6 +174,12 @@ menu "LVGL configuration"
 			help
 				Align the start address of draw_buf addresses to this bytes.
 
+			config LV_DRAW_USE_GLOBAL_OPA_AND_MATRIX
```
XuNeo: Will separate the config be better, considering some GPU could have no global alpha support?

FASTSHIFT: Separate configuration may cause the layer logic to become complicated. If the GPU does not support it, the `style_opa_layered` effect will not take effect. Users can be recommended to use `style_opa` to achieve similar effects.

kisvegabor:
It can't be true :smile: 

Okay, to be serious. 
 **global opacity**
> Completely solve the performance problems caused by style_opa_layered.

Let's say we have a layer with having a white background with 100% alpha, a red rectangle with also 100% alpha and blue rectangle having 50% alpha. We know that the layer needs to blended with 20%  alpha, so we set the global alpha to 20%. 
 - When the white background is rendered the 'A' channel of the layer will be 20%. 
 - What happens when the red rectangle is rendered? Will it fully overwrite the white resulting in 20% red only?
 - What happens with the 50% blue? If it partially overlaps red, how will it behave?
 
**matrix**
So matrix the end matrix is more flexible description of scale, rotate, skew, etc. So will still need to have layer, we just tell how to blend it in a more flexible way, right? if so, why don't we put it in `lv_draw_image_dsc_t`?

FASTSHIFT:
> It can't be true 😄
> 
> Okay, to be serious. **global opacity**
> 
> > Completely solve the performance problems caused by style_opa_layered.
> 
> Let's say we have a layer with having a white background with 100% alpha, a red rectangle with also 100% alpha and blue rectangle having 50% alpha. We know that the layer needs to blended with 20% alpha, so we set the global alpha to 20%.
> 
> * When the white background is rendered the 'A' channel of the layer will be 20%.
> * What happens when the red rectangle is rendered? Will it fully overwrite the white resulting in 20% red only?
> * What happens with the 50% blue? If it partially overlaps red, how will it behave?
> 
> **matrix** So matrix the end matrix is more flexible description of scale, rotate, skew, etc. So will still need to have layer, we just tell how to blend it in a more flexible way, right? if so, why don't we put it in `lv_draw_image_dsc_t`?

The layer is indeed very flexible, but it has an obvious disadvantage, that is, each drawing must be drawn to a buffer off the screen (this buffer is often located in PSRAM or SDRAM), and then the buffer is blended to the framebuffer as a picture. This will introduce an additional image rendering overhead.
On embedded devices, due to limited memory bandwidth, large memory copying needs to be avoided as much as possible. The introduction of matrices can achieve real-time rendering transform effects without any performance loss.

FASTSHIFT:
> Aaaaaah, I understand now 😅
> 
> So it's a real time thing without actually creating layers. E.g. a rounded rectangle will be transformed on the fly by the vector GPU.
> 
> In understand the matrix part but the global alpha is still not clear.
> 
> 1. `lv_obj_set_style_opa(arc, 127, 0)`:
>    image
> 2. `lv_obj_set_style_opa_layered(arc, 127, 0)`:
>    image
> 
> Wont global alpha produce an output like 1?

I re-read the vg_lite driver and found that global alpha will only act on image-related rendering. Vector drawing global alpha such as path filling is invalid, so let us still use the original implementation of `style_opa_layered`.

kisvegabor:
Okay, let's focus on the matrix part. I few notes

- `blend_mode` is not considered
-  `layer->matrix` needs to saved in the `draw_task`s because when there is an operating system and the draw tasks are executed later, the saved matrix can be already overwritten for other draw tasks. It's the very same as we need to save the current clip_area of the layer in the  draw_task, and than pass the clip_area to the draw_unit when it's actually rendered. 
- Once we introduce matrix for rendering, what about addig into each `lv_draw_<...>dsc_t`? It seems more flexible. We can say if the matrix is set it will be used, else the normal parameters.

FASTSHIFT:
> Okay, let's focus on the matrix part. I few notes
> 
> * `blend_mode` is not considered
> * `layer->matrix` needs to saved in the `draw_task`s because when there is an operating system and the draw tasks are executed later, the saved matrix can be already overwritten for other draw tasks. It's the very same as we need to save the current clip_area of the layer in the  draw_task, and than pass the clip_area to the draw_unit when it's actually rendered.

Understood.

> * Once we introduce matrix for rendering, what about addig into each `lv_draw_<...>dsc_t`? It seems more flexible. We can say if the matrix is set it will be used, else the normal parameters.

Does this mean that software rendering also needs to introduce matrices? Or should we say that software rendering still follows the old parameter scheme, and the matrix is ​​only used by the GPU?

kisvegabor:
The way how transformations are handled in SW render could work with matrix too. However it should be the subject of a minor version. Now we can just not accept draw tasks by the SW draw unit if they have matrix.

FASTSHIFT:
> layer->matrix needs to saved in the draw_tasks because when there is an operating system and the draw tasks are executed later, the saved matrix can be already overwritten for other draw tasks. It's the very same as we need to save the current clip_area of the layer in the draw_task, and than pass the clip_area to the draw_unit when it's actually rendered.

`layer->matrix` should be read-only for draw tasks. Do we need to consider multi-threading issues?

FASTSHIFT:
We can first consider incorporating a simple first version modification. When `LV_DRAW_TRANSFORM_USE_MATRIX` is turned off by default, it will not affect any previous rendering behavior.

FASTSHIFT:
> * Once we introduce matrix for rendering, what about addig into each `lv_draw_<...>dsc_t`? It seems more flexible. We can say if the matrix is set it will be used, else the normal parameters.

I think `lv_draw_<...>dsc_t `does not need to save matrix, because the two sets of transfrom information are easy to confuse people, and a lot of judgments need to be written to distinguish which one to use. As far as I can tell, `layer->matrix `can meet the needs very well and has the least impact on the framework.

kisvegabor:
> layer->matrix should be read-only for draw tasks. Do we need to consider multi-threading issues?

One draw task is drawn by a single thread so no multi-threading issues should happen

> I think lv_draw_<...>dsc_t does not need to save matrix, because the two sets of transfrom information are easy to confuse people, and a lot of judgments need to be written to distinguish which one to use. As far as I can tell, layer->matrix can meet the needs very well and has the least impact on the framework.

See:
> layer->matrix needs to saved in the draw_tasks because when there is an operating system and the draw tasks are executed later, the saved matrix can be already overwritten for other draw tasks. It's the very same as we need to save the current clip_area of the layer in the draw_task, and than pass the clip_area to the draw_unit when it's actually rendered.

So we need to save the matrix in the draw task anyway, because if there are nested matrices the draw_task will see the latest in the layer and its own.

nicusorcitu:
> I re-read the vg_lite driver and found that global alpha will only act on image-related rendering. Vector drawing global alpha such as path filling is invalid😅, so let us still use the original implementation of style_opa_layered.
>
I can take care and introduce the blend modes for all blit and draw operations. All I need is the upstream support to be there.

FASTSHIFT:
> So we need to save the matrix in the draw task anyway, because if there are nested matrices the draw_task will see the latest in the layer and its own.

What if I save a matrix [here](https://github.com/lvgl/lvgl/pull/5010/files#diff-5581480daafb376387d3e85ecf37216d8b0e35f7726171f1084a05b3a436be60R36)?

Then [here](https://github.com/lvgl/lvgl/pull/5010/files#diff-93d1ca44ca2f05d6fc161f3fd247adae3f60a2ebf5135fe2f408a0e3109a9609R101) we multiply the `global_matrix` and `layer->matrix` in the vg_lite unit:
```c
vg_lite_identity(&u->global_matrix);
vg_lite_translate(-layer->buf_area.x1, -layer->buf_area.y1, &u->global_matrix);
lv_matrix_multiply((lv_matrix_t*)&u->global_matrix, &layer->matrix);
```

kisvegabor:
> What if I save a matrix [here](https://github.com/lvgl/lvgl/pull/5010/files#diff-5581480daafb376387d3e85ecf37216d8b0e35f7726171f1084a05b3a436be60R36)?

I think it won't work. Imagine this:
1. LVGL is about the draw the background which takes a long time. VG-Lite takes this task
2. While VG-Lite is working, LVGL adds A, B, C, D draw_tasks with `matrix_1`, and and E, F, G draw_tasks with `matrix_2`
3. When VG-Lite dispatches the next task, e.g `A` it will see that `matrix_2` is set in the layer as it was the last. 

So we need to preserve the matrix for each draw task as it's always overwritten in the layer.

FASTSHIFT:
Not stale.

kisvegabor:
Not stale

FASTSHIFT:
> So we need to preserve the matrix for each draw task as it's always overwritten in the layer.

Done, let's continue.

FASTSHIFT:
@kdschlosser 

Hi, I seem to have encountered a CI error in mpy. I tried looking through the submission records, but I couldn't find any suspicious clues. Can you help take a look?
image

I noticed a similar problem was encountered in the PR discussion here: https://github.com/lvgl/lvgl/pull/4997

FASTSHIFT:
@PGNetHun 
HI, I encountered a CI error here:
image

Noticed that this modification turns on `LV_USE_LOG`, is it relevant?
https://github.com/lvgl/lv_binding_micropython/pull/298

FASTSHIFT:
> @PGNetHun HI, I encountered a CI error here: image
> 
> Noticed that this modification turns on `LV_USE_LOG`, is it relevant? [lvgl/lv_binding_micropython#298](https://github.com/lvgl/lv_binding_micropython/pull/298)

Fixed: https://github.com/lvgl/lv_binding_micropython/pull/320

XuNeo:
Not stale.

FASTSHIFT:
Close for now and consider implementing this feature later.

nicusorcitu:
@cristian-stoica

kisvegabor:
As now we can test VG-Lite could you add a test for it too? As it's not so trivial it shall be documented too.

If ThorVG could be used as a draw engine having an example would be huge as well.

FASTSHIFT:
> As now we can test VG-Lite could you add a test for it too? As it's not so trivial it shall be documented too.
> 
> If ThorVG could be used as a draw engine having an example would be huge as well.

No problem, we can merge https://github.com/lvgl/lvgl/pull/6264 first, and then I'll add new test cases based on the mainline.

kisvegabor:
I've just merged https://github.com/lvgl/lvgl/pull/6264

FASTSHIFT:
ThorVG rendering crash found in automated testing: https://github.com/thorvg/thorvg/issues/2391

FASTSHIFT:
Since the GPU does not support outputting images in AL88 and L8 color formats, it will fallback to software rendering. Software rendering does not support real-time transformation of matrices.
Therefore, when `LV_DRAW_TRANSFORM_USE_MATRIX` is enabled, using the framebuffer in AL88 and L8 color formats will cause the transform style to be invalid.

FASTSHIFT:
Since merge conflicts may occur, please merge this PR first: https://github.com/lvgl/lvgl/pull/6344.

FASTSHIFT:
We tested it on hardware and used the global matrix solution to draw the transform style. We can get nearly **5 times** the performance improvement, and since there is no need to create a temporary drawing buffer for the layer, the memory usage is also greatly reduced (especially for larger widgets).

@kisvegabor @liamHowatt 
Please help review this PR~

src/misc/lv_matrix.h
```diff
@@ -0,0 +1,125 @@
...
+
+/*********************
+ *      INCLUDES
+ *********************/
+
+#include "../lv_conf_internal.h"
+
+#if LV_USE_MATRIX
+
+#include <stdbool.h>
```
XuNeo: use lv_types.h instead.

FASTSHIFT: done

XuNeo (dismissed):

tests/ref_imgs_vg_lite/widgets/obj_flag_overflow_visible_1_4.png
kisvegabor, line 1: Overflow visible is not handled properly here. The purple square should clip the blue button as on the original image.

tests/ref_imgs_vg_lite/draw/render/al88/demo_render_layer_normal_opa_128.png
kisvegabor, line 1: These images are incorrect.

tests/ref_imgs_vg_lite/draw/render/l8/demo_render_layer_normal_opa_128.png
kisvegabor, line 1: Some transformations are missing.

kisvegabor (commented):
Awesome! The code itself looks good, but the screenshots show some errors.

How can I test it on PC? I guess we needed to have the ThorVG draw unit support.

FASTSHIFT:
> Awesome! The code itself looks good, but the screenshots show some errors.
> 
> How can I test it on PC? I guess we needed to have the ThorVG draw unit support.

You can test this by turning on the following options:
```kconfig
LV_USE_DRAW_VG_LITE=y 
LV_VG_LITE_USE_GPU_INIT=y
LV_VG_LITE_USE_ASSERT=y
LV_USE_THORVG=y
LV_USE_THORVG_INTERNAL=y
LV_USE_VG_LITE_THORVG=y 
LV_DRAW_BUF_ALIGN=64
```

tests/ref_imgs_vg_lite/widgets/obj_flag_overflow_visible_1_4.png
FASTSHIFT: This is a known defect because VG-Lite does not support using path to clip path. It can only use the path's own bounding box (a rectangular area relative to the path's own coordinate system) to clip itself. Currently, it can only be circumvented in the application to avoid being clipped in the transform.

tests/ref_imgs_vg_lite/draw/render/l8/demo_render_layer_normal_opa_128.png
FASTSHIFT: See https://github.com/lvgl/lvgl/pull/4883#issuecomment-2167520960

tests/ref_imgs_vg_lite/draw/render/al88/demo_render_layer_normal_opa_128.png
FASTSHIFT: See https://github.com/lvgl/lvgl/pull/4883#issuecomment-2167520960

kisvegabor: Can you handle the fallback in runtime based on the layer's color format?

tests/ref_imgs_vg_lite/widgets/obj_flag_overflow_visible_1_4.png
kisvegabor: What falling back to non-matrix render when overflow_visible is used?

tests/ref_imgs_vg_lite/draw/render/al88/demo_render_layer_normal_opa_128.png
FASTSHIFT: At present, it seems impossible to evaluate whether vg-lite supports the corresponding color format in the refr obj process. Do you have any ideas?

FASTSHIFT: However, at present, the L8 and AL88 color formats have not been used in our products, and the priority is not high. I think we can ignore that flaw for now.

liamHowatt (commented):
Seems good. Just a question: have you thought about whether `typedef struct {float m[3][3];} lv_matrix_t;` should use `lv_value_precise_t` to honor the `LV_USE_FLOAT` setting?

On the other hand, have you thought about / tested how `lv_matrix_transform_precise_point` and `lv_matrix_transform_area` behaves when `LV_USE_FLOAT` is `0`, or considered enforcing it to be `1` like `#if !LV_USE_FLOAT #error #endif`?

FASTSHIFT:
> Seems good. Just a question: have you thought about whether `typedef struct {float m[3][3];} lv_matrix_t;` should use `lv_value_precise_t` to honor the `LV_USE_FLOAT` setting?
> 
> On the other hand, have you thought about / tested how `lv_matrix_transform_precise_point` and `lv_matrix_transform_area` behaves when `LV_USE_FLOAT` is `0`, or considered enforcing it to be `1` like `#if !LV_USE_FLOAT #error #endif`?

Very good suggestion, LV_USE_FLOAT is also almost mandatory if lv_matrix is to be used, I have added the dependency.

tests/src/lv_test_conf.h
```diff
@@ -104,10 +104,6 @@ typedef void * lv_user_data_t;
 #define  LV_DPI_DEF         130
 #endif
 
-#if defined(LVGL_CI_USING_SYS_HEAP)
-#undef LV_USE_FLOAT
```
FASTSHIFT, line 114: @kisvegabor 
Why does this [commit ](https://github.com/lvgl/lvgl/commit/84c8cf8573e41cd969375cf8937e4399ee85b846) require `#undef LV_USE_FLOAT`?

FASTSHIFT:
@liamHowatt @kisvegabor 
Is there anything else that needs to be fixed? Can we merge it into a version first ;)

tests/ref_imgs_vg_lite/widgets/obj_flag_overflow_visible_1_4.png
FASTSHIFT: > What falling back to non-matrix render when overflow_visible is used?

@kisvegabor 
I added a clipping check, which will fallback to traditional drawing when rotation + clipping is detected.
It looks pretty close to being correct, but there is still one extra section being rendered.

src/core/lv_refr.c
```diff
@@ -952,6 +952,29 @@ static void refr_obj_matrix(lv_layer_t * layer, lv_obj_t * obj)
     layer->_clip_area = clip_area_ori;
 }
 
+static bool refr_check_obj_clip_overflow(lv_layer_t * layer, lv_obj_t * obj)
+{
+    if(lv_obj_get_style_transform_rotation(obj, 0) == 0) {
```
kisvegabor, line 969: I haven't tested it but it might be required to test `lv_obj_has_flag(obj, LV_OBJ_FLAG_OVERFLOW_VISIBLE)` too.

tests/src/lv_test_conf.h
kisvegabor: Just run a config without floats too.

FASTSHIFT: I added a mandatory dependency of `LV_USE_FLOAT` to `LV_USE_MATRIX` (see [here](https://github.com/lvgl/lvgl/pull/4883#issuecomment-2184946203)), and I found that only by removing this `#undef LV_USE_FLOAT` compilation test can it pass properly. Can I delete this `#undef LV_USE_FLOAT`?

kisvegabor: Okay, let's delete it for now. We need to restructure tests later anyway.

src/core/lv_refr.c
FASTSHIFT: It seems that the determination of multi-level rotation + clipping is a relatively complicated problem. This is the real-time clipping effect in testcase, sometimes correct and sometimes incorrect.

https://github.com/lvgl/lvgl/assets/26767803/e3b2eb90-2d08-47fd-a781-3a1bc71533eb

FASTSHIFT: This is to turn off the contrast between `LV_DRAW_TRANSFORM_USE_MATRIX`.

https://github.com/lvgl/lvgl/assets/26767803/c22fc76e-7b25-4202-979c-f9ff4a707c6c

FASTSHIFT: @kisvegabor 

https://github.com/lvgl/lvgl/assets/26767803/07e78a72-b2d3-4fb9-ba2b-af45c56be899

I tried adding:  lv_obj_has_flag(obj, LV_OBJ_FLAG_OVERFLOW_VISIBLE) to the `refr_check_obj_clip_overflow` function:

```c
static bool refr_check_obj_clip_overflow(lv_layer_t * layer, lv_obj_t * obj)
{
    if(lv_obj_get_style_transform_rotation(obj, 0) == 0) {
        return false;
    }

    if(parent && lv_obj_has_flag(parent, LV_OBJ_FLAG_OVERFLOW_VISIBLE)) {
        return true;
    }

    ...
}
```

It looks pretty close, but sometimes there's a little bit more around the edges.

liamHowatt: I'm having a bit of trouble reproducing the "little bit more around the edges". The left button is always correctly clipped for me. What combination of `LV_USE_MATRIX`, `LV_USE_VECTOR_GRAPHIC`, and `LV_DRAW_TRANSFORM_USE_MATRIX` are you using? `1`, `1`, `1`?

FYI when I disable `LV_USE_VECTOR_GRAPHIC` I get this and a crash when I move the mouse over
image

FASTSHIFT: @liamHowatt 
LV_USE_MATRIX=1
LV_DRAW_TRANSFORM_USE_MATRIX=1
LV_USE_VECTOR_GRAPHIC=1
LV_USE_THORVG_INTERNAL=1

FASTSHIFT:
@kisvegabor @liamHowatt 
Can we consider merging this PR as a preliminary release first? This problem is solved by PR after the known problem, because if it is not merged for too long, it will easily cause merge conflicts.
By default, `LV_DRAW_TRANSFORM_USE_MATRIX` is not turned on and does not affect previously normal users :)

src/core/lv_refr.c
liamHowatt: This is very strange. I copied lv_conf_template.h and enabled those configs (as well as float, SDL, assert, and clib malloc+string+sprintf) and I still get the screenshot above. What could be the difference between my config and yours + the CI's?

Here is what I'm running:
```c
static void ext_draw_size_event_cb(lv_event_t * e)
{
    lv_event_set_ext_draw_size(e, 100);
}

static void btn_clicked_event_cb(lv_event_t * e)
{
    uint32_t * cnt = lv_event_get_user_data(e);
    (*cnt)++;
}

static void matrix_example(void)
{
    lv_obj_t * obj_main = lv_obj_create(lv_screen_active());
    lv_obj_set_size(obj_main, 400, 300);
    lv_obj_set_style_bg_color(obj_main, lv_palette_main(LV_PALETTE_RED), 0);
    lv_obj_add_flag(obj_main, LV_OBJ_FLAG_OVERFLOW_VISIBLE);
    lv_obj_center(obj_main);
    lv_obj_add_event_cb(obj_main, ext_draw_size_event_cb, LV_EVENT_REFR_EXT_DRAW_SIZE, NULL);

    lv_obj_t * obj_child_1 = lv_obj_create(obj_main);
    lv_obj_set_size(obj_child_1, 200, 200);
    lv_obj_set_style_bg_color(obj_child_1, lv_palette_main(LV_PALETTE_PURPLE), 0);
    lv_obj_align(obj_child_1, LV_ALIGN_LEFT_MID, -100, 0);

    lv_obj_t * btn_1 = lv_button_create(obj_child_1);
    lv_obj_set_size(btn_1, 100, 100);
    lv_obj_align(btn_1, LV_ALIGN_LEFT_MID, -75, 0);
    lv_obj_remove_flag(btn_1, LV_OBJ_FLAG_SCROLL_ON_FOCUS);
    uint32_t cnt_1;
    lv_obj_add_event_cb(btn_1, btn_clicked_event_cb, LV_EVENT_CLICKED, &cnt_1);

    lv_obj_t * label_1 = lv_label_create(btn_1);
    lv_label_set_text(label_1, "Button 1");
    lv_obj_center(label_1);

    lv_obj_t * obj_child_2 = lv_obj_create(obj_main);
    lv_obj_set_size(obj_child_2, 200, 200);
    lv_obj_set_style_bg_color(obj_child_2, lv_palette_main(LV_PALETTE_ORANGE), 0);
    lv_obj_add_flag(obj_child_2, LV_OBJ_FLAG_OVERFLOW_VISIBLE);
    lv_obj_align(obj_child_2, LV_ALIGN_RIGHT_MID, 100, 0);
    lv_obj_add_event_cb(obj_child_2, ext_draw_size_event_cb, LV_EVENT_REFR_EXT_DRAW_SIZE, NULL);

    lv_obj_t * btn_2 = lv_button_create(obj_child_2);
    lv_obj_set_size(btn_2, 100, 100);
    lv_obj_align(btn_2, LV_ALIGN_RIGHT_MID, 75, 0);
    lv_obj_remove_flag(btn_2, LV_OBJ_FLAG_SCROLL_ON_FOCUS);
    uint32_t cnt_2;
    lv_obj_add_event_cb(btn_2, btn_clicked_event_cb, LV_EVENT_CLICKED, &cnt_2);

    lv_obj_t * label_2 = lv_label_create(btn_2);
    lv_label_set_text(label_2, "Button 2");
    lv_obj_center(label_2);

    cnt_1 = 0;
    cnt_2 = 0;

    // /*The clipped part of the left button (shouldn't trigger click event)*/
    // lv_test_mouse_click_at(100, 220);

    // /*The non clipped part of the left button (should trigger click event)*/
    // lv_test_mouse_click_at(140, 220);

    // /*The left part of the right button (should trigger click event)*/
    // lv_test_mouse_click_at(650, 220);

    // /*The outer part of the right button (should trigger click event as obj_child_2 has LV_OBJ_FLAG_OVERFLOW_VISIBLE)*/
    // lv_test_mouse_click_at(690, 220);

    // TEST_ASSERT_EQUAL_UINT32(1, cnt_1);
    // TEST_ASSERT_EQUAL_UINT32(2, cnt_2);

    // TEST_ASSERT_EQUAL_SCREENSHOT("widgets/obj_flag_overflow_visible_1_1.png");

    // /*Test if the overflowing parts are rendered correctly after scrolling too*/
    // lv_obj_scroll_by_bounded(obj_main, -20, 0, LV_ANIM_OFF);
    // TEST_ASSERT_EQUAL_SCREENSHOT("widgets/obj_flag_overflow_visible_1_2.png");

    // lv_obj_scroll_by_bounded(obj_child_2, -30, 0, LV_ANIM_OFF);
    // TEST_ASSERT_EQUAL_SCREENSHOT("widgets/obj_flag_overflow_visible_1_3.png");

    /*Test with rotation*/
    lv_obj_set_style_transform_rotation(obj_main, 300, 0);
    lv_obj_set_style_transform_pivot_x(obj_main, 200, 0);
    lv_obj_set_style_transform_pivot_y(obj_main, 150, 0);

    lv_obj_set_style_transform_rotation(obj_child_1, 300, 0);
    lv_obj_set_style_transform_pivot_x(obj_child_1, 100, 0);
    lv_obj_set_style_transform_pivot_y(obj_child_1, 100, 0);

    lv_obj_set_style_transform_rotation(obj_child_2, 300, 0);
    lv_obj_set_style_transform_pivot_x(obj_child_2, 100, 0);
    lv_obj_set_style_transform_pivot_y(obj_child_2, 100, 0);

    lv_obj_set_style_transform_rotation(btn_1, 300, 0);
    lv_obj_set_style_transform_pivot_x(btn_1, 100, 0);
    lv_obj_set_style_transform_pivot_y(btn_1, 100, 0);

    lv_obj_set_style_transform_rotation(btn_2, 300, 0);
    lv_obj_set_style_transform_pivot_x(btn_2, 100, 0);
    lv_obj_set_style_transform_pivot_y(btn_2, 100, 0);

    cnt_1 = 0;
    cnt_2 = 0;

    // /*The clipped part of the left button (shouldn't trigger click event)*/
    // lv_test_mouse_click_at(185, 40);

    // /*The non clipped part of the left button (should trigger click event)*/
    // lv_test_mouse_click_at(210, 80);

    // /*The left part of the right button (should trigger click event)*/
    // lv_test_mouse_click_at(590, 370);

    // /*The outer part of the right button (should trigger click event as obj_child_2 has LV_OBJ_FLAG_OVERFLOW_VISIBLE)*/
    // lv_test_mouse_click_at(600, 430);

    // /*The clipped part of the right button (clipped because it's out of the red panel's ext draw size, shouldn't trigger click event)*/
    // lv_test_mouse_click_at(645, 430);

    // TEST_ASSERT_EQUAL_UINT32(1, cnt_1);
    // TEST_ASSERT_EQUAL_UINT32(2, cnt_2);
    // TEST_ASSERT_EQUAL_SCREENSHOT("widgets/obj_flag_overflow_visible_1_4.png");
}
```

FASTSHIFT:
@liamHowatt 
This is my config file: [lv_conf.txt](https://github.com/user-attachments/files/16236667/lv_conf.txt)

Use your `matrix_example` to reproduce this:
8119b5dfd38b333b41a89bd2c697c3e0

liamHowatt:
Thank you. The difference was that you have `LV_USE_DRAW_VG_LITE` and `LV_USE_VG_LITE_THORVG` enabled. To clarify, does it seem correct that these are required?

FASTSHIFT:
> Thank you. The difference was that you have `LV_USE_DRAW_VG_LITE` and `LV_USE_VG_LITE_THORVG` enabled. To clarify, does it seem correct that these are required?

This configuration is correct.
`LV_USE_DRAW_VG_LITE` means to turn on the VG-Lite rendering backend.
`LV_USE_VG_LITE_THORVG` means to turn on the VG-Lite emulator and use the ThorVG software to simulate VG-Lite rendering.

liamHowatt:
> This configuration is correct.

:+1:

kisvegabor:
I'm testing right now. Enabled the configs as described [here](https://github.com/lvgl/lvgl/pull/4883#issuecomment-2182248285) and added `lv_obj_set_style_transform_skew_x` to a button. If `LV_DRAW_TRANSFORM_USE_MATRIX` is enbled it crashes with:
```
    #0 0x55a7082d6626 in lv_color_24_24_mix ../lvgl/src/draw/sw/blend/lv_draw_sw_blend_to_rgb888.c:846
    #1 0x55a7082d13aa in lv_draw_sw_blend_color_to_rgb888 ../lvgl/src/draw/sw/blend/lv_draw_sw_blend_to_rgb888.c:280
    #2 0x55a7082b3f0c in lv_draw_sw_blend ../lvgl/src/draw/sw/blend/lv_draw_sw_blend.c:103
    #3 0x55a7082e1ba4 in lv_draw_sw_box_shadow ../lvgl/src/draw/sw/lv_draw_sw_box_shadow.c:191
    #4 0x55a7082d7e28 in execute_drawing ../lvgl/src/draw/sw/lv_draw_sw.c:438
    #5 0x55a7082d7516 in execute_drawing_unit ../lvgl/src/draw/sw/lv_draw_sw.c:309
    #6 0x55a7082d7c4c in render_thread_cb ../lvgl/src/draw/sw/lv_draw_sw.c:416
    #7 0x55a7080c7ef8 in generic_callback ../lvgl/src/osal/lv_pthread.c:162
    #8 0x7f5e8f3cfac2 in start_thread nptl/pthread_create.c:442
    #9 0x7f5e8f46184f  (/lib/x86_64-linux-gnu/libc.so.6+0x12684f)
```

My `lv_conf.h` file is [lv_conf.h.zip](https://github.com/user-attachments/files/16264509/lv_conf.h.zip) (I use SDL)

---

Enabling ThorVG and/or VG-Lite requires multiple steps and the advantage of this feature is not obvious just by reading `lv_conf.h`. 

To make it simpler I think we should add a dedicated ThorVG renderer and create a page for ThorVG and generic VG-Lite driver [here](https://docs.lvgl.io/master/overview/renderers/index.html). Could you add a brief docs for both?

FASTSHIFT:
> I'm testing right now. Enabled the configs as described [here](https://github.com/lvgl/lvgl/pull/4883#issuecomment-2182248285) and added `lv_obj_set_style_transform_skew_x` to a button. If `LV_DRAW_TRANSFORM_USE_MATRIX` is enbled it crashes with:
> 
> ```
>     #0 0x55a7082d6626 in lv_color_24_24_mix ../lvgl/src/draw/sw/blend/lv_draw_sw_blend_to_rgb888.c:846
>     #1 0x55a7082d13aa in lv_draw_sw_blend_color_to_rgb888 ../lvgl/src/draw/sw/blend/lv_draw_sw_blend_to_rgb888.c:280
>     #2 0x55a7082b3f0c in lv_draw_sw_blend ../lvgl/src/draw/sw/blend/lv_draw_sw_blend.c:103
>     #3 0x55a7082e1ba4 in lv_draw_sw_box_shadow ../lvgl/src/draw/sw/lv_draw_sw_box_shadow.c:191
>     #4 0x55a7082d7e28 in execute_drawing ../lvgl/src/draw/sw/lv_draw_sw.c:438
>     #5 0x55a7082d7516 in execute_drawing_unit ../lvgl/src/draw/sw/lv_draw_sw.c:309
>     #6 0x55a7082d7c4c in render_thread_cb ../lvgl/src/draw/sw/lv_draw_sw.c:416
>     #7 0x55a7080c7ef8 in generic_callback ../lvgl/src/osal/lv_pthread.c:162
>     #8 0x7f5e8f3cfac2 in start_thread nptl/pthread_create.c:442
>     #9 0x7f5e8f46184f  (/lib/x86_64-linux-gnu/libc.so.6+0x12684f)
> ```
> 
> My `lv_conf.h` file is [lv_conf.h.zip](https://github.com/user-attachments/files/16264509/lv_conf.h.zip) (I use SDL)
> 
> Enabling ThorVG and/or VG-Lite requires multiple steps and the advantage of this feature is not obvious just by reading `lv_conf.h`.
> 
> To make it simpler I think we should add a dedicated ThorVG renderer and create a page for ThorVG and generic VG-Lite driver [here](https://docs.lvgl.io/master/overview/renderers/index.html). Could you add a brief docs for both?

Please set `LV_USE_OS` to `LV_OS_NONE`. VG-Lite only supports single-threaded use.

FASTSHIFT:
> Enabling ThorVG and/or VG-Lite requires multiple steps and the advantage of this feature is not obvious just by reading lv_conf.h.

The purpose of wrapping ThorVG with VG-Lite API is to reduce the debugging time on hardware (because the program needs to be downloaded back and forth), especially memory problems can be located very quickly using ASAN.

FASTSHIFT:
> To make it simpler I think we should add a dedicated ThorVG renderer and create a page for ThorVG and generic VG-Lite driver [here](https://docs.lvgl.io/master/overview/renderers/index.html). Could you add a brief docs for both?

No problem, I'll add documentation to explain.

kisvegabor:
> Please set LV_USE_OS to LV_OS_NONE. VG-Lite only supports single-threaded use.

Thanks it's running now, but for this:
```c
    lv_obj_set_style_transform_skew_x(btn2, 30, 0);
```
I get
image

FASTSHIFT:
> > Please set LV_USE_OS to LV_OS_NONE. VG-Lite only supports single-threaded use.
> 
> Thanks it's running now, but for this:
> 
> ```c
>     lv_obj_set_style_transform_skew_x(btn2, 30, 0);
> ```
> 
> I get image

You need to turn on the configuration `LV_VG_LITE_USE_BOX_SHADOW`, use VG-Lite to draw simulated shadows, otherwise it will fallback to software rendering.

FASTSHIFT:
> To make it simpler I think we should add a dedicated ThorVG renderer and create a page for ThorVG and generic VG-Lite driver [here](https://docs.lvgl.io/master/overview/renderers/index.html). Could you add a brief docs for both?

@kisvegabor 
Documentation added: https://github.com/lvgl/lvgl/pull/6557

kisvegabor:
Thanks, it's working now! However I've noticed that if I hover the mouse over the skew button it has artifacts. (The mouse cusrsor is an simple lv_obj). Is it a know issue?

matrix

FASTSHIFT:
> Thanks, it's working now! However I've noticed that if I hover the mouse over the skew button it has artifacts. (The mouse cusrsor is an simple lv_obj). Is it a know issue?
> 
> matrix [ matrix ](https://private-user-images.githubusercontent.com/7599318/352082687-5704a0ff-68a3-4eb6-bfd0-13dcc357c862.gif?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3MjE5MDc5NTEsIm5iZiI6MTcyMTkwNzY1MSwicGF0aCI6Ii83NTk5MzE4LzM1MjA4MjY4Ny01NzA0YTBmZi02OGEzLTRlYjYtYmZkMC0xM2RjYzM1N2M4NjIuZ2lmP1gtQW16LUFsZ29yaXRobT1BV1M0LUhNQUMtU0hBMjU2JlgtQW16LUNyZWRlbnRpYWw9QUtJQVZDT0RZTFNBNTNQUUs0WkElMkYyMDI0MDcyNSUyRnVzLWVhc3QtMSUyRnMzJTJGYXdzNF9yZXF1ZXN0JlgtQW16LURhdGU9MjAyNDA3MjVUMTE0MDUxWiZYLUFtei1FeHBpcmVzPTMwMCZYLUFtei1TaWduYXR1cmU9NDcxMTgwOTYwOGRhMjc4NGFmYTBkZTBhM2E2MmMwOGJkN2IwYWVmMGE5Y2MxNjNlZTY4OGZmNGEwZDc0NWQzZCZYLUFtei1TaWduZWRIZWFkZXJzPWhvc3QmYWN0b3JfaWQ9MCZrZXlfaWQ9MCZyZXBvX2lkPTAifQ.s6XnwWNmbJIparRAcxy5qEipW0Ss5FdWbEMetyxL-_Y) [ ](https://private-user-images.githubusercontent.com/7599318/352082687-5704a0ff-68a3-4eb6-bfd0-13dcc357c862.gif?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3MjE5MDc5NTEsIm5iZiI6MTcyMTkwNzY1MSwicGF0aCI6Ii83NTk5MzE4LzM1MjA4MjY4Ny01NzA0YTBmZi02OGEzLTRlYjYtYmZkMC0xM2RjYzM1N2M4NjIuZ2lmP1gtQW16LUFsZ29yaXRobT1BV1M0LUhNQUMtU0hBMjU2JlgtQW16LUNyZWRlbnRpYWw9QUtJQVZDT0RZTFNBNTNQUUs0WkElMkYyMDI0MDcyNSUyRnVzLWVhc3QtMSUyRnMzJTJGYXdzNF9yZXF1ZXN0JlgtQW16LURhdGU9MjAyNDA3MjVUMTE0MDUxWiZYLUFtei1FeHBpcmVzPTMwMCZYLUFtei1TaWduYXR1cmU9NDcxMTgwOTYwOGRhMjc4NGFmYTBkZTBhM2E2MmMwOGJkN2IwYWVmMGE5Y2MxNjNlZTY4OGZmNGEwZDc0NWQzZCZYLUFtei1TaWduZWRIZWFkZXJzPWhvc3QmYWN0b3JfaWQ9MCZrZXlfaWQ9MCZyZXBvX2lkPTAifQ.s6XnwWNmbJIparRAcxy5qEipW0Ss5FdWbEMetyxL-_Y)

This is a known issue. First, the quality of the skew effect rendered by ThorVG's rasterizer is not good enough. Second, the redraw area calculation does not take the skew style into account (currently it is not a high priority and no one uses this style yet). This can be solved by manually expanding the redraw area.
Currently, the most commonly used transform styles are scale and rotate. It is recommended to focus on testing these two styles.

kisvegabor:
Okay, I assumed that this is the case.

> currently it is not a high priority and no one uses this style yet

How do you know that? :smile: Just joking, I agree, these are not high priority features.

kisvegabor (approved):

FASTSHIFT:
> How do you know that? 😄 Just joking, I agree, these are not high priority features.

Because currently the interactive effects on mobile devices are basically scale and rotate, and skew effects are relatively rare.

XuNeo (approved):

FASTSHIFT:
Great! If find new problems later, I will try to send a PR to fix them.

FASTSHIFT:
It looks like the documentation compilation had a little problem, missing the ability to enable `LV_USE_MATRIX`:
https://github.com/lvgl/lvgl/actions/runs/10094997064/job/27914088965

@kisvegabor 
In which `lv_conf` should I add this configuration?

kisvegabor:
I've just fixed it here: https://github.com/lvgl/lv_web_emscripten/commit/92610364ca29835b3f0fcf79b9f1ae671667535e
Let's see if it's enough.

FASTSHIFT:
> I've just fixed it here: [lvgl/lv_web_emscripten@9261036](https://github.com/lvgl/lv_web_emscripten/commit/92610364ca29835b3f0fcf79b9f1ae671667535e) Let's see if it's enough.

Thank you very much!

PR #4898, opened 2023-11-29
clangd: IWYU pragmas for proper auto-import

voltangle:

This PR adds `// IWYU pragma: private, include <lvgl/lvgl.h">` statements all over internal header files that are re-exported by `lvgl.h`. This allows `clangd` to properly auto-import a correct, `<lvgl/lvgl.h>` header, instead of an internal one.

### Checkpoints
 
### DIFF
--- a/src/core/lv_global.h
+++ b/src/core/lv_global.h
@@ -2,7 +2,7 @@
  * @file lv_global.h
  *
  */
-// IWYU pragma: private, include <lvgl/lvgl.h">
+// IWYU pragma: private, include <lvgl/lvgl.h>
 
 #ifndef LV_GLOBAL_H
 #define LV_GLOBAL_H
--- a/src/core/lv_global.h
+++ b/src/core/lv_global.h
@@ -2,6 +2,7 @@
  * @file lv_global.h
  *
  */
+// IWYU pragma: private, include <lvgl/lvgl.h">
 
 #ifndef LV_GLOBAL_H
 #define LV_GLOBAL_H
--- a/src/core/lv_group.h
+++ b/src/core/lv_group.h
@@ -2,7 +2,7 @@
  * @file lv_group.h
  *
  */
-// IWYU pragma: private, include <lvgl/lvgl.h">
+// IWYU pragma: private, include <lvgl/lvgl.h>
 
 #ifndef LV_GROUP_H
 #define LV_GROUP_H
--- a/src/core/lv_group.h
+++ b/src/core/lv_group.h
@@ -2,6 +2,7 @@
  * @file lv_group.h
  *
  */
+ // IWYU pragma: private, include <lvgl/lvgl.h">
 
 #ifndef LV_GROUP_H
 #define LV_GROUP_H
--- a/src/core/lv_group.h
+++ b/src/core/lv_group.h
@@ -2,7 +2,7 @@
  * @file lv_group.h
  *
  */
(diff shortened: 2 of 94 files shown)


kisvegabor:
Hi,

> This allows clangd to properly auto-import a correct, <lvgl/lvgl.h> header, instead of an internal one.

Could you tell in more detail what the exact problem is that this PR solves?

ggoraa:
> Could you tell in more detail what the exact problem is that this PR solves?

Definitely!

You see, I'm using `clangd` as an LSP, and it has a feature - auto-include, like a lot of other LSPs. But, when I actually use it, instead of importing the "umbrella" header `<lvgl/lvgl.h>`, it imports an internal one, from `src`, which is definitely not ideal.

Here's a video demonstration:

https://github.com/lvgl/lvgl/assets/61655029/06114108-5268-4ace-8112-24cc73620621

kisvegabor:
I see thank you. Do you what is the policy of clangs when it decides which header to choose?

I'm just asking because it would be great to make it work out of the box.

Besides I wonder if it would be enough to add the comment only to `lvgl.h`. Regular users should need only it.

ggoraa:
https://clangd.llvm.org/design/include-cleaner#umbrella-headers

XuNeo:
I haven't had the opportunity to utilize this feature yet. Is there a way to circumvent the need to alter each private header?

Would it be feasible to only make modifications to lvgl.h to designate the headers as export? For reference, [see the documentation on IWYU pragmas](https://github.com/include-what-you-use/include-what-you-use/blob/master/docs/IWYUPragmas.md#iwyu-pragma-export)

XuNeo:
I haven't had the opportunity to utilize this feature yet. Is there a way to circumvent the need to alter each private header?

Would it be feasible to only make modifications to lvgl.h to designate the headers as export? For reference, [see the documentation on IWYU pragmas](https://github.com/include-what-you-use/include-what-you-use/blob/master/docs/IWYUPragmas.md#iwyu-pragma-export)

ggoraa:
Yeah, no, sadly it doesn't work with `clangd`
only the "redirect" comments do
shouldn't be a neovim problem, because it's all handled by `clangd` itself

XuNeo:
I still haven't the chance to try it. Today I saw a file in `skia` repo to include directives for IWYU that we can do the same.

See explains here. https://github.com/include-what-you-use/include-what-you-use/blob/master/docs/IWYUMappings.md

> Include mappings support a special wildcard syntax for the first entry:
> 
> { include: ["@<internal/.*>", "private", "<public>", "public"] }

See files from skia. https://github.com/google/skia/blob/main/toolchain/linux_trampolines/IWYU_mapping.imp

From the video you provides, we can map all includes starting with `src/` to `lvgl/lvgl.h`
> { include: ["@<src/.*>", "private", "<lvgl/lvgl.h>", "public"] }

ggoraa:
I hate this bot, honestly

ggoraa:
I hate this bot, honestly
@kisvegabor can you reopen this please? And why do you guys even have this "close when stale" bot?) Imo it's useless, and usually just gets in the way

kisvegabor:
Reopened. 

> And why do you guys even have this "close when stale" bot?)

Else we would have a lot of abandoned open issues and it would be hard to filter out the active relevant ones.  If an issue was important we can pin it or reopen it. For example in case of this issue I assumed that no one has time or interest to continue the investigation.

PR #4907, opened 2023-12-01
fix: workaround MSVC C4576 error

### LINKED ISSUE #4905: MSVC will return error C4576 for the following code
https://github.com/lvgl/lvgl/blob/ba229b0708fe3b893a179da3b3635a807b046c4b/src/misc/lv_area.h#L267-L279

Kenji Mouri

XuNeo:

Fix #4905

Simply initialize a struct and return it instead
### Checkpoints
 
### DIFF
--- a/src/misc/lv_area.h
+++ b/src/misc/lv_area.h
@@ -266,16 +266,20 @@ void lv_point_transform(lv_point_t * p, int32_t angle, int32_t scale_x, int32_t 
 
 static inline lv_point_t lv_point_from_precise(const lv_point_precise_t * p)
 {
-    return (lv_point_t) {
+    lv_point_t point = {
         (int32_t)p->x, (int32_t)p->y
     };
+
+    return point;
 }
 
 static inline lv_point_precise_t lv_point_to_precise(const lv_point_t * p)
 {
-    return (lv_point_precise_t) {
+    lv_point_precise_t point = {
         p->x, p->y
     };
+
+    return point;
 }
 
 static inline void lv_point_set(lv_point_t * p, int32_t x, int32_t y)


kdschlosser:
Question on this. Doesn't this turn into twice the memory use due to the explicit declaration of  a local variable? I know the memory is used dynamically and it ends up getting freed, but wouldn't it cause more memory fragmentation? The way it would there is no explicit allocation to a variable so the only allocation is going to be what is returned. The return it's self is not a pointer to a memory address so a second allocation takes place for the copy. Better to do the return like is was no?

I don't know how the compiler does it's optimizations and if this would be something that would be optimized away resulting in the same machine code. I would image that would depend on the compiler being used. Maybe someone with more knowledge with how a compiler would compile this code could chime in.

kisvegabor:
These variables are passed via either
- CPU internal registers (no real allocation)
- stack (not fragmented as the last fame is always freed)

kdschlosser:
i would think fragmentation would occur. instead of having the single allocation at the return now there is 2. the memory that gets freed is the local variable and that creates a hole. again I do not know the inner working of C compilers but it seems like there would be additional work that needs to be done and also additional memory use.

I know that in Python code doing this

```python

def func():
    return some_class()
```

is far better in both memory use and also speed than this

```python
def func():
    sc = some_class()
    return sc
```

kisvegabor:
All these small data are stored in CPU registers so no real memory allocation happens. Even if it were, it's never dynamic automatically but just stack operations. And stack cannot be fragmented by definition.

However, these code is optimized out for sure anyway.

kdschlosser:
i am still learning

PR #4921, opened 2023-12-04
fix(obj): readjust scroll after layout when child removed

niklasf:
Backported #4916 to lvgl 8.


kisvegabor:
I think we are safe to do this. Thank you!

PR #4922, opened 2023-12-04
fix(style): make the style parameter const in lv_obj_add/remove_style

niklasf:
Can we backport this to lvgl 8, also? Cherry-picked your commit.


niklasf:
Sorry, nevermind. Looks like it's not so simple in lvgl 8.

PR #4924, opened 2023-12-04
feat(tool): add compressed binary image support to C-array

XuNeo:

This PR enables PNG to compressed C-array conversion. 

Key updates include:
  - move `compress` parameter to `to_bin` method, since it's only used in lvgl binary image
  - C-array map data is now static
  - compression method is appended to C-array variable name, preventing naming conflicts.

### Checkpoints
 
### DIFF
--- a/scripts/LVGLImage.py
+++ b/scripts/LVGLImage.py
@@ -579,14 +579,12 @@ class LVGLImage:
 
         varname = path.basename(filename).split('.')[0]
         varname = varname.replace("-", "_")
-        varname += f"_{self.cf.name.lower()}"
-        if self.stride != (self.w * self.cf.bpp + 7) // 8:
-            varname += f"_stride{self.stride}"
-        if compress is not CompressMethod.NONE:
-            varname += f"_{compress.name}"
+        varname = varname.replace(".", "_")
 
-        flags = 0
-        flags |= 0x08 if compress != CompressMethod.NONE else 0
+        flags = "0"
+        if compress is not CompressMethod.NONE:
+            flags += " | LV_IMAGE_FLAGS_COMPRESSED"
+
         compressed = LVGLCompressData(self.cf, compress, self.data)
 
         header = f'''
@@ -615,7 +613,7 @@ uint8_t {varname}_map[] = {{
 
 const lv_img_dsc_t {varname} = {{
   .header.cf = LV_COLOR_FORMAT_{self.cf.name},
-  .header.flags = {hex(flags)},
+  .header.flags = {flags},
   .header.w = {self.w},
   .header.h = {self.h},
   .header.stride = {self.stride},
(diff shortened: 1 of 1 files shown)


kisvegabor:
> compression method is appended to C-array variable name, preventing naming conflicts

I think it's not required. The developer who is creating the UI is probably not interested in the compression method. And when the compression method changes for any reason it shouldn't be required to update all usage of the image in the UI.

What about adding an "output-name" parameter to the script to allow defining the names explicitly.

XuNeo:
> What about adding an "output-name" parameter to the script to allow defining the names explicitly.

Update to use filename as var name. File name can be specified from `img.to_c_array(filename.c)`. So it makes sense to keep filename as unique, thus var name unique.

```python
    img.to_c_array("output/cogwheel-abc.c")  # file name is used as c var name

```

kisvegabor:
I meant allowing for the user to specify the output name. E.g. 
`LVGLImage.py <some params> --output-name my_image1` or 
`LVGLImage.py <some params> --output-name "my_image1 my_image2 my_image3"` in case of multiple files

XuNeo:
> I meant allowing for the user to specify the output name. E.g. `LVGLImage.py <some params> --output-name my_image1` or `LVGLImage.py <some params> --output-name "my_image1 my_image2 my_image3"` in case of multiple files

There's a much simpler solution without messing around with large params. Here's a script I plan for CI to auto-generate test images.

```python

def generate_test_images():
    invalids = (ColorFormat.UNKNOWN, ColorFormat.TRUECOLOR, ColorFormat.TRUECOLOR_ALPHA)
    formats = [i for i in ColorFormat if i not in invalids]
    png_path = os.path.join(lvgl_test_dir, "test_images/pngs")
    pngs = list(Path(png_path).rglob("*.[pP][nN][gG]"))
    print(f"png files: {pngs}")

    align_options = [1, 64]

    for align in align_options:
        for compress in CompressMethod:
            compress_name = compress.name if compress != CompressMethod.NONE else "UNCOMPRESSED"
            outputs = os.path.join(lvgl_test_dir, f"test_images/stride_align{align}/{compress_name}/")
            os.makedirs(outputs, exist_ok=True)
            for fmt in formats:
                for png in pngs:
                    img = LVGLImage().from_png(png, cf=fmt, background=0xffffff)
                    img.adjust_stride(align=16)
                    img.set_compress(compress)
                    output = os.path.join(outputs, f"{Path(png).stem}_{fmt.name}.bin")
                    img.to_bin(output)
                    output = os.path.join(outputs, f"{Path(png).stem}_{fmt.name}.c")
                    img.to_c_array(output)
                    print(f"converting {os.path.basename(png)}, format: {fmt.name}, compress: {compress_name}")

```

kisvegabor:
Aaaahhaa, very nice! I didn't think of using the image converter in a Python script. I need to break my C chains :smile:

PR #4929, opened 2023-12-05
feat(needle):add needle to scale

### LINKED ISSUE #4917: The issue of using scale to replace meter
### Introduce the problem
In v9, the **meter**  has been removed and replaced with **scale.** However, if you want to use **needle,** currently this is achieved by adding a needle image and setting an angle. In V8, the meter can directly set numerical values, and the needle can display the corresponding position, making it very convenient to use.
### Suggested solution
Therefore, should we restore the **meter** control or add a **needle** control? This way, users can also reduce the difficulty of use.

lhdjply:

#4917
feat(needle):add needle to scale

image

### Checkpoints
 

lhdjply:
1. Currently, using images to create needs has not been implemented yet.
2. Is there a way to add a circle to the starting point of the line by setting the style?

kisvegabor:
Thank you for the PR.

I think it'd be more flexible to approach it like this:
```c
lv_obj_t * scale = lv_scale_create(parent);

lv_obj_t * line = lv_line_create(scale);
//Add styles to the line

lv_scale_set_line_needle_value(scale, line, value, needle_length);
```

This way we could add multiple needles to the scale as well.

lhdjply:
make sense

src/widgets/scale/lv_scale.c
```diff
@@ -182,6 +182,47 @@ void lv_scale_set_round_props(lv_obj_t * obj, uint32_t angle_range, int32_t rota
     lv_obj_invalidate(obj);
 }
 
+void lv_scale_set_line_needle_value(lv_obj_t * obj, lv_obj_t * needle_line, uint32_t needle_length, int32_t value)
+{
+    uint32_t angle;
+    lv_value_precise_t scale_width, scale_height;
+    lv_value_precise_t needle_length_x, needle_length_y;
+    float cosx_angle_value, sinx_angle_value;
```
kisvegabor: `float` shouldn't be used. Here a simple `int32_t` should be enough too, as we need only 1 px precision.

src/widgets/scale/lv_scale.c
```diff
@@ -182,6 +182,47 @@ void lv_scale_set_round_props(lv_obj_t * obj, uint32_t angle_range, int32_t rota
...
+    else {
+        angle = scale->angle_range * (value - scale->range_min) / (scale->range_max - scale->range_min);
+    }
+
+    scale_width = (lv_value_precise_t)(lv_obj_get_style_width(obj, LV_PART_MAIN));
+    scale_height = (lv_value_precise_t)(lv_obj_get_style_height(obj, LV_PART_MAIN));
+
+    cosx_angle_value = (float)(lv_trigo_cos(scale->rotation + angle)) / 32768;
+    sinx_angle_value = (float)(lv_trigo_sin(scale->rotation + angle)) / 32768;
+    needle_length_x = (lv_value_precise_t)(needle_length * cosx_angle_value);
```
kisvegabor: This should wourk without `float`
```c 
needle_length_x = (needle_length * lv_trigo_cos(scale->rotation + angle))  >> LV_TRIGO_SHIFT;
```

lhdjply:
Operation error, changed the hash value of someone else's commit. I closed this PR and resubmitted a new one

Issue #4342, opened 2023-07-05
[Parallel rendering] Optimization


kisvegabor:
Continue #4016

The parallel rendering update is merged to `master` but now its performance is worse than v8. 
We need to identify the bottlenecks and optimize them.

nicusorcitu:
Nice having this topic opened! I was also about to create one.

kisvegabor:
I've tried out the current master on an STM32F769 board and found these:
- I needed to [add a `volatile`](https://github.com/lvgl/lvgl/commit/388a5d7f975a91eb964e419e8d3ab81fd8a306c5) somewhere to make rendering work well with O3 optimization. I don't understand why.
- I tested with `lv_demo_benchmark(LV_DEMO_BENCHMARK_MODE_RENDER_AND_DRIVER);` and got very high FPS. Better than v8
- Using the widget's demo the perfomance felt good too (not worse than v8)
- In the benchmark of v8 there is no MODEs for `lv_demo_benchmark` so probably it's not a fair comparision
- We need the same benchmark for both v8 and v9 (I think v9's benchmark can be easily backported to v8, but I haven't tried it yet)

This the profiler log from v9. There is nothing extraordinary in it. 
![image](https://github.com/lvgl/lvgl/assets/7599318/5e3fb246-fc81-4d20-8fd3-6adfcd7d9bda)

When you test v9, be sure to disable OS (`LV_USE_OS  == LV_OS_NONE`) and enable O3 optimization for both v8 an v9.

nicusorcitu:
Just run the V9 release, CPU only, O3, on RT1170. The CPU is a single core 1GHz. GPU is disabled for now.
The results are attached without and with the FreeRTOS (e.g. with two threads lvgl_main and lvgl_draw SW).
[v9_cpu_os_freertos.txt](https://github.com/lvgl/lvgl/files/11969417/v9_cpu_os_freertos.txt)
[v9_cpu_os_none.txt](https://github.com/lvgl/lvgl/files/11969418/v9_cpu_os_none.txt)
I do not think the task switching takes so much, I will rework the task sync mechanism and see if any difference. But till then, 
I am just curious, on pthread (under linux) is the same big difference?

nicusorcitu:
How do you get the profiler logs?

kisvegabor:
On Linux there is some difference but not that much. Sometimes it was faster without OS and something it was faster with pthread.

This is my way of thinking:
[...]

PR #4944, opened 2023-12-06
chore(draw): fix typo

FASTSHIFT:

From the discussion: https://github.com/lvgl/lvgl/issues/4342#issuecomment-1699134058, the larger the preference_score value, the slower the rendering.

### Checkpoints
 
### DIFF
--- a/src/draw/lv_draw.h
+++ b/src/draw/lv_draw.h
@@ -91,7 +91,7 @@ typedef struct _lv_draw_task_t {
      * Set to which extent `preferred_draw_unit_id` is good at this task.
      * 80: means 20% better (faster) than software rendering
      * 100: the default value
-     * 110: means 10% better (faster) than software rendering
+     * 110: means 10% worse (slower) than software rendering
      */
     uint8_t preference_score;
 


FASTSHIFT:
The smaller the value of `preference_score`, the stronger the performance. I feel that this design is a bit confusing. Consider defining a `SCORE_MAX=255` to represent the strongest rendering performance? instead of 0.

kisvegabor:
My idea was to give a meaning to the score relative to the SW render but in the practice we won't have so many draw units and there won't be a big competition. Instead we will define what we want in an exact way, and not only based on performance but power consumption, freeing up the CPU, etc.

So I agree to make it work the opposite way. 

An other note: I've realized that we don't need to store the preferred score in `draw_task` but we can do something like:
```c
        /*Let the draw units assign the draw task*/
        uint8_t score = 100;
        t->preferred_draw_unit_id = 0;
        lv_draw_unit_t * u = info->unit_head;
        while(u) {
            if(u->evaluate_cb) u->evaluate_cb(u, t, &score);
            u = u->next;
        }
```

If you agree, could you update the PR accordingly?

FASTSHIFT:
> My idea was to give a meaning to the score relative to the SW render but in the practice we won't have so many draw units and there won't be a big competition. Instead we will define what we want in an exact way, and not only based on performance but power consumption, freeing up the CPU, etc.
> 
> So I agree to make it work the opposite way.

No problem, I will send PR.

> 
> An other note: I've realized that we don't need to store the preferred score in `draw_task` but we can do something like:
> 
> ```c
>         /*Let the draw units assign the draw task*/
>         uint8_t score = 100;
>         t->preferred_draw_unit_id = 0;
>         lv_draw_unit_t * u = info->unit_head;
>         while(u) {
>             if(u->evaluate_cb) u->evaluate_cb(u, t, &score);
>             u = u->next;
>         }
> ```
> 

I think it is more appropriate to keep the preference score in `lv_draw_task_t`, because considering that more parameters may need to be expanded in the future, the public interface can be modified as little as possible.

FASTSHIFT:
> > My idea was to give a meaning to the score relative to the SW render but in the practice we won't have so many draw units and there won't be a big competition. Instead we will define what we want in an exact way, and not only based on performance but power consumption, freeing up the CPU, etc.
> > So I agree to make it work the opposite way.
> 
> No problem, I will send PR.
> 
> > An other note: I've realized that we don't need to store the preferred score in `draw_task` but we can do something like:
> > ```c
> >         /*Let the draw units assign the draw task*/
> >         uint8_t score = 100;
> >         t->preferred_draw_unit_id = 0;
> >         lv_draw_unit_t * u = info->unit_head;
> >         while(u) {
> >             if(u->evaluate_cb) u->evaluate_cb(u, t, &score);
> >             u = u->next;
> >         }
> > ```
> 
> I think it is more appropriate to keep the preference score in `lv_draw_task_t`, because considering that more parameters may need to be expanded in the future, the public interface can be modified as little as possible.

kisvegabor:
> I think it is more appropriate to keep the preference score in lv_draw_task_t, because considering that more parameters may need to be expanded in the future, the public interface can be modified as little as possible

Good point! What about having
```c
typedef struct
{
  uint8_t score;
}lv_draw_evaluate_data_t;
```

It would keep the door open for changes without API break, yet we don't have to store it in draw tasks.

FASTSHIFT:
I still think it is better to put it in `lv_draw_task_t`, because `lv_draw_task_t` runs through the context of the entire drawing task process, which is very convenient for debugging. I also prefer simple parameter passing.

kisvegabor:
Ok, 1 byte  of extra data/draw task shouldn't hurt much. :slightly_smiling_face:

PR #4953, opened 2023-12-07
fix(qrcode): Fix QR code picking a version too big for width

wizard97:

The current `lv_qrcode` arbitrarily increases the QR code version based on the formula `uint32_t version_extend = remain / (scale << 2);`.  This is not robust and often will bump the QR code version to one that no longer fits the screen width.

Instead find the largest QR code version that still maintains the scale of the smallest QR code version that can fit all the data.

### Checkpoints
 
### DIFF
--- a/src/libs/qrcode/lv_qrcode.c
+++ b/src/libs/qrcode/lv_qrcode.c
@@ -118,13 +118,15 @@ lv_result_t lv_qrcode_update(lv_obj_t * obj, const void * data, uint32_t data_le
     if(qr_size <= 0) return LV_RESULT_INVALID;
     int32_t scale = img_dsc->header.w / qr_size;
     if(scale <= 0) return LV_RESULT_INVALID;
-    int32_t remain = img_dsc->header.w % qr_size;
 
-    /* The qr version is incremented by four point */
-    uint32_t version_extend = remain / (scale << 2);
-    if(version_extend && qr_version < qrcodegen_VERSION_MAX) {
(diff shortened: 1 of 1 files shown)


src/libs/qrcode/lv_qrcode.c
```diff
@@ -118,13 +118,15 @@ lv_result_t lv_qrcode_update(lv_obj_t * obj, const void * data, uint32_t data_le
...
-        qr_version = qr_version + version_extend > qrcodegen_VERSION_MAX ?
-                     qrcodegen_VERSION_MAX : qr_version + version_extend;
+    // Pick the largest QR code that still maintains scale.
+    for(int32_t i = qr_version + 1; i < qrcodegen_VERSION_MAX; i++) {
+        if(qrcodegen_version2size(i) * scale <= img_dsc->header.w) {
+            qr_version = i;
+        }
+        else {
+            break;
+        }
```
XuNeo: Could you apply below one?  It looks a little cleaner.
Also please use `/*Comment with no space before and after*/` comment style instead of `// comment`
```suggestion
    for(int32_t i = qr_version + 1; i < qrcodegen_VERSION_MAX; i++) {
        if(qrcodegen_version2size(i) * scale > img_dsc->header.w)
            break;

        qr_version = i;
     }
```

wizard97: All set!

kisvegabor:
Thank you!

PR #4959, opened 2023-12-07
docs(tjpgd): update tjpgd docs

niklasf:
Since lvgl/lv_lib_split_jpg@a2ee0bba87fde13e173b425c77851599ec0ffaba JPG decoding is no longer limited to 16 bit color depth.

### DIFF
--- a/docs/libs/tjpgd.rst
+++ b/docs/libs/tjpgd.rst
@@ -2,43 +2,28 @@
 TJPGD decoder
 =============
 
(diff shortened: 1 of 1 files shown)


kisvegabor:
Actually there is no such SJPG anymore in v9. There is the normal JPG image which is decoded in 8x8 tiles.

niklasf:
Amended to also remove mentions of SJPG.

kisvegabor:
Great, thank you!

PR #4964, opened 2023-12-08
fix(snapshot): set data_size on returned dsc

niklasf:
Before, `data_size` would always be 0.

    lv_img_dsc_t * snapshot = lv_snapshot_take(screen, LV_COLOR_FORMAT_RGB888);
    LV_ASSERT(snapshot->data_size > 0); // fail

Also align naming of parameter name and variables (`buf` instead of `buff`, as in function names).

### DIFF
--- a/src/others/snapshot/lv_snapshot.c
+++ b/src/others/snapshot/lv_snapshot.c
@@ -77,11 +77,11 @@ uint32_t lv_snapshot_buf_size_needed(lv_obj_t * obj, lv_color_format_t cf)
  * @param cf        color format for generated image.
  * @param dsc       image descriptor to store the image result.
  * @param buf       the buffer to store image data.
- * @param buff_size provided buffer size in bytes.
+ * @param buf_size  provided buffer size in bytes.
  * @return          LV_RESULT_OK on success, LV_RESULT_INVALID on error.
  */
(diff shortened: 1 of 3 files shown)


kisvegabor:
Thank you! :+1:

src/others/snapshot/lv_snapshot.c
```diff
@@ -98,7 +98,8 @@ lv_result_t lv_snapshot_take_to_buf(lv_obj_t * obj, lv_color_format_t cf, lv_ima
             return LV_RESULT_INVALID;
     }
 
-    if(lv_snapshot_buf_size_needed(obj, cf) > buff_size || buff_size == 0) return LV_RESULT_INVALID;
+    uint32_t buf_size_needed = lv_snapshot_buf_size_needed(obj, cf);
+    if(buf_size_needed > buf_size || buf_size == 0) return LV_RESULT_INVALID;
```
XuNeo: I think `buf_size_needed == 0` should also return invalid. That will happen when taking snapshot for a zero sized obj.

niklasf: Updated. The `buf_size == 0` check is now implied.

kisvegabor:
Looks good!

PR #4972, opened 2023-12-09
fix(snapshot): set data_size on returned dsc

niklasf:
Backports #4964 to lvgl 8.

### DIFF
--- a/src/extra/others/snapshot/lv_snapshot.c
+++ b/src/extra/others/snapshot/lv_snapshot.c
@@ -73,15 +73,15 @@ uint32_t lv_snapshot_buf_size_needed(lv_obj_t * obj, lv_img_cf_t cf)
 
 /** Take snapshot for object with its children, save image info to provided buffer.
  *
- * @param obj    The object to generate snapshot.
- * @param cf     color format for generated image.
- * @param dsc    image descriptor to store the image result.
- * @param buf    the buffer to store image data.
- * @param buff_size provided buffer size in bytes.
+ * @param obj      The object to generate snapshot.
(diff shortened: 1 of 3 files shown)


src/extra/others/snapshot/lv_snapshot.c
```diff
@@ -148,6 +148,7 @@ lv_res_t lv_snapshot_take_to_buf(lv_obj_t * obj, lv_img_cf_t cf, lv_img_dsc_t *
     lv_mem_free(draw_ctx);
 
     dsc->data = buf;
+    dsc->data_size = buf_size_needed;
```
PGNetHun, line 151: I think this should be also `buf_size`

PGNetHun (changes requested):

src/extra/others/snapshot/lv_snapshot.c
niklasf: `buf_size` is the available buffer size as passed into the function. But only `buf_size_needed` bytes will actually have valid image data.

kisvegabor: ~I agree to use `buf_size_needed` to zero only the really required buffer.~

kisvegabor: Sorry, I missed something. So the questions is if need to use the buf size passed by the user (the user might want to know the size he was using) or set it for the image (LVGL might want to know the real buffer size of image).

I vote for 2) so `dsc->data_size = buf_size_needed;`.

I merge this PR now, if you think that it would be important to discuss it, please open an issue.

PR #4988, opened 2023-12-12
fix: Empty struct definitions are not allowed in MSVC

glueckm:
Microsoft Visual Studio does not allow the definition of empty structs. At least one field is reauired in the struct cache_data_t


kisvegabor:
Actually that `typedef` is not used. Could you please remove it?

PR #4997, opened 2023-12-13
adds MicroPython test to lvgl

kdschlosser:
I don't know why this is needed but here it is. Red the top of the `__init__.py` file for instructions on how to use.


kisvegabor:
Thank you, Kevin! :pray: 

A few notes:
- Please use the [this](https://github.com/lvgl/lvgl/blob/master/tests/ref_imgs/binding.png) as reference image
- Could you update the workflow [here](https://github.com/lvgl/lvgl/blob/master/.github/workflows/build_micropython.yml#L67-L74) to run this test instead of `run.sh`. This way we could see if it's really passing the CI.

kisvegabor:
@kdschlosser I've tried the script locally and integrated to CI but I got the same error on both places: 
image

kdschlosser:
give it a shot now,

kisvegabor:
Thanks! Seems better but still failing:
image

kisvegabor:
Note that, I've replaced the reference image with the one from the `ref_imgs` folder.

kdschlosser:
add --debug to the command line for it so I can see what the output is.

kdschlosser:
out of question. did you disable LVGL's logging before compiling to run these tests? Logging needs to be disabled in the micropython build other wise the test is going to fail.

kisvegabor:
I get this log locally: 
[a.txt](https://github.com/lvgl/lvgl/files/13715046/a.txt)

It seems to be polluted, although the logs are disabled.

kdschlosser:
the --debug is dumping all kinds of data. When I say disable logging that leans LV_USE_LOG needs to be set to 0. Setting the logging level to NONE is not going to do it. The test program checks and sees if the logging functions have been added or not.

I am going to push some changes that provide a bit more detail as to what is happening. It's the funky way that MicroPython handles errors when the code runs that is causing a little bit of as trip up. I have it sorted out locally so an error would get spit out that was meaningful. I had thought that when an error occurred it would get output in a manner that would be able to be picked up by the test program. Turns out I was wrong on that one so I had to build in a mechanism to facilitate that aspect of it. 


This stuff you are seeing [31;1m is just ansi codes to recolor the text that gets output from the debugging process. That is not what is being output by MicroPython. 

I updated the code to have better error handling and I cleaned up a few things as well.

kdschlosser:
OH and this MICROPY_LV_USE_LOG doesn't do anything. The lv_conf.h file has to be manually changed. There was nothing ever built into the build system to be able to change the settings from the command line. I can work on that if you would like. It shouldn't be to hard to have make iterate over the passed in parameters to set things  correctly. Would need to update the lv_conf.h file to check for set macros and if they are unset then set them otherwise leave them how they are already set.

kdschlosser:
we an now see the error in the last run that took place.

```
File "<stdin>", line 16, in <module>
RuntimeError: Logging in LVGL MUST be disabled to run the tests.
```

kdschlosser:
OK so the logging error is solved. Next error is that the binding doesn't have the same fonts  used in the test. So that needs to be corrected. That is something you are going to have to do because the changes will need to be merged in order for it to work properly.

kdschlosser:
The test is closing gracefully and without any errors like open files or orphaned subprocesses. The changes I made to the CI didn't spit out any errors at all. have to check and see if the build artifacts actually got uploaded.

kisvegabor:
Still not good. :( 

Does it pass locally for you using the screen shot from the `refr_imgs` folder?

kisvegabor:
I've disabled [logging](https://github.com/lvgl/lv_binding_micropython/commit/f4604ead8db75e141b23b5e1bde90158ea734d47) manually to be sure, re-run the tests, but still fails.

kdschlosser:
It's failing this time from a frame data mismatch. I don't know enough about git workflows and I don't know why the artifacts are not populating. I was hoping you would be able to figure that part out. The artifacts will contain the images of what is being output. need to check them. to see what's wrong.

kdschlosser:
i suspect it is with the font I added that was missing. It is the 24 point one. Maybe changing the config file in the binding so the font is available.

kisvegabor:
I enabled the font but it still fails.

It's quite simple to upload the artifacts. See an example [here](https://github.com/lvgl/lvgl/blob/master/.github/workflows/ccpp.yml#L39-L46). If you tell me where is the file to upload I can add it.

kdschlosser:
I added to the workflow already. The path is there. everything in the artifacts directory is what needs to be uploaded. That is where all frame data in binary form and also the frame png files get saved to. It also makes an animated png of all of the frames so you can see what is happening in an animation too.

kdschlosser:
https://github.com/lvgl/lvgl/pull/4997/commits/29b1f76ba36a19bedcaaf235bf8f0ca232ed7002#diff-94082afd73e0a27785f11ed4162f6281943bd204d66dd7343745e850081e8e97R83

kdschlosser:
The logging is no longer the issue. I also fixed how the errors are handled so we actually know why it is failing. now it is correctly reporting that the failure is being caused by no matching frame data being found.

You can check the artifacts folder when you run it locally to see what is going on with the frame data.

kisvegabor:
Ah, I see now. There should be only one frame, as there is only on reference image. I see that the opacity of the button changes on each frame. Obviously all can't match the same reference image. 

I've overwritten the original `binding.png` with `frame9.png` and teh C test were passing, meaning `frame9.png` is correct.

So, please compare only the very last image and it should work.

kdschlosser:
It's a timing thing. If you rename the result.apng file to result.png and open the file in your browser you will then see what is going on.

The problem is a timing thing. The test has an animation that is running on that one button. That is the reason why I have it output the animation png file. Smart huh???? The test simply isn't running long enough to catch the needed frame. No need to change the target image file if you have other tests that use that image. I can just extend the length of time the test runs for to capture the needed frame.

kisvegabor:
I feel like it's more complex than what we need. You can just wait for long enough, and export a single image.

kdschlosser:
the BIN files is the actual RGB pixel data in RAW format. I had it output that as well so if there is some kind of a byte order issue you would be able to see it in there.

can never have too much output when running tests. There is no need to supply the --debug switch to the command line either. The test will always output a debugging log file to that artifacts folder for inspection. I corrected the issues with the exception stack getting passed to the testing framework. If there is an error it will show up in that log file as well as in the failure reason.

It's actually a pretty cool design because I don't have to worry about setting exit codes to tell the test program why the failure occurred and it not providing any useful information on what exactly has failed. I am able to capture the actual exception stack and pass it right to the test code so that stack data gets set for the test failure reason. This will tell you the exact location and reason for the failure.

kdschlosser:
what is the length of time that we need to wait in order to capture the right frame? Technically speaking the code shouldn't have to "wait". I should be able to use lv_tick_inc and set the time that has passed manually and then collect the frame correct?

kisvegabor:
I like to keep things simple and do what is really needed but I don't want to limit you creativity 🙂

It should work before taking the last snapshot:
```c
lv_tick_inc(1000);
lv_refr_now(NULL);
```

Let's merge it first (when the test passes), but in an update [here](https://github.com/lvgl/lvgl/pull/4997/files#diff-ec5b6a5b31ffa0d1dfa7410aafb547d8478cdac1a1ecdcd05acd7a3d147eec3fR255) some delay and `lv_timer_handler()` should be added. See [here](https://github.com/lvgl/lvgl/blob/master/tests/src/test_cases/test_bindings.c#L168) and [here](https://github.com/lvgl/lvgl/blob/master/tests/src/lv_test_helpers.c#L5-L10). 
Adding `gc.collect` on a lot of places would be great as well.

kdschlosser:
You have to remember that I wrote this as a test framework for the binding because I was under the impression that is what it was going to be used for. I did not realize the intention was for it to be used in LVGL and only for the purposes of capturing this one single frame. I had thought the purpose of doing all of this was to completely remove LVGL's reliance on the binding.


So I needed to adjust the script when I fully understood its intended purpose. Personally I think that instead of having this test placed into LVGL. the full test framework that I wrote should be put into the binding. The comparison image and the actual code that runs in MicroPython would reside with LVGL but LVGL would be able to pass the path to the files via the command line to the test framework located in the binding.

kdschlosser:
`lv_test_wait(300);` is an external function call and I have know knowledge of what is actually taking place when the call is made. You handed me a file and said "port this to MicroPython" and I did that with what was handed to me. Since that function doesn't exist in the binding code it was not added.

I BLAME YOU!!  I'm joking. I should have asked about it and my intention was to do that and I forgot to ask.

I am going to remove the looping code at the end and call lv_tick_inc setting the increment to 300 and then call lv_task_handler. That will allow the single frame to get captured and I am hoping it will be the correct one. I do not know exactly how the animations work. Will it fast forward the position of the animation based on the time fed into lv_tick_inc or will it advance only a single step? Guess I will find out.

kisvegabor:
The animation will fast forward. It always uses the absolute time and calculates where the animation should be at that time.

I leave it up to you how the framework is organized. What really matters for me:
- compare the screenshot with the one from the `ref_imgs` folder
- have the script itself in the `lvgl` repo to make it easy to adjust in case of LVGL changes.

kdschlosser:
OK it is working now using the image that you originally supplied. It only captures the single frame. I am going to remove the apng requirement in my next commit so don't merge this yet.

kisvegabor:
Working!!!! :heart: :tada: Thank you!

I'll merge it when all CI passes and after that remove the MP examples from LVGL.

kdschlosser:
OK so the only thing that needs to be located in LVGL is the actual micropy.py script and not the testing framework. The framework should not need to be adjusted at all. For the time being so you can get this release done leave the framework with LVGL. Once I get a framework in place that will handle running a more complex set of test cases for the binding we will be able to remove the framework from LVGL and adjust the CI accordingly.

The framework in the binding once finished will have the ability to run tests by passing the test location via the command line. I an not a big fan of having duplicate code especially if the code is in more than a single repo. While the framework code would not be identical between LVGL and the binding, it would be close enough to where is should be all in a single place so it can be maintained easily.

does that sound like a plan?

kisvegabor:
If you are planning to add more tests to the micropython repo and use the same framework it absolutely makes sense to move the framework there. 

Ideally there should be only one framework, but you will see in the practice how feasible it is. 

So, yes, it sounds like a plan! :+1:

kdschlosser:
sorry about this being a pain to do. That was my failure in fully understanding what your goal was and me writing the test program for a different use. That is my fault, I should have asked more questions instead of making assumptions. 

I would like to open up a discussion somewhere about handling the testing of the framework and gets some peoples input on it. There could be some edge case scenarios we did not think of.


I have a really cool idea tho. Remember how I mentioned the sheer size of the examples page in the documentation? It requires an extremely large amount of data to be transferred. This is crafty what I am thinking. Change the examples in the documentation to be static images, when the mouse hovers over the image it will change to the live example. This way the example only gets sent if it is actually being viewed. That very same static image would be the image used in the test for the frame buffer comparison.. This is an ideal way to provide some better performance when loading the documentation. It gives the documentation a place holder until the binding has been released so it has the matching LVGL version and it will air in giving a more complete set of tests because we can use the examples as test cases. 

If a change to an example causes the displayed output to change, change the static image to match the output and that will take care of the tests without anyone having to mess about with altering code specific to testing. Examples will have to be updated if there are API changes, the tests are the way to know which examples would need to be updated due to a failure in that test.

The framework will run all tests even if one of them fails. it will tell you which ones failed making it easier to locate examples that need to be updated. No need to go and check each and every example to ensure functionality. that is what the tests will do for you.

That really simplifies things big time. It also ensures that the examples are always working. I will tell you that if I go to look at a piece of software and the examples for the software don't function I won't even bother going any further. I will go and look at something else. If the examples don't function it is a really good indicator as to how things are going to work in the software. 

Those examples should be 100% working when a release of the binding is made. PERIOD! no exceptions and no excuses. Remember when I first started using LVGL and one of the first issues I opened up was a list of all of the examples that didn't work right. It was a really long list i have to say.... Doesn't make things look good for LVGL or the binding when that kind of thing happens.

kdschlosser:
we need  streamline things and make maintaining easier to do. Right now it is not an easy thing. There is a very large list that has to be done through to make sure everything is running right before a release is made. If we can automate some of that to tell us what is broken it should be done. Right now LVGL for the most part has a matching set of examples with the binding. One of the things that will be looked at in the tests is the LVGL examples folder. it needs to align with the examples folder in the binding. This is only because of the documentation in LVGL using the examples in the binding. so they need to match up so when an example in LVGL gets added or removed you don't end up with a dead link or an example that never runs.

kdschlosser:
I am going to write a script that goes over the documentation in LVGL and updates all of the API changes. You don't have to worry about going in and changing every reference of `lv_disp` to `lv_display`. The script will use the header file you made in order to know what the patterns are and what the replacement is.

That will make things a whole lot easier on you. It will take me less than an hour to key that up. It would take you days to manually make those changes.

kdschlosser:
another thing I can do to further automate the process is writing a script that looks at every commit that has taken place since the last release. It will collect a list of header files that have changed and search for references to the objects contained in each header file. if there are references in one of the documentation files then the file name will be printed out as needing to be updated.

If there are new header files that have been added and no references are found to any of the objects in that header file the file name of the header will be printed out stating that documentation needs to be written. 

This is only for the statically written documentation. The stuff that is generated for the API section would not be included in this.

This script would not run as part of the documentation build system It would be standalone so you can go over what does and doesn't need to be done. 

would something like that be useful?

kdschlosser:
@kisvegabor 

The CI's have passed.

kdschlosser:
I had thought the examples in the docs were generated from the binding. It's good that they are not. so you are correct then they do not need to be a 1 to 1. Adding the examples to the tests is not a big deal to do and it would ensure that all of the examples are in working order before a release is made. It would be a good thing to run them through the tests along with any other things that should be tested.

kisvegabor:
I think it's important to find a sweet spot between enough examples to be useful, but not too much for maintenance.

kdschlosser:
> I think it's important to find a sweet spot between enough examples to be useful, but not too much for maintenance.

agreed.

Perhaps you should consider adding a new forum area for people to submit code examples of how to do different things. That kind of information is always helpful to have.

kdschlosser:
I am glad I was able to hammer out this test program for you. couple little hiccups to deal with but nothing too hard to fix. Nice thing is it will catch any and all errors and I didn't have to specifically code in handling of the plethora of errors that can happen.

PR #4998, opened 2023-12-13
lvgl:add lv_display_remove_event_cb_with_user_data

jianglianfang:

Add lv_display_remove_event_cb_with_user_data becuase in many cases we don’t know the index corresponding to the event.

### Checkpoints
 
### DIFF
--- a/src/display/lv_display.c
+++ b/src/display/lv_display.c
@@ -699,6 +699,25 @@ bool lv_display_delete_event(lv_display_t * disp, uint32_t index)
     return lv_event_remove(&disp->event_list, index);
 }
 
+uint32_t lv_display_remove_event_cb_with_user_data(lv_display_t * disp, lv_event_cb_t event_cb, void * user_data)
+{
+    LV_ASSERT_NULL(disp);
+
+    uint32_t event_cnt = lv_display_get_event_count(disp);
+    uint32_t removed_count = 0;
+    int32_t i;
+
+    for(i = event_cnt - 1; i >= 0; i--) {
+        lv_event_dsc_t * dsc = lv_display_get_event_dsc(disp, i);
+        if(dsc && dsc->cb == event_cb && dsc->user_data == user_data) {
+            lv_display_delete_event(disp, i);
+            removed_count ++;
+        }
+    }
+
+    return removed_count;
+}
+
 lv_result_t lv_display_send_event(lv_display_t * disp, lv_event_code_t code, void * param)
 {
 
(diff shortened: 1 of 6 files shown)


src/display/lv_display.h
```diff
@@ -452,6 +452,8 @@ lv_event_dsc_t * lv_display_get_event_dsc(lv_display_t * disp, uint32_t index);
  */
 bool lv_display_delete_event(lv_display_t * disp, uint32_t index);
 
+uint32_t lv_display_remove_event_cb_with_user_data(lv_display_t * disp, lv_event_cb_t event_cb, void * user_data);
```
kisvegabor, line 462: Please add comments

kisvegabor (commented):
We have the index based approach to make LVGL better work with MicroPython, however for C user it's really useful.

Could you add the same for `lv_obj` and `lv_indev`?

src/display/lv_display.h
jianglianfang: OK, done

jianglianfang: sure, I will add it for lv_obj and lv_indev soon in another commit

src/display/lv_display.c
```diff
@@ -699,6 +699,25 @@ bool lv_display_delete_event(lv_display_t * disp, uint32_t index)
...
 
+uint32_t lv_display_remove_event_cb_with_user_data(lv_display_t * disp, lv_event_cb_t event_cb, void * user_data)
+{
+    LV_ASSERT_NULL(disp);
+
+    uint32_t event_cnt = lv_display_get_event_count(disp);
+    uint32_t i;
+    uint32_t removed_count = 0;
+
+    for(i = 0; i < event_cnt; i++) {
```
kisvegabor: It just came to my mind that it won't work well if there are multiple events to delete after each other. Let's say we want to remove all `event_C_*` from this list:
0. event_A_1
1. event_B_1
2. event_C_1
3. event_C_2
4. event_C_3

When C_1 is deleted with i==2 the list will change like this:
0. event_A_1
1. event_B_1
2. event_C_2
3. event_C_3

In the next iteration `i`  will be `3`, so C_2 is skipped. 

A simple solution would be something like `if(i > 0) i--` after each deletion, or start from the end of the list.

jianglianfang: Yeah, I think I might have overlooked that.

kisvegabor (approved):
Looks good to me, thank you!

PR #4999, opened 2023-12-13
lvgl:add vsync_poll

jianglianfang:

add vsync poll to get the time of vysnc.

### Checkpoints
- [x ] Run `code-format.py` from the scripts folder. [astyle](http://astyle.sourceforge.net/install.html) needs to be installed.
 
### DIFF
--- a/src/dev/nuttx/lv_nuttx_libuv.c
+++ b/src/dev/nuttx/lv_nuttx_libuv.c
@@ -28,6 +28,7 @@ typedef struct {
     int fd;
     bool polling;
     uv_poll_t fb_poll;
+    uv_poll_t vsync_poll;
 } lv_nuttx_uv_fb_ctx_t;
 
 typedef struct {
@@ -49,6 +50,7 @@ static void lv_nuttx_uv_timer_cb(uv_timer_t * handle);
 static int  lv_nuttx_uv_timer_init(lv_nuttx_uv_t * uv_info, lv_nuttx_uv_ctx_t * uv_ctx);
 static void lv_nuttx_uv_timer_deinit(lv_nuttx_uv_ctx_t * uv_ctx);
 
+static void lv_nuttx_uv_vsync_poll_cb(uv_poll_t * handle, int status, int events);
 static void lv_nuttx_uv_disp_poll_cb(uv_poll_t * handle, int status, int events);
 static void lv_nuttx_uv_disp_refr_req_cb(lv_event_t * e);
 static int  lv_nuttx_uv_fb_init(lv_nuttx_uv_t * uv_info, lv_nuttx_uv_fb_ctx_t * fb_ctx);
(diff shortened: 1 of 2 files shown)


src/dev/nuttx/lv_nuttx_libuv.h
```diff
@@ -25,6 +25,8 @@ extern "C" {
  *      DEFINES
  *********************/
 
+#define LV_EVENT_VSYNC (_LV_EVENT_LAST + 1)
```
XuNeo: Let's define it in `lv_event.h`.

kisvegabor: You can use `lv_event_register_id` too.

In the practice what do you exactly do in the `LV_EVENT_VSYNC` event?

XuNeo: To optimize the refresh rate, it's crucial to synchronize the entire refresh process with the display's hardware vsync signal. If not, the refresh frequency driven by CPU could fall out of sync with display's frequency, eventually resulting in the loss of a frame in period.

kisvegabor: I see, so you call the display refresh timer manually on VSYNC event. 

I wonder if an event is really required for this. What about having a helper `lv_nuttx_refresh_display(disp)` which is called from `lv_nuttx_uv_vsync_poll_cb` and it calls `_lv_display_refr_timer()`. Or maybe `lv_nuttx_uv_vsync_poll_cb` could call `_lv_display_refr_timer()` directly.

I'm not using NuttX driver on daily basis, so probably you know it better, I'm just asking. :slightly_smiling_face:

XuNeo: The VSYNC event is better when we consider multiple LVGL instances. The virtual displays(instances) rely on VSYNC signal to sync with physical display.

kisvegabor: > The VSYNC event is better when we consider multiple LVGL instances. 

Why? Calling `_lv_display_refr_timer(disp)` on vsync interrupt seems multi-instance independent.

Note that, if `lv_event_register_id` is used `LV_EVENT_VSYNC` needs to be added to `lv_globals` for multi instance support.

XuNeo: The multiple LVGL instances will only have one connected to physical display. Thus adding a LVGL level VSYNC can easily forward this information to other instances.

Besides, the refresh timer should also be driven by VSYNC signal and we can unify this behavior.

kisvegabor: I see. Can we reuse `LV_EVENT_REFRESH` for this? [By definition](https://github.com/lvgl/lvgl/blob/master/src/misc/lv_event.h#L79) this event is never sent by LVGL but it's for the user to indicate something requires a "refresh".

bjsylvia: `LV_EVENT_VSYNC` is only defined as an event for receiving hardware vblank signal, and it is not strongly bound to actual refresh or other tasks. It is recommended to keep the current precise definition.

XuNeo: VSYNC is a standard event for GUI. It would be better not mixing it with `REFRESH`. The `refresh` should be an action driven by VSYNC.

For other GUI libs, I can find info about VSYNC too.
https://support.touchgfx.com/docs/basic-concepts/operating-system#touchgfx-os-wrappers
https://source.android.com/docs/core/graphics/implement-vsync

kisvegabor: In this case le's create a new event code for in in `lv_event.h`

jianglianfang: moved LV_EVENT_VSYNC to lv_event.h

XuNeo (approved):

PR #5012, opened 2023-12-14
fix(canvas): fix canvas unaligned access buffer

FASTSHIFT:

Use draw buf interface.

cc @XuNeo 

### Checkpoints
 
### DIFF
--- a/src/widgets/canvas/lv_canvas.c
+++ b/src/widgets/canvas/lv_canvas.c
@@ -69,14 +69,15 @@ void lv_canvas_set_buffer(lv_obj_t * obj, void * buf, int32_t w, int32_t h, lv_c
     LV_ASSERT_NULL(buf);
 
     lv_canvas_t * canvas = (lv_canvas_t *)obj;
+    uint32_t stride = lv_draw_buf_width_to_stride(w, cf);
 
-    canvas->buf_unaligned = buf;
-    canvas->dsc.header.cf = cf;
-    canvas->dsc.header.w  = w;
-    canvas->dsc.header.h  = h;
-    canvas->dsc.header.stride  = lv_draw_buf_width_to_stride(w, cf);
-    canvas->dsc.data      = lv_draw_buf_align(buf, cf);
-    canvas->dsc.data_size = w * h * lv_color_format_get_size(cf);
+    canvas->buf_unaligned     = buf;
+    canvas->dsc.header.cf     = cf;
+    canvas->dsc.header.w      = w;
+    canvas->dsc.header.h      = h;
+    canvas->dsc.header.stride = stride;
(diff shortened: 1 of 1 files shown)


src/widgets/canvas/lv_canvas.c
```diff
@@ -69,14 +69,15 @@ void lv_canvas_set_buffer(lv_obj_t * obj, void * buf, int32_t w, int32_t h, lv_c
...
+    uint32_t stride = lv_draw_buf_width_to_stride(w, cf);
 
-    canvas->buf_unaligned = buf;
-    canvas->dsc.header.cf = cf;
-    canvas->dsc.header.w  = w;
-    canvas->dsc.header.h  = h;
-    canvas->dsc.header.stride  = lv_draw_buf_width_to_stride(w, cf);
-    canvas->dsc.data      = lv_draw_buf_align(buf, cf);
-    canvas->dsc.data_size = w * h * lv_color_format_get_size(cf);
+    canvas->buf_unaligned     = buf;
```
XuNeo, line 74: We need a better way to deal with unaligned buffer. It requires user understand the limitation behind it.

I suggest adding API `lv_canvas_set_buffer(lv_obj_t *, lv_draw_buff_t *)`;

XuNeo (commented):
It looks good for now. Let's migrate to `draw_buf` API later.

src/widgets/canvas/lv_canvas.c
kisvegabor: I understand the motivation and agree to be more transparent about it and not do hidden magic.

In `lv_canvas_set_buffer(lv_obj_t *, lv_draw_buff_t *)`  `draw_buf` is just a local variable which can be dropped after calling `lv_canvas_set_buffer()`?.
Usually I'm not big fan of `struct`s used to initialize something because it's not full intuitive what the user needs to do with the `struct`:  can be local, needs to malloced and freed on delete, can it be reused for the next initialization, can be modified? It can be documented, but it'd be better to make it intuitive.

How can we initialize a `draw_buf_t` by using a user provided buffer? And how do we deal with address alignement?

XuNeo: > How can we initialize a `draw_buf_t` by using a user provided buffer? And how do we deal with address alignement?

So we should not init a `draw_buf_t` manually, and force to use `lv_draw_buf_create`.


> it's not full intuitive what the user needs to do with the `struct`

I agree with you, but I cannot see a better method since we have so many limitations.

kisvegabor: I like `LV_DRAW_BUF_DEFINE` in #5022. With that the user can still use `lv_draw_buf_create` is she needs.

FASTSHIFT:
PR continues here: https://github.com/lvgl/lvgl/pull/5022

PR #5020, opened 2023-12-15
Implemented ST7789 LCD driver

zjanosy:
Implemented driver model for LCD controllers connected via SPI/parallel port. Implemented support for ST7789.


src/dev/display/lcd/lv_lcd_panel_commands.h
```diff
@@ -0,0 +1,54 @@
+/*
```
kisvegabor: As these are specific to st7789, please move these commands in `lv_st7789.c`.

List only the really used commands should be enough. If someone needs to use a custom command they need to look into the datasheet anyway.

src/dev/display/st7789/lv_st7789.c
```diff
@@ -0,0 +1,363 @@
...
+
+/**********************
+ *  STATIC VARIABLES
+ **********************/
+
+/**********************
+ *      MACROS
+ **********************/
+
+inline lv_st7789_driver_t *lv_st7789_get_driver(lv_display_t *disp)
```
kisvegabor:
```suggestion
static inline lv_st7789_driver_t *lv_st7789_get_driver(lv_display_t *disp)
```

And please move it to the static functions

src/dev/display/st7789/lv_st7789.c
```diff
@@ -0,0 +1,363 @@
...
+{
+	return (lv_st7789_driver_t *)lv_display_get_driver_data(disp);
+}
+
+/**********************
+ *   GLOBAL FUNCTIONS
+ **********************/
+
+lv_display_t *lv_st7789_create(uint32_t hor_res, uint32_t ver_res, lv_st7789_send_cmd_cb_t send_cmd, lv_st7789_send_cmd_cb_t send_color)
```
kisvegabor:
```suggestion
lv_display_t *lv_st7789_create(uint32_t hor_res, uint32_t ver_res, lv_st7789_send_cmd_cb_t send_cmd_cb, lv_st7789_send_cmd_cb_t send_color_cb)
```

src/dev/display/st7789/lv_st7789.c
```diff
@@ -0,0 +1,363 @@
...
+    	}, 1);
+    send_cmd(drv, LCD_CMD_DISPON, NULL, 0);
+}
+
+/**
+ * Turn display on/off
+ * @param drv			LCD driver object
+ * @param on			false: turn off, true: turn on
+ */
+static void set_onoff(lv_st7789_driver_t * drv, bool on)
```
kisvegabor: Please remove it if it's not used.

src/dev/display/st7789/lv_st7789.c
```diff
@@ -0,0 +1,363 @@
...
+    /* TODO: implement resolution change */
+    LV_UNUSED(hor_res);
+    LV_UNUSED(ver_res);
+
+    /* handle rotation */
+    switch (rot) {
+	case LV_DISPLAY_ROTATION_0:
+		set_swap_xy(drv, false);
+		set_mirror(drv, false, false);
+#if xCONFIG_LCD_TOUCH_ENABLED
```
kisvegabor: Remove left over code. Touch shouldn't be handled here.

kisvegabor (commented):
Thank you, Zoltan. It looks good in general, just a few minor issue caught by the CI should be fixed. 

Please run `code-format.py` in the scripts folder to format the code.

src/dev/display/lcd/lv_lcd_panel_commands.h
zjanosy: These commands are common for various controllers, not ST7789 specific.

kisvegabor: But not all, so we can't put it into a generic LCD file. 
It seems roughly 10 commands are really used in st7789.c. I think it's okay to define them locally to avoid any conflicts later.

src/dev/display/st7789/lv_st7789.h
```diff
@@ -0,0 +1,86 @@
...
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+/*********************
+ *      INCLUDES
+ *********************/
+
+#include "src/display/lv_display.h"
```
kisvegabor: It should be included with a relative path.

kisvegabor:
Not stale, will be finished in January.

src/dev/display/ili9341/lv_ili9341.c
```diff
@@ -0,0 +1,118 @@
+/**
+ * @file lv_ili9341.c
+ *
+ */
+
+/*********************
+ *      INCLUDES
+ *********************/
+#include "lv_ili9341.h"
+
```
kisvegabor, line 10: Please add `LV_USE_ILI9341` guard here and and also in `lv_conf_template.h` (scripts/lv_conf_internal_gen.py needs to run as well)

src/dev/display/ili9341/lv_ili9341.c
```diff
@@ -0,0 +1,118 @@
...
+ **********************/
+
+static void send_cmd(lv_lcd_generic_mipi_driver_t *drv, uint8_t cmd, uint8_t *param, size_t param_size);
+static lv_lcd_generic_mipi_driver_t *get_driver(lv_display_t *disp);
+
+/**********************
+ *  STATIC CONSTANTS
+ **********************/
+
+// init commands based on LovyanGFX
```
kisvegabor: Use `/**/` comemnts

src/dev/display/ili9341/lv_ili9341.h
```diff
@@ -0,0 +1,88 @@
...
+
+/**
+ * Prototype of a platform-dependent callback to transfer commands and data to the LCD controller.
+ * @param disp			display object
+ * @param cmd			command buffer (can handle 16 bit commands as well)
+ * @param cmd_size		number of bytes of the command
+ * @param param			parameter buffer
+ * @param param_size	number of bytes of the parameters
+ */
+typedef void (*lv_ili9341_send_cmd_cb_t)(lv_display_t *disp, uint8_t *cmd, size_t cmd_size, uint8_t *param, size_t param_size);
```
kisvegabor:
```suggestion
typedef lv_ili9341_send_cmd_cb_t lv_lcd_send_cmd_cb_t;
```

src/dev/display/st7735/lv_st7735.c
```diff
@@ -0,0 +1,170 @@
...
+
+static void send_cmd(lv_lcd_generic_mipi_driver_t *drv, uint8_t cmd, uint8_t *param, size_t param_size);
+static lv_lcd_generic_mipi_driver_t *get_driver(lv_display_t *disp);
+
+/**********************
+ *  STATIC CONSTANTS
+ **********************/
+
+static const uint8_t init_cmd_list[] = {
+#if 0
```
kisvegabor: `#if 0`s should be removed and only the final code should remain.
If the other options might be used by the users as well, and API functions or other direct way to adjust shall be provided.

src/dev/display/lcd/lv_lcd_generic_mipi.h
```diff
@@ -0,0 +1,224 @@
...
+
+/* flags for lv_lcd_xxx_create() */
+#define LV_LCD_FLAG_NONE								0x00000000UL
+#define LV_LCD_FLAG_MIRROR_X							0x00000001UL
+#define LV_LCD_FLAG_MIRROR_Y							0x00000002UL
+#define LV_LCD_FLAG_BGR									0x00000008UL
+#define LV_LCD_FLAG_RGB666								0x00000010UL
+
+/* command list */
+#define LV_LCD_CMD_SPECIAL	0xff
```
kisvegabor: As it's used only as a delay:
```suggestion
#define LV_LCD_CMD_DELAY_MS	0xfe
```

kisvegabor (commented):
Thank you for the update. 

Please also run `scripts/code-format.py`.

PGNetHun:
Hello!
Thx for the drivers!
Could you please provide example / doc how to use it, how to initialize?
I see that callbacks must be given (example: `send_color_cb` ), but don’t know what it should be exactly, how to implement, for example on ESP32 where SPI data  can be sent with DMA.
My other question is about RGB565 byte swap: ST7789 display needs that in ESP32 platform, and I don’t see where it is done.
Thx again for the display drivers, I think when this PR is merged, and MicroPython is aligned to use it, then it will further improve FPS (or less CPU usage) in MicroPython binding, and replacing the actual hybrid drivers.

src/dev/display/st7796/lv_st7796.c
```diff
@@ -0,0 +1,135 @@
...
+ * @file lv_st7796.c
+ *
+ */
+
+/*********************
+ *      INCLUDES
+ *********************/
+#include "lv_st7796.h"
+
+#if LV_USE_ST7796
```
kisvegabor, line 11: Please add these guards to the header files as well.

kisvegabor (commented):
Sometime the code formatter needs to be tun multiple times to remove all double empty lines. 

As these controllers don't contain any HW dependent parts we can enable them for the CI, so that they will be built and we will see if there are any compile time warnings,
So please enable the controllers [here](https://github.com/lvgl/lvgl/blob/master/tests/src/lv_test_conf_full.h#L107).

Please add docs for each controller. In an example section some dummy callbacks shall be added well.

zjanosy:
@PGNetHun 

> Could you please provide example / doc how to use it, how to initialize? 

I will add the documentation soon.

> I see that callbacks must be given (example: `send_color_cb` ), but don’t know what it should be exactly, how to implement, for example on ESP32 where SPI data can be sent with DMA.

I will include a sample implementation. Currently the drivers have been implemented only on STM32, but ESP32 will be added soon.

> My other question is about RGB565 byte swap: ST7789 display needs that in ESP32 platform, and I don’t see where it is done.

On the STM32 platform the SPI controller can swap the bytes. I need to check how the ESP32 SPI controller handles this. We may need to do it in software.

kisvegabor:
>> My other question is about RGB565 byte swap: ST7789 display needs that in ESP32 platform, and I don’t see where it is done.

> On the STM32 platform the SPI controller can swap the bytes. I need to check how the ESP32 SPI controller handles this. We may need to do it in software.

As it depends on the periphery if it can swap the bytes, `lv_draw_sw_rgb565_swap` can be called from the  `send_color_cb` if needed.

PGNetHun:
> @PGNetHun According to our plans after these drivers [this PR](https://github.com/lvgl/lvgl/pull/4507) will be finalized. Does it help with MicroPython somehow?

Well, I'm a bit lost. 
There is an another thread about (MicroPython) display drivers: https://forum.lvgl.io/t/micropython-display-drivers-part-2/14131

... so at the moment I see 3 different approaches:
1. the current MicroPython specific drivers (including the ESP32 hybrid display driver `ili9xxx.py`): https://github.com/lvgl/lv_binding_micropython/tree/master/driver
2. this PR and related [ESP driver](https://github.com/lvgl/lvgl/pull/4507)
3. the above mentioned forum thread: https://forum.lvgl.io/t/micropython-display-drivers-part-2/14131

I prefer having C display drivers due to high performance requirement (and less C <-> MicroPython context switch during display flushes)

zjanosy:
@PGNetHun 

> Well, I'm a bit lost.
> There is an another thread about (MicroPython) display drivers: https://forum.lvgl.io/t/micropython-display-drivers-part-2/14131
> 
> ... so at the moment I see 3 different approaches:
> 
> the current MicroPython specific drivers (including the ESP32 hybrid display driver ili9xxx.py): https://github.com/lvgl/lv_binding_micropython/tree/master/driver
> this PR and related https://github.com/lvgl/lvgl/pull/4507
> the above mentioned forum thread: https://forum.lvgl.io/t/micropython-display-drivers-part-2/14131
> 
The ESP-LCD driver was an initial attempt to implement a unified LCD driver architecture for LVGL. However, we decided to develop the idea further, and the current PR is the result of this. The ESP-LCD driver is now obsolete, and will be replaced by an example implementation of the current driver model.
I don't know much about the MicroPython drivers, but hopefully we can move towards a common solution for all platforms. The goal is to make LVGL work out of the box with the most common LCD controllers, without the need for using 3rd party drivers.

docs/integration/driver/display/st7789.rst
```diff
@@ -0,0 +1,248 @@
+=============================
+ST7789 LCD Controller driver
+=============================
+
+Overview
+-------------
+
+The `ST7789 <>`__ SOC is a popular TFT LCD controller chip made by Sitronix. It supports TFT LCD panels with a maximum resolution of 240x320 pixels, and 262k (18 bit) colors.
```
kisvegabor: Link to the datasheet?

docs/integration/driver/display/st7789.rst
```diff
@@ -0,0 +1,248 @@
...
+		...
+	}
+
+	/* Send large array of pixel data to the LCD. If necessary, this function has to do the byte-swapping. This function can do the transfer in the background. */
+	int32_t my_lcd_send_color(lv_display_t *disp, uint8_t *cmd, size_t cmd_size, uint8_t *param, size_t param_size)
+	{
+		...
+	}
+
+The only difference between the ``my_lcd_send_cmd()`` and ``my_lcd_send_color()`` functions is that ``my_lcd_send_cmd()`` is used to send short commands and it is expected
```
kisvegabor: Can we have a send_color function like this by sending the required command before calling it?
```c
int32_t my_lcd_send_color(lv_display_t *disp, uint8_t * buf, size_t buf_size);
```

docs/integration/driver/display/st7789.rst
```diff
@@ -0,0 +1,248 @@
...
+The driver can be used to create multiple displays. In such a configuration the callbacks must be able to distinguish between the displays. Usually one would
+implement a separate set of callbacks for each display. Also note that the user must take care of arbitrating the bus when multiple devices are connected to it.
+
+Example
+-------
+
+This is an example implementation on an STM32F746 using STM32CubeIDE and the ST HAL libraries.
+
+.. code:: c
+
```
kisvegabor, line 55: Let's create a docs page for the generic mipi driver as well and move the example there too.

docs/integration/driver/display/st7789.rst
```diff
@@ -0,0 +1,248 @@
...
+Create flags
+^^^^^^^^^^^^
+
+The third argument of the ``lv_st7789_create()`` function is a flag array. This can be used to configure the orientation and RGB ordering of the panel if the default settings
+do not work for you. In particular, the ST7789 driver accepts the following flags:
+
+.. code:: c
+
+	#define LV_LCD_FLAG_NONE                                0x00000000UL
```
kisvegabor: We don't need to show the values of the flags here, just list their names.

docs/integration/driver/display/st7789.rst
zjanosy: Currently the my_lcd_send_cmd() function sends both the command and the data, and there is no way to send only a command.
We could have defined a send_cmd() and a send_data() callback, and pass a flag to send_data() to tell that the data contains pixel data, which should be sent using DMA (and maybe should be byte-swapped).
We could even pass a 'transfer_ready_cb' to send_data(), which would be used to call the lv_display_flush_ready() from the driver (currently this must be called by the user, which is not the very best solution). It is not difficult to modify the code, if you prefer this method.

kisvegabor: Let's leave it as it is.

kisvegabor (commented):
The updated docs looks good!

src/dev/display/st7735/lv_st7735.c
```diff
@@ -0,0 +1,128 @@
+/**
+ * @file lv_st7735.c
+ *
+ */
+
+/*********************
+ *      INCLUDES
+ *********************/
+#include "lv_st7735.h"
```
kisvegabor, line 9: The docs is still missing for this one.

kisvegabor:
Please [enable these controllers](https://github.com/lvgl/lvgl/blob/master/tests/src/lv_test_conf_full.h#L107) for tests, so they can be built by the CI.

kisvegabor (approved):
Looks good, thank you. 
If anything comes up probably we can extend these without API braking the API.

PR #5032, opened 2023-12-18
fix(draw_vector): fix build break when thorvg disabled

FASTSHIFT:

```
lvgl/demos/lv_demos.c:74:37: error: ‘lv_demo_vector_graphic’ undeclared here (not in a function)
   74 |     { "vector_graphic", .entry_cb = lv_demo_vector_graphic },
      |                                     ^~~~~~~~~~~~~~~~~~~~~

/usr/bin/ld: lib/liblvgl.a(lv_draw_sw.c.o): in function `execute_drawing':
lvgl/src/draw/sw/lv_draw_sw.c:338: undefined reference to `lv_draw_sw_vector'
```

cc @onecoolx 

### Checkpoints
 
### DIFF
--- a/src/draw/sw/lv_draw_sw.c
+++ b/src/draw/sw/lv_draw_sw.c
@@ -333,7 +333,7 @@ static void execute_drawing(lv_draw_sw_unit_t * u)
         case LV_DRAW_TASK_TYPE_MASK_RECTANGLE:
             lv_draw_sw_mask_rect((lv_draw_unit_t *)u, t->draw_dsc, &t->area);
             break;
-#if LV_USE_VECTOR_GRAPHIC
+#if LV_USE_VECTOR_GRAPHIC && (LV_USE_THORVG_EXTERNAL || LV_USE_THORVG_INTERNAL)
         case LV_DRAW_TASK_TYPE_VECTOR:
             lv_draw_sw_vector((lv_draw_unit_t *)u, t->draw_dsc);
             break;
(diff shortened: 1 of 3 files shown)


src/draw/sw/lv_draw_sw.c
```diff
@@ -333,7 +333,7 @@ static void execute_drawing(lv_draw_sw_unit_t * u)
         case LV_DRAW_TASK_TYPE_MASK_RECTANGLE:
             lv_draw_sw_mask_rect((lv_draw_unit_t *)u, t->draw_dsc, &t->area);
             break;
-#if LV_USE_VECTOR_GRAPHIC
+#if LV_USE_VECTOR_GRAPHIC && (LV_USE_THORVG_EXTERNAL || LV_USE_THORVG_INTERNAL)
```
XuNeo, line 336: Should we use LV_USE_THORVG instead.

src/draw/sw/lv_draw_sw.h
```diff
@@ -162,7 +162,7 @@ void lv_draw_sw_transform(lv_draw_unit_t * draw_unit, const lv_area_t * dest_are
                           int32_t src_w, int32_t src_h, int32_t src_stride,
                           const lv_draw_image_dsc_t * draw_dsc, const lv_draw_image_sup_t * sup, lv_color_format_t cf, void * dest_buf);
 
-#if LV_USE_VECTOR_GRAPHIC
+#if LV_USE_VECTOR_GRAPHIC && (LV_USE_THORVG_EXTERNAL || LV_USE_THORVG_INTERNAL)
```
XuNeo, line 165: Should we use LV_USE_THORVG instead.

FASTSHIFT: I think for now, there is no need to introduce new macros.

kisvegabor: I agree, `(LV_USE_THORVG_EXTERNAL || LV_USE_THORVG_INTERNAL)` is used only on a few places

XuNeo: Let's go in this way. 
LV_USE_LZ4_INTERNAL/EXTERNAL has similar problem, but I have defined LV_USE_LZ4 too.

kisvegabor: I would remove `LV_USE_LZ4` too to make the user set only 1 config option.

PR #5034, opened 2023-12-18
fix(barcode): use draw buffer to set canvas to meet align requirement 

XuNeo:

Continue of #5012 #5022 
Only with draw buffer interface, the start address and stride alignment can meet requirement.

### Checkpoints
 
### DIFF
--- a/src/libs/barcode/lv_barcode.c
+++ b/src/libs/barcode/lv_barcode.c
@@ -195,20 +195,10 @@ static void lv_barcode_destructor(const lv_obj_class_t * class_p, lv_obj_t * obj
 {
     LV_UNUSED(class_p);
 
-    lv_image_dsc_t * img = lv_canvas_get_image(obj);
-    lv_cache_lock();
-    lv_cache_invalidate_by_src(img, LV_CACHE_SRC_TYPE_POINTER);
-    lv_cache_unlock();
+    lv_draw_buf_t * draw_buf = lv_canvas_get_draw_buf(obj);
+    if(draw_buf == NULL) return;
 
-    if(!img->data) {
-        LV_LOG_INFO("canvas buffer is NULL");
-        return;
-    }
-
-    LV_LOG_INFO("free canvas buffer: %p", lv_canvas_get_buf(obj));
-
-    lv_draw_buf_free((void *)lv_canvas_get_buf(obj));
-    img->data = NULL;
+    lv_draw_buf_destroy(draw_buf);
 }
 
 static bool lv_barcode_change_buf_size(lv_obj_t * obj, int32_t w, int32_t h)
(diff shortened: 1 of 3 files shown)


src/widgets/canvas/lv_canvas.h
```diff
@@ -63,7 +63,9 @@ lv_obj_t * lv_canvas_create(lv_obj_t * parent);
  * @param h height of the canvas
  * @param cf color format. `LV_COLOR_FORMAT...`
  */
-void lv_canvas_set_buffer(lv_obj_t * canvas, void * buf, int32_t w, int32_t h, lv_color_format_t cf);
+void lv_canvas_set_buffer(lv_obj_t * obj, void * buf, int32_t w, int32_t h, lv_color_format_t cf);
+
+void lv_canvas_set_draw_buf(lv_obj_t * obj, lv_draw_buf_t * draw_buf);
```
kisvegabor, line 76: Please add comments here and for `lv_canvas_get_draw_buf`.

kisvegabor (commented):
I like the approach, especially because the `lv_canvas_set_buffer` is available too.

For consistency `lv_canvas_set_draw_buffer` might be a better name, but it's doesn't matches `lv_draw_buf_t`.

What about a simple search replace of `draw_buf` to `draw_buffer`?

kisvegabor:
`lv_image_create()` is already used by the `lv_image` widget, however `lv_image_buf_*` could work. 

It really opens to door for some simplification and removal of duplication. 

Let's try it, if it really looks good we are still on time to add this change. 

I merge it now, to no block the development of with nitpicking on naming.

kisvegabor:
Just to confirm: I'll merge it by keeping the 3 commits. Ok?

XuNeo:
@kisvegabor @FASTSHIFT Please help to review the merge the code.
Please do not squash the two commits if that's OK.

src/libs/barcode/lv_barcode.c
```diff
@@ -195,43 +195,28 @@ static void lv_barcode_destructor(const lv_obj_class_t * class_p, lv_obj_t * obj
...
-        return;
-    }
-
-    LV_LOG_INFO("free canvas buffer: %p", lv_canvas_get_buf(obj));
+    lv_draw_buf_t * draw_buf = lv_canvas_get_draw_buf(obj);
+    if(draw_buf == NULL) return;
 
-    lv_draw_buf_free((void *)lv_canvas_get_buf(obj));
-    img->data = NULL;
+    lv_draw_buf_destroy(draw_buf);
```
kisvegabor, line 201: Shouldn't it be handled in `lv_canvas`?

XuNeo: The draw buffer is allocated in `bardcode`, thus should be destroyed in `barcode` too. Canvas could also use static draw buffer, thus cannot destroy externally provided buffer.

PR #5045, opened 2023-12-19
feat(draw-sw): add Helium acceleration to draw-sw

GorgonMeducer:

Add Helium acceleration to draw-sw. 

Ready to review and merge.

### Checkpoints
 

src/draw/sw/blend/neon/lv_blend_neon.h
```diff
@@ -226,10 +228,11 @@ static inline void _lv_color_blend_to_rgb565_with_mask_neon(_lv_draw_sw_blend_fi
         .mask_stride = dsc->mask_stride
     };
     lv_color_blend_to_rgb565_with_mask_neon(&asm_dsc);
+    return true;
 }
 
 extern void lv_color_blend_to_rgb565_mix_mask_opa_neon(asm_dsc_t * dsc);
-static inline void _lv_color_blend_to_rgb565_mix_mask_opa_neon(_lv_draw_sw_blend_fill_dsc_t * dsc)
+static inline bool _lv_color_blend_to_rgb565_mix_mask_opa_neon(_lv_draw_sw_blend_fill_dsc_t * dsc)
```
kisvegabor:
```suggestion
static inline lv_result_t _lv_color_blend_to_rgb565_mix_mask_opa_neon(_lv_draw_sw_blend_fill_dsc_t * dsc)
```

src/draw/sw/blend/lv_draw_sw_blend_to_argb8888.c
```diff
@@ -96,90 +98,97 @@ LV_ATTRIBUTE_FAST_MEM void lv_draw_sw_blend_color_to_argb8888(_lv_draw_sw_blend_
...
-
-                dest_buf[x + 12] = color32;
-                dest_buf[x + 13] = color32;
-                dest_buf[x + 14] = color32;
-                dest_buf[x + 15] = color32;
-            }
-            for(; x < w; x ++) {
-                dest_buf[x] = color32;
-            }
+        if(!LV_DRAW_SW_COLOR_BLEND_TO_ARGB8888(dsc))
```
kisvegabor: It could be even cleaner if in case of `LV_DRAW_SW_ASM_NONE` is defined these
```c
#defeine LV_DRAW_SW_COLOR_BLEND_TO_ARGB8888(dsc) false;
```

This way no extra `#ifdef` guard is required here.

kisvegabor (commented):
Thank you! 

You make a PR "draft" to indicate it's not ready to merge.

src/draw/sw/blend/lv_draw_sw_blend_to_argb8888.c
GorgonMeducer: Good idea. I will make these change later.

GorgonMeducer: Done

src/draw/sw/blend/neon/lv_blend_neon.h
GorgonMeducer: Done

env_support/cmsis-pack/lv_conf_cmsis.h
```diff
@@ -61,6 +61,16 @@
  *(Not so important, you can adjust it to modify default sizes and spaces)*/
 #define LV_DPI_DEF 130     /*[px/inch]*/
 
+/*customize tick-get */
+#if defined(__PERF_COUNTER__) && __PERF_COUNTER__
+    #define LV_GLOBAL_INIT(__GLOBAL_PTR)                                    \
```
kisvegabor, line 66: It makes sense to me.
@bjsylvia please take a look as well.

GorgonMeducer: Yes, with this method, the support for ESP timer and Arduino timer can be added back.

GorgonMeducer:
@kisvegabor I found no test in the benchmark to cover the blending of rgb888 to rgb565.

kisvegabor:
The benchmark doesn't cover all cases. For this you can use `lv_demo_render(LV_DEMO_RENDER_SCENE_IMAGE_NORMAL, opa);`
image

GorgonMeducer:
> The benchmark doesn't cover all cases. For this you can use `lv_demo_render(LV_DEMO_RENDER_SCENE_IMAGE_NORMAL, opa);`
> 
> image
> 
> 

The LCD on my prototype testing platform is only 320×240 and is too small to run this demo. 

Anyway, I will figure out a way to validate.

kisvegabor:
I recommend commenting out parts of `lv_demo_render` to make it fit to 320x240.

GorgonMeducer:
> I recommend commenting out parts of `lv_demo_render` to make it fit to 320x240.

Don't worry, I will use Renesas EK-RA8D1 board for the test (Cortex-M85, 480×854 rgb888)

GorgonMeducer:
@kisvegabor I have added the initial Helium support to blend. The coverage is the same or more than the level of LVGL8.

Please review it and merge it when you think it is OK.

I will create a separate PR to accelerate the transform in `lv_draw_sw_img.c` that is not covered by the neon version.

GorgonMeducer:
It is strange about the `Verify code formatting` part. If you check the error details, the error (extra lines) doesn't exist.

image

src/draw/sw/helium/lv_blend_helium.h
```diff
@@ -0,0 +1,1347 @@
+/**
+ * @file lv_blend_to_helium.h
```
XuNeo: It looks this file should be removed. The actual file used is in draw/sw/blend/helium

src/draw/sw/blend/helium/lv_blend_helium.h
```diff
@@ -0,0 +1,1346 @@
...
+extern "C" {
+#endif
+
+/*********************
+ *      INCLUDES
+ *********************/
+
+//#include "../../../../lv_conf_internal.h"
+#include "lvgl.h"
+
```
XuNeo: Let's use relative path instead. Include the whole "lvgl.h" seems unnecessary.

src/draw/sw/helium/lv_blend_helium.h
GorgonMeducer: Solved

src/draw/sw/blend/helium/lv_blend_helium.h
GorgonMeducer: Done

kisvegabor:
@GorgonMeducer Could you test it with a meaningful amount input-output color format combinations?  If so have you compared its performance  with pure SW rendering?

GorgonMeducer:
I have test this with benchmark, Widegts and render test. 

I saw similar performance update in LVGL8.

kisvegabor (approved):

kisvegabor:
Great! What was the average performance gain?  Could you share the end screen of the benchmark with and without Helium? (No logging is added now :slightly_frowning_face: ) Or just the first row?

Please confirm if this PR is ready to merge.

GorgonMeducer:
This PR is ready to merge.

GorgonMeducer:
I disable the screen update before running the benchmark and resuming the screen update. I get the render-only performance:

Without Helium:
image
image


With Helium:
image
image

GorgonMeducer:
NOTE: 320*240 RGB565    **1/4 FB**    LV_DEF_REFR_PERIOD=2ms

Without Helium:
image


With Helium:
image

src/draw/sw/blend/helium/lv_blend_helium.h
```diff
@@ -0,0 +1,1345 @@
...
+static inline lv_result_t _lv_rgb888_blend_normal_to_rgb565_helium(_lv_draw_sw_blend_image_dsc_t * dsc,
+                                                                   uint32_t src_px_size)
+{
+    if(src_px_size == 3) {
+        return LV_RESULT_INVALID;
+    }
+
+    arm_2d_size_t draw_size = {dsc->dest_w, dsc->dest_h};
+    int16_t des_stride = dsc->dest_stride / sizeof(uint16_t);
+    int16_t src_stride = dsc->src_stride / sizeof(uint32_t);
```
XuNeo, line 314: Is it working that use sizeof(uint16_t) for RGB888?
LVGL use byte as stride unit. Because aligning up stride to like 64byte will have 64 % sizeof(rgb888) != 0. So`des_stride` calculation will be wrong.

src/draw/sw/blend/helium/lv_blend_helium.h
```diff
@@ -0,0 +1,1345 @@
...
+        return LV_RESULT_INVALID;
+    }
+
+    lv_memzero(tmp_msk, des_stride * dsc->dest_h);
+
+    /* get mask */
+    __arm_2d_impl_gray8_colour_filling_channel_mask_opacity(tmp_msk,
+                                                            des_stride,
+                                                            (uint32_t *)((uintptr_t)(dsc->src_buf) + 3),
+                                                            src_stride,
```
XuNeo, line 957: Could you explain why `(uintptr_t)(dsc->src_buf) + 3` is needed?

src/draw/sw/blend/helium/lv_blend_helium.h
```diff
@@ -0,0 +1,1345 @@
...
+    asm_dsc_t asm_dsc = {
+        .dst_buf = dsc->dest_buf,
+        .dst_w = dsc->dest_w,
+        .dst_h = dsc->dest_h,
+        .dst_stride = dsc->dest_stride,
+        .src_buf = &dsc->color
+    };
+
+    lv_color_blend_to_argb8888_helium(&asm_dsc);
+#endif
```
XuNeo, line 1046: Will it be supported in future or did you encounter problems thus comment the code out?

kisvegabor:
>  25% performance gain, similar to LVGL8

That's great! I wonder what is the bottleneck with Helium in case of e.g. the  "Empty screen" test.

> I cannot figure out the relationship between the indicated FPS and the ms. e.g. 258 FPS (19ms+0ms)

See [here](https://github.com/lvgl/lvgl/issues/4624#issuecomment-1863008072). I think `LV_DEF_REFR_PERIOD` should be set to a reasonable value (e.g. 16ms -> 60FPS) to see if the drops below it. However something is strange. However something is strange: on the end screen there is 7ms flush time:
image
But in the table 0 ms flush time is reported. Maybe it's a bug in the benchmark. I'll check it out.

src/draw/sw/blend/helium/lv_blend_helium.h
GorgonMeducer: It is reserved for the future

src/draw/sw/blend/helium/lv_blend_helium.h
GorgonMeducer: It is a feature in arm-2d that a specified channel in 32bit format can be used as mask. For ccca8888 (i.e.  argb8888) the alpha channel is the 4th Byte, hence we use "+3"

src/draw/sw/blend/helium/lv_blend_helium.h
GorgonMeducer: Sorry, what's the problem? I know LVGL using byte for stride, and arm-2d uses pixel for stride, so here we need a proper conversion. 

Here source pixel size is 4, so to convert LVGL stride into pixel stride, we have to divide 4, i.e. sizeof(uint32_t). 

I didn't see any problem here. This acceleration has been visually validated on LCD.

GorgonMeducer:
Can we merge?

src/draw/sw/blend/helium/lv_blend_helium.h
XuNeo: Stride can be any number of bytes, for example 100x100 RGB888 with stride aligned up to 64, it cannot be converted to pixel stride, because it's not dividable by 3.

GorgonMeducer:
> on the end screen there is 7ms flush time:
But in the table 0 ms flush time is reported. Maybe it's a bug in the benchmark. I'll check it out.

This isn't a bug. As I disabled the LCD flush during benchmark running period and resume the LCD flush when summary is ready. 

The true problem is, Why LCD flushing time is countered as rendering time but not the flushing time. 

Please check all my screen shots, you will notice what I said: 

a. If I don't disable LCD flush, the Empty screen tests report 50ms 
b. If I disable the LCD flush, the Empty screen tests report 6～7ms (given the fact that the raw LCD flush time is 43ms for the whole screen)

src/draw/sw/blend/helium/lv_blend_helium.h
GorgonMeducer: as you can see, I only accelerate the condition where pixel size is 4. Even you align up to 64, there is no problem at all. 
Did you see I filtered the (src_px_size == 3) condition? 

image

GorgonMeducer: @kisvegabor Any suggestion about this?

XuNeo: Get it. It won't accelerate RGB888 because SIMD processes in 4 bytes.

GorgonMeducer:
Thank you very much. Happy new year.

kisvegabor:
Sorry, for the late reply. Just got to my computer at the Christmas holiday. :slightly_smiling_face: 

image

Let's analyse what happens here:
- The only thing which was updated on the screen was the FPS indicator, in every 300ms or so.
- The total refresh time is 16+7 = 23ms -> ~40FPS
- So in a 300 ms period there were 23 ms for 1 real refresh and `(300-23) / LV_DEF_REFR_PERIOD = (300-23) / 2 =  138` "empty" refreshes. So 1+138 refresh in 300 ms, which is 463 for 1000 ms. And you got 460 FPS which is quite the same.

In summary, LVGL really attempted to update the display 460 times in a second, however most of the times there was nothing to do.  That's why I [suggested](https://github.com/lvgl/lvgl/pull/5045#issuecomment-1869285217) to set the `LV_DEF_REFR_PERIOD` to reasonable  value (16 or 33) else the small value will modify the FPS in an unfair way.

GorgonMeducer:
@kisvegabor 
> So in a 300 ms period there were 23 ms for 1 real refresh and (300-23) / LV_DEF_REFR_PERIOD = (300-23) / 2 =  138 "empty" refreshes. So 1+138 refresh in 300 ms, which is 463 for 1000 ms. And you got 460 FPS which is quite the same.

OK. I understand now. But is this algorithm counterintuitive? Why not just use the 1 real refresh time to calculate the FPS, i.e. 1000 / (16+7)?

kisvegabor:
> Why not just use the 1 real refresh time to calculate the FPS, i.e. 1000 / (16+7)?

Let say the rendering + flushing time were 1 ms. If we used this as FPS (1000 / 1) we displayes 1000 FPS which doesn't show the actual Frames per second. If there ere not refreshes, or  the refresh time was 0 ms, we needed to show 1000 / 0 = infinite FPS.  IMO they show different information:
- average times: not directly related to the FPS, just show how many ms LVGL spent with these activities 
- FPS: how many times LVGL tried to refreshed the display in a second 

> 
@kisvegabor

So in a 300 ms period there were 23 ms for 1 real refresh and (300-23) / LV_DEF_REFR_PERIOD = (300-23) / 2 = 138 "empty" refreshes. So 1+138 refresh in 300 ms, which is 463 for 1000 ms. And you got 460 FPS which is quite the same.

OK. I understand now. But is this algorithm counterintuitive? Why not just use the 1 real refresh time to calculate the FPS, i.e. 1000 / (16+7)?

> For the next question, why does Rendering-time contain the LCD-flushing time? As you can see, when I disabled(bypassed) the LCD update at the driver level when running the benchmark, the empty screen took much less time than the case in which I didn't bypass the LCD flushing.

What are the exact times for disabled an not disabled in case of the empty screen?  Render time shouldn't change, only the flush time.

GorgonMeducer:
> Render time shouldn't change, only the flush time.

The phenomenon is the render time is different:

Enable flush:
image

Disable flush
image

> What are the exact times for disabled and not disabled flush in case of the empty screen? 

```c
void on_benchmark_complete(void)
{
    disp_enable_update();
}

void main(void)
{
    ...
    disp_disable_update();
    lv_demo_benchmark();
...
}
```

I have made a small modification to `lv_demo_benchmark.c`

```c
__attribute__((weak))
void on_benchmark_complete(void)
{
}

static void next_scene_timer_cb(lv_timer_t * timer)
{
    LV_UNUSED(timer);

    scene_act++;

    load_scene(scene_act);
    if(scenes[scene_act].scene_time == 0) {
        lv_timer_delete(timer);
        summary_create();
        on_benchmark_complete();
    }
    else {
        lv_timer_set_period(timer, scenes[scene_act].scene_time);
    }
}
```

As you can see:
a. I disabled the update before calling `lv_demo_benchmark`
b. I enabled the update after finishing all tests.

Did you see the problem?

kisvegabor:
Ahhh, it was really a bug in LVGL. I fixed it in #5150. Please review and test it.

Issue #4835, opened 2023-11-20
Plans for caching in v9


niklasf:
### Introduce the problem

It looks like the new built-in cache (`src/misc/lv_cache_builtin.h`) is meant to be used with a relatively low number of elements. It iterates over all entries to find a match.

The older `src/misc/lv_lru.h` hashes keys to place elements in buckets.

Suppose there is a use case for caching many small bits of information. Which cache should be used?

### Examples and cases

* The bitmap cache for Tiny TTF may hold lots of entries.
* Tiny TTF may benefit from also caching glyph metrics (keyed by *font + font size + unicode letter + next unicode letter*).

### Suggested solution

What's the plan for caching in v9? Is `lv_lru` legacy and everything should use `lv_cache`, and should the built-in cache be changed to better deal with large numbers of entries? Or is `lv_cache` meant for small numbers of potentially big entries (like draw buffers), and `lv_lru` remains to be used for other cases?

kisvegabor:
We have discussed it with @XuNeo and @FASTSHIFT that we should have multiple cache instances each having its own internal behavior, aging, storage logic, search implementation, etc. 

We can have a cache instance for all images and other for letters, and a third e.g. for SDL textures.

W-Mai:
I thought about it and conceived a new cache framework.

## API that may be used by the newly designed cache

```cpp
lv_cache_t * lv_cache_create(lv_cache_class_t * cache_class,
                                size_t node_size, size_t max_size,
                                lv_cache_compare_cb_t compare_cb,
                                lv_cache_alloc_cb_t alloc_cb,
                                lv_cache_free_cb_t free_cb);

void * lv_cache_get(lv_cache_t * cache, const void * key, void * user_data);
void   lv_cache_reset(lv_cache_t * cache, const void * key, void * user_data);
void   lv_cache_clear(lv_cache_t * cache, void * user_data);
void   lv_cache_set_max_size(lv_cache_t * cache, size_t max_size, void * user_data);
size_t lv_cache_get_max_size(lv_cache_t * cache, void * user_data);
size_t lv_cache_get_size(lv_cache_t * cache, void * user_data);
size_t lv_cache_get_free_size(lv_cache_t * cache, void * user_data);

void   lv_cache_set_compare_cb(lv_cache_t * cache, lv_cache_compare_cb_t compare_cb, void * user_data);
[...]

PR #5049, opened 2023-12-19
feat(cache): new cache framework

W-Mai:

New cache framework from https://github.com/lvgl/lvgl/issues/4835

### Checkpoints
 

src/draw/lv_image_decoder.c
```diff
@@ -46,6 +51,12 @@ static uint32_t img_width_to_stride(lv_image_header_t * header);
 void _lv_image_decoder_init(void)
 {
     _lv_ll_init(img_decoder_ll_p, sizeof(lv_image_decoder_t));
+    img_cache_p = lv_cache_create(&lv_cache_class_lru_rb,
+    sizeof(lv_image_decoder_cache_data_t), 128, (lv_cache_ops_t) {
+        .compare_cb = (lv_cache_compare_cb_t)image_decoder_cache_compare_cb,
+        .create_cb = NULL,
```
XuNeo, line 62: Do we need create_cb for every entry? 
It should be transparent for users and use unified `lv_cache_entry_t` in `lv_cache.c`. 
`lv_image_cache.c` will use in this way
```
int lv_image_cache_add(src, img_decoded) {
    lv_cache_add(image_cache_instance, src, key_len, img_decoded, data_len, weigth);
}
```

src/libs/bin_decoder/lv_bin_decoder.c
```diff
@@ -150,171 +153,25 @@ lv_result_t lv_bin_decoder_info(lv_image_decoder_t * decoder, const void * src,
...
+    if(create_res == false) return LV_RESULT_INVALID;
     if(dsc->decoded == NULL) return LV_RESULT_OK; /*Need to read via get_area_cb*/
 
     /*Add it to cache*/
-    t = lv_tick_elaps(t);
-    lv_cache_lock();
-    lv_cache_entry_t * cache = lv_cache_add(NULL, 0, decoder->cache_data_type, dsc->header.w * dsc->header.h * 4);
-    if(cache == NULL) {
-        lv_cache_unlock();
+    lv_image_decoder_cache_data_t search_key;
```
XuNeo: For image cache, we don't have to store the `src_type` in search_key, since we can detect the type.

src/libs/bin_decoder/lv_bin_decoder.c
```diff
@@ -150,171 +153,25 @@ lv_result_t lv_bin_decoder_info(lv_image_decoder_t * decoder, const void * src,
...
-                free_decoder_data(dsc);
-                return LV_RESULT_INVALID;
-#endif
-            }
-
-            res = LV_RESULT_OK;
-        }
-    }
-
-    if(res != LV_RESULT_OK) {
```
XuNeo, line 285: Instead of modifying all existing code logic, the cache API should simply be called after decoding and add data to cache. Using the callback mechanism changes how the user expects code flows.
```c
    if(res != LV_RESULT_OK) {
    }
#if LV_USE_CACHE_OR_SIMILAR
    else {
        lv_image_cache_add(src, img_decoded, weight);
    }
#endif
```

src/libs/bin_decoder/lv_bin_decoder.c
W-Mai: If you check the src type again, it will make the process complicated and slow.

src/draw/lv_image_decoder.c
W-Mai: No, create callback is for cache instance, not for entry.

src/libs/bin_decoder/lv_bin_decoder.c
W-Mai: That's how it has been changed now.

src/libs/bin_decoder/lv_bin_decoder.c
XuNeo: It won't cost much since it only compare if the first character is printable or not.

Use `src` only will make user application much convenient to invalidate cache:
```c
/* From user's application code */
/*Image changed, invalidate cache now*/
lv_image_cache_invalide(src);
```

W-Mai: ok fine

src/misc/cache/lv_cache.c
```diff
@@ -0,0 +1,209 @@
...
+}
+void lv_cache_release(lv_cache_t* cache, lv_cache_entry_t* entry, void* user_data)
+{
+    LV_ASSERT_NULL(entry);
+
+    lv_mutex_lock(&cache->lock);
+    lv_cache_entry_release_data(entry, user_data);
+
+    if(lv_cache_entry_get_ref(entry) == 0 && lv_cache_entry_is_invalid(entry)) {
+        lv_cache_drop(cache, lv_cache_entry_get_data(entry), user_data);
```
terry0012: lv_cache_drop will lock again...

W-Mai: done

src/draw/lv_image_decoder.c
```diff
@@ -137,6 +148,11 @@ lv_result_t lv_image_decoder_open(lv_image_decoder_dsc_t * dsc, const void * src
         if(dsc->header.stride == 0) dsc->header.stride = img_width_to_stride(&dsc->header);
 
         dsc->decoder = decoder;
+        dsc->cache = img_cache_p;
```
XuNeo, line 159: It's a global image cache, thus no need to store cache instance in every decoded image.

src/draw/lv_image_decoder.c
```diff
@@ -233,6 +249,34 @@ void lv_image_decoder_set_close_cb(lv_image_decoder_t * decoder, lv_image_decode
     decoder->close_cb = close_cb;
 }
 
+void lv_image_decoder_set_cache_free_cb(lv_image_decoder_t * decoder, lv_cache_free_cb_t cache_free_cb)
+{
+    decoder->cache_free_cb = cache_free_cb;
+}
+
+lv_cache_entry_t * lv_image_decoder_add_to_cache(lv_image_decoder_t * decoder,
+                                                 lv_image_decoder_cache_data_t * search_key,
```
XuNeo:
```suggestion
                                                 void * src,
```

src/draw/lv_image_decoder.c
```diff
@@ -233,6 +249,34 @@ void lv_image_decoder_set_close_cb(lv_image_decoder_t * decoder, lv_image_decode
...
+                                                 const lv_draw_buf_t * decoded, void * user_data)
+{
+    lv_cache_entry_t * cache_entry = lv_cache_add(img_cache_p, search_key, NULL);
+    if(cache_entry == NULL) {
+        return NULL;
+    }
+
+    lv_image_decoder_cache_data_t * cached_data;
```
XuNeo: It's an image cache, let's call it `lv_image_cache_data_t`

src/draw/lv_image_decoder.c
```diff
@@ -243,3 +287,59 @@ static uint32_t img_width_to_stride(lv_image_header_t * header)
...
+        }
+        return 0;
+    }
+    else if(lhs->src_type == LV_IMAGE_SRC_FILE && rhs->src_type == LV_IMAGE_SRC_VARIABLE) {
+        return 1;
+    }
+    else if(lhs->src_type == LV_IMAGE_SRC_VARIABLE && rhs->src_type == LV_IMAGE_SRC_FILE) {
+        return -1;
+    }
+    return 0;
```
XuNeo:
```suggestion
    return lhs->src_type > rhs->src_type ? 1 : -1;
```

src/draw/lv_image_decoder.c
```diff
@@ -243,3 +287,59 @@ static uint32_t img_width_to_stride(lv_image_header_t * header)
...
+}
+
+static void image_decoder_cache_free_cb(lv_image_decoder_cache_data_t * entry, void * user_data)
+{
+    LV_UNUSED(user_data); /*Unused*/
+
+    const lv_image_decoder_t * decoder = entry->decoder;
+    if(decoder && decoder->cache_free_cb) {
+        decoder->cache_free_cb(entry, user_data);
```
XuNeo, line 370: Should store the cache_free_cb in cache instance, and use global image cache here to free.

src/draw/lv_image_decoder.h
```diff
@@ -111,10 +112,20 @@ typedef struct _lv_image_decoder_t {
...
+    lv_cache_free_cb_t cache_free_cb;
     void * user_data;
 } lv_image_decoder_t;
 
+typedef struct _lv_image_decoder_cache_data_t {
+    const void * src;
+    lv_image_src_t src_type;
+
+    const lv_draw_buf_t * decoded;
+    const lv_image_decoder_t * decoder;
```
XuNeo, line 127: Could it be destroyed after data added to cache?

src/draw/lv_image_decoder.h
```diff
@@ -111,10 +112,20 @@ typedef struct _lv_image_decoder_t {
...
     lv_image_decoder_close_f_t close_cb;
-    uint32_t cache_data_type;
+
+    lv_cache_free_cb_t cache_free_cb;
     void * user_data;
 } lv_image_decoder_t;
 
+typedef struct _lv_image_decoder_cache_data_t {
+    const void * src;
+    lv_image_src_t src_type;
```
XuNeo, line 124: Remove.

src/draw/lv_image_decoder.h
```diff
@@ -149,8 +160,10 @@ typedef struct _lv_image_decoder_dsc_t {
      * Can be set in `open` function or set NULL.*/
     const char * error_msg;
 
+    lv_cache_t * cache;
```
XuNeo, line 165: Let's remove it, we can get global image cache instance.

src/draw/sdl/lv_draw_sdl.c
```diff
@@ -58,13 +60,57 @@ static SDL_Texture * layer_get_texture(lv_layer_t * layer);
 /**********************
  *   GLOBAL FUNCTIONS
  **********************/
+static bool sdl_texture_cache_create_cb(cache_data_t * cached_data, void * user_data)
```
XuNeo, line 62: It looks like an entry's callback rather than instance's callback.

src/draw/lv_image_decoder.c
W-Mai: In this way, I don't need to include lv_global and define the image_cache macro again in each decoder.

src/draw/lv_image_decoder.c
W-Mai: I think the original way is more clear

src/draw/sdl/lv_draw_sdl.c
W-Mai: Its function is to provide instance with a callback to create entry.

src/draw/lv_image_decoder.h
W-Mai: It's a global instance

src/draw/lv_image_decoder.h
W-Mai: need

src/draw/lv_image_decoder.c
W-Mai: great idea!

src/misc/cache/lv_cache.h
```diff
@@ -0,0 +1,68 @@
...
+#define LV_CACHE1_H
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+/*********************
+ *      INCLUDES
+ *********************/
+#include "_lv_cache_internal.h"
```
kisvegabor:
```suggestion
#include "lv_cache_private.h"
```
"private" is already used by `lv_display/indev_private.h`.

src/misc/cache/lv_cache.h
```diff
@@ -0,0 +1,68 @@
...
+ *********************/
+
+/**********************
+ *      TYPEDEFS
+ **********************/
+
+/**********************
+ * GLOBAL PROTOTYPES
+ **********************/
+lv_cache_t * lv_cache_create(const lv_cache_class_t * cache_class,
```
kisvegabor, line 35: Please add API comments to these functions.

src/misc/cache/lv_cache_entry.h
```diff
@@ -0,0 +1,56 @@
...
+ *********************/
+
+/**********************
+ *      TYPEDEFS
+ **********************/
+
+/**********************
+ * GLOBAL PROTOTYPES
+ **********************/
+uint32_t lv_cache_entry_get_size(const uint32_t node_size);
```
kisvegabor, line 32: Please add API comments here too.

src/misc/cache/lv_image_cache.c
```diff
@@ -0,0 +1,55 @@
...
+void lv_image_cache_drop(const void * src)
+{
+    if(src == NULL) {
+        lv_cache_drop_all(img_cache_p, NULL);
+        return;
+    }
+
+    lv_image_cache_data_t search_key = {
+        .src = src,
+        .src_type = lv_image_src_get_type(src),
```
kisvegabor, line 49: Do we need the type here? We can get it at any time from `src`.

src/draw/sdl/lv_draw_sdl.c
```diff
@@ -11,6 +11,7 @@
 #include LV_SDL_INCLUDE_PATH
 #include <SDL2/SDL_image.h>
 
+#include <src/misc/cache/_lv_cache_lru_rb.h>
```
kisvegabor: Please use relative include path.

kisvegabor (commented):
In general it looks good! I think the long discussions payed off :slightly_smiling_face: 

Great work!

W-Mai:
@kisvegabor ok, thanks for your comment! I'll resolved those comment a moment later.

src/misc/cache/lv_image_cache.c
W-Mai: But it will consume a certain amount of time, although it is very short. I want to use 4 bytes of space in exchange for a short time.

kisvegabor: Okay, probably it's a good trade off.

src/misc/cache/lv_cache.h
W-Mai: I will upload a new pr to write all the related api docs and code comments after this pr is everything ok 😉

W-Mai:
The CI has passed now

lv_conf_template.h
```diff
@@ -273,7 +273,7 @@
  *Used by image decoders such as `lv_lodepng` to keep the decoded image in the memory.
  *Data larger than the size of the cache also can be allocated but
  *will be dropped immediately after usage.*/
-#define LV_CACHE_DEF_SIZE       0
+#define LV_CACHE_DEF_SIZE       32
```
kisvegabor: Please add what the unit is. (byte, kB, etc)

src/misc/cache/lv_cache.h
```diff
@@ -0,0 +1,67 @@
...
+                             size_t node_size, size_t max_size,
+                             lv_cache_ops_t ops);
+void lv_cache_destroy(lv_cache_t * cache, void * user_data);
+
+lv_cache_entry_t * lv_cache_acquire(lv_cache_t * cache, const void * key, void * user_data);
```
kisvegabor, line 40: Please don't forget to add the API comments here.

kisvegabor (commented):
I'll test it tomorrow. A few minor comments until that :slightly_smiling_face:

src/draw/lv_image_decoder.c
```diff
@@ -27,7 +28,11 @@
  **********************/
 
 static uint32_t img_width_to_stride(lv_image_header_t * header);
+static lv_cache_compare_res_t image_decoder_cache_compare_cb(const lv_image_cache_data_t * lhs,
+                                                             const lv_image_cache_data_t * rhs);
+static void image_decoder_cache_free_cb(lv_image_cache_data_t * entry, void * user_data);
 
+static lv_result_t try_cache(lv_image_decoder_dsc_t * dsc);
```
XuNeo:
```suggestion
#if LV_CACHE_DEF_SIZE > 0
    static lv_cache_compare_res_t image_decoder_cache_compare_cb(const lv_image_cache_data_t * lhs,
                                                                 const lv_image_cache_data_t * rhs);
    static void image_decoder_cache_free_cb(lv_image_cache_data_t * entry, void * user_data);

    static lv_result_t try_cache(lv_image_decoder_dsc_t * dsc);
#endif
```

src/draw/lv_image_decoder.c
XuNeo: But it's use to create entry.
See `static bool tiny_ttf_cache_create_cb(tiny_ttf_cache_data_t * node, void * user_data)`
The first parameter targets to entry data.

lv_conf_template.h
W-Mai: Currently it's based on count, it's working fine now, I'll add the cache method based on memory size at a later stage

src/misc/cache/lv_cache.h
W-Mai: I'll be adding documentation to all the APIs I've added, and I'll be adding comments to some of the key code as well!

src/draw/lv_image_decoder.c
```diff
@@ -233,6 +258,36 @@ void lv_image_decoder_set_close_cb(lv_image_decoder_t * decoder, lv_image_decode
...
 }
 
+void lv_image_decoder_set_cache_free_cb(lv_image_decoder_t * decoder, lv_cache_free_cb_t cache_free_cb)
+{
+    decoder->cache_free_cb = cache_free_cb;
+}
+
+#if LV_CACHE_DEF_SIZE > 0
+lv_cache_entry_t * lv_image_decoder_add_to_cache(lv_image_decoder_t * decoder,
+                                                 lv_image_cache_data_t * search_key,
```
XuNeo, line 267: Naming conflict. The type is called `cache_data`, but variable is called `key`.
I think it will confuse people.

src/draw/lv_image_decoder.c
```diff
@@ -233,6 +258,36 @@ void lv_image_decoder_set_close_cb(lv_image_decoder_t * decoder, lv_image_decode
     decoder->close_cb = close_cb;
 }
 
+void lv_image_decoder_set_cache_free_cb(lv_image_decoder_t * decoder, lv_cache_free_cb_t cache_free_cb)
+{
+    decoder->cache_free_cb = cache_free_cb;
+}
+
+#if LV_CACHE_DEF_SIZE > 0
+lv_cache_entry_t * lv_image_decoder_add_to_cache(lv_image_decoder_t * decoder,
```
XuNeo, line 266: Let's move it to `lv_image_cache.c` and include it from `lv_image_decoder.c`

src/draw/lv_image_decoder.c
```diff
@@ -233,6 +258,36 @@ void lv_image_decoder_set_close_cb(lv_image_decoder_t * decoder, lv_image_decode
...
+                                                 lv_image_cache_data_t * search_key,
+                                                 const lv_draw_buf_t * decoded, void * user_data)
+{
+    lv_cache_entry_t * cache_entry = lv_cache_add(img_cache_p, search_key, NULL);
```
XuNeo, line 270: `search_key` should be deep copied to cache entry because if image path is used as key, it's normally on stack.

Precisely, the issue will happen on lvgl/src/misc/lv_rb.c:88 where the pointer of `key` is stored to node. Which may become wild pointer.

The copy of `key` is also a shallow copy, thus `src` pointer still can become wild. 
https://github.com/lvgl/lvgl/blob/2deef84ff77ffcae89c6c3fb8467e050765ec0b9/src/misc/cache/_lv_cache_lru_rb.c#L92C5-L92C5

src/draw/lv_image_decoder.h
```diff
@@ -111,10 +112,20 @@ typedef struct _lv_image_decoder_t {
     lv_image_decoder_open_f_t open_cb;
     lv_image_decoder_get_area_cb_t get_area_cb;
     lv_image_decoder_close_f_t close_cb;
-    uint32_t cache_data_type;
+
+    lv_cache_free_cb_t cache_free_cb;
```
XuNeo, line 116: From the name `lv_cache_free_cb_t`, it should belong to `lv_image_cache.c` where image cache instance will use this callback. I understand bin decoder is not unified with other decoders, so the free callback is not unified yet. Let's optimize it so we can use same `free_cb` for all decoders.

src/draw/lv_image_decoder.h
```diff
@@ -268,6 +281,14 @@ void lv_image_decoder_set_get_area_cb(lv_image_decoder_t * decoder, lv_image_dec
  */
 void lv_image_decoder_set_close_cb(lv_image_decoder_t * decoder, lv_image_decoder_close_f_t close_cb);
 
+void lv_image_decoder_set_cache_free_cb(lv_image_decoder_t * decoder, lv_cache_free_cb_t cache_free_cb);
```
XuNeo, line 286: #if LV_CACHE_DEF_SIZE > 0
#endif

src/draw/sw/lv_draw_sw_img.c
```diff
@@ -70,6 +69,7 @@ void lv_draw_sw_layer(lv_draw_unit_t * draw_unit, const lv_draw_image_dsc_t * dr
     new_draw_dsc.src = &img_dsc;
 
     lv_draw_sw_image(draw_unit, &new_draw_dsc, coords);
+    lv_image_cache_drop(&img_dsc);
```
XuNeo, line 72: Add a todo comment here, it should be optimized that this image should not go to cache at all.

src/libs/bin_decoder/lv_bin_decoder.c
```diff
@@ -1008,42 +864,149 @@ static lv_result_t decompress_image(lv_image_decoder_dsc_t * dsc, const lv_image
...
 }
 
-static lv_result_t try_cache(lv_image_decoder_dsc_t * dsc)
+static bool bin_decoder_decode_data(lv_image_decoder_dsc_t * dsc)
 {
-    lv_cache_lock();
+    lv_image_decoder_t * decoder = dsc->decoder;
+
+    lv_fs_res_t res = LV_RESULT_INVALID;
+    // uint32_t t = lv_tick_get();
```
XuNeo: So current cache interface lacks of `weight`. It should be a parameter of `lv_cache_add` as discussed on meeting.

src/libs/bin_decoder/lv_bin_decoder.c
```diff
@@ -1008,42 +864,149 @@ static lv_result_t decompress_image(lv_image_decoder_dsc_t * dsc, const lv_image
...
 
-static void cache_invalidate_cb(lv_cache_entry_t * entry)
+static void bin_decoder_cache_free_cb(lv_image_cache_data_t * cached_data, void * user_data)
 {
+    LV_UNUSED(user_data); /*Unused*/
+
     lv_image_decoder_dsc_t fake = { 0 };
-    fake.user_data = entry->user_data;
+    fake.user_data = cached_data->user_data;
     free_decoder_data(&fake);
```
XuNeo, line 1025: Let separate the `free_decoder_data` with `free_image_cache_entry`. So we can unify it with other decoders.

src/misc/cache/lv_cache_entry.c
```diff
@@ -0,0 +1,167 @@
...
+}
+bool lv_cache_entry_is_invalid(lv_cache_entry_t * entry)
+{
+    LV_ASSERT_NULL(entry);
+    return entry->is_invalid;
+}
+void * lv_cache_entry_get_data(lv_cache_entry_t * entry)
+{
+    LV_ASSERT_NULL(entry);
+    return (uint8_t *)entry - entry->node_size;
```
XuNeo, line 95: This whole logic is used to simplify users' mallocing a entry and free it, which can be easily done before call custom `free_cache_entry_cb` etc. 

So I suggest malloc entry, and malloc user's data separately. Return entry->data directly from `lv_cache_entry_get_data `, and no need for `lv_cache_entry.c` any more.

XuNeo (commented):
The cache implementation is generally well-done and it's evident that a significant amount of time was invested in refining it. However, I believe there are areas where the design could be simplified:

- cache_class: Unlike widgets, users rarely need to extend the current cache methods. By removing this, we could simplify numerous callbacks.

- cache_entry: The entry could be a simple, transparent struct for the user, where a data pointer is provided for use. The current design employs a variable-length struct, which adds unnecessary complexity.

Let's proceeding with current implementation to avoid delays, and we can make these simplifications later.

src/draw/lv_image_decoder.c
W-Mai: I don't think so, because the current cache is designed so that the key and data are shared in the same structure.

src/draw/lv_image_decoder.c
W-Mai: It's not the right time yet, it's still dependent on decoder

src/draw/lv_image_decoder.c
W-Mai: All of the scenarios where we use the path key now involve making a new copy of the memory, and then freeing the path key's memory at release time, which is how the code in the current project operates. So don't worry about that, where pointers are concerned, everyone makes a good advance copy.

src/misc/cache/lv_cache_entry.c
W-Mai: I don't want the user to be able to manually modify the entry's data, he has to go through an API action

src/libs/bin_decoder/lv_bin_decoder.c
W-Mai: I have an idea about this, but we can't use this feature now. I'll add it back later.

src/draw/lv_image_decoder.c
XuNeo: So that means user need to make sure the `key` is not on stack and always make a deep copy before adding to cache. Instead of `everyone make an advance copy`, cache manager should do it.

src/misc/cache/lv_cache_entry.c
XuNeo: I meant adding `void * data` to `lv_cache_entry_t `. User still call `lv_cache_entry_get_data ` to retrieve data.

lv_conf_template.h
kisvegabor: I think it's an important step, so please add the size based operation before merging it.

src/draw/lv_image_decoder.c
kisvegabor: I think `lhs` and `rhs` should be renamed to something more descriptive and easier to understand.

src/draw/lv_image_decoder.c
kisvegabor: It works like [this](https://github.com/lvgl/lvgl/blob/master/src/draw/lv_draw_label.c#L87-L90) in `lv_draw_label_dsc_t` where the text can be local or const too.

src/draw/lv_image_decoder.c
W-Mai: Generally speaking, the two nodes of the comparison function are described as left operands and right operands, or left-hand side and right-hand side, so I continued the common habit and used the abbreviations `lhs` and `rhs` of left-hand side and right-hand side. Do you have any other better ideas?

kisvegabor:
Please add at least a few tests with to be sure that it's working well.

W-Mai:
> Please add at least a few tests with to be sure that it's working well.

OK! I will add it.

Before that, I want to confirm the requirements for some test cases.

1. Is it a unit test to test the cache function separately?
2. Do you need to test with graphics?
3. At present, the testcase related to graphics and decoder in ci can pass, and the demo and example of lvgl can also work fine.

So what are the test cases I should write related to?

src/draw/lv_image_decoder.c
kisvegabor: I would call it `data1` and `data2` :smile: 
But I leave it up to you, it's really not that important in case of a `static` function.

W-Mai: Ha, 🤪, then I'll use `lhs` and `rhs` first. I think this is more interesting.

kisvegabor:
> Is it a unit test to test the cache function separately?

Yes, ideally the cache should be testes as separated as possible. 

> Do you need to test with graphics?

Only a few ones.

> At present, the testcase related to graphics and decoder in ci can pass, and the demo and example of lvgl can also work fine.

I think these should be explicitly tested for the cache:
- what happens if the cache is full: 
   - will it really drop an old entry
   - when some entries are dropped can we still access to the others? 
   - will it drop multiple entries to make space for a new one? 
- what if the entry to cache is too large and can't fit into the defined size?
- are the memory leaks when the cache pool is small and we use it extensively (a lot entries are dropped and other are allocated)
- render test1: 
   1. create an image widget from a "image1.png" file
   2. remove "image1.png" and rename "image2.png" to "image1.png"
   3. screenshot test: still "image1.png" should be shown
   4. invalidate "image1.png"
   5. screenshot test: now "image2.png" should be shown
- render test2:
   - almost the same but with a canvas which not cached, so the changes are shown without any extra action 
- something for fonts?

These are just ideas that I would try, however I'm sure that you know better what the interesting and cases to check are.

src/draw/lv_image_decoder.c
kisvegabor: Okay :laughing:

lv_conf_template.h
W-Mai: fine, I'v added size based cache back now 😁

terry0012:
> > Please add at least a few tests with to be sure that it's working well.
> 
> OK! I will add it.
> 
> Before that, I want to confirm the requirements for some test cases.
> 
> 1. Is it a unit test to test the cache function separately?
> 2. Do you need to test with graphics?
> 3. At present, the testcase related to graphics and decoder in ci can pass, and the demo and example of lvgl can also work fine.
> 
> So what are the test cases I should write related to?

1. add a unit test for cache.
2. write a test case with image and text, and change the image and text in period?

W-Mai:
I added a simple unit test. We urgently need to use the new cache framework in our project. At present, there are no bugs or inconveniences.

As for the remaining cache test cases, I want to make up for them later (I won't run away, I will definitely add them 😁. DON'T BE AFRAID😘)

kisvegabor:
I don't afraid :slightly_smiling_face: 
Let's fix the conflict, merge it, and update it gradually later.

PR #5051, opened 2023-12-19
refactor(image): add magic to image header

### LINKED ISSUE #5041: Support YUV color format
### Introduce the problem

To support video and YUV animation, we need to expand the color format. However, 

YUV is divided into many different formats according to different storage methods such as Planar/Semi-Planar/Packed. 

There are currently 10 commonly used YUV formats according to [https://github.com/lemenkov/libyuv/blob/main/include/libyuv/video_common.h](url), but the max number of color format in lv_image_header_t is 31 as below. 

```c
  typedef struct {
    uint32_t cf : 5;            /*Color format: See `lv_color_format_t`*/
    uint32_t always_zero : 3;   /*It the upper bits of the first byte. Always zero to look like a
                                  non-printable character*/

    uint32_t reserved_1: 8;     /*Reserved by LVGL for later use*/
    uint32_t flags: 16;         /*Image flags, see `lv_image_flags_t`*/

    uint32_t w: 16;
    uint32_t h: 16;
    uint32_t stride: 16;        /*Number of bytes in a row*/
    uint32_t reserved_2: 16;    /*Reserved to be used later*/
   } lv_image_header_t;
```

We cannot enumerate all YUV formats.

### Suggested solution

1. **expand color format.**
  But `always_zero` is used to differentiate between image src types, we can't expand cf.
```c
lv_image_src_t lv_image_src_get_type(const void * src)
{
    lv_image_src_t img_src_type = LV_IMAGE_SRC_UNKNOWN;

    if(src == NULL) return img_src_type;
    const uint8_t * u8_p = src;

    /*The first byte shows the type of the image source*/
    if(u8_p[0] >= 0x20 && u8_p[0] <= 0x7F) {
        img_src_type = LV_IMAGE_SRC_FILE; /*If it's an ASCII character then it's file name*/
    }
    else if(u8_p[0] >= 0x80) {
        img_src_type = LV_IMAGE_SRC_SYMBOL; /*Symbols begins after 0x7F*/
    }
    else {
        img_src_type = LV_IMAGE_SRC_VARIABLE; /*`lv_image_dsc_t` is draw to the first byte < 0x20*/
    }

    if(LV_IMAGE_SRC_UNKNOWN == img_src_type) {
        LV_LOG_WARN("unknown image type");
    }

    return img_src_type;
}
```
2.  as discussed in https://github.com/lvgl/lvgl/issues/4869, add `LV_COLOR_FORMAT_YUV` in `_lv_color_format_t` and put the detail information of YUV in buf_sup.
```c
typedef struct {
    lv_image_header_t header;           /*A header describing the basics of the image*/
    uint32_t data_size;                 /*Size of the data in bytes*/
    const void * buf;
    const void * buf_sup;           /*Optional supplementary buffer*/
}
```

How do you think ?

XuNeo:

Fix #5041
So the cf field is not limited to 5bits, to add more color format options. A workaround is included, so the existing bin images can still work.

### Checkpoints
 
### DIFF
--- a/examples/libs/gif/lv_example_gif_1.py
+++ b/examples/libs/gif/lv_example_gif_1.py
@@ -11,7 +11,7 @@ fs_driver.fs_register(fs_drv, 'S')
 #
 image_bulb_gif = lv.image_dsc_t(
     {
-        "header": {"always_zero": 0, "w": 0, "h": 0,  "cf": lv.COLOR_FORMAT.RAW},
+        "header": {"w": 0, "h": 0,  "cf": lv.COLOR_FORMAT.RAW},
         "data_size": 0,
         "data": img_bulb_gif_map,
     }
--- a/examples/libs/lodepng/lv_example_lodepng_1.py
+++ b/examples/libs/lodepng/lv_example_lodepng_1.py
@@ -5,7 +5,7 @@ from img_wink_png import img_wink_png_map
 
 image_wink_png = lv.image_dsc_t(
     {
-        "header": {"always_zero": 0, "w": 50, "h": 50,  "cf": lv.COLOR_FORMAT.RAW_ALPHA},
+        "header": {"w": 50, "h": 50,  "cf": lv.COLOR_FORMAT.RAW_ALPHA},
         "data_size": 5158,
         "data": img_wink_png_map,
     }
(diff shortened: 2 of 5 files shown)


kisvegabor:
MicroPython merged, submodules are updated, CI is passing, merging this one.

PR #5057, opened 2023-12-19
feat(indev) Add crown support to pointer input device

kisvegabor:

It nicely integrates the the typical watch crown features with the pointer input device. To test it with SDL change `LV_SDL_MOUSEWHEEL_MODE` to `1`.

cc @XuNeo @bjsylvia 

### Checkpoints
 
### DIFF
--- a/src/core/lv_obj_style_gen.c
+++ b/src/core/lv_obj_style_gen.c
@@ -762,6 +762,14 @@ void lv_obj_set_style_bitmap_mask_src(lv_obj_t * obj, const lv_image_dsc_t * val
     };
     lv_obj_set_local_style_prop(obj, LV_STYLE_BITMAP_MASK_SRC, v, selector);
 }
+
+void lv_obj_set_style_rotary_sensitivity(lv_obj_t * obj, uint32_t value, lv_style_selector_t selector)
+{
+    lv_style_value_t v = {
+        .num = (int32_t)value
+    };
+    lv_obj_set_local_style_prop(obj, LV_STYLE_ROTARY_SENSITIVITY, v, selector);
+}
 #if LV_USE_FLEX
 
 void lv_obj_set_style_flex_flow(lv_obj_t * obj, lv_flex_flow_t value, lv_style_selector_t selector)
--- a/src/drivers/sdl/lv_sdl_mouse.c
+++ b/src/drivers/sdl/lv_sdl_mouse.c
@@ -35,6 +35,9 @@ typedef struct {
     int16_t last_x;
     int16_t last_y;
     bool left_button_down;
+#if LV_SDL_MOUSEWHEEL_MODE == 1
+    int32_t diff;
+#endif
 } lv_sdl_mouse_t;
 
 /**********************
@@ -76,6 +79,10 @@ static void sdl_mouse_read(lv_indev_t * indev, lv_indev_data_t * data)
     data->point.x = dsc->last_x;
     data->point.y = dsc->last_y;
     data->state = dsc->left_button_down ? LV_INDEV_STATE_PRESSED : LV_INDEV_STATE_RELEASED;
+#if LV_SDL_MOUSEWHEEL_MODE == 1
+    data->enc_diff = dsc->diff;
+    dsc->diff = 0;
+#endif
 }
 
 static void release_indev_cb(lv_event_t * e)
@@ -101,7 +108,11 @@ void _lv_sdl_mouse_handler(SDL_Event * event)
         case SDL_MOUSEMOTION:
             win_id = event->motion.windowID;
             break;
-
+#if LV_SDL_MOUSEWHEEL_MODE == 1
+        case SDL_MOUSEWHEEL:
+            win_id = event->wheel.windowID;
+            break;
+#endif
         case SDL_FINGERUP:
         case SDL_FINGERDOWN:
         case SDL_FINGERMOTION:
@@ -174,8 +185,19 @@ void _lv_sdl_mouse_handler(SDL_Event * event)
             indev_dev->last_x = (int16_t)((float)hor_res * event->tfinger.x / zoom);
             indev_dev->last_y = (int16_t)((float)ver_res * event->tfinger.y / zoom);
             break;
+        case SDL_MOUSEWHEEL:
+#if LV_SDL_MOUSEWHEEL_MODE == 1
+#ifdef __EMSCRIPTEN__
+            /*Escripten scales it wrong*/
+            if(event->wheel.y < 0) dsc->diff++;
+            if(event->wheel.y > 0) dsc->diff--;
+#else
+            indev_dev->diff = -event->wheel.y;
+#endif  /*LV_SDL_MOUSEWHEEL_MODE*/
+#endif /*LV_SDL_MOUSEWHEEL_MODE == 1*/
+            break;
     }
-    lv_indev_read(indev);
+    //    lv_indev_read(indev);
 }
 
 #endif /*LV_USE_SDL*/
--- a/src/drivers/sdl/lv_sdl_mouse.c
+++ b/src/drivers/sdl/lv_sdl_mouse.c
@@ -197,7 +197,7 @@ void _lv_sdl_mouse_handler(SDL_Event * event)
 #endif /*LV_SDL_MOUSEWHEEL_MODE == 1*/
             break;
     }
-    //    lv_indev_read(indev);
+    lv_indev_read(indev);
 }
 
 #endif /*LV_USE_SDL*/
--- a/src/drivers/sdl/lv_sdl_mouse.c
+++ b/src/drivers/sdl/lv_sdl_mouse.c
@@ -35,7 +35,7 @@ typedef struct {
     int16_t last_x;
     int16_t last_y;
     bool left_button_down;
-#if LV_SDL_MOUSEWHEEL_MODE == 1
+#if LV_SDL_MOUSEWHEEL_MODE == LV_SDL_MOUSEWHEEL_MODE_CROWN
     int32_t diff;
 #endif
 } lv_sdl_mouse_t;
@@ -79,7 +79,7 @@ static void sdl_mouse_read(lv_indev_t * indev, lv_indev_data_t * data)
     data->point.x = dsc->last_x;
     data->point.y = dsc->last_y;
     data->state = dsc->left_button_down ? LV_INDEV_STATE_PRESSED : LV_INDEV_STATE_RELEASED;
-#if LV_SDL_MOUSEWHEEL_MODE == 1
+#if LV_SDL_MOUSEWHEEL_MODE == LV_SDL_MOUSEWHEEL_MODE_CROWN
     data->enc_diff = dsc->diff;
     dsc->diff = 0;
 #endif
@@ -108,7 +108,7 @@ void _lv_sdl_mouse_handler(SDL_Event * event)
         case SDL_MOUSEMOTION:
             win_id = event->motion.windowID;
             break;
-#if LV_SDL_MOUSEWHEEL_MODE == 1
+#if LV_SDL_MOUSEWHEEL_MODE == LV_SDL_MOUSEWHEEL_MODE_CROWN
         case SDL_MOUSEWHEEL:
             win_id = event->wheel.windowID;
             break;
@@ -186,7 +186,7 @@ void _lv_sdl_mouse_handler(SDL_Event * event)
             indev_dev->last_y = (int16_t)((float)ver_res * event->tfinger.y / zoom);
             break;
         case SDL_MOUSEWHEEL:
-#if LV_SDL_MOUSEWHEEL_MODE == 1
+#if LV_SDL_MOUSEWHEEL_MODE == LV_SDL_MOUSEWHEEL_MODE_CROWN
 #ifdef __EMSCRIPTEN__
             /*Escripten scales it wrong*/
             if(event->wheel.y < 0) dsc->diff++;
@@ -194,7 +194,7 @@ void _lv_sdl_mouse_handler(SDL_Event * event)
 #else
             indev_dev->diff = -event->wheel.y;
 #endif  /*__EMSCRIPTEN__*/
-#endif /*LV_SDL_MOUSEWHEEL_MODE == 1*/
+#endif /*LV_SDL_MOUSEWHEEL_MODE == LV_SDL_MOUSEWHEEL_MODE_CROWN*/
             break;
     }
     lv_indev_read(indev);
--- a/src/drivers/sdl/lv_sdl_mouse.c
+++ b/src/drivers/sdl/lv_sdl_mouse.c
@@ -193,7 +193,7 @@ void _lv_sdl_mouse_handler(SDL_Event * event)
             if(event->wheel.y > 0) dsc->diff--;
 #else
             indev_dev->diff = -event->wheel.y;
-#endif  /*LV_SDL_MOUSEWHEEL_MODE*/
+#endif  /*__EMSCRIPTEN__*/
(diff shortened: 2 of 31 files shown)


src/indev/lv_indev.c
```diff
@@ -1275,6 +1271,37 @@ static void indev_proc_release(lv_indev_t * indev)
...
+
+    indev_obj_act = obj;
+
+    bool interactive = lv_obj_is_group_def(obj);
+
+    if(interactive) {
+        uint32_t k = indev->pointer.diff < 0 ? LV_KEY_LEFT : LV_KEY_RIGHT;
+        uint32_t cnt = LV_ABS(indev->pointer.diff);
+        while(cnt) {
+            send_event(LV_EVENT_KEY, &k);
```
XuNeo: Should we define a new event and send the diff directly. Only in this way, we can make sure every turn of wheel is counted in.

XuNeo (commented):
The code looks really clean. I'll try it later. Thank you.

src/indev/lv_indev.c
kisvegabor: In the read callback or [the event](https://github.com/lvgl/lvgl/pull/5057/files#diff-63956c41f2f8adef0b3fa8fb8d8621c777559f0aeb54f42388e68895d3c15fcdR195) you can set the diff to 3 for example. And the loop will send the key 3 times.

XuNeo: Yes. I mean no need to loop. Typically, a watch crown could generate 1024 diff per rotation.

W-Mai: It may be more appropriate to pack the diff together and send a total number.

kisvegabor: It's doable, but it requires to update all the widget receiving KEY events. I suggest merging it as it is and add support the diff cunt event parameter in a new PR later.

bjsylvia:
@kisvegabor I used this patch to run on a Mac environment, but there are crashes related to scrolling animation when using the scroll wheel. I will try it on Ubuntu later.

kisvegabor:
I'm using on Linux as well, but SDL should be the same on both places.

src/indev/lv_indev.c
bjsylvia: we need a new scroll event to distinguish its origin from the crown device.

bjsylvia:
@kisvegabor I'm trying to run it on Ubuntu and it works well. 
On the Mac system, there is no crash after compiling without relying on Xcode Clang. 

Regarding implementation, there is a suggestion: for the operation of the digital crown, LVGL only need to report the corresponding **event device type** and **scroll event**, without the need for conversion to key events. 

If the upper layer application or component needs to respond to such scroll events to change button status or adjust progress, it should be handled by the upper layer application logic. 
For example, on Google WearOS it notifies the scroll event from a rotary encoder device : 
https://developer.android.com/training/wearables/user-input/rotary-input
https://developer.android.com/reference/kotlin/androidx/compose/ui/input/rotary/RotaryScrollEvent

While on Apple WatchOS it reports digital crown rotation event : 
https://developer.apple.com/documentation/swiftui/digitalcrownevent

kisvegabor:
You can get the event device type by `lv_indev_get_type(lv_event_get_indev(e))` but I'm not sure if we should use a scroll event here for two reasons:
1. We don't have a scroll event now :slightly_smiling_face: `LV_EVENT_SCROLLED` is sent when the widget was already scrolled just to notify anyone who is interested in it
2. It's not really a scroll but an increment or decrement. 

What would be the  advantage of approaching from the direction of scroll? I understand that it used in Android, however we haven't followed Android's approach so far and selectively picking a few things might cause conflicts.

XuNeo:
From the meeting conclusion, using a 32bit number for key event together with its repeat count can eliminate the loop to send key event one by one.

However, the crown or rotary encoder has its own feature differs with keyboard. For example the sensitivity when scrolling list and roller.

For roller, press keyboard LEFT or RIGHT should focus next item. However, using rotary encoder, we may say rotate 10 steps will trigger a refocus to next item in roller.

Android and IOS have treated rotary encoder differently than keyboard, they must have reasons. It would be better we investigate the usage further.

kisvegabor:
As the feature doesn't any API we don't need to merge it for v9. It means we have some more time think about it. 

The sensitivity of the roller is a great example for why we need to treat it separately. Where shall we store this sensitivity value? As a style property? Or a normal widget property?

Probably we need a similar sensitivity for [scrolling](https://github.com/lvgl/lvgl/pull/5057/files#diff-35a05ebbf5595318ed115bcabc7f70e6ae0b4f1f3ace3629e154a7434fe680f1R1296-R1297) too.

XuNeo:
Not stale.

kisvegabor:
In summary what we should do is 
- having a new `LV_EVENT_ROTARY_SCROLL` and the `ticks` parameter. 
- send this event from `LV_INDEV_ENCODER` too 
- handle `LV_EVENT_ROTARY_SCROLL` in the existing widgets 
- add a sensitivity style property (1 by default for roller, dropdown, etc)
- make the scroll size depend on the ticks and the sensitivity of the widget

I think it's a good idea to merge this functionality into the pointer input device type as they share the focused object and the crown is usually used in together with a touch pad.

kisvegabor:
I added the ROTARY event. Finally added these to `lv_indev_t`:
```c
/**< Rotary diff count will be divided by this value when widgets are adjusted*/
int32_t rotary_adjust_divider;

/**< Rotary diff count will be multiples by this value when scrolling to get scroll throw size*/
int32_t rotary_scroll_sensitvity;
```

src/indev/lv_indev.c
```diff
@@ -1314,6 +1328,36 @@ static void indev_proc_release(lv_indev_t * indev)
     }
 }
 
+static void indev_proc_pointer_diff(lv_indev_t * indev)
+{
+    lv_obj_t * obj = indev->pointer.last_pressed;
```
XuNeo, line 1327: `    lv_obj_t * obj = indev->pointer.last_pressed;`
Should the user manually click the screen when a new page is loaded, in order to make the rotary works by default?

src/indev/lv_indev.c
```diff
@@ -1314,6 +1328,36 @@ static void indev_proc_release(lv_indev_t * indev)
...
+    bool editable = lv_obj_is_editable(obj);
+
+    if(editable) {
+        int32_t sensitivity = indev->rotary_adjust_divider;
+        int32_t diff = indev->pointer.diff / sensitivity;
+        send_event(LV_EVENT_ROTARY, &diff);
+    }
+    else {
+
+        int32_t sensitivity = indev->rotary_scroll_sensitvity;
```
XuNeo: Sensitivity could be a value specified per widget. For example an that has large items may want to scroll faster.

XuNeo (commented):
Thank you Gabor. I just have few comments want to confirm.

src/indev/lv_indev.c
kisvegabor: I was thinking about it too. What about setting it automatically to the screen when a new screen is loaded?

src/indev/lv_indev.c
kisvegabor: I think both are needed in this case. The number of clicks per turn is indev specific so you should be able to set up a unified behavior regardless to the rotary characteristics. 

We can add a `rotary_scroll_sensitivity;` style property.

Do we need a widget specific setting [here ](https://github.com/lvgl/lvgl/pull/5057/files/849512eb67de9449e9efa0fabe1d4bdb8ebcd1da#diff-35a05ebbf5595318ed115bcabc7f70e6ae0b4f1f3ace3629e154a7434fe680f1R1342)too? E.g. to adjust a slider with larger range faster.

XuNeo: > We can add a `rotary_scroll_sensitivity;` style property.

Agree. 



> Do we need a widget specific setting [here ](https://github.com/lvgl/lvgl/pull/5057/files/849512eb67de9449e9efa0fabe1d4bdb8ebcd1da#diff-35a05ebbf5595318ed115bcabc7f70e6ae0b4f1f3ace3629e154a7434fe680f1R1342)too? E.g. to adjust a slider with larger range faster.

Yes. Application can adjust the speed as they want by style.

src/indev/lv_indev.c
XuNeo: How about searching a scrollable obj instead if `last_pressed` is NULL?

src/indev/lv_indev.c
XuNeo: > We can add a rotary_scroll_sensitivity; style property.

Agree.

> Do we need a widget specific setting here too? E.g. to adjust a slider with larger range faster.

Yes. In this way, application can adjust the speed as they want by style.

kisvegabor:
I've slightly reworked it:
- Configured the base sensitivity in `indev->rotary_sensitvity` with 1/256 unit
- Configure the per widget sensitivity with the `rotary_sensitvity` style property with 1/256 unit again. For scrollable widgets you can set it to a larger value (set dpi / 4 in the default theme), but it's 256 by default, so slider, dropdown, etc are stepping one.

The final diff is calculated like:
```c
  uint32_t indev_sensitivity = indev->rotary_sensitvity;
  uint32_t obj_sensitivity = lv_obj_get_style_rotary_sensitivity(indev_obj_act, 0);
  int32_t diff = (int32_t)((int32_t)indev->pointer.diff * indev_sensitivity * obj_sensitivity + 32768) >> 16;
```

bjsylvia (approved):

src/core/lv_obj_style_gen.c
```diff
@@ -761,6 +761,14 @@ void lv_obj_set_style_bitmap_mask_src(lv_obj_t * obj, const lv_image_dsc_t * val
     };
     lv_obj_set_local_style_prop(obj, LV_STYLE_BITMAP_MASK_SRC, v, selector);
 }
+
+void lv_obj_set_style_rotary_sensitivity(lv_obj_t * obj, uint32_t value, lv_style_selector_t selector)
```
PGNetHun, line 766: Well, I think the rotary sensitivity is a behavioural property, not a UI style attribute.
But since there is no other way to store behavioural properties for objects, "hacking" into style properties is OK, too.

Btw it would be good to have a way to store object behavioural attributes, as there could be other use cases also to store such properties. One example of behavioural attribute is "checkable", which is at the moment a state, and only some objects make use of it.

src/drivers/sdl/lv_sdl_mouse.c
```diff
@@ -174,6 +185,17 @@ void _lv_sdl_mouse_handler(SDL_Event * event)
...
             break;
+        case SDL_MOUSEWHEEL:
+#if LV_SDL_MOUSEWHEEL_MODE == 1
+#ifdef __EMSCRIPTEN__
+            /*Escripten scales it wrong*/
+            if(event->wheel.y < 0) dsc->diff++;
+            if(event->wheel.y > 0) dsc->diff--;
+#else
+            indev_dev->diff = -event->wheel.y;
+#endif  /*LV_SDL_MOUSEWHEEL_MODE*/
```
PGNetHun:
```suggestion
#endif  /* __EMSCRIPTEN__ */
```

src/themes/default/lv_theme_default.c
```diff
@@ -263,6 +264,7 @@ static void style_init(my_theme_t * theme)
     lv_style_set_text_font(&theme->styles.scr, theme->base.font_normal);
     lv_style_set_pad_row(&theme->styles.scr, PAD_SMALL);
     lv_style_set_pad_column(&theme->styles.scr, PAD_SMALL);
+    lv_style_set_rotary_sensitivity(&theme->styles.scr, theme->disp_dpi / 4 * 256);
```
PGNetHun, line 267: `4 * 256` sounds like a magic number.
I assume you have tested with SDL, so I wonder if it's a good value for real devices, too.

src/drivers/sdl/lv_sdl_mouse.c
```diff
@@ -35,6 +35,9 @@ typedef struct {
     int16_t last_x;
     int16_t last_y;
     bool left_button_down;
+#if LV_SDL_MOUSEWHEEL_MODE == 1
```
PGNetHun: I think using "constants" (in our case new `SD_MOUSEWHEEL_MODE_ENCODER` and `SDL_MOUSEWHEEL_MODE_CROWN` macros) is better and more descriptive than repeating the values `0` and `1` in files.
Now we know what `0` and `1` mean, but later we will forget, and have to remember, or investigate what they mean.

PGNetHun (changes requested):

src/core/lv_obj_style_gen.c
kisvegabor: > But since there is no other way to store behavioural properties for objects, "hacking" into style properties is OK, too.

It might make sense to say the sensitivity is a design choice too. :slightly_smiling_face: 
 
> Btw it would be good to have a way to store object behavioural attributes, as there could be other use cases also to store such properties. One example of behavioural attribute is "checkable", which is at the moment a state, and only some objects make use of it.

Nitpicking: it's a flag.

I was also thinking about it. Maybe storing widget properties as we store style properties? Probably we could save some RAM this way, as most of the properties are not used.

src/themes/default/lv_theme_default.c
kisvegabor: `theme->disp_dpi / 4` is 1/4 inch, which is an empirical value. `* 256` is just for scaling as `rotary_sensitivity` uses 1/256 unit.

src/drivers/sdl/lv_sdl_mouse.c
kisvegabor: True, I'll fix these.

src/core/lv_obj_style_gen.c
PGNetHun: Yeah, checkable is flag, checked is state :)

PGNetHun: Well, the actual flags are stored quite compact (1 bit/flag => sum 4 bytes), so at the moment I can't suggest any better option.

XuNeo (dismissed):
Thank you Gabor! All looks good to me.

kisvegabor:
I've updated the docs and added `LV_SDL_MOUSEWHEEL_MODE_ENCODER/CROWN` as @PGNetHun suggested. `enum`s can't be used here as the `enum` values are not resolved at the pre-processing stage so they can't be used in `#if`s.

src/drivers/sdl/lv_sdl_window.c
```diff
@@ -253,7 +253,9 @@ static void sdl_event_handler(lv_timer_t * t)
     SDL_Event event;
     while(SDL_PollEvent(&event)) {
         _lv_sdl_mouse_handler(&event);
+#if LV_SDL_MOUSEWHEEL_MODE == 0
```
PGNetHun: Forget to add `LV_SDL_MOUSEWHEEL_MODE_ENCODER`

src/drivers/sdl/lv_sdl_mousewheel.h
```diff
@@ -14,7 +14,7 @@ extern "C" {
  *      INCLUDES
  *********************/
 #include "lv_sdl_window.h"
-#if LV_USE_SDL
+#if LV_USE_SDL && LV_SDL_MOUSEWHEEL_MODE == 0
```
PGNetHun: Forget to add `LV_SDL_MOUSEWHEEL_MODE_ENCODER`

PGNetHun (changes requested):

kisvegabor:
@PGNetHun 
Fixed, thanks!

PGNetHun (approved):

XuNeo (approved):

PR #5061, opened 2023-12-19
refact(anim_timeline): refact anim timeline design

FASTSHIFT:

1. Hide the `lv_anim_timeline_t` data structure
2. Rename `lv_anim_timeline_stop` -> `lv_anim_timeline_pause`
3. Use a single anim to control the overall animation progress instead of independent anim
4. Add progress record, you can pause and resume animation from any time
5. Remove all global variables of the anim timeline example and add slider linkage

https://github.com/lvgl/lvgl/assets/26767803/34630f43-2dd6-4e5e-8324-bb817e8aa547

### Checkpoints
 
### DIFF
--- a/src/misc/lv_anim_timeline.h
+++ b/src/misc/lv_anim_timeline.h
@@ -19,22 +19,13 @@ extern "C" {
  *      DEFINES
  *********************/
 
+#define LV_ANIM_TIMELINE_PROGRESS_MAX 0xFFFF
+
 /**********************
  *      TYPEDEFS
  **********************/
 
-/*Data of anim_timeline_dsc*/
-typedef struct {
-    lv_anim_t anim;
-    uint32_t start_time;
-} lv_anim_timeline_dsc_t;
-
-/*Data of anim_timeline*/
-typedef struct {
-    lv_anim_timeline_dsc_t * anim_dsc;  /**< Dynamically allocated anim dsc array*/
-    uint32_t anim_dsc_cnt;              /**< The length of anim dsc array*/
-    bool reverse;                       /**< Reverse playback*/
-} lv_anim_timeline_t;
+typedef struct _lv_anim_timeline_t lv_anim_timeline_t;
 
 /**********************
 * GLOBAL PROTOTYPES
@@ -58,7 +49,7 @@ void lv_anim_timeline_delete(lv_anim_timeline_t * at);
  * @param start_time    the time the animation started on the timeline, note that start_time will override the value of delay.
  * @param a             pointer to an animation.
  */
-void lv_anim_timeline_add(lv_anim_timeline_t * at, uint32_t start_time, lv_anim_t * a);
+void lv_anim_timeline_add(lv_anim_timeline_t * at, uint32_t start_time, const lv_anim_t * a);
 
 /**
  * Start the animation timeline.
@@ -68,10 +59,10 @@ void lv_anim_timeline_add(lv_anim_timeline_t * at, uint32_t start_time, lv_anim_
 uint32_t lv_anim_timeline_start(lv_anim_timeline_t * at);
 
 /**
- * Stop the animation timeline.
+ * Pause the animation timeline.
  * @param at    pointer to the animation timeline.
  */
-void lv_anim_timeline_stop(lv_anim_timeline_t * at);
+void lv_anim_timeline_pause(lv_anim_timeline_t * at);
 
 /**
  * Set the playback direction of the animation timeline.
@@ -100,6 +91,13 @@ uint32_t lv_anim_timeline_get_playtime(lv_anim_timeline_t * at);
  * @return return true if it is reverse playback.
  */
 bool lv_anim_timeline_get_reverse(lv_anim_timeline_t * at);
+
+/**
+ * Get the progress of the animation timeline.
+ * @param at    pointer to the animation timeline.
+ * @return return value 0~65535 to map 0~100% animation progress.
+ */
+uint16_t lv_anim_timeline_get_progress(lv_anim_timeline_t * at);
 
 /**********************
  *      MACROS
--- a/docs/overview/anim.rst
+++ b/docs/overview/anim.rst
@@ -147,6 +147,9 @@ timeline.
 
 It supports forward and backward playback of the entire animation group,
 using :cpp:expr:`lv_anim_timeline_set_reverse(at, reverse)`.
+Note that if you want to play in reverse from the end of the timeline,
+you need to call :cpp:expr:`lv_anim_timeline_set_progress(at, LV_ANIM_TIMELINE_PROGRESS_MAX)`
+after adding all animations and before starting to play.
 
 Call :cpp:expr:`lv_anim_timeline_stop(at)` to stop the animation timeline.
 
@@ -159,8 +162,9 @@ duration of the entire animation timeline.
 Call :cpp:expr:`lv_anim_timeline_get_reverse(at)` function to get whether to
 reverse the animation timeline.
 
-Call :cpp:expr:`lv_anim_timeline_delete(at)` function to delete the animation
-timeline.
+Call :cpp:expr:`lv_anim_timeline_delete(at)` function to delete the animation timeline.
+**Note**: If you need to delete an object during animation, be sure to delete the
+anim timeline before deleting the object. Otherwise, the program may crash or behave abnormally.
 
 .. image:: /misc/anim-timeline.png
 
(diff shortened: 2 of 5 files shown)


src/misc/lv_anim_timeline.h
```diff
@@ -101,6 +92,13 @@ uint32_t lv_anim_timeline_get_playtime(lv_anim_timeline_t * at);
  */
 bool lv_anim_timeline_get_reverse(lv_anim_timeline_t * at);
 
+/**
+ * Get the progress of the animation timeline.
+ * @param at    pointer to the animation timeline.
+ * @return return value 0~65535 to map 0~100% animation progress.
+ */
+uint16_t lv_anim_timeline_get_progress(lv_anim_timeline_t * at);
```
kisvegabor, line 100: It's not  implemented in the C file

kisvegabor:
Sounds and look good!

src/misc/lv_anim_timeline.h
FASTSHIFT: Done

FASTSHIFT:
image
The test program asserts here, and it is found that `anim_timeline_exec_cb` does not seem to be called in `lv_test_wait(1)`, and is being investigated.

FASTSHIFT:
> image The test program asserts here, and it is found that `anim_timeline_exec_cb` does not seem to be called in `lv_test_wait(1)`, and is being investigated.

It should be a clock accuracy problem. `lv_test_wait(1)` is too short, causing the animation value calculated by `path_cb` to not change, resulting in the coordinates not being updated.

kisvegabor:
However it was working previously. Or were there changes affecting this part as well?

kisvegabor:
Thank you for looking into it. I spent a good amount of time to make it work well for each millisecond. It would be nice if these precision could be maintained.

FASTSHIFT:
> Thank you for looking into it. I spent a good amount of time to make it work well for each millisecond. It would be nice if these precision could be maintained.

The accuracy issue is solved, but the animation overlap issue doesn't seem to work well since both overlapping animations will set the value :(

FASTSHIFT:
To sum up, using an anim to control the progress of all animations can achieve animation pauses and true reverse playback (including the animation curve is also reversed, instead of simply exchanging start/end).
However, it will also cause animation overlap conflicts, and if obj is deleted immediately while playing the animation, it will also cause a crash, because the life cycle of anim_timeline will not automatically end when obj is deleted.

kisvegabor:
I can help to debuwith it. Can you send a minimal code example to demonstrate the issue? Just with two overlapping animations.

FASTSHIFT:
> I can help to debuwith it. Can you send a minimal code example to demonstrate the issue? Just with two overlapping animations.

Thank you so much! The test code is the original code copied from `test_anim_timeline`. It will assert in the third `TEST_ASSERT_EQUAL`.

```c
#define TEST_ASSERT_EQUAL(a, b) LV_ASSERT(a == b)

void lv_test_wait(uint32_t ms)
{
    lv_tick_inc(ms);
    lv_timer_handler();
    lv_refr_now(NULL);
}

void test_anim_timeline_reverse(void)
{
    lv_obj_t * obj = lv_obj_create(lv_scr_act());
    lv_obj_set_size(obj, 100, 100);

    lv_obj_set_pos(obj, 70, 70);

    /* Reverse this:
     *       |------1------|
     *                             |-----Y-----|
     *                                   |--------Z--------|
     *   0  200           1000   1500   2000  2500        3500    ms
     *   ?  50            100     200         300         700     value (px)
     */

    lv_anim_timeline_t * anim_timeline = lv_anim_timeline_create();
    lv_anim_timeline_set_reverse(anim_timeline, true);

    lv_anim_t a1;
    lv_anim_init(&a1);
    lv_anim_set_exec_cb(&a1, (lv_anim_exec_xcb_t)lv_obj_set_x);
    lv_anim_set_var(&a1, obj);
    lv_anim_set_values(&a1, 50, 100);
    lv_anim_set_duration(&a1, 800);
    lv_anim_set_early_apply(&a1, false);
    lv_anim_timeline_add(anim_timeline, 200, &a1);

    lv_anim_set_values(&a1, 200, 300);
    lv_anim_set_duration(&a1, 1000);
    lv_anim_timeline_add(anim_timeline, 1500, &a1);

    /*Overlap with the previous*/
    lv_anim_set_values(&a1, 400, 700);
    lv_anim_set_duration(&a1, 1500);
    lv_anim_timeline_add(anim_timeline, 2000, &a1);

    lv_anim_timeline_set_progress(anim_timeline, LV_ANIM_TIMELINE_PROGRESS_MAX);
    lv_anim_timeline_start(anim_timeline);    /*0 (3500)*/

    lv_refr_now(NULL);
    TEST_ASSERT_EQUAL(700, lv_obj_get_x(obj));

    lv_test_wait(999);                   /*999 (2501)*/
    TEST_ASSERT_EQUAL(500, lv_obj_get_x(obj));

    lv_test_wait(201);                   /*1200 (2300)*/
    TEST_ASSERT_EQUAL(280, lv_obj_get_x(obj));

    lv_test_wait(300);                   /*1500 (2000)*/
    TEST_ASSERT_EQUAL(250, lv_obj_get_x(obj));

    lv_test_wait(500);                    /*2000 (1500)*/
    TEST_ASSERT_EQUAL(200, lv_obj_get_x(obj));

    /*There is a gap*/
    lv_test_wait(100);                    /*2100 (1400)*/
    TEST_ASSERT_EQUAL(200, lv_obj_get_x(obj));

    /*Nothing should change it*/
    lv_obj_set_x(obj, 10);

    lv_test_wait(100);                    /*2200 (1300)*/
    TEST_ASSERT_EQUAL(10, lv_obj_get_x(obj));

    lv_test_wait(300);                    /*2500 (1000)*/
    TEST_ASSERT_EQUAL(100, lv_obj_get_x(obj));

    lv_test_wait(400);                    /*2900 (600)*/
    TEST_ASSERT_EQUAL(75, lv_obj_get_x(obj));

    lv_test_wait(400);                    /*3300 (200)*/
    TEST_ASSERT_EQUAL(50, lv_obj_get_x(obj));

    /*There is a gap*/
    lv_test_wait(100);                    /*3400 (100)*/
    TEST_ASSERT_EQUAL(50, lv_obj_get_x(obj));

    /*Nothing should change it*/
    lv_obj_set_x(obj, 20);
    lv_test_wait(100);
    TEST_ASSERT_EQUAL(20, lv_obj_get_x(obj));

    lv_test_wait(1000);
    TEST_ASSERT_EQUAL(20, lv_obj_get_x(obj));
}
```

It is necessary to comment out the timestamp settings of SDL and allow the timestamps to accumulate freely:
image

kisvegabor:
I think I understand what's going on
```c
    /* Reverse this:
     *       |------1------|
     *                             |-----2-----|
     *                                   |--------3--------|
     *   0  200           1000   1500   2000  2500        3500    ms
     *   ?  50            100     200    400   300         700     value (px)
     */
```

When not reversed, there is a jump at t=2000: anim2 is around value=250, but anim3 overwrites it to value=400. At t=2300 still both are active but anim3 is stronger. 

In the current (`master`) implementation when reversed, this line in the test
```c
    lv_test_wait(201);                   /*1200 (2300)*/
    TEST_ASSERT_EQUAL(280, lv_obj_get_x(obj));
``` 
suggest that at 2300 `anim2` is stronger. Which might or might not make sense.
1. When reversed the order of animations should be reversed too, so anim2 will be applied later and it will have higher precedence. (`master` behavior)
2. Or the order should be the same and the animation should be played just backwards. (this PR behavior)

I think 2) is more intuitive. If you agree the all we need to do is update the excepted values in test to reflect to this behavior.
 
---

Use case: let's say there only a single overshooting X animation (from 100 to 300) in the time line. When played normally the widget will move with an overshot at the end. It's fine. 

However what should happen when played backwards?   
1. In this PR will start with the overshot
2. Or is it more expected to play it at end (more complicated)

It think 2) is less intuitive and if it's needed a new timeline shall be created for the opposite direction.

---

Deleting the timeline with the widget: 
I think it's complicated to solve because it's not clear whet to do if a timeline uses 3 widgets and only one of them is deleted. Drop the whole timeline? Or only the related animation should be dropped.

I think we can leave it to the user, just document it.

FASTSHIFT:
> I think 2) is more intuitive. If you agree the all we need to do is update the excepted values in test to reflect to this behavior.

I think the same thing as you, I will update the testcase.

> I think we can leave it to the user, just document it.

OK, let me update the documentation.

FASTSHIFT:
@kisvegabor 
The testcase and documentation have been modified. Now we still have `lv_example_anim_timeline_1.py` that has not been modified. Can anyone help?

Issue #5127, opened 2023-12-29
how use the LV_DRAW_BUF_DEFINE to load the image?


xinsuinizhuan:
![图片](https://github.com/lvgl/lvgl/assets/40679769/f31e9f60-3478-4199-a2b8-31afe4525b42)

i also inlucde the lv_draw_buf.h file,but  some compile error

xinsuinizhuan:
I also add the .c files, but still could not link:
![图片](https://github.com/lvgl/lvgl/assets/40679769/32a5b558-5953-44f6-b653-71c19a8799af)

XuNeo:
It looks like you haven't include the file to build. It complains no symbol can be found.

xinsuinizhuan:
> It looks like you haven't include the file to build. It complains no symbol can be found.

how to include the file to build? I also inlcude the .c file in my project.

xinsuinizhuan:
![图片](https://github.com/lvgl/lvgl/assets/40679769/c6c25984-610c-44b8-b434-de5e051b1bfa)

XuNeo:
I'm not familiar with vscode. But the error suggests the symbol is not linked in. You can try to add faulty code to img_hand.c and compile to see if it reports error.

xinsuinizhuan:
> I'm not familiar with vscode. But the error suggests the symbol is not linked in. You can try to add faulty code to img_hand.c and compile to see if it reports error.

I  add the faulty img_hand.c file to my project, if have some head file to add, or set some define to 1?

xinsuinizhuan:
I resolve the problem， i create the preject is c++, and the main file is .cpp file, the .cpp file to load the imge .c buffer file, load false, it need to add extern "C".

kisvegabor:
As the issue is resolved, I close it.

Issue #5054, opened 2023-12-19
Use `image_buf` instead of `draw_buf`


XuNeo:
### Introduce the problem

The newly introduced `draw_buf` has identical definition with `lv_image_dst_t`. 
We can use the existing `lv_image_buf` naming. The `lv_image_buf_free` is currently not used by `lvgl`.

> `lv_image_create()` is already used by the `lv_image` widget, however `lv_image_buf_*` could work. 
> 
> It really opens to door for some simplification and removal of duplication. 
> 
> Let's try it, if it really looks good we are still on time to add this change. 
> 
> I merge it now, to no block the development of with nitpicking on naming.

_Originally posted by @kisvegabor in https://github.com/lvgl/lvgl/issues/5034#issuecomment-1860407825_

XuNeo:
cc @FASTSHIFT .
I think currently people cannot distinguish draw buffer from image. 
See the title of this issue. https://github.com/lvgl/lvgl/issues/5127

Separating draw buffer with image seems not a good choice from my point of view.

kisvegabor:
I agree to unify `lv_image_buf` and `lv_draw_buf`. Less building blocks is better.

kisvegabor:
no stale

kisvegabor:
Being discussed in #5496 too

PR #5132, opened 2023-12-29
feat(draw_buf): add draw_buf init API

XuNeo:

It's useful to convert existing memory to draw buffer to match other APIs. Init a draw buf means the memory comes from user, and user should take care of it instead of calling draw_buf_destroy.

### Checkpoints
 
### DIFF
--- a/src/draw/lv_draw_buf.c
+++ b/src/draw/lv_draw_buf.c
@@ -135,6 +135,27 @@ void lv_draw_buf_copy(void * dest_buf, uint32_t dest_w, uint32_t dest_h, const l
     }
 }
 
+lv_result_t lv_draw_buf_init(lv_draw_buf_t * draw_buf, uint32_t w, uint32_t h, lv_color_format_t cf, uint32_t stride,
+                             void * data, uint32_t data_size)
+{
+    LV_ASSERT_NULL(draw_buf);
+    if(draw_buf == NULL) return LV_RESULT_INVALID;
+
+    lv_memzero(draw_buf, sizeof(lv_draw_buf_t));
+    if(stride == 0) stride = lv_draw_buf_width_to_stride(w, cf);
+    if(stride * h > data_size) {
+        LV_LOG_WARN("Data size too small, required: %" LV_PRId32 ", provided: %" LV_PRId32, stride * h,
+                    data_size);
+        return LV_RESULT_INVALID;
+    }
+
+    lv_image_header_init(&draw_buf->header, w, h, cf, stride, 0);
+    draw_buf->data = lv_draw_buf_align(data, cf);
+    draw_buf->unaligned_data = data;
+    draw_buf->data_size = data_size;
+    return LV_RESULT_OK;
+}
+
 lv_draw_buf_t * lv_draw_buf_create(uint32_t w, uint32_t h, lv_color_format_t cf, uint32_t stride)
 {
     lv_draw_buf_t * draw_buf = lv_malloc_zeroed(sizeof(lv_draw_buf_t));
--- a/src/draw/lv_draw_buf.h
+++ b/src/draw/lv_draw_buf.h
@@ -174,12 +174,26 @@ void lv_draw_buf_copy(void * dest_buf, uint32_t dest_w, uint32_t dest_h, const l
  * that meets specified requirements.
  *
  * @param w         the buffer width in pixels
- * @param h
+ * @param h         the buffer height in pixels
  * @param cf        the color format for image
  * @param stride    the stride in bytes for image. Use 0 for automatic calculation based on
  *                  w, cf, and global stride alignment configuration.
  */
 lv_draw_buf_t * lv_draw_buf_create(uint32_t w, uint32_t h, lv_color_format_t cf, uint32_t stride);
+
+/**
+ * Initialize a draw buf with the given buffer and parameters.
+ * @param draw_buf  the draw buf to initialize
+ * @param w         the buffer width in pixels
+ * @param h         the buffer height in pixels
+ * @param cf        the color format
+ * @param stride    the stride in bytes. Use 0 for automatic calculation
+ * @param data      the buffer used for drawing. Unaligned `data` will be aligned internally
+ * @param data_size the size of the buffer in bytes
+ * @return          return LV_RESULT_OK on success, LV_RESULT_INVALID otherwise
+ */
+lv_result_t lv_draw_buf_init(lv_draw_buf_t * draw_buf, uint32_t w, uint32_t h, lv_color_format_t cf, uint32_t stride,
+                             void * data, uint32_t data_size);
 
 /**
  * Duplicate a draw buf with same image size, stride and color format. Copy the image data too.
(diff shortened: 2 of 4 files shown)


XuNeo:
I want to raise the discussion about `draw_buf` lifecycle. Current draw buf actually contains two pointers to free. One is the struct itself, another is the buffer itself. Destroy a draw buffer means free them all.

However, we have seen situations to convert existing memory to a draw buffer interface. It becomes tricky to say `destroy` a draw buffer, since it's not actually `created` before.

Using an `init` API, emphasis that memory is provided by user, and it just initializes the struct. Both memories should be managed by user.

FASTSHIFT:
My point of view is to think of lv_draw_buf_t as the same object as lv_obj_t, and all are allocated on the heap. draw_buf->data is one of the attributes of lv_draw_buf_t, which is analogous to lv_img->src. lv_obj_t never allows static allocation.

kisvegabor:
I see these main parts to consider:
- image decoding
- canvas
- if `draw_buf` and `image_dsc` will be unified consider normal const images too

@XuNeo knows the current use cases of `draw_buf` better than me, but I think especially for const images we can't say that `draw_buf` is always allocated as the widgets.

In general I think `lv_draw_buf_create` could set a flag to indicate that the `draw_buf` was allocated. If it was, the widgets, canvas, image decoder, etc, should free it. Similarly `lv_draw_buf_malloc()` could also mark the buffer as malloc-ed and subject to be freed. 

A function like `lv_draw_buf_clean_up(draw_buf)` could free what needs to be freed.

FASTSHIFT:
If static allocation of draw_buf is allowed, we may not be able to hide user access to draw_buf members.

FASTSHIFT:
Although the data structures of lv_image_dsc_t and lv_draw_buf_t are similar, I think they are responsible for describing different contents. lv_image_dsc_t is used to store read-only image information, and users allow static allocation. lv_draw_buf_t is a structure used to describe and transfer all image information in lvgl, including display, layer, font, and decoded image. It is readable and writable. We can provide a conversion method from lv_image_dsc_t to lv_draw_buf_t, because there is no guarantee that lv_draw_buf_t will frequently add more complex data structures in the future. If lv_image_dsc_t and lv_draw_buf_t are unified, the image resources exported by the image conversion tool will be inconvenient and forward-compatible.

kisvegabor:
> If static allocation of draw_buf is allowed, we may not be able to hide user access to draw_buf members.

What about `lv_draw_buf_set_buf(draw_buf, buf, flags)`? `flags` can tell to whether automatically free the buffer or not. 

> the image resources exported by the image conversion tool will be inconvenient and forward-compatible.

Fair point. We can think of a gradual an experimental change. Let's say we assume it's possible to unify the image_dsc and draw_buf, but we do add only `typedef lv_draw_buf_t lv_image_dsc_t` first and treat them the same internally. In a few month we can update the converter to output `lv_draw_buf_t` which should work just the same.

XuNeo:
It seems we have to rely on `flags`.  Since we are pretty sure `draw buffer` is different than `image`, let's add more specific flags only for `draw buf`.

Flags like `LV_IMAGE_FLAGS_ALLOCATED` should only be suitable for `draw buffer`.

XuNeo:
Let's focus on the lifecycle management of draw buffer here, and discuss unififying draw buf and image buf in https://github.com/lvgl/lvgl/issues/5054.

xinsuinizhuan:
I want use a image, i also think to load it by a image path firstly,  but now it is a image buff and have no head file, and i could complie find it.

xinsuinizhuan:
Or make a .h head file, define all the image buffers, as a stantic, make a startup function, then load the image buffer, so if i inchude the .h head file, find it by compile normaly.

XuNeo:
> I want use a image, i also think to load it by a image path firstly, then when the parent widget destroy, destroy it.But now it is a image buff and load it in a .c file, have no head file, i include the .c file in my project, but could not find it normaly by compiles.

Let's discuss the issue you met in https://github.com/lvgl/lvgl/issues/5127.

XuNeo:
I included a use case of `lv_draw_buf_init` which explains the idea "Init a draw buf means the memory comes from user, and user should take care of it instead of calling draw_buf_destroy."

src/draw/lv_draw_buf.c
```diff
@@ -134,6 +134,28 @@ void lv_draw_buf_copy(void * dest_buf, uint32_t dest_w, uint32_t dest_h, const l
     }
 }
 
+lv_result_t lv_draw_buf_init(lv_draw_buf_t * draw_buf, uint32_t w, uint32_t h, lv_color_format_t cf, uint32_t stride,
+                             void * ptr, uint32_t data_size)
```
PGNetHun: `*ptr` should be `*data`

src/draw/lv_draw_buf.h
```diff
@@ -181,6 +181,20 @@ void lv_draw_buf_copy(void * dest_buf, uint32_t dest_w, uint32_t dest_h, const l
...
+ * @param draw_buf  the draw buf to initialize
+ * @param w         the buffer width in pixels
+ * @param h
+ * @param cf        the color format for image
+ * @param stride    the stride in bytes for image. Use 0 for automatic calculation.
+ * @param ptr       the buffer to use for image data. If NULL, a buffer will be allocated.
+ * @param data_size the size of the buffer in bytes
+ * @return          LV_RESULT_OK: success
+ */
+lv_result_t lv_draw_buf_init(lv_draw_buf_t * draw_buf, uint32_t w, uint32_t h, lv_color_format_t cf, uint32_t stride,
```
PGNetHun, line 195: `*ptr` should be `*data`

src/draw/lv_draw_buf.h
```diff
@@ -181,6 +181,20 @@ void lv_draw_buf_copy(void * dest_buf, uint32_t dest_w, uint32_t dest_h, const l
...
 lv_draw_buf_t * lv_draw_buf_create(uint32_t w, uint32_t h, lv_color_format_t cf, uint32_t stride);
 
+/**
+ * Initialize a draw buf with the given buffer and parameters.
+ * @param draw_buf  the draw buf to initialize
+ * @param w         the buffer width in pixels
+ * @param h
+ * @param cf        the color format for image
+ * @param stride    the stride in bytes for image. Use 0 for automatic calculation.
+ * @param ptr       the buffer to use for image data. If NULL, a buffer will be allocated.
```
PGNetHun: `ptr` should be `data`.

The text "_use for image data_" is misleading I think, because we could misinterpret it with the image object. Better text would be: "_the buffer to use for the drawing data._"

PGNetHun (changes requested):

src/draw/lv_draw_buf.h
PGNetHun: I think we should generally avoid using the term "image" in case of `lv_draw_buf`, because it is used for image objects (`lv_image`), and for me an image represents a whole image (decoded PNG or JPG, etc).

**It's just my opinion.**

I suggest to use the term "drawing", as the file name reflects that also: `lv_draw_buf`

Examples: "_color format for image_" -> "_color format for drawing_",  "_image data_" -> "_drawing data_", etc.

What do you think?

XuNeo: I agree. I have updated to use `drawing` or not mention `image` at all. Please help to review it.

PGNetHun (approved):

FASTSHIFT (approved):

PR #5136, opened 2023-12-30
fix(display): fix inconsistency in the API

kisvegabor:

### Checkpoints
 

kisvegabor:
@kdschlosser, please take a look.

kdschlosser:
it looks good... I haven't tested it at all yet. I do have a question tho. does LVGL support multiple displays being attached at one time?

kisvegabor:
> does LVGL support multiple displays being attached at one time?

Yes, see [here](https://docs.lvgl.io/master/overview/disp.html#multiple-display-support).

kdschlosser:
If there is multi display support then this code is not right

```
void lv_display_load_screen(lv_display_t * disp, lv_obj_t * scr)
{
    LV_UNUSED(disp); /*Not required, added just for consistency*/
    lv_screen_load_anim(scr, LV_SCR_LOAD_ANIM_NONE, 0, 0, false);
}
```

Because you are not loading a screen to a specific display you are loading it to whatever display is currently being used. adding the display as a parameter only for the purposes of the API should not be done. The screen should actually get added to the display that was passed into it. There should be a requirement for the display to be valid. the lv_screen_load should locate the display that is currently in use and pass that when it makes the call to lv_display_load_screen.

kisvegabor:
The screen is already created on a given display. So from the screen widget LVGL get the related display, e.g. with `lv_obj_get_disp` (I'm updating it to `lv_obj_get_display`)

FASTSHIFT (dismissed):

kdschlosser:
OK so when I do the following..

```c
lv_obj_t *new_screen = lv_create_obj(NULL);
lv_display_load_screen(some_display, new_screen);
```

What display does the new screen get attached to? If I have more than a single display `some_display` might not be the one that is active. but LVGL is making the assumption that when I want to load a screen that means that I want to see it on the active display and that might not be the case.

see what I am getting at now?

src/display/lv_display.c
```diff
@@ -492,8 +492,9 @@ lv_obj_t * lv_display_get_screen_prev(lv_display_t * disp)
     return disp->prev_scr;
 }
 
-void lv_display_load_scr(lv_obj_t * scr)
+void lv_display_load_screen(lv_display_t * disp, lv_obj_t * scr)
```
XuNeo: Need to add to `lv_api_map.h`

src/core/lv_obj_pos.c
```diff
@@ -282,7 +282,7 @@ void lv_obj_mark_layout_as_dirty(lv_obj_t * obj)
     scr->scr_layout_inv = 1;
 
     /*Make the display refreshing*/
-    lv_display_t * disp = lv_obj_get_disp(scr);
+    lv_display_t * disp = lv_obj_get_display(scr);
```
XuNeo, line 285: Need add `lv_obj_get_display ` to api map.

src/indev/lv_indev.h
```diff
@@ -137,9 +137,9 @@ lv_indev_state_t lv_indev_get_state(const lv_indev_t * indev);
 
 lv_group_t * lv_indev_get_group(const lv_indev_t * indev);
 
-struct _lv_display_t * lv_indev_get_disp(const lv_indev_t * indev);
+struct _lv_display_t * lv_indev_get_display(const lv_indev_t * indev);
 
-void lv_indev_set_disp(lv_indev_t * indev, struct _lv_display_t * disp);
+void lv_indev_set_display(lv_indev_t * indev, struct _lv_display_t * disp);
```
XuNeo, line 134: Add `lv_indev_get_display` to api map.

src/display/lv_display.c
kisvegabor: Finally I removed this function as loading a screen is not related to the display. `lv_screen_load()` can be used directly.

src/core/lv_obj_pos.c
kisvegabor: Added

src/indev/lv_indev.h
kisvegabor: Added

kisvegabor:
> What display does the new screen get attached to?

`lv_obj_create(NULL)` will create a screen on the current default display. Which is the first display the your create or what you set with `lv_display_set_default`. There is no way to cross-load screens among displays. A screen belongs to the display where your have created it. 

To resolve this inconsistency I removed `lv_display_load_screen` and users can use `lv_screen_load` directly.

XuNeo (approved):
Need to solve the conflicts firstly.

FASTSHIFT (dismissed):

kisvegabor:
> Need to solve the conflicts firstly.

Resolved

src/display/lv_display.h
```diff
@@ -329,6 +329,15 @@ lv_obj_t * lv_display_get_screen_active(lv_display_t * disp);
...
 /**
+<<<<<<< HEAD
+=======
+ * Make a screen active
+ * @param scr       pointer to a screen
+ */
+void lv_display_load_scr(lv_obj_t * scr);
+
+/**
+>>>>>>> master
```
XuNeo:
```suggestion
```

PGNetHun (approved):

XuNeo (approved):

FASTSHIFT (approved):

FASTSHIFT (approved):

PR #5144, opened 2023-12-31
chore(docs): fix some outdated information

PetteriAimonen:
Just a few small fixes to documentation that I stumbled upon.
No functional changes.

As an aside, is it intentional that [https://docs.lvgl.io/latest/](https://docs.lvgl.io/latest/en/html/intro/index.html) is v7.11.0-dev and not the latest version? Most google hits for lvgl end up there, and I was confused for a moment until I noticed the version number in sidebar, as I only looked at the URL.

Oh, and happy new year to the very cool library!

### DIFF
--- a/docs/overview/coord.rst
+++ b/docs/overview/coord.rst
@@ -440,9 +440,13 @@ the given layout.
 Built-in layout
 ---------------
 
-LVGL comes with two very powerful layouts: - Flexbox - Grid
+LVGL comes with two very powerful layouts:
+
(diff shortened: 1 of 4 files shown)


kisvegabor:
Hi Petteri,

Thank you for the fixes and also for pointing out the `latest` issue. We kept it pointing to 7.11 for historical reasons, but I think enough time passed and it's ok to make it point to `master`. So I updated it.

kisvegabor (approved):

FASTSHIFT (approved):

PR #5161, opened 2024-01-02
chore: move lv_obj_t define to lv_types.h

XuNeo:

Continue of #4380. 
Micropython should now has no complain, and it's also reasonable to move `lvgl` defined types to `lv_types.h`.

### Checkpoints

### DIFF
--- a/docs/others/obj_property.rst
+++ b/docs/others/obj_property.rst
@@ -24,8 +24,8 @@ Two APIs are provided to get/set widget properties. It can be enabled by setting
         };
     } lv_property_t;
 
-    lv_result_t lv_obj_set_property(struct _lv_obj_t * obj, const lv_property_t * value);
-    lv_property_t lv_obj_get_property(struct _lv_obj_t * obj, lv_prop_id_t id);
+    lv_result_t lv_obj_set_property(lv_obj_t * obj, const lv_property_t * value);
+    lv_property_t lv_obj_get_property(lv_obj_t * obj, lv_prop_id_t id);
 
 
 Property ID
--- a/docs/overview/draw.rst
+++ b/docs/overview/draw.rst
@@ -267,7 +267,7 @@ documentation.
 
     typedef struct {
         lv_draw_ctx_t * draw_ctx;           /**< Draw context*/
-        const struct _lv_obj_class_t * class_p;     /**< The class that sent the event */
+        const lv_obj_class_t * class_p;     /**< The class that sent the event */
         uint32_t type;                      /**< The type if part being draw. Element of `lv_<name>_draw_part_type_t` */
         lv_area_t * draw_area;              /**< The area of the part being drawn*/
         lv_draw_rect_dsc_t *
--- a/docs/overview/img.rst
+++ b/docs/overview/img.rst
@@ -244,7 +244,7 @@ open/close the PNG files. It should look like this:
     * @param header    image information is set in header parameter
     * @return          LV_RESULT_OK: no error; LV_RESULT_INVALID: can't get the info
     */
-   static lv_result_t decoder_info(struct _lv_image_decoder_t * decoder, const void * src, lv_image_header_t * header)
+   static lv_result_t decoder_info(lv_image_decoder_t * decoder, const void * src, lv_image_header_t * header)
    {
      /*Check whether the type `src` is known by the decoder*/
      if(is_png(src) == false) return LV_RESULT_INVALID;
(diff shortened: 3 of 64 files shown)


XuNeo:
It turns out the micropython is still not working. I'll pause the update here and let's discuss before continuing.

XuNeo:
It turns out the micropython is still not working. I'll pause the update here and let's discuss before continuing.

kisvegabor:
> and it's also reasonable to move lvgl defined types to lv_types.h.

Why is it more reasonable? What problem does it solve?

XuNeo:
> Why is it more reasonable? 

Because `lv_types.h` is used in lvgl to define types and exported to` lvgl.h`? 
It's a central place that can solve header file circular dependency issues explained in https://github.com/lvgl/lvgl/pull/4380.

> What problem does it solve?

Use `lv_xxx_t` in code instead of `struct _lv_xxx_t` has below benefits. The `_` prefix is also meaning "internally use only". 

- Readability: Using typedefs can make the code easier to read and understand. 
- Consistency: It can make the code more uniform and easier to work with.

```
/*Example of exposed API comparison */
void lv_obj_delete(struct _lv_obj_t * obj);
void lv_obj_delete(lv_obj_t * obj);

```

kisvegabor:
I see that `lv_obj_t` feels special because it's used a lot, but if we do it with that, for consistency we needed to do the same with other types as well. E.g. `lv_timer_t`, `lv_display_t`, `lv_anim_t`, which doesn't seem like a god idea.

I think `*_private.h` is a cleaner way. We already do this with `lv_display_private.h` and `lv_indev_private.h` to hide the actual `typedef struct`s from `lvgl.h` and it works with the MicroPython binding too.

I can imagine private headers for all larger modules, such as
- obj 
- timer
- anim
- event
- style
- all widgets?

XuNeo:
> E.g. `lv_timer_t`, `lv_display_t`, `lv_anim_t`, which doesn't seem like a god idea.

Those types are also used by user. And current API exposes struct _lv_timer_t instead of lv_timer_t;
Typedef them in lv_types however can both hide the struct and provide a cleaner way.

We cannot expose _private.h to user.

kisvegabor:
See 
- `struct _lv_display_t {}` in [`lv_display_private.h`](https://github.com/lvgl/lvgl/blob/master/src/display/lv_display_private.h#L33-L151)
- `lv_display_t` in [`lv_display.h`](https://github.com/lvgl/lvgl/blob/master/src/display/lv_display.h#L36)

I mean the same for other modules too. This way the large struct of `lv_display_t` is not exposed to the users and bindings.

XuNeo:
> See
> 
> * `struct _lv_display_t {}` in [`lv_display_private.h`](https://github.com/lvgl/lvgl/blob/master/src/display/lv_display_private.h#L33-L151)
> * `lv_display_t` in [`lv_display.h`](https://github.com/lvgl/lvgl/blob/master/src/display/lv_display.h#L36)

Will it solve the circular dependency issue I mentioned?
For example, currently lv_display.h included lv_draw.h, but lv_draw.h also also need lv_display_t which then need include lv_display.h.

XuNeo:
> It turns out the micropython is still not working. I'll pause the update here and let's discuss before continuing.

The issue is finally fixed in https://github.com/lvgl/lv_binding_micropython/pull/308

kisvegabor:
> Will it solve the circular dependency issue I mentioned?

On it's own not, but adding only 
```c
struct _lv_display_t;
typedef struct _lv_display_t lv_display_t;
```
and similar typedefs looks much better.

XuNeo:
> and similar typedefs looks much better.

Great. I think I can continue now. The main update is only find and replace after all.
The only difference your solution is move below defs to their corresponding header, after all modules are correctly separated to public/private headers.

> struct _lv_display_t;
> typedef struct _lv_display_t lv_display_t;

kisvegabor:
> The only difference your solution is move below defs to their corresponding header, after all modules are correctly separated to public/private headers.

Actually we can move these to `lv_types.h` to avoid circular depencies:
```c
//lv_types.h

struct _lv_display_t;
typedef struct _lv_display_t lv_display_t;

struct _lv_indev_t;
typedef struct _indev_t lv_indev_t;

struct _lv_obj_t;
typedef struct _lv_obj_t lv_obj_t;

...
```

XuNeo:
All changes have been made, and squashed.
Manually run below CI checks and both of them passes locally.

```
python3 lib/lv_bindings/lvgl/tests/micropy_test/__init__.py --artifact-path=lib/lv_bindings/lvgl/tests/micropy_test/artifacts --mpy-path=ports/unix/build-standard/micropython
```

`./lib/lv_bindings/tests/run.sh`.

kisvegabor (approved):

kisvegabor:
The CI is passed, merging. 

I'm happy that the the circular dependency hell is resolved in a so elegant way. Thank you! :blush:

PR #5175, opened 2024-01-04
refact(display): Use draw_buf to replace raw pointers

FASTSHIFT:

Use the draw_buf structure to unify the buffer description in lvgl.

cc @XuNeo 

### Checkpoints

### DIFF
--- a/src/display/lv_display.c
+++ b/src/display/lv_display.c
@@ -393,17 +393,12 @@ void lv_display_set_draw_buffers(lv_display_t * disp, void * buf1, void * buf2, 
     if(disp == NULL) return;
 
     LV_ASSERT_MSG(buf1 == lv_draw_buf_align(buf1, disp->color_format), "buf1 must be aligned");
-
-    uint32_t stride = lv_draw_buf_width_to_stride(disp->hor_res, disp->color_format);
-
-    lv_image_header_init(&disp->buf_1.header, disp->hor_res, disp->ver_res, disp->color_format, stride, 0);
     disp->buf_1.data = buf1;
     disp->buf_1.unaligned_data = buf1;
     disp->buf_1.data_size = buf_size_in_bytes;
 
     if(buf2) {
         LV_ASSERT_MSG(buf2 == lv_draw_buf_align(buf2, disp->color_format), "buf2 must be aligned");
-        lv_image_header_init(&disp->buf_2.header, disp->hor_res, disp->ver_res, disp->color_format, stride, 0);
         disp->buf_2.data = buf2;
         disp->buf_2.unaligned_data = buf2;
         disp->buf_2.data_size = buf_size_in_bytes;
--- a/src/display/lv_display.c
+++ b/src/display/lv_display.c
@@ -392,10 +392,20 @@ void lv_display_set_draw_buffers(lv_display_t * disp, void * buf1, void * buf2, 
     if(disp == NULL) disp = lv_display_get_default();
     if(disp == NULL) return;
 
-    disp->buf_1 = buf1;
-    disp->buf_2 = buf2;
-    disp->buf_act = buf1;
-    disp->buf_size_in_bytes = buf_size_in_bytes;
+    LV_ASSERT_MSG(buf1 == lv_draw_buf_align(buf1, disp->color_format), "buf1 must be aligned");
+    LV_ASSERT_MSG(buf2 == lv_draw_buf_align(buf2, disp->color_format), "buf2 must be aligned");
+
+    lv_image_header_init(&disp->buf_1.header, 0, 0, disp->color_format, 0, 0);
+    disp->buf_1.data = buf1;
+    disp->buf_1.unaligned_data = buf1;
+    disp->buf_1.data_size = buf_size_in_bytes;
+
+    lv_image_header_init(&disp->buf_2.header, 0, 0, disp->color_format, 0, 0);
+    disp->buf_2.data = buf2;
+    disp->buf_2.unaligned_data = buf2;
+    disp->buf_2.data_size = buf_size_in_bytes;
+
+    disp->buf_act = &disp->buf_1;
     disp->render_mode = render_mode;
 }
 
@@ -463,7 +473,7 @@ LV_ATTRIBUTE_FLUSH_READY bool lv_display_flush_is_last(lv_display_t * disp)
 
 bool lv_display_is_double_buffered(lv_display_t * disp)
 {
-    return disp->buf_2 != NULL;
+    return disp->buf_2.data != NULL;
 }
 
 /*---------------------
--- a/src/display/lv_display.c
+++ b/src/display/lv_display.c
@@ -393,17 +393,21 @@ void lv_display_set_draw_buffers(lv_display_t * disp, void * buf1, void * buf2, 
     if(disp == NULL) return;
 
     LV_ASSERT_MSG(buf1 == lv_draw_buf_align(buf1, disp->color_format), "buf1 must be aligned");
-    LV_ASSERT_MSG(buf2 == lv_draw_buf_align(buf2, disp->color_format), "buf2 must be aligned");
 
-    lv_image_header_init(&disp->buf_1.header, 0, 0, disp->color_format, 0, 0);
+    uint32_t stride = lv_draw_buf_width_to_stride(disp->hor_res, disp->color_format);
+
+    lv_image_header_init(&disp->buf_1.header, disp->hor_res, disp->ver_res, disp->color_format, stride, 0);
     disp->buf_1.data = buf1;
     disp->buf_1.unaligned_data = buf1;
     disp->buf_1.data_size = buf_size_in_bytes;
 
-    lv_image_header_init(&disp->buf_2.header, 0, 0, disp->color_format, 0, 0);
-    disp->buf_2.data = buf2;
-    disp->buf_2.unaligned_data = buf2;
-    disp->buf_2.data_size = buf_size_in_bytes;
+    if(buf2) {
+        LV_ASSERT_MSG(buf2 == lv_draw_buf_align(buf2, disp->color_format), "buf2 must be aligned");
+        lv_image_header_init(&disp->buf_2.header, disp->hor_res, disp->ver_res, disp->color_format, stride, 0);
+        disp->buf_2.data = buf2;
+        disp->buf_2.unaligned_data = buf2;
+        disp->buf_2.data_size = buf_size_in_bytes;
+    }
 
     disp->buf_act = &disp->buf_1;
     disp->render_mode = render_mode;
--- a/src/core/lv_refr.c
+++ b/src/core/lv_refr.c
@@ -345,11 +345,11 @@ void _lv_display_refr_timer(lv_timer_t * tmr)
         return;
     }
 
-    if(disp_refr->buf_size_in_bytes == 0) {
+    if(disp_refr->buf_act->data_size == 0) {
         LV_LOG_WARN("disp->buf_size_in_bytes is 0");
         return;
     }
-    if(disp_refr->buf_1 == NULL) {
+    if(disp_refr->buf_1.data == NULL) {
         LV_LOG_WARN("No draw buffer");
         return;
     }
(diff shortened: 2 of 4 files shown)


src/display/lv_display.c
```diff
@@ -392,10 +392,24 @@ void lv_display_set_draw_buffers(lv_display_t * disp, void * buf1, void * buf2,
...
-    disp->buf_act = buf1;
-    disp->buf_size_in_bytes = buf_size_in_bytes;
+    LV_ASSERT_MSG(buf1 == lv_draw_buf_align(buf1, disp->color_format), "buf1 must be aligned");
+
+    uint32_t stride = lv_draw_buf_width_to_stride(disp->hor_res, disp->color_format);
+
+    lv_image_header_init(&disp->buf_1.header, disp->hor_res, disp->ver_res, disp->color_format, stride, 0);
+    disp->buf_1.data = buf1;
+    disp->buf_1.unaligned_data = buf1;
+    disp->buf_1.data_size = buf_size_in_bytes;
```
XuNeo: A good example of to use lv_draw_buf_init API here. User provides raw buffer and initialized to draw_buf, thus user owns data.

kisvegabor:
Since [fixing the forward declaration](https://github.com/lvgl/lv_micropython/commits/feat/multi-instance/) the ESP32 build of MictroPython build fails. 

(We haven't noticed that in the #5161 because I rerun only the failed Unix test.)

cc @kdschlosser

kdschlosser:
I know how to fix that problem. I have had it happen to me. I just have to remember what in the heck caused it.

This was a change in LVGL that caused the issue or was it a change in the binding?

XuNeo:
Let's discuss the build break in https://github.com/lvgl/lvgl/issues/5182, where I have included a temp solution.

kdschlosser:
I don't think your change would have caused this issue. It is a naming problem I have had this issue before and I don't remember what was causing it to happen.

The other problem I have it because of how this whole binding is set up  I m not able to fork the thing to make any changes to it because I already have a fork of micropython which is what this binding is. It won't allow me to have 2 forks of the same thing. It is too much of a headache to get the thing to work properly by adding remotes and all of that stuff. So I am not even able to fix it and submit a PR for the fix.

C47D (approved):

kisvegabor (approved):

kisvegabor:
The build break should be fixed.
@XuNeo, please review this PR.

kdschlosser:
where was it broken at?

It is still failing the tests.

kisvegabor:
I re run the test here manually. Now it uses the updated MicroPython repos. 

On the main page you can see it's passing all tests:
image

kdschlosser:
HEYYYYYY That is what we have been wanting to do to be able to easily identify what is and is not an array. It would be nice instead of defining a structure that only has the specific intention of handling the buffers if it was more generic in design so it could be used in other places where arrays are passed to functions or set as fields in structures.

something along the lines of 


```c
typedef struct {
    uint32_t data_size;     /*Total buf size in bytes*/
    union {
        void * data;
        void * unaligned_data;  /*Unaligned address of `data`, used internally by lvgl*/
    };
} lv_buf_array_t;


typedef lv_buf_array_t lv_uint8_array_t;
typedef lv_buf_array_t lv_char_array_t;
typedef lv_buf_array_t lv_byte_array_t;
typedef lv_buf_array_t lv_uint16_array_t;
typedef lv_buf_array_t lv_uint32_array_t;
```

use the specific typedefs for the expected data type that is to be used in a field or a function argument.

XuNeo (commented):
LGTM, once layer migrates to draw buffer, it can be further optimized.
The `lv_draw_buf_init` API could be used here too.

kisvegabor:
> That is what we have been wanting to do to be able to easily identify what is and is not an array

For this we are using
```
uint8_t * name //pointer
uint8_t name[] //array
```

There might be some places where it is missed, but mostly it should be there.

kdschlosser:
the problem is in a binding there is no way to know if 

```
lv_obj_t *
```

means an array of lv_obj_t objects or a pointer to a single lv_obj_t object.

```
void * some_obj
```

It gets even more complicated with the above because that could be a pointer to a single item or it could be a pointer to an array of items.
doing this

```
uint8_t name[]
```

is invalid in structures. You cannot declare a variable length array like that. In a function call yes but in a structure no. With a structure the array muct be declared as a pointer and there needs to be another field in the structure that tell you what the length of the structure is. There is no API convention for the naming of the fields. If there was we would be able to determine the what is what

as an example
```
typedef struct _lv_draw_image_dsc_t {
    lv_draw_dsc_base_t base;

    const void * src;
    lv_image_header_t header;

    int32_t rotation;
    int32_t scale_x;
    int32_t scale_y;
    int32_t skew_x;
    int32_t skew_y;
    lv_point_t pivot;

    lv_color_t recolor;
    lv_opa_t recolor_opa;

    lv_opa_t opa;
    lv_blend_mode_t blend_mode : 4;

    uint16_t antialias      : 1;
    uint16_t tile           : 1;
    lv_draw_image_sup_t * sup;
} lv_draw_image_dsc_t;
```

what is `const void * src` ??? Is it a pointer so a single object?, is it an array? what is it? There is no field in the structure that could be used to determine if it is an array or not... Or is there? Oh there is bit it is a field of a structure that is within this structure. It's not named in a manner that would make it identifiable as being able to be used to know if `src` is an array or a single object.


This 2 concepts might be one and the same in C code but they are not in Python. This is something that I believe has caused some issues even in the MicroPython binding and is one of the reasons why I am not able to do an in place modification of an array that is within a structure. I have to actually recreate all of the elements in the array and apply them to the the structure. there is no ability to change an array of structures using item assignment. 

Using a wrapper for anything that is supposed to be an array would make it undeniable as to what it is that is being expected. This only needs to exist for public structures or any function calls where the size of the buffer gets passed in the function call. There is no additional memory used in these cases. It is a formality to be able to programmatically know what is being expected.

Even in C code by looking at a header file and seeing this

```
typedef struct {
   _lv_obj_style_t * styles;
} same_structure;
```

you are not able to tell me what the `styles` field is. Is it an array of `_lv_obj_style_t` objects or is it a pointer to a single `_lv_obj_style_t` object. Those are 2 wildly different things when dealing with other programing languages.

In CPython to create an array you have to use the following code

```
(_lv_obj_style_t * array_len)(*array_elements)
```

where as a pointer to a single object is done as follows

```
obj = _lv_obj_style_t()
ctypes.pointer(obj)
```

```
obj = _lv_obj_style_t()
some_function(ctypes.byref(obj)
```

an array that gets fulled by the function.
```
array = (_lv_obj_style_t * array_len)()
some_function(array, array_len)
```

in C code these are almost the the same thing because a pointer is simply the first memory address and by indexing the pointer you are able to add the size to the index number to know what the starting memory address is for the next element if there is one. This always works regardless so if it is not an array and you do that you end up reading memory that is not apart of that object. That leads to undefined behavior, the program may not necessarily crash because of it but there are no protections to stop you from doing that. In higher level languages an array is an array and that is what it is. It cannot be anything other than that. The concept of a pointer doesn't exist you cannot point to memory locations and you cannot do things like this.


```
typedef struct {
    int some_field1;
    int some_field2; 
    int some_field3;
} some_struct;

typedef struct {
    uint16_t some_other_field1;
    uint16_t reserved1;
    uint16_t some_other_field2;
    uint16_t reserved2;
    uint16_t some_other_field3;
    uint16_t reserved3;
} some_other_struct;


some_struct *test = NULL;
// fills struct pointer with data
function(test);

another_function((some_other_struct *) (test)->some_other_field3);
```

This is able to be done because C code is dealing with memory locations and it uses what is in those memory locations exactly how it is.  in C code the type are rules one what to do with data in a memory location. They are not rules that say the data in a memory location can only be a certain type. setting a type for something does not mean that the underlying data is etched in stone to be that specific type. It is only 1's and 0's in memory. Kind of like putting on a set of eyeglasses. It changes how the data is viewed but it doesn't define that the data in a memory location can only be one specific type.

if I have an array that is defined like this.
```
uint8_t array1[2][] = { {0xFF, 0x00}, {0xFF, 0x00}, {0xFF, 0x00}, {0xFF, 0x00}};
uint16_t array2[] = {0xFF00, 0xFF00, 0xFF00, 0xFF00 }
```
Those are technically speaking the exact same thing. in memory they both look like this.
```
1111 1111 0000 0000 1111 1111 0000 0000 1111 1111 0000 0000 1111 1111 0000 0000
```
using a higher level language we have no way of dealing with that. An int in python is all of these types in C uin8_t, uint16_t, uint32_t, uint64_t, int8_t, int16_t, int32_t, int64_t. In Python there are ways to access specific memory locations but ONLY if you know exactly what the data is and what type is to be applied to it and in the case of an array I need to know explicitly what the length of the array is. If there is no way to know what the size is then there is no way to handle picking the data up to be used. Remember the size issue i mentioned before when making structures private that are used in public functions. The start and stop positions need to be known in order to deal with pointers. I don't need to know what the data actually is I just need to know where the start and stop points are and that is determined by knowing the size. end point = start point + size. by making a structure private so it is not able to be seen means there is no way to calculate the size of the structure from reading a header file. Can you tell me what the size of `lv_display_t` is without opening up the private header file? Remember we are only able to "visually" inspect the header files. The code is not being compiled or ran. without being able to visually see it there is no way to know what the size is.

kisvegabor:
Clear, thanks. 

If we are talking only about public struct fields, instead of having `<type_name>_array_t` adding an `_array` postfix the the variable name (not type name) seems more reasonable to me.

E.g.
```c
typedef struct {
   uint32_t a;
   void  * b; //Pointer to a single element
   lv_obj_t * obj_array; //Array of widgets
}lv_xyz_t;
˙```

The problem is that 2 techniques used the tell the array size:
1. Just store it next to the array (we can have convention for that)
2. The last element of the array is a trailing item with some special value (E.g. `NULL`).

To better understand the scope of the problem it would be great to see what the public `struct`s are now where are having arrays. Can you list them and a simple way?

**Comment by kdschlosser** [comment]
> To better understand the scope of the problem it would be great to see what the public structs are now where are having arrays. > Can you list them and a simple way?

If I could we wouldn't be having this discussion ... LMAO

**Comment by kisvegabor** [comment]
I think these are independent: knowing how to fix vs how many occurrences are there. 

Anyway, now most of the main `struct` definitions  were moved to `lv_types.h` so we can move the actual `typedef`s to `lv_*_private.h` files . However `lv_draw_label_dsc_t`, `lv_draw_rect_dsc_t`, and similar structs are still public and we don't have set/get API for.

**Comment by kdschlosser** [comment]
question about the forward declarations. Since that is being done is it going to take care of this??...


```
void some_function(struct _lv_obj_t * obj);
```

Is that going to end up becoming

```
void some_function(_lv_obj_t * obj);
```

????

That would make it a whole lot easier when reading the code for a binding and it will also make it easier when generating that JSON. It is one of the animals I am dealing with right now with the JSON output.


The other thing is how are the forward declarations going to look? are they like this?

```
struct _lv_obj_t;
typedef struct _lv_obj_t lv_obj_t;
```

so the need to have any function using the `_lv_obj_t` type would not need to be done and it is simply for the purposes of being able to create the correct types a head of time.

I think it would be wise to do this with all structures regardless of them being private or not. Then there is no need to even think about if a structure has been declared or not when making changes to LVGL. The only structures/unions you would not concern yourself with doing this are ones that are truly private. and they are not used in any structure or union that is used in a public function.

**Comment by kisvegabor** [comment]
> question about the forward declarations. Since that is being done is it going to take care of this??...

It will look like this `void some_function(lv_obj_t * obj);`

Note that
```c
struct _lv_obj_t;
typedef struct _lv_obj_t lv_obj_t;
```

Is useful if the type `lv_obj_t` in this case is used *only* as a pointer. To access the fields we need the full declaration. And I think right now the users are accessing only the `lv_draw_...dsc_t` structs by field.

kdschlosser:
OK so functions will no longer have any types in it that begin with an underscore. and same goes for any structures that need to be directly accessed by the user.

would there be forward declarations for structures that even have fields in them that the user does access? I ask this because if a public structure needs to reference itself.

```
struct _some_struct {
    struct _some_struct * field;
};
```

will turn into

```
struct _some_struct {
    some_struct * field;
};
```

is that correct? This would be a huge help when reading the header files in order to create the JSON API data.

kisvegabor:
> would there be forward declarations for structures that even have fields in them that the user does access? I ask this because if a public structure needs to reference itself.

Seems doable however it's quite hard to find all occurrences by hand. Can the script output the affected structs?

Issue #5090, opened 2023-12-24
Setting LV_DRAW_BUF_STRIDE_ALIGN to a value rather than 1 will cause display glitches


GorgonMeducer:
### Perform all steps below and tick them with [x]
- [x] Read the [FAQ](https://docs.lvgl.io/master/intro/index.html#faq)
- [x] Check the related part of the [Documentation](https://docs.lvgl.io/)
- [x] Update lvgl to the latest version
- [x] Reproduce the issue in a [Simulator](https://docs.lvgl.io/latest/en/html/get-started/pc-simulator.html)

### Describe the bug
I do not fully understand the purpose of the option `LV_DRAW_BUF_STRIDE_ALIGN`, but when I set it to `4`, there are obvious display glitches.

![image](https://github.com/lvgl/lvgl/assets/13148491/e2dd9786-7e9a-4c97-bbd1-1b10da4c2482)

I tried fixing it in #5091, but it doesn't work.

XuNeo:
`LV_DRAW_BUF_STRIDE_ALIGN` determines the bytes alignment of draw buffer stride.
E.g. if `64` is used, for a draw buffer with 100x100, ARGB8888, stride will be `ALIGN_UP((100*4), 64)` = 512.

For CPU draw unit and GPU that has no such requirement, `LV_DRAW_BUF_STRIDE_ALIGN` can be configured to 1.

GorgonMeducer:
The problem is, 4 is a valid opinion but it causes glitches.

XuNeo:
Could you show the steps to reproduce it?

GorgonMeducer:
Set the macro I mentioned to 4 and run benchmark test. 

If you turn on performance monitor, you will saw similar glitches very often.

kisvegabor:
It works well in the [Eclipse simpualtor](https://github.com/lvgl/lv_port_pc_eclipse) so I assume that it's a driver issue.
Do you consider the stride align = 4 in the flush function too?

GorgonMeducer:
@kisvegabor, The driver supports aligned access. As long as there is no bus fault or hardfault, even an unaligned address is used, the data would be written to the correct address.

GorgonMeducer:
Happy new year!

kisvegabor:
Happy new year :slightly_smiling_face: 

[...]

PR #5176, opened 2024-01-04
chore(porting): Add code helper to clip the display buffer

GorgonMeducer:

To follow the discussion in #5090 

### Checkpoints

### DIFF
--- a/examples/porting/lv_port_disp_template.c
+++ b/examples/porting/lv_port_disp_template.c
@@ -116,28 +116,28 @@ static void disp_flush(lv_display_t * disp_drv, const lv_area_t * area, uint8_t 
 {
     if(disp_flush_enabled) {
 
-    /*
-     * If you meet ALL the criteria below, please set the value 0 to 1 to enable the code segment.
-     *     - If your display device, e.g. a lot of SPI LCD, supports the Window Mode, and
-     *     - You want to use the window mode together with LV_DISPLAY_RENDER_MODE_PARTIAL
-     *     - The macro LV_DRAW_BUF_STRIDE_ALIGN is NOT 1
-     */
-    #if 0
-        /* clip the frame buffer(pointed by px_map) with the width calculated from the area, as 
+        /*
+         * If you meet ALL the criteria below, please set the value 0 to 1 to enable the code segment.
+         *     - If your display device, e.g. a lot of SPI LCD, supports the Window Mode, and
+         *     - You want to use the window mode together with LV_DISPLAY_RENDER_MODE_PARTIAL
+         *     - The macro LV_DRAW_BUF_STRIDE_ALIGN is NOT 1
+         */
+#if 0
+        /* clip the frame buffer(pointed by px_map) with the width calculated from the area, as
          * the buffer stride might be larger than width. */
         do {
             lv_color_format_t color_format = lv_display_get_color_format(disp_drv);
             int32_t width = area->x2 - area->x1 + 1;
             int32_t height = area->y2 - area->y1 + 1;
-            int32_t stride = lv_draw_buf_width_to_stride( width, color_format);
+            int32_t stride = lv_draw_buf_width_to_stride(width, color_format);
             width *= lv_color_format_get_bpp(color_format) >> 3;
 
             if(width == stride) {
                 break;
             }
 
-            uint8_t *src_ptr = px_map;
-            uint8_t *des_ptr = px_map;
+            uint8_t * src_ptr = px_map;
+            uint8_t * des_ptr = px_map;
 
             for(int32_t y = 0; y < height; y++) {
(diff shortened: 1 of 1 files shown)


kisvegabor:
@XuNeo Can't we use/have a `lv_draw_buf_` to readjust the stride of a buffer in place?

If not I suggest having a helper in `lv_draw_sw.c` similar to `lv_draw_sw_rgb565_swap()`.

GorgonMeducer:
> If not I suggest having a helper in lv_draw_sw.c similar to lv_draw_sw_rgb565_swap().

Adding a helper is also a good idea.

GorgonMeducer:
@kisvegabor @XuNeo 
Any updates about this?

XuNeo:
> @XuNeo Can't we use/have a `lv_draw_buf_` to readjust the stride of a buffer in place?

Current `lv_draw_buf_adjust_stride` will create a new draw buf, we can change it to only adjust in place and return error if not possible to do so. The caller then need to decide to create a new draw buf and do draw buffer copy instead(with expected stride set).

XuNeo:
@kisvegabor I think flush_cb should be update too to use draw_buf instead of pixel_map, which also affects porting code.

After that, https://github.com/lvgl/lv_binding_micropython/pull/311 also needs attention.

kisvegabor:
> I think flush_cb should be update too to use draw_buf instead of pixel_map, which also affects porting code.

I'm not entirely sure about it. I think the simplicity of the `flush_cb` is a great value.  

In the rare cases when advanced users need to take care of stride, they can get it from the width of the area. The color format stored in the display. But it"s required only for the 1% of the users (I guess) and wouldn't make the API more complicated for the 99%.

GorgonMeducer:
> But it"s required only for the 1% of the users (I guess) and wouldn't make the API more complicated for the 99%.

When people use GPUs that have alignment requirements and LCDs with window mode, we have to handle this.
So, how many devices would have GPU and LCD with window mode at same time? It is definitely more than 1%. I would bet at least 33%.

GorgonMeducer:
> True, however if there is a GPU probably there is an LCD controller periphery too. And we can integrate it into LVGL. (Actually it's part of LVGL's deal with the vendors that we will add their LCD controllers).

That's why I say 33%.  LTDC requires a lot of pins, but LCDs with window mode are usually very cheap (so many people would use them) and only need a few pins for SPI. 

I just need a solution for this scenario.  Should we add a helper? (similar to rgb565_swap)

XuNeo:
`flush_cb(draw_buf, area)` won't make things more complicated. They can deal with it as usual simply by `pixel_map = draw_buf->data` because the stride must be the simpler case.

Otherwise, for partial update mode with GPU support, `draw_buf` is a must.

kisvegabor:
> flush_cb(draw_buf, area) won't make things more complicated. They can deal with it as usual simply by pixel_map = draw_buf->data because the stride must be the simpler case.

I feel this redundant as both `draw_buf` and `area` contains the width and height.

> Otherwise, for partial update mode with GPU support, draw_buf is a must.

Can't we convert the width of the area to stride?

GorgonMeducer:
> Can't we convert the width of the area to stride?

LCD driver needs the width but not stride to set the window. We also know that stride can be bigger than width * pixel size

kisvegabor:
I mean:
```c
int32_t w = lv_area_get_width(area);
int32_t stride = lv_draw_buf_width_to_stride(w, lv_display_get_color_format(display))
```

GorgonMeducer:
> However I don't thinks this should be done before calling flush_cb. Better leaving it to driver to make decision.

No problem. I also think it is the developers' work to use the helper in their driver.

kisvegabor:
The questions is what to do with the parameters of `flush_cb`. I'll comment about it in https://github.com/lvgl/lvgl/issues/5273

XuNeo:
> Will you introduce this helper, i.e. `lv_draw_buf_adjust_stride_in_place ` ?

I plan to modify the behavior of `lv_draw_buf_adjust_stride` to do in place. Create a new draw buffer secretly in this API is not good. Better for user to manually create the new buffer when stride adjust fails.

@kisvegabor Hi Gabor, do you agree?

kisvegabor:
> I plan to modify the behavior of lv_draw_buf_adjust_stride to do in place. Create a new draw buffer secretly in this API is not good. Better for user to manually create the new buffer when stride adjust fails.

I perfectly agree.

GorgonMeducer:
Do we have any update for this? Any helper? solution ?
@XuNeo @kisvegabor

kisvegabor:
I consider `lv_draw_buf_adjust_stride` an internal API, so we can added after the release.

GorgonMeducer:
OK, no problem.

PR #5177, opened 2024-01-04
chore(draw_buf): add invalid cf and stride check

XuNeo:

If cf and stride is invalid value 0, use default cf for reshape, and use global stride instead of 0. Add error log when malloc failed.

### Checkpoints

### DIFF
--- a/src/draw/lv_draw_buf.c
+++ b/src/draw/lv_draw_buf.c
@@ -147,6 +147,8 @@ lv_draw_buf_t * lv_draw_buf_create(uint32_t w, uint32_t h, lv_color_format_t cf,
     void * buf = lv_draw_buf_malloc(size, cf);
     LV_ASSERT_MALLOC(buf);
     if(buf == NULL) {
+        LV_LOG_WARN("No memory: %"LV_PRIu32"x%"LV_PRIu32", cf: %d, stride: %"LV_PRIu32", %"LV_PRIu32"Byte, ",
+                    w, h, cf, stride, size);
         lv_free(draw_buf);
         return NULL;
     }
@@ -156,6 +158,7 @@ lv_draw_buf_t * lv_draw_buf_create(uint32_t w, uint32_t h, lv_color_format_t cf,
     draw_buf->header.cf = cf;
     draw_buf->header.flags = LV_IMAGE_FLAGS_MODIFIABLE | LV_IMAGE_FLAGS_ALLOCATED;
     draw_buf->header.stride = stride;
+    draw_buf->header.magic = LV_IMAGE_HEADER_MAGIC;
     draw_buf->data = lv_draw_buf_align(buf, cf);
     draw_buf->unaligned_data = buf;
     draw_buf->data_size = size;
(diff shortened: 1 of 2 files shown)


src/draw/lv_draw_buf.c
```diff
@@ -147,6 +147,8 @@ lv_draw_buf_t * lv_draw_buf_create(uint32_t w, uint32_t h, lv_color_format_t cf,
     void * buf = lv_draw_buf_malloc(size, cf);
     LV_ASSERT_MALLOC(buf);
     if(buf == NULL) {
+        LV_LOG_ERROR("No memory: %"LV_PRIu32"x%"LV_PRIu32", cf: %d, stride: %"LV_PRIu32", %"LV_PRIu32"Byte, ",
```
kisvegabor: It's not necessarily a warning. LVGL can try malloc-ing next time. 
Let's make is a warning or info.

XuNeo: I have changed to warning for now. Let's get back to it when standalone image cache is added, which will give us the ability to try mallocing on failed.

src/draw/lv_draw_buf.c
```diff
@@ -147,6 +147,8 @@ lv_draw_buf_t * lv_draw_buf_create(uint32_t w, uint32_t h, lv_color_format_t cf,
     void * buf = lv_draw_buf_malloc(size, cf);
     LV_ASSERT_MALLOC(buf);
     if(buf == NULL) {
+        LV_LOG_WARN("No memory: %"LV_PRIu32"x%"LV_PRIu32", cf: %d, stride: %"LV_PRIu32", %"LV_PRIu32"Byte, ",
+                     w, h, cf, stride, size);
         lv_free(draw_buf);
```
PGNetHun, line 152: `lv_draw_buf_free` should be used instead of `lv_free`

src/draw/lv_draw_buf.c
```diff
@@ -184,6 +187,8 @@ lv_draw_buf_t * lv_draw_buf_reshape(lv_draw_buf_t * draw_buf, lv_color_format_t
 {
     if(draw_buf == NULL) return NULL;
 
+    /*If color format is invalid, keep using the original format.*/
```
PGNetHun: Correct text:

> If color format is not defined or unknown, keep using the original color format.

... but I think we could also omit the comment.

PGNetHun (changes requested):

src/draw/lv_draw_buf.c
XuNeo: lv_free is corresponding to `lv_malloc_zeroed ` in line 140. It's the struct itself rather than the buffer.

src/draw/lv_draw_buf.c
XuNeo: LVGL never used `cf == 0` , so I tend to comment here to explain the special usage. I'll update the comment. Thank you!

PGNetHun (approved):

kisvegabor (approved):

PR #5190, opened 2024-01-04
fix declspecs usage in msvc

ghost:
This PR https://github.com/lvgl/lvgl/pull/4704 tryd to address the issue of the missing decl specs.
However there is an error there.

When compiling lvgl you need to use dllimport. When compiling (external) cmake targets you need to use dllexport.

BUILD_SHARED_LIBS is defined when building dlls. So LV_ATTRIBUTE_EXTERN_DATA will allways point to __declspec\(dllimport\).

I cloned the latest master and copied the lv_conf.h to the root directory. Then I open VS 2022 and build everything. It seems to work now with VS 2022.

Thx guys for your reviews.


CMakeLists.txt
```diff
@@ -19,9 +19,8 @@ else()
...
-  if (BUILD_SHARED_LIBS)
-    target_compile_definitions(lvgl PUBLIC LV_ATTRIBUTE_EXTERN_DATA=__declspec\(dllimport\))
-  else()
-    target_compile_definitions(lvgl PUBLIC LV_ATTRIBUTE_EXTERN_DATA=__declspec\(dllexport\))
-  endif()
-endif()
\ No newline at end of file
+  target_compile_definitions(lvgl
+    PRIVATE LV_ATTRIBUTE_EXTERN_DATA=__declspec\(dllimport\)
+    PUBLIC LV_ATTRIBUTE_EXTERN_DATA=__declspec\(dllexport\)
```
kisvegabor: Using both `PRIVATE` and `PUBLIC` is not that intuitive. Please add some comments to explain what a happens here and why we do this.

kisvegabor (commented):
Thank you! 

@kdschlosser Please, could you review and test the changes?

CMakeLists.txt
ghost: @kisvegabor  I added some explanations. I hope this makes it clearer.

kisvegabor (dismissed):
Looks good to me. Let's wait for @kdschlosser's feedback.

FASTSHIFT (dismissed):

kdschlosser:
with MSVC when compiling a shared library the _DLL macro will be set and when static it will not be. That is your way of knowing what compilations is being done without needing to add any kind of special handling for it. 

I really believe this specific thing needs to be in the source for LVGL and not in the cmake files. It should be in lv_types.h file. modifying how things get compiled using macros set in the compiler options at the command line is one thing. declaring types that way... ehhhh..... i don't really care for doing that. Firstly because an IDE is not going to be able to follow the data path. it is going to come to a dead end to a macro that looks as if it is unused. someone making a modification is not going to know what is happening. I call that voodoo magic code. Just because it can be done doesn't mean it should be done. 

It is very easy to check and see if the compiler is MSVC and then checking the _DLL macro from a source or header file that is getting compiled.

kdschlosser:
see here.

https://github.com/kdschlosser/lvgl/blob/cb838ddfc563cc1812dfc9cbb780a1c6e3bd09f4/src/misc/lv_types.h#L39

Ju1He1:
> see here.
> 
> https://github.com/kdschlosser/lvgl/blob/cb838ddfc563cc1812dfc9cbb780a1c6e3bd09f4/src/misc/lv_types.h#L39


Hey your solution can not work because there is an error. It is not `_MSVC_VER `but `MSC_VER`.
After fixing this i tried to compile the latest lvgl master with the lv_conf taken from here (https://github.com/lvgl/lv_port_pc_visual_studio) and I'm getting lots of inconsistent dll linker errors.
Did you try building your solution on windows?

I'm open to any solution which works on windows :)

Ju1He1:
> that was just an example. are you using cmake to compile it? because you would still need to use cmake otherwise none of the functions will get exported properly.

I'm using latest VS 2022 (with Cmake) to build lvgl ;)
I checked your solution. According to the docs it should work (not sure why it's not working for me). However keep in mind that the _DLL Makro is only set when building shared libs. If (for whatever reason) someone is linking lvgl statically on windows your solution will no longer work.

@kisvegabor  @kdschlosser  I pushed some changes, and replaced PUBLIC with INTERFACE to improve differentiation between import and export. For me I don't care if we are using my solution or kdschlosser's solution. 
I only want to make it work on windows.
Your turn to decide ;)

kisvegabor:
@Ju1He1
I suggest trying to add 
```c
#ifdef MSVC_VER
    #ifdef _DLL
        #define LV_ATTRIBUTE_EXTERN_DATA __declspec(dllimport)
    #else
        #define LV_ATTRIBUTE_EXTERN_DATA __declspec(dllexport)
    #endif
#else
    #define LV_ATTRIBUTE_EXTERN_DATA
#endif
```

to `lv_conf.h` instead of `lv_types.h` to be sure all parts of LVGL sees it.

Ju1He1:
> ```c
> #ifdef MSVC_VER
>     #ifdef _DLL
>         #define LV_ATTRIBUTE_EXTERN_DATA __declspec(dllimport)
>     #else
>         #define LV_ATTRIBUTE_EXTERN_DATA __declspec(dllexport)
>     #endif
> #else
>     #define LV_ATTRIBUTE_EXTERN_DATA
> #endif
> ```

thx for your reply. There are some errors in that code
I added this to lv_conf.h and commented out my Cmake Stuff
```
#ifdef _MSC_VER
  #ifdef _DLL
    #define LV_ATTRIBUTE_EXTERN_DATA __declspec(dllexport)
  #else
    #define LV_ATTRIBUTE_EXTERN_DATA __declspec(dllimport)
  #endif
  #else
  #define LV_ATTRIBUTE_EXTERN_DATA
#endif //_MSC_VER
```

Unfortunately this does not work. I'm getting lots of "unresolved symbol errors".

I did some digging.
The examples are actually libraries. So _DLL is set. This is why symbols are not getting imported properly. This probably results in the linker errors

Also another possible issue:
According to the MS docs
"_DLL Defined as 1 when the /MD or /MDd (Multithreaded DLL) compiler option is set. Otherwise, undefined."
=> CMake appends these flags implicitly, I'm not so sure about other build systems. 


For me it looks like we can only get this to work if we are adding these options to cmake (like in the PR). Or am I missing something?

kisvegabor:
I agree with @kdschlosser to make it work in a compiler independent way, however if it's too hacky, we can go with the simper solution which already makes a lot of people happy. :slightly_smiling_face: 

So I merge tis PR now. If there will be any better ideas we can improve it at any time.

Thank you!

kisvegabor (approved):

PR #5154, opened 2024-01-02
feat(freetype): merge freetype outline and image into one context

W-Mai:
### Description of the feature or fix

Merge freetype outline and image into one context.

Now you can create bitmap based freetype font and vector graph based outline freetype font at same time by using these flag:

```cpp
enum {
    LV_FREETYPE_FONT_RENDER_MODE_BITMAP = 0,
    LV_FREETYPE_FONT_RENDER_MODE_OUTLINE = 1,
};
```cpp
// create a bitmap based freetype font
lv_freetype_font_create("./res/MiSans-Regular.ttf", LV_FREETYPE_FONT_RENDER_MODE_BITMAP,
                                                24,
                                                LV_FREETYPE_FONT_STYLE_ITALIC);

// create a vector graph freetype font
lv_freetype_font_create("./res/MiSans-Regular.ttf", LV_FREETYPE_FONT_RENDER_MODE_OUTLINE,
                                                24,
                                                LV_FREETYPE_FONT_STYLE_NORMAL);
```

snapshot:

image

## Note:
If your environment doesn't support vector drawing, you will get a result like this
image

### Checkpoints
- [x] Run `code-format.py` from the scripts folder. [astyle](http://astyle.sourceforge.net/install.html) needs to be installed.
- [ ] Update the [Documentation](https://github.com/lvgl/lvgl/tree/master/docs) if needed
- [ ] Add [Examples](https://github.com/lvgl/lvgl/tree/master/examples) if relevant. 
- [ ] Add [Tests](https://github.com/lvgl/lvgl/blob/master/tests/README.md) if applicable.
- [ ] If you added new options to `lv_conf_template.h` run [lv_conf_internal_gen.py](https://github.com/lvgl/lvgl/blob/release/v8.3/scripts/lv_conf_internal_gen.py) and update [Kconfig](https://github.com/lvgl/lvgl/blob/release/v8.3/Kconfig).
 
Be sure the following conventions are followed:
- [ ] Follow the [Styling guide](https://github.com/lvgl/lvgl/blob/master/docs/CODING_STYLE.md)
- [ ] Prefer `enum`s instead of macros. If inevitable to use `define`s export them with `LV_EXPORT_CONST_INT(defined_value)` right after the `define`.
- [ ] In function arguments prefer `type name[]` declaration for array parameters instead of `type * name`
- [ ] Use typed pointers instead of `void *` pointers
[...]

PR #5196, opened 2024-01-05
feat(freetype): add new freetype testcase

W-Mai:

add new freetype testcase

freetype1

### Checkpoints


W-Mai:
this pr is depended on https://github.com/lvgl/lvgl/pull/5154
before merge this pr, please merge https://github.com/lvgl/lvgl/pull/5154 first.
and I will rebase this pr to latest code.

W-Mai:
😮‍💨 The cmake version provided by the `run-on-arch-action` is a little old. 
Wait for me to downgrade some cmake functions.

W-Mai:
I have no idea about this ci fail.

PGNetHun:
Because you change also the code, not just adding new tests, please update PR title to: `feat(freetype)`

PGNetHun:
> I have no idea about this ci fail.

It must be related to the newly added **libasan6** library.

Build error:
```
/usr/bin/ld: cannot find liblsan_preinit.o: No such file or directory
  /usr/bin/ld: cannot find -llsan
```
Google first result: https://github.com/google/sanitizers/issues/695

kisvegabor:
Similar but independent topic in TinyTTF and `lv_binfont_load` I have changed the `create` API like this:
```c
lv_result_t lv_tiny_ttf_create_file(lv_font_t * font, const char * path, int32_t font_size);
```

This way if a font is used in many styles, it can be easily reinitialized with different size or other settings without updating the LVGL styles. If a new font is allocated in the `create` function all styles needs to be updated manually. Therefore I recommend updating the API of FreeType like this as well.

W-Mai:
> > I have no idea about this ci fail.
> 
> It must be related to the newly added **libasan6** library.
> 
> Build error:
> 
> ```
> /usr/bin/ld: cannot find liblsan_preinit.o: No such file or directory
>   /usr/bin/ld: cannot find -llsan
> ```
> 
> Google first result: [google/sanitizers#695](https://github.com/google/sanitizers/issues/695)

感谢你的回复，不过这些方法我都尝试过了，现在的代码已经是修改过的了，但是依然没有通过CI。
我好像有了一些其他的头绪，正在尝试解决这个可恶的问题。

Thank you for your reply😁, but I have tried all these methods. The current code has been modified, but it still has not passed CI.

I seem to have some other clues and I'm trying to solve this fxxxxxking problem.😇

W-Mai:
It seems that there are a little bit different in freetype from x86 version and arm version

x86 snapshot:
freetype1

armv7 snapshot:
freetype1_err

black point is the difference:
image

W-Mai:
I found that if I only link -fsanitize=leak, the link will fail, but if I link -fsanitize=address,leak at the same time, it will work fine.

W-Mai:
> It seems that there are a little bit different in freetype from x86 version and arm version

In the end, I found the answer. The accuracy of freetype at 32-bit and 64-bit is also different, so the drawing results are different, so I kept the different drawing results under 32-bit and 64-bit.

W-Mai:
> Similar but independent topic in TinyTTF and `lv_binfont_load` I have changed the `create` API like this:
> 
> ```c
> lv_result_t lv_tiny_ttf_create_file(lv_font_t * font, const char * path, int32_t font_size);
> ```
> 
> This way if a font is used in many styles, it can be easily reinitialized with different size or other settings without updating the LVGL styles. If a new font is allocated in the `create` function all styles needs to be updated manually. Therefore I recommend updating the API of FreeType like this as well.

Let me see, what you're saying is that after you've created the font outside, it's better to pass it in to create font?

Both @FASTSHIFT and I think it would be a better choice to return a `lv_font_t *`.

1. Passing in a font, the incoming font may be constructed on the stack, may be from malloc, or global, and the life cycle is not easy to control. If it is a font on the stack, it may cause unnecessary memory leakage.

2. Return a font that makes it easier to manage the life cycle of the created font.

3. From the name, create is more like generating a new object from the inside, similar to lv_obj_create.

So we both think return a `lv_font_t *` is a more better way.

What do you think of this point?

kisvegabor:
You can still do this:
```c
lv_font_t * f1 = lv_malloc(sizeof(lv_font_t));
lv_freetype_create(f1, ...);

//or
static lv_font_t f1; //or global
lv_freetype_create(&f2, ...);
```

So you have more control and flexibility. 

However I do agree with your comment on naming. `lv_freetype_init_...(lv_font_t * font, ...)` might be a batter name.

What about having both like this:
```c
lv_freetype_init_font_from_file(lv_font_t * font, ...);

//Helper
lv_font_t * lv_freetype_create_font_from_file(...)
{
  lv_font_t * f = lv_malloc(sizeof(lv_font_t));
  lv_freetype_init_font_from_file(f, ...);
  return f;
}

void lv_freetype_delete_font(lv_font_t * f)
{
   //freetype specific destroy
   lv_free(f)
}
```

W-Mai:
> You can still do this:
> 
> ```c
> lv_font_t * f1 = lv_malloc(sizeof(lv_font_t));
> lv_freetype_create(f1, ...);
> 
> //or
> static lv_font_t f1; //or global
> lv_freetype_create(&f2, ...);
> ```
> 
> So you have more control and flexibility.
> 
> However I do agree with your comment on naming. `lv_freetype_init_...(lv_font_t * font, ...)` might be a batter name.
> 
> What about having both like this:
> 
> ```c
> lv_freetype_init_font_from_file(lv_font_t * font, ...);
> 
> //Helper
> lv_font_t * lv_freetype_create_font_from_file(...)
> {
>   lv_font_t * f = lv_malloc(sizeof(lv_font_t));
>   lv_freetype_init_font_from_file(f, ...);
>   return f;
> }
> 
> void lv_freetype_delete_font(lv_font_t * f)
> {
>    //freetype specific destroy
>    lv_free(f)
> }
> ```

Good idea, very feasible!
Use init and create to distinguish between the two, and wrap them with helper functions to make the operation more flexible.

We can reconstruct our functions later with this in mind.

PR #5204, opened 2024-01-05
refactor(display): use draw_buf_t for display

XuNeo:

Change parameter of `lv_display_set_draw_buffers ` completely to `lv_draw_buf_t`.

Usage example:
```c
        if(dsc->buf1) lv_draw_buf_destroy(dsc->buf1);
        dsc->buf1 = lv_draw_buf_create(hor_res, ver_res, cf, 0);
        LV_ASSERT_NULL(dsc->buf1);

#if LV_SDL_BUF_COUNT == 2
        if(dsc->buf2) lv_draw_buf_destroy(dsc->buf2);
        dsc->buf2 = lv_draw_buf_create(hor_res, ver_res, cf, 0);
        LV_ASSERT_NULL(dsc->buf2);
#endif
        lv_display_set_draw_buffers(disp, dsc->buf1, dsc->buf2);
```

Or
```c
    static lv_color32_t test_fb[(HOR_RES + LV_DRAW_BUF_STRIDE_ALIGN - 1) * VER_RES + LV_DRAW_BUF_ALIGN];
    lv_display_t * disp = lv_display_create(HOR_RES, VER_RES);
    static lv_draw_buf_t draw_buf;
    lv_draw_buf_init(&draw_buf, HOR_RES, VER_RES, LV_COLOR_FORMAT_ARGB8888, 0, test_fb, sizeof(test_fb));
    lv_display_set_draw_buffers(disp, &draw_buf, NULL);
```
### Checkpoints

### DIFF
--- a/src/display/lv_display.h
+++ b/src/display/lv_display.h
@@ -231,10 +231,15 @@ int32_t lv_display_get_dpi(const lv_display_t * disp);
  * @param disp              pointer to a display
  * @param buf1              first buffer
  * @param buf2              second buffer (can be `NULL`)
+ */
+void lv_display_set_draw_buffers(lv_display_t * disp, lv_draw_buf_t * buf1, lv_draw_buf_t * buf2);
+
+/**
+ * Set display render mode
+ * @param disp              pointer to a display
  * @param render_mode       LV_DISPLAY_RENDER_MODE_PARTIAL/DIRECT/FULL
  */
-void lv_display_set_draw_buffers(lv_display_t * disp, void * buf1, void * buf2, uint32_t buf_size_in_bytes,
-                                 lv_display_render_mode_t render_mode);
+void lv_display_set_render_mode(lv_display_t * disp, lv_display_render_mode_t render_mode);
 
 /**
  * Set the flush callback which will be called to copy the rendered image to the display.
--- a/docs/CHANGELOG.rst
+++ b/docs/CHANGELOG.rst
@@ -22,7 +22,8 @@ If you are updating a v8 project to v9, special care must be taken as some parts
 
 So pay extra attention to these:
 
-- :cpp:func:`lv_display_set_draw_buffers(display, buf1, buf2, buf_size_byte, mode)` is more or less the equivalent of ``lv_disp_draw_buf_init(&draw_buf_dsc, buf1, buf2, buf_size_px)`` from v8, however in **v9 the buffer size is set in bytes**.
+- :cpp:func:`lv_display_set_draw_buffers(display, buf1, buf2)` is more or less the equivalent of ``lv_disp_draw_buf_init(&draw_buf_dsc, buf1, buf2, buf_size_px)`` from v8, however in **v9 the buffer type is lv_draw_buf_t**.
+- Make sure to call :cpp:func:`lv_display_set_render_mode(display, mode)` to set correct render mode.
 - In v9 ``lv_color_t`` is always RGB888 regardless of ``LV_COLOR_DEPTH``.
 - ``lv_conf.h`` has been changed a lot, so don't forget to update it from ``lv_conf_template.h``
 - Be sure ``<stdint.h>`` is **not** included in ``lv_conf.h``. In v9 we have some assembly parts for even better performance and a random include there would mess up the assembly part.
@@ -78,9 +79,10 @@ Display API
 
    lv_display_t * disp = lv_display_create(hor_res, ver_res)
    lv_display_set_flush_cb(disp, flush_cb);
-   lv_display_set_draw_buffers(disp, buf1, buf2, buf_size_in_bytes, mode);
+   lv_display_set_draw_buffers(disp, buf1, buf2);
+   lv_display_set_render_mode(disp, mode);
 
--  Note that now **buf size is in bytes and not pixels**
+-  Note that now **draw buffer can be malloced by `lv_draw_buf_create` or initialized by `lv_draw_buf_init`**
 -  ``mode`` can be:
 
    -  ``LV_DISPLAY_RENDER_MODE_PARTIAL`` This way the buffers can be
--- a/src/core/lv_refr.c
+++ b/src/core/lv_refr.c
@@ -345,7 +345,8 @@ void _lv_display_refr_timer(lv_timer_t * tmr)
         return;
     }
 
-    if(disp_refr->buf_act == NULL || disp_refr->buf_act->data == NULL || disp_refr->buf_act->data_size == 0) {
+    lv_draw_buf_t * buf_act = disp_refr->buf_act;
+    if(!(buf_act && buf_act->data && buf_act->data_size)) {
         LV_LOG_WARN("No draw buffer");
         return;
     }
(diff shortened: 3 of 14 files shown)


src/display/lv_display.h
```diff
@@ -231,10 +231,15 @@ int32_t lv_display_get_dpi(const lv_display_t * disp);
...
+void lv_display_set_draw_buffers(lv_display_t * disp, lv_draw_buf_t * buf1, lv_draw_buf_t * buf2);
+
+/**
+ * Set display render mode
+ * @param disp              pointer to a display
  * @param render_mode       LV_DISPLAY_RENDER_MODE_PARTIAL/DIRECT/FULL
  */
-void lv_display_set_draw_buffers(lv_display_t * disp, void * buf1, void * buf2, uint32_t buf_size_in_bytes,
-                                 lv_display_render_mode_t render_mode);
+void lv_display_set_render_mode(lv_display_t * disp, lv_display_render_mode_t render_mode);
```
kisvegabor, line 242: Why do we need a separate function for this? I think the rendering mode is coupled with the buffers so they shall be set together. it means less room for error on the user's side.

Are there any uses cases when the render mode is changed but the buffers are not?

XuNeo: I thought from name of the API `lv_display_set_draw_buffers`, it should only update draw buffers. Render mode is kind of independent of buffer. For example, if the buffer is full screen sized, user can choose to use all render mode freely.

There's use case in SDL when window resizes, the draw buffer is updated but not render mode.

XuNeo:
I found the existing SDL driver will crash with certain combination of buffer cound, render mode and stride align.

I tend to make the API change firstly and go back to the existing crash issue later.

src/display/lv_display.h
kisvegabor: I see your point, but I'm afraid a lot of users will forget to set the render mode... 

Let's keep them separated, and show it in all examples and docs. It should be enough too.

XuNeo: Doc updated.

kisvegabor:
Any idea why the MicroPython port is failing? cc @kdschlosser?

XuNeo:
CI has been fixed.
```c

buf = lv.draw_buf_create(WIDTH, HEIGHT, lv.COLOR_FORMAT.RGB888, 0)

disp_drv.set_draw_buffers(buf, None)
disp_drv.set_render_mode(lv.DISPLAY_RENDER_MODE.FULL)
```

kisvegabor (approved):

XuNeo:
> Do we have to update the porting part? @XuNeo

Unfortunately yes. You can follow the examples in this PR for SDL/linuxfb etc. It's quite simple and straightforward.

XuNeo:
That would be much appreciated. Thank you!
I have only tested SDL driver.

GorgonMeducer:
No problem, will do.

GorgonMeducer:
@XuNeo Done

kisvegabor:
I know it's late, but when I reviewed #5260 I started to think about that exactly what problem does it solve if we provide `lv_draw_buf_t`s to `lv_display_set_draw_buf`.

```
uint8_t buf[_LV_DRAW_BUF_SIZE(w,h, cf)]
lv_display_set_draw_buf(disp, buf, NULL);
```
could do the same. 

In case of partial mode the draw_buf is continuously reshaped anyway so setting the stride upfront is not needed. In full refresh and direct mode we can calculate the stride from the horizontal resolution of the display. If the resolution changes the draw bufs are reshaped too. 

Likewise, if the color format of the display's color format changes.

Considering all these I can't see the added value of this change, but it makes the initialization much more complicated.

Finally, I suggest reverting it.

XuNeo:
From the code micro python used, it actually simplifies initialization, especially when you taking stride align, address align into consideration.


```python
buf = lv.draw_buf_create(WIDTH, HEIGHT, lv.COLOR_FORMAT.RGB888, 0)

disp_drv.set_draw_buffers(buf, None)
disp_drv.set_render_mode(lv.DISPLAY_RENDER_MODE.FULL)
```

vs.
```python
color_size = lv.color_format_get_size(disp_drv.get_color_format())

buf = bytearray(WIDTH * HEIGHT * color_size)
disp_drv.set_draw_buffers(
    buf,
    None,
    WIDTH * HEIGHT * color_size,
    lv.DISPLAY_RENDER_MODE.FULL
    )
```


Another motivation is the also related to stride/address align during init. It need to make adjustment internally and we don't know if caller is aware of it. 

```c
void lv_display_set_draw_buffers(lv_display_t * disp, lv_draw_buf_t * buf1, lv_draw_buf_t * buf2)
{
    if(disp == NULL) disp = lv_display_get_default();
    if(disp == NULL) return;

    LV_ASSERT_MSG(buf1 == lv_draw_buf_align(buf1, disp->color_format), "buf1 must be aligned");
    disp->buf_1.data = buf1;
    disp->buf_1.unaligned_data = buf1;
    disp->buf_1.data_size = buf_size_in_bytes;
```

kisvegabor:
I consider MicroPython as a special case as you have minimal control over the memory management. In C/C++ you can decide for where to malloc, use static/global variables, use only an address in external SRAM, etc.

Having an aligned buffer is a real concern however, e.g. for a full frame buffer `lv_draw_buf_create`, or `LV_DRAW_BUF_DEFINE` is not really suitable because most probably you don't want to allocate the large buffer in internal RAM.

Once the user needs to allocate it somewhere, somehow he need to take of the alignment himself. From here, setting stride or size using macros is acceptable I think.

We can also update `_LV_DRAW_BUF_SIZE` to add `LV_DRAW_BUF_STRIDE_ALIGN - 1` to the required size. This way `lv_display_set_draw_buffers` can try to align the buffer and reduce the buffer size accordingly.

Just passing two arrays feel way simpler to me. As LVGL is used by novice developers and makers too, keeping these key APIs as straightforwrd and simple as possible is very important for me.

FASTSHIFT:
It is reasonable to use a unified data structure to describe the drawing buffer.
First, it can keep the original information of the drawing buffer intact in the context, and will not cause inconsistency with the original data due to automatic conversion in the middle.
Second, more parameters may be added in the future due to the need to adapt to different hardware. It is more convenient to expand parameters in the structure, which will hardly damage the existing API design and facilitate future function expansion.

kisvegabor:
Just to be clear, I mean only passing arrays to `lv_display_set_draw_buffers`. I agree to use `lv_draw_buf_t` internally in `lv_display_t`. However I still don't see the added value of the extra complication on the API level because the information stored in the user created `lv_draw_buf_t`s 
1. will be overwritten when the `lv_draw_buf_t`s are reshaped
2. can be calculated (width to stride) or get from other data (`display->color_format`)

> Second, more parameters may be added in the future due to the need to adapt to different hardware. It is more convenient to expand parameters in the structure, which will hardly damage the existing API design and facilitate future function expansion.

I don't think it's real issue. Setting up the buffers is not something new that we don't have experience with. We can say with a high confidence that all we need is:
1. the buffer it self (an array)
2. buffer size
3. color format (stored in `lv_display_t`)
4. stride (can be calculated from the current width)

It's hard to imagine anything else here.

In light of these, if there is no very clear advantage of having `lv_draw_buf_t` in this API I would like keep the original API with array.

PR #5216, opened 2024-01-08
Add automatic refs

kdschlosser:
There is still more work to be done on this. It automatically generates the reference declarations. I have also updated all of the `:ref:` directives that are currently found in the documentation so they point to where they need to. We still have to go through and locate whatever remains of the old markdown links that point to internal pages and update those.

That is all that is left with the references. There are also sell formatting changes that need to be made and that can be in another PR if wanted or we can add that to this PR. 

There is still a bucket lead of things that are not documented that need to be documented. There is also some corrections that have to be made so docstrings are not pointing to things that no longer exist or have been renamed.


kdschlosser:
I did also want to mention that I added support for spell checking. It is not turned on and I am sure we would have to put together a dictionary due to all of the special verbage. If you decide you want to give that a go, lemme know and I will turn it on.

docs/get-started/quick-overview.rst
```diff
@@ -16,7 +14,7 @@ Instead of porting LVGL to embedded hardware straight away, it's highly
 recommended to get started in a simulator first.
 
 LVGL is ported to many IDEs to be sure you will find your favorite one.
-Go to the :ref:`simulator` section to get ready-to-use projects that can be run
+Go to the :ref:`integration/ide/pc-simulator/simulator on pc [0-1]` section to get ready-to-use projects that can be run
```
kisvegabor: What does `[0-1]` mean?

docs/widgets/obj.rst
```diff
@@ -139,26 +125,10 @@ explicitly selected with :cpp:func:`lv_disp_set_default`.
 
 To get an object's screen use the :cpp:expr:`lv_obj_get_screen(obj)` function.
 
-.. _lv_obj_overview_events:
-
-Events
```
PGNetHun, line 144: I think this section could remain here for **Base object**

docs/get-started/quick-overview.rst
kisvegabor: Just found the answer [here](https://github.com/lvgl/lvgl/pull/5193#issuecomment-1880025879).

> it's the file path without the file extension followed by an additional "/" and then the section title. At the end there is a marker of [0-0]
what this marker means is [section-level] How the sections are counted up is when a level drops. This is done so the names cannot overlap.

It will cause maintenance issues when a new section/level is inserted and we need to updated all the related links in the docs. I would rather accept to take care of not having not have conflicting headings on a page. Or at worst case if there will be some, a link will point to the wrong section of the correct page. So it's a small issue happening rarely.

kdschlosser: I can remove those identifiers if you like it's not a big deal to do. but there will be naming collisions. I will make an update to the docs to show you another way of addressing the naming collisions.  I still like the generated names even if the `[*-*]` is not used. I had thought about the same thing with the maintenance aspect of it and I didn't care for it either but thought it still needed to be shown. There are 2 other ways of handling it. One is to rename sections that are duplicates. I don't personally care for this way because is removes uniformity of the documentation. There is another way, I will make those changes so you can see what the other way is.

kdschlosser:
Quick questions.

What happened to the arm2d renderer?? is that still supported?
What happened to the stm32 dma2d renderer, is that still supported?
It also looks like there are nxp vglite renderers, `/src/draw/nxp/vglite` and `/src/draw/vg_lite`. did someone forget to remove  `/src/draw/vg_lite`?

I am going through and cleaning up the API sections that link using some of the references because a lot of them will now aitomatically populate, just have to change the filename of the rst file so it matched the name of the header file (without the `lv_` prefix).

kisvegabor:
> What happened to the arm2d renderer?? is that still supported?

It will be added in the near future.

> What happened to the stm32 dma2d renderer, is that still supported?

I'm still having meeting with ST in this regard.

> It also looks like there are nxp vglite renderers, /src/draw/nxp/vglite and /src/draw/vg_lite. did someone forget to remove /src/draw/vg_lite?

`/src/draw/vg_lite` is generic VG-Lite implementation while `/src/draw/nxp/vg_lite` is dedicated to NXP. We are examining if the can be unified.

docs/CHANGELOG.rst
```diff
@@ -32,12 +32,12 @@ Main new features
...
+- Better parallel rendering architecture. See the details :ref:`here <porting/draw/custom gpu>`
 - Built in display and touch driver: SDL, Linux Frame buffer,  NuttX LCD and touch drivers, ST7789 and ILI9341 driver are available and more will come soon
-- :ref:`others/observer/observer [0-1]` allows to bind data to UI elements and create a uniform and easy to maintain API
+- :ref:`others/observer/index/observer` allows to bind data to UI elements and create a uniform and easy to maintain API
 - GitHub CodeSpace integration makes possible to run LVGL in an Online VSCode editor with 3 click. See more `here <https://blog.lvgl.io/2023-04-13/monthly-newsletter>`__
 - Add vector graphics support via ThorVG. It can be used to draw vector graphics to a `Canvas <https://github.com/lvgl/lvgl/blob/master/examples/widgets/canvas/lv_example_canvas_8.c>`__
-- :ref:`widgets/image/image (lv_image) [0-1]` supports aligning, stretching or tiling the image source if the widget is larger or smaller.
+- :ref:`widgets/image/image (lv_image)` supports aligning, stretching or tiling the image source if the widget is larger or smaller.
```
kisvegabor, line 40: What about using the string from the URL as it is? It would make it simple to just copy paste it.
```suggestion
- :ref:`widgets/image.html#overview` supports aligning, stretching or tiling the image source if the widget is larger or smaller.
```

docs/CHANGELOG.rst
```diff
@@ -32,12 +32,12 @@ Main new features
 - Run time display color format adjustment with RGB888 support
 - Built-in support ``pthread``, ``FreeRTOS`` and other (RT)OSes which are used during rendering
 - Built-in support LVGL's, C library, and other stdlib's
-- Better parallel rendering architecture. See the details :ref:`here <porting/draw/custom gpu [0-1]>`
+- Better parallel rendering architecture. See the details :ref:`here <porting/draw/custom gpu>`
 - Built in display and touch driver: SDL, Linux Frame buffer,  NuttX LCD and touch drivers, ST7789 and ILI9341 driver are available and more will come soon
-- :ref:`others/observer/observer [0-1]` allows to bind data to UI elements and create a uniform and easy to maintain API
+- :ref:`others/observer/index/observer` allows to bind data to UI elements and create a uniform and easy to maintain API
```
kisvegabor, line 37: Why do we have `index` here?

kdschlosser: It's to correct the duplicate "label" issue. The section names should remain the same as they are. If you read the documentation you will understand why they do. In order to handle that the observer documentation has been broken up into several files so the section names are not duplicates. This is because they are in different files.

Look at the file/folder structure in the docs and it will make more sense.

kdschlosser:
OK cool to know about the other renderers. I will add todo's for the renderers that are not added and I will add a document file for the generic vg_lite renderer for the time being. If they get merged into a single component then we can deal with it then.

docs/CHANGELOG.rst
kdschlosser: I modified the document builder code that inserts the API documentation so it will now read index files. This allows for nesting TOC trees into documentation where there is an API section after a TOC tree directive. I use the folder name as the name to find any matching header files when dealing with index.rst files.

kdschlosser:
As far as the vglite and vg_lite is concerned. Possibly a better naming convention to tell the 2 apart? 

`lv_draw_buf_vglite_init_handlers`
`lv_draw_buf_vg_lite_init_handlers`

ehhhhh..... Really similar

docs/CHANGELOG.rst
kdschlosser: I don't understand what you are saying... 

this 

`widgets/image.html#overview` is an invalid reference. I am not even able to define it because the # is an illegal character. Well not an an easy way anyhow, I would have to write a special handler for sphinx in order to create those kinds of references and they could not be defined manually only programmatically. What I mean by that is I would not be able to use `.. _some_ref#some section:` in the rst files. How the references are currently done is the documents folder gets copied to a temp directory and the script edits those files and adds adds the reference declarations as if you had actually keyed them in yourself. It is nice this way because I am able to easily dump a list of the references or go into the temp folder and look at what it is doing to see if there are any problems. Hiding everything behind the scenes so to speak makes it really hard to locate where problems are coming from.

docs/CHANGELOG.rst
kisvegabor: Only the "Observer" section was duplicated, right?

docs/CHANGELOG.rst
kisvegabor: What about this idea?

docs/CHANGELOG.rst
kdschlosser: I should be able to remove the "index" from the references if you would like. I don't think that any name collisions would occur if I do.

kisvegabor:
> As far as the vglite and vg_lite is concerned. Possibly a better naming convention to tell the 2 apart?

Ahh, yeah... @nicusorcitu Could you add the `nxp_vglite` instead of `vglite`  to the public APIs to differentiate your implementation from the generic one?

docs/CHANGELOG.rst
kdschlosser: No there were a couple of other locations that also had duplicated sections. If it was only the one I would have renamed the headers. This is why I thought it would be best to go the route I went, It is also not just one section in a file that is bumping heads. some of the files have 3-4 sections that had naming conflicts.  

again I can remove the need to have "index" in the ref directive if you like.

kdschlosser:
and for sake of consistency I would also rename the pxp ones to nxp_pxp

kdschlosser:
OR... instead of having to do all of that work maybe renaming the vg_lite ones to generic_vglite

kdschlosser:
IDK if there are other "generic" renderers that could be added in the future. planning for the future would be to change the folder structure to `draw/generic/vglite` This would allow for expansion if any additional generic renderers get added.

kisvegabor:
> lv_draw_{mfg}_{accelerator_model}

It sounds good to me.

docs/CHANGELOG.rst
kisvegabor: It's a pity that `#` is illegal. What about some other separator between the page and  the section? E.g. `this_is/a_page__section`

docs/CHANGELOG.rst
PGNetHun: Is it possible to rollback this `observer` to store all the observer doc in 1 file?

src/core/lv_obj_tree.h
```diff
@@ -224,7 +224,7 @@ void lv_obj_tree_walk(lv_obj_t * start_obj, lv_obj_tree_walk_cb_t cb, void * use
 
 /**
  * Iterate through all children of any object and print their ID.
- * @param start_obj     start integrating from this object
+ * @param start_ob     start integrating from this object
```
PGNetHun, line 227: `start_obj` name was ok in comment, the parameter name is wrong

docs/widgets/obj.rst
PGNetHun: oh, I see this "Events" section is moved down, so ignore my previous comment

PGNetHun (commented):
The observer doc should be rolled back to have everything about observer in 1 doc file.

PR #5263, opened 2024-01-10
feat(libjpeg_turbo): add JPEG image EXIF data parsing

liujp-arch:
Parse the rotation information of a JPEG image

### Checkpoints


src/draw/lv_image_buf.h
```diff
@@ -92,6 +92,7 @@ typedef struct {
     uint32_t flags: 16;         /*Image flags, see `lv_image_flags_t`*/
     uint32_t cf : 8;            /*Color format: See `lv_color_format_t`*/
     uint32_t magic: 8;          /*Magic number. Must be LV_IMAGE_HEADER_MAGIC*/
+    uint32_t rotation : 16;     /*Image angle of rotation*/
```
PGNetHun: I think the source image's rotation is not relevant for LVGL drawing feature, so it should not be stored in LVGL image buf, because we don't process that meta info. That is only a meta / EXIF info.

-> The decoder should decode the image in the correct angle, so LVGL can show it correctly.

src/draw/lv_image_buf.h
```diff
@@ -103,6 +104,7 @@ typedef struct {
     uint32_t h: 16;
     uint32_t stride: 16;        /*Number of bytes in a row*/
     uint32_t reserved_2: 16;    /*Reserved to be used later*/
+    uint32_t rotation : 16;     /*Image angle of rotation*/
```
PGNetHun: Same as before: rotation is meta info (EXIF data) in source image, LVGL does not process that info. 
For rotating images there is a transform feature in LVGL, but in your case decoder should decode the image in the correct angle.

src/widgets/image/lv_image.c
```diff
@@ -219,6 +219,7 @@ void lv_image_set_src(lv_obj_t * obj, const void * src)
     img->w        = header.w;
     img->h        = header.h;
     img->cf       = header.cf;
+    img->rotation = header.rotation;
```
PGNetHun: Same as mentioned earlier: source image rotation meta info is not relevant for `lv_image`

PGNetHun (changes requested):

PGNetHun (changes requested):
Please do not store the rotation EXIF data in LVGL image object, instead the decoder should decode the image in the correct angle.

kisvegabor:
> Please do not store the rotation EXIF data in LVGL image object, instead the decoder should decode the image in the correct angle.

I agree.

src/draw/lv_image_buf.h
liujp-arch: Okay, I will fix it.

XuNeo:
Interesting.
What API should we use for decoder to rotate an image(lv_draw_buf_t * decoded)?

PGNetHun:
> Interesting. What API should we use for decoder to rotate an image(lv_draw_buf_t * decoded)?

Well, I think we should not use LVGL to rotate an image, instead the image should be decoded in the correct angle/rotation based on the rotation info in source image meta/EXIF data.
So, the decoder is responsible to decode in the correct angle.

tests/ref_imgs/libs/freetype_1.lp64.png
PGNetHun, line 1: Why is this reference image changed?

tests/ref_imgs/libs/jpg_2.png
PGNetHun, line 1: The reference image is wrong

PGNetHun (changes requested):
To test the code changes, could you please add a small test JPEG file (and the reference PNG file to compare)?
The JPEG file should have that specific EXIF data for which you have created this PR.

tests/ref_imgs/libs/freetype_1.lp64.png
liujp-arch: The old file failed to test on my local device.

tests/ref_imgs/libs/jpg_2.png
PGNetHun, line 1: I doubt that the LVGL logo image on the right side is correct.
The test JPEG contains EXIF with 180 degree rotation, but image is not rotated when displaying it.
Not rotated even in other JPEG viewers (here on Github site (Chrome browser), IrfanView, Paint.NET, etc), so I assume the test image itself is saved wrong.
(online EXIF viewer: https://www.metadata2go.com/view-metadata )

src/libs/libjpeg_turbo/lv_libjpeg_turbo.c
```diff
@@ -37,16 +38,27 @@ static lv_result_t decoder_open(lv_image_decoder_t * decoder, lv_image_decoder_d
...
 /**********************
  *  STATIC VARIABLES
  **********************/
+const int JPEG_EXIF = 0x45786966; /* Exif data structure tag */
+const int JPEG_BIG_ENDIAN_TAG = 0x4d4d;
 
 /**********************
  *      MACROS
  **********************/
+#define TRANS_32_VALUE(big_endian, data) big_endian ? \
```
PGNetHun, line 58: In other systems transforming between little/big endian is called byte swap, so please rename to:
`SWAP_BYTES_32`
`SWAP_BYTES_16`

There is also a macro [`LV_BIG_ENDIAN_SYSTEM`](https://github.com/lvgl/lvgl/blob/master/lv_conf_template.h#L319) in `lv_conf_template.h` to get current system's endianness, so no need to swap bytes if endianness matches.

src/libs/libjpeg_turbo/lv_libjpeg_turbo.c
```diff
@@ -94,22 +106,15 @@ static lv_result_t decoder_info(lv_image_decoder_t * decoder, const void * src,
     if(src_type == LV_IMAGE_SRC_FILE) {
         const char * fn = src;
 
-        lv_fs_file_t f;
-        lv_fs_res_t res = lv_fs_open(&f, fn, LV_FS_MODE_RD);
-        if(res != LV_FS_RES_OK) {
-            LV_LOG_WARN("Can't open file: %s", fn);
+        uint32_t data_size;
+        uint8_t * data = alloc_file(fn, &data_size);
```
PGNetHun: **Performance degradation issue:** the whole file is read by `alloc_file()`, although this function first checks whether file is JPEG (reads first bytes), and only when JPEG signature is found, reads other parts of file and returns image meta data (width, height, etc...)

This `decoder_info()` function is called on source files to detect whether the file is JPEG, so it is called on PNG files (and other file types) too, and when the whole file is read and JPEG signature is not found, then reading the whole file was waste of resources (memory and FS calls).
Please keep the original logic, to read only the first X. bytes to check whether file is really a JPEG.

tests/src/test_assets/test_img_lvgl_logo_with_exif.jpg
PGNetHun, line 1: Is the JPEG correct?
In Chrome browser, IrfanView and Paint.NET the image is upside-down.
But maybe the JPEG decoders in these applications don't handle JPEG rotation info correctly, only some Photo-editors.

src/libs/libjpeg_turbo/lv_libjpeg_turbo.c
```diff
@@ -369,8 +393,10 @@ static lv_draw_buf_t * decode_jpeg_file(const char * filename)
             jpeg_read_scanlines(&cinfo, buffer, 1);
 
             /* Assume put_scanline_someplace wants a pointer and sample count. */
-            lv_memcpy(cur_pos, buffer[0], stride);
-            cur_pos += decoded->header.stride;
+            dispose_buffer(cur_pos, buffer[0], line_index, cinfo.output_width, cinfo.output_height, decoded->header.stride,
```
PGNetHun: Can you please rename function to _rotate_buffer()_? 

_Dispose_ means to me releasing a resource, cleaning up, for example in .NET we use for that purpose.
But there is also a [dispose pattern](https://en.wikipedia.org/wiki/Dispose_pattern) for releasing resources, so dispose is not the best name for rotating image parts.

src/libs/libjpeg_turbo/lv_libjpeg_turbo.c
```diff
@@ -37,16 +38,27 @@ static lv_result_t decoder_open(lv_image_decoder_t * decoder, lv_image_decoder_d
                                 const lv_image_decoder_args_t * args);
 static void decoder_close(lv_image_decoder_t * decoder, lv_image_decoder_dsc_t * dsc);
 static lv_draw_buf_t * decode_jpeg_file(const char * filename);
-static bool get_jpeg_size(const char * filename, uint32_t * width, uint32_t * height);
+static uint8_t * alloc_file(const char * filename, uint32_t * size);
```
PGNetHun: Because you are already editing this file, could you please rename `alloc_file()` to `read_file()` ?
This name describes better what the function does: reads the file content into memory.

PGNetHun (changes requested):

tests/ref_imgs/libs/jpg_2.png
liujp-arch: Well, Most JPEG image viewers have parsed and applied the Exif information, so they display the rotated image, such as browsers, IrfanView, and so on. If we want to view the original information of the image, we can use GIMP or Photoshop, or we can use test_jpgd to test the image, and the test should return failure.

src/libs/libjpeg_turbo/lv_libjpeg_turbo.c
liujp-arch: It's a good idea, but the macro LV_BIG_ENDIAN_SYSTEM is not applicable here. The endianness contained in Exif are only used for the current image, it may be inconsistent with the system's endianness.

src/libs/libjpeg_turbo/lv_libjpeg_turbo.c
liujp-arch: Well, if we need to read Exif information of a file, the original logic will open and close the file twice, I think it's not a good choice.

src/libs/libjpeg_turbo/lv_libjpeg_turbo.c
liujp-arch: thanks for the feedback, I'll update them later.

tests/src/test_assets/test_img_lvgl_logo_with_exif.jpg
liujp-arch: I think it is correct, we can open a JPG image in binary mode and view the rotation information based on the Exif format. [https://www.media.mit.edu/pia/Research/deepview/exif.html](url)

PGNetHun: I've opened the image in Gimp, and it popped up this window:
image

Was your intention to show up the image deliberately upside-down?
I thought that you would store the image rotated, mark it as 180 degree rotated, and the decoder should show in correct angle: 
image

That would be better to avoid misunderstandings why the second image is upside-down.

PGNetHun (changes requested):

src/libs/libjpeg_turbo/lv_libjpeg_turbo.c
PGNetHun: I **do** think we should keep the original logic (read only first X. bytes), because [lv_image_decoder.c -> lv_image_decoder_get_info()](https://github.com/lvgl/lvgl/blob/master/src/draw/lv_image_decoder.c#L93) function iterates through the decoders to find an applicable one (it tries decoders one-by-one), and if this libjpeg_turbo decoder is the first in the list and reads always the whole file (size could be also >100 kB) and it can't find the JPEG signature (because the file we wanted to show is for example PNG), then it has read the whole file needlessly. LVGL FS cache is not a global cache (at the moment), so the cached buffer is released after file is closed, and the next decoder also reads part of the file.

A performance problem popped up also related to this `decoder->info_cb()` callback, that it reads continously file content:
https://forum.lvgl.io/t/about-latest-multi-instance-updates/14258/2

tests/src/test_assets/test_img_lvgl_logo_with_exif.jpg
liujp-arch: To avoid any misunderstanding, I will replace the current image with a new one.

src/libs/libjpeg_turbo/lv_libjpeg_turbo.c
liujp-arch: Okay, I see. I will keep the original logic.

PGNetHun (approved):
Only 2 CI builds fail on this libjpeg changes.

PGNetHun (changes requested):
Please fix CI build errors -> libjpeg test fails.
It seems there is a memory leak somehow.

liujp-arch:
looks like need more memory actually.

PGNetHun (dismissed):

kisvegabor:
I've checked the [test image](https://github.com/lvgl/lvgl/pull/5263/files#diff-bda4cca649f1fec8dc422169f231021a9cbc350acabb97b4d4e4942efcf4ed67), found 2 non rotated image, and thoght something is not working.  And them saw [this comment](https://github.com/lvgl/lvgl/pull/5263#discussion_r1453045873). 

I think showing the 180° rotated image as normal is less intuitive.

XuNeo:
If I understand correctly, this PR introduces capability to rotate JPEG image as instructed. 
So the better test case would be a 180 rotated logo . Seeing the image is rotated means it works.

PGNetHun:
> If I understand correctly, this PR introduces capability to rotate JPEG image as instructed. So the better test case would be a 180 rotated logo . Seeing the image is rotated means it works.

I assume the intention was to show a rotated image correct: 
for example a portrait image, which is stored 90 degree rotated in JPEG.

In this PR maybe a label could be added below the 2 images (what show up correct, without upside-down):
- not rotated
- 180 degree rotated

kisvegabor:
Ah,it seems I'm really misunderstanding something. Let's say I take a portrait picture with my camera. I guess the EXIF rotation will be 90°. But if we ignore the EXIF data, what image will be shown? A portrait or a landscape?

liujp-arch:
it‘s actually like this, 

when we take a picture with camera(i mean really camera), we may not always hold the camera in such a position that the camera top corresponds to top of the scene, if we hold the camera with orientation flag '6' in above image, the Exif rotation will be 90°. it's amount to take a portrait picture with a phone camera.

kisvegabor:
All clear, thank you! 

If by processing the EXIF data we get exactly the same output as with a normal image, I suggest using this the [original reference image](https://github.com/lvgl/lvgl/blob/master/tests/ref_imgs/libs/jpg_2.png) for all 4 EXIF options.

W-Mai (approved):

terry0012:
@liujp-arch @kisvegabor  
Are there any other points of concern regarding this feature? Can we merge it?

My personal suggestion is to build a set of basic function APIs for image processing in future, such as rotation and scaling, similar to the draw API. This will unify their usage in image decoding and other parts, and also make it easier to accelerate them using SIMD in the future.

PGNetHun:
The actual code is OK (for me), there was only 1 suggestion:
have different input JPEG files with different rotations, but only 1 reference image.
See: https://github.com/lvgl/lvgl/pull/5263#issuecomment-1900056348

But I think we can do it also later, after merging this PR.

terry0012:
> All clear, thank you!
> 
> If by processing the EXIF data we get exactly the same output as with a normal image, I suggest using this the [original reference image](https://github.com/lvgl/lvgl/blob/master/tests/ref_imgs/libs/jpg_2.png) for all 4 EXIF options.

image

@kisvegabor you mean all the 4 EXIF options images should be decoded to the left image in the last line ?

kisvegabor:
> @kisvegabor you mean all the 4 EXIF options images should be decoded to the left image in the last line ?

Yes, we can have only a single reference image like this:
image

and all 4 variants can be compared to that.

kisvegabor:
I think it's a useful feature and only a little is missing. Did not give up now! 🙂

liujp-arch:
I need to do more testing on the rotation of the image. I will update once it's finished.

tests/ref_imgs/libs/jpg_2.png
PGNetHun, line 1: Well, I thought that you rotate the image RAW data, set the orientation in EXIF (like in real camera images), so here in test reference image (where we see that EXIF rotation functionality works) every image is shown in correct orientation like this:
image

PGNetHun (approved):

FASTSHIFT (approved):

PR #5272, opened 2024-01-11
feat(display): assert when buffer is not aligned

XuNeo:

Default to 1 to make it behave same to lvgl v8.
To avoid confusions raised in lvgl/lv_binding_micropython#313

### Checkpoints

### DIFF
--- a/src/display/lv_display.c
+++ b/src/display/lv_display.c
@@ -398,27 +398,29 @@ void lv_display_set_draw_buffers(lv_display_t * disp, lv_draw_buf_t * buf1, lv_d
 void lv_display_set_buffers(lv_display_t * disp, void * buf1, void * buf2, uint32_t buf_size,
                             lv_display_render_mode_t render_mode)
 {
+    LV_ASSERT_MSG(buf1 != NULL, "Null buffer");
+    lv_color_format_t cf = lv_display_get_color_format(disp);
     uint32_t w = lv_display_get_horizontal_resolution(disp);
     uint32_t h = lv_display_get_vertical_resolution(disp);
-    LV_ASSERT(w != 0 && h != 0);
 
-    lv_color_format_t cf = lv_display_get_color_format(disp);
+    LV_ASSERT_MSG(w != 0 && h != 0, "display resolution is 0");
+
+    /* buf1 or buf2 is not aligned according to LV_DRAW_BUF_ALIGN */
+    LV_ASSERT_MSG(buf1 == lv_draw_buf_align(buf1, cf), "buf1 is not aligned: %p", buf1);
+    LV_ASSERT_MSG(buf2 == NULL || buf2 == lv_draw_buf_align(buf2, cf), "buf2 is not aligned: %p", buf2);
+
     uint32_t stride = lv_draw_buf_width_to_stride(w, cf);
     if(render_mode == LV_DISPLAY_RENDER_MODE_PARTIAL) {
         /* for partial mode, we calculate the height based on the buf_size and stride */
         h = buf_size / stride;
     }
-    else if(stride * h > buf_size) {
-        LV_LOG_ERROR("%s mode requires screen sized buffer(s)",
-                     render_mode == LV_DISPLAY_RENDER_MODE_FULL ? "FULL" : "DIRECT");
-        LV_ASSERT(0);
+    else {
+        LV_ASSERT_MSG(stride * h < buf_size, "%s mode requires screen sized buffer(s)",
+                      render_mode == LV_DISPLAY_RENDER_MODE_FULL ? "FULL" : "DIRECT");
         return;
     }
 
-    if(h == 0) {
-        LV_ASSERT_MSG(h != 0, "the buffer is too small");
-        return;
-    }
+    LV_ASSERT_MSG(h != 0, "the buffer is too small");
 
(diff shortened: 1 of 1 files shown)


kisvegabor:
Even SW render requires 4 byte alignment to read/write `uint32_t` data. Can we assume that it's fulfilled naturally?

kisvegabor:
To approach is differently, if all memories are 4 byte aligned by default `LV_DRAW_BUF_ALIGN   4` shouldn't change anything. But at least it acts as a check if the address is really ok.

PGNetHun (approved):

XuNeo:
For some CPU, the unaligned access is supported. I wonder if align to 4 will cause trouble like the one in micropython.

Set to 4 however will only cover the issue if unaligned access is not supported. The caller must be aware that it needs additional bytes in order to adjust alignment.

kisvegabor:
> For some CPU, the unaligned access is supported. 

True, but probably way slower. So using `4` as default value would prevent hard to find performance issues too.

>  I wonder if align to 4 will cause trouble like the one in micropython.

To be honest I don't understand the MicroPython issue. Is it a problem if the allocated buffer is aligned?  With `LV_DRAW_BUF_ALIGN  4` nothing should happen (I assume `esp.heap_caps_malloc` returns at least word aligned memory)

> Set to 4 however will only cover the issue if unaligned access is not supported. The caller must be aware that it needs additional bytes in order to adjust alignment.

In `lv_display_set_*_buffers` I think an ASSERT without any extra magic if `LV_DRAW_BUF_ALIGN` is not fulfilled and `LV_DRAW_BUF_ALIGN  4` would be a good default value/behavior.

PGNetHun:
Yes, hopefully `esp.heap_caps_malloc` function (by passing `MALLOC_CAP_DMA` flag) returns 32-bit aligned memory. 

Unfortunately `MALLOC_CAP_32BIT` flag is not present in `esp.MALLOC_CAP` enum in ESP-IDF 4.4.5:
image

Related document:
https://docs.espressif.com/projects/esp-idf/en/latest/esp32/api-reference/system/mem_alloc.html#dma-capable-memory

kisvegabor:
If not 32 bit aligned is returned, I think it's better to assert. IMO expecting at least 4 byte align is quite reasonable.

XuNeo:
> I think it's better to assert.

Do you mean we should assert in `lv_display_set_buffers` instead of secretly align it?

kisvegabor:
> Do you mean we should assert in lv_display_set_buffers instead of secretly align it?

Yes, I think we can expect the user to do the alignment externally instead of doing hidden magic.

kisvegabor (dismissed):

PGNetHun (approved):

src/display/lv_display.c
```diff
@@ -398,11 +398,29 @@ void lv_display_set_draw_buffers(lv_display_t * disp, lv_draw_buf_t * buf1, lv_d
 void lv_display_set_buffers(lv_display_t * disp, void * buf1, void * buf2, uint32_t buf_size,
                             lv_display_render_mode_t render_mode)
 {
+    LV_ASSERT(buf1 != NULL);
+    if(buf1 == NULL) {
+        LV_LOG_ERROR("buf1 is NULL");
```
FASTSHIFT: LV_ASSERT_NULL(buf1);

src/display/lv_display.c
```diff
@@ -398,11 +398,29 @@ void lv_display_set_draw_buffers(lv_display_t * disp, lv_draw_buf_t * buf1, lv_d
...
     uint32_t w = lv_display_get_horizontal_resolution(disp);
     uint32_t h = lv_display_get_vertical_resolution(disp);
     LV_ASSERT(w != 0 && h != 0);
 
-    lv_color_format_t cf = lv_display_get_color_format(disp);
+    if(buf1 != lv_draw_buf_align(buf1, cf)) {
+        /* buf1 is not aligned according to LV_DRAW_BUF_ALIGN */
+        LV_LOG_ERROR("buf1 is not aligned");
+        return;
+    }
```
FASTSHIFT: LV_ASSERT_MSG(buf1 == lv_draw_buf_align(buf1, cf), "buf1 is not aligned");

src/display/lv_display.c
```diff
@@ -398,11 +398,29 @@ void lv_display_set_draw_buffers(lv_display_t * disp, lv_draw_buf_t * buf1, lv_d
...
+
+    if(buf2 && buf2 != lv_draw_buf_align(buf2, cf)) {
+        /* buf2 is not aligned according to LV_DRAW_BUF_ALIGN */
+        LV_LOG_ERROR("buf2 is not aligned");
+        return;
+    }
```
FASTSHIFT: if(buf2 != NULL)
{
    LV_ASSERT_MSG(buf2 == lv_draw_buf_align(buf2, cf), "buf2 is not aligned");
}

XuNeo: You have to add return when failed. Thus no need to use LV_ASSERT_MSG at all.

XuNeo:
Added `LV_ASSERT_MSG` so it will assert if enabled.
CI needs https://github.com/lvgl/lvgl/pull/5392 to pass.

kisvegabor (approved):

FASTSHIFT (approved):

PR #5284, opened 2024-01-11
test(obj): Add initial tests for `lv_obj_move_to_index`

C47D:

Improve `lv_obj_move_to_index` as ravioli marks it's complexity as 9.
Planned to be merged into master after v9 release

### Checkpoints

### DIFF
--- a/src/core/lv_obj_tree.c
+++ b/src/core/lv_obj_tree.c
@@ -210,7 +210,9 @@ void lv_obj_move_to_index(lv_obj_t * obj, int32_t index)
     }
 
     const uint32_t parent_child_count = lv_obj_get_child_count(parent);
+    /* old_index only can be 0 or greater, this point can not be reached if the parent is null */
     const int32_t old_index = lv_obj_get_index(obj);
+    LV_ASSERT(0 <= old_index);
 
     if(index < 0) {
         index += parent_child_count;
@@ -416,14 +418,16 @@ int32_t lv_obj_get_index(const lv_obj_t * obj)
     LV_ASSERT_OBJ(obj, MY_CLASS);
 
     lv_obj_t * parent = lv_obj_get_parent(obj);
-    if(parent == NULL) return 0xFFFFFFFF;
+    if(parent == NULL) return -1;
 
     int32_t i = 0;
     for(i = 0; i < parent->spec_attr->child_cnt; i++) {
         if(parent->spec_attr->children[i] == obj) return i;
     }
 
-    return -1; /*Shouldn't happen*/
+    /*Shouldn't reach this point*/
+    LV_ASSERT(0);
+    return -1;
 }
 
 int32_t lv_obj_get_index_by_type(const lv_obj_t * obj, const lv_obj_class_t * class_p)
--- a/src/core/lv_obj_tree.c
+++ b/src/core/lv_obj_tree.c
@@ -202,22 +202,28 @@ void lv_obj_move_to_index(lv_obj_t * obj, int32_t index)
 {
     LV_ASSERT_OBJ(obj, MY_CLASS);
 
+    /* Check parent validity */
     lv_obj_t * parent = lv_obj_get_parent(obj);
-
     if(!parent) {
         LV_LOG_WARN("parent is NULL");
         return;
     }
 
-    if(index < 0) {
-        index = lv_obj_get_child_count(parent) + index;
-    }
-
+    const uint32_t parent_child_count = lv_obj_get_child_count(parent);
     const int32_t old_index = lv_obj_get_index(obj);
 
-    if(index < 0) return;
-    if(index >= (int32_t) lv_obj_get_child_count(parent)) return;
-    if(index == old_index) return;
+    if(index < 0) {
+        index += parent_child_count;
+    }
+
+    if((index < 0)
+       /* Index is same or bigger than parent child count */
+       || (index >= (int32_t) parent_child_count)
+       /* If both previous and new index are the same */
+       || (index == old_index)) {
+
+        return;
+    }
 
     int32_t i = old_index;
     if(index < old_index) {
--- a/src/core/lv_obj_tree.c
+++ b/src/core/lv_obj_tree.c
@@ -210,7 +210,7 @@ void lv_obj_move_to_index(lv_obj_t * obj, int32_t index)
     }
 
     const uint32_t parent_child_count = lv_obj_get_child_count(parent);
-    /* old_index only can be 0 or greater, this point can not be reached if the parent is null */
+    /* old_index only can be 0 or greater, this point can not be reached if the parent is not null */
     const int32_t old_index = lv_obj_get_index(obj);
     LV_ASSERT(0 <= old_index);
 
--- a/src/core/lv_obj_tree.c
+++ b/src/core/lv_obj_tree.c
@@ -216,6 +216,7 @@ void lv_obj_move_to_index(lv_obj_t * obj, int32_t index)
         index += parent_child_count;
     }
 
+    /* Index was negative and the absolute value is greater than parent child count */
     if((index < 0)
        /* Index is same or bigger than parent child count */
        || (index >= (int32_t) parent_child_count)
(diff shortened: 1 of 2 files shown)


src/core/lv_obj_tree.c
```diff
@@ -202,22 +202,28 @@ void lv_obj_move_to_index(lv_obj_t * obj, int32_t index)
...
+    const uint32_t parent_child_count = lv_obj_get_child_count(parent);
+    const int32_t old_index = lv_obj_get_index(obj);
+
     if(index < 0) {
-        index = lv_obj_get_child_count(parent) + index;
+        index += parent_child_count;
     }
 
-    const int32_t old_index = lv_obj_get_index(obj);
+    if((index < 0)
```
C47D, line 222: When this condition can be TRUE? If the `index < 0` is TRUE above, this is not possible, isn't?

C47D:
@kisvegabor I'm seeing something not expected in the coverage report of this function. There are some lines that are shown as not executed, but they are (I've debugged the test cases)

image

src/core/lv_obj_tree.c
kisvegabor: Can be `-1` on error 
- [here](https://github.com/lvgl/lvgl/blob/master/src/core/lv_obj_tree.c#L412) (should be `return -1`)
- or [here](https://github.com/lvgl/lvgl/blob/master/src/core/lv_obj_tree.c#L419)

kisvegabor:
I' think it's because of -O3 optimization. Can you try with -O0?

C47D:
> I' think it's because of -O3 optimization. Can you try with -O0?

Hi, I wasn't able to find the optimization configuration, is this in `CMakeList.txt` on the root directory or lvgl or the one available in the `tests` dir?

src/core/lv_obj_tree.c
C47D: So, should I replace the return value (0xFFFFFFFF) with -1?

```diff
diff --git a/src/core/lv_obj_tree.c b/src/core/lv_obj_tree.c
index be1c797f1..3b569ce46 100644
--- a/src/core/lv_obj_tree.c
+++ b/src/core/lv_obj_tree.c
@@ -415,7 +415,7 @@ int32_t lv_obj_get_index(const lv_obj_t * obj)
     LV_ASSERT_OBJ(obj, MY_CLASS);
 
     lv_obj_t * parent = lv_obj_get_parent(obj);
-    if(parent == NULL) return 0xFFFFFFFF;
+    if(parent == NULL) return -1;
 
     int32_t i = 0;
     for(i = 0; i < parent->spec_attr->child_cnt; i++) {
```

kisvegabor:
You can try `cmake .. -DCMAKE_BUILD_TYPE=Debug`

C47D:
> You can try `cmake .. -DCMAKE_BUILD_TYPE=Debug`

Debug build type is the type configured.

C47D:
I cleaned the tests report directory and tried to generate the report again and I'm getting an empty report, maybe the report generation broke somehow and the code coverage wasn't following the latest changes.

After executing this command in the tests directory
```console
python3 main.py --clean --report test --test-suite tree
```
I get an almost empty report directory
```console
$ ls report
coverage.xml  index.css  index.html
```
And the empty report:
image

C47D:
I'm getting closer, see the message at the end
```console
====================================================================
Building: test_sysheap: Test config, system heap, 32 bit color depth
====================================================================

-- The C compiler identification is GNU 11.4.0
-- The CXX compiler identification is GNU 11.4.0
-- Performing Test FLAG_SUPPORTED_FOR_C_--COVERAGE
-- Performing Test FLAG_SUPPORTED_FOR_C_--COVERAGE - Failed
```
Same goes for
```console
-- Performing Test FLAG_SUPPORTED_FOR_C_-PEDANTIC-ERRORS
-- Performing Test FLAG_SUPPORTED_FOR_C_-PEDANTIC-ERRORS - Failed

-- Performing Test FLAG_SUPPORTED_FOR_C_-WERROR
-- Performing Test FLAG_SUPPORTED_FOR_C_-WERROR - Failed

-- Performing Test FLAG_SUPPORTED_FOR_C_-FSANITIZE=ADDRESS
-- Performing Test FLAG_SUPPORTED_FOR_C_-FSANITIZE=ADDRESS - Failed
```

kisvegabor:
I've also noticed those messages but to be honest I don't know what they are. Recently the CMakeFiles were updated by @XuNeo, @FASTSHIFT, and  @W-Mai. Could you comment on it?

XuNeo:
Those flag tests are added by @W-Mai , `Failed` means this flag is not supported by compiler used, thus won't be added to CFLAG.
The message is only for hint, it's not the reason of failure.

W-Mai:
I added the compiler's flag support capability check function, and there may be some problems with this code. I haven't tested the use of coverage. Let me take a look at this problem.

C47D:
Thanks, it's been a while since the last time I used coverage, so I didn't tried to go back in time to see where it started failing.

Additional information about my setup:
- using gcc 11
- I had to install freetype recently, using the install-prerequisites script. This was last week I think.

kisvegabor:
> should we open a new issue for this coverage thing?

Yes, please!

W-Mai:
The fix PR is here #5356

C47D:
That was quick, thanks @W-Mai I will test it later today.

C47D:
Now we have full coverage of `lv_obj_move_to_index`

image

@kisvegabor Is the comment added in d6d576c correct?
`/* Index was negative and the absolute value is greater than parent child count */`

src/core/lv_obj_tree.c
```diff
@@ -416,14 +418,16 @@ int32_t lv_obj_get_index(const lv_obj_t * obj)
...
+    if(parent == NULL) return -1;
 
     int32_t i = 0;
     for(i = 0; i < parent->spec_attr->child_cnt; i++) {
         if(parent->spec_attr->children[i] == obj) return i;
     }
 
-    return -1; /*Shouldn't happen*/
+    /*Shouldn't reach this point*/
+    LV_ASSERT(0);
```
C47D, line 429: We shouldn't reach this point, that's why I've added the `LV_ASSERT(0)`, which will always fails if reached and assertions are enabled.

src/core/lv_obj_tree.c
```diff
@@ -210,7 +210,9 @@ void lv_obj_move_to_index(lv_obj_t * obj, int32_t index)
     }
 
     const uint32_t parent_child_count = lv_obj_get_child_count(parent);
+    /* old_index only can be 0 or greater, this point can not be reached if the parent is null */
     const int32_t old_index = lv_obj_get_index(obj);
+    LV_ASSERT(0 <= old_index);
```
C47D, line 215: Fail assertion if `old_index` is `-1`, this only happens  when parent is NULL, but this was checked above.

src/core/lv_obj_tree.c
XuNeo: Assert could be disabled... thus we should add return -1 at the end.

C47D: Yep, that's why the return statement was kept

src/core/lv_obj_tree.c
```diff
@@ -202,22 +202,31 @@ void lv_obj_move_to_index(lv_obj_t * obj, int32_t index)
...
+    /* Check parent validity */
     lv_obj_t * parent = lv_obj_get_parent(obj);
-
     if(!parent) {
         LV_LOG_WARN("parent is NULL");
         return;
     }
 
+    const uint32_t parent_child_count = lv_obj_get_child_count(parent);
+    /* old_index only can be 0 or greater, this point can not be reached if the parent is null */
```
kisvegabor:
```suggestion
    /* old_index only can be 0 or greater, this point can not be reached if the parent is not null */
```

kisvegabor (commented):
Thank you all. I've found only a minor typo.

kisvegabor (approved):

kisvegabor:
It should be good to go! Thank you!

PR #5286, opened 2024-01-11
fix(display): fix the support for partial mode in lv_display_set_buffers

GorgonMeducer:

The `lv_display_set_buffers` uses screen height in partial mode.  This isn't right.

### Checkpoints

### DIFF
--- a/src/display/lv_display.c
+++ b/src/display/lv_display.c
@@ -407,7 +407,7 @@ void lv_display_set_buffers(lv_display_t * disp, void * buf1, void * buf2, uint3
     uint32_t stride = lv_draw_buf_width_to_stride(w, cf);
     if(LV_DISPLAY_RENDER_MODE_PARTIAL == render_mode) {
         h = buf_size / stride;
-        if (0 == h) {
+        if(0 == h) {
             LV_LOG_ERROR("the buffer is too small for PARTIAL mode");
             LV_ASSERT(0);
             return;
--- a/src/display/lv_display.c
+++ b/src/display/lv_display.c
@@ -405,15 +405,19 @@ void lv_display_set_buffers(lv_display_t * disp, void * buf1, void * buf2, uint3
 
     lv_color_format_t cf = lv_display_get_color_format(disp);
     uint32_t stride = lv_draw_buf_width_to_stride(w, cf);
-    if(stride * h > buf_size && render_mode != LV_DISPLAY_RENDER_MODE_PARTIAL) {
+    if(LV_DISPLAY_RENDER_MODE_PARTIAL == render_mode) {
+        h = buf_size / stride;
+        if (0 == h) {
+            LV_LOG_ERROR("the buffer is too small for PARTIAL mode");
+            LV_ASSERT(0);
+            return;
+        }
+    }
+    else if(stride * h > buf_size) {
         LV_LOG_ERROR("%s mode requires screen sized buffer(s)",
                      render_mode == LV_DISPLAY_RENDER_MODE_FULL ? "FULL" : "DIRECT");
         LV_ASSERT(0);
         return;
-    }
-
-    if(LV_DISPLAY_RENDER_MODE_PARTIAL == render_mode) {
-        h = buf_size / stride;
     }
 
     lv_draw_buf_init(&disp->_static_buf1, w, h, cf, stride, buf1, buf_size);
--- a/src/display/lv_display.c
+++ b/src/display/lv_display.c
@@ -407,7 +407,11 @@ void lv_display_set_buffers(lv_display_t * disp, void * buf1, void * buf2, uint3
     uint32_t stride = lv_draw_buf_width_to_stride(w, cf);
     if(render_mode == LV_DISPLAY_RENDER_MODE_PARTIAL) {
         h = buf_size / stride;
-        LV_ASSERT_MSG(h != 0, "the buffer is too small for PARTIAL mode");
+        if(h == 0) {
+            LV_LOG_ERROR("the buffer is too small for PARTIAL mode");
+            LV_ASSERT(0);
+            return;
+        }
     }
     else if(stride * h > buf_size) {
         LV_LOG_ERROR("%s mode requires screen sized buffer(s)",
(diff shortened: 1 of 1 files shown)


kisvegabor (dismissed):
@XuNeo, please review.

src/display/lv_display.c
```diff
@@ -412,6 +412,10 @@ void lv_display_set_buffers(lv_display_t * disp, void * buf1, void * buf2, uint3
         return;
     }
 
+    if(LV_DISPLAY_RENDER_MODE_PARTIAL == render_mode) {
```
XuNeo: move up to 408 and change 408 to else if

XuNeo (commented):
Thank you, misse it somehow.

src/display/lv_display.c
GorgonMeducer: Done

GorgonMeducer:
@XuNeo Please check. Thanks.

src/display/lv_display.c
```diff
@@ -405,7 +405,15 @@ void lv_display_set_buffers(lv_display_t * disp, void * buf1, void * buf2, uint3
 
     lv_color_format_t cf = lv_display_get_color_format(disp);
     uint32_t stride = lv_draw_buf_width_to_stride(w, cf);
-    if(stride * h > buf_size && render_mode != LV_DISPLAY_RENDER_MODE_PARTIAL) {
+    if(LV_DISPLAY_RENDER_MODE_PARTIAL == render_mode) {
```
FASTSHIFT: render_mode == LV_DISPLAY_RENDER_MODE_PARTIAL
Follow the overall style of lvgl.

src/display/lv_display.c
```diff
@@ -405,7 +405,15 @@ void lv_display_set_buffers(lv_display_t * disp, void * buf1, void * buf2, uint3
...
+        h = buf_size / stride;
+        if(0 == h) {
+            LV_LOG_ERROR("the buffer is too small for PARTIAL mode");
+            LV_ASSERT(0);
+            return;
```
FASTSHIFT: uint32_t min_h = buf_size / stride;
LV_ASSERT_MSG(min_h != 0, "the buffer is too small for PARTIAL mode");

XuNeo: Parameter `h` must be updated too. Change assert message to `> 0`
```
h = buf_size / stride;
LV_ASSERT_MSG(h > 0, "the buffer is too small for PARTIAL mode");
```

GorgonMeducer: If we use min_h, how could we update the h?

src/display/lv_display.c
GorgonMeducer: done

XuNeo (approved):

src/display/lv_display.c
```diff
@@ -405,7 +405,11 @@ void lv_display_set_buffers(lv_display_t * disp, void * buf1, void * buf2, uint3
 
     lv_color_format_t cf = lv_display_get_color_format(disp);
     uint32_t stride = lv_draw_buf_width_to_stride(w, cf);
-    if(stride * h > buf_size && render_mode != LV_DISPLAY_RENDER_MODE_PARTIAL) {
+    if(render_mode == LV_DISPLAY_RENDER_MODE_PARTIAL) {
+        h = buf_size / stride;
+        LV_ASSERT_MSG(h != 0, "the buffer is too small for PARTIAL mode");
+    }
```
kisvegabor, line 411: Add `return` here in case of asserts are disabled?

GorgonMeducer: I added return before, but @FASTSHIFT asked me to remove it...

GorgonMeducer: Added return ...again... @kisvegabor

XuNeo:
Emm, that reminds me the stuck issue on other platforms. Could we change default assert handler to exit()?

src/display/lv_display.c
XuNeo: I think below is more suitable.
```c
h = buf_size / stride;
LV_ASSERT_MSG(h > 0, "the buffer is too small for PARTIAL mode");
return;
```

XuNeo (commented):
Thanks. I think the return should added just below assert message incase LV_ASSERT is not killing execution.

src/display/lv_display.c
PGNetHun: > I think below is more suitable.
> 
> ```c
> h = buf_size / stride;
> LV_ASSERT_MSG(h > 0, "the buffer is too small for PARTIAL mode");
> return;
> ```

well, I think in this case it will always `return`

PGNetHun (approved):

src/display/lv_display.c
kisvegabor: > well, I think in this case it will always return

Wow, you are right! This?
```c
    if(render_mode == LV_DISPLAY_RENDER_MODE_PARTIAL) {
        h = buf_size / stride;
    }
    ...
    if(h == 0) {
       LV_ASSERT_MSG(h != 0, "the buffer is too small");
       return; 
    }   

```

GorgonMeducer: > well, I think in this case it will always return

No, that's not right.

GorgonMeducer: We calculate the h for partial mode based on the stride and buf size.

GorgonMeducer:
Updated.

XuNeo (approved):

src/display/lv_display.c
PGNetHun: I wonder whether `return;` is still needed after `LV_ASSERT()` & `LV_ASSERT_MSG()` lines, because in current configuration those macros end up in infinite loop:
https://github.com/lvgl/lvgl/blob/master/src/misc/lv_assert.h#L37
https://github.com/lvgl/lvgl/blob/master/lv_conf_template.h#L232

(but correct me if I'm wrong)

I hope this fix gets in `master` branch soon, because it's needed for [MicroPython LVGL v9 update PR](https://github.com/lvgl/lv_binding_micropython/pull/311).

PGNetHun (changes requested):

src/display/lv_display.c
kisvegabor: The ASSERTS can be disabled. In this case they do nothing but we still should return on error.

kisvegabor (approved):

GorgonMeducer:
Thank you!

PR #5300, opened 2024-01-12
Tuning API of `lv_array` module

XuNeo:

Miscellaneous `lv_array` API update.
The API names have been updated slightly to match up with STL vector.

This was done preparing for modules `lv_events` etc.

### Checkpoints

### DIFF
--- a/src/draw/vg_lite/lv_draw_vg_lite_vector.c
+++ b/src/draw/vg_lite/lv_draw_vg_lite_vector.c
@@ -236,25 +236,25 @@ static void lv_path_to_vg(lv_vg_lite_path_t * dest, const lv_vector_path_t * src
 
     uint32_t pidx = 0;
     for(uint32_t i = 0; i < src->ops.size; i++) {
-        lv_vector_path_op_t * op = LV_ARRAY_GET(&src->ops, i, uint8_t);
+        lv_vector_path_op_t * op = lv_array_at(&src->ops, i);
         switch(*op) {
             case LV_VECTOR_PATH_OP_MOVE_TO: {
-                    const lv_fpoint_t * pt = LV_ARRAY_GET(&src->points, pidx, lv_fpoint_t);
+                    const lv_fpoint_t * pt = lv_array_at(&src->points, pidx);
                     CMP_BOUNDS(pt);
                     lv_vg_lite_path_move_to(dest, pt->x, pt->y);
                     pidx += 1;
                 }
                 break;
             case LV_VECTOR_PATH_OP_LINE_TO: {
-                    const lv_fpoint_t * pt = LV_ARRAY_GET(&src->points, pidx, lv_fpoint_t);
+                    const lv_fpoint_t * pt = lv_array_at(&src->points, pidx);
                     CMP_BOUNDS(pt);
                     lv_vg_lite_path_line_to(dest, pt->x, pt->y);
                     pidx += 1;
                 }
                 break;
             case LV_VECTOR_PATH_OP_QUAD_TO: {
-                    const lv_fpoint_t * pt1 = LV_ARRAY_GET(&src->points, pidx, lv_fpoint_t);
-                    const lv_fpoint_t * pt2 = LV_ARRAY_GET(&src->points, pidx + 1, lv_fpoint_t);
+                    const lv_fpoint_t * pt1 = lv_array_at(&src->points, pidx);
+                    const lv_fpoint_t * pt2 = lv_array_at(&src->points, pidx);
                     CMP_BOUNDS(pt1);
                     CMP_BOUNDS(pt2);
                     lv_vg_lite_path_quad_to(dest, pt1->x, pt1->y, pt2->x, pt2->y);
@@ -262,9 +262,9 @@ static void lv_path_to_vg(lv_vg_lite_path_t * dest, const lv_vector_path_t * src
                 }
                 break;
             case LV_VECTOR_PATH_OP_CUBIC_TO: {
-                    const lv_fpoint_t * pt1 = LV_ARRAY_GET(&src->points, pidx, lv_fpoint_t);
-                    const lv_fpoint_t * pt2 = LV_ARRAY_GET(&src->points, pidx + 1, lv_fpoint_t);
-                    const lv_fpoint_t * pt3 = LV_ARRAY_GET(&src->points, pidx + 2, lv_fpoint_t);
+                    const lv_fpoint_t * pt1 = lv_array_at(&src->points, pidx);
+                    const lv_fpoint_t * pt2 = lv_array_at(&src->points, pidx + 1);
+                    const lv_fpoint_t * pt3 = lv_array_at(&src->points, pidx + 2);
                     CMP_BOUNDS(pt1);
                     CMP_BOUNDS(pt2);
                     CMP_BOUNDS(pt3);
--- a/src/draw/vg_lite/lv_draw_vg_lite_vector.c
+++ b/src/draw/vg_lite/lv_draw_vg_lite_vector.c
@@ -254,7 +254,7 @@ static void lv_path_to_vg(lv_vg_lite_path_t * dest, const lv_vector_path_t * src
                 break;
             case LV_VECTOR_PATH_OP_QUAD_TO: {
                     const lv_fpoint_t * pt1 = lv_array_at(&src->points, pidx);
-                    const lv_fpoint_t * pt2 = lv_array_at(&src->points, pidx);
+                    const lv_fpoint_t * pt2 = lv_array_at(&src->points, pidx + 1);
                     CMP_BOUNDS(pt1);
                     CMP_BOUNDS(pt2);
                     lv_vg_lite_path_quad_to(dest, pt1->x, pt1->y, pt2->x, pt2->y);
--- a/src/misc/lv_array.c
+++ b/src/misc/lv_array.c
@@ -45,63 +45,116 @@ void lv_array_init(lv_array_t * array, uint32_t capacity, uint32_t element_size)
     LV_ASSERT_MALLOC(array->data);
 }
 
-void lv_array_destroy(lv_array_t * array)
+void lv_array_deinit(lv_array_t * array)
 {
     if(array->data) {
         lv_free(array->data);
         array->data = NULL;
     }
+
+    array->size = 0;
+    array->capacity = 0;
 }
 
-void lv_array_copy(lv_array_t * target, const lv_array_t * array)
+void lv_array_copy(lv_array_t * target, const lv_array_t * source)
 {
-    if(lv_array_is_empty(array)) {
+    if(lv_array_is_empty(source)) {
         return;
     }
-    lv_array_destroy(target);
-    lv_array_init(target, array->capacity, array->element_size);
-    lv_memcpy(target->data, array->data, array->size * array->element_size);
-    target->size = array->size;
+    lv_array_deinit(target);
+    lv_array_init(target, source->capacity, source->element_size);
+    lv_memcpy(target->data, source->data, source->size * source->element_size);
+    target->size = source->size;
 }
 
-void lv_array_clear(lv_array_t * array)
+lv_result_t lv_array_remove(lv_array_t * array, uint32_t index)
 {
-    array->size = 0;
+    if(index >= array->size) {
+        return LV_RESULT_INVALID;
+    }
+
+    /*Shortcut*/
+    if(index == array->size - 1) {
+        lv_array_resize(array, array->size - 1);
+        return LV_RESULT_OK;
+    }
+
+    uint8_t * start = lv_array_at(array, index);
+    uint8_t * remaining = start + array->element_size;
+    uint32_t remaining_size = (array->size - index - 1) * array->element_size;
+    lv_memcpy(start, remaining, remaining_size);
+    lv_array_resize(array, array->size - 1);
+    return LV_RESULT_OK;
+}
+
+lv_result_t lv_array_erase(lv_array_t * array, uint32_t start, uint32_t end)
+{
+    if(end > array->size) {
+        end = array->size;
+    }
+
+    if(start >= end) {
+        return LV_RESULT_INVALID;
+    }
+
+    /*Shortcut*/
+    if(end == array->size) {
+        lv_array_resize(array, start);
+        return LV_RESULT_OK;
+    }
+
+    uint8_t * start_p = lv_array_at(array, start);
+    uint8_t * remaining = start_p + (end - start) * array->element_size;
+    uint32_t remaining_size = (array->size - end) * array->element_size;
+    lv_memcpy(start_p, remaining, remaining_size);
+    lv_array_resize(array, array->size - (end - start));
+    return LV_RESULT_OK;
 }
 
 void lv_array_resize(lv_array_t * array, uint32_t new_capacity)
 {
-    if(new_capacity > array->size) {
-        if(new_capacity > array->capacity) {
-            uint8_t * data = lv_malloc(new_capacity * array->element_size);
-            lv_memcpy(data, array->data, array->size * array->element_size);
-            lv_free(array->data);
-            array->data = data;
-            array->capacity = new_capacity;
-        }
-    }
-    else {
+    uint8_t * data = lv_realloc(array->data, new_capacity * array->element_size);
+    LV_ASSERT_NULL(data);
+    array->data = data;
+    array->capacity = new_capacity;
+    if(array->size > new_capacity) {
         array->size = new_capacity;
     }
 }
 
-bool lv_array_append(lv_array_t * array, const uint8_t * element)
+lv_result_t lv_array_concat(lv_array_t * array, const lv_array_t * other)
 {
-    if(array->size >= array->capacity) {
-        // array is full
-        return false;
+    LV_ASSERT_NULL(array->data);
+    uint32_t size = other->size;
+    if(array->size + size > array->capacity) {
+        /*array is full*/
+        return LV_RESULT_INVALID;
     }
 
-    LV_ASSERT_NULL(array->data);
-    uint8_t * data = (uint8_t *)(array->data + array->size * array->element_size);
-    lv_memcpy(data, element, array->element_size);
-    array->size++;
-    return true;
+    uint8_t * data = array->data + array->size * array->element_size;
+    lv_memcpy(data, other->data, array->element_size * size);
+    array->size += size;
+    return LV_RESULT_OK;
 }
 
-uint8_t * lv_array_get(const lv_array_t * array, uint32_t index)
+lv_result_t lv_array_push_back(lv_array_t * array, const void * element)
 {
-    if(index > (array->size - 1)) {
+    LV_ASSERT_NULL(array->data);
+
+    if(array->size == array->capacity) {
+        /*array is full*/
+        lv_array_resize(array, array->capacity + 1);
+    }
+
+    uint8_t * data = array->data + array->size * array->element_size;
+    lv_memcpy(data, element, array->element_size);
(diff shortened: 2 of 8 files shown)


src/misc/lv_array.h
```diff
@@ -20,7 +20,7 @@ extern "C" {
 /*********************
  *      DEFINES
  *********************/
-#define DEFAULT_CAPS 8
+#define LV_ARRAY_DEFAULT_CAPS 8
```
PGNetHun: Please use full name: `LV_ARRAY_DEFAULT_CAPACITY`

src/misc/lv_array.h
```diff
@@ -38,46 +38,159 @@ typedef struct {
  * GLOBAL PROTOTYPES
  **********************/
 
+/**
+ * Init. an array.
```
PGNetHun: no need for the dot after _init_:
Simply: _Init an array._

src/misc/lv_array.h
```diff
@@ -38,46 +38,159 @@ typedef struct {
...
+/**
+ * Resize the array to the given capacity.
+ * @note if the new capacity is smaller than the current size, the array will be truncated.
+ * @param array pointer to an `lv_array_t` variable
+ * @param new_capacity the new capacity of the array
+ */
+void lv_array_resize(lv_array_t * array, uint32_t new_capacity);
+
+/**
+ * Deinit. an array. Free the allocated memory.
```
PGNetHun: No need for the dots in the sentence.
Simply: _Deinit the array, and free the allocated memory._

src/misc/lv_array.h
```diff
@@ -38,46 +38,159 @@ typedef struct {
...
+bool lv_array_concat(lv_array_t * array, const lv_array_t * other);
 
-uint32_t lv_array_length(const lv_array_t * array);
+/**
+ * Push back element. Adds a new element to the end of the array.
+ * @param array pointer to an `lv_array_t` variable
+ * @param element pointer to the element to add
+ * @return true: success; false: error
+ */
+bool lv_array_push_back(lv_array_t * array, const void * element);
```
PGNetHun: To follow the naming convention of other programming languages, I think the best name would be:
- `lv_array_append()`
- `lv_array_add()`

I prefer the first one, because it really means to add an element to the end of the array

src/misc/lv_array.h
```diff
@@ -38,46 +38,159 @@ typedef struct {
...
 void lv_array_clear(lv_array_t * array);
 
-void lv_array_resize(lv_array_t * array, uint32_t new_capacity);
+/**
+ * Remove the element at the specified position in the array.
+ * @param array pointer to an `lv_array_t` variable
+ * @param index the index of the element to remove
+ * @return true: success; false: error
+ */
+bool lv_array_pop(lv_array_t * array, uint32_t index);
```
PGNetHun: In other languages `pop` means to **get and remove** an item from a list (or stack), but in this case we just remove the item at the given index.

So I think the function name should reflect that: `lv_array_remove(lv_array_t * array, uint32_t index)`

src/misc/lv_array.h
```diff
@@ -38,46 +38,159 @@ typedef struct {
...
+/**
+ * Remove from the array either a single element or a range of elements ([start, end)).
+ * @note This effectively reduces the container size by the number of elements removed.
+ * @note When start equals to end, the function has no effect.
+ * @param array pointer to an `lv_array_t` variable
+ * @param start the index of the first element to be removed
+ * @param end the index of the first element that is not to be removed
+ * @return true: success; false: error
+ */
+bool lv_array_erase(lv_array_t * array, uint32_t start, uint32_t end);
```
PGNetHun: To align to previos function to remove item, I think this function name could be: 
`lv_array_remove_range(lv_array_t * array, uint32_t start, uint32_t end)`

src/misc/lv_array.h
```diff
@@ -38,46 +38,159 @@ typedef struct {
...
+ * @return a pointer to the requested element, NULL if `index` is out of range
+ */
+void * lv_array_at(const lv_array_t * array, uint32_t index);
+
+/**
+ * Returns a pointer to the first element in the array.
+ * @param array pointer to an `lv_array_t` variable
+ * @return a pointer to the first element in the array
+ */
+static inline void * lv_array_front(const lv_array_t * array)
```
PGNetHun, line 171: Maybe rename to `lv_array_first()`

src/misc/lv_array.h
```diff
@@ -38,46 +38,159 @@ typedef struct {
...
+{
+    return lv_array_at(array, 0);
+}
+
+/**
+ * Returns a pointer to the last element in the array.
+ * @param array pointer to an `lv_array_t` variable
+ */
+static inline void * lv_array_back(const lv_array_t * array)
```
PGNetHun, line 180: Maybe rename to `lv_array_last()`

PGNetHun (changes requested):
Some comments.

These 2 functions are not tested:
- `lv_array_pop`
- `lv_array_erase`

I think some function names could be renamed, too, to align the naming convention to other programming languages.

PS:
I suggest to hold back this PR (and related https://github.com/lvgl/lvgl/pull/5301) till release of LVGL v9, and merge only after it, because I think we should not do "big changes" till release, only bugfixes (A.K.A. code freeze), so LVGL can be stabilized.

src/misc/lv_array.c
```diff
@@ -45,71 +45,143 @@ void lv_array_init(lv_array_t * array, uint32_t capacity, uint32_t element_size)
...
     lv_memcpy(data, element, array->element_size);
     array->size++;
     return true;
 }
 
-uint8_t * lv_array_get(const lv_array_t * array, uint32_t index)
+void * lv_array_at(const lv_array_t * array, uint32_t index)
 {
-    if(index > (array->size - 1)) {
+    if(index >= array->capacity) {
```
PGNetHun: it should be: `array->size`
or do you want to allow devs to access unused elements in the array? (indexes between `size` and `capacity`)

using `array->capacity` would lead to a "bug":
if an array has 4 elements, and the capacity is 8, then the following call would allow to assign value at an unused index, and the size of array still remains 4, so assigned element would be still outside of used range:
`lv_array_assign(array, 5, "value");`

PGNetHun (changes requested):

src/misc/lv_array.h
XuNeo: This is also a standard STL function.
https://cplusplus.com/reference/vector/vector/erase/

src/misc/lv_array.h
XuNeo: I'm following C++ STL, thought it's more popular to use. The original name is actually `_append`.
What do you think? Should we follow certain standard?

PGNetHun: Hmmm, I see.
Then keep STL naming.

src/misc/lv_array.h
PGNetHun: Yes, then keep it.

src/misc/lv_array.c
```diff
@@ -45,71 +45,143 @@ void lv_array_init(lv_array_t * array, uint32_t capacity, uint32_t element_size)
...
+    array->size += size;
+    return true;
+}
+
+bool lv_array_push_back(lv_array_t * array, const void * element)
+{
     LV_ASSERT_NULL(array->data);
-    uint8_t * data = (uint8_t *)(array->data + array->size * array->element_size);
+
+    if(array->size == array->capacity) {
```
PGNetHun, line 144: I think this push back function should increase automatically the array if capacity is reached, just like it is mentioned in STL vector function: 
https://cplusplus.com/reference/vector/vector/push_back/

So, then in other [PR](https://github.com/lvgl/lvgl/pull/5301/files#diff-52ff566468d87f6016ee458a4a89dffd561e014d31b1d16af51fd752570dd3afR99) you don't need to increase array capacity before calling push_back function.

PGNetHun (changes requested):

src/misc/lv_array.c
XuNeo: I was thinking the same, but in reality where array is used by lv_vector path, it normally doubles array size instead of increase one by one. So in the end, all APIs won't increase the array capacity automatically.

src/misc/lv_array.c
```diff
@@ -45,71 +45,143 @@ void lv_array_init(lv_array_t * array, uint32_t capacity, uint32_t element_size)
...
 {
     array->size = 0;
 }
 
-void lv_array_resize(lv_array_t * array, uint32_t new_capacity)
+/**
+ * Remove the element at the specified position in the array.
+ * @param array pointer to an `lv_array_t` variable
+ * @param index the index of the element to remove
+ */
```
PGNetHun: Missing return comment

src/misc/lv_array.c
```diff
@@ -45,71 +45,143 @@ void lv_array_init(lv_array_t * array, uint32_t capacity, uint32_t element_size)
...
+bool lv_array_remove(lv_array_t * array, uint32_t index)
+{
+    if(index >= array->size) {
+        return false;
+    }
+
+    /*Shortcut*/
+    if(index == array->size - 1) {
+        lv_array_resize(array, array->size - 1);
+        return false;
```
PGNetHun: In case the last item of an array is removed we can return `true`, because it succeeded.

But should we really resize the array after every remove? Or just after a given threshold (capacity - size >= THRESHOLD).

PS: there is a potential bug when calling `lv_array_remove(array, 0)`, when it had only 1 item, see [my comment](https://github.com/lvgl/lvgl/pull/5300/files#r1451530900)

src/misc/lv_array.c
```diff
@@ -45,71 +45,143 @@ void lv_array_init(lv_array_t * array, uint32_t capacity, uint32_t element_size)
...
+    uint32_t remaining_size = (array->size - end) * array->element_size;
+    lv_memcpy(start_p, remaining, remaining_size);
+    lv_array_resize(array, array->size - (end - start));
+    return true;
+}
+
+void lv_array_resize(lv_array_t * array, uint32_t new_capacity)
+{
+    uint8_t * data = lv_realloc(array->data, new_capacity * array->element_size);
+    LV_ASSERT_NULL(data);
```
PGNetHun, line 117: In case `lv_array_remove(array, 0)` is called (removing last item in array), it invokes `lv_array_resize(array, 0)`, and so `lv_realloc()` will return `NULL`, and this function will fail.
Please correct me if I'm wrong.
There could be also a test case for adding an item to an array with capacity=1, then removing that item.

PGNetHun (changes requested):
Please when you have finished fixes, or answered all feedback comments, re-request review.

src/misc/lv_array.c
PGNetHun: Sorry, in case `lv_realloc()` is called with size 0, then in actual LVGL implementation it will return the address of `memory_zero` and not `NULL` (but maybe I'm wrong):
1. https://github.com/lvgl/lvgl/blob/master/src/stdlib/lv_mem.c#L131
2. https://github.com/lvgl/lvgl/blob/master/src/stdlib/lv_mem.c#L26
3. https://github.com/lvgl/lvgl/blob/master/src/core/lv_global.h#L85

-> `array->data` will point to `memory_zero`.
Or am I wrong?

PGNetHun (changes requested):

src/misc/lv_array.h
kisvegabor: "append" sounds better to me too, but let's make it consistent with STL.

src/misc/lv_array.h
```diff
@@ -38,46 +38,151 @@ typedef struct {
...
-#define LV_ARRAY_APPEND(array, element) lv_array_append((array), (uint8_t*)(element))
-
-#define LV_ARRAY_GET(array, index, type) (type*)lv_array_get((array), (index))
-
-#define LV_ARRAY_SET(array, index, data, type) \
-    do { \
-        type* elem = (type*)lv_array_get((array), (index)); \
-        *elem = *((type*)(data)); \
-    } while(0)
+#define LV_ARRAY_APPEND_VALUE(array, value) lv_array_push_back((array), &(value))
```
kisvegabor: Do we really need it?

src/misc/lv_array.c
kisvegabor: >  in case lv_realloc() is called with size 0, then in actual LVGL implementation it will return the address of memory_zero and not NULL

I confirm it. It's because allocating e.g. a zero sized array means that the pointer is used and initialized but it happened to have 0 size. While NULL means an error, not set, etc. Similarly to Python or JS:
```py
>>> a = []
>>> a
[]
>>> b = None
>>> b
>>> 
```

src/misc/lv_array.c
kisvegabor: I think both are valid use cases:
1. Having more control on how to update the capacity
2. Having a simple way to just append a value

 Can we have an other function like `lv_array_push_back_with_resize` ?

XuNeo:
Thank you guys for the feedback. I'll update them later when get time.

src/misc/lv_array.c
XuNeo: I intend to do this way because I saw the original test case relies on this feature(with buggy code). I'm not sure if we should leave this backdoor open. I does can simplify some situations.

Note, 
```c
void test_array_set_get(void)
{
    uint32_t v = 100;
    lv_array_assign(&array, 0, &v);
    uint32_t * r = lv_array_at(&array, 0);
    TEST_ASSERT_EQUAL_UINT32(100, *r);
}
```

Original buggy code, note when size is zero, it won't return NULL.

```c
uint8_t * lv_array_get(const lv_array_t * array, uint32_t index)
{
    if(index > (array->size - 1)) {
        return NULL;
    }

    LV_ASSERT_NULL(array->data);
    return array->data + index * array->element_size;
}
```

src/misc/lv_array.c
XuNeo: I have changed `lv_array_push_back` to automatically resize with clear comment to state it.
For situations need better performance, a pre-resized array could be used.
For situations like lv_event, it can automatically resize.

src/misc/lv_array.c
XuNeo: It should return true. Fixed.

For performance, we can optimize them later with a THRESHOLD. I suspect it involves many APIs. Please open a new issue that we can follow up.

src/misc/lv_array.h
XuNeo: No need. Removed.

src/misc/lv_array.c
kisvegabor: Pre-resize is a good idea!

src/misc/lv_array.c
XuNeo: I have updated to use `array->size ` instead. Test cases updated too.
Returning pointer to non-existent element could be dangerous. For now, there's no such case need this feature.

Using `array->size` also matches with STL.

XuNeo:
`lv_result_t` should be used instead of `bool` to indicate operation fails or not. Updated.

kisvegabor (dismissed):

XuNeo:
Conflicts solved.

src/draw/vg_lite/lv_draw_vg_lite_vector.c
```diff
@@ -236,35 +236,35 @@ static void lv_path_to_vg(lv_vg_lite_path_t * dest, const lv_vector_path_t * src
...
                     CMP_BOUNDS(pt);
                     lv_vg_lite_path_line_to(dest, pt->x, pt->y);
                     pidx += 1;
                 }
                 break;
             case LV_VECTOR_PATH_OP_QUAD_TO: {
-                    const lv_fpoint_t * pt1 = LV_ARRAY_GET(&src->points, pidx, lv_fpoint_t);
-                    const lv_fpoint_t * pt2 = LV_ARRAY_GET(&src->points, pidx + 1, lv_fpoint_t);
+                    const lv_fpoint_t * pt1 = lv_array_at(&src->points, pidx);
+                    const lv_fpoint_t * pt2 = lv_array_at(&src->points, pidx);
```
PGNetHun: Fix to: `pidx + 1`

src/misc/lv_array.c
```diff
@@ -45,86 +45,128 @@ void lv_array_init(lv_array_t * array, uint32_t capacity, uint32_t element_size)
...
 
-bool lv_array_append(lv_array_t * array, const uint8_t * element)
+lv_result_t lv_array_concat(lv_array_t * array, const lv_array_t * other)
 {
-    if(array->size >= array->capacity) {
-        // array is full
-        return false;
+    LV_ASSERT_NULL(array->data);
+    uint32_t size = other->size;
+    if(array->size + size > array->capacity) {
```
PGNetHun, line 129: Maybe we should allow the array to auto resize.

I think whenever we use this `lv_array`, we don't want to care about whether it is full or not, we just add elements, concat another array, get element at given index, remove element, etc.... like in any other programming language, for example JavaScript or Python .... so what I want to say: we just want to use it like a "service", and don't want to pre-check the capacity every time we want to add a new element or concat another array. It's not the caller's responsibility.
I don't see any reason why we should reject array concat when the target is full.

Or is this behaviour of this concat function intentionally?

PS:
I see, that `lv_array_push_back()` automatically resizes the array.

PGNetHun (changes requested):

src/draw/vg_lite/lv_draw_vg_lite_vector.c
XuNeo: Thank you for your careful review, I didn't realize the code have changed. Nice catch!

src/misc/lv_array.c
XuNeo: The array module doesn't have auto-resize capability originally. But since push_back does in this way, concat should too. Updated.

As for the performance issue when removing element, it's a tradeoff between size and efficiency.
I cannot decide what to do yet in this PR. It would be better for a new issue to discuss.

PGNetHun: I think it is OK now, even removing element.
If there will be a performance issue, because of many push and removes (continous resize and memory allocation), then we could think on having a minimum capacity threshold, so if remove function is called, then it will not resize the array if size is smaller than the threshold.

XuNeo: I want to keep the  old API behavior same as the original one.
The newly introduced API does have auto resize now.

src/misc/lv_array.h
```diff
@@ -38,46 +40,161 @@ typedef struct {
...
+ */
+lv_result_t lv_array_erase(lv_array_t * array, uint32_t start, uint32_t end);
 
-bool lv_array_is_full(const lv_array_t * array);
-/**********************
- *      MACROS
- **********************/
+/**
+ * Concatenate two arrays. Adds new elements to the end of the array.
+ * @note The destination array must be large enough to hold the result.
```
PGNetHun: Because the array is auto-resized when concatenating, we don't need this `@note` anymore. 
Or change to: _The destination array is automatically resized to hold the new items._

src/misc/lv_array.c
PGNetHun: Great, thanks, it now works perfectly (as arrays in other prog.languages)
When you think it is done, and CI builds finished successfully, could you please re-request review? So I can mark our conversation comments resolved. Thank you!
image

PS: 
it is recommended to re-request review from **all** previous reviewers to review again the whole PR (or just recent commits).

XuNeo:
The comment of `lv_array_concat` has been updated to `The destination array is automatically expanded as necessary.`

PGNetHun (approved):

kisvegabor (approved):

kisvegabor:
Let's merge it now, and improve resizing on shrink if needed in a new PR.

PR #5301, opened 2024-01-12
feat(event): use lv_array for event descriptors

XuNeo:

Most logics can be eliminated by using lv_array.

### Checkpoints

### DIFF
--- a/src/misc/lv_event.c
+++ b/src/misc/lv_event.c
@@ -65,14 +65,15 @@ lv_result_t lv_event_send(lv_event_list_t * list, lv_event_t * e, bool preproces
     if(list == NULL) return LV_RESULT_OK;
 
     uint32_t i = 0;
-    for(i = 0; i < list->cnt; i++) {
-        if(list->dsc[i].cb == NULL) continue;
-        bool is_preprocessed = (list->dsc[i].filter & LV_EVENT_PREPROCESS) != 0;
+    lv_event_dsc_t * dsc = lv_array_at(list, 0);
+    for(i = 0; i < lv_array_size(list); i++) {
+        if(dsc[i].cb == NULL) continue;
+        bool is_preprocessed = (dsc[i].filter & LV_EVENT_PREPROCESS) != 0;
         if(is_preprocessed != preprocess) continue;
-        lv_event_code_t filter = list->dsc[i].filter & ~LV_EVENT_PREPROCESS;
+        lv_event_code_t filter = dsc[i].filter & ~LV_EVENT_PREPROCESS;
         if(filter == LV_EVENT_ALL || filter == e->code) {
-            e->user_data = list->dsc[i].user_data;
-            list->dsc[i].cb(e);
+            e->user_data = dsc[i].user_data;
+            dsc[i].cb(e);
             if(e->stop_processing) return LV_RESULT_OK;
 
             /*Stop if the object is deleted*/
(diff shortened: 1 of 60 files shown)


PGNetHun:
Just a remark:
this PR has to be merged **after** PR https://github.com/lvgl/lvgl/pull/5300

src/misc/lv_event.c
```diff
@@ -65,14 +65,15 @@ lv_result_t lv_event_send(lv_event_list_t * list, lv_event_t * e, bool preproces
     if(list == NULL) return LV_RESULT_OK;
 
     uint32_t i = 0;
-    for(i = 0; i < list->cnt; i++) {
-        if(list->dsc[i].cb == NULL) continue;
-        bool is_preprocessed = (list->dsc[i].filter & LV_EVENT_PREPROCESS) != 0;
+    lv_event_dsc_t * dsc = lv_array_at(list, 0);
```
PGNetHun, line 68: No need for this, because in the for loop we can access items (`lv_event_dsc_t`) directly using `lv_array_at(list, i)`.
Just predefine `dsc` variable to point to first element:
`lv_event_dsc_t *dsc = list->data;`

The other problem is that when array has no elements, and `lv_array_at()` is fixed to return `NULL` if there is no element at given position ([PR comment](https://github.com/lvgl/lvgl/pull/5300#discussion_r1451459360)), then  `dsc` would be `NULL`.

src/misc/lv_event.c
```diff
@@ -86,27 +87,29 @@ lv_result_t lv_event_send(lv_event_list_t * list, lv_event_t * e, bool preproces
...
+
+    lv_event_dsc_t dsc = { 0 };
+    dsc.cb = cb;
+    dsc.filter = filter;
+    dsc.user_data = user_data;
 
-    list->dsc[list->cnt - 1].cb = cb;
-    list->dsc[list->cnt - 1].filter = filter;
-    list->dsc[list->cnt - 1].user_data = user_data;
+    lv_array_resize(list, lv_array_size(list) + 1);
```
PGNetHun: At the moment there is a possible double memory allocation issue:
if `list` is empty:
1. then it is initialized to capacity 1 in line 91: `lv_array_init(list, 1, sizeof(lv_event_dsc_t));`
2. then here list is resized to 2: `lv_array_resize(list, lv_array_size(list) + 1);`

So, I think lv_array_push_back() function should increase the array automatically, and so if list is initialized with capacity 1 (and size=0), then there will be no double memory allocation.
https://github.com/lvgl/lvgl/pull/5300#discussion_r1451475640

PGNetHun (changes requested):

src/misc/lv_event.c
XuNeo: > for loop we can access items (lv_event_dsc_t) directly using lv_array_at(list, i)

Init dsc firstly using lv_array_at thus we can avoid calling lv_array_at in the loop, thus possibly better performance.

> lv_array_at() is fixed to return NULL

``for(i = 0; i < lv_array_size(list)``` make sure we don't access NULL.

src/misc/lv_event.c
XuNeo: Added `else` to eliminate the extra dsc memory.

kisvegabor:
Looks good, waiting for #5300

XuNeo:
Rebased on https://github.com/lvgl/lvgl/pull/5300 to run CI firstly.

PGNetHun (approved):
Wow, this is a great code simplification :)

kisvegabor (approved):

kisvegabor:
> Wow, this is a great code simplification :)

I agree :blush:

Issue #4926, opened 2023-12-04
Finalize v9


kisvegabor:
Continue #3298 (see for the changes too)

### Schedule
1. **December 4**: Feature stop, start updating the docs and testing
2. **December 18**: Release candidate version and call to test
3. **January 15**: Release v9.0

### 1. Features to finalize
We don't add any new features but finalize these ongoing ones 
- [ ] Image decoder fine tuning (@XuNeo )
- [ ] Online image converter (@W-Mai)
- [ ] Cache update (@W-Mai and @FASTSHIFT #4835 )
- [x] Scale needle handling (@lhdjply ? #4917)
- [ ] Pointer scroll support (@kisvegabor / @XuNeo #4909)
- [ ] Add ESP-LCD support (@zjanosy  #4507)
- [ ] Add support for at least 3 display controller (@zjanosy)
- [x] Remove the MicroPython examples and replace them with a test (@kdschlosser #4645)

### 2. Prepare for release candidate (by December 18)
Docs review
- [x] [Porting](https://docs.lvgl.io/master/porting/index.html) (@kisvegabor )
- [x] [Drivers](https://docs.lvgl.io/master/integration/index.html) ( :warning: Help wanted)
- [x] [Widgets](https://docs.lvgl.io/master/widgets/index.html) (@kisvegabor )
- [x] [Overview](https://docs.lvgl.io/master/overview/index.html) (especially the image part)  ( :warning: Help wanted)

Demos end examples
- [x] [Examples](https://docs.lvgl.io/master/examples.html) (@kisvegabor )
- [ ] Review and update demos (@kisvegabor )

Others
- [x] Migration guide (@kisvegabor)
- [x] **Call to test and announce release candidate**  (@kisvegabor )
 
### 3. Fixes and last touches (by January 15)
Mainly for January
- [ ] CI update with more config options (@XuNeo and @kisvegabor)
- [ ] Final NXP drivers (@nicusorcitu)
- [ ] Arm2D support? (@GorgonMeducer )
- [ ] Project update (E.g. Eclipse, VSCode, etc)  ( :warning:  Help wanted)
- [x] Changelog (@kisvegabor)
- [ ] **Release** (@kisvegabor )

nicusorcitu:
Is screen rotation still a task for v9?

GorgonMeducer:
OK, I will update the cmsis-pack by end of this week.
In the coming two weeks, I will finish the porting for Helium.

[...]

PR #5323, opened 2024-01-15
chore(cmsis-pack): prepare for LVGL9 release

GorgonMeducer:

- Creating cmsis-pack for LVGL9 release.
- Other minor updates

### Checkpoints


GorgonMeducer:
@kisvegabor 
I have created dedicated header files for arm-2d and included them in `lv_draw_sw_helium.h` and `lv_blend_helium.h`.
Could this relief your concerns?

README.md
```diff
@@ -1,6 +1,3 @@
-**IMPORTANT NOTE** The next major version (v9.0.0) is developed in the master branch.
-The last stable version is available in the [release/v8.3](https://github.com/lvgl/lvgl/tree/release/v8.3) branch.
-
 ---
```
kisvegabor:
```suggestion

```

kisvegabor (commented):
I think the root of my confusion with Helium/Arm2D is that
1. Now we have Arm2D/Helium to accelerate SW render via intrinsics: I think it's not helium because Arm2D can use anything under the hood. That is we could use the current Arm2D intrinsics on a not-helium MCU too, right?
2. We can have an Arm2D draw unit: I'm not sure if it's planned/needed but if so, it would be also Arm2D.

So I wonder why do we needed the world "helium" in connection with Arm2D. Please let me know if I'm missing something?

GorgonMeducer:
For the concerns you have:
> I think it's not helium because Arm2D can use anything under the hood. That is we could use the current Arm2D intrinsics on a not-helium MCU too, right?

In theory, yes, but I have added macro to ensure that **only when Helium is available in a device, the arm-2d intrinsics would be used.** 

For now, there is no reason to enable arm-2d acceleration for non-helium MCU. It gains no benefits at all.

> We can have an Arm2D draw unit: I'm not sure if it's planned/needed but if so, it would be also Arm2D.

Yes, it is planned. Let me explain it in detail:

- Arm-2D provides acceleration for both Synchronose acceleration and Asyncrhonouse acceleration
    * Helium and Helium-based ACI (Arm Custom Instruction Set) are synchronous acceleration. I use this synchronous acceleration in the draw-sw for blend and transform.
    * Arm also provides an IP called DMA350 which can be used as a skeleton of a 2D-GPU that silicon vendors can expand its 2D processing capability with their extensions. As a 2D-GPU, it offloads CPUs. The Arm-2D Draw Unit is used as the driver for this 2D-GPU. 

As you can see, the one accelerating draw-sw and the Arm-2D Draw Unit play different roles and use different aspects of arm-2d to accelerate LVGL9 in any possible way.

We plan to add the support for DMAC-350 this year and once it is ready, we will implement the Arm-2D DrawUnit and enable the DMAC-350 as a 2D GPU for LVGL.

I hope this answers you questions.

kisvegabor:
Thank you for the clarification. The sync-async part is clear too. 

> In theory, yes, but I have added macro to ensure that only when Helium is available in a device, the arm-2d intrinsics would be used.
For now, there is no reason to enable arm-2d acceleration for non-helium MCU. It gains no benefits at all.

I feel like it would be more future proof the call the Sync version Arm2D as it might happen that it will support other magics besides Helium. What do you think?

GorgonMeducer:
> I wonder why do we needed the world "helium" in connection with Arm2D

For draw-sw, I believe the **only valuable part of arm-2d is the helium acceleration**. I have and will continue to make sure that **only** the helium part is used.

Further more, I would like to use the arm-2d helium part as the default helium implementation of the LVGL intrinsics. This means:

- People can choose some of the LVGL intrinsics and add their own helium implementation
    * no need to remove the arm-2d
    * Just define the intrinsics macros (directly or via `LV_DRAW_SW_HELIUM_CUSTOM_INCLUDE `)

GorgonMeducer:
> I feel like it would be more future proof the call the Sync version Arm2D as it might happen that it will support other magics besides Helium. What do you think?

No problem at all. What do you suggest to do?

kisvegabor:
Can/will Arm2D support Neon or general non-SIMD assembly intrinsics in the future? I so, I wouldn't mention helium at all in folder names and config defines and just call it Arm2D.

GorgonMeducer:
>Can/will Arm2D support Neon or general non-SIMD assembly intrinsics in the future?

- Arm-2D has the plan to add NEON support, but it won't happen this year.

- Arm-2D defines intrincis in low-level, and we use helium to implement those intrinsics.  In other words, arm-2d supports non-SIMD assembly intrinsics in theory and you can use any way to implement those intrinsics in arm-2d. 

I am dividing the macro `LV_USE_DRAW_ARM2D` into two separated ones, i.e. `LV_USE_DRAW_ARM2D_SYNC` and `LV_USE_DRAW_ARM2D_ASYNC`.

lvgl.h
```diff
@@ -16,7 +16,7 @@ extern "C" {
 #define LVGL_VERSION_MAJOR 9
 #define LVGL_VERSION_MINOR 0
 #define LVGL_VERSION_PATCH 0
-#define LVGL_VERSION_INFO "dev"
```
kisvegabor, line 19: In line with [this comment](https://github.com/lvgl/lvgl/issues/4926#issuecomment-1891789674) please use the `"rc"` tag for now.

GorgonMeducer: Done

kisvegabor:
> I am dividing the macro LV_USE_DRAW_ARM2D into two separated ones, i.e. LV_USE_DRAW_ARM2D_SYNC and LV_USE_DRAW_ARM2D_ASYNC.

I think it better shows the what is what. The last commit looks good. I think the last missing thing is to rename helium folders to arm2d.

GorgonMeducer:
OK, no problem. 
I will make the change.

GorgonMeducer:
@kisvegabor Done. 
Please check.

lv_conf_template.h
```diff
@@ -133,8 +133,8 @@
     #endif
 #endif
 
-/* Use Arm-2D on Cortex-M based devices. Please only enable it for Helium Powered devices for now */
-#define LV_USE_DRAW_ARM2D 0
+/* Use Arm-2D to accelerate the sw render */
+#define LV_USE_DRAW_ARM2D_SYNC 0
```
kisvegabor: If it's used only with the SW render please move it to  e.g. line 103.

kisvegabor (commented):
What happens if `LV_USE_DRAW_ARM2D_SYNC` is enabled but `LV_DRAW_SW_ASM_HELIUM` is not?

GorgonMeducer:
No Helium acceleration would be added to draw-sw

lv_conf_template.h
GorgonMeducer: OK, no problem

library.properties
```diff
@@ -1,5 +1,5 @@
 name=lvgl
-version=9.0.0-dev
+version=9.0.0
```
kisvegabor:
```suggestion
version=9.0.0-rc
```

library.json
```diff
@@ -1,6 +1,6 @@
 {
 	"name": "lvgl",
-	"version": "9.0.0-dev",
+	"version": "9.0.0",
```
kisvegabor:
```suggestion
	"version": "9.0.0-rc",
```

lv_conf_template.h
```diff
@@ -1,6 +1,6 @@
 /**
  * @file lv_conf.h
- * Configuration file for v9.0.0-dev
+ * Configuration file for v9.0.0
```
```suggestion
 * Configuration file for v9.0.0-rc
```

kisvegabor (commented):
Thank you, Gabriel. 

I think it's much cleaner now. Just the last touch with the "-rc" tags.

GorgonMeducer:
You can merge it now as it is updated to the latest changes.

kisvegabor (approved):

kisvegabor:
Thank you! I'm writing the news letter now :slightly_smiling_face:

PR #5355, opened 2024-01-17
feat(draw): add lv_draw_sw_rgb565_to_l1 for simpler conversation to monochrome

kisvegabor:

Related to https://github.com/lvgl/lvgl/issues/4926#issuecomment-1893601117

@espzav
Please this function. I'm not sure if the bit order is correct for real display controllers. 

### Checkpoints

### DIFF
--- a/src/draw/sw/lv_draw_sw.c
+++ b/src/draw/sw/lv_draw_sw.c
@@ -227,15 +227,17 @@ void lv_draw_sw_rgb565_to_l1(void * buf, uint32_t buf_size_px)
     lv_color16_t * in = buf;
     uint32_t i;
     for(i = 0; i < buf_size_px; i += 8) {
-        *out = 0;
-        if(in[0].blue > 16)(*out) |= (1 << 7);
-        if(in[1].blue > 16)(*out) |= (1 << 6);
-        if(in[2].blue > 16)(*out) |= (1 << 5);
-        if(in[3].blue > 16)(*out) |= (1 << 4);
-        if(in[4].blue > 16)(*out) |= (1 << 3);
-        if(in[5].blue > 16)(*out) |= (1 << 2);
-        if(in[6].blue > 16)(*out) |= (1 << 1);
-        if(in[7].blue > 16)(*out) |= (1 << 0);
+        uint8_t tmp = 0;
+        if(in[0].blue > 16)tmp |= (1 << 7);
+        if(in[1].blue > 16)tmp |= (1 << 6);
+        if(in[2].blue > 16)tmp |= (1 << 5);
+        if(in[3].blue > 16)tmp |= (1 << 4);
+        if(in[4].blue > 16)tmp |= (1 << 3);
+        if(in[5].blue > 16)tmp |= (1 << 2);
+        if(in[6].blue > 16)tmp |= (1 << 1);
+        if(in[7].blue > 16)tmp |= (1 << 0);
+
+        *out = tmp;
 
         in += 8;
         out += 1;
--- a/src/draw/sw/lv_draw_sw.c
+++ b/src/draw/sw/lv_draw_sw.c
@@ -216,6 +216,32 @@ void lv_draw_sw_rgb565_swap(void * buf, uint32_t buf_size_px)
 
 }
 
+void lv_draw_sw_rgb565_to_l1(void * buf, uint32_t buf_size_px)
+{
+    if(buf_size_px & 0x7) {
+        LV_LOG_WARN("buf_size_px must be multiple of 8");
+        return;
+    }
+
+    uint8_t * out = buf;
+    lv_color16_t * in = buf;
+    uint32_t i;
+    for(i = 0; i < buf_size_px; i += 8) {
+        *out = 0;
+        if(in[0].blue > 16)(*out) |= (1 << 7);
+        if(in[1].blue > 16)(*out) |= (1 << 6);
+        if(in[2].blue > 16)(*out) |= (1 << 5);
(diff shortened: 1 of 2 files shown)


espzav:
@kisvegabor Thank you very much for this function! I am working on monochromatic screen now, but I have some issues. 

Not working `LV_DISPLAY_RENDER_MODE_FULL`. When I put it like here: `lv_display_set_buffers(disp, buf1, buf2, disp_cfg->buffer_size, LV_DISPLAY_RENDER_MODE_FULL);` I got:
```
0x400dc5e5: lv_display_set_buffers at C:/WORK/GIT/esp-bsp/components/esp_lvgl_port/examples/i2c_oled/components/lvgl__lvgl/src/display/lv_display.c:411
```
When I do this:
```
lv_display_set_buffers(disp, buf1, buf2, disp_cfg->buffer_size, LV_DISPLAY_RENDER_MODE_PARTIAL);
lv_display_set_render_mode(disp, LV_DISPLAY_RENDER_MODE_FULL);
```
I got this:
```
0x400db855: layer_reshape_draw_buf at C:/WORK/GIT/esp-bsp/components/esp_lvgl_port/examples/i2c_oled/components/lvgl__lvgl/src/core/lv_refr.c:579 (discriminator 2)
```

When I use `LV_DISPLAY_RENDER_MODE_PARTIAL`, I see something, but not the text, which I should. 

I am working on it and I will inform you, when I find something new.

espzav:
> It's not supported for display color format. If `LV_COLOE_DEPTH` is `16` the default color format should be `lv_display_set_color_format(disp, LV_COLOR_FORMAT_RGB565);`

Ok, then I cannot use `LV_DISPLAY_RENDER_MODE_FULL` ? 

> Have you used the `LV_EVENT_INVALIDATE_AREA` to round the areas to 8 byte boundary?

No, I thought, that everything will be solved in `flush_cb`.

kisvegabor:
> Ok, then I cannot use LV_DISPLAY_RENDER_MODE_FULL

You can for sure! But I assume the display has multiples of 8 horizontal resolution to make it work nicely with bits. 

 > No, I thought, that everything will be solved in flush_cb.

If you are using FULL mode, no extra action needed . `LV_EVENT_INVALIDATE_AREA` is not required either as LVGL will refresh the the whole screen anyway.

XuNeo:
Shouldn't we use lv_draw_buf_t as parameter to take stride into consideration?

I have planned to use below API to support different color format. There's converting between ARGB8888 and RGB565 at least.

We can take stride into consideration in this way.

```c
 /* @note  `dest` and `src` should have same color format. Color converting is not supported fow now.
 */
void lv_draw_buf_copy(lv_draw_buf_t * dest, const lv_area_t * dest_area,
                      const lv_draw_buf_t * src, const lv_area_t * src_area);

```

kisvegabor:
Using `darw_buf` would require some boilerplate in `flush_cb` (Yeah, I know, if added the `draw_buf` parameter in `flush_cb` there weren't any boilerplates. )

I also feel like this converter function is too arbitrary. We should have a more generic solution. What about moving this function into the docs as an example code, and figure out a good solution in a separate  issue?

XuNeo:
Haaa,
Agree. Most users won't have stride headache anyway. Let's do it in simplest method.

kisvegabor:
It's great that it's working now! However I don't understand few parts in your functions. E.g. in `c += 16*ver_res` form where does 16 come from?

espzav:
> It's great that it's working now! However I don't understand few parts in your functions. E.g. in `c += 16*ver_res` form where does 16 come from?

Sorry, it was very fast implementation. The 16 is 8*2 (8 - move by 8 color pixels, 2 - bytes of one color pixel)

I solved rotation today and reworked it to this: 

```
static void _lvgl_port_transform_monochrome(lv_display_t *display, const lv_area_t *area, uint8_t *color_map)
{
    uint8_t *buf = color_map;
    lv_color16_t *color = (lv_color16_t *)color_map;
    uint16_t hor_res = lv_display_get_physical_horizontal_resolution(display);
    uint16_t ver_res = lv_display_get_physical_vertical_resolution(display);
    uint16_t res = hor_res;
    bool swap_xy = (lv_display_get_rotation(display) == LV_DISPLAY_ROTATION_90 || lv_display_get_rotation(display) == LV_DISPLAY_ROTATION_270);

    int x1 = area->x1;
    int x2 = area->x2;
    int y1 = area->y1;
    int y2 = area->y2;

    int out_x, out_y;
    for (int y = y1; y <= y2; y++) {
        for (int x = x1; x <= x2; x++) {
            bool chroma_color = (color[hor_res * y + x].blue > 16);

            if (swap_xy) {
                out_x = y;
                out_y = x;
                res = ver_res;
            } else {
                out_x = x;
                out_y = y;
                res = hor_res;
            }

            /* Write to the buffer as required for the display.
            * It writes only 1-bit for monochrome displays mapped vertically.*/
            buf = color_map + res * (out_y >> 3) + (out_x);
            if (chroma_color) {
                (*buf) &= ~(1 << (out_y % 8));
            } else {
                (*buf) |= (1 << (out_y % 8));
            }
        }
    }
}
```

I don't know, if it helps to you and if you want to add into LVGL. It is specific for monochromatic screens with drivers SSD1306 / SH1107 and similar.  I added into ESP LVGL port (https://github.com/espressif/esp-bsp/pull/268).

kisvegabor:
I think we should add L1 and L8 rendering instead of these helpers. We will work on it in the near future so I close this PR for now.

PR #5387, opened 2024-01-18
optinally install demos and libs as well

ghost:
At the moment only lvgl files are installed. If you would like to build demos and examples as well, you have to install these files as well.


kisvegabor:
Thank you! 

I fell like CMake gets complex enough to have it own file in the docs. Could you add a new `build` folder [here](https://github.com/lvgl/lvgl/tree/master/docs/integration) with `cmake.rst` and list the features that we have now?

Ju1He1:
> Thank you!
> 
> I fell like CMake gets complex enough to have it own file in the docs. Could you add a new `build` folder [here](https://github.com/lvgl/lvgl/tree/master/docs/integration) with `cmake.rst` and list the features that we have now?

Hey, I added some docs.

Locally I'm using CMakePresets to build this project since this makes my live much easier. :D
I would make the project much easier to build for users since CMakePresets are nicely supported with many IDEs. It basically "works out of the box" :)

But If this is not desired I can of course remove the CMakePresets file ;)

kisvegabor:
Thank you! I didn't know about CMakePreset so far, but it looks really great. 

However I'm no sure about `lv_conf_linux/win.h`. It adds a constant maintenance burden. What about updating the default `lv_conf_temlpate.h` like
```c
/*I don't what __WIN__ could be in reality */
#define LV_USE_FS_WIN32   __WIN__ ? 1 : 0  
...
#define LV_USE_WINDOWS  __WIN__ ? 1 : 0  
```

cc @W-Mai

W-Mai:
> Thank you! I didn't know about CMakePreset so far, but it looks really great.
> 
> However I'm no sure about `lv_conf_linux/win.h`. It adds a constant maintenance burden. What about updating the default `lv_conf_temlpate.h` like
> 
> ```c
> /*I don't what __WIN__ could be in reality */
> #define LV_USE_FS_WIN32   __WIN__ ? 1 : 0  
> ...
> #define LV_USE_WINDOWS  __WIN__ ? 1 : 0  
> ```
> 
> cc @W-Mai

I haven't used CMakePreset either, but it looks great!

I agree with this point. It is too expensive to maintain so many config files.😭

Ju1He1:
Alright. I Removed the Windows and Linux configs and adapted the presets.
It builds now locally with MSVC and in my WSL (Tested with GCC 12).

I decided to go with `MSC_VER` instead of `WIN32`.
I think `WIN32` is also set when you are crosscompiling on windows to some embedded target, because WIN32 only refers to the host machine, not the actual target.

However the first test is failing. 
@kisvegabor  Could you tell me why?

lv_conf_template.h
```diff
@@ -12,7 +12,7 @@
  */
 
 /* clang-format off */
-#if 0 /*Set it to "1" to enable content*/
+#if 1 /*Set it to "1" to enable content*/
```
kisvegabor: It should be `0` by default.

lv_conf_template.h
```diff
@@ -348,7 +348,10 @@
 #define LV_EXPORT_CONST_INT(int_value) struct _silence_gcc_warning /*The default value just prevents GCC warning*/
 
 /*Prefix all global extern data with this*/
-#define LV_ATTRIBUTE_EXTERN_DATA
+#ifndef LV_ATTRIBUTE_EXTERN_DATA
+  #define LV_ATTRIBUTE_EXTERN_DATA
+#endif // !LV_ATTRIBUTE_EXTERN_DATA
```
```suggestion
#endif /*LV_ATTRIBUTE_EXTERN_DATA*/
```

lv_conf_template.h
```diff
@@ -603,7 +606,7 @@
 #endif
 
 /*API for CreateFile, ReadFile, etc*/
-#define LV_USE_FS_WIN32 0
+#define LV_USE_FS_WIN32   _MSC_VER ? 1 : 0 
 #if LV_USE_FS_WIN32
     #define LV_FS_WIN32_LETTER '\0'     /*Set an upper cased letter on which the drive will accessible (e.g. 'A')*/
```
kisvegabor: We should also set a letter by default.
```suggestion
    #define LV_FS_WIN32_LETTER 'A'     /*Set an upper cased letter on which the drive will accessible (e.g. 'A')*/
```

kisvegabor:
If you click on
image

it tells
image

So basically the `lv_conf_internal.h` is not in sync with `lv_conf_template.h`.

lv_conf_template.h
ghost: what is the point of setting `"LV_CONF_PATH": "${sourceDir}/lv_conf_template.h"`.
But I gues we could set it back to 0 i gues :)

Ju1He1:
@kisvegabor  By commenting out the stuff in the config I'm getting lots of warnings
`"Possible failure to include lv_conf.h, please read the comment in this file if you get errors"`
Build seams to run though, but is this a desired bhevaiour?

src/lv_conf_internal.h
```diff
@@ -2027,15 +2030,15 @@
     #ifdef CONFIG_LV_USE_FS_WIN32
         #define LV_USE_FS_WIN32 CONFIG_LV_USE_FS_WIN32
     #else
-        #define LV_USE_FS_WIN32 0
+        #define LV_USE_FS_WIN32   _MSC_VER ? 1 : 0
```
W-Mai: This `_MSC_VER` is reporting so many warnings, can you try to fix it?
Like the existence of a code like this writing above?
```c
#ifndef _MSC_VER
    #define _MSC_VER 0
#endif
```

kisvegabor: We can add it to beginning of `lv_conf_internal.h` (generated by [lv_conf_internal_gen.py](https://github.com/lvgl/lvgl/blob/master/scripts/lv_conf_internal_gen.py))

kisvegabor:
> "Possible failure to include lv_conf.h, please read the comment in this file if you get errors"
Build seams to run though, but is this a desired bhevaiour?

I don't know you why you see it. Let's see if [this](https://github.com/lvgl/lvgl/pull/5387#discussion_r1462821778) changes something.

Ju1He1:
> > "Possible failure to include lv_conf.h, please read the comment in this file if you get errors"
> > Build seams to run though, but is this a desired bhevaiour?
> 
> I don't know you why you see it. Let's see if [this](https://github.com/lvgl/lvgl/pull/5387#discussion_r1462821778) changes something.

this is because of this

```
/* clang-format off */
#if 0 /*Set it to "1" to enable content*/

#ifndef LV_CONF_H
#define LV_CONF_H
```

We have two options:
1) move the header guard out of the #if 0 section
2) remove LV_CONF_PATH from preset. 

Which option would you prefer?

lv_conf_template.h
```diff
@@ -355,7 +355,10 @@
 #define LV_EXPORT_CONST_INT(int_value) struct _silence_gcc_warning /*The default value just prevents GCC warning*/
 
 /*Prefix all global extern data with this*/
-#define LV_ATTRIBUTE_EXTERN_DATA
+#ifndef LV_ATTRIBUTE_EXTERN_DATA
```
kisvegabor: Do we need the extra `ifndef` guard here?

ghost: yes because If invoked via cmake LV_ATTRIBUTE_EXTERN_DATA is passed by the command line.
Setting this define will overwrite the command line invocation and resulting in msvc linker errors.
Or do you have another suggestion?

Ju1He1:
> I would prefer
> 
> ```c
> /* clang-format off */
> 
> #ifndef LV_CONF_H
> #define LV_CONF_H
> 
> #if 0 /*Set it to "1" to enable content*/
> ```
> 
> however I don't see the root of the problem. It works for years like this.

the problem is that  lv_conf_internal checks for
` #if !defined(LV_CONF_H) `
this header guard is not defined if we set it to 0

Ju1He1:
> I would prefer
> 
> ```c
> /* clang-format off */
> 
> #ifndef LV_CONF_H
> #define LV_CONF_H
> 
> #if 0 /*Set it to "1" to enable content*/
> ```
> 
> however I don't see the root of the problem. It works for years like this.

sorry this does not work.
if you move it lots of declarations like LV_ATTRIBUTE_LARGE_RAM_ARRAY which are defined in lv_conf_template.h are missing. (I did regenerate the internal config)
So we must stick with this compiler warning or comment in the template.

Or do you have another suggestion ?

kisvegabor:
In your `lv_conf.h` it should be `#if 1`, but in the template it should be `#if 0`.

However , if you would like to not use `lv_conf.h`, `LV_CONF_SKIP` shall be defined in the compiler

src/lv_conf_internal.h
ghost: I did add this there

Ju1He1:
alright I added LV_CONF_SKIP to my preset. It builds on Windows and WSL (GCC 12)

Ju1He1:
@kisvegabor  looks like I broke something. Unfortunately I have no clue what :O
Can you help me figure out what?

kisvegabor:
As the compile issues are related to ThorVG, I assume the problem is somewhere [here](https://github.com/lvgl/lvgl/pull/5387/files#diff-e93911b6a5fdc8d825b6f11e0db30727559c34691c4687025a0f82b5c6a06189R144-R148).

Note that in the ThorVG test [these special options](https://github.com/lvgl/lvgl/blob/master/tests/src/lv_test_conf.h#L69-L85) are enabled.

cc @W-Mai @FASTSHIFT do you have any idea for the CI issue?

Ju1He1:
@kisvegabor  I have reverted the config changes. I don't need them anyway, I'm using my own config. For now this project builds with cmake on MSVC 17 :)

docs/building/cmake.rst
```diff
@@ -0,0 +1,75 @@
...
+cmake
+==========
+
+Overview
+********
+This project uses CMakePresets to ensure an easy build.  Find out more on Cmake Presets here:
+https://cmake.org/cmake/help/latest/manual/cmake-presets.7.html
+
+Prerequsites
+---------
```
kisvegabor: This decorations should have the same length as the tilte

kisvegabor (commented):
Thanks!
Just a minor  comment.

docs/building/cmake.rst
ghost: so like this?

kisvegabor: Almost :slightly_smiling_face: 

Let me show!

docs/building/cmake.rst
```diff
@@ -0,0 +1,75 @@
...
+
+- VS 2022
+- VS Code
+- CLion
+
+Simply load this project into your IDE and select your desired preset and you are good to go.
+
+
+Build with CMake GUI
+----------
```
kisvegabor:
```suggestion
Build with CMake GUI
--------------------
```

docs/building/cmake.rst
```diff
@@ -0,0 +1,75 @@
...
+All optional targets can be disabled by setting the proper cache variables.
+If you use cmake to install lvgl 3 folders will be created.
+
+- include/lvgl (contains all public headers)
+- bin (contains all binaries (\*.dll))
+- lib (contains all precompiled source files (\*.lib))
+
+
+Build with Command line
+----------
```
kisvegabor:
```suggestion
Build with Command line
-----------------------
```

kisvegabor (approved):

FASTSHIFT (approved):

PR #5389, opened 2024-01-19
fix(vector) : add path bounding and matrix transform functions. 

onecoolx:

### Checkpoints

### DIFF
--- a/src/draw/lv_draw_vector.c
+++ b/src/draw/lv_draw_vector.c
@@ -315,18 +315,17 @@ void lv_vector_path_close(lv_vector_path_t * path)
 void lv_vector_path_get_bounding(const lv_vector_path_t * path, lv_area_t * area)
 {
     uint32_t len = lv_array_size(&path->points);
-    lv_fpoint_t * p0 = lv_array_at(&path->points, 0);
-    float x1 = p0->x;
-    float x2 = p0->x;
-    float y1 = p0->y;
-    float y2 = p0->y;
+    lv_fpoint_t * p = lv_array_at(&path->points, 0);
+    float x1 = p[0].x;
+    float x2 = p[0].x;
+    float y1 = p[0].y;
+    float y2 = p[0].y;
 
     for(uint32_t i = 1; i < len; i++) {
-        lv_fpoint_t * p = lv_array_at(&path->points, i);
-        if(p->x < x1) x1 = p->x;
-        if(p->y < y1) y1 = p->y;
-        if(p->x > x2) x2 = p->x;
-        if(p->y > y2) y2 = p->y;
+        if(p[i].x < x1) x1 = p[i].x;
+        if(p[i].y < y1) y1 = p[i].y;
+        if(p[i].x > x2) x2 = p[i].x;
+        if(p[i].y > y2) y2 = p[i].y;
     }
 
     area->x1 = (int32_t)x1;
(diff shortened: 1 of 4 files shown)


src/draw/lv_draw_vector.c
```diff
@@ -295,6 +312,29 @@ void lv_vector_path_close(lv_vector_path_t * path)
...
+{
+    uint32_t len = lv_array_size(&path->points);
+    lv_fpoint_t * p0 = lv_array_at(&path->points, 0);
+    float x1 = p0->x;
+    float x2 = p0->x;
+    float y1 = p0->y;
+    float y2 = p0->y;
+
+    for(uint32_t i = 1; i < len; i++) {
+        lv_fpoint_t * p = lv_array_at(&path->points, i);
```
XuNeo: Instead of calling lv_arraya_at in the loop, which will always do sanity check, we can do this way since it's an actual array.
```c
    lv_fpoint_t * p = lv_array_at(&path->points, 0);
    for(uint32_t i = 1; i < len; i++) {
        p[i].x = 
```

We can add a helper similar to STL vector to get the first element.
`lv_array_front(&array);`

https://en.cppreference.com/w/cpp/container/vector/front

src/draw/lv_draw_vector.c
```diff
@@ -195,6 +195,23 @@ void lv_matrix_multiply(lv_matrix_t * matrix, const lv_matrix_t * m)
     _multiply_matrix(matrix, m);
 }
 
+void lv_matrix_transform_point(const lv_matrix_t * matrix, lv_fpoint_t * point)
```
kisvegabor, line 198:
```suggestion
void lv_point_transform_by_matrix(lv_fpoint_t * point, const lv_matrix_t * matrix)
```

src/draw/lv_draw_vector.c
```diff
@@ -195,6 +195,23 @@ void lv_matrix_multiply(lv_matrix_t * matrix, const lv_matrix_t * m)
...
+{
+    float x = point->x;
+    float y = point->y;
+
+    point->x = x * matrix->m[0][0] + y * matrix->m[1][0] + matrix->m[0][2];
+    point->y = x * matrix->m[0][1] + y * matrix->m[1][1] + matrix->m[1][2];
+}
+
+void lv_matrix_transform_path(const lv_matrix_t * matrix, lv_vector_path_t * path)
```
kisvegabor, line 207:
```suggestion
void lv_path_transform_by_matrix(lv_vector_path_t * path, const lv_matrix_t * matrix)
```

src/draw/lv_draw_vector.c
onecoolx: I think it's weird that there is a single function called 'lv_point_xxxx' here.

onecoolx: I think it classified as a method of matrix， how about that？

src/draw/lv_draw_vector.c
onecoolx: done

C47D:
Will we add tests for this?

W-Mai (approved):

src/draw/lv_draw_vector.c
kisvegabor: Okay, let's keep it as it is.

kisvegabor:
> Will we add tests for this?

Some tests would be really great!

onecoolx (commented):
add test cases

kisvegabor (approved):

kisvegabor:
Thanks!

PR #5398, opened 2024-01-19
perf(vg_lite): add asynchronous rendering support

FASTSHIFT:

Add VG-Lite flush commit trigger threshold. GPU will try to batch these many draw tasks. No need to enable independent rendering threads.
### Checkpoints

### DIFF
--- a/lv_conf_template.h
+++ b/lv_conf_template.h
@@ -176,6 +176,9 @@
 /* Enable VG-Lite assert. */
 #define LV_VG_LITE_USE_ASSERT 0
 
+/* VG-Lite flush commit trigger threshold. GPU will try to batch these many draw tasks. */
+#define LV_VG_LITE_FLUSH_MAX_COUNT 8
+
 #endif
 
 /*=======================
--- a/src/draw/vg_lite/lv_draw_vg_lite.c
+++ b/src/draw/vg_lite/lv_draw_vg_lite.c
@@ -68,6 +68,7 @@ void lv_draw_vg_lite_init(void)
     unit->base_unit.dispatch_cb = draw_dispatch;
     unit->base_unit.evaluate_cb = draw_evaluate;
     unit->base_unit.delete_cb = draw_delete;
+    lv_array_init(&unit->img_dsc_pending, 4, sizeof(lv_image_decoder_dsc_t));
 
     lv_vg_lite_path_init(unit);
 
@@ -134,21 +135,15 @@ static void draw_execute(lv_draw_vg_lite_unit_t * u)
             break;
     }
 
-#if LV_USE_PARALLEL_DRAW_DEBUG
-    /* Layers manage it for themselves. */
-    if(t->type != LV_DRAW_TASK_TYPE_LAYER) {
-    }
-#endif
-
-    LV_VG_LITE_CHECK_ERROR(vg_lite_finish());
+    lv_vg_lite_flush(draw_unit);
 }
 
 static int32_t draw_dispatch(lv_draw_unit_t * draw_unit, lv_layer_t * layer)
 {
-    lv_draw_vg_lite_unit_t * draw_vg_lite_unit = (lv_draw_vg_lite_unit_t *)draw_unit;
+    lv_draw_vg_lite_unit_t * u = (lv_draw_vg_lite_unit_t *)draw_unit;
 
     /* Return immediately if it's busy with draw task. */
-    if(draw_vg_lite_unit->task_act) {
+    if(u->task_act) {
         return 0;
     }
 
@@ -157,6 +152,7 @@ static int32_t draw_dispatch(lv_draw_unit_t * draw_unit, lv_layer_t * layer)
 
     /* Return 0 is no selection, some tasks can be supported by other units. */
     if(!t || t->preferred_draw_unit_id != VG_LITE_DRAW_UNIT_ID) {
+        lv_vg_lite_finish(draw_unit);
         return -1;
     }
 
@@ -171,16 +167,16 @@ static int32_t draw_dispatch(lv_draw_unit_t * draw_unit, lv_layer_t * layer)
     }
 
     t->state = LV_DRAW_TASK_STATE_IN_PROGRESS;
-    draw_vg_lite_unit->base_unit.target_layer = layer;
-    draw_vg_lite_unit->base_unit.clip_area = &t->clip_area;
-    draw_vg_lite_unit->task_act = t;
+    u->base_unit.target_layer = layer;
+    u->base_unit.clip_area = &t->clip_area;
+    u->task_act = t;
 
-    draw_execute(draw_vg_lite_unit);
+    draw_execute(u);
 
-    draw_vg_lite_unit->task_act->state = LV_DRAW_TASK_STATE_READY;
-    draw_vg_lite_unit->task_act = NULL;
+    u->task_act->state = LV_DRAW_TASK_STATE_READY;
+    u->task_act = NULL;
 
-    /* The draw unit is free now. Request a new dispatching as it can get a new task. */
+    /*The draw unit is free now. Request a new dispatching as it can get a new task*/
     lv_draw_dispatch_request();
 
     return 1;
@@ -217,6 +213,7 @@ static int32_t draw_evaluate(lv_draw_unit_t * draw_unit, lv_draw_task_t * task)
 static int32_t draw_delete(lv_draw_unit_t * draw_unit)
 {
     lv_draw_vg_lite_unit_t * unit = (lv_draw_vg_lite_unit_t *)draw_unit;
+    lv_array_deinit(&unit->img_dsc_pending);
     lv_vg_lite_path_deinit(unit);
     lv_vg_lite_decoder_deinit();
     return 1;
--- a/Kconfig
+++ b/Kconfig
@@ -355,6 +355,13 @@ menu "LVGL configuration"
 				default n
 				depends on LV_USE_DRAW_VG_LITE
 
+			config LV_VG_LITE_FLUSH_MAX_COUNT
+				int "VG-Lite flush commit trigger threshold."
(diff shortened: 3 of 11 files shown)


src/draw/vg_lite/lv_draw_vg_lite.c
```diff
@@ -69,6 +73,10 @@ void lv_draw_vg_lite_init(void)
     unit->base_unit.evaluate_cb = draw_evaluate;
     unit->base_unit.delete_cb = draw_delete;
 
+#if LV_USE_OS
+    LV_ASSERT(lv_thread_init(&unit->thread, LV_THREAD_PRIO_HIGH, render_thread_cb, 8 * 1024, unit) == LV_RES_OK);
```
W-Mai: 8 * 1024
Should this be configureable?

src/draw/vg_lite/lv_draw_vg_lite.c
```diff
@@ -134,21 +135,15 @@ static void draw_execute(lv_draw_vg_lite_unit_t * u)
...
     }
 
-#if LV_USE_PARALLEL_DRAW_DEBUG
-    /* Layers manage it for themselves. */
-    if(t->type != LV_DRAW_TASK_TYPE_LAYER) {
-    }
-#endif
-
-    LV_VG_LITE_CHECK_ERROR(vg_lite_finish());
+    lv_vg_lite_flush(draw_unit);
```
kisvegabor, line 138: What if `flush_count == 3` and there are no more draw tasks. What will trigger the flushing of these 3 comamnds?

lv_conf_template.h
```diff
@@ -176,6 +176,9 @@
 /* Enable VG-Lite assert. */
 #define LV_VG_LITE_USE_ASSERT 0
 
+/* VG-Lite flush commit trigger threshold */
```
```suggestion
/* VG-Lite flush commit trigger threshold. VGlLite will try to batch these many draw tasks */
```

- How can we determine the ideal value? Can it be 256 too? 
- How large the command buffer is?

src/draw/vg_lite/lv_draw_vg_lite.c
```diff
@@ -157,6 +152,7 @@ static int32_t draw_dispatch(lv_draw_unit_t * draw_unit, lv_layer_t * layer)
 
     /* Return 0 is no selection, some tasks can be supported by other units. */
     if(!t || t->preferred_draw_unit_id != VG_LITE_DRAW_UNIT_ID) {
+        lv_vg_lite_finish(draw_unit);
         return -1;
     }
```
FASTSHIFT, line 157: @kisvegabor 
If there are no more drawing tasks, it will wait for the GPU drawing to complete and clear flush_count.

lv_conf_template.h
FASTSHIFT: > How can we determine the ideal value? Can it be 256 too?

8 is an experience value.
We plan to implement this set of logic in the underlying driver in the future. Check whether the GPU is in idle state in `vg_lite_flush`: if the GPU is in idle, the rendering will be submitted immediately. If the GPU is busy, the rendering task will be queued to wait for the next time examine.
In LVGL, just simply call `vg_lite_flush`.

> How large the command buffer is?

Currently we configure the cmd buffer to be 64K x2. These two buffers are ping-pong buffers. When the GPU is reading one of the buffers, the user can write to the other buffer and swap at the appropriate time. The cmd buf length occupied by different drawing commands is different, ranging from tens to hundreds of bytes.

kisvegabor (approved):
I'll merge it after the release.

FASTSHIFT:
> I'll merge it after the release.

In fact, merge now also works, does not affect the main line function :)

src/draw/vg_lite/lv_draw_vg_lite.c
```diff
@@ -157,6 +152,7 @@ static int32_t draw_dispatch(lv_draw_unit_t * draw_unit, lv_layer_t * layer)
 
     /* Return 0 is no selection, some tasks can be supported by other units. */
     if(!t || t->preferred_draw_unit_id != VG_LITE_DRAW_UNIT_ID) {
+        lv_vg_lite_finish(draw_unit);
```
nicusorcitu, line 155: why to wait for GPU completion here?

src/draw/vg_lite/lv_draw_vg_lite.c
nicusorcitu, line 138: Calling vg_lite_flush() after each task does not necessary mean that command gets batched. Instead, after each vg_lite_flush() you will force the other command buffer to take over. As GPU have only 2 command buffers, if a 3th vg_lite_flush() comes too fast while the first vg_lite_flush() is not yet complete - e.g. the command buffer 1 is not complete - then the 3'th vg_lite_flush() will act as a vg_lite_finish().

This mechanism you implemented does not work as you intend to.
What is the improvement with this new code?

nicusorcitu: Remember that vg_lite_flush() will switch the two GPU command buffers.

src/draw/vg_lite/lv_draw_vg_lite.c
FASTSHIFT: When all the drawing tasks are finished, it will go here (usually about to be sent), and you can wait here for the GPU to finish the unrendered work.

src/draw/vg_lite/lv_draw_vg_lite.c
FASTSHIFT: This function is a double encapsulation of `vg_lite_flush`.
In `lv_vg_lite_flush` I added the count. When `lv_vg_lite_flush` is called more than a certain number of times, `vg_lite_flush` is called to tell the GPU to start rendering, so that the GPU can process several small tasks together to improve efficiency.

The next time `vg_lite_flush` is called, usually the GPU has already processed it and can immediately start a new rendering.

nicusorcitu: I understand now.

But you batch by force a number of LV_VG_LITE_FLUSH_MAX_COUNT tasks! Even the first time.

In my tests (running the benchmark v9) I found that GPU gets busy for a couple of 4-5 tasks only when it has to deal with filling the screen area - big task. Otherwise the GPU completes any task right away.

So the GPU gets IDLE most of the time and you do not have to batch anymore commands. This is why I suggest the solution I implemented here:
https://github.com/lvgl/lvgl/blob/3199231b39b267c91733225a1e521d2784ee74a8/src/draw/nxp/vglite/lv_vglite_utils.c#L80
In this way I am making sure the GPU is busy as much as possible. If the GPU is idle then makes no sense to queue even more commands.

src/draw/vg_lite/lv_draw_vg_lite.c
nicusorcitu: You can go here also if there are tasks which are not supported by VGlite but only by the CPU. So you do not have to wait for GPU completion.

src/draw/vg_lite/lv_draw_vg_lite.c
FASTSHIFT: What if we implement this logic inside `vg_lite_flush`? Let the driver determine the GPU status by itself. If the GPU is busy, it will join the queue. If it is idle, it will start rendering immediately.

src/draw/vg_lite/lv_draw_vg_lite.c
FASTSHIFT: We basically implement full GPU rendering, and the CPU basically does not need to participate in the rendering work.

src/draw/vg_lite/lv_draw_vg_lite.c
nicusorcitu: That is an API change behavior. In documentation it says it "submits the command to the GPU without waiting for completion". You can not just return because it might not be come another vg_lite_flush() to really submit it to the GPU.

There are other applications beside lvgl on top of vglite or in some cases customers can use vglite api directly to design the application. I would not ask for a change.

src/draw/vg_lite/lv_draw_vg_lite.c
nicusorcitu: If there is a single thread, then it does not really matter if everything can be drawn by vglite.
The main lvgl thread will call the vglite dispatcher. If the task is supported, will be passed to vglite. Else, the CPU dispatcher will get called and CPU will take care of it.

I will try to run your implementation on my side and see how it behaves.

src/draw/vg_lite/lv_draw_vg_lite.c
FASTSHIFT: Hmm...that makes sense. May I ask if you added the `VG_LITE_GPU_IDLE_STATE` enumeration in `vglite_run` yourself? Or is it already implemented in the SDK?

nicusorcitu: We asked for the API implementation to the VSI. You should see it as well in the next driver releases. I use 4.0.58 version.
```
vg_lite_get_parameter(VG_LITE_GPU_IDLE_STATE, 1, (vg_lite_pointer)&gpu_idle);
```

FASTSHIFT: Great! I'll switch to the same plan..

nicusorcitu: I would also let the VGLite thread implementation with the possibility to run it single thread (v8 like).
https://github.com/lvgl/lvgl/blob/152dc0b2beafa12e9030dc614d5bc3535e1e9c6a/src/draw/nxp/vglite/lv_draw_vglite.c#L113
And check all the places where the ```#if LV_USE_OS``` is used in this file. Simply setting the LV_USE_OS to LV_OS_NONE will make it run single thread (as you have it now).

PR #5404, opened 2024-01-19
feat(image_decoder): simplify `lv_image_decoder_get_info/open` workflow

W-Mai:

1. Simplify `lv_image_decoder_get_info/open` workflow
2. The logic of `open_cb` call failure has been modified. By default, we assume that the decoder when `get_info` is successful must be `open` successfully. If it fails, other decoders **should not** continue to decode, thus simplifying the process and accelerating the speed of failure.

### Notes

### DIFF
--- a/src/draw/lv_image_decoder.c
+++ b/src/draw/lv_image_decoder.c
@@ -97,6 +97,7 @@ lv_result_t lv_image_decoder_open(lv_image_decoder_dsc_t * dsc, const void * src
     lv_memzero(dsc, sizeof(lv_image_decoder_dsc_t));
 
     if(src == NULL) return LV_RESULT_INVALID;
+    dsc->src = src;
     dsc->src_type = lv_image_src_get_type(src);
 
 #if LV_CACHE_DEF_SIZE > 0
@@ -112,8 +113,7 @@ lv_result_t lv_image_decoder_open(lv_image_decoder_dsc_t * dsc, const void * src
     if(dsc->decoder == NULL) return LV_RESULT_INVALID;
 
     /*Duplicate the source if it's a file*/
-    if(dsc->src_type == LV_IMAGE_SRC_FILE) dsc->src = lv_strdup(src);
-    else dsc->src = src;
+    if(dsc->src_type == LV_IMAGE_SRC_FILE) dsc->src = lv_strdup(dsc->src);
 
     /*Make a copy of args*/
     dsc->args = args ? *args : (lv_image_decoder_args_t) {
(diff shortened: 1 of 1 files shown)


W-Mai:
this pr relies on other prs
merge order:
1. #5402  
2. #5405 
3. #5404

kisvegabor:
Do we need to merge these before releasing v9 on Monday? If not strictly required I'd merge them after the release.

W-Mai:
I feel like the changes aren’t that big 🫣

src/draw/lv_image_decoder.c
```diff
@@ -78,100 +86,52 @@ void _lv_image_decoder_deinit(void)
...
+    if(dsc->decoder == NULL) return LV_RESULT_INVALID;
 
-    lv_image_decoder_t * decoder;
+    /*Duplicate the source if it's a file*/
+    if(dsc->src_type == LV_IMAGE_SRC_FILE) dsc->src = lv_strdup(src);
+    else dsc->src = src;
 
-    static const lv_image_decoder_args_t def_args = {
+    /*Make a copy of args*/
+    dsc->args = args ? *args : (lv_image_decoder_args_t) {
```
XuNeo, line 119:
```c
    static const lv_image_decoder_args_t def_args = {
        .stride_align = LV_DRAW_BUF_STRIDE_ALIGN != 1,
        .premultiply = false,
        .no_cache = false,
        .use_indexed = false,
    };

    /*Make a copy of args */
    dsc->args = args ? *args : def_args;
```

See 
https://github.com/lvgl/lvgl/pull/4907
https://stackoverflow.com/questions/33270731/error-c4576-in-vs2015-enterprise/33271165#33271165

XuNeo (commented):
I agree with the change. If the decoder can get info, it must also can open it and should not fallback to other decoders.

One of the issue we met before is when the corresponding decoder fails to decode it will `fallback` to other decoders.

For example, GPU may need additional process to bin image by registering a new custom bin decoder fails due to various reasons, it should exit but not fall back to built-in bin decoder.

PGNetHun:
> Do we need to merge these before releasing v9 on Monday? If not strictly required I'd merge them after the release.

Yes I agree, it's risky to change anything in image decoding process right before release.
But we need this and the related PRs (especially https://github.com/lvgl/lvgl/pull/5407 ) for a better performance of image drawing.

PGNetHun (approved):
Overall OK.

src/draw/lv_image_decoder.c
kisvegabor: I think it should work if the CI passes. Let's update it if it causes issues somewhere.

kisvegabor (approved):

PR #5425, opened 2024-01-22
fix(draw): widen lv_img_header_t dimensions from 11 to 16 bits

niklasf:
Was observing snapshots larger than 2048px getting mysteriously truncated in LVGL 8. LVGL 9 already widened these:

https://github.com/lvgl/lvgl/blob/09cb87cdc6a0168a98bc0a3182a8439b13249ead/src/draw/lv_image_buf.h#L90-L91

### DIFF
--- a/src/draw/lv_img_buf.h
+++ b/src/draw/lv_img_buf.h
@@ -114,9 +114,9 @@ typedef uint8_t lv_img_cf_t;
 #if LV_BIG_ENDIAN_SYSTEM
 typedef struct {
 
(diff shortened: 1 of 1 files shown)


kisvegabor:
Thank you for reporting it. 
I think changing such a fundamental thing in v8 can be risky. Is it an option for you to update to v9?

niklasf:
Mhh ... yeah. I guess it's time to update in any case. I can use this as a private patch in the meantime.

PR #5433, opened 2024-01-23
fix: fix warnings with -flto on GCC 11

niklasf:
* Avoid type punning in kern pair binary search.
* Inline delete_core() into lv_anim_delete (used only once) and avoid invalid function pointer cast.
* %p formatter requires void pointer.
* Work around false-positive -Wmaybe-uninitialized wrt. lv_style_get_prop().

### DIFF
--- a/src/core/lv_obj_style.c
+++ b/src/core/lv_obj_style.c
@@ -932,7 +932,7 @@ static void trans_anim_cb(void * _tr, int32_t v)
                 break;
         }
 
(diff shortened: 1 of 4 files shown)


C47D (approved):
Let's others more familiar with the code base review the delete part, everything else looks good to me.

kisvegabor (approved):

kisvegabor:
Thank you!

PR #5445, opened 2024-01-23
Fix VA_ARGS in MSVC

ghost:
The latest lvgl Version does not compile with MSVC. See isssue
https://github.com/lvgl/lvgl/issues/5441 for more details.
This PR fixes this.
@kisvegabor  Could you create a new release, to fix the MSVC build?


FASTSHIFT (dismissed):

XuNeo:
Is this issue introduced in https://github.com/lvgl/lvgl/issues/5392?

Adding formated message is really useful. Could we solve MSVC issue like this?
```
#if defined(_MSC_VER)
#define MYLOG(fmt, ...)    LOG_HELPER("mylog: " fmt, __VA_ARGS__)
#else
#define MYLOG(fmt, ...)    LOG_HELPER("mylog: " fmt, ##__VA_ARGS__)
#endif
```

Reference: https://stackoverflow.com/questions/65013414/msvc-treatment-of-va-args-vs-other-compilers

CMakeLists.txt
```diff
@@ -33,4 +33,15 @@ if (MSVC)
...
+    GCC and Clang have added some (non std compliant) extensions to their preprocessor.
+    ##VA_ARGS__ is one of them. It is used in lvgl to support variadic macros.
+    if no argument is passed to the macro, the MSVC preprocessor will not remove the trailing comma.
+    This will result in a syntax error.
+    To fix this we need to enable the /Zc:preprocessor flag. This will make the MSVC preprocessor behave like the GCC/Clang preprocessor.
+    In C++20 and C23 __VA_OPT__ is introduced. Making this behaviour standard.
+    But for now we will stick with this preprocessor flag.
+  ]]
+  target_compile_options(lvgl PUBLIC /Zc:preprocessor)
 endif()
```
W-Mai:
```suggestion
endif()

```

need a blank line at the end of the file

W-Mai (changes requested):

FASTSHIFT:
> Is this issue introduced in #5392?
> 
> Adding formated message is really useful. Could we solve MSVC issue like this?
> 
> ```
> #if defined(_MSC_VER)
> #define MYLOG(fmt, ...)    LOG_HELPER("mylog: " fmt, __VA_ARGS__)
> #else
> #define MYLOG(fmt, ...)    LOG_HELPER("mylog: " fmt, ##__VA_ARGS__)
> #endif
> ```
> 
> Reference: https://stackoverflow.com/questions/65013414/msvc-treatment-of-va-args-vs-other-compilers

@MouriNaruto Which option do you prefer?

MouriNaruto:
@FASTSHIFT I think we should use the @XuNeo way to solve the issue.

Note: I have met the issue, but can be compiled when "/std:c17" is defined. But I think there are many people use old MSVC toolchain.

Kenji Mouri

MouriNaruto:
I need do some explainations.

According to https://learn.microsoft.com/en-us/cpp/build/reference/zc-preprocessor?view=msvc-170, we know "The /Zc:preprocessor option is available starting in Visual Studio 2019 version 16.5. An earlier, incomplete version of the new preprocessor option is available in versions of Visual Studio starting in Visual Studio 2017 version 15.8. For more information, see [/experimental:preprocessor](https://learn.microsoft.com/en-us/cpp/build/reference/experimental-preprocessor?view=msvc-170)."

But many MSVC toolchain users don't want to use conformance mode in some cases, especially for interoperate with the legacy projects.

Kenji Mouri

CMakeLists.txt
ghost: done

Ju1He1:
> > Is this issue introduced in #5392?
> > Adding formated message is really useful. Could we solve MSVC issue like this?
> > ```
> > #if defined(_MSC_VER)
> > #define MYLOG(fmt, ...)    LOG_HELPER("mylog: " fmt, __VA_ARGS__)
> > #else
> > #define MYLOG(fmt, ...)    LOG_HELPER("mylog: " fmt, ##__VA_ARGS__)
> > #endif
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Reference: https://stackoverflow.com/questions/65013414/msvc-treatment-of-va-args-vs-other-compilers
> 
> @MouriNaruto Which option do you prefer?

this will not work!
You need to remove the trailing "," if no Arguments are preset! This is only done with the compiler switch :)

MouriNaruto:
> @MouriNaruto We could read the MSVC version with [this](https://cmake.org/cmake/help/latest/variable/MSVC_VERSION.html)
> 
> However I do not see a point there. With the legacy mode lvglv9 will not compile. You would have to remove the whole formatting stuff just because some people use very old MSVC toolchains ;)

According to https://en.cppreference.com/w/c/preprocessor/replace, we know "Note: Some compilers offer an extension that allows ## to appear after a comma and before __VA_ARGS__, in which case the ## does nothing when __VA_ARGS__ is non-empty, but removes the comma when __VA_ARGS__ is empty: this makes it possible to define macros such as [fprintf](http://en.cppreference.com/w/c/io/fprintf) ([stderr](http://en.cppreference.com/w/c/io/std_streams), format, ##__VA_ARGS__)."

So, ##__VA_ARGS__ is actually an extension not the standard.

Kenji Mouri

Ju1He1:
> > @MouriNaruto We could read the MSVC version with [this](https://cmake.org/cmake/help/latest/variable/MSVC_VERSION.html)
> > However I do not see a point there. With the legacy mode lvglv9 will not compile. You would have to remove the whole formatting stuff just because some people use very old MSVC toolchains ;)
> 
> According to https://en.cppreference.com/w/c/preprocessor/replace, we know "Note: Some compilers offer an extension that allows ## to appear after a comma and before __VA_ARGS__, in which case the ## does nothing when __VA_ARGS__ is non-empty, but removes the comma when __VA_ARGS__ is empty: this makes it possible to define macros such as [fprintf](http://en.cppreference.com/w/c/io/fprintf) ([stderr](http://en.cppreference.com/w/c/io/std_streams), format, ##__VA_ARGS__)."
> 
> So, "##__VA_ARGS__" itself is actually an extension for some compilers like GCC and LLVM, not in the C99 standard.
> 
> Kenji Mouri

yes of course you are right ;) ##__VA_ARGS__ is not standard. GCC and clang implemented this, MSVC did their own stuff.
So with the traditional mode you will have to differentiate between 0 Args and > 0 Args. So you basically need 2 different functions. MS explained the comma elision there.
https://learn.microsoft.com/en-us/cpp/preprocessor/preprocessor-experimental-overview?view=msvc-160#comma-elision-in-variadic-macros

IMHO this is not worth the overhead. If someone would still like to use traditional (legacy) mode, we should disable formatting completely.

BTW: VA_OPT is standard and available from C23 and C++20

MouriNaruto:
> > > @MouriNaruto We could read the MSVC version with [this](https://cmake.org/cmake/help/latest/variable/MSVC_VERSION.html)我们可以用这个来阅读 MSVC 版本
> > > However I do not see a point there. With the legacy mode lvglv9 will not compile. You would have to remove the whole formatting stuff just because some people use very old MSVC toolchains ;)但是，我看不出有什么意义。使用传统模式时，lvglv9 将无法编译。您将不得不删除整个格式化内容，因为有些人使用非常旧的 MSVC 工具链;)
> > 
> > 
> > According to https://en.cppreference.com/w/c/preprocessor/replace, we know "Note: Some compilers offer an extension that allows ## to appear after a comma and before **VA_ARGS**, in which case the ## does nothing when **VA_ARGS** is non-empty, but removes the comma when **VA_ARGS** is empty: this makes it possible to define macros such as [fprintf](http://en.cppreference.com/w/c/io/fprintf) ([stderr](http://en.cppreference.com/w/c/io/std_streams), format, ##**VA_ARGS**)."根据 https://en.cppreference.com/w/c/preprocessor/replace，我们知道“注意：一些编译器提供了一个扩展，允许 ## 出现在逗号之后和VA_ARGS之前，在这种情况下，当 ## 不为空时VA_ARGS不执行任何操作，但在 VA_ARGS 为空时删除逗号：这使得定义宏成为可能，例如 fprintf （ stderr， format， ##VA_ARGS）。
> > So, "##**VA_ARGS**" itself is actually an extension for some compilers like GCC and LLVM, not in the C99 standard.因此，“##VA_ARGS”本身实际上是某些编译器（如 GCC 和 LLVM）的扩展，而不是 C99 标准中的扩展。
> > Kenji Mouri 毛利健二
> 
> yes of course you are right ;) ##**VA_ARGS** is not standard. GCC and clang implemented this, MSVC did their own stuff.是的，你当然是对的;)##VA_ARGS不是标准的。GCC 和 clang 实现了这一点，MSVC 做了自己的事情。 So with the traditional mode you will have to differentiate between 0 Args and > 0 Args. So you basically need 2 different functions. MS explained the comma elision there.因此，在传统模式下，您必须区分 0 Args 和 > 0 Args。所以你基本上需要 2 个不同的功能。MS 在那里解释了逗号省略。 https://learn.microsoft.com/en-us/cpp/preprocessor/preprocessor-experimental-overview?view=msvc-160#comma-elision-in-variadic-macros
> 
> IMHO this is not worth the overhead. If someone would still like to use traditional (legacy) mode, we should disable formatting completely.恕我直言，这不值得开销。如果有人仍然想使用传统（传统）模式，我们应该完全禁用格式化。
> 
> BTW: VA_OPT is standard and available from C23 and C++20顺便说一句：VA_OPT是标准配置，可从 C23 和 C++20 购买

But it's macro and the compiler will be expanded all at the compile time. Also, old LVGL support both format and old compilers.

Kenji Mouri

MouriNaruto:
Also, we shouldn't desire people use C23 at present. And LVGL is a project which needs to consider the pure C experience because the usage scenario.

Kenji Mouri

MouriNaruto:
> @MouriNaruto so what is your suggestion/solution?

In my opinion, maybe revert the commit change from https://github.com/lvgl/lvgl/commit/24115615bc49cc80602579ecfe33a0e200b82ffe is a good solution.

Kenji Mouri

Ju1He1:
> > @MouriNaruto so what is your suggestion/solution?
> 
> In my opinion, maybe revert commit [2411561](https://github.com/lvgl/lvgl/commit/24115615bc49cc80602579ecfe33a0e200b82ffe) changes is a good solution.
> 
> Kenji Mouri

at least optionally disable it. The extra information is helpfull ;)
Not sure how to do this though

Ju1He1:
@MouriNaruto  any PR from your side?

MouriNaruto:
https://github.com/lvgl/lvgl/pull/5453

Maybe this PR looks OK.

Kenji Mouri

Ju1He1:
> #5453
> 
> Maybe this PR looks OK.
> 
> Kenji Mouri

puh this is exactly what I meant. You have to create 2 different macros depending if you are passing 0 arguments or more than zero arguments. IMHO not very user friendly

MouriNaruto:
> @MouriNaruto any PR from your side?

I'm actually not care that because I use C17 mode in MSVC.

But I know have a solution for workaround is better.  But according to the reality, we should not ask people to use extra compiler command line options especially for pure C style projects. (Many people choose pure C style project for out-of-box and don't worry about something like the dependency)

Kenji Mouri

FASTSHIFT:
fixed: https://github.com/lvgl/lvgl/pull/5453, Let's use a more general solution to avoid having to :)

kisvegabor:
Closing as it should be solved by #5453

PR #4404, opened 2023-07-24
fix(attr): refactor LV_ATTRIBUTE_* for function attributes

fronders:
### Description of the feature or fix

Continue of #4391

removed function attributes from prototypes and kept only on function declarations or static inline functions:
- `LV_ATTRIBUTE_TICK_INC`
- `LV_ATTRIBUTE_TIMER_HANDLER`
- `LV_ATTRIBUTE_FLUSH_READY`
- `LV_ATTRIBUTE_FAST_MEM`

This fixes an issue in ESP-IDF where section attributes use  `__COUNTER__` macro, and this results in prototype and declaration having different sections assigned

Issue #4391, opened 2023-07-21
Setting LV_ATTRIBUTE_FAST_MEM spawns linker warnings on ESP-IDF v5.1


fronders:
### Perform all steps below and tick them with [x]
- [x] Read the [FAQ](https://docs.lvgl.io/master/intro/index.html#faq)
- [x] Check the related part of the [Documentation](https://docs.lvgl.io/)
- [x] Update lvgl to the latest version
- [ ] Reproduce the issue in a [Simulator](https://docs.lvgl.io/latest/en/html/get-started/pc-simulator.html)

### Describe the bug
I'm using ESP-IDF v5.1 (stable) with lvgl v8.3.8 (from component manager). When I enable `Set IRAM as LV_ATTRIBUTE_FAST_MEM` linker produces a bunch of warnings about conflicting iram sections like this:

```
Building C object esp-idf/lvgl__lvgl/CMakeFiles/__idf_lvgl__lvgl.dir/src/misc/lv_math.c.obj
../managed_components/lvgl__lvgl/src/misc/lv_math.c:49:1: warning: ignoring attribute 'section (".iram1.3")' because it conflicts with previous 'section (".iram1.0")' [-Wattributes]
   49 | {
      | ^
In file included from ../managed_components/lvgl__lvgl/src/misc/lv_math.c:9:
../managed_components/lvgl__lvgl/src/misc/lv_math.h:47:31: note: previous declaration here
   47 | LV_ATTRIBUTE_FAST_MEM int16_t lv_trigo_sin(int16_t angle);
      |                               ^~~~~~~~~~~~
```

You can clearly see different sections are assigned to function and to its prototype because `IRAM_ATTR` expands with `_SECTION_ATTR_IMPL(".iram1", __COUNTER__)` macro that uses `__COUNTER__`. This was not the case for v5.0 supposedly because v5.1 updated to GCC12.2.0 and this was ignored

[This issue comment](https://github.com/espressif/esp-idf/pull/10773#issuecomment-1432602363) mentions what can be a fix for this problem:

> solution we would like to have is to remove IRAM_ATTR from function declarations in the header file, except for the static inline definitions

And indeed all ESP-IDF sources follow that practice

---

Here is the full output of the linkage:

```
esptool.py v4.6.2
Creating esp32 image...
Merged 1 ELF section
Successfully created esp32 image.
Generated /home/fronders/esp/mp3-player/build/bootloader/bootloader.bin
[...]

PR #5467, opened 2024-01-24
Removed LV_ATTRIBUTE_FAST_MEM function attribute from prototypes and kept only on function declarations or static inline functions

ktand:

Removed LV_ATTRIBUTE_FAST_MEM function attribute from prototypes and kept only on function declarations and static inline functions. 

This fixes an issue in ESP-IDF where section attributes use __COUNTER__ macro, and this results in prototype and declaration having different sections assigned

This was previously fixed in the 8.x release.

See #4391 and #4404

### DIFF
--- a/src/draw/lv_draw_label.c
+++ b/src/draw/lv_draw_label.c
@@ -66,7 +66,7 @@ void lv_draw_glyph_dsc_init(lv_draw_glyph_dsc_t * dsc)
     lv_memzero(dsc, sizeof(lv_draw_glyph_dsc_t));
 }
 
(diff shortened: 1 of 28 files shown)


kisvegabor (approved):

kisvegabor:
Awesome, thank you! 

How much performance improvement do you see when `LV_ATTRIBUTE_FAST_MEM` is used?

ktand:
> Awesome, thank you!
> 
> How much performance improvement do you see when `LV_ATTRIBUTE_FAST_MEM` is used?

I have not measured it. Will get back when I have some numbers.

PR #5479, opened 2024-01-25
Fix accepted chars issue on big endian systems

### LINKED ISSUE #5454: Text area accepted chars not working on big endian system
### LVGL version

v8.3.4

### What happened?

When adding accepted chars to a textarea, no characters are accepted on my big endian system.

The reason is the following line in lv_textarea.c:
`uint32_t c_uni = _lv_txt_encoded_next((const char *)&c, NULL);`

The char pointer will here point to 0 and c_uni will be 0. To fix it I made the change shown below to lv_textarea.c, but I am not sure if this is the best solution. I basically swapped the byte order of the uint_32_t that is added in lv_textarea_add_char. What makes it a bit tricky is that if lv_textarea_add_char is called from lv_textarea_add_text, the byte order seems to be correct, so in that case we cannot swap the byte order.

```
uint32_t c2 = c;
#if LV_BIG_ENDIAN_SYSTEM
    if(c != 0) while(*letter_buf == 0) ++letter_buf;

    // The byte order may or may not need to be swapped here, since lv_textarea_add_text is ordering bytes correctly.
    // Assume swapping is needed if highest byte is zero. May not be foolproof.
    if ((c != 0) && ((c & 0xff000000) == 0) ) {
        c2 = ((c >> 24) & 0xff) | // move byte 3 to byte 0
            ((c << 8) & 0xff0000) | // move byte 1 to byte 2
            ((c >> 8) & 0xff00) | // move byte 2 to byte 1
            ((c << 24) & 0xff000000); // byte 0 to byte 3
    }
#endif

    lv_res_t res = insert_handler(obj, letter_buf);
    if(res != LV_RES_OK) return;

    uint32_t c_uni = _lv_txt_encoded_next((const char *)&c2, NULL);
```

### How to reproduce?

```lv_obj_t * ta = lv_textarea_create(lv_screen_active());
lv_textarea_set_accepted_chars(ta, "+-.01234567890");```

### PR DESCRIPTION
**Pull request by jimbom**
Fixes: #5454

When adding accepted chars to a textarea, no characters are accepted on big-endian systems.

In the function lv_textarea_add_char, the character to add is taken as a uint32_t parameter, called c. When encoded, this is casted to a char pointer. On big endian systems this pointer will point to the wrong byte in the uint32_t, and the wrong encoded character, c_uni, will be used when checking for valid characters.

The root of the issue is the following line:
`uint32_t c_uni = _lv_txt_encoded_next((const char *)&c, NULL);`

The reference to uint32_t c should not be casted to a char pointer. To change this, I believe require pretty big changes to the code, since _lv_txt_encoded_next requires a const char *.

The solution/workaround made here is to swap the byte order of, c in case of big-endian system. But if lv_textarea_add_char is called from lv_textarea_add_text, the byte order of c is already correct, so then the bytes order shall not be swapped. Here we assume that the byte order shall be swapped if the most significant byte of c is zero. This worked in my case but will not give the correct value of c_uni if c is a 4 byte code. However, since I believe accepted characters will always be a char, I think this will never be an issue.

### Notes

### DIFF
--- a/src/widgets/textarea/lv_textarea.c
+++ b/src/widgets/textarea/lv_textarea.c
@@ -121,7 +121,7 @@ void lv_textarea_add_char(lv_obj_t * obj, uint32_t c)
     if(c != 0) while(*letter_buf == 0) ++letter_buf;
 
     // The byte order may or may not need to be swapped here to get correct c_uni below,
-    // since lv_textarea_add_text is ordering bytes correctly bfore calling lv_textarea_add_char.
+    // since lv_textarea_add_text is ordering bytes correctly before calling lv_textarea_add_char.
     // Assume swapping is needed if MSB is zero. May not be foolproof.
     if((c != 0) && ((c & 0xff000000) == 0)) {
         c2 = ((c >> 24) & 0xff) | // move byte 3 to byte 0
--- a/src/widgets/textarea/lv_textarea.c
+++ b/src/widgets/textarea/lv_textarea.c
@@ -116,14 +116,25 @@ void lv_textarea_add_char(lv_obj_t * obj, uint32_t c)
 
     const char * letter_buf = (char *)&u32_buf;
 
+    uint32_t c2 = c;
 #if LV_BIG_ENDIAN_SYSTEM
     if(c != 0) while(*letter_buf == 0) ++letter_buf;
+
+    // The byte order may or may not need to be swapped here to get correct c_uni below,
+    // since lv_textarea_add_text is ordering bytes correctly bfore calling lv_textarea_add_char.
+    // Assume swapping is needed if MSB is zero. May not be foolproof.
+    if((c != 0) && ((c & 0xff000000) == 0)) {
+        c2 = ((c >> 24) & 0xff) | // move byte 3 to byte 0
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: src/widgets/textarea/lv_textarea.c [hunk 1]
```diff
@@ -116,14 +116,25 @@ void lv_textarea_add_char(lv_obj_t * obj, uint32_t c)
 
     const char * letter_buf = (char *)&u32_buf;
 
+    uint32_t c2 = c;
 #if LV_BIG_ENDIAN_SYSTEM
     if(c != 0) while(*letter_buf == 0) ++letter_buf;
+
+    // The byte order may or may not need to be swapped here to get correct c_uni below,
+    // since lv_textarea_add_text is ordering bytes correctly bfore calling lv_textarea_add_char.
```
kisvegabor:
```suggestion
    // since lv_textarea_add_text is ordering bytes correctly before calling lv_textarea_add_char.
```

Please use `/* */` comments.

kisvegabor (commented):
Thank you! it's great that v9 is tested on a Big endian machine too.

kisvegabor (approved):

kisvegabor:
Thank you! :clap:

PR #5481, opened 2024-01-25
feat(display): add save screenshot to file support

FASTSHIFT:

Added screenshot saving file function to facilitate debugging of images during the drawing process.

Thanks to draw_buf, which unifies the data structure of the drawing buffer, it is extremely convenient to expand related functions later.

### Notes

### DIFF
--- a/src/misc/lv_utils.c
+++ b/src/misc/lv_utils.c
@@ -9,6 +9,7 @@
 #include <stddef.h>
 
 #include "lv_utils.h"
+#include "lv_fs.h"
 
 /*********************
  *      DEFINES
@@ -55,6 +56,38 @@ void * _lv_utils_bsearch(const void * key, const void * base, uint32_t n, uint32
         }
     }
     return NULL;
+}
+
+lv_result_t lv_draw_buf_save_to_file(const lv_draw_buf_t * draw_buf, const char * path)
+{
+    lv_fs_file_t file;
+    lv_fs_res_t res = lv_fs_open(&file, path, LV_FS_MODE_WR);
+    if(res != LV_FS_RES_OK) {
+        LV_LOG_ERROR("create file %s failed", path);
+        return LV_RESULT_INVALID;
+    }
+
+    /*Image content modified, invalidate image cache.*/
+    lv_image_cache_drop(path);
+
+    uint32_t bw;
+    res = lv_fs_write(&file, &draw_buf->header, sizeof(draw_buf->header), &bw);
+    if(res != LV_FS_RES_OK || bw != sizeof(draw_buf->header)) {
+        LV_LOG_ERROR("write draw_buf->header failed");
+        lv_fs_close(&file);
+        return LV_RESULT_INVALID;
+    }
+
+    res = lv_fs_write(&file, draw_buf->data, draw_buf->data_size, &bw);
+    if(res != LV_FS_RES_OK || bw != draw_buf->data_size) {
+        LV_LOG_ERROR("write draw_buf->data failed");
+        lv_fs_close(&file);
+        return LV_RESULT_INVALID;
+    }
+
+    lv_fs_close(&file);
+    LV_LOG_TRACE("saved draw_buf to %s", path);
+    return LV_RESULT_OK;
 }
 
 /**********************
--- a/src/misc/lv_utils.c
+++ b/src/misc/lv_utils.c
@@ -10,6 +10,7 @@
 
 #include "lv_utils.h"
 #include "lv_fs.h"
+#include "cache/lv_image_cache.h"
 
 /*********************
  *      DEFINES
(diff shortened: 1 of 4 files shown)


PGNetHun (changes requested):
Sorry, but I think **this PR should not be added** to display and draw_buf, because:
1. display/draw_buf should contain only tightly related functions, and not any unrelated things. Meaning: do not add function to create a file in file system to write actual screen buffer into it. That should be done in a separate code (higher level of architecture).
2. -> Please avoid the "god object" anti-pattern!
3. There is already a feature to take screenshots: [snapshot](https://docs.lvgl.io/master/others/snapshot.html)
4. Do not add a function that will be used (only) for testing / debugging
5. display/draw_buf should not know about FS module
6. if FS is not enabled in `lv_conf.h`, then saving would throw error

XuNeo:
> That should be done in a separate code (higher level of architecture).

How about adding utils directly in `lvgl/src/others/utils/lv_utils.h`

It's really useful to debug drawing issues, we have been using the same method since v8 together with obj screenshot. 

The code is not only used for debugging but can also be used by application. For example, current app view can be saved to file system as an app preview image, and then be used in background task manager.

PGNetHun:
Yes, if that is really needed, then it can be added to `lv_utils.h`; but FS-related calls have definitely nothing to do in display module / drawing layer (from archictecture perspective). If that function would save/duplicate the draw_buf, then we could keep it there, but not a file system saving function.
These not core functionalities should go definitely in separate modules, like `lv_utils.h`, etc.
Keep architecture and code structure clean, and separate core functionality from "higher-level" codes.

src/misc/lv_utils.c
```diff
@@ -57,6 +58,35 @@ void * _lv_utils_bsearch(const void * key, const void * base, uint32_t n, uint32
     return NULL;
 }
 
+lv_result_t lv_draw_buf_save_file(const lv_draw_buf_t * draw_buf, const char * path)
+{
+    lv_fs_file_t file;
```
XuNeo, line 64: Should we invalidate the image cache here?

src/misc/lv_utils.c
```diff
@@ -57,6 +58,35 @@ void * _lv_utils_bsearch(const void * key, const void * base, uint32_t n, uint32
...
+
+    uint32_t bw;
+    res = lv_fs_write(&file, &draw_buf->header, sizeof(draw_buf->header), &bw);
+    if(res != LV_FS_RES_OK || bw != sizeof(draw_buf->header)) {
+        LV_LOG_ERROR("write draw_buf->header failed");
+        lv_fs_close(&file);
+        return LV_RESULT_INVALID;
+    }
+
+    res = lv_fs_write(&file, draw_buf->data, draw_buf->data_size, &bw);
```
XuNeo, line 82: The draw_buf->data may also be changed by GPU, the moment CPU wants to read it, should we invalidate D-cache here too?

kisvegabor:
What about making it a snapshot feature, like `lv_snapshot_take_to_file`? It would be more generic, not only for a display. 

It would be even better to have a `lv_draw_buf_save_to_file` helper.

XuNeo:
> What about making it a snapshot feature, like `lv_snapshot_take_to_file`? It would be more generic, not only for a display.

It could also be really useful. Let's prepare it in another PR after https://github.com/lvgl/lvgl/pull/5487 is reviewed.

src/misc/lv_utils.c
PGNetHun: The quickiest solution would be to duplicate draw buf.
Or to “lock” draw buf, that nobody changes it.
Btw thanks for moving this function out of draw buf core functionality.

XuNeo: Duplicate the draw buffer is a CPU action, which still needs to consider cache issue. I'm not sure if we have a central place to take care of cache synchronization and we are free to use it by CPU here if so.

PGNetHun (commented):
Because file operations (especially saving many kilobytes to flash) can take many hundred miliseconds, it is better to duplicate the draw buf in memory, and saving that to file. Memory copy finishes faster, so even if source draw buf is not locked, the chance of “transient” state is less than saving the source directly to file system.

So, you have to decide: block drawing for a few ten-hundred miliseconds to save draw buf to filesystem, or duplicate it (more memory needed) and save that snapshot.

src/misc/lv_utils.c
XuNeo: Added.

src/misc/lv_utils.c
XuNeo: I decides to ignore cache issue here, and leave to user code to take care of it since only up level code know if there's cache inconsistency.

XuNeo:
LVGL operates on a single thread, which prevents the use of its API in a separate thread for file writing. It is advisable to allow the user to determine the best approach, whether that involves duplicating a new draw buffer for a different thread or creating a static loading UI that displays `saving file...`.

This API could serve as an initial step for users who wish to optimize it. Additionally, this API is not solely beneficial for debugging. It can also be utilized to save PNG decoded images as .bin images, for instance.

PGNetHun (approved):

PR #5496, opened 2024-01-26
feat(draw_buf): distinguish between lv_image_dsc_t and lv_draw_buf_t

### LINKED ISSUE #5054: Use `image_buf` instead of `draw_buf`

XuNeo:

Based on https://github.com/lvgl/lvgl/pull/5487 Fix https://github.com/lvgl/lvgl/issues/5054
lv_image_dsc_t should only by used to describe a constant image
resource.

lv_draw_buf_t is used to describe an image in RAM.

lv_image_buf.c/.h is now removed.

### Notes


src/draw/lv_draw_buf.h
```diff
@@ -25,17 +25,127 @@ extern "C" {
...
+ *  It must not be a valid ASCII character nor larger than 0x80. See `lv_image_src_get_type`.
+ */
+#define LV_IMAGE_HEADER_MAGIC (0x19)
+LV_EXPORT_CONST_INT(LV_IMAGE_HEADER_MAGIC);
+
 /**********************
  *      TYPEDEFS
  **********************/
 
+typedef enum _lv_image_flags_t {
```
PGNetHun: Well, I think removing `lv_image_buf`, then simply copy-pasting it's enums and things is not the best way.
This change made code more confusing: why are there "image" related enums in `lv_draw_buf` ?
I think we should keep the original `lv_image_buf`, or move into `lv_image` OR change the "image" names here in `lv_draw_buf` to avoid this deceptive naming here.
But I think sometimes it's better not to touch an already working and clean code.... there is no need to refactor everything.

PGNetHun (changes requested):

src/draw/lv_draw_buf.h
XuNeo: See discussion here. https://github.com/lvgl/lvgl/issues/5054

The primary distinction to note is that lv_image_dsc_t is intended for image resources, while the draw buffer is used for internally allocated intermediate images. It's crucial to maintain this separation clear to prevent confusion about the appropriate usage and timing for each.

Removing the lv_image_buf.c/.h is because we can only do modification to draw buffer. 
The image flag  in draw buffer does confuse a little. Do you have better idea?

kisvegabor: I also find "image flags" in "draw buf" a little bit confusing. However I understand that it's in line wit our endeavor to use only `draw_buf` in the end. 

Can we have e.g. `LV_IMAGE_FLAGS_PREMULTIPLIED` or other flags on a draw buf?

If so, to proceed gradually I suggest:
- Renaming everything in this file to "... draw buf ..." 
- Add the image related names to `lv_api_map.h` or to this file with typedef an defines.

XuNeo: Well, I think it's not ideal. Because a draw buffer is an image, so it can have image flags. 
An image is not an draw buffer, but it can have image flags since it's an image.

It has similar idea in OpenGL, see https://stackoverflow.com/questions/13909404/whats-the-difference-between-image-and-texture-in-opengl. An image loaded to RAM that GPU can use later is a texture. We use name draw buffer instead.

PGNetHun: my concern is not only about image flags, but also the other image structs moved to this draw buf file.
image header struct is also similar to draw buf struct, and some other functions are moved also to here, still with `lv_image_` prefix.
It makes confusing to differ the “real image” from a “simple draw buffer”.

btw the binary images (stored in C array or separate file) will also contain stride, and “empty vertical space” (unused bytes), when image width does not fit exactly in 4/8/16 bytes alignment?

XuNeo: I actually wanted to rename it to lv_draw_buf_xxx, because only draw buffer can be modifiable, but am concerned about API breaking.  Do you think we should rename them too?

> btw the binary images (stored in C array or separate file) will also contain stride, and “empty vertical space” 
 (unused bytes), when image width does not fit exactly in 4/8/16 bytes alignment?

The padding is only added as instructed in the python script. It's also not needed by the software renderer and image decoder.

src/draw/lv_draw_buf.h
```diff
@@ -247,7 +350,41 @@ static inline void lv_draw_buf_from_image(lv_draw_buf_t * buf, const lv_image_ds
...
+ *   - for `LV_COLOR_FORMAT_I2`: 0..3
+ *   - for `LV_COLOR_FORMAT_I4`: 0..15
+ *   - for `LV_COLOR_FORMAT_I8`: 0..255
+ * @param c the color to set in lv_color32_t format
+ */
+void lv_draw_buf_set_palette(lv_draw_buf_t * draw_buf, uint8_t id, lv_color32_t c);
+
+static inline void lv_image_buf_set_palette(lv_image_dsc_t * dsc, uint8_t id, lv_color32_t c)
+{
+    LV_LOG_WARN("Legacy API, use lv_draw_buf_set_palette instead.");
```
kisvegabor: I suggest moving all legacy API `lv_api_map.h`.

Maybe deprecated is a better word for it.

kisvegabor (commented):
It seems to me we have types and functions for 
- draw_buf only: the manage modifiable buffers
- images only: e.g. `lv_image_dsc_t`
- both: `LV_IMAGE_FLAG_...`

I'm not sure if we need any distinction between images and draw buffer except the `LV_IMAGE_FLAGS_MODIFIABLE` flag.

Of course, we need some backward compatibility layer in the core LVGL, but I would keep only draw_buf.

src/draw/lv_draw_buf.h
PGNetHun: Yes, the word "deprecated" is better.

src/draw/lv_draw_buf.c
```diff
@@ -374,6 +381,20 @@ lv_result_t lv_draw_buf_premultiply(lv_draw_buf_t * draw_buf)
     return LV_RESULT_OK;
 }
 
+void lv_draw_buf_set_palette(lv_draw_buf_t * draw_buf, uint8_t id, lv_color32_t c)
```
PGNetHun: I know that this function was copy-pasted, but some remarks:

1. No need to shorten parameter names, will not affect final compiled binary size.
Yes, we have function comment describing what it is, but if possible avoid mental mapping (clean code tip).
Simply write: `color` instead of `c`.

2. Parameter `id`: I would use the name `index` instead of `id`, as it is more commonly used in relation to palette colors. But I'm OK also with `id`, just wanted to mention.
See: https://en.wikipedia.org/wiki/Palette_(computing) and https://en.wikipedia.org/wiki/Indexed_color

src/lv_api_map_v8.h
```diff
@@ -249,6 +249,64 @@ static inline void lv_obj_move_background(lv_obj_t * obj)
...
+ * @return          LV_RESULT_OK on success, LV_RESULT_INVALID on error.
+ */
+static inline lv_result_t lv_snapshot_take_to_buf(lv_obj_t * obj, lv_color_format_t cf, lv_image_dsc_t * dsc,
+                                                  void * buf,
+                                                  uint32_t buf_size)
+{
+    lv_draw_buf_t draw_buf;
+    LV_LOG_WARN("Deprecated API, use lv_snapshot_take_to_draw_buf instead.");
+    lv_draw_buf_init(&draw_buf, 1, 1, cf, buf_size, buf, buf_size);
+    lv_result_t res = lv_snapshot_take_to_draw_buf(obj, cf, &draw_buf);
```
kisvegabor: Note these warnings.

kisvegabor (commented):
Let's approach the `draw_buf` and `image_dsc` differently. 

`image_dsc` is the same as `draw_buf` just more limited. We can even say that `draw_buf` is a superset/extension of `image_dsc`. 

If you agree with that we can keep `lv_image_buf.h` (maybe rename to `lv_image_dsc.h`) and include it `lv_draw_buf.h`.

src/lv_api_map_v8.h
XuNeo: I have removed the warning logs and added lv_image_dsc.h

src/draw/lv_draw_buf.h
XuNeo: The image related API now goes out of lv_draw_buf.h

src/lv_api_map_v8.h
kisvegabor: The warnings are still here.

kisvegabor (commented):
I like it!

src/lv_api_map_v8.h
XuNeo: Ah, get it. Now should be fixed.

kisvegabor (dismissed):
@PGNetHun Could you also take a look at it?

src/draw/lv_draw_buf.c
```diff
@@ -161,7 +161,15 @@ lv_result_t lv_draw_buf_init(lv_draw_buf_t * draw_buf, uint32_t w, uint32_t h, l
         return LV_RESULT_INVALID;
     }
 
-    lv_image_header_init(&draw_buf->header, w, h, cf, stride, 0);
+    lv_image_header_t * header = &draw_buf->header;
+    lv_memzero(header, sizeof(*header));
```
PGNetHun: No need to clear `draw_buf->header`, as it is already cleared (the whole `draw_buf`) in line 156.

src/draw/lv_draw_buf.c
```diff
@@ -161,7 +161,15 @@ lv_result_t lv_draw_buf_init(lv_draw_buf_t * draw_buf, uint32_t w, uint32_t h, l
...
+    header->w = w;
+    header->h = h;
+    header->cf = cf;
+    header->stride = stride;
+    header->flags = 0;
+    header->magic = LV_IMAGE_HEADER_MAGIC;
```
PGNetHun, line 170: Out of scope:
I suggest to rename `magic` to `signature`.
The name `magic` (number) is generic, but `signature` is more specific about it's purpose, and it is more commonly used term.

src/draw/lv_draw_image.c
```diff
@@ -219,6 +219,34 @@ void _lv_draw_image_tiled_helper(lv_draw_unit_t * draw_unit, const lv_draw_image
     lv_image_decoder_close(&decoder_dsc);
 }
 
+void _lv_image_buf_get_transformed_area(lv_area_t * res, int32_t w, int32_t h, int32_t angle, uint16_t scale_x,
+                                        uint16_t scale_y,
+                                        const lv_point_t * pivot)
```
PGNetHun: No need for the 3rd line for `pivot`, move `scale_x`, `scale_y`, `pivot` to 2nd line:
```
void _lv_image_buf_get_transformed_area(lv_area_t * res, int32_t w, int32_t h, int32_t angle, 
                                        uint16_t scale_x,  uint16_t scale_y, const lv_point_t * pivot)
```

src/draw/lv_draw_image.h
```diff
@@ -138,6 +138,20 @@ void _lv_draw_image_normal_helper(lv_draw_unit_t * draw_unit, const lv_draw_imag
...
+ * @param res store the coordinates here
+ * @param w width of the rectangle to transform
+ * @param h height of the rectangle to transform
+ * @param angle angle of rotation
+ * @param scale_x zoom in x direction, (256 no zoom)
+ * @param scale_y zoom in y direction, (256 no zoom)
+ * @param pivot x,y pivot coordinates of rotation
+ */
+void _lv_image_buf_get_transformed_area(lv_area_t * res, int32_t w, int32_t h, int32_t angle, uint16_t scale_x,
+                                        uint16_t scale_y,
```
PGNetHun: No need for 3 lines, move `scale_x`, `scale_y` and `pivot` parameters to 2nd line.

src/draw/lv_image_dsc.h
```diff
@@ -13,21 +13,17 @@ extern "C" {
...
  *      DEFINES
  *********************/
 
-#define _LV_ZOOM_INV_UPSCALE 5
-
 /** Magic number for lvgl image, 9 means lvgl version 9
  *  It must not be a valid ASCII character nor larger than 0x80. See `lv_image_src_get_type`.
  */
 #define LV_IMAGE_HEADER_MAGIC (0x19)
+LV_EXPORT_CONST_INT(LV_IMAGE_HEADER_MAGIC);
```
PGNetHun, line 26: As mentioned before, I suggest to rename `magic` to `signature`, as it better describes it's purpose than the generic name `magic`.
I assume this is an internal LVGL "magic number", so no API change is needed when renamed. Or do we need?

src/draw/lv_image_dsc.h
```diff
@@ -38,43 +34,38 @@ typedef enum _lv_image_flags_t {
...
-    LV_IMAGE_FLAGS_COMPRESSED       = (1 << 3),
+    LV_IMAGE_FLAGS_COMPRESSED       = 0x08,
+
+    /*Below flags are applicable only for draw buffer header.*/
 
     /**
      * The image is allocated from heap, thus should be freed after use.
      */
-    LV_IMAGE_FLAGS_ALLOCATED        = (1 << 4),
+    LV_IMAGE_FLAGS_ALLOCATED        = 0x0100,
```
PGNetHun: There is one more `0` in value.
It should be: `0x0010`

src/draw/lv_image_dsc.h
```diff
@@ -38,43 +34,38 @@ typedef enum _lv_image_flags_t {
...
+
+    /**
+     * If the image data is malloced and can be processed in place.
+     * In image decoder post processing, this flag means we modify it in-place.
+     */
+    LV_IMAGE_FLAGS_MODIFIABLE       = 0x0200,
```
PGNetHun: There is one more `0` in value.
It should be: `0x0020`

src/lv_api_map_v8.h
```diff
@@ -249,6 +249,68 @@ static inline void lv_obj_move_background(lv_obj_t * obj)
 #define lv_image_decoder_built_in_open      lv_bin_decoder_open
 #define lv_image_decoder_built_in_close     lv_bin_decoder_close
 
+/**
```
PGNetHun: I'm not sure that these should be here, and not in their original "place":
- `lv_snapshot_free`
- `lv_snapshot_take_to_buf`
- `lv_image_buf_set_palette`
- `lv_image_buf_free`

Functions (even if deprecated) should stay in their own component, and be removed from there... otherwise if there is a refactoring of the component, then these "moved" functions will be forgotten easily.

PGNetHun (changes requested):

src/draw/lv_draw_buf.c
XuNeo: I think `magic` is acceptable based on explaining `A unique value with unexplained meaning `. 

https://en.wikipedia.org/wiki/List_of_file_signatures

> data used to identify or verify the content of a file. Such signatures are also known as [magic numbers](https://en.wikipedia.org/wiki/Magic_number_(programming)#In_files) or Magic Bytes.

src/draw/lv_image_dsc.h
XuNeo: This is intended to leave some bit available for `const image` flags.
The `ALLOCATED` bit and below ones are applicable only for draw buffer header.

PGNetHun: But then `LV_IMAGE_FLAGS_ALLOCATED` and `LV_IMAGE_FLAGS_USER1` will have same value: `0x0100`
Same with `LV_IMAGE_FLAGS_MODIFIABLE` and `LV_IMAGE_FLAGS_USER2`:  `0x0200`

src/lv_api_map_v8.h
XuNeo: Leaving the deprecated functions split in several headers actually make it easier to be forgotten when refactoring. The day we decides to remove deprecated API, we can simply delete api map directly.

src/draw/lv_image_dsc.h
XuNeo: Ah, I see. Will fix it now.

src/lv_api_map_v8.h
PGNetHun: To move out (deprecated) functions from their module/component to somewhere else makes code more **fragmented**. 
The functions and data that belong together should be placed as close as possible (in OO into an object). 
(in C# we just add `[Obsolete]` attribute to methods, but **not moving out** of their object)
The day we decide to remove deprecated functions, we can search for word "obsolete", and remove them from their component. 
As I see the purpose of this `lv_api_map_v8.h` is mainly to have "aliases" to make some v8 function names to work with v9, and not the container file for deprecated functions.

kisvegabor: I think it was my proposal somewhere to move these to `lv_api_map_v8.h` but now I agree with @PGNetHun.

It's better for the users to see the depreciated and the normal functions on one place.  We can keep using `lv_api_map*.h` as a backward compatibility layer.

src/others/snapshot/lv_snapshot.h
```diff
@@ -67,33 +67,33 @@ lv_result_t lv_snapshot_reshape_draw_buf(lv_obj_t * obj, lv_draw_buf_t * draw_bu
...
 /**
- * Legacy API, use lv_snapshot_take_to_draw_buf instead.
  * Take snapshot for object with its children, save image info to provided buffer.
  * @param obj       the object to generate snapshot.
  * @param cf        color format for generated image.
  * @param dsc       image descriptor to store the image result.
  * @param buf       the buffer to store image data. It must meet align requirement.
  * @param buf_size  provided buffer size in bytes.
  * @return          LV_RESULT_OK on success, LV_RESULT_INVALID on error.
+ * @deprecated      Use lv_snapshot_take_to_draw_buf instead.
```
PGNetHun, line 89: In some places `@deprecated` is the first comment line, here it is simply added after `@return`

PGNetHun (approved):

src/others/snapshot/lv_snapshot.h
XuNeo: I think generated doc will include `deprecated` note, no matter where it locates. Let's see if it's true.

PR #5511, opened 2024-01-27
feat(st7789): Updated documentation. Added code example and step-by-step guide for STM32

zjanosy:

### Notes


examples/porting/lv_port_lcd_stm32_template.c
```diff
@@ -0,0 +1,183 @@
...
+ **********************/
+
+void lv_port_display_init(void)
+{
+    /* Initialize LCD I/O */
+    if(lcd_io_init() != 0)
+        return;
+
+    /* Create the LVGL display object and the ST7789 LCD display driver */
+    lcd_disp = lv_st7789_create(MY_DISP_HOR_RES, MY_DISP_VER_RES, 0, lcd_send_cmd, lcd_send_color);
```
LeCyberDucky: Instead of the `0`, I would use `LV_LCD_FLAG_NONE`, to make it clearer that flags for this purpose exist.

zjanosy: I'll change it, thanks for the suggestion!

examples/porting/lv_port_lcd_stm32_template.c
```diff
@@ -0,0 +1,183 @@
...
+/*********************
+ *      DEFINES
+ *********************/
+#ifndef MY_DISP_HOR_RES
+    #warning Please define or replace the macro MY_DISP_HOR_RES with the actual screen width, default value 320 is used for now.
+    #define MY_DISP_HOR_RES    320
+#endif
+
+#ifndef MY_DISP_VER_RES
+    #warning Please define or replace the macro MY_DISP_HOR_RES with the actual screen height, default value 240 is used for now.
```
PGNetHun: Rename:  `MY_DISP_HOR_RES`  ->  `MY_DISP_VER_RES`

PGNetHun (approved):
Looks good, found only 1 copy-paste issue

examples/porting/lv_port_lcd_stm32_template.c
zjanosy: Good catch! Actually I copied this from lv_port_disp_template.c. I fixed it in both places.

PGNetHun (approved):

examples/porting/lv_port_lcd_stm32_template.c
```diff
@@ -0,0 +1,183 @@
...
+    lv_display_set_rotation(lcd_disp, LV_DISPLAY_ROTATION_270);     /* set landscape orientation */
+
+    /* Example: two dynamically allocated buffers for partial rendering */
+    lv_color_t * buf1 = NULL;
+    lv_color_t * buf2 = NULL;
+
+    uint32_t buf_size = MY_DISP_HOR_RES * MY_DISP_VER_RES / 10 * lv_color_format_get_size(lv_display_get_color_format(
+                                                                                              lcd_disp));
+
+    buf1 = lv_malloc(buf_size);
```
kisvegabor, line 80: Use global or static variables instead of malloc.

examples/porting/lv_port_lcd_stm32_template.c
```diff
@@ -0,0 +1,183 @@
...
+
+/*Copy this file as "lv_port_disp.c" and set this value to "1" to enable content*/
+#if 0
+
+/*********************
+ *      INCLUDES
+ *********************/
+/* Include STM32Cube files here, e.g.:
+#include "stm32f7xx_hal.h"
+*/
```
kisvegabor, line 15:
```suggestion
#if defined STM32F3
#include "stm32f3xx_hal.h"
#elif defined STM32F4
#include "stm32f4xx_hal.h"
#elif defined STM32F7
#include "stm32f7xx_hal.h"
#elif defined STM32H7
#include "stm32xxx_hal.h" /*Include your hal header */
#endif
```

examples/porting/lv_port_lcd_stm32_template.h
```diff
@@ -0,0 +1,49 @@
...
+
+/**********************
+ *      TYPEDEFS
+ **********************/
+
+/**********************
+ * GLOBAL PROTOTYPES
+ **********************/
+/* Initialize low level display driver */
+void lv_port_disp_init(void);
```
kisvegabor, line 37: As we can have other STM port files too (e.g. LTDC) let name it
```suggestion
void lv_port_lcd_stm32_init(void);
```

docs/integration/driver/display/lcd_stm32_guide.rst
```diff
@@ -0,0 +1,320 @@
...
+pin  configuration LCD     user label
+==== ============= ======= ==========
+PA4  GPIO_Output   CS	   LCD_CS
+PA5  SPI1_SCK	   SCK	   --
+PA7  SPI1_MOSI	   SDI     --
+PA15 GPIO_Output   RESET   LCD_RESET
+PB10 GPIO_Output   DC      LCD_DCX
+==== ============= ======= ==========
+
+Step-by-step instructions
```
kisvegabor, line 46: Do we really need such a detailed guide here? 

I think we should explain only the LVGL related things and not general topic (such as clock config). For such a detailed and specific configuration we should have a blog post or YouTube video instead that we can link from here.

I suggest updating [this file](https://docs.lvgl.io/master/integration/chip/stm32.html) instead where we explain:
1. How to integrate LVGL
   - clone LVGL
   - lv_tick 
2. STM devices can be used with 
   - LVGL's built in display controller drivers
      - 1-2 paragraph of description about what to configure in Cube IDE (note that STm devices can be used with outher IDEs as well)  
      - link to `lv_port_stm32_lcd`
   - X-CUBE-DISPLAY (just one paragraph)
   - LTDC: 1-2 paragraphs. 

I hope we can have an official collaboration with ST in the future. However, until that I would like focus on our Official partners. 

For example it would be nice to have the same description for ESP as well.

examples/porting/lv_port_lcd_stm32_template.c
zjanosy: Please note that this code has not been tested on other platforms than the STM32F746. We can add these headers, but what if there are subtle differences between those? Since this is an example, I think it is reasonable to expect that the users will adapt it to their platform.

examples/porting/lv_port_lcd_stm32_template.h
zjanosy: This has been copied from lv_port_disp_template.c. Since 99% of the users will only use a single display, it does not really matter if the templates use the same function name (actually, it makes sense). But since this is just a template, the users can change it if there is any conflict.

docs/integration/driver/display/lcd_stm32_guide.rst
zjanosy: I think it is very useful to provide a step-by-step guide like this, because the difficult part is not how to integrate LVGL, but how to configure the hardware. While this could be a blog post or YouTube video, including it in the LVGL documentation will immediately help novice users with their LVGL project, saving the hassle of searching the Internet for solutions (which may be outdated and may not be correct). Having this guide would have saved me days of work... :-)

The reason for making the guide so detailed is that leaving out any of these steps (e.g., clock or DMA configuration) will make LVGL perform sub-optimally. And that would cause user complaints of LVGL in general or discarding LVGL alltogether.

examples/porting/lv_port_lcd_stm32_template.c
zjanosy: OK, I'll change it.

examples/porting/lv_port_lcd_stm32_template.c
kisvegabor: In this case can we add it as an example to one of our lv_port_stm32* repos and link it in the docs?

docs/integration/driver/display/lcd_stm32_guide.rst
kisvegabor: Okay let's leave it as it is but please add link to it [here](
https://docs.lvgl.io/master/integration/chip/stm32.html)

kisvegabor:
Not stale.

kisvegabor (approved):
I'm not sure we take the best possible approach, but we can update the examples at any time later. It will be quite valuable for the users as it is so I suggest merging this PR.

PR #5548, opened 2024-01-31
fix(draw): fix the invalidation of scaled areas

### LINKED ISSUE #5447: Use the `lv_obj_set_style_transform_scale` function to scale, and there will still be residual borders when zoomed out.
### LVGL version

v9.0.0

### What happened?

Use the `lv_obj_set_style_transform_scale` function to scale, and there will still be residual borders when zoomed out.

https://github.com/lvgl/lvgl/assets/18370433/5fa0469b-1846-4527-9bb2-171f46333a7f

### How to reproduce?

```c
static void anim_cb(void * obj, int32_t value)
{
  lv_obj_set_style_transform_scale(obj, value, LV_PART_MAIN);
}

void test_demo(void)
{
  lv_obj_t *obj=lv_obj_create(lv_screen_active());
  lv_obj_set_size(obj,200,200);
  lv_obj_set_style_bg_color(obj,lv_palette_main(LV_PALETTE_BLUE),LV_PART_MAIN);
  lv_anim_t aa;
  lv_anim_init(&aa);
  lv_anim_set_var(&aa, obj);
  lv_anim_set_values(&aa, 0, 256);
  lv_anim_set_time(&aa, 500);
  lv_anim_set_playback_time(&aa, 500);
  lv_anim_set_repeat_count(&aa, LV_ANIM_REPEAT_INFINITE);
  lv_anim_set_path_cb(&aa, lv_anim_path_linear);
  lv_anim_set_exec_cb(&aa, anim_cb);
  lv_anim_start(&aa);
}
```

kisvegabor:

fix #5447

### Notes


kisvegabor:
cc @lhdjply

lhdjply:
I tested it. It's an improvement over before. However, there is still some residue after the bezel is scaled.
### test code
```c
static void anim_cb(void * obj, int32_t value)
{
  lv_obj_set_style_transform_scale(obj, value, LV_PART_MAIN);
}

void test_demo(void)
{
  lv_obj_t *obj=lv_obj_create(lv_screen_active());
  lv_obj_set_size(obj,200,200);
  lv_obj_set_style_bg_color(obj,lv_palette_main(LV_PALETTE_BLUE),LV_PART_MAIN);
  lv_anim_t aa;
  lv_anim_init(&aa);
  lv_anim_set_var(&aa, obj);
  lv_anim_set_values(&aa, 0, 256);
  lv_anim_set_time(&aa, 500);
  lv_anim_set_playback_time(&aa, 500);
  lv_anim_set_repeat_count(&aa, LV_ANIM_REPEAT_INFINITE);
  lv_anim_set_path_cb(&aa, lv_anim_path_linear);
  lv_anim_set_exec_cb(&aa, anim_cb);
  lv_anim_start(&aa);
}
```

lhdjply:
I tested it. It's an improvement over before. However, there is still some residue after the bezel is scaled.

https://github.com/lvgl/lvgl/assets/18370433/41e1991f-609c-4c4b-81c1-53502ffd8e3e


### test code
```c
static void anim_cb(void * obj, int32_t value)
{
  lv_obj_set_style_transform_scale(obj, value, LV_PART_MAIN);
}

void test_demo(void)
{
  lv_obj_t *obj=lv_obj_create(lv_screen_active());
  lv_obj_set_size(obj,200,200);
  lv_obj_set_style_bg_color(obj,lv_palette_main(LV_PALETTE_BLUE),LV_PART_MAIN);
  lv_anim_t aa;
  lv_anim_init(&aa);
  lv_anim_set_var(&aa, obj);
  lv_anim_set_values(&aa, 0, 256);
  lv_anim_set_time(&aa, 500);
  lv_anim_set_playback_time(&aa, 500);
  lv_anim_set_repeat_count(&aa, LV_ANIM_REPEAT_INFINITE);
  lv_anim_set_path_cb(&aa, lv_anim_path_linear);
  lv_anim_set_exec_cb(&aa, anim_cb);
  lv_anim_start(&aa);
}
```

src/core/lv_obj_pos.c
```diff
@@ -783,15 +788,12 @@ void lv_obj_get_transformed_area(const lv_obj_t * obj, lv_area_t * area, bool re
 {
     lv_point_t p[4] = {
         {area->x1, area->y1},
```
lhdjply, line 790: I seem to have fixed it with the following code, but I don't know why.
```suggestion
        {area->x1 - 1, area->y1 - 1},
        {area->x1 - 1, area->y2 + 2},
        {area->x2 + 2, area->y1 - 1},
        {area->x2 + 2, area->y2 + 2},
```

kisvegabor:
Oh, thank you! I'm on it.

kisvegabor:
Huh, it seems working. To be honest I'm not so happy with it because it's hacky here and there. 

The current image transformation algorithm has some precision issues due to the integer math. I have some ideas to improve it, but it will require much more time.

lhdjply:
I think we can merge first to fix the bug, and then slowly step by step, optimize the code, through the accuracy.

IAMMX:
I tested it. It doesn't seem to fix it. It's still the original result.
### Original picture
The image format is ARGB8888.
295810594-7354024f-896b-4cec-8ac0-f4f58ecc85b8
### Code
```c
LV_IMG_DECLARE(img_test);
void img_scale_test(void)
{
  lv_obj_t *logo1;
  logo1 = lv_image_create(lv_screen_active());
  lv_image_set_src(logo1, &img_test);
  lv_obj_align(logo1,LV_ALIGN_CENTER,-200,0);
  lv_image_set_scale(logo1, 256);

  lv_obj_t *logo2;
  logo2 = lv_image_create(lv_screen_active());
  lv_image_set_src(logo2, &img_test);
  lv_obj_align(logo2,LV_ALIGN_CENTER,200,0);
  lv_image_set_scale(logo2, 255);
}
```
### Result
Screenshot_20240206_131321_com huawei browser

kisvegabor:
@IAMMX 
I've debugged it an it turned out that it's a rounding error related to LVGL stores some coordinates internally with 1 px resolution.  This is what happens:
- The image has 240px width, meaning x1 = 0, x2 = 239
- When we scale it down to 255/256 around pivot_x = 120 we get x1 = 0.5 -> 0; x2 = 118.5 -> 238.
- So the scaled image is on x1=0; x2= 238;
- When LVGL calculates the inverse scale during transformations, (with pivot_x = 120 )it will get x2 = (238-120) * 256 / 255 + 120 = 238.4. It's stored with higher precision, be we should have get 239 here
- So the precision was lost when we stored the transformed area with 1 px precision

Probably it can be improved, but it would be larger scale update. 

As a workaround you can do 2 things:
- add a few transparent pixel around the image
- calling `lv_image_set_pivot(logo, 2, 2)` also solved the issue for me in this particular config as this way we do the calculations with 239 instead of 119. (meaning less error)

IAMMX:
> add a few transparent pixel around the image
This method can work, but every time you create the image, be careful to leave some transparent areas.

> calling lv_image_set_pivot(logo, 2, 2) also solved the issue for me in this particular config as this way we do the calculations with 239 instead of 119. (meaning less error)
This method will not work if it is used for picture scaling animations.

kisvegabor:
> `lv_image_set_pivot(logo, 2, 2)`

It doesn't change anything, but I meant `lv_image_set_pivot(logo, 0, 0)`

---

Unfortunately I don't have a good solution for the accuracy issue now. :slightly_frowning_face:

lhdjply:
In 4dcd66e, the image layer_transform_2.png is added, but in the code, no `TEST_ASSERT_EQUAL_SCREENSHOT("draw/layer_transform_2.png")` is found;

kisvegabor:
True, thank you. Removed.

lhdjply (approved):

src/core/lv_obj_pos.h
```diff
@@ -340,6 +340,16 @@ void lv_obj_move_children_by(lv_obj_t * obj, int32_t x_diff, int32_t y_diff, boo
...
 
+/**
+ * Transform an array of points using the angle and zoom style properties of an object
+ * @param obj           pointer to an object whose style properties should be used
+ * @param p             the array of points to transform, the result will be written back here too
+ * @param p_count       number of points in the array
+ * @param recursive     consider the transformation properties of the parents too
+ * @param inv           do the inverse of the transformation (-angle and 1/zoom)
+ */
+void lv_obj_transform_point_array(const lv_obj_t * obj, lv_point_t p[], size_t p_count, bool recursive, bool inv);
```
PGNetHun: Can we have please full names for parameters? (clean code: choose descriptive and unambiguous names)
`p` -> `points`
`p_count` -> `count`
`inv` -> `inverse`

Same for `lv_obj_transform_point`.

src/core/lv_obj_pos.c
```diff
@@ -783,15 +788,12 @@ void lv_obj_get_transformed_area(const lv_obj_t * obj, lv_area_t * area, bool re
...
-        {area->x1, area->y2},
-        {area->x2, area->y1},
-        {area->x2, area->y2},
+        {area->x1, area->y2 + 1},
```
PGNetHun, line 793: Just a question:
is it really correct to add always 1 pixel, or can it be avoided by changing way handling areas?
Because it's kind of _code smell_ to add always +1 pixel to `x2` and `y2`; previous code did not do that; and as I remember I saw also other places add/substract 1 pixel from/to `x2` / `y2`.

src/misc/lv_area.h
```diff
@@ -265,6 +260,10 @@ void lv_area_align(const lv_area_t * base, lv_area_t * to_align, lv_align_t alig
 void lv_point_transform(lv_point_t * p, int32_t angle, int32_t scale_x, int32_t scale_y, const lv_point_t * pivot,
                         bool zoom_first);
 
+void lv_point_array_transform(lv_point_t * p, size_t p_count, int32_t angle, int32_t scale_x, int32_t scale_y,
```
PGNetHun: Can we have meaningful parameter names here, too?
`p` -> `points`
`p_count` -> `count`

(same for `lv_point_transform` function)

PGNetHun (commented):
Just some minor naming related comments.
I've not tested changes, but see that reference images are updated, so I assume code is OK.

Bonus test case (optional):
Can you please add also 1 test:
show a frame of a [sprite/texture atlas](https://en.wikipedia.org/wiki/Texture_atlas), and rotate it.
I use sprites in my app (using `lv_image` `offset_x`, `offset_y`, and rotation), and the sprite is shown wrong (cut of image is after rotation, so parts of next frame appear, too).
Example:
animation of ants: use a sprite PNG as source for `lv_img` and `lv_animation` (+ callback to set image `offset_y`) for frame animation, and rotate the `lv_img` to the walking direction.
ant

The bug:
image

src/core/lv_obj_pos.h
kisvegabor: I agree, fixed.

src/core/lv_obj_pos.c
kisvegabor: Honestly, it's really a mess. E.g. width == 100 means, x1 = 0; x2 = 99. If zoom to 1000% it should be x1 = 0; x2=999.9 (~1000). So as the last pixel is zoomed we need assume that it's there. Combining this with rounding issues due to integer math, it's at least not intuitive. 

It would require a new iteration on the whole transformation logic to make it better somehow.

kisvegabor:
> Can you please add also 1 test:
show a frame of a [sprite/texture atlas](https://en.wikipedia.org/wiki/Texture_atlas), and rotate it.

Well, I haven't realized it so far, but there is no such a feature in LVGL now. Always the whole image is rotated and the clip area of the destination area is just a normal (not rotated) rectangle described by 2 points. To support what you have described we needed to apply a clip area on the source image too. It's possible but not supported now. 

As workaround you can transform the image as a layer:
```c
    lv_obj_set_style_transform_rotation(img1, 450, 0);
    lv_obj_set_style_transform_pivot_x(img1, 0, 0);
    lv_obj_set_style_transform_pivot_y(img1, 0, 0);
```
It's slower and uses more memory, but for small images it shouldn't be a big issue.

IAMMX:
Screenshot_20240216_150522_com huawei browser

Why did you delete all these test pictures?

PGNetHun (dismissed):

kisvegabor:
Thank you for pointing it out. I've just added back the reference images images.

IAMMX:
Since you have already merged master, please remove aeb60a9. Because the issue has been fixed in 46e338e.

kisvegabor:
I have reverted it. We will squash commits so don't be scared off the long history :slightly_smiling_face:

lhdjply:
This branch has conflicts that must be resolved

kisvegabor:
Fixed

src/core/lv_obj_pos.h
```diff
@@ -340,14 +340,25 @@ void lv_obj_move_children_by(lv_obj_t * obj, int32_t x_diff, int32_t y_diff, boo
  */
 void lv_obj_transform_point(const lv_obj_t * obj, lv_point_t * p, bool recursive, bool inv);
 
+/**
+ * Transform an array of points using the angle and zoom style properties of an object
+ * @param obj           pointer to an object whose style properties should be used
+ * @param points        the array of points to transform, the result will be written back here too
+ * @param count         number of points in the array
+ * @param recursive     consider the transformation properties of the parents too
+ * @param inverse       do the inverse of the transformation (-angle and 1/zoom)
```
XuNeo: How about use a single `flag` parameter defined as "`LV_OBJ_TRANSFORM_POINT_INVERSE | LV_OBJ_TRANSFORM_POINT_RECURSIVE`"?

It could improve code readability considering below usage.
`lv_obj_transform_point_array(obj, points, n, true, true)`

examples/widgets/image/lv_example_image_1.c
```diff
@@ -7,10 +7,15 @@ void lv_example_image_1(void)
     lv_obj_t * img1 = lv_image_create(lv_screen_active());
     lv_image_set_src(img1, &img_cogwheel_argb);
     lv_obj_align(img1, LV_ALIGN_CENTER, 0, 0);
+    lv_image_set_scale_x(img1, 512);
+    lv_image_set_scale_y(img1, 128);
+    lv_image_set_rotation(img1, 10);
```
XuNeo: Is this change intended? `example_image_1` is the start point to use image, better leave it as simple as possible.

src/core/lv_obj_pos.c
```diff
@@ -808,7 +810,11 @@ void lv_obj_invalidate_area(const lv_obj_t * obj, const lv_area_t * area)
 
     lv_area_t area_tmp;
     lv_area_copy(&area_tmp, area);
+
     if(!lv_obj_area_is_visible(obj, &area_tmp)) return;
+    if(obj->spec_attr && obj->spec_attr->layer_type == LV_LAYER_TYPE_TRANSFORM) {
+        lv_area_increase(&area_tmp, 5, 5);
```
XuNeo, line 820: Need some comments for this magic number 5.

src/core/lv_obj_pos.c
```diff
@@ -1143,9 +1149,9 @@ static void transform_point(const lv_obj_t * obj, lv_point_t * p, bool inv)
 
     if(inv) {
         angle = -angle;
-        scale_x = (256 * 256) / scale_x;
-        scale_y = (256 * 256) / scale_y;
+        scale_x = (256 * 256 + scale_x - 1) / scale_x;
```
XuNeo, line 1158: Do we need to check `scale_x == 0`?

kisvegabor:
Thank you, Neo. All fixed.

XuNeo (approved):

FASTSHIFT (approved):

C47D (approved):

PR #5562, opened 2024-02-01
feat(fs): implement littlefs lfs.h driver support

PierreRambaud:

Not sure it's the best way, but this is how I use LittleFS.h with Arduino on my side :)
Waiting for your feedback :pray: 
I'm sorry for not adding tests, never did it in C :/

### Notes

### DIFF
--- a/docs/libs/fs.rst
+++ b/docs/libs/fs.rst
@@ -14,6 +14,7 @@ LVG has built in support for:
 - POSIX (Linux and Windows using POSIX function .e.g ``open``, ``read``)
 - WIN32 (Windows using Win32 API function .e.g ``CreateFileA``, ``ReadFile``)
 - MEMFS (read a file from a memory buffer)
+- LITTLEFS (a little fail-safe filesystem designed for microcontrollers)
 
 You still need to provide the drivers and libraries, this extension
 provides only the bridge between FATFS, STDIO, POSIX, WIN32 and LVGL.
--- a/docs/libs/lfs.rst
+++ b/docs/libs/lfs.rst
@@ -1,0 +1,59 @@
+.. _lfs:
+
+==============
+littlefs
+==============
+
+littlefs is a little fail-safe filesystem designed for microcontrollers.
+
+Detailed introduction: https://github.com/littlefs-project/littlefs
+
+
+Usage
+-----
+
+Enable :c:macro:`LV_USE_FS_LITTLEFS` and define a :c:macro`LV_FS_LITTLEFS_LETTER` in ``lv_conf.h``.
+
+When enabled :c:macro:`lv_littlefs_set_handler` can be used to set up a mount point.
+
+Example
+-------
+
+.. code:: c
+    #include "lfs.h"
+
+    // configuration of the filesystem is provided by this struct
+    const struct lfs_config cfg = {
+        // block device operations
+        .read  = user_provided_block_device_read,
+        .prog  = user_provided_block_device_prog,
+        .erase = user_provided_block_device_erase,
+        .sync  = user_provided_block_device_sync,
+
+        // block device configuration
+        .read_size = 16,
+        .prog_size = 16,
+        .block_size = 4096,
+        .block_count = 128,
+        .cache_size = 16,
+        .lookahead_size = 16,
+        .block_cycles = 500,
+    };
+
+    // mount the filesystem
+    int err = lfs_mount(&lfs, &cfg);
+
+    // reformat if we can't mount the filesystem
+    // this should only happen on the first boot
+    if (err) {
+        lfs_format(&lfs, &cfg);
+        lfs_mount(&lfs, &cfg);
+    }
+
+    lv_littlefs_set_handler(&lfs);
+
+
+API
+---
+
+
--- a/src/libs/fsdrv/lv_fs_littlefs.c
+++ b/src/libs/fsdrv/lv_fs_littlefs.c
@@ -1,0 +1,188 @@
+#include "../../../lvgl.h"
+#if LV_USE_FS_LITTLEFS
+
+#include "lfs.h"
+#include "../../core/lv_global.h"
+
+typedef struct LittleFile {
+    lfs_file_t file;
+} LittleFile;
+
+/**********************
+ *  STATIC PROTOTYPES
+ **********************/
+static void * fs_open(lv_fs_drv_t * drv, const char * path, lv_fs_mode_t mode);
+static lv_fs_res_t fs_close(lv_fs_drv_t * drv, void * file_p);
+static lv_fs_res_t fs_read(lv_fs_drv_t * drv, void * file_p, void * buf, uint32_t btr, uint32_t * br);
+static lv_fs_res_t fs_write(lv_fs_drv_t * drv, void * file_p, const void * buf, uint32_t btw, uint32_t * bw);
+static lv_fs_res_t fs_seek(lv_fs_drv_t * drv, void * file_p, uint32_t pos, lv_fs_whence_t whence);
+static lv_fs_res_t fs_tell(lv_fs_drv_t * drv, void * file_p, uint32_t * pos_p);
+
+void lv_littlefs_set_handler(lfs_t * lfs)
+{
+    lv_fs_drv_t * drv = lv_fs_get_drv(LV_FS_LITTLEFS_LETTER);
+    drv->user_data = lfs;
+}
+
+/**
+ * Register a driver for the LittleFS File System interface
+ */
+void lv_fs_littlefs_init(void)
+{
+    lv_fs_drv_t * fs_drv = &(LV_GLOBAL_DEFAULT()->littlefs_fs_drv);
+    lv_fs_drv_init(fs_drv);
+
+    fs_drv->letter = LV_FS_LITTLEFS_LETTER;
+    fs_drv->open_cb = fs_open;
+    fs_drv->close_cb = fs_close;
+    fs_drv->read_cb = fs_read;
+    fs_drv->write_cb = fs_write;
+    fs_drv->seek_cb = fs_seek;
+    fs_drv->tell_cb = fs_tell;
+
+    fs_drv->dir_close_cb = NULL;
+    fs_drv->dir_open_cb = NULL;
+    fs_drv->dir_read_cb = NULL;
+
+    lv_fs_drv_register(fs_drv);
+}
+
+/**********************
+ *   STATIC FUNCTIONS
+ **********************/
+
+/**
+ * Open a file
+ * @param drv       pointer to a driver where this function belongs
+ * @param path      path to the file beginning with the driver letter (e.g. S:/folder/file.txt)
+ * @param mode      read: FS_MODE_RD, write: FS_MODE_WR, both: FS_MODE_RD | FS_MODE_WR
+ * @return          a file descriptor or NULL on error
+ */
+static void * fs_open(lv_fs_drv_t * drv, const char * path, lv_fs_mode_t mode)
+{
+    LV_UNUSED(drv);
+
+    int flags;
+    if(mode == LV_FS_MODE_WR)
+        flags = LFS_O_WRONLY;
+    else if(mode == LV_FS_MODE_RD)
+        flags = LFS_O_RDONLY;
+    else if(mode == (LV_FS_MODE_WR | LV_FS_MODE_RD))
+        flags = LFS_O_RDWR;
+
+    LittleFile * lf = (LittleFile *)lv_malloc(sizeof(LittleFile));
+    LV_ASSERT_NULL(lf);
+
+    lfs_t * lfs = (lfs_t *)drv->user_data;
+    int err = lfs_file_open(lfs, &lf->file, path, flags);
+    if(err) {
+        return NULL;
+    }
+
+    return (void *)lf;
+}
+
+/**
+ * Close an opened file
+ * @param drv       pointer to a driver where this function belongs
+ * @param file_p    pointer to a file_t variable. (opened with fs_open)
+ * @return          LV_FS_RES_OK: no error or  any error from @lv_fs_res_t enum
+ */
+static lv_fs_res_t fs_close(lv_fs_drv_t * drv, void * file_p)
+{
+    LV_UNUSED(drv);
+    LittleFile * lf = (LittleFile *)file_p;
+
+    lfs_t * lfs = (lfs_t *)drv->user_data;
+    lfs_file_close(lfs, &lf->file);
+    lv_free(lf);
+
+    return LV_FS_RES_OK;
+}
+
+/**
+ * Read data from an opened file
+ * @param drv       pointer to a driver where this function belongs
+ * @param file_p    pointer to a file_t variable.
+ * @param buf       pointer to a memory block where to store the read data
+ * @param btr       number of Bytes To Read
+ * @param br        the real number of read bytes (Byte Read)
+ * @return          LV_FS_RES_OK: no error or any error from @lv_fs_res_t enum
+ */
+static lv_fs_res_t fs_read(lv_fs_drv_t * drv, void * file_p, void * buf, uint32_t btr, uint32_t * br)
+{
+    LV_UNUSED(drv);
+    LittleFile * lf = (LittleFile *)file_p;
+
+    lfs_t * lfs = (lfs_t *)drv->user_data;
+    *br = lfs_file_read(lfs, &lf->file, (uint8_t *)buf, btr);
+
+    return (int32_t)(*br) < 0 ? LV_FS_RES_UNKNOWN : LV_FS_RES_OK;
+}
+
+/**
+ * Write into a file
+ * @param drv       pointer to a driver where this function belongs
+ * @param file_p    pointer to a file_t variable
+ * @param buf       pointer to a buffer with the bytes to write
+ * @param btw       Bytes To Write
+ * @param bw        the number of real written bytes (Bytes Written)
+ * @return          LV_FS_RES_OK: no error or  any error from @lv_fs_res_t enum
+ */
+static lv_fs_res_t fs_write(lv_fs_drv_t * drv, void * file_p, const void * buf, uint32_t btw, uint32_t * bw)
+{
+    LV_UNUSED(drv);
+    LittleFile * lf = (LittleFile *)file_p;
+
+    lfs_t * lfs = (lfs_t *)drv->user_data;
+    *bw = lfs_file_write(lfs, &lf->file, (uint8_t *)buf, btw);
+
+    return (int32_t)(*bw) < 0 ? LV_FS_RES_UNKNOWN : LV_FS_RES_OK;
+}
+
+/**
+ * Set the read write pointer. Also expand the file size if necessary.
+ * @param drv       pointer to a driver where this function belongs
+ * @param file_p    pointer to a file_t variable. (opened with fs_open )
+ * @param pos       the new position of read write pointer
+ * @param whence    tells from where to interpret the `pos`. See @lv_fs_whence_t
+ * @return          LV_FS_RES_OK: no error or any error from @lv_fs_res_t enum
+ */
+static lv_fs_res_t fs_seek(lv_fs_drv_t * drv, void * file_p, uint32_t pos, lv_fs_whence_t whence)
+{
+    LV_UNUSED(drv);
+    int mode;
+    if(whence == LV_FS_SEEK_SET)
+        mode = LFS_SEEK_SET;
+    else if(whence == LV_FS_SEEK_CUR)
+        mode = LFS_SEEK_CUR;
+    else if(whence == LV_FS_SEEK_END)
+        mode = LFS_SEEK_END;
+
+    LittleFile * lf = (LittleFile *)file_p;
+
+    lfs_t * lfs = (lfs_t *)drv->user_data;
+    int rc = lfs_file_seek(lfs, &lf->file, pos, mode);
+
+    return rc < 0 ? LV_FS_RES_UNKNOWN : LV_FS_RES_OK;
+}
+
+/**
+ * Give the position of the read write pointer
+ * @param drv       pointer to a driver where this function belongs
+ * @param file_p    pointer to a file_p variable
+ * @param pos_p     pointer to store the result
+ * @return          LV_FS_RES_OK: no error or any error from @lv_fs_res_t enum
+ */
+static lv_fs_res_t fs_tell(lv_fs_drv_t * drv, void * file_p, uint32_t * pos_p)
+{
+    LV_UNUSED(drv);
+    LittleFile * lf = (LittleFile *)file_p;
+
+    lfs_t * lfs = (lfs_t *)drv->user_data;
+    *pos_p = lfs_file_tell(lfs, &lf->file);
+
+    return (int32_t)(*pos_p) < 0 ? LV_FS_RES_UNKNOWN : LV_FS_RES_OK;
+}
+
+#endif
--- a/Kconfig
+++ b/Kconfig
@@ -1057,6 +1057,13 @@ menu "LVGL configuration"
 			default 0
 			depends on LV_USE_FS_MEMFS
 
+		config LV_USE_FS_LITTLEFS
+			bool "File system on top of littlefs API"
+		config LV_FS_LITTLEFS_LETTER
+			int "Set an upper cased letter on which the drive will accessible (e.g. 'A' i.e. 65)"
+			default 0
+			depends on LV_USE_FS_LITTLEFS
+
 		config LV_USE_LODEPNG
 			bool "PNG decoder library"
 
--- a/docs/libs/index.rst
+++ b/docs/libs/index.rst
@@ -22,3 +22,4 @@
     rlottie
     ffmpeg
     rle
+    lfs
--- a/examples/porting/lv_port_fs_template.c
+++ b/examples/porting/lv_port_fs_template.c
@@ -64,7 +64,6 @@ void lv_port_fs_init(void)
      * Register the file system interface in LVGL
      *--------------------------------------------------*/
 
-    /*Add a simple drive to open images*/
     static lv_fs_drv_t fs_drv;
     lv_fs_drv_init(&fs_drv);
 
@@ -195,8 +194,8 @@ static lv_fs_res_t fs_seek(lv_fs_drv_t * drv, void * file_p, uint32_t pos, lv_fs
 /**
  * Give the position of the read write pointer
  * @param drv       pointer to a driver where this function belongs
- * @param file_p    pointer to a file_t variable.
- * @param pos_p     pointer to to store the result
+ * @param file_p    pointer to a file_t variable
+ * @param pos_p     pointer to store the result
  * @return          LV_FS_RES_OK: no error or  any error from @lv_fs_res_t enum
  */
 static lv_fs_res_t fs_tell(lv_fs_drv_t * drv, void * file_p, uint32_t * pos_p)
--- a/lv_conf_template.h
+++ b/lv_conf_template.h
@@ -641,6 +641,12 @@
     #define LV_FS_MEMFS_LETTER '\0'     /*Set an upper cased letter on which the drive will accessible (e.g. 'A')*/
 #endif
 
+/*API for LittleFs. */
+#define LV_USE_FS_LITTLEFS 0
+#if LV_USE_FS_LITTLEFS
+    #define LV_FS_LITTLEFS_LETTER '\0'     /*Set an upper cased letter on which the drive will accessible (e.g. 'A')*/
+#endif
+
 /*LODEPNG decoder library*/
 #define LV_USE_LODEPNG 0
 
--- a/src/core/lv_global.h
+++ b/src/core/lv_global.h
@@ -160,6 +160,10 @@ typedef struct _lv_global_t {
     lv_fs_drv_t win32_fs_drv;
 #endif
 
+#if LV_USE_FS_LITTLEFS
+    lv_fs_drv_t littlefs_fs_drv;
+#endif
+
 #if LV_USE_FREETYPE
     struct _lv_freetype_context_t * ft_context;
 #endif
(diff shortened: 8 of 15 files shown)


src/libs/fsdrv/lv_fs_littlefs.c
```diff
@@ -0,0 +1,165 @@
+#include "../../../lvgl.h"
+#if LV_USE_FS_LITTLEFS != '\0'
+
```
terry0012, line 3: #if LV_USE_FS_LITTLEFS

src/libs/fsdrv/lv_fs_littlefs.c
```diff
@@ -0,0 +1,165 @@
...
+#include <LittleFS.h>
+#include "../../core/lv_global.h"
```
terry0012, line 5: where is LittleFS.h ?
need 3rd library support?

src/libs/fsdrv/lv_fs_littlefs.c
```diff
@@ -0,0 +1,165 @@
...
+
+/**
+ * Register a driver for the LittleFS File System interface
+ */
+void lv_port_littlefs_init(void)
+{
+    fs_init();
+
+    /*Add a simple drive to open images*/
+    static lv_fs_drv_t fs_drv;
```
terry0012: use global var for fs_drv, reference to fatfs.

`lv_fs_drv_t * fs_drv_p = &(LV_GLOBAL_DEFAULT()->littlefs_fs_drv);`

and define `littlefs_fs_drv`   in `_lv_global_t` as below:

```
#if LV_USE_FS_LITTLEFS
    lv_fs_drv_t littlefs_fs_drv;
#endif
```

src/libs/fsdrv/lv_fs_littlefs.c
```diff
@@ -0,0 +1,165 @@
...
+static lv_fs_res_t fs_close(lv_fs_drv_t * drv, void * file_p);
+static lv_fs_res_t fs_read(lv_fs_drv_t * drv, void * file_p, void * buf, uint32_t btr, uint32_t * br);
+static lv_fs_res_t fs_write(lv_fs_drv_t * drv, void * file_p, const void * buf, uint32_t btw, uint32_t * bw);
+static lv_fs_res_t fs_seek(lv_fs_drv_t * drv, void * file_p, uint32_t pos, lv_fs_whence_t whence);
+static lv_fs_res_t fs_tell(lv_fs_drv_t * drv, void * file_p, uint32_t * pos_p);
+
+/**
+ * Register a driver for the LittleFS File System interface
+ */
+void lv_port_littlefs_init(void)
```
PGNetHun: You need to init this driver in `lv_init.c` like FAT FS driver: https://github.com/lvgl/lvgl/blob/master/src/lv_init.c#L248

src/libs/fsdrv/lv_fs_littlefs.c
```diff
@@ -0,0 +1,165 @@
...
+    return (int32_t)(*br) < 0 ? LV_FS_RES_UNKNOWN : LV_FS_RES_OK;
+}
+
+/**
+ * Write into a file
+ * @param drv       pointer to a driver where this function belongs
+ * @param file_p    pointer to a file_t variable
+ * @param buf       pointer to a buffer with the bytes to write
+ * @param btw       Bytes To Write
+ * @param bw        the number of real written bytes (Bytes Written). NULL if unused.
```
PGNetHun: If `bw` parameter can be `NULL`, then - I guess - you need NULL checking in the `return` line

src/libs/fsdrv/lv_fs_littlefs.c
```diff
@@ -0,0 +1,165 @@
...
+    static lv_fs_drv_t fs_drv;
+    lv_fs_drv_init(&fs_drv);
+
+    fs_drv.letter = LV_FS_LITTLEFS_LETTER;
+    fs_drv.open_cb = fs_open;
+    fs_drv.close_cb = fs_close;
+    fs_drv.read_cb = fs_read;
+    fs_drv.write_cb = fs_write;
+    fs_drv.seek_cb = fs_seek;
+    fs_drv.tell_cb = NULL;
```
PGNetHun: Please implement `tell_cb` also, as many internal code (and 3rd party libraries) use it, for example the image decoders:
- https://github.com/lvgl/lvgl/blob/master/src/libs/gif/gifdec.c#L779
- https://github.com/lvgl/lvgl/blob/master/src/libs/libpng/lv_libpng.c#L205
- https://github.com/lvgl/lvgl/blob/master/src/libs/lodepng/lodepng.c#L374
- https://github.com/lvgl/lvgl/blob/master/src/libs/libjpeg_turbo/lv_libjpeg_turbo.c#L218

PGNetHun (changes requested):

PGNetHun:
@PierreRambaud ,
Please after the commit is added, and you think it can be reviewed, then re-request review from every reviewer in github:
image

src/libs/fsdrv/lv_fs_littlefs.c
PierreRambaud: Yes it's based on: https://github.com/esp8266/Arduino/blob/master/libraries/LittleFS/src/LittleFS.h
I don't know if there is another available implementation

src/libs/fsdrv/lv_fs_littlefs.c
PierreRambaud: Thanks, I implement that but I've to do more test to be sure everything's ok :)

src/libs/fsdrv/lv_fs_littlefs.c
PierreRambaud: Thank you for your help :pray:

src/libs/fsdrv/lv_fs_littlefs.c
PGNetHun: we should not use and refer Arduino related code (or in this case LittleFS arduino wrapper) in LVGL, because LVGL can be used with other OS / frameworks, too. For example with FreeRTOS, unix, MicroPython, etc…

If we want to use a 3rd party library, then refer to (or copy) it’s original (and possibly platform-independent) code. 
In case of LittleFS: https://github.com/littlefs-project/littlefs

For example in MicroPython framework they use the original code: https://github.com/micropython/micropython/tree/master/lib/littlefs

PierreRambaud: I'm gonna use the littlefs project you linked! You're totally right about the Arduino, didn't know what's the process for this kind of dependency

kisvegabor: Note that we shouldn't add LittleFS itself to LVGL (similarly we don't add FATFS either). It's the user responsibility to add LittleFS anyhow he/she wants. E.g. by a component manager or manually.

src/libs/fsdrv/lv_fs_littlefs.c
```diff
@@ -0,0 +1,181 @@
...
+    LittleFile * lf = (LittleFile *)file_p;
+    lf->file.seek(pos, mode);
+    return LV_FS_RES_OK;
+}
+
+/**
+ * Give the position of the read write pointer
+ * @param drv       pointer to a driver where this function belongs
+ * @param file_p    pointer to a FIL variable.
+ * @param pos_p     pointer to to store the result
```
PGNetHun: double _to_ words

src/libs/fsdrv/lv_fs_littlefs.c
```diff
@@ -0,0 +1,181 @@
...
+    lf->file.close();
+
+    delete lf;
+    return LV_FS_RES_OK;
+}
+
+/**
+ * Read data from an opened file
+ * @param drv       pointer to a driver where this function belongs
+ * @param file_p    pointer to a file_t variable.
```
PGNetHun, line 106: no need for the ending point, as other lines don't contain it either

src/libs/fsdrv/lv_fs_littlefs.c
```diff
@@ -0,0 +1,181 @@
...
+
+    LittleFile * lf = (LittleFile *)file_p;
+    lf->file.seek(pos, mode);
+    return LV_FS_RES_OK;
+}
+
+/**
+ * Give the position of the read write pointer
+ * @param drv       pointer to a driver where this function belongs
+ * @param file_p    pointer to a FIL variable.
```
PGNetHun: description of `file_p` should be: _pointer to a file_t variable_

src/libs/fsdrv/lv_fs_littlefs.c
```diff
@@ -0,0 +1,181 @@
...
+ * @param pos_p     pointer to to store the result
+ * @return LV_FS_RES_OK: no error, the file is read
```
PGNetHun: no need for: _the file is read_

PGNetHun (changes requested):
Just some minor comment fixes needed, and OK.

src/libs/fsdrv/lv_fs_littlefs.c
PierreRambaud: Woot, implementing the based lfs.h library is harder than expected :sweat_smile: 
I'm struggling with platformio:
```
.pio/libdeps/m5stack-core2/lfs/runners/test_runner.c:23:10: fatal error: execinfo.h: No such file or directory

******************************************************************
* Looking for execinfo.h dependency? Check our library registry!
*
* CLI  > platformio lib search "header:execinfo.h"
* Web  > https://registry.platformio.org/search?q=header:execinfo.h
*
******************************************************************

 #include <execinfo.h>
          ^~~~~~~~~~~~
compilation terminated.
```

Also, is there a way to declare custom lfs configuration? I want to make lfs_config more generic and don't know if it should be configured in lv_conf, or elsewhere :thinking: 
```c
static void fs_init(void)
{
  const struct lfs_config cfg = {};

  lfs_mount(&lfs, &cfg);
}
```

Because of my M5Stack Core2 configuration, I think I have to backport this configuration to be sure everything is working well with the new implementation, I would like to reuse the same configuration as it's defined in this file
https://github.com/esp8266/Arduino/blob/master/libraries/LittleFS/src/LittleFS.h#L62


At least, I update the code until I find a way to make it work on my project ^^

terry0012 (approved):

src/libs/fsdrv/lv_fs_littlefs.c
```diff
@@ -0,0 +1,191 @@
...
+#include "../../core/lv_global.h"
+
+struct LittleFile {
+    lfs_file_t file;
+};
+
+/**********************
+ *  STATIC PROTOTYPES
+ **********************/
+static lfs_t lfs;
```
XuNeo: Please move it to lv_logbal too.
You can add macro to make it easier to use.
```
#define lfs LV_GLOBAL_DEFAULT()->lfs

```

src/libs/fsdrv/lv_fs_littlefs.c
kisvegabor: I think in LVGL we should assume that LittleFS is already there and initialized. So all the special configs should remain outside of LVGL.

src/libs/fsdrv/lv_fs_littlefs.c
kisvegabor: As described [here](https://github.com/lvgl/lvgl/pull/5562#discussion_r1477777390), it think it should not be part of LVGL. If you need to pass a reference to `lfs`, we should add a function like `lv_littlefs_set_handler()` that the user can call before actually using LittleFS with LVGL.

src/libs/fsdrv/lv_fs_littlefs.c
PierreRambaud: I can't find an easy way to use what's defined in Arduino's configuration, everything is private/protected :(
If we assume that LVGL should not be aware of the configuration, maybe in a first step we can use the "LittleFS.h" dependency ?
Regarding this [comment](https://github.com/lvgl/lvgl/pull/5562#discussion_r1477779716)  what do you mean by `lv_littlefs_set_handler`? Is there an example somewhere?

kisvegabor:
Sorry for the late reply Pierre. I've just check out how LittleFS is integrated into Arduino and I cannot see a good way to create a wrapper around it. As a reference this is in the example to write to a file:
```c

writeFile(LITTLEFS, "/mydir/hello2.txt", "Hello2");

...

void writeFile(fs::FS &fs, const char * path, const char * message){
    Serial.printf("Writing file: %s\r\n", path);

    File file = fs.open(path, FILE_WRITE);
    if(!file){
        Serial.println("- failed to open file for writing");
        return;
    }
    if(file.print(message)){
        Serial.println("- file written");
    } else {
        Serial.println("- write failed");
    }
    file.close();
}
```

In Arduino is it common that file I/O function are used via a class? (fs.open, fs.close, fs.write etc). If so, maybe what we need is a generic Arduino driver.

PierreRambaud:
> Member

Yes, it's common to use classes in [Arduino libraries](https://www.arduino.cc/reference/en/libraries/).

I'm afraid that I don't understand what you mean by having a generic Arduino driver.

I usually play with [SD](https://www.arduino.cc/reference/en/libraries/sd/), [EEPROM](https://docs.arduino.cc/learn/built-in-libraries/eeprom), and LittleFS. SD uses its own [File class](https://www.arduino.cc/reference/en/libraries/sd/name/), and LittleFS uses `fs::File`.

What about having three drivers:
- based on `LittleFS.h` called arduino_littlefs
- based on `SD.h` called arduino_sd
- based on `lfs.h` called littlefs

kisvegabor:
I hoped that all FS related drivers are using the same concept of `fs::File`. I see that it's not that case :slightly_frowning_face: 

> What about having three drivers:

It's getting complicated. However as we have too many issues related to File system and Arduino, I'm afraid our best bet is to add all these. 

Let's have a PR for `lfs.h` and an other for the 2 Arduino libs. 

And now I'm lost a little bit. Which one we are doing in this PR? :smile:

PierreRambaud:
> I hoped that all FS related drivers are using the same concept of `fs::File`. I see that it's not that case 🙁
> 
> > What about having three drivers:
> 
> It's getting complicated. However as we have too many issues related to File system and Arduino, I'm afraid our best bet is to add all these.
> 
> Let's have a PR for `lfs.h` and an other for the 2 Arduino libs.
> 
> And now I'm lost a little bit. Which one we are doing in this PR? 😄

This one will be about the `lfs.h` implementation. One thing I'm not sure about is what you said a few days ago regarding  `lv_littlefs_set_handler`, what do you have in mind?
I'm sorry if my questions seem stupid, but I'm not a C nor a C++ developer, and I want to understand what I'm trying to do properly.
When this one is ok, I'll be glad to do the ArduinoSD and ArduinoLittleFS driver for you :pray:

kisvegabor:
> This one will be about the lfs.h implementation. 

:+1: 

> One thing I'm not sure about is what you said a few days ago regarding lv_littlefs_set_handler, what do you have in mind?

E.g. in `fs_read` we have [this](https://github.com/lvgl/lvgl/pull/5562/files#diff-cfb0a29e9176c1b2d437ed93288081d5840419f8983c61d80e0371d9777e93acR122): `*br = lfs_file_read(&lfs, &lf->file, (uint8_t *) buf, btr);` `lfs` is mounted and initialized by a single config [here](https://github.com/lvgl/lvgl/pull/5562/files#diff-cfb0a29e9176c1b2d437ed93288081d5840419f8983c61d80e0371d9777e93acR54-R59). 

I wonder if it were more flexible and cleaner if we let the user to mount and configure lfs as he wants, and just pass `&lfs` to our LittleFS driver. Something like:
```c
void lv_littlefs_set_handler(lfs_t * lfs)
{
  lv_fs_drv_t * drv = lv_fs_get_drv(LV_FS_LITTLEFS_LETTER);
  drv->user_data = lfs;
}

static lv_fs_res_t fs_read(lv_fs_drv_t * drv, void * file_p, void * buf, uint32_t btr, uint32_t * br)
{
    lfs_t * lfs = drv->user_data; //Get the saved driver handler. I don't what it's type is so I used void * as an example.
    LittleFile * lf = (LittleFile *)file_p;

    *br = lfs_file_read(lfs, &lf->file, (uint8_t *) buf, btr);

    return (int32_t)(*br) < 0 ? LV_FS_RES_UNKNOWN : LV_FS_RES_OK;
}
```

What do you think?


> I'm sorry if my questions seem stupid, but I'm not a C nor a C++ developer, and I want to understand what I'm trying to do properly.

Not stupid at all! Actually I'm not sure if `lv_littlefs_set_handler` is a good idea andI'm really curious what you think!

> When this one is ok, I'll be glad to do the ArduinoSD and ArduinoLittleFS driver for you 🙏

Thank you in advance :blush:

PierreRambaud:
> E.g. in `fs_read` we have [this](https://github.com/lvgl/lvgl/pull/5562/files#diff-cfb0a29e9176c1b2d437ed93288081d5840419f8983c61d80e0371d9777e93acR122): `*br = lfs_file_read(&lfs, &lf->file, (uint8_t *) buf, btr);` `lfs` is mounted and initialized by a single config [here](https://github.com/lvgl/lvgl/pull/5562/files#diff-cfb0a29e9176c1b2d437ed93288081d5840419f8983c61d80e0371d9777e93acR54-R59).
> 
> I wonder if it were more flexible and cleaner if we let the user to mount and configure lfs as he wants, and just pass `&lfs` to our LittleFS driver. Something like:
> 
> ```c
> void lv_littlefs_set_handler(lfs_t * lfs)
> {
>   lv_fs_drv_t * drv = lv_fs_get_drv(LV_FS_LITTLEFS_LETTER);
>   drv->user_data = lfs;
> }
> 
> static lv_fs_res_t fs_read(lv_fs_drv_t * drv, void * file_p, void * buf, uint32_t btr, uint32_t * br)
> {
>     lfs_t * lfs = drv->user_data; //Get the saved driver handler. I don't what it's type is so I used void * as an example.
>     LittleFile * lf = (LittleFile *)file_p;
> 
>     *br = lfs_file_read(lfs, &lf->file, (uint8_t *) buf, btr);
> 
>     return (int32_t)(*br) < 0 ? LV_FS_RES_UNKNOWN : LV_FS_RES_OK;
> }
> ```
> 
> What do you think?
> 
> > I'm sorry if my questions seem stupid, but I'm not a C nor a C++ developer, and I want to understand what I'm trying to do properly.
> 
> Not stupid at all! Actually I'm not sure if `lv_littlefs_set_handler` is a good idea andI'm really curious what you think!
> 
> > When this one is ok, I'll be glad to do the ArduinoSD and ArduinoLittleFS driver for you 🙏
> 
> Thank you in advance 😊

From what I know about C, I think it's a good idea. So I removed the `fs_init` method, but I think the behavior around `lv_littlefs_set_handler` should be documented somewhere. What about `docs/integration/drivers` directory?

src/libs/fsdrv/lv_fs_littlefs.c
```diff
@@ -0,0 +1,190 @@
...
+    lv_fs_drv_t * drv = lv_fs_get_drv(LV_FS_LITTLEFS_LETTER);
+    drv->user_data = lfs;
+}
+
+/**
+ * Register a driver for the LittleFS File System interface
+ */
+void lv_fs_littlefs_init(void)
+{
+    /*Add a simple drive to open images*/
```
PGNetHun: No need for this comment, as not only images can be opened through LVGL FS API, but fonts too

src/libs/fsdrv/lv_fs_littlefs.c
```diff
@@ -0,0 +1,190 @@
...
+
+    return rc < 0 ? LV_FS_RES_UNKNOWN : LV_FS_RES_OK;
+}
+
+/**
+ * Give the position of the read write pointer
+ * @param drv       pointer to a driver where this function belongs
+ * @param file_p    pointer to a file_p variable
+ * @param pos_p     pointer to store the result
+ * @return LV_FS_RES_OK: no error
```
PGNetHun: Could you please fix indentation of `@return` commit?

src/libs/fsdrv/lv_fs_littlefs.c
```diff
@@ -0,0 +1,190 @@
+#include "../../../lvgl.h"
+#if LV_USE_FS_LITTLEFS
+
+#include <lfs.h>
```
PGNetHun: I wonder where this `lfs.h` comes from.

To help other developers how to install and use LittleFS, could you please add new document to the 3rd party libraries folder? 
https://github.com/lvgl/lvgl/tree/master/docs/libs

It is enough to write down how to install / clone LittleFS, like in `libpng` doc:
https://docs.lvgl.io/master/libs/libpng.html

You can also mention in File systems doc that LittleFS is now supported, too: https://github.com/lvgl/lvgl/blob/master/docs/libs/fs.rst

PGNetHun:
> From what I know about C, I think it's a good idea. So I removed the `fs_init` method, but I think the behavior around `lv_littlefs_set_handler` should be documented somewhere. What about `docs/integration/drivers` directory?

Yes, a new document about this LittleFS integration is a good idea, and I think it could be added to 3rd party libraries folder.
Please see my comment: https://github.com/lvgl/lvgl/pull/5562#discussion_r1491619560

Is that OK for other reviewers?

src/libs/fsdrv/lv_fs_littlefs.c
PierreRambaud: I'm not a good English writer (nor a french writer as well :( ), but if you think something is mandatory and missing inside the documentation, tell me :pray:

docs/libs/fs.rst
```diff
@@ -14,6 +14,7 @@ LVG has built in support for:
 - POSIX (Linux and Windows using POSIX function .e.g ``open``, ``read``)
 - WIN32 (Windows using Win32 API function .e.g ``CreateFileA``, ``ReadFile``)
 - MEMFS (read a file from a memory buffer)
+- LITTLEFS (on-board file system for your ESP32 & ESP8266)
```
PGNetHun: as I know LittleFS is platform independent, so you can remove "ESP32 & ESP8266", OR replace whole sentence to:
_A little fail-safe filesystem designed for microcontrollers._

(first line of https://github.com/littlefs-project/littlefs/blob/master/README.md )

docs/libs/lfs.rst
PGNetHun, line 1: Please mention this `lfs` file in [docs/libs/index.rst](https://github.com/lvgl/lvgl/blob/master/docs/libs/index.rst), so it will show up in [3rd party libraries index page](https://docs.lvgl.io/master/libs/index.html)

PGNetHun (changes requested):
Just 2 small doc changes, and it's OK from my side.

docs/libs/lfs.rst
PierreRambaud: Thank didn't know about this. I better understand how the documentation is working :muscle:

PGNetHun (dismissed):
OK, approved.
Thank you for the hard work!

IAMMX:
Add the relevant configuration in `Kconfig`

PGNetHun (dismissed):

PierreRambaud:
I'm sorry, I don't understand why the test is failing as I didn't touch it :sweat_smile:

PGNetHun:
> I'm sorry, I don't understand why the test is failing as I didn't touch it 😅

That failing "Verify code formatting" check is not your fault, it is already fixed, so I re-ran it.

src/libs/fsdrv/lv_fs_littlefs.c
```diff
@@ -0,0 +1,188 @@
...
+        flags = LFS_O_RDWR;
+
+    lfs_t * lfs = drv->user_data;
+    lfs_file_t file;
+    int err = lfs_file_open(&lfs, &file, path, flags);
+    if(err) {
+        return NULL;
+    }
+
+    LittleFile * lf = new LittleFile{file};
```
XuNeo: Can we use pure C code?
```c
    LittleFile * lf = lv_malloc(sizeof(*if));
    LV_ASSERT_NULL(lf);

    int err = lfs_file_open(&lfs, &lf->file, path, flags);
    if(err) {
        return NULL;
    }

**Review by XuNeo — COMMENTED** [review]
Thank you. I just found two minor issues. Is it OK to write C++ code in `.c` file?
I think it can be replace by lv_malloc and lv_free.

File: src/libs/fsdrv/lv_fs_littlefs.c (hunk 16 above)
Comment (PierreRambaud) [reply src/libs/fsdrv/lv_fs_littlefs.c]: Nice catch :scream: 
Totally forget to change this part when moving to c :(

File: src/libs/fsdrv/lv_fs_littlefs.c [hunk 17]
```diff
@@ -0,0 +1,188 @@
...
+ * @return          LV_FS_RES_OK: no error or  any error from @lv_fs_res_t enum
+ */
+static lv_fs_res_t fs_close(lv_fs_drv_t * drv, void * file_p)
+{
+    LV_UNUSED(drv);
+    LittleFile * lf = (LittleFile *)file_p;
+
+    lfs_t * lfs = drv->user_data;
+    lfs_file_close(&lfs, &lf->file);
+    delete lf;
```
XuNeo: Should use lv_free here correspondingly. :)

XuNeo (dismissed):

src/libs/fsdrv/lv_fs_littlefs.c
```diff
@@ -0,0 +1,188 @@
...
+    drv->user_data = lfs;
+}
+
+/**
+ * Register a driver for the LittleFS File System interface
+ */
+void lv_fs_littlefs_init(void)
+{
+    lv_fs_drv_t * fs_drv_p = &(LV_GLOBAL_DEFAULT()->littlefs_fs_drv);
+    lv_fs_drv_init(&fs_drv);
```
lhdjply: May I ask where this variable `fs_drv` comes from, is not defined `fs_drv_p`?

src/libs/fsdrv/lv_fs_littlefs.c
```diff
@@ -0,0 +1,188 @@
...
+ * @param file_p    pointer to a file_t variable. (opened with fs_open)
+ * @return          LV_FS_RES_OK: no error or  any error from @lv_fs_res_t enum
+ */
+static lv_fs_res_t fs_close(lv_fs_drv_t * drv, void * file_p)
+{
+    LV_UNUSED(drv);
+    LittleFile * lf = (LittleFile *)file_p;
+
+    lfs_t * lfs = drv->user_data;
+    lfs_file_close(&lfs, &lf->file);
```
lhdjply: Is it not a pointer variable that is defined? Why use the fetch address symbol? A lot of the pr is used in this way.

lhdjply:
Before requesting pr, please run the code locally.

src/libs/fsdrv/lv_fs_littlefs.c
lhdjply:
```c
lv_fs_drv_t * fs_drv = &(LV_GLOBAL_DEFAULT()->littlefs_fs_drv); 
lv_fs_drv_init(fs_drv); 
fs_drv->letter = LV_FS_LITTLEFS_LETTER; 
fs_drv->open_cb = fs_open; 
fs_drv->close_cb = fs_close;
 fs_drv->read_cb = fs_read; 
fs_drv->write_cb = fs_write; 
fs_drv->seek_cb = fs_seek;
 fs_drv->tell_cb = fs_tell; 
fs_drv->dir_close_cb = NULL; 
fs_drv->dir_open_cb = NULL; 
fs_drv->dir_read_cb = NULL; 
lv_fs_drv_register(fs_drv);
```

PierreRambaud:
> Before requesting pr, please run the code locally.

I did it with the latest changes. Sorry for the inconvenience, it's a little bit hard to use lfs.h without Arduino for me :(

src/libs/fsdrv/lv_fs_littlefs.c
lhdjply: Please modify according to the examples I provided.

src/libs/fsdrv/lv_fs_littlefs.c
```diff
@@ -0,0 +1,188 @@
...
+
+    int flags;
+    if(mode == LV_FS_MODE_WR)
+        flags = LFS_O_WRONLY;
+    else if(mode == LV_FS_MODE_RD)
+        flags = LFS_O_RDONLY;
+    else if(mode == (LV_FS_MODE_WR | LV_FS_MODE_RD))
+        flags = LFS_O_RDWR;
+
+    LittleFile * lf = lv_malloc(sizeof(*lf));
```
lhdjply:
```c
LittleFile * lf = （LittleFile *）lv_malloc(sizeof(LittleFile));
```

XuNeo: Hi, there's no need to cast (void*) pointer here. The original code looks good to me.

XuNeo (dismissed):

XuNeo (dismissed):
Thank you. :)

src/libs/fsdrv/lv_fs_littlefs.c
PierreRambaud: I need to change a few things to be sure everything is still ok. But I think @lhdjply is right

PierreRambaud:
Still working on it, finalizing my tests :pray:

lhdjply (approved):

XuNeo (approved):
I don't think it's necessary to cast from void * pointer but it's OK.

PierreRambaud:
> I don't think it's necessary to cast from void * pointer but it's OK.

unfortunately, I had to do it, otherwise I receive this error at compile time `a value of type "void *" cannot be used to initialize an entity of type "LittleFile *"C/C++(144)`

PierreRambaud:
@kisvegabor Don't hesitate to tell me if something's missing. I prepare the Arduino LittleFS.h version on beside

kisvegabor (approved):
Thank you Pierre, it looks great! 

I know a lot of people who will be very happy when the the Arduino LFS and SD wrappers will be added too. :blush:

PierreRambaud:
> Contributor

I created a [PR](https://github.com/lvgl/lvgl/pull/6301)  for you, maybe you can also test it and tell me if everything is ok for you too :)  (We can also continue the discussion inside the PR instead of staying here)

PR #5596, opened 2024-02-05
feat(draw-sw): add simple Helium acceleration

PeterBee97:
Similar to Neon, this patch added Helium acceleration to all blending functions with minimal code. Still, the assembly code will produce slightly different result (normally 1bit error) from C rendering results due to difference in implementation, but visually similar to the eye:
cbimage (8)
The argb8888 dst is also supported with parallel division by integer newton iteration method. Thanks to tail predication and scatter-gathering functions, the code is conciser than Neon version and should also be more efficient. However, speed is not yet tested as the development was done on QEMU.
### Notes


PeterBee97:
Should be able to work together with arm2d. @GorgonMeducer please help review and test it! :)

PeterBee97:
> Do you consider adding other Helium acceleration? For example, for transform, rotate screen etc.
> 
I'll look into transforms later.

> By the way, would you please provide a macro option to only use arm-2d acceleration without the ASM version you added? (There is an option only to use your helium assembly version without arm-2d).
> 
I've moved arm2d header inclusion out of lv_blend_helium.h since arm2d already has its individual config LV_USE_DRAW_ARM2D_SYNC, so now the two accelerator can be toggled independently.

kisvegabor:
Amazing! However it's not entirely clear how to select between pure ASM and Arm2D. I guess you need to set `LV_USE_DRAW_SW_ASM LV_DRAW_SW_ASM_HELIUM` in both cases. And if you enable `LV_USE_DRAW_ARM2D_SYNC` Arm2D will be used, if it's disabled the plain ASM?

PeterBee97:
> Amazing! However it's not entirely clear how to select between pure ASM and Arm2D. I guess you need to set `LV_USE_DRAW_SW_ASM LV_DRAW_SW_ASM_HELIUM` in both cases. And if you enable `LV_USE_DRAW_ARM2D_SYNC` Arm2D will be used, if it's disabled the plain ASM?

That's what I previously thought, but since arm2d cannot override all of asm functions right now enabling LV_DRAW_SW_ASM_HELIUM means asm functions cannot be shutdown entirely (which is bad for Gabriel's test). Having the two options decoupled means we can have arm2d, or asm, or both (where asm is the backup). Or is there better idea to let arm2d work by itself? Need some help here. @kisvegabor

GorgonMeducer:
@PeterBee97  @kisvegabor 
> That's what I previously thought, but since arm2d cannot override all of asm functions right now enabling LV_DRAW_SW_ASM_HELIUM means asm functions cannot be shutdown entirely (which is bad for Gabriel's test). 

The original solution is the best one. I only need a macro option for testing purpose only.

PeterBee97:
Added macro ```USE_NATIVE_ASSEMBLY``` to disable functions internally.

kisvegabor (approved):
Clear, looks good to me. We need to document these somewhere though. 

I'll also do some tests today.

src/draw/sw/blend/helium/lv_blend_helium.h
```diff
@@ -0,0 +1,1310 @@
...
+#include "../../../../lv_conf_internal.h"
+
+/* detect whether helium is available based on arm compilers' standard */
+#if defined(__ARM_FEATURE_MVE) && __ARM_FEATURE_MVE
+
+#ifdef LV_DRAW_SW_HELIUM_CUSTOM_INCLUDE
+#include LV_DRAW_SW_HELIUM_CUSTOM_INCLUDE
+#endif
+
+#define USE_NATIVE_ASSEMBLY 1
```
kisvegabor: It should come from a compiler define.

Please also change the name to `LV_USE_NATIVE_ASSEMBLY`

kisvegabor (changes requested):
I've just found something.

src/draw/sw/blend/helium/lv_blend_helium.h
PeterBee97: Sure, moved to lv_conf.

GorgonMeducer:
Looks good for me.

kisvegabor (approved):
At this point I'm not sure ` #define  LV_USE_DRAW_SW_ASM     LV_DRAW_SW_ASM_...` is still required. As ASM implementation can overlap probably we should enable them differently. 

Let's improve it in an other PR later.

PR #5614, opened 2024-02-07
Updated Kconfig for default using with Espressif IDF

espzav:

We need to select some `Kconfig` options as default for using with ESP-IDF. It is not good for us to select these defaults in each project/example `sdkconfig.defaults`.

### DIFF
--- a/Kconfig
+++ b/Kconfig
@@ -69,9 +69,14 @@ menu "LVGL configuration"
 
 	endmenu
 
+	config LV_USE_CLIB
+		bool
+		default n
+
 	menu "Memory settings"
 		choice
 			prompt "Malloc functions source"
+			default LV_USE_CLIB_MALLOC if LV_USE_CLIB
 			default LV_USE_BUILTIN_MALLOC
 
 		config LV_USE_BUILTIN_MALLOC
@@ -93,6 +98,7 @@ menu "LVGL configuration"
 
 		choice
 			prompt "String functions source"
+			default LV_USE_CLIB_STRING if LV_USE_CLIB
 			default LV_USE_BUILTIN_STRING
 
 		config LV_USE_BUILTIN_STRING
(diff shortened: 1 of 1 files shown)


espzav:
@tore-espressif @kisvegabor PTAL

tore-espressif:
@espzav Looks good from ESP side :)

igrr:
@espzav could you please check other similar settings as well? For example, LV_USE_CLIB_SPRINTF. Basically we can default to POSIX or libc options, where available.

espzav:
> @espzav could you please check other similar settings as well? For example, LV_USE_CLIB_SPRINTF. Basically we can default to POSIX or libc options, where available.

I tried CONFIG_LV_USE_CLIB_MALLOC, CONFIG_LV_USE_CLIB_STRING and CONFIG_LV_USE_CLIB_SPRINTF and it seems, that it is working. 
As @kisvegabor said here https://github.com/lvgl/lvgl/pull/5617#issuecomment-1931918070, it seems, that we cannot use `default y if IDF_CMAKE` there can be added new bool like `LV_USE_CLIB` and we can select it in `esp_lvgl_port` component `Kconfig`.

PGNetHun:
Duplicate of https://github.com/lvgl/lvgl/pull/5617

espzav:
@kisvegabor I updated this PR, is it better solution for you? I removed duplicity with #5617

tore-espressif:
> LV_USE_CLIB and we can select it in esp_lvgl_port component Kconfig

This sounds reasonable to me. We add a new option that is hidden and not platform specific

Kconfig
```diff
@@ -69,9 +69,14 @@ menu "LVGL configuration"
...
 	endmenu
 
+	config LV_USE_CLIB
+		bool
+		default n
+
 	menu "Memory settings"
 		choice
 			prompt "Malloc functions source"
+			default LV_USE_CLIB_MALLOC if LV_USE_CLIB
```
PGNetHun, line 79: There will be 2 defaults if `LV_USE_CLIB` ?

igrr: We could make the 2nd one depend on !LV_USE_CLIB, but IIRC Kconfig picks the first default value which matches the conditions. So leaving the last one without any conditions should be fine.

PR #5639, opened 2024-02-12
Windows: Size window to have a client size of requested rectangle

danie1kr:

Creating a window in Windows only vaguely respects the `hor_res` and `ver_res` parameters.
With this proposal, the window decoration is better taken into account, creating a window that is more close to the requested size.

### Notes

### DIFF
--- a/src/drivers/windows/lv_windows_display.c
+++ b/src/drivers/windows/lv_windows_display.c
@@ -126,6 +126,10 @@ static unsigned int __stdcall lv_windows_display_thread_entrypoint(
         window_style &= ~(WS_SIZEBOX | WS_MAXIMIZEBOX | WS_THICKFRAME);
     }
 
+    // ask Windows how big a window needs to be for hor_res and ver_res
+    RECT rect = { 0, 0, data->hor_res, data->ver_res };
+    AdjustWindowRect(&rect, window_style, false);
+
     HWND window_handle = CreateWindowExW(
                              WS_EX_CLIENTEDGE,
                              L"LVGL.Window",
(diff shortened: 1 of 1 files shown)


kisvegabor:
cc @MouriNaruto

MouriNaruto:
I don't think you have read the implementation carefully. The current implementation will set the raw size from CreateWindowExW to LVGL display in the WM_CREATE message.

https://github.com/lvgl/lvgl/blob/e6dc2f05d239b914617f19e1db019ebb35d83449/src/drivers/windows/lv_windows_context.c#L488-L493

And finally do the size adjustment in the WM_SIZE message which is automatically called after the WM_CREATE message.

https://github.com/lvgl/lvgl/blob/e6dc2f05d239b914617f19e1db019ebb35d83449/src/drivers/windows/lv_windows_context.c#L522-L579

I don't use AdjustWindowRect when creating window because it's not reliable.

image
First, [AdjustWindowRect document](https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-adjustwindowrect)  said that to specify an extended window style, use the AdjustWindowRectEx function.

image
Second, [AdjustWindowRectEx document](https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-adjustwindowrectex) said that this API is not DPI aware, and should not be used if the calling thread is per-monitor DPI aware. For the DPI-aware version of this API, see AdjustWindowsRectExForDPI.

Also, AdjustWindowsRectExForDPI is not reliable for enabling the DWM composition which is forced beginning with Windows 8. Especially the Windows 10 frameless-style window which is the DWM hacking.

So, this is why I chose only to pass the raw window size before. Because we need to let WM_CREATE message know the raw value for postprocessing.

Kenji Mouri

danie1kr:
Thanks for the insights @MouriNaruto.
Then I will have to live with a smaller window. At least for me, the created window is smaller by a noticeable margin.
How can there not be a way in windows to easily calculate what we will get...

danie1kr:
Thanks for the insights @MouriNaruto.
Then I will have to live with a smaller window. At least for me, the created window is smaller by a noticeable margin.
How can there not be a way in windows to easily calculate what we will get...

PR #5662, opened 2024-02-16
fix(evdev): force processing of button events

### LINKED ISSUE #5211: Touch screen input events missed
### LVGL version

v8.3

### What happened?

Sometimes very short taps on the touch screen are not recognized by lvgl.
I'm using evdev as input device and I can see the events coming in when watching `evtest` output. However, the lv_obj does not receive events.
When the touch is longer (like if you would press a physical button), everything works as expected.
This happens only 50% of the time on very quick tapping.

Here's a video of three taps, left shows debugging output from my program in lvgl event callbacks, right side shows output of `evtest`. You can see that the first two taps are being recognized by lvgl but the last one isn't (no output on left side).

https://github.com/lvgl/lvgl/assets/1481786/da605b47-512d-43c3-8bf9-10c6f13279cd

### How to reproduce?

_No response_

stwirth:

Closes #5211.
When multiple button press/release events are queued up for processing, this change ensures that all of them are processed. Without this, a quick tap could be missed (see #5211).

I have tested a similar change on v8.3 but have not updated my system to v9 yet, so any help with testing this is appreciated :)

### DIFF
--- a/src/drivers/evdev/lv_evdev.c
+++ b/src/drivers/evdev/lv_evdev.c
@@ -114,7 +114,12 @@ static void _evdev_read(lv_indev_t * indev, lv_indev_data_t * data)
 
     /*Update dsc with buffered events*/
     struct input_event in = { 0 };
+    bool got_button_event = false;
     while(read(dsc->fd, &in, sizeof(in)) > 0) {
+        if(in.type == EV_SYN && got_button_event) {
+            /* force processing of button events to make sure we don't miss any, see #5211 */
+            break;
+        }
         if(in.type == EV_REL) {
             if(in.code == REL_X) dsc->root_x += in.value;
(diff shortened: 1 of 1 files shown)


kisvegabor:
Thank you! I've just tried it out with a touch pad and noticed an issue. When try to move the cursor with touch pad it is recognized as a press and it drags the containers. 

Have you tried it with a touch pad?

stwirth:
@kisvegabor no, I'm only working with a touch screen. I just changed the commit to use `if (in.code == BTN_TOUCH) got_button_event = true;`. Can you try again please?

kisvegabor:
Still the same. :slightly_frowning_face: 
However I found that my touchpad is handled as an "absolute" device. Where I touch the touch pas the cursor jumps there. 

Can it be an issue in the LVGL driver or in my Linux driver?

PR #5702, opened 2024-02-20
feat(draw): optimize helium asm

PeterBee97:
Optimize helium asm code
- Blend rgb565 without conversion: Blending function is now the same as `lv_color_16_16_mix`
- Remove unnecessary color reload: color fill should be faster for all format
Now 16bpp color fill is 1 instruction per loop (besides letp) as it should be, and 2 for opaque image BLIT-ting.

### Notes

### DIFF
--- a/src/draw/sw/blend/helium/lv_blend_helium.S
+++ b/src/draw/sw/blend/helium/lv_blend_helium.S
@@ -17,7 +17,6 @@ reciprocal:
 
 .text
 .syntax unified
-.altmacro
 .p2align 2
 
 TMP         .req r0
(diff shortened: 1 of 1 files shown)


XuNeo (dismissed):

GorgonMeducer:
This Helium assembly code use GAS macro which cannot be well supported in LLVM and AC6.
Would you please do the following changes to improve the portability?

- `.ifc` is needed to compare strings.
- Clang doesn't support `.func` and `.endfunc` the docs say that this is only relevant for stabs format debugging so these can probably be removed.
- Clang doesn't support the alternate `&` concatenation operator, it needs to be `\()` instead.

For now, the Helium assembly you added cannot be used with either LLVM or Arm Compiler 6, we would detect the compiler and disable the assembly accordingly.

Once you solved the compatibility issue, we can allow it work in LLVM and AC6.

Please take the following code as example:

```asm
.macro blender src_bpp, dst_bpp, mask, opa, mode
nop
.endm

.macro export name, src_bpp, dst_bpp, mask, opa, mode
.thumb_func
.global \name
\name\():
    blender     src_bpp, dst_bpp, mask, opa, mode
.endm

.macro export_set src, dst, src_bpp, dst_bpp, mode
.ifc \src,color
    export lv_\()src\()_blend_to_\()dst\()_helium, src_bpp, dst_bpp, 0, 0, mode
.else
nop
.endif
.endm

```

kisvegabor:
Thank you, I've tried out the changes. 

With RGB565 now it's faster, but has exactly the same speed as the SW renderer. 

In XRGB8888 format it's very slow, I saw about 5 FPS. :slightly_frowning_face:

PeterBee97:
Abandoned altmacro mode and now it can be compiled by clang. Also did some optimizations to 32bpp load & store. Would you look at this again? @kisvegabor @GorgonMeducer

kisvegabor:
@PeterBee97 With the latest code there is no compiler error, but the Clang hangs after compiling a lot of files. So I cannot verify the changes. :slightly_frowning_face: 

Once it's assembly, for performance measurement I can use any compiler, right?

kisvegabor:
After ten minutes it continued compiling and prints a loooooot of error like these:
image

Does it say anyhing to you?

PeterBee97:
The error is with the previous version, and with the latest change there should be no more "&" which could cause clang to output too many errors and hang. I have modified them according to GorgonMeducers's advice. Could you please pull the latest change again? @kisvegabor

kisvegabor:
Sorry for the late feedback.

The current version compiles well with LLVM. With RGB565 the performance is the same as with SW render, but with XRGB8888 Helium is ~50% slower. 

In your tests have you measured faster rendering?

PeterBee97:
> In your tests have you measured faster rendering?

Sorry I do not have suitable hardware at hand. I am considering the RT-Thread vision board which just came out featuring the same Renesas Cortex-M85 MCU. Maybe we can merge this PR and continue to optimize later?

FASTSHIFT (approved):

kisvegabor (approved):
I agree to merge it as it contains important updates to compile under LLVM.
Thank you for the fixes so far. Please let us know if you could test it on your end.

Issue #5391, opened 2024-01-19
The proposal of the minimal portable free-standing core implementation


MouriNaruto:
In the development of LVGL v9, there are lots of components added into LVGL. It's really good for lots of developers.

According the usage scenario (embedded, IoT and bare-metal) of LVGL, LVGL needs to keep the ability of running with no external dependencies. So, LVGL should not depend C standard library functions so much. And I know the efforts from LVGL, LVGL makes its standard library layer (for example, the “lv_strcmp” is a good example) to achieve that.

But I know it's not practical to make all components dependent on the LVGL standard library because not all components need to ask strictly. I think it's wise to let LVGL to have a minimal portable free-standing core will make embedded, IoT and bare-metal developers happy.

I think the the minimal portable free-standing core implementation should have these.

- only depends on LVGL standard library, for C standard header, should only depends on something without C standard function definitions but provides portability like “stdint.h” etc 
- exclude platform support because these cases (with limited C standard header support scenarios) you need to implement the platform support yourself in most cases,
- with demos and examples
- components which don't need to ask strictly have a separate source folder like the demos and examples do

Kenji Mouri

kisvegabor:
Hi,

I think it affects only the content of the [libs](https://github.com/lvgl/lvgl/tree/master/src/libs) folder. Other code (written by us) could use LVGL's stdio wrapper without any issues.

MouriNaruto:
> Hi,
> 
> I think it affects only the content of the [libs](https://github.com/lvgl/lvgl/tree/master/src/libs) folder. Other code (written by us) could use LVGL's stdio wrapper without any issues.

@kisvegabor I agree with that. Maybe LVGL source tree can have the folder name like third-party in the future. I think it's necessary to split the libs for written by LVGL and written by third-party.

Kenji Mouri

MouriNaruto:
> Actually the `libs` folder has this purpose. We have only non-LVGL written libs there.

@kisvegabor 

But actually, there are something like the "fsdrv" and "lv_bin_decoder" folders (which are LVGL written libs) in the "libs" folder.

I also think maybe LVGL can have a united “platform” or "hal" folder for "devices", "osal" and "stdlib" etc folders.

Kenji Mouri

kisvegabor:
> But actually, there are something like the "fsdrv" and "lv_bin_decoder" folders (which are LVGL written libs) in the "libs" folder.

[...]

PR #5720, opened 2024-02-22
fix(stdlib): fix issues when using LVGL TLSF memory pool to manage more than 4 GiB of memory

MouriNaruto:

People will know that I'm developing UEFI applications with LVGL in recent days after read https://github.com/lvgl/lvgl/issues/5391#issuecomment-1909771955.

Because UEFI memory pool will cause many memory fragments. So, my project uses LVGL builtin TLSF memory pool to manage the memory. And I found if I use LVGL TLSF memory pool to manage more than 4 GiB of memory, I cannot get the correct result via  lv_mem_monitor API.

I also found the original LVGL implementation have considered about the 64-bit environments and many lv_mem APIs use size_t to manage size. So, I think it should not break the ABI for 32-bit environments if we use size_t type instead of uint32_t type to fix the issue. (For 64-bit LVGL users needs to recompile. But I think most people don't use 64-bit LVGL instance in the current stage. It should be acceptable.)

### Notes

Kenji Mouri

### DIFF
--- a/src/stdlib/lv_mem.c
+++ b/src/stdlib/lv_mem.c
@@ -74,9 +74,9 @@ void * lv_malloc(size_t size)
 #if LV_LOG_LEVEL <= LV_LOG_LEVEL_INFO
         lv_mem_monitor_t mon;
         lv_mem_monitor(&mon);
-        LV_LOG_INFO("used: %6d (%3d %%), frag: %3d %%, biggest free: %6d",
-                    (int)(mon.total_size - mon.free_size), mon.used_pct, mon.frag_pct,
-                    (int)mon.free_biggest_size);
+        LV_LOG_INFO("used: %zu (%3d %%), frag: %3d %%, biggest free: %zu",
+                    mon.total_size - mon.free_size, mon.used_pct, mon.frag_pct,
+                    mon.free_biggest_size);
 #endif
         return NULL;
     }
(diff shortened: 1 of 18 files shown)


MouriNaruto:
I think there are some issues should be also fixed for LVGL. (Need some time to do.)

Kenji Mouri

MouriNaruto:
It looks like all CIs are passed. It should be continued.

Kenji Mouri

XuNeo (dismissed):

src/stdlib/lv_mem.c
```diff
@@ -74,9 +74,9 @@ void * lv_malloc(size_t size)
 #if LV_LOG_LEVEL <= LV_LOG_LEVEL_INFO
         lv_mem_monitor_t mon;
         lv_mem_monitor(&mon);
-        LV_LOG_INFO("used: %6d (%3d %%), frag: %3d %%, biggest free: %6d",
-                    (int)(mon.total_size - mon.free_size), mon.used_pct, mon.frag_pct,
-                    (int)mon.free_biggest_size);
+        LV_LOG_INFO("used: %zu (%3d %%), frag: %3d %%, biggest free: %zu",
+                    (size_t)(mon.total_size - mon.free_size), mon.used_pct, mon.frag_pct,
```
FASTSHIFT: remove size_t cast.

src/stdlib/lv_mem.c
```diff
@@ -103,9 +103,9 @@ void * lv_malloc_zeroed(size_t size)
...
```
FASTSHIFT: remove size_t cast.

MouriNaruto:
image
Here is the screenshot of just using LVGL TLSF memory pool to manage 6 GiB of memory in my test virtual machine with this fix.

Kenji Mouri

FASTSHIFT (changes requested):

src/stdlib/lv_mem.c
MouriNaruto: OK. I have modified it.

Kenji Mouri

src/stdlib/lv_mem.c
MouriNaruto: OK. I have modified it.

Kenji Mouri

FASTSHIFT (approved):

kisvegabor (approved):
Awesome, thank you!

PR #5725, opened 2024-02-23
fix(vg_lite): fix GPU draw unit fallback not finish

FASTSHIFT:

When using single-thread + GPU rendering, drawing operations that are not supported by the GPU need to fallback to software rendering.

At present, it is found that there are some problems in the draw fallback process, that is, the CPU does not wait for the GPU rendering to finish before rendering.
For example, the draw task queue currently has two tasks, the first task is dispatched to the GPU, and the second task is dispatched to the CPU.

The dispatch fallback process before modification:
dispatch_cb -> vg_lite_dispatch -> lv_draw_get_next_available_task -> vg_lite_draw -> vg_lite_flush ->
dispatch_cb -> sw_dispatch -> lv_draw_get_next_available_task -> sw draw

Modified process:
dispatch_cb -> vg_lite_dispatch -> lv_draw_get_next_available_task -> vg_lite_draw -> vg_lite_flush ->
dispatch_cb -> vg_lite_dispatch -> lv_draw_get_next_available_task -> task == NULL -> vg_lite_finish ->
dispatch_cb -> sw_dispatch -> lv_draw_get_next_available_task -> sw draw

I'm actually not sure whether this modification is correct, but after testing, the rendering behavior is as expected. If you have any questions, please correct me.

### Notes

### DIFF
--- a/src/draw/lv_draw.c
+++ b/src/draw/lv_draw.c
@@ -253,8 +253,19 @@ bool lv_draw_dispatch_layer(lv_display_t * disp, lv_layer_t * layer)
         lv_draw_unit_t * u = _draw_info.unit_head;
         while(u) {
             int32_t taken_cnt = u->dispatch_cb(u, layer);
+#if LV_USE_OS
             if(taken_cnt >= 0) render_running = true;
             u = u->next;
+#else
+            if(taken_cnt >= 0) {
+                render_running = true;
+                break;
+            }
+            else {
+                /*If the unit is not suitable for the layer, try the next unit*/
+                u = u->next;
+            }
+#endif
         }
     }
 


W-Mai (approved):

kisvegabor:
I think the issue is somewhere else. `lv_draw_get_next_available_task` checks if there is `LV_DRAW_TASK_STATE_QUEUED` tasks which is not overlapped by any other not ready task. 

From our previous discussions I remember that you mark a task as `READY` immediately as you take it. In the code of VGLite it looks good, but maybe there is something non obvious that I can't see.

If you mark a task READY too early in VGLite, `lv_draw_get_next_available_task` will think that an overlapping task is available and SW might start rendering it in parallel.

FASTSHIFT:
> I think the issue is somewhere else. `lv_draw_get_next_available_task` checks if there is `LV_DRAW_TASK_STATE_QUEUED` tasks which is not overlapped by any other not ready task.
> 
> From our previous discussions I remember that you mark a task as `READY` immediately as you take it. In the code of VGLite it looks good, but maybe there is something non obvious that I can't see.
> 
> If you mark a task READY too early in VGLite, `lv_draw_get_next_available_task` will think that an overlapping task is available and SW might start rendering it in parallel.

How to modify this code?
```c
static int32_t draw_dispatch(lv_draw_unit_t * draw_unit, lv_layer_t * layer)
{
    lv_draw_vg_lite_unit_t * u = (lv_draw_vg_lite_unit_t *)draw_unit;

    /* Return immediately if it's busy with draw task. */
    if(u->task_act) {
        return 0;
    }

    /* Try to get an ready to draw. */
    lv_draw_task_t * t = lv_draw_get_next_available_task(layer, NULL, VG_LITE_DRAW_UNIT_ID);

    /* Return 0 is no selection, some tasks can be supported by other units. */
    if(!t || t->preferred_draw_unit_id != VG_LITE_DRAW_UNIT_ID) {
        lv_vg_lite_finish(u);
        return -1;
    }

    void * buf = lv_draw_layer_alloc_buf(layer);
    if(!buf) {
        return -1;
    }

    /* Return if target buffer format is not supported. */
    if(!lv_vg_lite_is_dest_cf_supported(layer->draw_buf->header.cf)) {
        return -1;
    }

    t->state = LV_DRAW_TASK_STATE_IN_PROGRESS;
    u->base_unit.target_layer = layer;
    u->base_unit.clip_area = &t->clip_area;
    u->task_act = t;

    draw_execute(u);

    u->task_act->state = LV_DRAW_TASK_STATE_READY;
    u->task_act = NULL;

    /*The draw unit is free now. Request a new dispatching as it can get a new task*/
    lv_draw_dispatch_request();

    return 1;
}
```

FASTSHIFT:
Can I push `lv_draw_task_t * t` into the pending list and uniformly set the task state to `LV_DRAW_TASK_STATE_READY` in `lv_vg_lite_finish`?

kisvegabor:
Now `lv_vg_lite_finish` is called only from `lv_draw_vg_lite_layer`. From what other places do you plan to call `lv_vg_lite_finish`?

FASTSHIFT:
On draw dispatch, after the GPU gets the last task, wait for the GPU rendering to complete.
```c
static int32_t draw_dispatch(lv_draw_unit_t * draw_unit, lv_layer_t * layer)
{
    ...

    /* Try to get an ready to draw. */
    lv_draw_task_t * t = lv_draw_get_next_available_task(layer, NULL, VG_LITE_DRAW_UNIT_ID);

    /* Return 0 is no selection, some tasks can be supported by other units. */
    if(!t || t->preferred_draw_unit_id != VG_LITE_DRAW_UNIT_ID) {
        lv_vg_lite_finish(u);
        return -1;
    }
    ...
}
```

kisvegabor:
If you keep calling `draw_dispatch` for batch execution, you need to make each task READY to let `lv_draw_get_next_available_task` give a new task. It works in a single threaded environment, but I suggest something else:
- When you assign a task VGLite set its state to `LV_DRAW_TASK_STATE_IN_PROGRESS`. 
- In `lv_draw_get_next_available_task` if it it sees that a task is IN_PROGRESS but it's assigned to the current draw unit let's consider it as finished

For example let's say you pick the screen fill first, mark it as IN_PROGRESS. You immediately call `lv_draw_get_next_available_task` again. The next draw task is a label which depends on the screen, but as the screen will be drawn by same draw unit anyway, it can assume that the label is available. When everything is ready you mark all the tasks as READY. 

The only issue is that in case of multi core SW rendering all draw units has the same ID, so we need to use a pointer to the draw unit to check which draw unit is assigned to the task.

src/draw/sw/lv_draw_sw.c
```diff
@@ -313,6 +315,11 @@ static int32_t dispatch(lv_draw_unit_t * draw_unit, lv_layer_t * layer)
         return -1;
     }
 
+    extern int flush_cnt;
+    if(flush_cnt > 0) {
+        LV_LOG_WARN("!!!!!!!!!!!!!!!!!!!!!!!!!!!!! flush_cnt: %d", flush_cnt);
```
PGNetHun: Funny, and very important warning message!!!!!!!!!!!!!!!  :-)
I think we can remove this "debugging" code part, or if it has a meaning to devs, then simply correct the text: make it clear and easy to understand why it is logged.

src/draw/vg_lite/lv_draw_vg_lite.c
```diff
@@ -112,36 +117,49 @@ static void draw_execute(lv_draw_vg_lite_unit_t * u)
     vg_lite_identity(&u->global_matrix);
     vg_lite_translate(-layer->buf_area.x1, -layer->buf_area.y1, &u->global_matrix);
 
+    LV_LOG_USER("draw task: %d", t->type);
```
PGNetHun: Please remove this line, or change to TRACE, because it looks like a "debugging" / trace log event.

src/draw/vg_lite/lv_draw_vg_lite.c
```diff
@@ -112,36 +117,49 @@ static void draw_execute(lv_draw_vg_lite_unit_t * u)
...
+
+    bool has_flushed = true;
+
     switch(t->type) {
         case LV_DRAW_TASK_TYPE_LABEL:
             lv_draw_vg_lite_label(draw_unit, t->draw_dsc, &t->area);
```
PGNetHun: Every other task type calls `lv_vg_lite_flush(u)`, but label not.
Is it a special case? (I'm not familiar with VG Lite)

src/draw/vg_lite/lv_draw_vg_lite.c
```diff
@@ -112,36 +117,49 @@ static void draw_execute(lv_draw_vg_lite_unit_t * u)
...
         case LV_DRAW_TASK_TYPE_TRIANGLE:
             lv_draw_vg_lite_triangle(draw_unit, t->draw_dsc);
+            lv_vg_lite_flush(u);
             break;
         case LV_DRAW_TASK_TYPE_MASK_RECTANGLE:
             lv_draw_vg_lite_mask_rect(draw_unit, t->draw_dsc, &t->area);
+            has_flushed = false;
             break;
 #if LV_USE_VECTOR_GRAPHIC
         case LV_DRAW_TASK_TYPE_VECTOR:
```
PGNetHun: This vector task also does not need flush?

src/draw/vg_lite/lv_draw_vg_lite.c
```diff
@@ -152,7 +170,7 @@ static void draw_execute(lv_draw_vg_lite_unit_t * u)
             break;
     }
 
-    lv_vg_lite_flush(u);
+    return has_flushed;
```
PGNetHun: What about inverting the logic?
```
bool flush = false;
switch (...) {
   case ....:
       lv_draw_vg_lite_.....();
       flush = true;
   case ....:
   .....
}

if ( flush ) lv_vg_lite_flush(u);
return flush;
```

Changing a boolean value needs less CPU instructions than calling same function in many places (and passing same parameter every time).
(it's just a micro-optimization, but if above suggested logic is correct for every switch-case, then some bytes can be spared in firmware)

src/draw/vg_lite/lv_draw_vg_lite.c
```diff
@@ -204,7 +226,7 @@ static int32_t draw_evaluate(lv_draw_unit_t * draw_unit, lv_draw_task_t * task)
     LV_UNUSED(draw_unit);
 
     switch(task->type) {
-        case LV_DRAW_TASK_TYPE_LABEL:
+        // case LV_DRAW_TASK_TYPE_LABEL:
```
PGNetHun: Accidentally left here commented (because of debugging), or it is really not needed? 
In latter case you could remove it, git history will have it anyway.

src/draw/vg_lite/lv_draw_vg_lite_label.c
```diff
@@ -223,7 +223,7 @@ static void draw_letter_bitmap(lv_draw_vg_lite_unit_t * u, const lv_draw_glyph_d
      * You need to wait for the GPU to finish using the buffer before releasing it.
      * Later, use the font cache for management to improve efficiency.
      */
-    lv_vg_lite_finish(u);
+    // lv_vg_lite_finish(u);
```
PGNetHun: Accidentally left here commented (because of debugging?), or it is really not needed?
In latter case you could remove it, git history will have it anyway.

src/draw/vg_lite/lv_vg_lite_utils.c
```diff
@@ -328,7 +328,7 @@ bool lv_vg_lite_is_src_cf_supported(lv_color_format_t cf)
 {
     switch(cf) {
         case LV_COLOR_FORMAT_A4:
-        case LV_COLOR_FORMAT_A8:
+        // case LV_COLOR_FORMAT_A8:
```
PGNetHun: Accidentally left here commented (because of debugging?), or it is really not needed?
In latter case you could remove it, git history will have it anyway.

src/draw/vg_lite/lv_vg_lite_utils.c
```diff
@@ -1028,13 +1028,18 @@ void lv_vg_lite_disable_scissor(void)
     LV_VG_LITE_CHECK_ERROR(vg_lite_disable_scissor());
 }
 
+int flush_cnt;
```
PGNetHun: I assume this is a debugging variable "left-over".  Or maybe not, and used somewhere else.
If not needed, can be removed.
If needed, then please move to top of file, and change to unsigned int (if it really can't be negative and there overflow will not cause problem).

src/draw/vg_lite/lv_vg_lite_utils.c
```diff
@@ -1028,13 +1028,18 @@ void lv_vg_lite_disable_scissor(void)
...
+
 void lv_vg_lite_flush(struct _lv_draw_vg_lite_unit_t * u)
 {
     LV_ASSERT_NULL(u);
     LV_PROFILER_BEGIN;
 
     u->flush_count++;
 
+    LV_LOG_USER("++ flush_count = %d", u->flush_count);
```
PGNetHun: I assume this is also a debugging left-over.
If needed, then please change to TRACE and simple text:
```suggestion
    LV_LOG_TRACE("flush_count = %d", u->flush_count);
```

src/draw/vg_lite/lv_vg_lite_utils.c
```diff
@@ -1063,12 +1068,17 @@ void lv_vg_lite_finish(struct _lv_draw_vg_lite_unit_t * u)
 
     LV_VG_LITE_CHECK_ERROR(vg_lite_finish());
 
+    LV_LOG_USER("/// flush_count: %d", u->flush_count);
```
PGNetHun: I guess this line is not needed, only a debugging left-over.
Or if still needed, then correct text (remove `///` )

PGNetHun (changes requested):

src/draw/sw/lv_draw_sw.c
```diff
@@ -289,6 +289,8 @@ static int32_t evaluate(lv_draw_unit_t * draw_unit, lv_draw_task_t * task)
     return 0;
 }
 
+extern int flush_cnt;
```
PGNetHun: I think this should go to top of file, so we can see immediately what variables are used.

PGNetHun (changes requested):

src/draw/sw/lv_draw_sw.c
FASTSHIFT: Ah, this is a test code, used to check whether the GPU drawing is completed before software rendering. I found that there are still problems, so I will convert it to a draft~

kisvegabor:
@FASTSHIFT What's the satus of this PR?

FASTSHIFT:
> @FASTSHIFT What's the satus of this PR?

I tried to add the state `LV_DRAW_TASK_STATE_RENDER_IN_PROGRESS`, but I found that state management is very complicated (and there is still an unfinished fallback), so I temporarily stopped the investigation and used the simple solution of this PR first...

If you have time, can you help implement your `LV_DRAW_TASK_STATE_QUEUED` solution?

kisvegabor:
> If you have time, can you help implement your LV_DRAW_TASK_STATE_IN_PROGRESS solution?

Yes, of course. As it can be really complicated I suggest having a call to discuss it live.

PR #5734, opened 2024-02-23
fix(draw_vector): check the point lenght of path in lv_vector_path_get_bounding

duliang2008:
check the path and area is not NULL.
check the len of points in path is not zero.

add param checks to avoid buff-overflow.

### DIFF
--- a/src/draw/lv_draw_vector.c
+++ b/src/draw/lv_draw_vector.c
@@ -325,7 +325,15 @@ void lv_vector_path_close(lv_vector_path_t * path)
 
 void lv_vector_path_get_bounding(const lv_vector_path_t * path, lv_area_t * area)
 {
+    if(path == NULL || area == NULL) {
+        return;
+    }
+
     uint32_t len = lv_array_size(&path->points);
+    if(len == 0) {
+        return;
+    }
+
     lv_fpoint_t * p = lv_array_front(&path->points);
     float x1 = p[0].x;
     float x2 = p[0].x;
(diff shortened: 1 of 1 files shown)


src/draw/lv_draw_vector.c
```diff
@@ -325,7 +325,15 @@ void lv_vector_path_close(lv_vector_path_t * path)
 
 void lv_vector_path_get_bounding(const lv_vector_path_t * path, lv_area_t * area)
 {
+    if(path == NULL || area == NULL) {
```
XuNeo: Should we assert here or leave the error to caller?

FASTSHIFT: LV_ASSERT_NULL(path);
LV_ASSERT_NULL(area);

src/draw/lv_draw_vector.c
```diff
@@ -325,7 +325,15 @@ void lv_vector_path_close(lv_vector_path_t * path)
...
+        return;
+    }
+
     uint32_t len = lv_array_size(&path->points);
+    if(len == 0) {
+        return;
```
FASTSHIFT, line 334: lv_memzero(area, sizeof(lv_area_t));
return;

FASTSHIFT (changes requested):

kisvegabor (dismissed):

FASTSHIFT (approved):

XuNeo (approved):

Issue #5735, opened 2024-02-23
Meson failing compile step when building projects using v9.0.0


KaruroChori:
### LVGL version

v9.0.0

### What happened?

Meson will fail to compile projects integrating this library at version v9.0.0.
The older 8.x branch works fine.
A significant number of types cannot be found and will lead to compile errors.

KaruroChori:
I was able to make it work with a Frankenstein-ed solution, using a modified version of the older makefile

```
# Option to define LV_LVGL_H_INCLUDE_SIMPLE, default: ON
option(LV_LVGL_H_INCLUDE_SIMPLE
       "Use #include \"lvgl.h\" instead of #include \"../../lvgl.h\"" ON)

# Option to define LV_CONF_INCLUDE_SIMPLE, default: ON
option(LV_CONF_INCLUDE_SIMPLE
       "Simple include of \"lv_conf.h\" and \"lv_drv_conf.h\"" ON)

# Option to set LV_CONF_PATH, if set parent path LV_CONF_DIR is added to
# includes
option(LV_CONF_PATH "Path defined for lv_conf.h")
get_filename_component(LV_CONF_DIR ${LV_CONF_PATH} DIRECTORY)

# Option to build shared libraries (as opposed to static), default: OFF
option(BUILD_SHARED_LIBS "Build shared libraries" OFF)

file(GLOB_RECURSE SOURCES ${LVGL_ROOT_DIR}/src/*.c ${LVGL_ROOT_DIR}/src/*.S)
file(GLOB_RECURSE EXAMPLE_SOURCES ${LVGL_ROOT_DIR}/examples/*.c)
file(GLOB_RECURSE DEMO_SOURCES ${LVGL_ROOT_DIR}/demos/*.c)
file(GLOB_RECURSE THORVG_SOURCES ${LVGL_ROOT_DIR}/src/libs/thorvg/*.cpp ${LVGL_ROOT_DIR}/src/others/vg_lite_tvg/*.cpp)

add_library(lvgl ${SOURCES})
add_library(lvgl::lvgl ALIAS lvgl)


target_compile_definitions(
  lvgl PUBLIC $<$<BOOL:${LV_LVGL_H_INCLUDE_SIMPLE}>:LV_LVGL_H_INCLUDE_SIMPLE>
              $<$<BOOL:${LV_CONF_INCLUDE_SIMPLE}>:LV_CONF_INCLUDE_SIMPLE>)

[...]

PR #5745, opened 2024-02-25
fix(env): fix meson build break

KaruroChori:
Linked to #5735.
I am not suggesting these as the actual changes to patch the current makefile in 9.0.

This is just me forcing my way to make it work.
I basically started from the 8.3 makefile which was fine in meson, and I added back most of the new features from the 9.0 version until it broke again.
Tested in meson 1.3.x.

### DIFF
--- a/env_support/cmake/custom.cmake
+++ b/env_support/cmake/custom.cmake
@@ -32,16 +32,9 @@ file(GLOB_RECURSE THORVG_SOURCES ${LVGL_ROOT_DIR}/src/libs/thorvg/*.cpp ${LVGL_R
 add_library(lvgl ${SOURCES})
 add_library(lvgl::lvgl ALIAS lvgl)
 
-if(NOT (CMAKE_C_COMPILER_ID STREQUAL "MSVC"))
-  target_compile_definitions(
-    lvgl PUBLIC $<$<BOOL:${LV_LVGL_H_INCLUDE_SIMPLE}>:LV_LVGL_H_INCLUDE_SIMPLE>
-                $<$<BOOL:${LV_CONF_INCLUDE_SIMPLE}>:LV_CONF_INCLUDE_SIMPLE>
-                $<$<COMPILE_LANGUAGE:ASM>:__ASSEMBLY__>)
-else()
-  target_compile_definitions(
-    lvgl PUBLIC $<$<BOOL:${LV_LVGL_H_INCLUDE_SIMPLE}>:LV_LVGL_H_INCLUDE_SIMPLE>
-                $<$<BOOL:${LV_CONF_INCLUDE_SIMPLE}>:LV_CONF_INCLUDE_SIMPLE>)
-endif()
+target_compile_definitions(
+  lvgl PUBLIC $<$<BOOL:${LV_LVGL_H_INCLUDE_SIMPLE}>:LV_LVGL_H_INCLUDE_SIMPLE>
+              $<$<BOOL:${LV_CONF_INCLUDE_SIMPLE}>:LV_CONF_INCLUDE_SIMPLE>)
 
 # Add definition of LV_CONF_PATH only if needed
 if(LV_CONF_PATH)


kisvegabor:
Looks good to me! Especially because the build tests have been passed. :slightly_smiling_face: 

The failing CI is unrelated. Please make the PR ready (not draft), so that we can approve it before merging.

kisvegabor (dismissed):

KaruroChori:
I marked it as a normal pull request.
However, I would like to stress it was not meant as such.
It is possible for it to introduce regressions (nothing major) which are not captured by automatic tests, and they might only manifest in others building workflow (as it did for me).

XuNeo:
Hi, Could you rebase to latest master to run CI again?

kisvegabor:
If it breaks something for someone, we will fix it and also add a test for that case. :slightly_smiling_face:

kisvegabor:
I've rebased it on `master`. There were some conflicts that I have resolved, so please test it again with Meson.

KaruroChori:
Great, I am currently abroad without my main workstation.
I will test it as soon as possible and update this PR.

IAMMX:
Screenshot_20240310_233638_com huawei browser
Why not fix the mainline version instead of the branch?

PGNetHun:
There are 624 files changed in this PR, which is strange.
Why not pointing PR to `master` branch, and then rebase to `master` ?
A lot of changes would disappear.
At the moment merging this PR to `release/v9.0` is risky, I see many changes here already "fixed" in `master` branch.

PGNetHun (commented):
There are 624 files changed in this PR, which is strange.
Why not pointing PR to `master` branch, and then rebase to `master` ?
A lot of changes would disappear.
At the moment merging this PR to `release/v9.0` is **risky**, I see many changes here already "fixed" in `master` branch.

KaruroChori:
I have no idea what happened and I am not really familiar with rebase-based workflows to be honest.
This PR was open as a draft just to show a possible fix for the build regression, and it consisted of one or two commits to change the makefile. It was opened against the release branch, because that is the one my project is using. 

I was asked to move it from draft to a normal PR.
After that, @kisvegabor pushed changes to rebase it against master. And this is where most of those 219 commits are coming from I think.

KaruroChori:
I changed target. Next week I will be able to test it and verify some changes later added by @kisvegabor against my original ones. If that works, I think this patch should be back-ported to the stable release branch, unless you have release guidelines against this.

PGNetHun:
Great, thanks to changing base to `master` branch, now only 1 file is changed, that's better to review.
I'm not familiar with cmake, sorry, I will skip reviewing this change.

env_support/cmake/custom.cmake
```diff
@@ -4,36 +4,28 @@ option(LV_LVGL_H_INCLUDE_SIMPLE
...
 target_compile_definitions(
   lvgl PUBLIC $<$<BOOL:${LV_LVGL_H_INCLUDE_SIMPLE}>:LV_LVGL_H_INCLUDE_SIMPLE>
-              $<$<BOOL:${LV_CONF_INCLUDE_SIMPLE}>:LV_CONF_INCLUDE_SIMPLE>
-              $<$<COMPILE_LANGUAGE:ASM>:__ASSEMBLY__>)
-
-# Add definition of LV_CONF_PATH only if needed
-if(LV_CONF_PATH)
-  target_compile_definitions(lvgl PUBLIC LV_CONF_PATH=${LV_CONF_PATH})
-endif()
+              $<$<BOOL:${LV_CONF_INCLUDE_SIMPLE}>:LV_CONF_INCLUDE_SIMPLE>)
```
XuNeo, line 37: I'm not sure of what's the original code does. Is it sure to remove them?
`              $<$<COMPILE_LANGUAGE:ASM>:__ASSEMBLY__>)`

Does the `LV_CONF_PATH` removed intentionally?

```
# Add definition of LV_CONF_PATH only if needed
if(LV_CONF_PATH)
  target_compile_definitions(lvgl PUBLIC LV_CONF_PATH=${LV_CONF_PATH})
endif()
```

env_support/cmake/custom.cmake
```diff
@@ -43,14 +35,21 @@ endif()
 # Include root and optional parent path of LV_CONF_PATH
 target_include_directories(lvgl SYSTEM PUBLIC ${LVGL_ROOT_DIR} ${LV_CONF_DIR})
 
+# Include /examples folder
```
XuNeo: Should remove this comment since below one is for `thorvg`.

env_support/cmake/custom.cmake
```diff
@@ -43,14 +35,21 @@ endif()
...
 
 if(NOT LV_CONF_BUILD_DISABLE_THORVG_INTERNAL)
     add_library(lvgl_thorvg ${THORVG_SOURCES})
     add_library(lvgl::thorvg ALIAS lvgl_thorvg)
     target_include_directories(lvgl_thorvg SYSTEM PUBLIC ${LVGL_ROOT_DIR}/src/libs/thorvg)
-    target_link_libraries(lvgl_thorvg PUBLIC lvgl)
+    if(LV_CONF_PATH)
```
XuNeo: Please add back the original `target_link_libraries(lvgl_thorvg PUBLIC lvgl)`. The compile definitions added to lvgl will also take effect on lvgl_thorvg.

kisvegabor:
Not stale

kisvegabor:
@KaruroChori How can we proceed with this update?

KaruroChori:
The combined changes done in the branch are still working with meson. I am now checking the changes in the revision.

env_support/cmake/custom.cmake
KaruroChori: Added back
```
# Add definition of LV_CONF_PATH only if needed
if(LV_CONF_PATH)
  target_compile_definitions(lvgl PUBLIC LV_CONF_PATH=${LV_CONF_PATH})
endif()
```
and it works. That COMPILE_LANGUAGE etc however is the culprit preventing the cmake file from working.

env_support/cmake/custom.cmake
KaruroChori: Added back in test and it works

env_support/cmake/custom.cmake
KaruroChori: Added back and tested to work

KaruroChori:
Updated the cmake on my branch. 
The only issue left is now the removed `  $<$<COMPILE_LANGUAGE:ASM>:__ASSEMBLY__>` which will fail to work on meson if placed back.

kisvegabor:
Thank you for looking into it again.

> The only issue left is now the removed   $<$<COMPILE_LANGUAGE:ASM>:__ASSEMBLY__> which will fail to work on meson if placed back.

Please post the error message. Hopefully we can find a solution.

KaruroChori:
[error.log](https://github.com/lvgl/lvgl/files/15049404/error.log)

kisvegabor:
@XuNeo @liamHowatt Does it say something to you?

liamHowatt:
@KaruroChori could you please share your "meson.build" file?

rtiemann:
Hi,

I've just started to play around with LVGL and ran into this issue when trying to integrate LVGL into a project built with Meson. Here is a minimal example for reproducing the issue:

    $ mkdir foo && cd foo
    $ git clone -b release/v9.1 https://github.com/lvgl/lvgl.git subprojects/lvgl
    $ cat <<EOF >meson.build
    project('testing', 'c')
    cmake = import('cmake')
    lvgl_cmake_opts = cmake.subproject_options()
    lvgl_cmake_opts.add_cmake_defines({'LV_CONF_PATH': meson.project_source_root() / 'lv_conf.h'})
    lvgl_proj = cmake.subproject('lvgl', options: lvgl_cmake_opts)
    EOF
    $ cat <<EOF >lv_conf.h
    #ifndef LV_CONF_H
    #define LV_CONF_H
    #endif
    EOF
    $ meson setup . build
    $ ninja -C build

The build fails because `__ASSEMBLY__` gets defined for C files, which causes the typedefs in `src/misc/lv_types.h` to be invisible to the C compiler. The CMake build correctly adds `-D__ASSEMBLY__` only to assembly files whereas Meson adds `-D__ASSEMBLY__` to all files in `ninja.build`. Looks like a bug in Meson's CMake module to me. I've tested with Meson 1.4.0 and latest master, and both fail.

Anyway, removing the  `$<$<COMPILE_LANGUAGE:ASM>:__ASSEMBLY__>` part from `env_support/cmake/custom.cmake` makes it work.

The only two assembly files I could find in the LVGL project (`src/draw/sw/blend/neon/lv_blend_neon.S` and `src/draw/sw/blend/helium/lv_blend_helium.S`) define `__ASSEMBLY__` themselves, so as far as I can tell, it should be safe to simply remove the corresponding part from `env_support/cmake/custom.cmake`.

kisvegabor:
> The only two assembly files I could find in the LVGL project (src/draw/sw/blend/neon/lv_blend_neon.S and src/draw/sw/blend/helium/lv_blend_helium.S) define __ASSEMBLY__ themselves, so as far as I can tell, it should be safe to simply remove the corresponding part from env_support/cmake/custom.cmake.

Thank you for investigating it! Removing `__ASSEMBLY__` seems safe because it's added in both `.S` files too. See [here](https://github.com/lvgl/lvgl/blob/master/src/draw/sw/blend/neon/lv_blend_neon.S#L6-L8).

So it seems [this](https://github.com/lvgl/lvgl/blob/master/env_support/cmake/custom.cmake#L35-L44) in the current Cmake file in master can be simplified like:
```cmake
target_compile_definitions(lvgl PUBLIC 
    $<$<BOOL:${LV_LVGL_H_INCLUDE_SIMPLE}>:LV_LVGL_H_INCLUDE_SIMPLE>
    $<$<BOOL:${LV_CONF_INCLUDE_SIMPLE}>:LV_CONF_INCLUDE_SIMPLE>)
```

What do you thnik?

rtiemann:
> So it seems [this](https://github.com/lvgl/lvgl/blob/master/env_support/cmake/custom.cmake#L35-L44) in the current Cmake file in master can be simplified like:
> What do you thnik?

Yes, I think this simplification should not break anything given how the project is structured. I've tried it on master a few minutes ago, and it Works On My Machine (TM). So if you could change this, it would be a very helpful change for Meson projects as long as Meson has problems with this construct. 🙂

The `__ASSEMBLY__` magic was added in b42cb4307458e2a5d0248756b716d967b2d8066c. Maybe the author can explain why this change was made in the first place (probably simply to avoid having to define that symbol in every newly added assembly file).

kisvegabor:
I've added my version of this PR. As it affects how the `.S` files are compiled, @PeterBee97, could you take look at it?

KaruroChori:
> @KaruroChori could you please share your "meson.build" file?

Sorry, it looks like github decided not to show me notifications for activities related to this PR, so I just assumed it went stale :/.
Happy to see the root cause was isolated.

kisvegabor:
@KaruroChori maybe you have unsubscribed my accident: 
image

kisvegabor (approved):

liamHowatt (approved):

Issue #5680, opened 2024-02-19
Usage of stdint, stddef, ...


MiSimon:
### Introduce the problem

In some header and source files there are includes for stdint, stddef, ... . In some environments these headers do not exist. 

It is possible to produce simple versions of these headers that include everything needed for lvgl. But since the includes happen in the headers and not only inside source files, these headers have to be provided with lvgl.h and the other headers.

### Proposal

I suggest to create lv_xxx headers for them (like lv_string.h) and use these in the header and source files. Inside this headers a define like LV_USE_OS can be used to either include the existing headers or to include specific headers.

An example for lv_stdint.h:

```
#if LV_USE_OS == LV_OS_CUSTOM
  #include LV_OS_CUSTOM_STDINT_INCLUDE
#else
  #include <stdint.h>
#endif
```

With defines in lv_conf.h:

```
#if LV_USE_OS == LV_OS_CUSTOM
    #define LV_OS_CUSTOM_STDINT_INCLUDE <my_stdint.h>
    #define LV_OS_CUSTOM_STDDEF_INCLUDE <my_stddef.h>
    #define LV_OS_CUSTOM_STDBOOL_INCLUDE <my_stdbool.h>
#endif
```

kisvegabor:
I see the problem. Just out of curiosity, do you develop an UEFI bios?

MiSimon:
Yes indeed, I use LVGL with UEFI for a couple of years now. 
Till now i used it in a company project with other ported libraries, to make porting easier there is a UEFI version of the standard library. 
Now i want to use it in a private project where i do not need/want an UEFI version of the standard library to keep it simple. The goal would be a driver like the windows/x11/... drivers integrated in LVGL.

MouriNaruto:
I also hope to have that feature.

In the current stage, I have made a minimal C runtime (My public but old implementation at https://github.com/ProjectMile/Mile.Uefi/tree/main/Mile.Uefi.RuntimeLibrary/Include) for LVGL to workaround that issue, lol.

Kenji Mouri

kisvegabor:
@MouriNaruto 
Would this on its own solve your issue too?
```
#if LV_USE_OS == LV_OS_CUSTOM
    #define LV_OS_CUSTOM_STDINT_INCLUDE <my_stdint.h>
    #define LV_OS_CUSTOM_STDDEF_INCLUDE <my_stddef.h>
[...]

PR #5767, opened 2024-02-28
refacter(conf): use defines for standard includes

MiSimon:

As discussed in #5680 the core files should be platform independent and not rely on standard headers and/or standard methods.
Most of the standard methods (strlen, memcmp, ...) have already been replaced with internal implementations, only strcat and memcmp were missing.

This pull requests:
- adds defines for all external header files that can be changed to fit the platform requirement
- adds lv_strcat and lv_memcmp
- replaces hard coded external header names with the new defines
- replaces calls to standard functions with calls to internal implementations


kisvegabor:
Hi,

Thank you! It looks good to me. I've just a fixed a few minor things.

src/core/lv_group.c
```diff
@@ -6,12 +6,11 @@
 /*********************
  *      INCLUDES
  *********************/
-#include <stddef.h>
-
 #include "lv_group.h"
 #include "../core/lv_obj.h"
 #include "../core/lv_global.h"
 #include "../indev/lv_indev.h"
+#include LV_STDDEF_INCLUDE
```
XuNeo: How about adding lv_stddef.h similar to lv_string.h etc.

src/stdlib/lv_string.h
```diff
@@ -114,6 +123,15 @@ char * lv_strdup(const char * src);
...
 char * lv_strdup(const char * src);
 
+/**
+ * @brief Copies the string pointed to by src, including the terminating null character,
+ *        to the end of the string pointed to by dst.
+ * @param dst Pointer to the destination string where the content is to be appended.
+ * @param src Pointer to the source of data to be copied.
+ * @return A pointer to the destination string, which is dst.
+ */
+char * lv_strcat(char * dst, const char * src);
```
XuNeo, line 125: Should we also add version of lv_strncat?

FASTSHIFT (dismissed):

src/stdlib/lv_string.h
MiSimon: It would make sense to replace all possible calls to lv_strcat with lv_strncat. IMO also all calls to lv_strcpy should be changed to lv_strncpy. 
But i think that should be done in a new PR.

src/core/lv_group.c
kisvegabor: What do we need from `stddef`? I think we use only `NULL` from there.

src/stdlib/lv_string.h
kisvegabor: Having only `lv_strncat` makes sense. 

I agree to to spare `lv_strncpy` for an other PR.

XuNeo (dismissed):

XuNeo:
While resolving the conflicts, it comes to my mind that we can include `stdbool`, `stdint` etc. from `lv_types.h` and use `lv_types.h` it where needed?

MiSimon:
I moved the includes to lv_types.h. My only concern is that now the way of including files is hidden for most contributors and many will continue to include them with the standard names.

XuNeo:
> many will continue to include them with the standard names.

On which platform the standard includes are not exist? Is it possible to include them to CI?

src/draw/vg_lite/lv_vg_lite_decoder.c
```diff
@@ -12,6 +12,8 @@
 #if LV_USE_DRAW_VG_LITE
 
 #include "lv_vg_lite_utils.h"
+#include <stdlib.h>
+#include <string.h>
```
XuNeo, line 16:
```suggestion
#include "../../misc/lv_types.h"
```

src/misc/lv_rb.h
```diff
@@ -14,8 +14,8 @@ extern "C" {
  *      INCLUDES
  *********************/
 #include "lv_types.h"
-#include "stdbool.h"
 #include "lv_assert.h"
+#include LV_STDBOOL_INCLUDE
```
XuNeo:
```suggestion
#include "lv_types.h"
```

MiSimon:
> On which platform the standard includes are not exist? Is it possible to include them to CI?

I use LVGL in UEFI and there are no standard libraries by default.

src/draw/vg_lite/lv_vg_lite_decoder.c
MiSimon: I left the standard headers in the 'not core' .c files on purpose. I didn't know which of this files (drawing and external libraries) require the standard includes through their dependencies.

XuNeo: I understand.
For `vg_lite`, it's normally the case we can use standard includes, so it's fine to leave it here.

XuNeo (dismissed):

src/misc/lv_types.h
```diff
@@ -15,8 +15,14 @@ extern "C" {
...
 
+#include LV_STDDEF_INCLUDE
+#include LV_STDBOOL_INCLUDE
+#include LV_INTTYPES_INCLUDE
+#include LV_LIMITS_INCLUDE
+#include LV_STDARG_INCLUDE
+
 #ifndef __ASSEMBLY__
-#include <stdint.h>
+#include LV_STDINT_INCLUDE
```
kisvegabor: Please add all includes in the `__ASSEMBLY__` guard.

kisvegabor:
> I use LVGL in UEFI and there are no standard libraries by default.

Having a CI as @XuNeo suggested would be great indeed. This way we could be sure that LVGL is always working for UEFI BIOS development. Is an UEFI BIOS development environment something that we can easily set up on a Linux based CI machine?

MiSimon:
> Having a CI as @XuNeo suggested would be great indeed. This way we could be sure that LVGL is always working for UEFI BIOS development. Is an UEFI BIOS development environment something that we can easily set up on a Linux based CI machine?

I have no experience with the github CI system, but if we can provide a small set of UEFI headers and a clang + llvm build environment it should be possible to run a simple build.

XuNeo (approved):

FASTSHIFT (approved):

kisvegabor:
> I have no experience with the github CI system, but if we can provide a small set of UEFI headers and a clang + llvm build environment it should be possible to run a simple build.

That would be amazing! Are you interested in contributing with such a thing?

src/stdlib/lv_string.h
liamHowatt: `lv_strncat` PR: #5927

MiSimon:
> That would be amazing! Are you interested in contributing with such a thing?

If i can take _tests/makefile_ as starting point i can create a version of it that contains the headers and compiles a simple application for UEFI (X64). As i said i have no experience with the github CI system, somebody would have to add the build to the CI system.

kisvegabor:
If you can just open a new issues and describe the required steps to set up the environment we can convert it to a CI.

MiSimon:
I have created a new PR #5964, the PR contains a simple Makefile and a simple header.

PR #5787, opened 2024-03-04
ci(micropython): catch event failure in the test

kdschlosser:
Adds to the MicroPython test to push an event to see if the callback gets called properly.

### DIFF
--- a/.github/workflows/build_micropython.yml
+++ b/.github/workflows/build_micropython.yml
@@ -75,7 +75,7 @@ jobs:
       if: success() && matrix.port == 'unix'
       run: |
         export XDG_RUNTIME_DIR=/tmp
-        python3 lib/lv_bindings/lvgl/tests/micropy_test/__init__.py --artifact-path=lib/lv_bindings/lvgl/tests/micropy_test/artifacts --mpy-path=ports/unix/build-standard/micropython
+        python3 lib/lv_bindings/lvgl/tests/micropy_test/__init__.py --artifact-path=lib/lv_bindings/lvgl/tests/micropy_test/artifacts --mpy-path=ports/unix/build-standard/micropython --debug
 
     - name: Upload Artifacts
       uses: actions/upload-artifact@v4
--- a/.github/workflows/build_micropython.yml
+++ b/.github/workflows/build_micropython.yml
@@ -75,7 +75,7 @@ jobs:
       if: success() && matrix.port == 'unix'
       run: |
         export XDG_RUNTIME_DIR=/tmp
-        python3 lib/lv_bindings/lvgl/tests/micropy_test/__init__.py --artifact-path=lib/lv_bindings/lvgl/tests/micropy_test/artifacts --mpy-path=ports/unix/build-standard/micropython --debug
+        python3 lib/lv_bindings/lvgl/tests/micropy_test/__init__.py --artifact-path=lib/lv_bindings/lvgl/tests/micropy_test/artifacts --mpy-path=ports/unix/build-standard/micropython
 
     - name: Upload Artifacts
       uses: actions/upload-artifact@v4
--- a/tests/micropy_test/micropy.py
+++ b/tests/micropy_test/micropy.py
@@ -110,6 +110,19 @@ def event_callback(e):
 
 
 @test_func_wrapper
+def test_event():
+    scr = lv.screen_active()
+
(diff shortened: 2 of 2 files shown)


kisvegabor:
Still failing :slightly_frowning_face:

kdschlosser:
OK so the question is what is the CI using for the binding? is it using the master branch code or it is using some other branch? where is the micropython code coming from that the CI is using. and what branch of lv_micropython was the update added to?

kdschlosser:
I might need to tweak the test. the failure that it is spitting out is not very helpful so I am going to turn on the debugging mode in the test and commit that change to this PR so I can see what is going on. then I will turn off the debugging and do another commit to this PR and make any changes that need to be made.

kisvegabor:
It uses lv_micropython master, and the lv_binding_micropython version as it's it set in lv_micropython. That's why we need to update lv_micropython when lv_binding_micropython changes.

Lvgl is used from the PR.

kdschlosser:
I was using the wrong function to send the event. I should have been using `lv_obj_send_event` and not `lv_event_send`

kdschlosser:
OK just have to tun off the debugging output and this should be done.

kdschlosser:
ok should be good to go.

kisvegabor (approved):

XuNeo (approved):
Hi,
I'm curious about how the `user_data` in `add_event_cb` can be used. It seems all examples pass None directly.

I'm wondering could we optimize the `global event_flag` with a local python variable etc.

```python
@test_func_wrapper
def test_event():
    scr = lv.screen_active()
    
    flag = False
    scr.add_event_cb(event_callback, lv.EVENT.CLICKED, flag)

    scr.send_event(lv.EVENT.CLICKED, None)

    if not flag:
        raise RuntimeError('Event failure')
```

PGNetHun:
> Hi, I'm curious about how the `user_data` in `add_event_cb` can be used. It seems all examples pass None directly.
> 
> I'm wondering could we optimize the `global event_flag` with a local python variable etc.
> 
> ```python
> @test_func_wrapper
> def test_event():
>     scr = lv.screen_active()
>     
>     flag = False
>     scr.add_event_cb(event_callback, lv.EVENT.CLICKED, flag)
> 
>     scr.send_event(lv.EVENT.CLICKED, None)
> 
>     if not flag:
>         raise RuntimeError('Event failure')
> ```

Hi!
I also don't use `user_data` to pass data to event handler, and giving back result through `user_data` is also less usable as simply changing a Python variable in class or anywhere else, or using other way to inform caller/other code about the result.
So I think `user_data` in event handlers is most cases useless, so that's why we mostly pass `None` to `add_event_cb()` functions.

kdschlosser:
when dealing with MicroPython I do not believe the userdata field is something that is able to be used. This is because MicroPython uses the userdata field in the structures to store the callback. I would have to look at the C code to see what it does. Setting the global works for the time being. Now there is a test for checking the event system to make sure it is working.

The other 1/2 of this PR was done in the micropython binding repo and that PR actually fixed more than just the events getting broken. It also fixed all of the structures that were not getting populated with fields as well. any forward declaration of a structure or union now gets condensed into a single object so the gen_mpy script will read the object properly.

kdschlosser:
To answer your question @XuNeo 

That will not work. it's because a copy of `flag` that gets passed and not a pointer to `flag`

```python
@test_func_wrapper
def test_event():
    scr = lv.screen_active()
    
    flag = False
    scr.add_event_cb(event_callback, lv.EVENT.CLICKED, flag)

    scr.send_event(lv.EVENT.CLICKED, None)

    if not flag:
        raise RuntimeError('Event failure')
```

This however would work

```python

@test_func_wrapper
def event_callback(e):
    e.param.flag = True

@test_func_wrapper
def test_event():
    scr = lv.screen_active()
    
    class SomeClass:
        flag = False

    scr.add_event_cb(event_callback, lv.EVENT.CLICKED, None)

    scr.send_event(lv.EVENT.CLICKED, SomeClass)

    if not SomeClass.flag:
        raise RuntimeError('Event failure')
```

Instead of using userdata we are using the param field that gets set directly to the event that gets created. IDK. In theory this should work, not 100% sure though. It would need to be tested. the type for param is `void * ` so I would think it would work because the mechanics of the gen_mpy script should know that it's a pointer and it would convert the passed in object to a pointer to store it and it would undo that when collecting it from the param field in the event structure.

PR #5793, opened 2024-03-05
feat(png): add png8 bit support

XuNeo:

For hardwares support indexed8 image, decode PNGs with palette to indexed color format, so GPU can use it directly.

### Notes

### DIFF
--- a/src/libs/lodepng/lodepng.c
+++ b/src/libs/lodepng/lodepng.c
@@ -5306,8 +5306,26 @@ static void decodeGeneric(unsigned char ** out, unsigned * w, unsigned * h,
     if(!state->error && scanlines_size != expected_size) state->error = 91; /*decompressed size doesn't match prediction*/
     lodepng_free(idat);
 
+    uint32_t offset = 0;
     if(!state->error) {
-        lv_draw_buf_t * decoded = lv_draw_buf_create(*w, *h, LV_COLOR_FORMAT_ARGB8888, 4 * *w);
+        LodePNGColorMode * color;
+        color = &state->info_png.color;
+        lv_draw_buf_t * decoded;
+        lv_color_format_t cf;
+        uint32_t stride;
+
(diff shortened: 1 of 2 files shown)


kisvegabor:
Good idea! Please add a test, and we are good to go.

PGNetHun (changes requested):
Sorry, but I think we should not change this 3rd party library file (`lodepng.c`) too much, we need to check first whether we can solve our problem with existing library functions.

As I see [lodepng_decode_memory](https://github.com/lvandeve/lodepng/blob/master/lodepng.h#L153) function has a `colortype` parameter, and [LCT_PALETTE](https://github.com/lvandeve/lodepng/blob/master/lodepng.h#L124) enum value might be worth to check, and investigate whether it helps us to accomplish what we want.

Btw this `lodepng.c` file is outdated (20201017 vs [latest 20230410](https://github.com/lvandeve/lodepng/blob/master/lodepng.h) ), so that's why we should also avoid heavily changing 3rd party library files;  upgrading will be simpler if not so many things are changed.

XuNeo:
I don't see how to make the draw buffer alloc function work without hacking into lodepng source code. I'd like adding original code plus lvgl patches applied.

kisvegabor:
> I'd like adding original code plus lvgl patches applied.

Do you mean store `lodepng.patch` next to the `lodepng.c` and apply it when `lodepng` is updated?

XuNeo:
> Do you mean store `lodepng.patch` next to the `lodepng.c` and apply it when `lodepng` is updated?

Yes, but the code in lvgl repo should be already patched, so users can use it out-of-box, just like current status.
In the mean time, if we need to upgrade lodepng. Then grab the latest version source code and patch it again.


For this PR, we still need to find a better way instead of hacking into lodepng to make such big modification. I'm currently investigating `libpng`, and hope can get some ideads.

PR #5800, opened 2024-03-05
feat(draw/sw): added support for LV_COLOR_FORMAT_L8

zjanosy:

Implemented rendering L8, RGB565, RGB888 and ARGB8888 into L8 canvas.
Implemented rendering L8 format image into L8, RGB565, RGB888 and ARGB8888 canvas.

### Notes
Documentation, examples and tests are WIP.


liamHowatt:
All changes make sense to me at first glance.

L8 cogwheel image looks okay on `COLOR_DEPTH` 8, 16, 24, 32.
L8,RGB565,RGB888,ARGB8888 cogwheel image looks okay on `COLOR_DEPTH` 8.
Blending untested.

It's a large space of possibilities to test. The eventual test cases will reveal any non-obvious problems.

razor5k:
Hi,
does it make sense to add integer rounding before divide operations?

//example
static inline uint8_t lv_color_luminance(lv_color_t c)
{
    return (uint8_t)((uint16_t)(77u * c.red + 151u * c.green + 28u * c.blue) >> 8);
}
//to
static inline uint8_t lv_color_luminance(lv_color_t c)
{
    return (uint8_t)((uint16_t)(77u * c.red + 151u * c.green + 28u * c.blue + 128) >> 8);
}

zjanosy:
> Hi, does it make sense to add integer rounding before divide operations?

I don't think it would be visible, and it would be a little bit slower. In fact, this formula is already an approximation, so I would not bother. Maybe an even simpler (and faster) approximation would do, like ((c.red << 1) + (c.green << 2) + c.blue) >> 3, although this may need some scaling to reach the maximum value of 255.
Here is a comparison of using the "real" formula and the crude approximation:
lenna_0 3R+0 59G+0 11B
lenna_0 25R+0 5G+0 13B
There is little visible difference, except that the latter is a bit darker (which could be fixed by an additional scaling, but then we loose the performance gain -- if there is any on modern architectures).

razor5k:
You made a valid point not to round. If you don't see a huge difference, it may be better to be faster.

kisvegabor:
Hey,

Thank you or the PR! It was already requested by a lot of people! :slightly_smiling_face: 

I've tested it like this:

```c
  lv_demo_widgets();

  lv_draw_buf_t * draw_buf = lv_snapshot_create_draw_buf(lv_screen_active(), LV_COLOR_FORMAT_L8);
  lv_snapshot_take_to_draw_buf(lv_screen_active(), LV_COLOR_FORMAT_L8, draw_buf);

  lv_obj_clean(lv_screen_active());
  lv_obj_t * canvas = lv_canvas_create(lv_screen_active());
  lv_canvas_set_draw_buf(canvas, draw_buf);
```

image
You can notice that
- the center of the slider is empty
- the shadow below the the buttons are missing

```c
lv_demo_render(LV_DEMO_RENDER_SCENE_FILL, LV_OPA_50);
```
image

ARGB as reference
image

---

I've updated the tests to support rendering to L8 and added the [reference images](https://github.com/lvgl/lvgl/pull/5800/files#diff-152c99c22d1d3b852eac1b5bc3ee2f133ae266fd2c7be90ec75584c72237b0d6) as well for [this test case](https://github.com/lvgl/lvgl/pull/5800/files#diff-c01a46513a453dcc584488d3b69009180e852a7b5f50364facae05ee4032a832). It seems there are a few places where the rendering is not correct. Could you take a look at these?

kisvegabor:
The CI passing for me locally, but here is an off by one error [here](https://github.com/lvgl/lvgl/actions/runs/8241211613/job/22538005682?pr=5800#step:5:1205).  Do you have any idea why the rendering can be platform dependent?

kisvegabor:
FYI, the issue is here:
image

So when a scaled ARGB8888 image is blended to L8.

kisvegabor:
And it's passing now. We will see if the passing or the failure have happened by "accident".

I've discussed it with @zjanosy that he will update `lv_demo_render` with blending A8 and RGB565A8 images too (so use these as source and not destination)

zjanosy:
I don't understand why the code formatting check fails. I have run the code format script several times. Actually, the file it fails in (examples/others/fragment/lv_example_fragment_1.c) is not part of my contribution.

kisvegabor:
Reminder, `lv_example_label_4` need to updated too.

tests/test_images/stride_align1/LZ4/test_A1_LZ4_align1.c
```diff
@@ -45,14 +45,14 @@ uint8_t test_A1_LZ4_align1_map[] = {
...
-const lv_img_dsc_t test_A1_LZ4_align1 = {
+const lv_image_dsc_t test_A1_LZ4_align1 = {
   .header.magic = LV_IMAGE_HEADER_MAGIC,
   .header.cf = LV_COLOR_FORMAT_A1,
   .header.flags = 0 | LV_IMAGE_FLAGS_COMPRESSED,
   .header.w = 71,
   .header.h = 60,
   .header.stride = 16,
-  .data_size = 347,
+  .data_size = sizeof(test_A1_LZ4_align1_map),
```
kisvegabor: These should be updated in a separate PR.  I can help with reverting if needed.

kisvegabor (commented):
Thank you! I couldn't compile it because `img_render_lvgl_logo_l8` is missing? Maybe you have forgotten pushing it.

tests/test_images/stride_align1/LZ4/test_A1_LZ4_align1.c
zjanosy: I'm not sure what to do with this. I think these were generated by the test script.

demos/render/assets/img_render_lvgl_logo_l8.c
```diff
@@ -0,0 +1,66 @@
+
+#if defined(LV_LVGL_H_INCLUDE_SIMPLE)
+#include "lvgl.h"
+#elif defined(LV_BUILD_TEST)
+#include "../lvgl.h"
```
kisvegabor: Use an include similar to the other images.

kisvegabor (commented):
I've just fixed the code formatting issue.

demos/render/assets/img_render_lvgl_logo_l8.c
zjanosy: Fixed.

kisvegabor (dismissed):
I've removed the accidentally changed test files. 

The rest is looking good!

src/core/lv_obj_scroll.c
```diff
@@ -309,7 +309,7 @@ void lv_obj_scroll_by(lv_obj_t * obj, int32_t dx, int32_t dy, lv_anim_enable_t a
         lv_anim_t a;
         lv_anim_init(&a);
         lv_anim_set_var(&a, obj);
-        lv_anim_set_completed_cb(&a, scroll_completed_completed_cb);
```
FASTSHIFT: Why was it modified? it comes from this fix: https://github.com/lvgl/lvgl/pull/5979

kisvegabor: I reverted it.

scripts/LVGLImage.py
```diff
@@ -819,6 +819,16 @@ def _png_to_alpha_only(self, cf: ColorFormat, filename: str):
                 rawdata += row
 
         self.set_data(cf, w, h, rawdata)
+    
+    def sRGB_to_linear(self, x):
+        if x < 0.04045:
+            return x/12.92
+        return pow((x+0.055)/1.055, 2.4)
```
XuNeo: Format a bit.
```suggestion
        return pow((x + 0.055) / 1.055, 2.4)
```

scripts/LVGLImage.py
```diff
@@ -819,6 +819,16 @@ def _png_to_alpha_only(self, cf: ColorFormat, filename: str):
...
+
+    def linear_to_sRGB(self, y):
+        if y <= 0.0031308:
+            return 12.92 * y
+        return 1.055 * pow(y, 1/2.4) - 0.055
```
XuNeo:
```suggestion
        return 1.055 * pow(y, 1 / 2.4) - 0.055
```

scripts/LVGLImage.py
```diff
@@ -819,6 +819,16 @@ def _png_to_alpha_only(self, cf: ColorFormat, filename: str):
                 rawdata += row
 
         self.set_data(cf, w, h, rawdata)
+    
+    def sRGB_to_linear(self, x):
+        if x < 0.04045:
+            return x/12.92
```
XuNeo:
```suggestion
            return x / 12.92
```

scripts/LVGLImage.py
```diff
@@ -831,8 +841,11 @@ def _png_to_luma_only(self, cf: ColorFormat, filename: str):
             A = row[3::4]
             for r, g, b, a in zip(R, G, B, A):
                 r, g, b, a = color_pre_multiply(r, g, b, a, self.background)
+                r = self.sRGB_to_linear(r/255.0)
+                g = self.sRGB_to_linear(g/255.0)
+                b = self.sRGB_to_linear(b/255.0)
```
XuNeo:
```suggestion
                r = self.sRGB_to_linear(r / 255.0)
                g = self.sRGB_to_linear(g / 255.0)
                b = self.sRGB_to_linear(b / 255.0)
```

scripts/LVGLImage.py
```diff
@@ -831,8 +841,11 @@ def _png_to_luma_only(self, cf: ColorFormat, filename: str):
...
                 luma = 0.2126 * r + 0.7152 * g + 0.0722 * b
-                rawdata += uint8_t(int(luma))
+                rawdata += uint8_t(int(self.linear_to_sRGB(luma)*255))
```
XuNeo:
```suggestion
                rawdata += uint8_t(int(self.linear_to_sRGB(luma) * 255))
```

src/misc/lv_color.h
```diff
@@ -110,6 +111,11 @@ typedef struct {
     uint8_t v;
 } lv_color_hsv_t;
 
+typedef struct {
+    uint8_t lumi;
+    uint8_t alpha;
+} lv_color16a_t;
```
XuNeo, line 117: Should it be `lv_color8a_t`?

scripts/LVGLImage.py
```diff
@@ -819,6 +819,16 @@ def _png_to_alpha_only(self, cf: ColorFormat, filename: str):
...
+        if x < 0.04045:
+            return x/12.92
+        return pow((x+0.055)/1.055, 2.4)
+
+    def linear_to_sRGB(self, y):
+        if y <= 0.0031308:
+            return 12.92 * y
+        return 1.055 * pow(y, 1/2.4) - 0.055
 
     def _png_to_luma_only(self, cf: ColorFormat, filename: str):
```
XuNeo, line 833: Is the luma here should be called lumi(nance) instead?

XuNeo (dismissed):

src/misc/lv_color.h
kisvegabor: `lv_color16a_t` is really misleading as `lv_color16_t` is RGB565. `lv_color_al88_t` would be cleanest, but it's different style. `lv_color8a_t` can be a good trade off.

zjanosy: While I agree that `lv_color16a_t` is not the best name, the naming scheme was that the number in the color format name indicates the number of bits of the value. Therefore I think color8a is a bit misleading, too, because it suggest that it is a 8 bit value, which it is not. Maybe `lv_color8a8_t` would be better (although `ARGB8888` is called `lv_color32_t`, and not something like `lv_color24a` or `lv_color24a8`).
Certainly, the cleanest solution would be to call them as the corresponding color format is called, i.e.:
`lv_color32_t` -> `lv_argb8888_t`
`lv_color16_t` -> `lv_rgb565_t`
`lv_color16a_t` -> `lv_al88_t`
But this would likely break too much code.

scripts/LVGLImage.py
zjanosy: Yeah, luma and luminance are different terms:
https://www.geeksforgeeks.org/difference-between-luma-and-luminance/
But honestly I don't know which is the correct name in our case.

src/misc/lv_color.h
kisvegabor: Actually we can rename these by adding `typedef lv_color_argb8888_t lv_color32_t` in `lv_api_map_v9.h`.

XuNeo: > lv_color32_t -> lv_argb8888_t
lv_color16_t -> lv_rgb565_t
lv_color16a_t -> lv_al88_t

I agree with it.  But it should be `lv_color16_t --> lv_al88_t` where 16 means it takes 16bits. There will be `lv_color[32|24|16|8]_t.`


> Actually we can rename these by adding typedef lv_color_argb8888_t lv_color32_t in lv_api_map_v9.h.

Let's go for it. It can be a separate PR. We can merge this massive achievement firstly.

XuNeo (approved):

kisvegabor (approved):

kisvegabor:
Merged, thank you!

GorgonMeducer:
I plan to add the Arm-2D (Helium) support for L8. In fact, in arm-2d, we call L8 the GRAY8.

kisvegabor:
Amazing, thank you in advance!

PR #5828, opened 2024-03-11
feat(tests): add freetype font stress tests

W-Mai:

Generate random unicodes and use random fonts to create random position labels, then randomly delete random label and related random fonts.

# MERGE BELOW 1ST

- #5834 

### Notes


W-Mai:
I found that ci errors were reported because gcc optimized the order of calculating random numbers during compilation, resulting in some results being reversed 🤦

W-Mai:
## DON'T MERGE NOW

I'm trying to upgrade system version to Ubuntu 22.04 to fix freetype version problem

W-Mai:
😭

1. The container of aarch64 supports ubuntu 22.04, 22.04 does not support libjpeg62-turbo-dev, which needs to be replaced by jibjpeg-turbo8-dev, but supports the latest version of freetype

2. The container of armv6 does not support ubuntu 22.04, you have to use bullseye, and then does not support libjpeg-turbo8-dev. You need to use jibjpeg62-turbo-dev, and it does not support freetype. Resulting in a few pixels difference in the drawing.

PGNetHun (changes requested):
Sorry, but I think this `rnd_unicodes` does not strictly belong to LVGL "production" code, I doubt that anybody will use it in his/her project; this library is used only for tests, so it is only a **test helper library** to generate random Unicode characters. 

So, no need to add a new config (`KConfig`, `lv_conf_template.h`) for it, or add to `lvgl.h` or to `src/libs/` folder, simply add the library to `tests/src/` folder or `tests/libraries`, and reference it from test files. 

PS:
we should separate the "production" codes/libraries and test codes/helper libraries.

W-Mai:
> Sorry, but I think this `rnd_unicodes` does not strictly belong to LVGL "production" code, I doubt that anybody will use it in his/her project; this library is used only for tests, so it is only a **test helper library** to generate random Unicode characters.
> 
> So, no need to add a new config (`KConfig`, `lv_conf_template.h`) for it, or add to `lvgl.h` or to `src/libs/` folder, simply add the library to `tests/src/` folder or `tests/libraries`, and reference it from test files.
> 
> PS: we should separate the "production" codes/libraries and test codes/helper libraries.

I agree with you, but now we don't seem to have such a directory to store the library used for testing.

W-Mai:
@PGNetHun How about now?
I create a new folder `test_libs` in `tests/src/` folder.
And refs theme in the CMakeFiles. So others can put all other third part libs in this folder, and can be include in all `test_cases`.

W-Mai:
The same errors happened. https://github.com/lvgl/lvgl/actions/runs/8274508818/job/22639971283?pr=5864#step:5:579

W-Mai:
There seems to be a problem with the image of Ubuntu 22.04,

- https://github.com/actions/runner-images/issues/9491
- https://github.com/DPDK/dpdk/commit/d0d7032a3f6e66c1a09b13ad42e9d7975db8c627
- https://github.com/apache/kvrocks/pull/2161

tests/src/lv_test_conf_full.h
```diff
@@ -80,6 +80,7 @@
 #define LV_USE_VECTOR_GRAPHIC   1
 #define LV_USE_PROFILER         1
 #define LV_PROFILER_INCLUDE     "lv_profiler_builtin.h"
+#define LV_USE_RND_UNICODES     1
```
PGNetHun: This is not needed any more.

tests/src/test_cases/libs/test_font_stress.c
```diff
@@ -0,0 +1,208 @@
+#if LV_BUILD_TEST
+#include "../lvgl.h"
+#include "rnd_unicodes/lv_rnd_unicodes.h"
```
PGNetHun: I think these includes should go inside `#if LV_USE_FREETYPE`, because if FreeType library is disabled, then we don't use these.

tests/src/test_cases/libs/test_font_stress.c
```diff
@@ -0,0 +1,208 @@
...
+
+        if(g_ctx.config.loop_cnt % CAPTURE_SKIP_FRAMES == 0) {
+            char buf[64];
+            TEST_FREETYPE_ASSERT_EQUAL_SCREENSHOT(i);
+            i++;
+        }
+    }
+}
+
+#else
```
PGNetHun, line 193: Can you please check, if this file is compliled without `#else` ?

PGNetHun (changes requested):
Thank you for moving test library out of production code libraries.

W-Mai:
CI fixing is here: 
- #5867

tests/src/test_cases/libs/test_font_stress.c
W-Mai: It seems that there must be a test function in each testcase file, so an empty dummy function is required.

If I remove the dummy function, it will cause

image

tests/src/test_cases/libs/test_font_stress.c
W-Mai: done

tests/src/lv_test_conf_full.h
W-Mai: Done

PGNetHun (approved):

FASTSHIFT (approved):

PR #5750, opened 2024-02-26
feat(docs): add Chinese translation (jump link)

100ask:
# Preview

image

**out_html**: [http://photos.100ask.net/lvgl/out_html.7z](http://photos.100ask.net/lvgl/out_html.7z)

PR #5839, opened 2024-03-12
feat(docs): add Chinese translation (jump link)

100ask:
I created this PR with a new branch, the old PR: #5750

### DIFF
--- a/docs/_ext/link_roles.py
+++ b/docs/_ext/link_roles.py
@@ -10,6 +10,9 @@ from collections import namedtuple
 from docutils import nodes
 from sphinx.transforms.post_transforms import SphinxPostTransform
 
+URL_BASE = {
+    "zh_CN": "https://lvgl.100ask.net/"
+}
 
 class translation_link(nodes.Element):
     """Node for "link_to_translation" role."""
@@ -27,10 +30,6 @@ class TranslationLinkNodeTransform(SphinxPostTransform):
     default_priority = 0
 
     def run(self, **kwargs):
-
-        URL_BASE = ''
-        URL_BASE_ZH_CN = 'https://lvgl.100ask.net/'
-
         # Only output relative links if building HTML
         for node in self.document.traverse(translation_link):
             if 'html' in self.app.builder.name:
@@ -38,13 +37,9 @@ class TranslationLinkNodeTransform(SphinxPostTransform):
                 (language, link_text) = text.split(':')
                 env = self.document.settings.env
                 docname = env.docname
-                doc_path = env.doc2path(docname, False)
+                #doc_path = env.doc2path(docname, False)
                 urlpath = os.environ['LVGL_URLPATH']+'/'
-
-                if language == "zh_CN":
-                    URL_BASE = URL_BASE_ZH_CN
-
-                return_path = URL_BASE+urlpath
+                return_path = URL_BASE.get(language, "") + urlpath
                 
                 url = '{}.html'.format(os.path.join(return_path, docname))
 
--- a/docs/_ext/link_roles.py
+++ b/docs/_ext/link_roles.py
@@ -1,0 +1,63 @@
+# based on http://protips.readthedocs.io/link-roles.html
+
+#from __future__ import print_function, unicode_literals
+
+import os
+import re
+import subprocess
+from collections import namedtuple
+
+from docutils import nodes
+from sphinx.transforms.post_transforms import SphinxPostTransform
+
+
+class translation_link(nodes.Element):
+    """Node for "link_to_translation" role."""
+
+
+# Linking to translation is done at the "writing" stage to avoid issues with the info being cached between builders
+def link_to_translation(name, rawtext, text, lineno, inliner, options={}, content=[]):
+    node = translation_link()
+    node['expr'] = (rawtext, text, options)
+    return [node], []
+
+
+class TranslationLinkNodeTransform(SphinxPostTransform):
+    # Transform needs to happen early to ensure the new reference node is also transformed
+    default_priority = 0
+
+    def run(self, **kwargs):
+
+        URL_BASE = ''
+        URL_BASE_ZH_CN = 'https://lvgl.100ask.net/'
+
+        # Only output relative links if building HTML
+        for node in self.document.traverse(translation_link):
+            if 'html' in self.app.builder.name:
+                rawtext, text, options = node['expr']
+                (language, link_text) = text.split(':')
+                env = self.document.settings.env
+                docname = env.docname
+                doc_path = env.doc2path(docname, False)
+                urlpath = os.environ['LVGL_URLPATH']+'/'
+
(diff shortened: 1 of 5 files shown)


kisvegabor (dismissed):
Looks good! Thank you! 😊

docs/_ext/link_roles.py
```diff
@@ -0,0 +1,56 @@
...
+        # Only output relative links if building HTML
+        for node in self.document.traverse(translation_link):
+            if 'html' in self.app.builder.name:
+                rawtext, text, options = node['expr']
+                (language, link_text) = text.split(':')
+                env = self.document.settings.env
+                docname = env.docname
+                doc_path = env.doc2path(docname, False)
+                urlpath = os.environ['LVGL_URLPATH']+'/'
+                return_path = "https://lvgl.100ask.net/"+urlpath
```
PGNetHun: Can you please move the hard-coded URL to top of this Python file as a variable (constant)?
So the URL will not be "hidden" in source code.

Like this:
```
URL_BASE = 'https://lvgl.100ask.net/'
....
    url_path = os.environ['LVGL_URLPATH'] + '/'
    return_path = URL_BASE + url_path

```

docs/integration/framework/arduino.rst
```diff
@@ -41,6 +44,7 @@ installed, follow these configuration steps:
 2. Go to ``lvgl`` and copy ``lv_conf_template.h`` as ``lv_conf.h`` into the Arduino Libraries directory next to the ``lvgl`` library folder.
 3. Open ``lv_conf.h`` and change the first ``#if 0`` to ``#if 1`` to enable the content of the file
 4. Set the color depth of you display in :c:macro:`LV_COLOR_DEPTH`
+5. Set :c:macro:`LV_TICK_CUSTOM`
```
PGNetHun: This line was removed in PR #5832 
Please rebase your branch to main, and re-request review. Thank you!

PGNetHun (changes requested):
1. Some minor changes needed
2. Sorry, but I think we should show the external link ( https://lvgl.100ask.net/ ) on the home page, or in header/footer **instead of adding same `link_to_translation` to every doc file**. New doc files will not contain it (ppl might forget to add), and btw there could be other translations too, not only chinese, so having a link in header/footer/sidenav to external sites with the translations is better.

100ask:
> 2\. **instead of adding same `link_to_translation` to every doc file**. New doc files will not contain it (ppl might forget to add),

Great suggestion! I will improve it today.

100ask:
> 2\. and btw there could be other translations too

What do you think? @kisvegabor

docs/_ext/link_roles.py
```diff
@@ -0,0 +1,63 @@
...
+
+
+class TranslationLinkNodeTransform(SphinxPostTransform):
+    # Transform needs to happen early to ensure the new reference node is also transformed
+    default_priority = 0
+
+    def run(self, **kwargs):
+
+        URL_BASE = ''
+        URL_BASE_ZH_CN = 'https://lvgl.100ask.net/'
```
PGNetHun: Please move to top of file: after the import lines, but **before class definitions**,
and you could change to dictionary, so no need for if-else in code:
```
URL_BASE = {
    "zh_CN": "https://lvgl.100ask.net/"
}
.....
return_path = URL_BASE.get(language, "") + urlpath
....
```

docs/add_translation.py
```diff
@@ -0,0 +1,30 @@
...
+
+
+def exec(temp_directory):
+    files = find_files(temp_directory, '.rst')
+
+    for rst_file in files:
+
+        temp_filename = rst_file + '.tmp'
+
+        with open(temp_filename, 'w', encoding='utf-8') as temp_file:
```
PGNetHun: All these lines can be written shorter, without a temp file:
```
def line_prepender(filename, line):
    with open(filename, 'r+') as f:
        content = f.read()
        f.seek(0, 0)
        f.write(line.rstrip('\r\n') + '\n' + content)
```
(source: https://stackoverflow.com/a/5917395 )

Just make sure "\n" is added twice after the inserted line (or remove `.rstrip()`).

Please also check, that file remains UTF-8 ([BOM characters](https://en.wikipedia.org/wiki/Byte_order_mark) are still there in the beginning of file)

PGNetHun (changes requested):
Good changes, thanks, only some minor comments are added.

PS: 
you don't have to squash commits in your branch, simply add a new commit, so reviewers can see what is changed since last review (since last commit).
Then after PR is approved, we do a squash merge, and in the end only 1 commit will go to `master` branch. But till then it is good to see all commits and change commits in feature branch.

docs/_ext/link_roles.py
100ask: Agree, I will improve.

docs/add_translation.py
100ask: I will try to improve it.

P.S. `lvgl/docs/add_translation.py` executed at compile time (temporary directory) without any actual modifications to the files in `lvgl/docs`.

kisvegabor:
> Sorry, but I think we should show the external link ( https://lvgl.100ask.net/ ) on the home page, or in header/footer instead of adding same link_to_translation to every doc file. 

What about near to the image link?

100ask:
> > Sorry, but I think we should show the external link ( https://lvgl.100ask.net/ ) on the home page, or in header/footer instead of adding same link_to_translation to every doc file.
> 
> What about near to the image link?

1. Not obvious, may not notice.
2. When reading on a small screen, it becomes difficult, even invisible:  

b6ebefb9-cf18-4511-87c4-e9a0c80a1fc0

docs/add_translation.py
```diff
@@ -0,0 +1,23 @@
...
+
+
+def exec(temp_directory):
+    files = find_files(temp_directory, '.rst')
+
+    for rst_file in files:
+        with open(rst_file, 'r+', encoding='utf-8') as f:
+            content = f.read()
+            f.seek(0, 0)
+            f.write(':link_to_translation:`zh_CN:[中文]`\n\n' + content)
```
PGNetHun, line 23: At the moment adding this hard-coded 'zh_CN' is OK, but during any new code development please think on code extendibility, too, if we want to re-use/extend it later; 
I mean in this case the language code could come from function parameter ( `exec(temp_directory, language_code)` ), so in the future it would be enough to call `exec()` again with different language, and the code adds the new language line.

PGNetHun (approved):

kisvegabor (approved):
It's unlikely that we will add more translations in the future due to maintenance issues. China is special because:
- We have a lot of users in China
- Many people don't know English
- Websites outside of China are super slow

docs/add_translation.py
100ask: Good idea, other translations(perhaps) can be easily added in the future.
I will improve it in the new PR.  #5874

PR #5851, opened 2024-03-13
fix(drm): Default to XRGB8888 framebuffer

marex:
The ARGB8888 framebuffer format for base canvas makes little sense as the base canvas is unlikely to be transparent and require alpha. Use XRGB8888 framebuffer format which is more widely supported by DRM drivers as base plane pixel format.

This makes e.g. i.MX8M Nano work by default.

---

This is identical fix to lv_drivers https://github.com/lvgl/lv_drivers/pull/282

### DIFF
--- a/src/drivers/display/drm/lv_linux_drm.c
+++ b/src/drivers/display/drm/lv_linux_drm.c
@@ -24,7 +24,7 @@
  *      DEFINES
  *********************/
 #if LV_COLOR_DEPTH == 32
-    #define DRM_FOURCC DRM_FORMAT_ARGB8888
+    #define DRM_FOURCC DRM_FORMAT_XRGB8888
 #elif LV_COLOR_DEPTH == 16
     #define DRM_FOURCC DRM_FORMAT_RGB565
 #else


W-Mai (approved):

PGNetHun (approved):

W-Mai (approved):

PGNetHun (approved):

kisvegabor (approved):
It really makes sense. And it's faster too!

marex:
> It really makes sense. And it's faster too!

In the meantime, I can imagine one use case where base plane with alpha would make sense, chroma keying (single color background with some object in front), but in that case, I think the user would configure the library accordingly.

On which platform is XRGB8888 base plane faster than ARGB8888 ? Base plane alpha should be resolved in the CRTC only, since that would be the chroma key . Unless, on this platform where it is now faster, the alpha is handled by reading in two planes unconditionally , while without alpha only one plane is read , that might explain why it is faster without alpha .

kisvegabor:
Note that v9 doesn't support chroma keying. We have dropped it as it wasn't really used. 

Rendering to ARGB8888 is slower in LVGL because when 2 colors are mixed LVGL needs to mix both the RGB colors and the alpha channel. As a reference [this function](https://github.com/lvgl/lvgl/blob/master/src/draw/sw/blend/lv_draw_sw_blend_to_argb8888.c#L574-L612) mixes colors with and ARGB destination. And [this](https://github.com/lvgl/lvgl/blob/master/src/draw/sw/blend/lv_draw_sw_blend_to_rgb888.c#L586-L602) is for an XRGRB8888.

marex:
> Note that v9 doesn't support chroma keying. We have dropped it as it wasn't really used.

OK

> Rendering to ARGB8888 is slower in LVGL because when 2 colors are mixed LVGL needs to mix both the RGB colors and the alpha channel.

I was writing about hardware variant of this, where the CRTC samples multiple input planes and does the mixing for you, this stuff:
- https://docs.kernel.org/gpu/drm-kms.html#overview
- https://docs.kernel.org/gpu/drm-kms.html#crtc-abstraction
- https://docs.kernel.org/gpu/drm-kms.html#plane-abstraction

There you can usually turn the input planes on/off , so the CRTC wouldn't sample them and waste DRAM bandwidth.

> As a reference [this function](https://github.com/lvgl/lvgl/blob/master/src/draw/sw/blend/lv_draw_sw_blend_to_argb8888.c#L574-L612) mixes colors with and ARGB destination. And [this](https://github.com/lvgl/lvgl/blob/master/src/draw/sw/blend/lv_draw_sw_blend_to_rgb888.c#L586-L602) is for an XRGRB8888.

Right, sure, software blending is obviously going to be more expensive.

kisvegabor:
Yeah, I see. Blending the final ARGB layer is really done by the hardware. I mean only the process of LVGL rendering.

PR #5852, opened 2024-03-13
fix(sdl): handle both LV_IMAGE_SRC_FILE and LV_IMAGE_SRC_VARIABLE

marex:
The SDL image draw code currently assumes that the image source is a filename and attempts to open that filename. This is not necessarily the case, e.g. the lv_demo_fb uses encoded images which are of type LV_IMAGE_SRC_VARIABLE and instead of filename, come with a buffer of pixels. Handle the later using SDL_CreateRGBSurfaceFrom().

### DIFF
--- a/src/draw/sdl/lv_draw_sdl.c
+++ b/src/draw/sdl/lv_draw_sdl.c
@@ -224,10 +224,41 @@ static bool draw_to_texture(lv_draw_sdl_unit_t * u, cache_data_t * data)
             break;
         case LV_DRAW_TASK_TYPE_IMAGE: {
                 lv_draw_image_dsc_t * image_dsc = task->draw_dsc;
-                const char * path = image_dsc->src;
-                SDL_Surface * surface = IMG_Load(&path[2]);
-                if(surface == NULL) {
-                    fprintf(stderr, "could not load image: %s\n", IMG_GetError());
+                lv_image_src_t type = lv_image_src_get_type(image_dsc->src);
+                SDL_Surface * surface = NULL;
+                if(type == LV_IMAGE_SRC_FILE) {
+                    const char * path = image_dsc->src;
+                    surface = IMG_Load(&path[2]);
+                    if(surface == NULL) {
+                        LV_LOG_ERROR("could not load image from file: %s", IMG_GetError());
+                        return false;
+                    }
+                }
+                else if(type == LV_IMAGE_SRC_VARIABLE) {
+                    lv_image_dsc_t * lvd = image_dsc->src;
+                    surface = SDL_CreateRGBSurfaceFrom(lvd->data,
+                                                       lvd->header.w, lvd->header.h,
+                                                       LV_COLOR_FORMAT_GET_BPP(lvd->header.cf),
+                                                       lvd->header.stride,
+#if SDL_BYTEORDER == SDL_LIL_ENDIAN
+                                                       0x00FF0000,
+                                                       0x0000FF00,
+                                                       0x000000FF,
+                                                       0xFF000000
+#else
+                                                       0x0000FF00,
+                                                       0x00FF0000,
+                                                       0xFF000000,
+                                                       0x000000FF
+#endif
+                                                      );
+                    if(surface == NULL) {
+                        LV_LOG_ERROR("could not load image from variable");
+                        return false;
+                    }
+                }
+                else {
+                    LV_LOG_WARN("image source type unknown");
                     return false;
                 }
 


src/draw/sdl/lv_draw_sdl.c
```diff
@@ -224,10 +224,34 @@ static bool draw_to_texture(lv_draw_sdl_unit_t * u, cache_data_t * data)
...
+                                                       0x00FF0000,
+                                                       0xFF000000,
+                                                       0x000000FF
+#endif
+                                                      );
+                }
+
                 if(surface == NULL) {
-                    fprintf(stderr, "could not load image: %s\n", IMG_GetError());
+                    fprintf(stderr, "could not load image\n");
```
PGNetHun: 1. Use `LV_LOG_ERROR` instead of `fprintf`
2. Move `LV_LOG_ERROR` and `return false;` to `if` - `else if` branches, so you can add the `else` branch also that should call `LV_LOG_WARN` that image type is unknown, and also return false.
3. You can add again `IMG_GetError` to source file branch's `LV_LOG_ERROR`

PGNetHun (changes requested):

src/draw/sdl/lv_draw_sdl.c
marex: @PGNetHun should be fixed now.

src/draw/sdl/lv_draw_sdl.c
```diff
@@ -224,10 +224,37 @@ static bool draw_to_texture(lv_draw_sdl_unit_t * u, cache_data_t * data)
...
+#endif
+                                                      );
+                }
+                else {
+                    LV_LOG_WARN("image source type unknown\n");
+                }
+
                 if(surface == NULL) {
-                    fprintf(stderr, "could not load image: %s\n", IMG_GetError());
+                    LV_LOG_ERROR("could not load image: %s\n", IMG_GetError());
```
PGNetHun: `IMG_GetError()` belongs to `IMG_Load`.

My suggestion was:
```
if(type == LV_IMAGE_SRC_FILE) {
 ...
 surface = IMG_Load(&path[2]);
 if(surface == NULL) {
   LV_LOG_ERROR(....., IMG_GetError());
   return false;
 }
}
else if(type == LV_IMAGE_SRC_VARIABLE) {
 ...
 surface = SDL_CreateRGBSurfaceFrom(....);
 if(surface == NULL) {
  LV_LOG_ERROR(.....);
  return false;
 }
}
else {
 LV_LOG_WARN(....);
 return false;
}
```

So, when the source is file, then we get more details from `IMG_GetError()`, when source is variable, then we just log the error, and when the source is unknown, the log warning/error, and return.
The reason I would like to see the LV_LOG_ERRORs separated is, that in case of file source we might have more info why loading it failed. 
It is better to log more AND relevant information then nothing... in later bug investigations the more info will help in the investigation.

PGNetHun (changes requested):

src/draw/sdl/lv_draw_sdl.c
marex: I see, should be all better now.

src/draw/sdl/lv_draw_sdl.c
```diff
@@ -224,10 +224,41 @@ static bool draw_to_texture(lv_draw_sdl_unit_t * u, cache_data_t * data)
...
-                SDL_Surface * surface = IMG_Load(&path[2]);
-                if(surface == NULL) {
-                    fprintf(stderr, "could not load image: %s\n", IMG_GetError());
+                lv_image_src_t type = lv_image_src_get_type(image_dsc->src);
+                SDL_Surface * surface = NULL;
+                if(type == LV_IMAGE_SRC_FILE) {
+                    const char * path = image_dsc->src;
+                    surface = IMG_Load(&path[2]);
+                    if(surface == NULL) {
+                        LV_LOG_ERROR("could not load image from file: %s\n", IMG_GetError());
```
PGNetHun: No need to add new line character: `\n`

src/draw/sdl/lv_draw_sdl.c
```diff
@@ -224,10 +224,41 @@ static bool draw_to_texture(lv_draw_sdl_unit_t * u, cache_data_t * data)
...
+                                                       0xFF000000
+#else
+                                                       0x0000FF00,
+                                                       0x00FF0000,
+                                                       0xFF000000,
+                                                       0x000000FF
+#endif
+                                                      );
+                    if(surface == NULL) {
+                        LV_LOG_ERROR("could not load image from variable\n");
```
PGNetHun: No need to add new line character: `\n`

src/draw/sdl/lv_draw_sdl.c
```diff
@@ -224,10 +224,41 @@ static bool draw_to_texture(lv_draw_sdl_unit_t * u, cache_data_t * data)
...
+                        return false;
+                    }
+                }
+                else {
+                    LV_LOG_WARN("image source type unknown\n");
```
PGNetHun: No need to add new line character: `\n`

PGNetHun (changes requested):
Just minor correction is needed, then PR is OK.

Btw you don't need to squash commits in your branch, just add a new commit with new changes, so reviewers can see what is changed since last review/commit. After you commit the new changes, please [re-request review](https://github.blog/changelog/2019-02-21-re-request-review-on-a-pull-request/) from reviewers.
When we merge, then all commits will be squashed into 1 merge commit to `master` branch.

marex:
> Btw you don't need to squash commits in your branch, just add a new commit with new changes, so reviewers can see what is changed since last review/commit.

There is a `Compare` button next to each new push, that gives you the diff of what changed. I can create commits with `git commit --fixup HEAD` if that makes it easier, it feels a bit like duplicating functionality though.

PGNetHun:
> > Btw you don't need to squash commits in your branch, just add a new commit with new changes, so reviewers can see what is changed since last review/commit.
> 
> There is a `Compare` button next to each new push, that gives you the diff of what changed. I can create commits with `git commit --fixup HEAD` if that makes it easier, it feels a bit like duplicating functionality though.

There is a **purpose** of having separate commits after code reviews: to see what is changed in the new commits, to see the difference to previous code review(s). 
If only 1 commit is in PR, then we don't see the history of changes you made, what you've fixed in the new commit(s) since the previous code review(s). Sometimes (basically most of the times) it is enough to review only the new changes, whether the commits went into the right direction, or not; so no need to review again the whole PR, all the files.

PGNetHun (approved):

kisvegabor:
Hi,

We can merge this PR as a temporary solution but the end goal would be to use LVGL's image decoders instead of `SDL_Image`.
In the end, I think, it's even simpler as we just need to draw the image as we draw an other primitive (rectangle, label, etc).

However, I'm not sure if we need to cache the transformed images (LVGL does the transformation once, so we cache the final result) or the un-transformed ones (SDL does the transformation on each rendering cycle).

So the question is: can you update this PR to use LVGL's image decoders? In the first let's make it work with the "LVGL does the transformation" case, as it's simpler.

marex:
> > > Btw you don't need to squash commits in your branch, just add a new commit with new changes, so reviewers can see what is changed since last review/commit.
> > 
> > 
> > There is a `Compare` button next to each new push, that gives you the diff of what changed. I can create commits with `git commit --fixup HEAD` if that makes it easier, it feels a bit like duplicating functionality though.
> 
> There is a **purpose** of having separate commits after code reviews: to see what is changed in the new commits, to see the difference to previous code review(s).

This is exactly what the `Compare`  button is for, scroll a bit up, there is this line:
```
marex force-pushed the sdl_image_var branch from e3d1c09 to eded26e          Compare
```
You don't need a new dedicated commit for that, git can show you the diff between any two commit hashes, they don't have to have any direct relationship. That's also what the `Compare` button does, the last one in this thread will lead you to this https://github.com/lvgl/lvgl/compare/e3d1c094db8e4fdceaa3d09bdf8d13ae42e08d54..eded26ecaf1d85ef3776667f825fbbda09f7e939 diff , i.e. the latest changes that I pushed .

> If only 1 commit is in PR, then we don't see the history of changes you made, what you've fixed in the new commit(s) since the previous code review(s). Sometimes (basically most of the times) it is enough to review only the new changes, whether the new commits went into the right direction, or not; so no need to review again the whole PR, all the files, only the newly changed ones.

I agree it is easier to review only the diff from last revision, but that is exactly what you get from that `Compare` button, is it not ?

I wrote already that I can create you fixup commits with `git commit --fixup ...`, those can be resolved with `git rebase --autosquash` before merging, if that makes it easier.

marex:
> Hi,
> 
> We can merge this PR as a temporary solution but the end goal would be to use LVGL's image decoders instead of `SDL_Image`. In the end, I think, it's even simpler as we just need to draw the image as we draw an other primitive (rectangle, label, etc).

This should be a bugfix, so It is probably better to have this in rather than the LVGL demo without pictures, which is how I came across this missing part.

> However, I'm not sure if we need to cache the transformed images (LVGL does the transformation once, so we cache the final result) or the un-transformed ones (SDL does the transformation on each rendering cycle).

Pardon my ignorance, I am not that deep in LVGL, what kind of transformation does this refer to ?

> So the question is: can you update this PR to use LVGL's image decoders? In the first let's make it work with the "LVGL does the transformation" case, as it's simpler.

I wonder, why is it desired to depends on SDL, or in fact system libraries in general, less ? If SDL can already load and parse the image, why not use that functionality ?

I can understand that on an embedded system, but the SDL backend seems to be available only on higher-end systems with actual OS underneath, right ?

kisvegabor:
> This should be a bugfix, so It is probably better to have this in rather than the LVGL demo without pictures, which is how I came across this missing part.

We can do that, but before merging let's conclude the points below. 

> Pardon my ignorance, I am not that deep in LVGL, what kind of transformation does this refer to ?

Sorry for being vague. I mean image rotation, scale, etc. LVGL has built in support for these so we can either use LVGL for these kind of image manipulations or use SDL instead. 

> I wonder, why is it desired to depends on SDL, or in fact system libraries in general, less ? If SDL can already load and parse the image, why not use that functionality ?

SDL is just an option. You can always use LVGL's software renderer. However if you have video card, you can sue SDL for faster rendering.

marex:
> > This should be a bugfix, so It is probably better to have this in rather than the LVGL demo without pictures, which is how I came across this missing part.
> 
> We can do that, but before merging let's conclude the points below.
> 
> > Pardon my ignorance, I am not that deep in LVGL, what kind of transformation does this refer to ?
> 
> Sorry for being vague. I mean image rotation, scale, etc. LVGL has built in support for these so we can either use LVGL for these kind of image manipulations or use SDL instead.
> 
> > I wonder, why is it desired to depends on SDL, or in fact system libraries in general, less ? If SDL can already load and parse the image, why not use that functionality ?
> 
> SDL is just an option. You can always use LVGL's software renderer. However if you have video card, you can sue SDL for faster rendering.

That is exactly my use case. In fact, I am using the SDL backend exactly because it has the ability to access the GPU. I have little interest in using software rendering or blitting,

kisvegabor (approved):

kisvegabor:
>> SDL is just an option. You can always use LVGL's software renderer. However if you have video card, you can sue SDL for faster rendering.

> That is exactly my use case. 

Great! 

Let's merge it as a temporal fix as you have suggested and migrate the LVGL's image decoders later.

Issue #5810, opened 2024-03-06
compile error when #define LV_USE_OS   LV_OS_FREERTOS


BahuMan:
### LVGL version

9.0.1

### What happened?

In Visual Studio + PlatformIO, when I change lv_conf.h to set **LV_USE_OS**  to **LV_OS_FREERTOS**, I get the following compile error:

```
In file included from C:/Users/Bart/.platformio/packages/framework-arduinoespressif32/tools/sdk/esp32s3/include/freertos/include/freertos/portable.h:51,
                 from C:/Users/Bart/.platformio/packages/framework-arduinoespressif32/tools/sdk/esp32s3/include/freertos/include/freertos/FreeRTOS.h:63,
                 from lib/lvgl/src/osal/lv_freertos.h:26,
                 from lib/lvgl/src/osal/lv_os.h:30,
                 from lib/lvgl/src/osal/lv_freertos.c:15:
lib/lvgl/src/osal/lv_freertos.c: In function 'prvCheckMutexInit':
C:/Users/Bart/.platformio/packages/framework-arduinoespressif32/tools/sdk/esp32s3/include/freertos/port/xtensa/include/freertos/portmacro.h:476:53: error: too few arguments to function 'vPortEnterCritical'
 #define portENTER_CRITICAL(mux)                     vPortEnterCritical(mux)
                                                     ^~~~~~~~~~~~~~~~~~
C:/Users/Bart/.platformio/packages/framework-arduinoespressif32/tools/sdk/esp32s3/include/freertos/include/freertos/task.h:222:35: note: in expansion of macro 'portENTER_CRITICAL'
 #define taskENTER_CRITICAL( x )   portENTER_CRITICAL( x )
                                   ^~~~~~~~~~~~~~~~~~
lib/lvgl/src/osal/lv_freertos.c:339:9: note: in expansion of macro 'taskENTER_CRITICAL'
         taskENTER_CRITICAL();
         ^~~~~~~~~~~~~~~~~~
C:/Users/Bart/.platformio/packages/framework-arduinoespressif32/tools/sdk/esp32s3/include/freertos/port/xtensa/include/freertos/portmacro.h:576:51: note: declared here 
 static inline void __attribute__((always_inline)) vPortEnterCritical(portMUX_TYPE *mux)

``` 

The problem seems to be in  lib/lvgl/src/osal/lv_freertos.c:339:9 because taskENTER_CRITICAL() is called without mutex argument.

I'm using the espressif32 platform.

[...]

PR #5862, opened 2024-03-13
Add support for FreeRTOS ported by Espressif

HanesSciarrone:

The FreeRTOS ported by Espressif changes the standard FreeRTOS API to support multi-core SoCs. Therefore, this change generate compatibility with the new Espressif API. 

Fixes issue #5810


Kconfig
```diff
@@ -135,12 +135,14 @@ menu "LVGL configuration"
 				bool "1: PTHREAD"
 			config LV_OS_FREERTOS
 				bool "2: FREERTOS"
+            config LV_OS_FREERTOS_ESP
```
kisvegabor: Check the indentation.

kisvegabor (commented):
Thank you! 

It just came to my mind that we could automatically detect the ESP environment via an `#ifdef`. E.g. `#ifdef ESP_PLATFORM` I'm not sure if really `ESP_PLATFORM` shall  be used, but we are using it in [Cmake](https://github.com/lvgl/lvgl/blob/master/CMakeLists.txt#L11C4-L11C16).

What do you think about it?

Kconfig
```diff
@@ -150,11 +152,12 @@ menu "LVGL configuration"
 			default 0 if LV_OS_NONE
 			default 1 if LV_OS_PTHREAD
 			default 2 if LV_OS_FREERTOS
-			default 3 if LV_OS_CMSIS_RTOS2
-			default 4 if LV_OS_RTTHREAD
-			default 5 if LV_OS_WINDOWS
+            default 3 if LV_OS_FREERTOS_ESP
```
PGNetHun: Use same indentation (TAB) in your new lines.

PGNetHun (changes requested):
Just minor indentation corrections are needed, the config and code changes seem OK.

src/osal/lv_freertos.c
```diff
@@ -336,8 +345,11 @@ static void prvCheckMutexInit(lv_mutex_t * pxMutex)
     if(pxMutex->xIsInitialized == pdFALSE) {
         /* Mutex initialization must be in a critical section to prevent two threads
          * from initializing it at the same time. */
+#if LV_USE_OS == LV_OS_FREERTOS
         taskENTER_CRITICAL();
-
+#elif LV_USE_OS == LV_OS_FREERTOS_ESP
+        taskENTER_CRITICAL(&critSectionMux);
+#endif
```
XuNeo: How about define a macro above to handle differences between FREERTOS and FREERTOS_ESP?

```c
        #define _enter_critical() /*need better name*/ taskENTER_CRITICAL();
        #define _exit_critical()
#elif LV_USE_OS == LV_OS_FREERTOS_ESP
        #define _enter_critical() taskENTER_CRITICAL(&critSectionMux);
        #define _exit_critical()
#endif
```

HanesSciarrone:
Hi @kisvegabor 

I've done some research on the ESP_PLATFORM macro. As far as I can see the macro is set to 1 when the CMake file is processed within the ESP-IDF build system [Documentation ESP-IDF](https://docs.espressif.com/projects/esp-idf/en/stable/esp32/api-guides/build-system.html#preset-component-variables). My doubt is if we use the Arduino framework instead of ESP-IDF, I'll test it in my development environment while writing in the issue to ask the users if they can test it also.

HanesSciarrone:
Hi @kisvegabor 

I've tested the approach you commented using the ESP_PLATFORM macro and it works perfectly. I think that it is a better approach because the modified files are less.

kisvegabor:
Great! Please update PR accordingly.

PGNetHun (approved):
Great, these new changes are even better!

HanesSciarrone:
@kisvegabor What do you mean by updating the PR accordingly? I already uploaded the changes when I replied to you.

XuNeo (approved):

kisvegabor (approved):
My bad, all look good! Thank you!

PR #5878, opened 2024-03-15
fix(SDL) change SDL to use the display color format

kdschlosser:
SDL used the COLOR_DEPTH macros and those are no longer used to define how the data gets written to the frame buffers. This corrects that so now the display color format gets used.


PGNetHun (dismissed):

src/drivers/sdl/lv_sdl_window.c
```diff
@@ -364,16 +364,21 @@ static void texture_resize(lv_display_t * disp)
...
-    SDL_PixelFormatEnum px_format = SDL_PIXELFORMAT_BGR24;
-#elif LV_COLOR_DEPTH == 16
-    SDL_PixelFormatEnum px_format = SDL_PIXELFORMAT_RGB565;
-#else
-#error("Unsupported color format")
-#endif
+    SDL_PixelFormatEnum px_format;
+
+    switch(lv_color_format_get_size(lv_display_get_color_format(disp)) * 8) {
+        case 32:
```
kisvegabor: Once we are updating it please use an ARGB format here if LVGL is configured to ARGB8888.

kisvegabor (changes requested):

src/drivers/sdl/lv_sdl_window.c
kdschlosser: I am not sure what exactly you want me to do. 

The only thing that has changed is the removal of the macros in favor of using the color format. Basically it was testing for 32 bit, 24 bit and 16 bit color using macros that really don't need to be set anymore and could actually differ than what the real color depth is. So, I switched it over to using the color format instead. I am testing the color formats to check how many bits there are and setting the same exact SDL color formats that were using before.

Are you saying that you want me to set the SDL color format properly? I can do that so it matches exactly with what the color format that LVGL is using in most cases if you want...

kdschlosser:
@kisvegabor 

I just made another commit that I think is what you were wanting.

Some of the LVGL color formats are not supported in SDL so those are left out. The most commonly used ones have a matching SDL format.

kdschlosser:
The unix build is using Ubuntu 20.04 and `libsdl2-dev` is being installed the problem is the SDL version that is being used is OLD!!! 
The package that is being installed is version 2.0.10 and the most recent version of SDL2 is 2.30.1

```shell
Error: ../../lib/lv_bindings/lvgl/src/drivers/sdl/lv_sdl_window.c:379:25: error: ‘SDL_PIXELFORMAT_XRGB8888’ undeclared (first use in this function); did you mean ‘SDL_PIXELFORMAT_ARGB8888’?
  379 |             px_format = SDL_PIXELFORMAT_XRGB8888;
```

I am thinking we need to install the package from the SDL github repo instead of using the package manager. IDK when the enumeration was added to SDL 2 as I am not able to look that far back on their git repo. It only goes to 2.23 I believe, not 2.0 that's for sure.

I don't see a newer version on the package index either, maybe I am missing it somewhere?

kdschlosser:
Update SDL2 has been updated but it is only available for Ubuntu 24.04, so I made a change to the workflow so it uses the latest Ubuntu version. Hopefully that will work and not break something else.

kdschlosser:
@kisvegabor 

It looks like the issue is now fixed. we can set the minimum SDL version requirement to 2.30.0. That is what is being used currently and it works properly. I will go digging into the SDL repo history for the file that enum is in and see if I am able to locate a minimum version number for SDL2 and add that to the docs. 

That's if this is what you wanted me to do with with the color format changes.

kdschlosser:
SDL 2.0.14 is the minimum that is needed to compile properly.

Ubuntu 20.04: 2.0.10
Ubuntu 22.04: 2.0.20
Ubuntu 23.10: 2.28.3
Ubuntu 24.04: 2.30.0

So Ubuntu 22.04 or newer is required if installing using the package manager.

.github/workflows/build_micropython.yml
```diff
@@ -8,7 +8,7 @@ jobs:
   build:
     if: ${{ github.event_name != 'pull_request' || github.repository != github.event.pull_request.head.repo.full_name }}
     name: Build ${{ matrix.port }} port
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-latest
```
PGNetHun: Sorry, but using "latest" from anything comes with the risk that in the future other 3rd party libraries will not be compatible with "latest" version... so please use ALWAYS specific version, which is 100% compatible with other libraries we use. 
I would say using ubuntu 22 is fine, or 24 if all the other libraries work with that, too (please test it).

PGNetHun (changes requested):

src/drivers/sdl/lv_sdl_window.c
```diff
@@ -364,17 +364,31 @@ static void texture_resize(lv_display_t * disp)
...
+            px_format = SDL_PIXELFORMAT_ARGB8888;
+            break;
+        case LV_COLOR_FORMAT_XRGB8888:
+            px_format = SDL_PIXELFORMAT_XRGB8888;
+            break;
+        case LV_COLOR_FORMAT_RGB565:
+            px_format = SDL_PIXELFORMAT_RGB565;
+            break;
+        default:
+            /*
```
kisvegabor: Please add an `LV_LOG_WARN()` here and `return;`

.github/workflows/build_micropython.yml
kisvegabor: `master` already uses `ubuntu-latest` (see https://github.com/lvgl/lvgl/blob/master/.github/workflows/ccpp.yml#L12) but specifying the version really makes sense.

cc @W-Mai

src/drivers/sdl/lv_sdl_window.c
kdschlosser: OK cool. I need to move the entire switch/case to the beginning of the function this way no work gets done before the warning. what about using LV_ASSERT_MSG? This way the error gets logged and it stops the program from running?

.github/workflows/build_micropython.yml
kdschlosser: I am in full agreement with setting the Ubuntu version. I did this just to check because I had found the version for SDL for Ubuntu 24.04 and while it wasn't 2.30.1 which is where I knew 100% that the enumerations were in I wanted to see if it would fix the problem. Once I knew the problem was fixed I then went digging to see what version of SDL had the change made in it and what would be the minimum version of Ubuntu would have the changes in it. Keep this as a reminder that that any changes made that use SDL are going to have to work with SDL version 2.0.20 which is ~2 years old which is pretty outdated. 

The other option is to compile SDL then we can use a newer version if wanted. IDK why SDL is not updated like it should be for the Ubuntu releases. They stop releasing the packages for newer versions of SDL for the different Ubuntu releases.


As it sits right now we know that everything works with Ubuntu 24.04 without a doubt. 24.04 has not been released as an officially yet but it is smart to run tests for it because it is going to be coming up. So maybe modifying the builds to run tests for 22.04 and also 24.04.

I am not sure what to do about the SDL issue and then having such an old release to the Ubuntu package index. unfortunately they do not build a Debian package when they do their releases.

kdschlosser:
I just noticed there is all kinds of things that needs to be changed in the SDL driver. It is not setting the buffer size correctly when the driver is created, it is using some arbitrary size of `32 * 1024` instead of using the passed in width and height. No color format is getting set so the default value is getting used. That's not so much of an issue unless the user wants to set it. If the user sets the color format after the driver has been created they are going to have to perform a window resize in order for the buffers to be resized properly. This is not so much of an issue if they change from a 32 bit color format to say the 16 bit one but it will definitely cause an issue if they change from the 16 bit color format to a 32 bit one. There is going to be frame buffer overruns. 

lv_observer is able to be used to handle monitoring for a color format change correct? If it can be when is the observer callback made during the call to lv_task_handler? It would need to be at the very beginning before anything gets written to the frame buffer.

When the display is resized the buffer does get sized properly, it is only when it is created that it is not sized properly. The `texture_resize` function needs to also get called when a change gets made to the color format.

kdschlosser:
question, car you realloc an array that has not been initialized?

example...
```
void *some_array;
realloc(some_array, 100);
```

will that work?. It looks like that is what is being done in the code. Then after the realloc is occurring on the frame buffer the buffer is being set to `0x00` using `memset` and then it is getting set to `0xFF` using `lv_memset`.

.github/workflows/build_micropython.yml
kdschlosser: @kisvegabor 

what was the solution for this change request?

kisvegabor: I've opened https://github.com/lvgl/lvgl/pull/6032 to see how Ubuntu 24.04 behaves.

kdschlosser: ok once that gets merged in I will update this PR and correct any merge conflicts.

kisvegabor: GitHub's CI don't want to queue some of the actions. I don't know if it's just a temporal issue, or related to ubuntu 24.04 somehow.

kdschlosser:
@kisvegabor 

Question for ya. when using XRGB which is basically the same thing as ARGB except the alpha channel is unused. What does LVGL fill that byte of data with? The pixel size for both color formats should still be 32 bit which means the frame buffer should be pixel values of `uint32_t`. I would image the `X` byte would be filled with `255` but maybe not.

SDL doesn't have support for XRGB below 5.0.20 but it does support ARGB so if the extra byte of data is populated with `255` I can make the changes needed to use ARGB when XRGB is what is selected in LVGL. SDL 2.0.10 is available for Ubuntu 20.04.x

kisvegabor:
It should be set to 255, however `SDL_pixels.h` I have this:
image

I've used `SDL_PIXELFORMAT_RGB888` for XRGB8888 so far and it worked well.

For RGB888 was used `SDL_PIXELFORMAT_BGR24` (as you can see in the code). I've tested these and they worked well.

kdschlosser:
It does but that is in version 2.0.20 and above. otherwise it is 

`
SDL_PIXELFORMAT_RGB888 =
        SDL_DEFINE_PIXELFORMAT(SDL_PIXELTYPE_PACKED32, SDL_PACKEDORDER_XRGB,
                               SDL_PACKEDLAYOUT_8888, 24, 4),

`

That is basically saying the buffer that is being passed in is going to be 4 bytes per pixel and only 3 of the bytes are to be used.

This is what I have for what color formats should point to what color formats

```
LV_COLOR_FORMAT_L8 = ???
LV_COLOR_FORMAT_I1 = SDL_PIXELFORMAT_INDEX1MSB
LV_COLOR_FORMAT_I2 = ???
LV_COLOR_FORMAT_I4 = SDL_PIXELFORMAT_INDEX4MSB

LV_COLOR_FORMAT_I8 = SDL_PIXELFORMAT_INDEX8
LV_COLOR_FORMAT_A8 = ???
LV_COLOR_FORMAT_RGB565 = SDL_PIXELFORMAT_RGB565
LV_COLOR_FORMAT_ARGB8565 = ???
LV_COLOR_FORMAT_RGB565A8 = ???
LV_COLOR_FORMAT_RGB888 = SDL_PIXELFORMAT_RGB24
LV_COLOR_FORMAT_ARGB8888 = SDL_PIXELFORMAT_ARGB8888
LV_COLOR_FORMAT_XRGB8888 = SDL_PIXELFORMAT_RGB888 / SDL_PIXELFORMAT_XRGB8888

LV_COLOR_FORMAT_I420 = SDL_PIXELFORMAT_IYUV  // Alias for I420

// Not sure on these next 3
LV_COLOR_FORMAT_I422 = SDL_DEFINE_PIXELFOURCC('I', '4', '2', '2')  YU16 ??  SDL_PIXELFORMAT_UYVY
LV_COLOR_FORMAT_I444 = SDL_DEFINE_PIXELFOURCC('I', '4', '4', '4')  YU24 ??
LV_COLOR_FORMAT_I400 = SDL_DEFINE_PIXELFOURCC('I', '4', '0', '0')
LV_COLOR_FORMAT_NV21 = SDL_PIXELFORMAT_NV21
LV_COLOR_FORMAT_NV12 = SDL_PIXELFORMAT_NV12
LV_COLOR_FORMAT_YUY2 = SDL_PIXELFORMAT_YUY2
LV_COLOR_FORMAT_UYVY = SDL_PIXELFORMAT_UYVY
```

FOURCC seems to to be the standard when it comes to pixel formats and I know that LVGL is written for memory use and speed but I wonder is possibly using the FOURCC naming convention and values would be the thing to use. I am not sure how many places in LVGL store the pixel format (color format). it being passed to a function shouldn't be that big of a deal because that is memory allocated onto the stack and it would get released once the function has finished correct? Right now LVGL uses a uint8_t to store the pixel format in lv_display_t.

Here are the common FOURCC values used in libyuv

https://github.com/lemenkov/libyuv/blob/main/include/libyuv/video_common.h


and here is another more complete list of FOURCC codes

https://github.com/torvalds/linux/blob/master/include/uapi/drm/drm_fourcc.h

kdschlosser:
@kisvegabor 

I made some changes. Have a look see and let me know what you think.

I added a `TODO` to fix the include for SDL_image. That will get fixed in a separate PR but I am unsure of how to go about doing it. 

The actual header file names stay the same doesn't matter the platform but the directory those header files are located in is platform specific. So how would I go about making an include if a directory gets set as a macro?

I am giving up with trying to fix the code formatting. IDK why it wants to change the indent level like it does. I think there is something wrong there.

kisvegabor:
The PR looks good, and I've fixed the code formatting too. You need to use astyle v3.4.12 (same as the CI). 

> I added a TODO to fix the include for SDL_image. That will get fixed in a separate PR but I am unsure of how to go about doing it.

`SDL_image` should be removed as it is, and LVGL's image decoder should be used instead. See [here](https://github.com/lvgl/lvgl/issues/6016).

kdschlosser:
>SDL_image should be removed as it is, and LVGL's image decoder should be used instead. See https://github.com/lvgl/lvgl/issues/6016.

This should be done in a separate PR.

If this one looks good we can merge it if you want.

Also, for future knowledge. compiling SDL2 is actually a breeze to do. they did a really good job with the build system and making it simple. It does have a lot of dependencies that have to be installed prior to compiling. So if you wanted to support a higher minimum version adding SDL to the CI to compile I can do easily.

kisvegabor:
> Also, for future knowledge. compiling SDL2 is actually a breeze to do. 

Good to know!

kisvegabor (approved):

kdschlosser:
I have been doing some work with SDL specifically the events and there is a way to get the events to populate without having to poll. The reason why I do not like the polling is because the events get queued. each time the application has to make a loop to collect an event. who known how long it will take for the application loop to come full circle to collect an event. Having the events push into the application using callbacks I believe is a better way to go about it.


I have also gotten double buffering to work using a thread to handle updating SDL with the buffer data. In order to get the events to work that gets run in a second thread. The trick to making it work is being able to "schedule" or push the callback so it happens in the application thread where LVGL is running. with MicroPython that's pretty easily done. I think we can use lv_timer to accomplish the same task as well. This would only take place when LVGL comes full circle. IDK if there is a way to force LVGL to read the indev data at the time the callback gets called.

kisvegabor:
@kdschlosser 
Could you send a separate PR about these?

kdschlosser:
I did figure out how to push an indev update but it is not able to be done from a thread. The question is if I am able to create timer objects from a thread? They would need to be only used a single time so no memory leaks can occur from that. I also don't know if there is a limit to how many timer objects are able to be queued and is the order in which they are run the same order in which they get added to the queue. The other question is if I call `lv_refr_now()` does the update take place diring that function call or does calling that function expire the refresh timer?

The objective is to get the events when they occur and update things as they happen instead of missing events, or processing events after the event has actually taken place. The current design uses the timers build into LVGL to check for events and then the events get iterated over until there are none left, so you end up with events that get skipped and if the loop is long you can end up with a lag in things like mouse movements.

Those things above are things I do not know about as far as the events go. I have a system that is working in the micropython binding that I have been working on that does work but not the most ideal way of doing it. I have reached out to the folks at MicroPython to see if I am able to do what I want to do and I have been trying some thing to get it running that way but without much luck. The way I have it working is better than what the current design is in terms of performance but it is not to the best it can be.

If I wrote it to keep everything self contained in LVGL it would be easier to make everything work because callbacks and things like that would be done to C functions and not Python functions. That makes a large difference in how it would operate.

The SDL port is also for devices that are able to run some form of an operating system like Linux, Windows or macOS so there isn't the issue of memory constraints, processor use constraints, or the ability to run multiple threads. so using a little bit more memory for storing things like callbacks is not that big of a deal to do...

user data is able to be passed to the callback function for the timer which is a good thing to be able to do.

OH I forgot to mention that I also added support for joysticks and controllers.

One of the things I have been meaning to ask is how this code works for the flushing callback when using SDL to handle the rendering?

This is what the code is that gets compiled with `LV_USE_DRAW_SDL` set to `1`
```c
static void flush_cb(lv_display_t * disp, const lv_area_t * area, uint8_t * px_map)
{
    /*IMPORTANT! It must be called to tell the system the flush is ready*/
    lv_display_flush_ready(disp);
}
```

kisvegabor:
> The question is if I am able to create timer objects from a thread?

Yes, assuming `lv_timer_handler` protected by mutex, but sure they don't mess up eachother.

>  I also don't know if there is a limit to how many timer objects are able to be queued and is the order in which they are run the same order in which they get added to the queue.

Only the memory is the limit. The call order is the necessarily the same as the creation order.  

> The other question is if I call lv_refr_now() does the update take place diring that function call or does calling that function expire the refresh timer?

During the call.

> This is what the code is that gets compiled with LV_USE_DRAW_SDL set to 1

For flushing I think  you need `flush_wait_cb`. See [here](https://docs.lvgl.io/master/porting/display.html#flush-wait-callback).

---

I was thinking about adding `lv_lock` and `lv_unlock` inside `lv_timer_handler()`. They would lock/unlock a mutex and the user could call `lv_lock/unlock` `anywhere they want too. Of course, they can work only if an `LV_USE_OS` is enabled, however we can add built-in support for SDL's OS layer too.

PR #5884, opened 2024-03-15
refactor(font): refactor font glyph data acquire method

W-Mai:

Refactor font glyph data acquire method

### Notes

### DIFF
--- a/src/draw/lv_draw_label.c
+++ b/src/draw/lv_draw_label.c
@@ -421,7 +421,7 @@ static void draw_letter(lv_draw_unit_t * draw_unit, lv_draw_glyph_dsc_t * dsc,  
             }
         }
 
-        dsc->glyph_data = (void *)lv_font_get_glyph_bitmap(&g, letter, draw_buf);
+        dsc->glyph_data = (void *) lv_font_get_glyph_bitmap(&g, draw_buf);
         dsc->format = dsc->glyph_data ? g.format : LV_FONT_GLYPH_FORMAT_NONE;
     }
     else {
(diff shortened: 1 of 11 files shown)


kisvegabor:
Without seeing clear benefits, I think it's the same case that I have described [here](https://github.com/lvgl/lvgl/pull/5888#issuecomment-2002007406).

W-Mai:
> Without seeing clear benefits, I think it's the same case that I have described [here](https://github.com/lvgl/lvgl/pull/5888#issuecomment-2002007406).

This modification will not affect the generation of font code. It will not cause code blocking elsewhere.

The main improvement point of this patch is that the `glyph index` needed to call the relevant API again. In fact, when the first time to `get glyph dsc`, the `glyph index` has been obtained, so the next step is to `get bitmap`. At that time, it can use the previously obtained `glyph index`, which reduces the acquisition of `glyph index` and reduces the consumption of time. So there is a gain.

At the same time, the callback function of the previous `get bitmap` essentially does not need the unicode param. This change will also optimize the code to make the code clearer.

kisvegabor:
Ah, I see it now. Very nice! :slightly_smiling_face: 
Have you measured the performance change?

kisvegabor (approved):

W-Mai:
> Ah, I see it now. Very nice! 🙂 Have you measured the performance change?

Aha, not yet. I want to do it, but I don't have much time to deal with it. Theoretically, time is optimized. Maybe we can ask the community for help to test this performance problem, and you can refer to what I did in this issue. #5901 

I think we need to compare the performance differences of default fonts, tinyttf, image fonts, and freetype fonts.

FASTSHIFT (approved):

Issue #5899, opened 2024-03-18
Text processing functions and API


sparkles43:
### Introduce the problem

Hi,

I am currently trying to do some UTF-8 text processing in my application. Due to the fact that I only require a minimal set of functions to process UTF-8 character I don't want to pull in another library but rather use the text processing capabilities lvgl offers on its own.

I have found the file [lv_text.h](https://github.com/lvgl/lvgl/blob/master/src/misc/lv_text.h#L259) (and [lv_txt.h](https://github.com/lvgl/lvgl/blob/release/v8.3/src/misc/lv_txt.h#L187) respectively) to contain global function pointers to useful text processing functions. Unfortunately all function pointers are labeled as private (prefixed with `_`) but are widely used in lvgl.

Is there any obstacle to making these function pointers part of the official API by removing the prefix? (I could imagine that it could increase maintenance efforts :thinking: )

Best regards,
sparkles

### Proposal

- Make function pointers part of the official API by removing the prefix `_` (for both lvgl v8 and lvgl v9)

kisvegabor:
The original purpose of adding the `_` was to indicate that a function is not part of the public API and to lower the size of the MicroPython binding. In v9 we started to establish the patter on having private headers instead (e.g. `lv_display_private.h`).

I think we can slowly transition to having private header files instead of `_` functions. 

cc @liamHowatt Could you open a PR for adding  `lv_text_private.h`? Note that `lv_text_private.h` shouldn't be included in `lvgl.h`, but only in the C files where needed.

sparkles43:
> The original purpose of adding the _ was to indicate that a function is not part of the public API ...

I guess my understanding was correct then.

> I think we can slowly transition to having private header files instead of _ functions.

Personally, I think private header files are a good way of preventing (more or less) accidental use of non-public APIs. :)

Regarding the original idea, can you give some feedback on making text processing part of the public API? I think it would be reasonable to handle strings in my UI application the same way lvgl does, i.e. have a common understanding of the text representations.

My initial proposal (to just drop the `_` prefix) is not a very good idea. The files should obviously be refactored to expose functions.

Anyways, I wonder why the text processing functions were exposed as extern function pointers. Was that to reduce calling overheads or was it rather used as a convenient way to switch between ASCII functions and UTF-8 functions in the lv_txt.c?

PR #5913, opened 2024-03-19
chore(text): move "_" funcs to private h

liamHowatt:

In response to #5899.

Part of the gradual transition to having private header files instead of `_` functions.

- First commit:
  - Move private `_` text functions to private header.
  - Include the private header in source files that depend on the private functions.
  - Make `extern` text function pointers `const`.
- Second commit (do we want this, @kisvegabor?):
  - Bulk rename private functions. Remove the leading `_`s.

### Notes


XuNeo:
Let's add the old APIs started with `_` to api_map.h too in case if anyone actually has used them.

liamHowatt:
I had to include the private header in the api map, which is included by lvgl.h so the private header is ultimately included everywhere lvgl.h is but at least the intent is clear. 🙂️

XuNeo:
Emmm, I think what @sparkles43 wants is to make `lv_text_` public, we may add those directly to `lv_txt.h`?

kisvegabor:
I think it's important to have only the really essential API function included from `lvgl.h`. LVGL's API is already very large.

Therefore I suggest this:
- Bulk rename `_lv_text` to `lv_text`
- Do not add API mapping as the `_lv` function weren't part of the public API so far either
- If anyone needs to use the `lv_text_*` functions they can include `lv_text_private.h` in their code.

What do you think?

XuNeo:
I think functions in `private.h` should never be used in user application, and that's most developer would/should follow.

For the lv_text case, the static `lv_text_utf8_` APIs can be moved to a public place like utils/lv_utf8.h and make it public since it doesn't depend on other modules. And there's a real need for it now.

sparkles43:
> I think functions in private.h should never be used in user application, and that's most developer would/should follow.

I agree with @XuNeo. This is basically why I opened my issue about lv_text.

> - Bulk rename _lv_text to lv_text
> - Do not add API mapping as the _lv function weren't part of the public API so far either

I think the remaining points - @kisvegabor made - make good sense.

> For the lv_text case, the static lv_text_utf8_ APIs can be moved to a public place like utils/lv_utf8.h and make it public since it doesn't depend on other modules. And there's a real need for it now.

- Until now I thought of the internal switching as an advantage. But of course the user could do checks on his/her own to guarantee UTF-8 is used.
- If we moved the UTF8 implementation to a separate file `utils/lv_utf8.h` that would make the private text processing `lv_text.h` depend on a public API which would make changes to the private API portion harder (e.g. format change away from `uint32_t` for UTF-8 characters would require an adapter in the `lv_text.c`)
  - I guess we would move the complexity of the API to a lower level in that case `utils/lv_utf8.h`, `utils/lv_ascii.h`

I am not strictly against the proposal but in this case we should take a look at which variant implies what maintenance efforts. Additionally we should think about the likelihood of changes to the API 🤔

kisvegabor:
Keeping the API smaller is very important. Ideally only the real core features (widgets, events, animation, styles, etc) should be public.

> I think functions in private.h should never be used in user application, and that's most developer would/should follow.

Normally it really shouldn't be used by the users, however if they want to use it for any reason they can do that. in general `private.h` files can be used by the developers e.g. if they are adding a new widget, or something similar. So I would phrase it like "Private headers shall be used only by the advanced users for advanced things".

> I am not strictly against the proposal but in this case we should take a look at which variant implies what maintenance efforts. Additionally we should think about the likelihood of changes to the API 🤔

At some point probably we will rework the whole typesetting interface, so these functions will be completely rewritten too. In light of that the more we hide is better.

kisvegabor (dismissed):

XuNeo (dismissed):

liamHowatt:
Should new private headers go in here?
https://github.com/liamHowatt/lvgl/blob/6127a82713485a26f7fb68e0dfe2d3e9ac639c41/src/lvgl_private.h

kisvegabor:
> Should new private headers go in here?
 
Ah, yes! The purpose of `lv_private.h` is to have a simple way to easily include the whole LVGL API like this:
```c
#include "lvgl.h"
#include "lv_private.h"
```

liamHowatt:
FYI here are all the *_private.h headers I could find in `master`.

```
./src/libs/freetype/lv_freetype_private.h
./src/indev/lv_indev_private.h ✅️
./src/display/lv_display_private.h ✅️
./src/misc/cache/lv_cache_entry_private.h
./src/misc/cache/lv_cache_private.h
```

XuNeo (approved):

kisvegabor (approved):

Issue #5906, opened 2024-03-18
LVGL9: lv_scale_set_line_needle_value supports only 1 object


s-hadinger:
### LVGL version

v9.0.0

### What happened?

`lv_scale_set_line_needle_value()` works only if there is only 1 scale object. As soon as you update a line, it changes all lines of all scale object.

The root cause is that the line points are defined as `static lv_point_precise_t needle_line_points[2];`. Since line object don't keep a local copy of the points array, the points coordinates are shared with all scale objects.

https://github.com/lvgl/lvgl/blob/1cb3e219a971c6f80a692928c00d91c61cfb5f84/src/widgets/scale/lv_scale.c#L183

Currently I do a not-so-clean workaround:
- call `lv_scale_set_line_needle_value()`
- read back the points from the `line` object
- copy the points to a structure allocated per line object
- set back the line points with `lv_line_set_points()`

liamHowatt:
Thanks @s-hadinger

I was able to reproduce with a modified version of `lv_example_scale_3`.

I think the fix could be as simple as this. A needle line array can live in each `lv_scale_t`.

```diff
diff --git a/src/widgets/scale/lv_scale.c b/src/widgets/scale/lv_scale.c
index 1e8df7ade..c21fc50ff 100644
--- a/src/widgets/scale/lv_scale.c
+++ b/src/widgets/scale/lv_scale.c
@@ -180,7 +180,6 @@ void lv_scale_set_line_needle_value(lv_obj_t * obj, lv_obj_t * needle_line, int3
     int32_t scale_width, scale_height;
     int32_t actual_needle_length;
     int32_t needle_length_x, needle_length_y;
-    static lv_point_precise_t needle_line_points[2];
 
     LV_ASSERT_OBJ(obj, MY_CLASS);
     lv_scale_t * scale = (lv_scale_t *)obj;
@@ -224,12 +223,12 @@ void lv_scale_set_line_needle_value(lv_obj_t * obj, lv_obj_t * needle_line, int3
     needle_length_x = (actual_needle_length * lv_trigo_cos(scale->rotation + angle)) >> LV_TRIGO_SHIFT;
     needle_length_y = (actual_needle_length * lv_trigo_sin(scale->rotation + angle)) >> LV_TRIGO_SHIFT;
 
-    needle_line_points[0].x = scale_width / 2;
-    needle_line_points[0].y = scale_height / 2;
-    needle_line_points[1].x = scale_width / 2 + needle_length_x;
[...]

PR #5937, opened 2024-03-21
feat(scale): mutiple line needles

liamHowatt:

Related issue: #5906

#### Feature

- More than one scale can have a line needle at the same time
- One scale can have multiple line needles

Scale image needles could do this already.

Doing this for line needles is nontrivial because line points are not stored in the line object. They need to be stored somewhere else. Until now, the line needle points array was `static`.

#### Changes

- Add a method to add line needles to the scale. A struct containing the line and line points gets added to the end of a linked list in scale, similar to how scale stores its scale sections.
- Change the line needle value setter to accept a line needle instead of a line. The passed line needle must come from the above mentioned line needle adder.
- Add a getter to get the line needle's underlying line object.
- Update `lv_example_scale_3`.

### Notes

### DIFF
--- a/examples/widgets/scale/lv_example_scale_6.c
+++ b/examples/widgets/scale/lv_example_scale_6.c
@@ -34,7 +34,6 @@ static void timer_cb(lv_timer_t * timer)
      */
 
     /* the scale will store the minute hand line points in `minute_hand_points` */
-    lv_line_set_points_mutable(minute_hand, minute_hand_points, 2); /* only needs to be set once */
     lv_scale_set_line_needle_value(scale, minute_hand, 60, minute);
     /* log the points that were stored in the array */
     LV_LOG_USER(
@@ -106,6 +105,7 @@ void lv_example_scale_6(void)
     lv_scale_set_rotation(scale, 270);
 
     minute_hand = lv_line_create(scale);
+    lv_line_set_points_mutable(minute_hand, minute_hand_points, 2);
 
     lv_obj_set_style_line_width(minute_hand, 3, 0);
     lv_obj_set_style_line_rounded(minute_hand, true, 0);
--- a/examples/widgets/scale/lv_example_scale_6.c
+++ b/examples/widgets/scale/lv_example_scale_6.c
@@ -1,0 +1,127 @@
+#include "../../lv_examples.h"
+#if LV_USE_SCALE && LV_BUILD_EXAMPLES
+
+#if LV_USE_FLOAT
+#define my_PRIprecise "f"
+#else
+#define my_PRIprecise LV_PRId32
+#endif
+
+static lv_obj_t * scale;
+static lv_obj_t * minute_hand;
+static lv_obj_t * hour_hand;
+static lv_point_precise_t minute_hand_points[2];
+static int32_t hour;
+static int32_t minute;
+
+static void timer_cb(lv_timer_t * timer)
+{
+    LV_UNUSED(timer);
+
+    minute++;
+    if(minute > 59) {
+        minute = 0;
+        hour++;
+        if(hour > 11) {
+            hour = 0;
+        }
+    }
+
+    /**
+     * the scale will store the needle line points in the existing
+     * point array if one was set with `lv_line_set_points_mutable`.
+     * Otherwise, it will allocate the needle line points.
+     */
+
+    /* the scale will store the minute hand line points in `minute_hand_points` */
+    lv_line_set_points_mutable(minute_hand, minute_hand_points, 2); /* only needs to be set once */
+    lv_scale_set_line_needle_value(scale, minute_hand, 60, minute);
+    /* log the points that were stored in the array */
+    LV_LOG_USER(
+        "minute hand points - "
+        "0: (%" my_PRIprecise ", %" my_PRIprecise "), "
+        "1: (%" my_PRIprecise ", %" my_PRIprecise ")",
+        minute_hand_points[0].x, minute_hand_points[0].y,
+        minute_hand_points[1].x, minute_hand_points[1].y
+    );
+
+    /* the scale will allocate the hour hand line points */
+    lv_scale_set_line_needle_value(scale, hour_hand, 40, hour * 5 + (minute / 12));
+}
+
+/**
+ * A round scale with multiple needles, resembing a clock
+ */
+void lv_example_scale_6(void)
+{
+    scale = lv_scale_create(lv_screen_active());
+
+    lv_obj_set_size(scale, 150, 150);
+    lv_scale_set_mode(scale, LV_SCALE_MODE_ROUND_INNER);
+    lv_obj_set_style_bg_opa(scale, LV_OPA_60, 0);
+    lv_obj_set_style_bg_color(scale, lv_color_black(), 0);
+    lv_obj_set_style_radius(scale, LV_RADIUS_CIRCLE, 0);
+    lv_obj_set_style_clip_corner(scale, true, 0);
+    lv_obj_center(scale);
+
+    lv_scale_set_label_show(scale, true);
+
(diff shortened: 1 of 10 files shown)


kisvegabor:
The original implementation needs to improved for sure, but I'm little bit concerned if we really need an other type here. 

Can we just say that it's the user's responsibility to set  point array for the line before using it in `lv_scale_set_line_needle_value`? If the points are not set we can even dynamically allocate it and add an event to line which frees the points on `DELETE`.

What do you think?

s-hadinger:
@kisvegabor I would prefer like you mention above: let the caller create the lv_line object and allocate the points.

liamHowatt:
What are both of your thoughts on what I pushed?

We can do the caller provided points but I could not think of a clear way to document the API.
```c
/**
 * Point the needle to the corresponding value through the line
 * @param obj              pointer to a scale object
 * @param needle_line      needle_line of the scale
 * @param needle_line_points user provided uninitialized 2-array for the line points -- must not go out of scope
 * @param needle_length    length of the needle
 *                         needle_length>0 needle_length=needle_length;
 *                         needle_length<0 needle_length=radius-|needle_length|;
 * @param value            needle to point to the corresponding value
 */
void lv_scale_set_line_needle_value(lv_obj_t * obj, lv_obj_t * needle_line, lv_point_precise_t * needle_line_points,
                                    int32_t needle_length, int32_t value);
```

Is this what you had in mind?

kisvegabor:
As `lv_obj_t * needle_line` is a real line widget, it could be used like this:
```c
static lv_point_precise_t points[2];
lv_obj_t * line = lv_line_create(scale);
lv_line_set_points(line, points, 2);

lv_scale_set_line_needle_value(scale, line, length, value);
```

liamHowatt:
Thanks for clarifying. I feel silly needing it spelled out but what would happen inside `lv_scale_set_line_needle_value`? Would it call a new getter on the `line` to get the pointer out of the `line` or would it access the `line`'s field directly? I toyed with that earlier and the problem I ran into was that the `line` stores the array as a `const` pointer. Since the assumption is valid in this special case, can we cast away the `const`?

kisvegabor:
Ah, that's not silly. Actually it's hard to resolve. We cannot remove the `const` from `lv_line_t` to allow using real const array. However converting the to non-const is not safe.

In lack of a better idea, I vote for the non safe way. If the array is `const` it will cause an immediate easy to debug crash. But we need to highlight the use of non-const array in the docs, in the example and the API comments too.

liamHowatt:
I added an experimental line points mutability API. It's the maximum amount of verbosity. Let's agree how much to cut out.

In total:
- `lv_line` points mutable setter, const points getter, fallible mutable points getter, is_mutable getter, point count getter.
- `lv_scale` checks if line has mutable 2-array and uses that. Otherwise allocates one which lasts for the lifetime of the line.
- New `lv_example_scale_6` which uses 2 line needles and demonstrates the line point options.
- New tests for `lv_line` and `lv_scale`.

image

examples/widgets/scale/lv_example_scale_6.c
```diff
@@ -0,0 +1,126 @@
...
+    }
+
+    /**
+     * the scale will store the needle line points in the existing
+     * point array if one was set with `lv_line_set_points_mutable`.
+     * Otherwise, it will allocate the needle line points.
+     */
+
+    /* the scale will store the minute hand line points in `minute_hand_points` */
+    lv_line_set_points_mutable(minute_hand, minute_hand_points, 2); /* only needs to be set once */
```
kisvegabor: Why is it called in the timer if needs to be called only once?

src/widgets/line/lv_line.h
```diff
@@ -27,9 +27,13 @@ extern "C" {
 /*Data of line*/
 typedef struct {
     lv_obj_t obj;
-    const lv_point_precise_t * point_array;     /**< Pointer to an array with the points of the line*/
+    union {
+        const lv_point_precise_t * constant;
+        lv_point_precise_t * mutable;
```
kisvegabor: Note this error.

src/widgets/line/lv_line.h
```diff
@@ -70,6 +82,35 @@ void lv_line_set_y_invert(lv_obj_t * obj, bool en);
...
+ * @return              number of points in array of points
+ */
+uint32_t lv_line_get_point_count(lv_obj_t * obj);
+
+/**
+ * Get the mutability of the stored point array pointer.
+ * @param obj           pointer to a line object
+ * @return              true: the point array pointer is mutable, false: constant
+ */
+bool lv_line_get_point_array_is_mutable(lv_obj_t * obj);
```
kisvegabor: What about this?
```suggestion
bool lv_line_is_point_array_mutable(lv_obj_t * obj);
```

Having both `get` and `is` a little bit stage to me. I vote for keeping `is` as there is no direct `set` counterpart.

src/widgets/line/lv_line.c
```diff
@@ -91,6 +89,42 @@ void lv_line_set_y_invert(lv_obj_t * obj, bool en)
...
+    return line->point_array_is_mutable;
+}
+
+lv_point_precise_t * lv_line_get_points_mutable(lv_obj_t * obj)
+{
+    LV_ASSERT_OBJ(obj, MY_CLASS);
+
+    lv_line_t * line = (lv_line_t *)obj;
+    if(!line->point_array_is_mutable) {
+        LV_LOG_ERROR("the line point array is not mutable");
```
kisvegabor: It should only a warning. `ERROR`s are rather used on critical problems or crashes.

kisvegabor (commented):
I like the idea of `lv_line_set_points_mutable`!

My only concern is that usually we do it the other was around. E.g.
```c
lv_label_set_text(label, buf); //buf is not const
lv_label_set_text_static(label, "hello"); //the text is constant
```

But we can keep your suggested API for backward compatibility.

examples/widgets/scale/lv_example_scale_6.c
liamHowatt: I wanted related functionality to be nearby so they can see all the steps in one place but yes I think it's a better demonstration to call it once outside the timer 👍️

liamHowatt:
> I like the idea of `lv_line_set_points_mutable`!
> 
> My only concern is that usually we do it the other was around. E.g.
> 
> ```c
> lv_label_set_text(label, buf); //buf is not const
> lv_label_set_text_static(label, "hello"); //the text is constant
> ```
> 
> But we can keep your suggested API for backward compatibility.

Using `_static` naming for consistency would be good but I think this is a bit different than the `set_text` cases. Technically `set_points` and `set_points_mutable` are both static setters since they don't make a copy. They technically should be called `set_points_static` and `set_points_static_mutable`.

Does that give you any ideas?

kisvegabor:
> Does that give you any ideas?

Yes, let's leave it as it is :slightly_smiling_face:

kisvegabor (approved):
LGTM!

xinsuinizhuan:
> I added an experimental line points mutability API. It's the maximum amount of verbosity. Let's agree how much to cut out.
> 
> In total:
> 
>     * `lv_line` points mutable setter, const points getter, fallible mutable points getter, is_mutable getter, point count getter.
> 
>     * `lv_scale` checks if line has mutable 2-array and uses that. Otherwise allocates one which lasts for the lifetime of the line.
> 
>     * New `lv_example_scale_6` which uses 2 line needles and demonstrates the line point storage options.
> 
>     * New tests for `lv_line` and `lv_scale`.
> 
> 
> image

where is the example? I need it.

liamHowatt:
(see my comment in the issue https://github.com/lvgl/lvgl/issues/5972#issuecomment-2025289110)

FASTSHIFT (approved):

PR #5939, opened 2024-03-21
fix(demos): update README to match lv_demos.c

marex:
Use lv_demos_show_help() both in lv_demos.c and demos/README, lv_demos_usage() does not exist.

### DIFF
--- a/demos/README.md
+++ b/demos/README.md
@@ -59,7 +59,7 @@
 
 ## Configure Demos Entry
 
-"demos/lv_demos.c" provides `lv_demos_create` and `lv_demos_usage` to simplify the creation of demos.
+"demos/lv_demos.c" provides `lv_demos_create` and `lv_demos_show_help` to simplify the creation of demos.
 
 If you build your main program named `lv_demos`, then you can run the widgets demo by running `lv_demos widgets` and the benchmark demo by running `lv_demos benchmark 1`.
 
@@ -93,7 +93,7 @@ int main(int argc, char ** argv)
   }
 
   if (!lv_demos_create(&argv[1], argc - 1)) {
-    lv_demos_usage();
+    lv_demos_show_help();
     goto demo_end;
   }
 


PGNetHun (approved):

liamHowatt (approved):

XuNeo (approved):

marex:
Why is the commit in lvgl master branch different from the one which has been submitted as part of this MR ?

This MR commit https://github.com/lvgl/lvgl/pull/5939/commits/95c7fa197c6998c4e7072b07e6c8826cd5036485 has proper commit message , the commit which is in master branch has had the commit message removed https://github.com/lvgl/lvgl/commit/9dec22dfaa01fe65904c65ece711e350624182a4 ... what is the purpose of removing commit messages during merge ?

kisvegabor:
The title (first line) of both commits are the same. Do you mean the lost extra description?
> Use lv_demos_show_help() both in lv_demos.c and demos/README,
lv_demos_usage() does not exist.

If so probably it was lost because we squash PRs into a single commit. Seemingly it creates a new commit even if there is only one commit in the PR. For this new commit we can set a commit message during merging. Usually it's the title of the PR. It's possible to add a longer description too, but it's not added by GitHub automatically, so we needed to handpick/write something there manually.

I see that having the info in the commit message would be useful, but at least the PR is linked there.

XuNeo:
We can avoid such information lost by changing merge strategy to `rebase and merge`  only. It requires committer squash relevant commits manually.

kisvegabor:
Making the commiter squash the commits might slow down to process, so we can decide it case by case.

Anyway, it's good to keep in mind to save the extra commit messages too.

marex:
> The title (first line) of both commits are the same. Do you mean the lost extra description?

Yes, that's called the commit message.

> > Use lv_demos_show_help() both in lv_demos.c and demos/README,
> > lv_demos_usage() does not exist.
> 
> If so probably it was lost because we squash PRs into a single commit. Seemingly it creates a new commit even if there is only one commit in the PR. For this new commit we can set a commit message during merging. Usually it's the title of the PR. It's possible to add a longer description too, but it's not added by GitHub automatically, so we needed to handpick/write something there manually.

As @XuNeo wrote already, rebase with fast forward merge would help here as it would retain the commit messages, but that rebase would also rewrite the commit hashes (commit hashes in the upstream tree and the ones in author repository would not be identical). The best approach is to merge with a merge commit, because then you also preserve the commit hashes of the commits from author.

> I see that having the info in the commit message would be useful, but at least the PR is linked there.

That isn't particularly useful when running git log and examining commits while offline.

btw I generally follow https://www.kernel.org/doc/html/latest/process/submitting-patches.html#describe-your-changes when creating commits, with various project specific tweaks as needed.

kisvegabor:
> The best approach is to merge with a merge commit, because then you also preserve the commit hashes of the commits from author.

Probably it's only my ignorance, but why is it important to have same commit hash in the upstream branch after merge as the author has in his feature branch?

marex:
> > The best approach is to merge with a merge commit, because then you also preserve the commit hashes of the commits from author.
> 
> Probably it's only my ignorance, but why is it important to have same commit hash in the upstream branch after merge as the author has in his feature branch?

Because then I (as an author of a commit, which is the state of a git repository when I finished developing the part and when I asked you to merge it into your upstream tree) can attest that whatever landed in your upstream tree (as one ancestor of the merge commit, which itself later becomes an ancestor of further development work i.e. more commits, from other authors) is exactly what I developed (and assigned the hash to) and not something that has been somehow modified in transit (that does not necessarily mean malicious intent, it could be data corruption or something else).

See https://lwn.net/Articles/811068/ part "How Git works, simplified" , I didn't find any article which would be less ... verbose ... and elaborate on the merging strategies. Maybe https://lwn.net/Articles/210045/ part "How merging operates" is also a good read.

marex:
> Do you need to "attest" which commit was yours due to company policies, potential legal issues, or personal motivation?

Currently I do not have to do this for any commit, but I cannot predict the future.

kisvegabor:
I see, thank you! As this the first time this concerns was raised, let's see if it's an issue for others too. If so, we will figure out somethings.

PR #5947, opened 2024-03-24
feat(hover): add Hover support for pointer device.

turoksama:

Add Hover support for pointer device: applied to top 3 layers.

Detects pointer device position and applied 2 new events to objects once when condition is matched: 
LV_EVENT_HOVERED and LV_EVENT_HOVER_LEAVE.

Objects should be CLICKERABLE(has LV_OBJ_FLAG_CLICKABLE) by calling: lv_obj_add_flag(obj, LV_OBJ_FLAG_CLICKABLE);
We can set styles for LV_STATE_HOVERED or add events LV_EVENT_HOVERED/LV_EVENT_HOVER_LEAVE.

Simple test codes:

void hoverd_event_cb(lv_event_t * e)
{
	int id = (int)e->user_data;
	lv_log("Object ID hovered: %d", id);
}

void hover_test(void)
{
	lv_obj_t *label_hovered= lv_label_create(lv_screen_active());
	lv_obj_add_flag(label_hovered, LV_OBJ_FLAG_CLICKABLE);
	lv_obj_add_event_cb(label_hovered, hoverd_event_cb, LV_EVENT_HOVERED, label_hovered->id);
	lv_obj_set_style_text_color(label_hovered, lv_color_hex(0x5be1b6), LV_PART_MAIN|LV_STATE_HOVERED);
}

### Notes


src/indev/lv_indev.c
```diff
@@ -657,12 +657,33 @@ static void indev_pointer_proc(lv_indev_t * i, lv_indev_data_t * data)
...
+        for(uint8_t h = 0; h < 3; h++, last++) {
+            lv_obj_t * hv = lv_indev_search_obj(*l++, &data->point);
+            if(*last != hv) {
+                if(lv_obj_is_valid(hv)) lv_obj_send_event(hv, LV_EVENT_HOVERED, i);
+                if(lv_obj_is_valid(*last)) lv_obj_send_event(*last, LV_EVENT_HOVER_LEAVE, i);
+                *last = hv;
+            }
+        }
+    }
+#else
```
XuNeo:
```suggestion
    if((i->pointer.last_point.x != data->point.x || i->pointer.last_point.y != data->point.y)) {
        /*Move the cursor if set and moved*/
        if(i->cursor != NULL) lv_obj_set_pos(i->cursor, data->point.x, data->point.y);
#if LV_USE_HOVER
        /*Hover detection*/
        static lv_obj_t * hv_last[3] = { NULL };
        lv_obj_t * layer[3];   //except layer bottom
        layer[0] = lv_layer_sys();
        layer[1] = lv_layer_top();
        layer[2] = lv_screen_active();
        lv_obj_t * last[3] = hv_last, *l[3] = layer;
        for(uint8_t h = 0; h < 3; h++, last++) {
            lv_obj_t * hv = lv_indev_search_obj(l[h] /* Double check if this's you want */, &data->point);
            if(*last != hv) {
                if(lv_obj_is_valid(hv)) lv_obj_send_event(hv, LV_EVENT_HOVERED, i);
                if(lv_obj_is_valid(*last)) lv_obj_send_event(*last, LV_EVENT_HOVER_LEAVE, i);
                *last = hv;
            }
        }
#endif
    }
```
Remove below `#else` section.

For lvgl v9, we cannot use static variable. You can add them lv_global.

Storing obj's pointer to an array needs to deal with obj delete event to avoid wild pointer.

Inside the for loop, do you mean checking  `if(last[h] != hv) {` instead ?

turoksama: #else section is used because HOVER is checked in every movements that position changed of pointer, it may have performance lost even HOVER is not enabled, pointer device will check `if((i->pointer.last_point.x` != data->point.x || i->pointer.last_point.y != data->point.y)) ` if cursor is not set. I think it's reasonable that people who doesn't need HOVER support. 
Using `lv_obj_t ** last = hv_last, ** l = layer;`  is  to replacing last[n] code, not for any other reasons.
Will try to put  `static lv_obj_t * hv_last[3] = { NULL };`   them in lv_global as you mentioned. :)

src/indev/lv_indev.c
```diff
@@ -657,12 +657,32 @@ static void indev_pointer_proc(lv_indev_t * i, lv_indev_data_t * data)
...
+        layer[1] = lv_layer_top();
+        layer[2] = lv_screen_active();
+        lv_obj_t ** last = lv_global.hv_last;
+        for(uint8_t h = 0; h < 3; h++, last++) {
+            lv_obj_t * hv = lv_indev_search_obj(layer[h], &data->point);
+            if(*last != hv) {
+                if(lv_obj_is_valid(hv)) lv_obj_send_event(hv, LV_EVENT_HOVERED, i);
+                if(lv_obj_is_valid(*last)) lv_obj_send_event(*last, LV_EVENT_HOVER_LEAVE, i);
+                *last = hv;
+            }
```
XuNeo: This looks more readable to me.

You still have to check what would happen if the obj stored in global->hv_last is deleted somehow.

```suggestion
        for(uint8_t h = 0; h < 3; h++) {
            lv_obj_t * hv = lv_indev_search_obj(layer[h], &data->point);
            if(last[h] != hv) {
                if(lv_obj_is_valid(hv)) lv_obj_send_event(hv, LV_EVENT_HOVERED, i);
                if(lv_obj_is_valid(last[h])) lv_obj_send_event(last[h], LV_EVENT_HOVER_LEAVE, i);
                last[h] = hv;
            }
```

kisvegabor:
First of all thank you for adding this long waited feature!

A few comments:
- Rename `LV_EVENT_HOVERED` to `LV_EVENT_HOVER_OVER`?
- Why do we need `lv_global.hv_last[3]`? There can be only a single hovered widget which can be found the very same way as the pressed widget is found [here](https://github.com/lvgl/lvgl/blob/master/src/indev/lv_indev.c#L1115-L1117), however we need to search for it only when the indev is released.
- If you need to store the last hovered widget, it shall be done [here](https://github.com/lvgl/lvgl/blob/master/src/indev/lv_indev_private.h#L77-L87), so that each indev can have its own hovered widget.
- We need to send the hover events only when the hovered widget changes.
- So there shouldn't be too many events, therefore I think `LV_USE_HOVER` is not required.

turoksama:
Thanks for your comments, it sounds more reasonable.
Here I have some questions about the layer_top and layer_sys, they seems to be the 2 layers at the top of layer_active_screen,  should they be processed prior to the 3rd and acts in parallel?

> Why do we need lv_global.hv_last[3]? There can be only a single hovered widget which can be found the very same way as the pressed widget is found [here](https://github.com/lvgl/lvgl/blob/master/src/indev/lv_indev.c#L1115-L1117), however we need to search for it only when the indev is released.

kisvegabor:
> Here I have some questions about the layer_top and layer_sys, they seems to be the 2 layers at the top of layer_active_screen, should they be processed prior to the 3rd and acts in parallel?


This is the order of layers:
1. sys layer
2. top layer
3. screen
4. bottom layer

The input device shall search in this order. If it finds a widget it should stop there and consider it as a touched widget.  So if something is found on the top layer, no need to search on the screen.

turoksama:
> The input device shall search in this order. If it finds a widget it should stop there and consider it as a touched widget. So if something is found on the top layer, no need to search on the screen.

I am uncertain whether we possess any magical “**” operations for these screen objects within _lv_display_t. As a potential optimization, if such operations do not exist, might we consider reordering them to enhance the efficiency of internal searches?

change:
    /** Screens of the display*/
    lv_obj_t ** screens;    /**< Array of screen objects.*/
    lv_obj_t * act_scr;     /**< Currently active screen on this display*/
    lv_obj_t * prev_scr;    /**< Previous screen. Used during screen animations*/
    lv_obj_t * scr_to_load; /**< The screen prepared to load in lv_screen_load_anim*/
    lv_obj_t * bottom_layer;    /**< @see lv_display_get_layer_bottom*/
    lv_obj_t * top_layer;       /**< @see lv_display_get_layer_top*/
    lv_obj_t * sys_layer;       /**< @see lv_display_get_layer_sys*/

into:

    /** Screens of the display*/
    lv_obj_t ** screens;    /**< Array of screen objects.*/
    lv_obj_t * sys_layer;       /**< @see lv_display_get_layer_sys*/
    lv_obj_t * top_layer;       /**< @see lv_display_get_layer_top*/
    lv_obj_t * act_scr;     /**< Currently active screen on this display*/
    lv_obj_t * bottom_layer;    /**< @see lv_display_get_layer_bottom*/
    lv_obj_t * prev_scr;    /**< Previous screen. Used during screen animations*/
    lv_obj_t * scr_to_load; /**< The screen prepared to load in lv_screen_load_anim*/

How do you feel about this idea?

src/indev/lv_indev.c
```diff
@@ -1267,6 +1246,23 @@ static void indev_proc_press(lv_indev_t * indev)
  */
 static void indev_proc_release(lv_indev_t * indev)
 {
+    if((indev->pointer.last_point.x != indev->pointer.act_point.x ||
+        indev->pointer.last_point.y != indev->pointer.act_point.y)) {
+        lv_obj_t ** layer = &indev->disp->sys_layer;
+        lv_obj_t ** last = &indev->pointer.last_hovered;
+        lv_obj_t * hved = NULL;
```
kisvegabor: Avoid ambiguous abbreviations   
```suggestion
        lv_obj_t * hovered = NULL;
```

src/indev/lv_indev.c
```diff
@@ -1267,6 +1246,23 @@ static void indev_proc_press(lv_indev_t * indev)
...
+        uint8_t layers = 4;
+        while(!hved && layers) {
+            hved = lv_indev_search_obj(*layer++, &indev->pointer.act_point);
+            if(*last != hved) {
+                if(lv_obj_is_valid(hved)) lv_obj_send_event(hved, LV_EVENT_HOVER_OVER, indev_act);
+                if(lv_obj_is_valid(*last)) lv_obj_send_event(*last, LV_EVENT_HOVER_LEAVE, indev_act);
```
kisvegabor: `lv_obj_is_valid` shouldn't be used here becasue it's slow and sometimes gives incorrect result. Instead `indev->pointer.last_hovered` shall be `NULL`-ed explicitly [here](https://github.com/lvgl/lvgl/blob/master/src/indev/lv_indev.c#L1567-L1569) and [here](https://github.com/lvgl/lvgl/blob/master/src/indev/lv_indev.c#L1388).

src/indev/lv_indev.c
```diff
@@ -1267,6 +1246,23 @@ static void indev_proc_press(lv_indev_t * indev)
  */
 static void indev_proc_release(lv_indev_t * indev)
 {
+    if((indev->pointer.last_point.x != indev->pointer.act_point.x ||
+        indev->pointer.last_point.y != indev->pointer.act_point.y)) {
+        lv_obj_t ** layer = &indev->disp->sys_layer;
+        lv_obj_t ** last = &indev->pointer.last_hovered;
+        lv_obj_t * hved = NULL;
+        uint8_t layers = 4;
+        while(!hved && layers) {
```
kisvegabor: Relying on the order of field in a struct is not safe and can result in hard to debug issues. It seems [`pointer_search_obj`](https://github.com/lvgl/lvgl/blob/master/src/indev/lv_indev.c#L1388) can be used here.

kisvegabor (commented):
Looks good in general, thank you! Just a few comments.

src/indev/lv_indev.c
turoksama: Settled.
Set last hovered object to NULL when the object is deleted, instead of passive search.

src/indev/lv_indev.c
turoksama: It seems pointer_search_obj is better than relying on the order of field in a struct. Thanks for your advice.

kisvegabor (commented):
Amazing! It's so simple now :slightly_smiling_face: 

Could you add a few test too? Just create a button and maybe some other widgets, use the [test's mouse API](https://github.com/lvgl/lvgl/blob/master/tests/src/lv_test_indev.h#L14-L18) and see if the correct events are called as you adjust the mouse's X/Y coordinates. You can just increment counters in the events to see if they were fired. (More info about the tests [here](https://github.com/lvgl/lvgl/blob/master/tests/README.md))

kisvegabor:
I've just fixed minor bug related to `LV_OBJ_FLAG_PRESS_LOCK` and added indev events. I hope you don mind that I pushed into your branch.

turoksama:
That's cool. Thanks for your commits.
I'll test it later.

src/indev/lv_indev.c
```diff
@@ -1247,6 +1259,25 @@ static void indev_proc_press(lv_indev_t * indev)
  */
 static void indev_proc_release(lv_indev_t * indev)
 {
+    if(indev->wait_until_release || /*Hover the new widget even if the coordinates didn't changed*/
+       (indev->pointer.last_point.x != indev->pointer.act_point.x ||
+        indev->pointer.last_point.y != indev->pointer.act_point.y)) {
+        lv_obj_t ** last = &indev->pointer.last_hovered;
```
XuNeo, line 1270: Let's use `lv_obj_t * last = indev->pointer.last_hovered;`, and change below one to `indev->pointer.last_hovered = hovered;`

XuNeo (commented):
If the hovered obj is deleted, the stored pointer becomes wild pointer. So it also needs to check obj deletion and process accordingly.

https://github.com/lvgl/lvgl/blob/500e16e96883c8b7a8b010e9f54e8bcfb488c10f/src/core/lv_obj_tree.c#L527-L535

kisvegabor:
Hey @turoksama,

Will you have time to finalize this PR in the near future? We are so close, it would be a pity to let it go :slightly_smiling_face:

turoksama:
Hi there, @kisvegabor 
Add some test cases for hover, would you mind to test it? Thanks.

src/indev/lv_indev.c
turoksama: Well, set NULL in lv_obj_delete as @kisvegabor mentioned is prefer to checking valid which is more like a hacking one. What do you think?

kisvegabor:
I've just fix the crash in the test, but it still fails due to an assert. 

I've added a simple test for deleting the hovered widget too.

kisvegabor:
I think you forgot to push the update.

kisvegabor:
I use command line, but I think the Sync button means updating your PR from the target branch (LVGL master in this case) 

So you can just commit and push the fixed test and it will appear here.

turoksama:
Is it ready to merge?

kisvegabor (approved):
Looks good to me, thank you for all the effort!

kisvegabor:
@XuNeo @FASTSHIFT could review this PR?

FASTSHIFT (approved):

XuNeo (approved):

PR #5951, opened 2024-03-25
fix(draw): fix the default draw thread stack is too large

FASTSHIFT:

Fix the memory waste caused by the default draw thread stack being too large, and add a stack size check when some components that consume stack space are opened.

cc @GorgonMeducer 

### Notes

### DIFF
--- a/Kconfig
+++ b/Kconfig
@@ -183,6 +183,13 @@ menu "LVGL configuration"
 				it is buffered into a "simple" layer before rendering. The widget can be buffered in smaller chunks.
 				"Transformed layers" (if `transform_angle/zoom` are set) use larger buffers and can't be drawn in chunks.
 
+		config LV_DRAW_THREAD_STACK_SIZE
+			int "Stack size of draw thread in bytes"
+			default 8192
+			depends on LV_USE_OS > 0
+			help
+				FreeType req. stack size >= 24KB.
+
 		config LV_USE_DRAW_SW
 			bool "Enable software rendering"
 			default y
@@ -197,11 +204,6 @@ menu "LVGL configuration"
 				> 1 requires an operating system enabled in `LV_USE_OS`
 				> 1 means multiply threads will render the screen in parallel
 
-		config LV_DRAW_THREAD_STACKSIZE
-			int "stack size of draw thread in byte"
-			default 32768
-			depends on LV_USE_OS > 0
-			
 		config LV_USE_DRAW_ARM2D_SYNC
 			bool "Enable Arm's 2D image processing library (Arm-2D) for all Cortex-M processors"
 			default n
--- a/Kconfig
+++ b/Kconfig
@@ -188,7 +188,7 @@ menu "LVGL configuration"
 			default 8192
 			depends on LV_USE_OS > 0
 			help
-				FreeType req. stack size >= 24KB.
+				If FreeType or ThorVG is enabled, it is recommended to set it to 32KB or more.
 
 		config LV_USE_DRAW_SW
 			bool "Enable software rendering"
--- a/lv_conf_template.h
+++ b/lv_conf_template.h
@@ -113,6 +113,11 @@
 /*The target buffer size for simple layer chunks.*/
 #define LV_DRAW_LAYER_SIMPLE_BUF_SIZE    (24 * 1024)   /*[bytes]*/
 
+/* The stack size of the drawing thread.
+ * NOTE: FreeType req. stack size >= 24KB.
+ */
+#define LV_DRAW_THREAD_STACK_SIZE    (8 * 1024)   /*[bytes]*/
+
 #define LV_USE_DRAW_SW 1
 #if LV_USE_DRAW_SW == 1
     /* Set the number of draw unit.
@@ -149,9 +154,6 @@
         #define  LV_DRAW_SW_ASM_CUSTOM_INCLUDE ""
     #endif
 #endif
-
-/* Set stack size of drawing thread. Unit is byte. If Thorvg is enabled, 128kB is required tested on simulator.*/
-#define LV_DRAW_THREAD_STACKSIZE 32768
 
 /* Use NXP's VG-Lite GPU on iMX RTxxx platforms. */
 #define LV_USE_DRAW_VGLITE 0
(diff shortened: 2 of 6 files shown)


W-Mai (approved):
Thank you for adding freetype stack limitations 😁

GorgonMeducer:
It looks good to me. Maybe we can add check for other components too.

XuNeo (dismissed):

kisvegabor:
Do you have any info about the stack usage of ThorVG?

FASTSHIFT:
> Do you have any info about the stack usage of ThorVG?

I haven't tested ThorVG on embedded devices yet. Have you tested it before? How many stacks were used?

kisvegabor:
I've used it for Lottie animations. I'll check the stack usage today.

kisvegabor:
I've tested with rendering the vector demo in 100x100 and 200x200 resolution. The extra stack usage was the same in both cases: +20kB. So let's say minimum 24kB is needed, but 32kB is recommended. 

In `lv_draw_sw.c` we can also check the set stack size based and if it's to small based on the settings an `#error` or `#warning` can be triggered.

On side note, in case of multi core rendering, we could dedicate a core to FreeType and/or ThorVG, so that we don't need to allocate that large stack multiple times.

FASTSHIFT:
> I've tested with rendering the vector demo in 100x100 and 200x200 resolution. The extra stack usage was the same in both cases: +20kB. So let's say minimum 24kB is needed, but 32kB is recommended.
> 
> In `lv_draw_sw.c` we can also check the set stack size based and if it's to small based on the settings an `#error` or `#warning` can be triggered.
> 
> On side note, in case of multi core rendering, we could dedicate a core to FreeType and/or ThorVG, so that we don't need to allocate that large stack multiple times.

Done，The ThorVG check has been added.

scripts/lv_conf_internal_gen.py
```diff
@@ -197,6 +197,12 @@
     #define LV_USE_THORVG  (LV_USE_THORVG_INTERNAL || LV_USE_THORVG_EXTERNAL)
 #endif
 
+#if LV_USE_OS
+    #if (LV_USE_FREETYPE || LV_USE_THORVG) && LV_DRAW_THREAD_STACK_SIZE < (24 * 1024)
+        #warning "LV_DRAW_THREAD_STACK_SIZE is too small for FreeType or ThorVG. It recommended to set it to 32KB or more."
```
XuNeo: typo: It **'s** recommended to ... But maybe better for this:

```suggestion
        #warning "Increase LV_DRAW_THREAD_STACK_SIZE to at least 32KB for FreeType or ThorVG."
```

FASTSHIFT: Done

XuNeo (dismissed):

Kconfig
```diff
@@ -183,6 +183,13 @@ menu "LVGL configuration"
 				it is buffered into a "simple" layer before rendering. The widget can be buffered in smaller chunks.
 				"Transformed layers" (if `transform_angle/zoom` are set) use larger buffers and can't be drawn in chunks.
 
+		config LV_DRAW_THREAD_STACK_SIZE
+			int "Stack size of draw thread in bytes"
+			default 8192
+			depends on LV_USE_OS > 0
```
kisvegabor, line 189: It also depends on `LV_USE_DRAW_SW` as its the stack size of the SW render task.

lv_conf_template.h
```diff
@@ -113,6 +113,11 @@
 /*The target buffer size for simple layer chunks.*/
 #define LV_DRAW_LAYER_SIMPLE_BUF_SIZE    (24 * 1024)   /*[bytes]*/
 
+/* The stack size of the drawing thread.
+ * NOTE: If FreeType or ThorVG is enabled, it is recommended to set it to 32KB or more.
+ */
+#define LV_DRAW_THREAD_STACK_SIZE    (8 * 1024)   /*[bytes]*/
```
kisvegabor, line 119: It's related only to the SW render.  It'd be better to rename it to `LV_DRAW_SW_THREAD_STACK_SIZE` and add a compatibility define in `lv_conf_internal.h`:
```c
#if defined(LV_DRAW_THREAD_STACKSIZE) && !defined(LV_DRAW_SW_THREAD_STACK_SIZE)
  #warning "LV_DRAW_THREAD_STACKSIZE was renamed to LV_DRAW_SW_THREAD_STACK_SIZE. Please update lv_conf.h or run menuconfig again."
  #define LV_DRAW_SW_THREAD_STACK_SIZE LV_DRAW_THREAD_STACKSIZE
#endif
```

Kconfig
FASTSHIFT: `LV_DRAW_THREAD_STACK_SIZE` should work as a general configuration, not just for software rendering.

lv_conf_template.h
FASTSHIFT: Done

Kconfig
kisvegabor: It's not necessarily true that all renderes need 8k RAM by default. A simpler one optimized of for size with less features might use only 2kB, but a heavier one might need much more.  

E.g. [PXP](https://github.com/lvgl/lvgl/blob/master/src/draw/nxp/pxp/lv_draw_pxp.c#L95) has a hard coded 2kB stack size. So `LV_DRAW_THREAD_STACK_SIZE` is not really a general configuration.

XuNeo (approved):

Kconfig
kisvegabor: Let's come to a conclusion on this before merging.

FASTSHIFT: I think it should be done the other way around. Most draw units should use `LV_DRAW_THREAD_STACK_SIZE` as stack size configuration instead of hard-coding. If the draw unit does not need to draw freetype fonts, it can be hard-coded in its own code.

kisvegabor: So there are 2 cases:
1. The GPU is limited and doesn't really depend on `lv_conf.h` :arrow_right: the stack size can be hardcoded (e.g. PXP)
2. The GPU is more powerful and can use FreeType or ThorVG if enabled :arrow_right:  it should use `LV_DRAW_THREAD_STACK_SIZE`.

If we approach it like this, I agree with current version. :+1:

kisvegabor (approved):

PR #5961, opened 2024-03-26
fix(animimg): fix repeat_count type mismatch

jiangxiaobaiyuenyxx:
fix(anim): fix lv_anim_set_repeat_count configuration parameter may be truncated


FASTSHIFT (approved):

terry0012 (approved):

W-Mai (approved):

XuNeo (approved):

kisvegabor:
Now we mostly have `uint32_t` and `int32_t` in the APIs.
It's not critical, but in light of that changing `count` to `uint32_t` in `lv_anim_set/get_repeat_count` would have been a better approach to have more unified types in the API.

PR #5964, opened 2024-03-26
feat(ci): Add CI test for UEFI build

MiSimon:

A simple makefile that builds LVGL without the standard library for UEFI.
Only the "core" source files are built, without any third party libraries.

### Notes
- RFC
- see #5767 for more information


MiSimon:
I wanted to extend the dummy test app to a minimal working UEFI application but therefore i need to define CONFIG_LV_COLOR_DEPTH=32 and this enables _LV_KCONFIG_PRESENT. 
```
#ifdef CONFIG_LV_COLOR_DEPTH
    #define _LV_KCONFIG_PRESENT
#endif
```

With _LV_KCONFIG_PRESENT all widgets are disabled per default.
```
#ifndef LV_USE_BUTTON
    #ifdef _LV_KCONFIG_PRESENT
        #ifdef CONFIG_LV_USE_BUTTON
            #define LV_USE_BUTTON CONFIG_LV_USE_BUTTON
        #else
            #define LV_USE_BUTTON 0
        #endif
    #else
        #define LV_USE_BUTTON        1
    #endif
#endif
```

Is there a way to work around this, without enabeling all widgets per `#define`?

kisvegabor:
Thank you for opening this PR!

Can you use `-DLV_COLOR_DEPTH` instead of `-DCONFIG_LV_COLOR_DEPTH`? (same for the other configs)

tests/makefile_uefi/Makefile
```diff
@@ -0,0 +1,52 @@
+LVGL_PATH := $(CURDIR)/../..
+ARCH = x86_64
+CC = clang
+$(info LVGL_PATH is set to: $(LVGL_PATH))
+
+CSRCS  = $(shell find $(LVGL_PATH)/src/core -type f -name '*.c')
```
kisvegabor: Can you use [`lvgl.mk`](https://github.com/lvgl/lvgl/blob/master/lvgl.mk) instead?

If not, what about using `find` in `src` directly? It would make maintenance simpler when new subfolders are added.

MiSimon: I had to change some other files for this. If lvgl.mk is used all none lvgl includes have to happen after the feature defines.

src/libs/tjpgd/tjpgd.h
```diff
@@ -9,6 +9,9 @@ extern "C" {
 #endif
 
 #include "tjpgdcnf.h"
+
+#if LV_USE_TJPGD
```
kisvegabor, line 14: `lv_conf_internal.h` needs to included.

MiSimon: Already changed that

kisvegabor:
Thank you! I've added a CI file and it compiles succesfully: :tada: 
image

cc @MouriNaruto

kisvegabor (dismissed):

MiSimon:
> Thank you! I've added a CI file and it compiles succesfully: 🎉

I think the CI is using the wrong makefile, it uses tests/makefile and not test/makefile_uefi

kisvegabor:
Ahh, you are right. I've just fixed it and it's still working :slightly_smiling_face:

tests/makefile_uefi/Makefile
```diff
@@ -0,0 +1,38 @@
...
+LDFLAGS = -target $(ARCH)-unknown-windows -nostdlib -Wl,-entry:efi_main -Wl,-subsystem:efi_application -fuse-ld=lld-link
+
+COBJS := $(patsubst $(LVGL_PATH)%,objs/%, $(CSRCS:.c=.o))
+
+test_file: $(COBJS)
+	$(CC) -o $@ $^ $(LDFLAGS)
+
+objs/%.o: $(LVGL_PATH)%.c
+	mkdir -p $(dir $@)
+	$(CC) $(CFLAGS) -o $@ -c $^
```
lhdjply, line 34: The compilation process is too long and not nice to look at. I suggest changing it to the following.
```suggestion
	@$(CC) $(CFLAGS) -o $@ -c $^
	@echo cc $^
```

MiSimon: Most of the time I look at the build output I do this to find a problem, therefore I am happy about every piece of information I can find there. 
But yes, your suggested change looks much cleaner. I took the other test makefile as template, I can change it but then I suggest to also change the other makefile(s?).

kisvegabor: I also vote for the current more verbose output, especially as we use it the CI into which we are looking only if there are errors.

lhdjply: Okay, I understand

examples/others/imgfont/lv_example_imgfont_1.c
```diff
@@ -1,8 +1,8 @@
 #include "../../lv_examples.h"
-#include <stdio.h>
 
 #if LV_BUILD_EXAMPLES
 #if LV_USE_IMGFONT
+#include <stdio.h>
```
XuNeo: Looks like we can remove it directly.

examples/widgets/list/lv_example_list_2.c
```diff
@@ -1,7 +1,6 @@
-#include <stdlib.h>
-
 #include "../../lv_examples.h"
 #if LV_USE_LIST && LV_BUILD_EXAMPLES
+#include <stdlib.h>
```
XuNeo: Let's remove it and use `lv_rand` instead.

```c
// rand() % cnt
lv_rand(0, cnt);
```

src/libs/qrcode/qrcodegen.h
```diff
@@ -317,3 +320,5 @@ int qrcodegen_getMinFitVersion(enum qrcodegen_Ecc ecl, size_t dataLen);
 #ifdef __cplusplus
 }
 #endif
+
+#endif
```
XuNeo: Please add new line at file end.
```bash
error: no newline at end of file [-Werror,-Wnewline-eof]
#endif```

File: src/libs/qrcode/qrcodegen.c [hunk 7]
```diff
@@ -1111,3 +1113,4 @@ int qrcodegen_version2size(int version)
 
     return ((version - 1) * 4 + 21);
 }
+#endif
```
XuNeo: ditto.

src/libs/lodepng/lodepng.h
```diff
@@ -30,10 +30,9 @@ freely, subject to the following restrictions:
 extern "C" {
 #endif
 
-#include <string.h> /*for size_t*/
-
 #include "../../../lvgl.h"
 #if LV_USE_LODEPNG
+#include <string.h> /*for size_t*/
```
XuNeo: size_t is defined in `stddef.h`, let's correct it.

You may also need to fix one warning in lv_lodepng.c, to use lv_memcmp instead of memcmp.

src/libs/tjpgd/tjpgd.c
```diff
@@ -1135,3 +1136,4 @@ JRESULT jd_decomp(
 
     return rc;
 }
+#endif
```
XuNeo: Please add new line at file end.

tests/makefile_uefi/Makefile
```diff
@@ -0,0 +1,38 @@
...
+
+clean:
+	rm -r $(CURDIR)/objs
+	rm test_file
```
XuNeo: ditto.

tests/makefile_uefi/efi.h
```diff
@@ -0,0 +1,160 @@
...
+_Static_assert(sizeof(int32_t)  == 4, "Size check for 'int32_t' failed.");
+_Static_assert(sizeof(uint32_t) == 4, "Size check for 'uint32_t' failed.");
+_Static_assert(sizeof(uint_fast32_t) == 4, "Size check for 'uint_fast32_t' failed.");
+_Static_assert(sizeof(int64_t)  == 8, "Size check for 'int64_t' failed.");
+_Static_assert(sizeof(uint64_t) == 8, "Size check for 'uint64_t' failed.");
+_Static_assert(sizeof(intptr_t) == sizeof(void *), "Size check for 'intptr_t' failed.");
+_Static_assert(sizeof(ptrdiff_t) == sizeof(void *), "Size check for 'ptrdiff_t' failed.");
+_Static_assert(sizeof(uintptr_t) == sizeof(void *), "Size check for 'uintptr_t' failed.");
+
+#endif
```
XuNeo: ditto.

src/libs/lodepng/lodepng.h
kisvegabor: We should rather use `LV_STDDEF_INCLUDE`.

examples/widgets/list/lv_example_list_2.c
MiSimon: Done

src/libs/qrcode/qrcodegen.h
MiSimon: Done

src/libs/qrcode/qrcodegen.c
MiSimon: Done

src/libs/tjpgd/tjpgd.c
MiSimon: Done

tests/makefile_uefi/Makefile
MiSimon: Done

examples/others/imgfont/lv_example_imgfont_1.c
MiSimon: Done

src/libs/lodepng/lodepng.h
MiSimon: Done

src/libs/qrcode/qrcodegen.h
MiSimon: Done, [ac21d65](https://github.com/lvgl/lvgl/pull/5964/commits/ac21d65814e15b40082cc42b8302c09a29518623) looks like it is removed but i cloned it and it is still there

src/libs/qrcode/qrcodegen.c
MiSimon: Done

tests/makefile_uefi/Makefile
MiSimon: Done

tests/makefile_uefi/efi.h
MiSimon: Done

src/libs/tjpgd/tjpgd.c
MiSimon: Done

kisvegabor (approved):

XuNeo (approved):

PR #5973, opened 2024-03-27
fix(display): update the color format of the draw buffers on color format change

### LINKED ISSUE #5889: LVGL9: lv_display_set_color_format bug
### LVGL version

~~v9.0.0~~
v9.1

### What happened?
@kisvegabor
If you create a simple bar and set the value, it will be rendered and flushed as expected, but when you change the value again ~~something bad happens and it gets corrupted/disappeared.~~ It gets asserted.

### How to reproduce?
```C
	lv_obj_t * bar = lv_bar_create(lv_screen_active());
	lv_obj_set_size(bar, 200, 20);
	lv_obj_center(bar);

	lv_bar_set_value(bar, 10, LV_ANIM_OFF);
	lv_refr_now(NULL);

	lv_bar_set_value(bar, 50, LV_ANIM_OFF);
	lv_refr_now(NULL);

	while(1){};

### PR DESCRIPTION
**Pull request by kisvegabor**

fixes #5889

@XuNeo 
We [reshape](https://github.com/lvgl/lvgl/blob/master/src/core/lv_refr.c#L578) the draw buffer before rendering, so I think the changes are ok, but please confirm it.

### Notes

### DIFF
--- a/src/display/lv_display.c
+++ b/src/display/lv_display.c
@@ -457,6 +457,8 @@ void lv_display_set_color_format(lv_display_t * disp, lv_color_format_t color_fo
 
     disp->color_format = color_format;
     disp->layer_head->color_format = color_format;
+    disp->buf_1->header.cf = color_format;
+    if(disp->buf_2) disp->buf_2->header.cf = color_format;
 
     lv_display_send_event(disp, LV_EVENT_COLOR_FORMAT_CHANGED, NULL);
 }
--- a/src/display/lv_display.c
+++ b/src/display/lv_display.c
@@ -457,7 +457,7 @@ void lv_display_set_color_format(lv_display_t * disp, lv_color_format_t color_fo
 
     disp->color_format = color_format;
     disp->layer_head->color_format = color_format;
-    disp->buf_1->header.cf = color_format;
+    if(disp->buf_1) disp->buf_1->header.cf = color_format;
     if(disp->buf_2) disp->buf_2->header.cf = color_format;
 
     lv_display_send_event(disp, LV_EVENT_COLOR_FORMAT_CHANGED, NULL);

### CODE REVIEW

**Comment by kisvegabor** [comment]
@kdschlosser With this change the C tests are passing but the MicroPython test fails. How can we check the reason of the failure?

**Review by XuNeo — COMMENTED** [review]
`lv_display_set_color_format` could be called before and after set the real `buf`.
So it's better to use
```c
if(disp->buf_1) disp->buf_1 = lv_draw_buf_reshape(disp->buf_1, color_format, hor_res, ver_res, LV_STRIDE_AUTO);
if(disp->buf_2) disp->buf_2 = lv_draw_buf_reshape(disp->buf_2, color_format, hor_res, ver_res, LV_STRIDE_AUTO);
```

We should also mention that do not mix use below two API sets
```c
/* Raw buffer manuplation */
lv_display_set_color_format();
lv_display_set_buffers();
```
```c
lv_display_set_draw_buffers()
```

XuNeo:
> buffers are always reshaped before rendering

True. You are correct, there's no need to call `lv_draw_buf_reshape` here.

kdschlosser:
This is the error. You just have to go and look at the CI for the micropython test and it will tell you why it is failing.

```python
  File "lib/lv_bindings/lvgl/tests/micropy_test/__init__.py", line 298, in test_1_image_compare
ValueError: not enough image data

During handling of the above exception, another exception occurred:

  File "lib/lv_bindings/lvgl/tests/micropy_test/__init__.py", line 318, in test_1_image_compare
AssertionError: Traceback (most recent call last):
  File "lib/lv_bindings/lvgl/tests/micropy_test/__init__.py", line 298, in test_1_image_compare
  File "/home/runner/.local/lib/python3.8/site-packages/PIL/Image.py", line 815, in frombytes
    raise ValueError(msg)
ValueError: not enough image data
```

It is telling us there is not enough image data. Meaning the color depth is somehow not right. The test script is collecting raw RGB data from the frame buffer so whatever change has been made is causing the wrong amount of data to be collected or the data that is being collected is not 3 byte RGB pixel data.

kdschlosser:
IDK where the artifacts are stored but in there should be a file with the raw pixel data that was collected so we can see what is going on.

kdschlosser:
It looks like you are trying to allow dynamic changing of the color format after the display driver has been initialized. Wouldn't the easiest way to be only to allow the color format to be set once and if the user wants to change the color format they would need to create a new display in LVGL in order to do that?

I say that because of buffer sizing. the frame buffers that have been passed in would need to be resized in order for this to work. In the case of someone that has passed in a raw buffer LVGL is not going to be able to handle that resizing and it would need to be done externally.

It could be explained but then how would the user go about changing the size of the buffer that is stored in LVGL? It would be easier to restrict a display's color format to only being set a single time.

kdschlosser:
Let me explain further on that.

When a display gets created in LVGL the user must tell the display what the frame buffers are. So lets use an RGB display as an example because it is easier to follow what I am saying.

Lets say the display is 480 x 320 and the color format for the display is RGB888. but for whatever reason the display when it was initialized is set to RGB565. That would mean the buffers are going to have a size of 480 x 320 x 2. Now if we allow dynamic changing of the color format and the user changes it to RGB888 guess what? You now have a buffer overrun condition because the size needed is going to be 480 x 320 x 3. If we flop flop it the other way around and the starting color format is RGB888 and we are changing it to RGB565 no you have  a wasted memory use scenario.

You also need to remember these are MCU's we are dealing with so creating a second set of buffers or even resizing the frame buffers might not even be possible due to memory constraints. having to add 153600 bytes to an already existing frame buffer while everything is up and running is going to be a hard thing to do because the entire frame buffer would need to occupy 460800 bytes of contiguous memory. Due to memory constraints the frame buffers are one of the things that get allocated first in order to secure that amount of contiguous memory because there might not be that must contiguous space available later on. This problem gets even more compounded when dealing with double buffering.

kisvegabor:
Artifacts are not created automatically, but required explicit steps in the CI. Like [this](https://github.com/lvgl/lvgl/blob/master/.github/workflows/ccpp.yml#L47-L52).

I think it was crashing because the buffers were `NULL` in `set_color_format` as you set the buffers only after that. See [here](https://github.com/lvgl/lvgl/blob/master/tests/micropy_test/micropy.py#L84-L88). I added a fix, let's see if it passes now.

In case of screen sized buffer, the user really might need to set new buffers, which is doable and a reasonable expectation from the user. It's better to keep this opportunity open then limiting to setting the color depth once.

kisvegabor:
And passed! Could you improve the log to indicate if there was a crash and where was it?

kdschlosser:
I will need to have the problem duplicated and debugging turned on in order to see why the error didn't surface like it should have.

XuNeo (approved):

FASTSHIFT (approved):

PR #5974, opened 2024-03-27
feat(draw_buff): add user custom draw buffer instance support

W-Mai:

Add user custom draw buffer instance support, and maintain the compatibility of the previous API.

Adding these APIs can separate fonts from normal picture draw buffs for management. Fonts use many small pieces of memory, which is easy to cause memory fragmentation.

New APIs here:
```cpp

void lv_draw_buf_init_with_default_handlers(lv_draw_buf_handlers_t * handlers);
void lv_draw_buf_init_handlers(lv_draw_buf_handlers_t * handlers,
                               lv_draw_buf_malloc_cb buf_malloc_cb,
                               lv_draw_buf_free_cb buf_free_cb,
                               lv_draw_buf_align_cb align_pointer_cb,
                               lv_draw_buf_invalidate_cache_cb invalidate_cache_cb,
                               lv_draw_buf_width_to_stride_cb width_to_stride_cb);

void * lv_draw_buf_align_user(const lv_draw_buf_handlers_t * handlers, void * buf, lv_color_format_t color_format);
void lv_draw_buff_invalidate_cache_user(const lv_draw_buf_handlers_t * handlers, const lv_draw_buf_t * draw_buf, const lv_area_t * area);
uint32_t lv_draw_buf_width_to_stride_user(const lv_draw_buf_handlers_t * handlers, uint32_t w, lv_color_format_t color_format);
lv_draw_buf_t * lv_draw_buf_create_user(const lv_draw_buf_handlers_t * handlers, uint32_t w, uint32_t h, lv_color_format_t cf, uint32_t stride);
void lv_draw_buf_destroy_user(const lv_draw_buf_handlers_t * handlers, lv_draw_buf_t * buf);
```

@FASTSHIFT @XuNeo 

### Notes

### DIFF
--- a/src/draw/lv_draw_buf.c
+++ b/src/draw/lv_draw_buf.c
@@ -14,7 +14,7 @@
 /*********************
  *      DEFINES
  *********************/
-#define handlers LV_GLOBAL_DEFAULT()->draw_buf_handlers
+#define default_handlers LV_GLOBAL_DEFAULT()->draw_buf_handlers
 
 /**********************
  *      TYPEDEFS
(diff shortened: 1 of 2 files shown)


XuNeo (approved):

FASTSHIFT (approved):

kisvegabor:
Do you agree that the `draw_buf` API can be considered as internal API? If so keep only the new variants with the `const lv_draw_buf_handlers_t * handlers` parameter, and remove the old ones. If the handler is `NULL` we can use the `default_handlers`.

W-Mai:
> Do you agree that the `draw_buf` API can be considered as internal API? If so keep only the new variants with the `const lv_draw_buf_handlers_t * handlers` parameter, and remove the old ones. If the handler is `NULL` we can use the `default_handlers`.

In general, `draw_buf` is indeed used internally, and almost no one uses this API in the user mode, but the vendor of `driver` modules similar to `decoder`s and `nuttx` will use the `draw_buf` related API, so the original API is retained. The API is recognized in order to be compatible with previous usage.

If only the API that need parameters is retained, most of the code using the default API will become complicated. I think.

kisvegabor:
Okay, let's keep it as it is. As part of #6011 we should add some hide these APIs as much as possible  and keep only the `typedef` for `lv_draw_buf_t`.

PR #5988, opened 2024-03-28
feat(drivers): account for the frame buffer virtual resolution deviating from the visible resolution.

pvginkel:
The Linux frame buffer flush_cb method now accounts for the virtual resolution deviating from the visible resolution. The offset into the paint buffer is calculated based on the visible resolution, and the offset into the frame buffer is calculated using the line_length, i.e. stride. The x offset and y offset are only used to calculate the position into the frame buffer, not the position into the paint buffer.

On Linux devices where the virtual resolution of the frame buffer deviate from the visible resolution, the flush_cb method would fail. In such a configuration, the line_length (stride) differs from the xres of the device. The line_length was used to calculate the offset into the paint buffer, causing issues.

I've also made a change to how the xoffset and yoffset are used. I've removed usage of these parameters to calculate the offset into the paint buffer. xoffset was used, and is not anymore. yoffset wasn't used in the first place. Since these parameters are offsets of the visible resolution into the virtual resolution, they couldn't have been needed for the paint buffer. That being said, I wasn't able to test this as I'm not sure how to configure them using fbset.

I've tested this with different resolutions, and both the direct and partial rendering modes.

### Notes

### DIFF
--- a/src/drivers/display/fb/lv_linux_fbdev.c
+++ b/src/drivers/display/fb/lv_linux_fbdev.c
@@ -312,8 +312,12 @@ static void flush_cb(lv_display_t * disp, const lv_area_t * area, uint8_t * colo
         return;
     }
 
-    uint32_t color_pos = (area->x1 + dsc->vinfo.xoffset) * px_size + area->y1 * dsc->finfo.line_length;
-    uint32_t fb_pos = color_pos + dsc->vinfo.yoffset * dsc->finfo.line_length;
+    uint32_t color_pos =
+        area->x1 * px_size +
+        area->y1 * dsc->vinfo.xres * px_size;
+    uint32_t fb_pos =
+        (area->x1 + dsc->vinfo.xoffset) * px_size +
+        (area->y1 + dsc->vinfo.yoffset) * dsc->finfo.line_length;
 
     uint8_t * fbp = (uint8_t *)dsc->fbp;
     int32_t y;
@@ -321,7 +325,7 @@ static void flush_cb(lv_display_t * disp, const lv_area_t * area, uint8_t * colo
         for(y = area->y1; y <= area->y2; y++) {
             lv_memcpy(&fbp[fb_pos], &color_p[color_pos], w * px_size);
             fb_pos += dsc->finfo.line_length;
-            color_pos += dsc->finfo.line_length;
+            color_pos += dsc->vinfo.xres * px_size;
         }
     }
     else {
--- a/src/drivers/display/fb/lv_linux_fbdev.c
+++ b/src/drivers/display/fb/lv_linux_fbdev.c
@@ -314,7 +314,7 @@ static void flush_cb(lv_display_t * disp, const lv_area_t * area, uint8_t * colo
 
     uint32_t color_pos =
         area->x1 * px_size +
-        area->y1 * dsc->vinfo.xres * px_size;
+        area->y1 * disp->hor_res * px_size;
     uint32_t fb_pos =
         (area->x1 + dsc->vinfo.xoffset) * px_size +
         (area->y1 + dsc->vinfo.yoffset) * dsc->finfo.line_length;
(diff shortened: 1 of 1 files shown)


kisvegabor (dismissed):
Thank you! 

I've tested it and I confirm that it's working well.

src/drivers/display/fb/lv_linux_fbdev.c
```diff
@@ -312,16 +312,20 @@ static void flush_cb(lv_display_t * disp, const lv_area_t * area, uint8_t * colo
...
+        (area->y1 + dsc->vinfo.yoffset) * dsc->finfo.line_length;
 
     uint8_t * fbp = (uint8_t *)dsc->fbp;
     int32_t y;
     if(LV_LINUX_FBDEV_RENDER_MODE == LV_DISPLAY_RENDER_MODE_DIRECT) {
         for(y = area->y1; y <= area->y2; y++) {
             lv_memcpy(&fbp[fb_pos], &color_p[color_pos], w * px_size);
             fb_pos += dsc->finfo.line_length;
-            color_pos += dsc->finfo.line_length;
+            color_pos += dsc->vinfo.xres * px_size;
```
Johennes: I'm coming here from #6002 where I tried to fix the same problem. Interestingly, this change here does not appear to resolve the issue I'm seeing. In my case `dsc->finfo.line_length` is equal to `dsc->vinfo.xres * px_size` so this doesn't make a difference. Using `disp->hor_res` instead of `dsc->vinfo.xres` fixes the rendering for me. @pvginkel would that also work in your case?

kisvegabor: It's great that the two discussion are joined and hopefully we will have a more robust solution in the end. :+1:

pvginkel: @Johennes I must assume that you've rotated the display then. I'm thinking that we replace `xres` with `hres` if rotation is `LV_DISPLAY_ROTATION_90` or `LV_DISPLAY_ROTATION_270`. Can you verify whether this assumption is correct, that the display rotation is set to one of these values?

Johennes: No, I'm actually not using rotation. I'm using `lv_display_set_physical_resolution` and `lv_display_set_resolution` to limit the LVGL display to a subsection of the framebuffer. That's why using the resolution from the framebuffer's `vinfo` breaks things for me.

pvginkel: ~~Would you be able to share the code? I must confess I have no idea how to determine the size of the paint buffer in your scenario.~~

I get it now, you had this in your original comment. Let me check if `disp->hor_res` works for me also.

pvginkel: Yes, this works for me also. I pushed a change. Can you test this in your app @Johennes?

Johennes: Yes, this works for me, too. Thanks a lot. :+1:

pvginkel:
I forgot to set the PR to draft. I've incorporated feedback from @Johennes. @PGNetHun the PR is ready for review now.

src/drivers/display/fb/lv_linux_fbdev.c
```diff
@@ -312,16 +312,20 @@ static void flush_cb(lv_display_t * disp, const lv_area_t * area, uint8_t * colo
         return;
     }
 
-    uint32_t color_pos = (area->x1 + dsc->vinfo.xoffset) * px_size + area->y1 * dsc->finfo.line_length;
-    uint32_t fb_pos = color_pos + dsc->vinfo.yoffset * dsc->finfo.line_length;
+    uint32_t color_pos =
+        area->x1 * px_size +
+        area->y1 * disp->hor_res * px_size;
```
kisvegabor: Please move it into the `if(LV_LINUX_FBDEV_RENDER_MODE == LV_DISPLAY_RENDER_MODE_DIRECT)` branch to indicate that it's used only there.

kisvegabor (commented):
Thank you, you just one minor note.

src/drivers/display/fb/lv_linux_fbdev.c
pvginkel: Done.

kisvegabor (approved):
Thank you!

FASTSHIFT (approved):

PR #5999, opened 2024-03-30
fix(example): LVGL_Arduino.ino millis() as tick source

KlausMu:

The example "LVGL_Arduino.ino" does not compile.
```
error: invalid conversion from 'long unsigned int (*)()' to 'lv_tick_get_cb_t' {aka 'unsigned int (*)()'} [-fpermissive]
     lv_tick_set_cb(millis);
                    ^~~~~~
```
Although data types are indeed the same (32 bit unsigned integer), it does not work this way.

Please see also https://forum.lvgl.io/t/updating-lvgl-library/15307

### DIFF
--- a/examples/arduino/LVGL_Arduino/LVGL_Arduino.ino
+++ b/examples/arduino/LVGL_Arduino/LVGL_Arduino.ino
@@ -67,6 +67,12 @@ void my_touchpad_read( lv_indev_t * indev, lv_indev_data_t * data )
      */
 }
 
+/*use Arduinos millis() as tick source*/
+uint32_t my_tick(void)
+{
+    return millis();
+}
+
 void setup()
 {
     String LVGL_Arduino = "Hello Arduino! ";
@@ -78,7 +84,7 @@ void setup()
     lv_init();
 
     /*Set a tick source so that LVGL will know how much time elapsed. */
-    lv_tick_set_cb(millis);
+    lv_tick_set_cb(my_tick);
 
     /* register print function for debugging */
 #if LV_USE_LOG != 0
--- a/examples/arduino/LVGL_Arduino/LVGL_Arduino.ino
+++ b/examples/arduino/LVGL_Arduino/LVGL_Arduino.ino
@@ -68,7 +68,7 @@ void my_touchpad_read( lv_indev_t * indev, lv_indev_data_t * data )
 }
 
 /*use Arduinos millis() as tick source*/
-uint32_t my_tick(void)
+static uint32_t my_tick(void)
 {
     return millis();
 }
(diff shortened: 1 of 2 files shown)


examples/arduino/LVGL_Arduino/LVGL_Arduino.ino
```diff
@@ -67,6 +67,12 @@ void my_touchpad_read( lv_indev_t * indev, lv_indev_data_t * data )
      */
 }
 
+/*use Arduinos millis() as tick source*/
+uint32_t my_tick(void)
+{
+    return millis();
```
W-Mai, line 73: LGTM
Should we add a force cast here for the return value of `millis()`?

C47D (approved):

examples/arduino/LVGL_Arduino/LVGL_Arduino.ino
KlausMu: You mean `return (uint32_t) millis();`?
I tested it with Arduino IDE and PlatformIO. There is no difference. It works with and without the type cast. In both cases, no compiler warnings.

kisvegabor (commented):
Please also update the docs here: https://docs.lvgl.io/master/porting/tick.html

Replace
> Arduino: lv_tick_set_cb(millis);

With
> Arduino: lv_tick_set_cb(my_tick_get_cb);, where my_tick_get_cb is:  `static uint32_t my_tick_get_cb(void) { return millis(); }

examples/arduino/LVGL_Arduino/LVGL_Arduino.ino
```diff
@@ -67,6 +67,12 @@ void my_touchpad_read( lv_indev_t * indev, lv_indev_data_t * data )
      */
 }
 
+/*use Arduinos millis() as tick source*/
+uint32_t my_tick(void)
```
kisvegabor:
```suggestion
static uint32_t my_tick(void)
```

KlausMu: sorry, I haven't seen that I simply could commit your suggestion. I added the suggestion in my branch.

KlausMu:
> Please also update the docs here: https://docs.lvgl.io/master/porting/tick.html
> 
> Replace
> 
> > Arduino: lv_tick_set_cb(millis);
> 
> With
> 
> > Arduino: lv_tick_set_cb(my_tick_get_cb);, where my_tick_get_cb is:  `static uint32_t my_tick_get_cb(void) { return millis(); }

Ok, updated the docs as well

kisvegabor (approved):
Thank you!

FASTSHIFT (approved):

PR #6001, opened 2024-03-30
feat(example): let "LVGL_Arduino.ino" easier to use

KlausMu:

While using the "LVGL_Arduino.ino" example, I found some minor improvements which could help people to get it running

- updated link to docs
- support to set the rotation
- omit warnings until first touch

### DIFF
--- a/examples/arduino/LVGL_Arduino/LVGL_Arduino.ino
+++ b/examples/arduino/LVGL_Arduino/LVGL_Arduino.ino
@@ -1,5 +1,5 @@
 /*Using LVGL with Arduino requires some extra steps:
- *Be sure to read the docs here: https://docs.lvgl.io/master/get-started/platforms/arduino.html  */
+ *Be sure to read the docs here: https://docs.lvgl.io/master/integration/framework/arduino.html  */
 
 #include <lvgl.h>
 
@@ -15,9 +15,10 @@
 //#include <examples/lv_examples.h>
 //#include <demos/lv_demos.h>
 
-/*Set to your screen resolution*/
+/*Set to your screen resolution and rotation*/
 #define TFT_HOR_RES   320
 #define TFT_VER_RES   240
+#define TFT_ROTATION  0
 
 /*LVGL draw into this buffer, 1/10 screen size usually works well. The size is in bytes*/
 #define DRAW_BUF_SIZE (TFT_HOR_RES * TFT_VER_RES / 10 * (LV_COLOR_DEPTH / 8))
@@ -58,6 +59,15 @@ void my_touchpad_read( lv_indev_t * indev, lv_indev_data_t * data )
 
     if(!touched) {
         data->state = LV_INDEV_STATE_RELEASED;
+        
+        // initially, data->point.x == hor. res and data->point.y == ver. res,
+        // which would lead to continous warnings from lv_indev.c until first touch
+        if((data->point.x < 0) || (data->point.x >= TFT_HOR_RES)) {
+            data->point.x = 0;
+        }
+        if((data->point.y < 0) || (data->point.y >= TFT_VER_RES)) {
+            data->point.y = 0;
+        }
     } else {
         data->state = LV_INDEV_STATE_PRESSED;
 
@@ -66,6 +76,11 @@ void my_touchpad_read( lv_indev_t * indev, lv_indev_data_t * data )
     }
      */
 }
+
+// would be better to have this in lv_tft_espi.h
+typedef struct {
+    TFT_eSPI * tft;
+} lv_tft_espi_t;
 
 void setup()
 {
@@ -89,6 +104,8 @@ void setup()
 #if LV_USE_TFT_ESPI
     /*TFT_eSPI can be enabled lv_conf.h to initialize the display in a simple way*/
     disp = lv_tft_espi_create(TFT_HOR_RES, TFT_VER_RES, draw_buf, sizeof(draw_buf));
+    ((lv_tft_espi_t *)(lv_display_get_driver_data(disp)))->tft->setRotation(TFT_ROTATION);
+
 #else
     /*Else create a display yourself*/
     disp = lv_display_create(TFT_HOR_RES, TFT_VER_RES);
--- a/examples/arduino/LVGL_Arduino/LVGL_Arduino.ino
+++ b/examples/arduino/LVGL_Arduino/LVGL_Arduino.ino
@@ -18,7 +18,7 @@
 /*Set to your screen resolution and rotation*/
 #define TFT_HOR_RES   320
 #define TFT_VER_RES   240
-#define TFT_ROTATION  0
+#define TFT_ROTATION  LV_DISPLAY_ROTATION_0
 
 /*LVGL draw into this buffer, 1/10 screen size usually works well. The size is in bytes*/
 #define DRAW_BUF_SIZE (TFT_HOR_RES * TFT_VER_RES / 10 * (LV_COLOR_DEPTH / 8))
@@ -68,11 +68,6 @@ void my_touchpad_read( lv_indev_t * indev, lv_indev_data_t * data )
      */
 }
 
-// would be better to have this in lv_tft_espi.h
-typedef struct {
-    TFT_eSPI * tft;
-} lv_tft_espi_t;
-
 void setup()
 {
     String LVGL_Arduino = "Hello Arduino! ";
@@ -95,7 +90,7 @@ void setup()
 #if LV_USE_TFT_ESPI
     /*TFT_eSPI can be enabled lv_conf.h to initialize the display in a simple way*/
     disp = lv_tft_espi_create(TFT_HOR_RES, TFT_VER_RES, draw_buf, sizeof(draw_buf));
-    ((lv_tft_espi_t *)(lv_display_get_driver_data(disp)))->tft->setRotation(TFT_ROTATION);
+    lv_display_set_rotation(disp, TFT_ROTATION);
 
 #else
     /*Else create a display yourself*/
--- a/examples/arduino/LVGL_Arduino/LVGL_Arduino.ino
+++ b/examples/arduino/LVGL_Arduino/LVGL_Arduino.ino
@@ -59,15 +59,6 @@ void my_touchpad_read( lv_indev_t * indev, lv_indev_data_t * data )
 
     if(!touched) {
         data->state = LV_INDEV_STATE_RELEASED;
-        
-        // initially, data->point.x == hor. res and data->point.y == ver. res,
-        // which would lead to continous warnings from lv_indev.c until first touch
-        if((data->point.x < 0) || (data->point.x >= TFT_HOR_RES)) {
-            data->point.x = 0;
-        }
-        if((data->point.y < 0) || (data->point.y >= TFT_VER_RES)) {
-            data->point.y = 0;
-        }
     } else {
         data->state = LV_INDEV_STATE_PRESSED;
 
--- a/src/drivers/display/tft_espi/lv_tft_espi.cpp
+++ b/src/drivers/display/tft_espi/lv_tft_espi.cpp
@@ -26,6 +26,7 @@ typedef struct {
  *  STATIC PROTOTYPES
  **********************/
 static void flush_cb(lv_display_t * disp, const lv_area_t * area, uint8_t * px_map);
+static void resolution_changed_event_cb(lv_event_t * e);
 
 /**********************
  *  STATIC VARIABLES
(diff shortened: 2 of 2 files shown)


examples/arduino/LVGL_Arduino/LVGL_Arduino.ino
```diff
@@ -58,6 +59,15 @@ void my_touchpad_read( lv_indev_t * indev, lv_indev_data_t * data )
 
     if(!touched) {
         data->state = LV_INDEV_STATE_RELEASED;
+        
+        // initially, data->point.x == hor. res and data->point.y == ver. res,
```
kisvegabor: It shouldn't be the case. Is it maybe an issue of the touch driver you are using?

examples/arduino/LVGL_Arduino/LVGL_Arduino.ino
```diff
@@ -89,6 +104,8 @@ void setup()
 #if LV_USE_TFT_ESPI
     /*TFT_eSPI can be enabled lv_conf.h to initialize the display in a simple way*/
     disp = lv_tft_espi_create(TFT_HOR_RES, TFT_VER_RES, draw_buf, sizeof(draw_buf));
+    ((lv_tft_espi_t *)(lv_display_get_driver_data(disp)))->tft->setRotation(TFT_ROTATION);
```
kisvegabor: We can improve the TFT_eSPI driver by internally adding a display event handler for `LV_EVENT_RESOLUTION_CHANGE` . In that event we can call `setRotation` using the current rotation set by `lv_display_set_rotation()` (We can use `lv_display_get_rotation` to get the current rotation). 

See [this](https://github.com/lvgl/lvgl/blob/master/src/drivers/display/lcd/lv_lcd_generic_mipi.c#L289-L312) as a reference.

examples/arduino/LVGL_Arduino/LVGL_Arduino.ino
```diff
@@ -58,6 +59,15 @@ void my_touchpad_read( lv_indev_t * indev, lv_indev_data_t * data )
...
         data->state = LV_INDEV_STATE_RELEASED;
+        
+        // initially, data->point.x == hor. res and data->point.y == ver. res,
+        // which would lead to continous warnings from lv_indev.c until first touch
+        if((data->point.x < 0) || (data->point.x >= TFT_HOR_RES)) {
+            data->point.x = 0;
+        }
+        if((data->point.y < 0) || (data->point.y >= TFT_VER_RES)) {
+            data->point.y = 0;
+        }
```
liamHowatt: I think the `initially, data->point.x == hor. res` is fairly hardware-specific. This detail adds a lot of info for a beginner to parse. Also, if a warning is emitted it's not a bad thing. The user can program around it case-by-case or even ignore the warning with no consequences. Remove this if you agree.

examples/arduino/LVGL_Arduino/LVGL_Arduino.ino
```diff
@@ -67,6 +77,11 @@ void my_touchpad_read( lv_indev_t * indev, lv_indev_data_t * data )
      */
 }
 
+// would be better to have this in lv_tft_espi.h
+typedef struct {
+    TFT_eSPI * tft;
+} lv_tft_espi_t;
+
```
liamHowatt: As you say, it would be good to have that typedef in the header, however it depends on an optional library type which won't always be available when building for non-tft so it needs to stay in the c file.

Regarding this re-declaration, it's better not to duplicate declarations and instead use the API to access what you need. Maybe in a personal application you could do this, but not in an example in the repo. Remove this if you agree.

examples/arduino/LVGL_Arduino/LVGL_Arduino.ino
liamHowatt: Did this work for you when you tested with different values of `TFT_ROTATION`? Changing the rotation would swap the hor+ver resolution, wouldn't it? I think it would be better to rotate the display using `lv_display_set_rotation` so the user doesn't have to manually swap the hor+ver resolution when they change the rotation.

```suggestion
    /*TFT_eSPI can be enabled lv_conf.h to initialize the display in a simple way*/
    disp = lv_tft_espi_create(TFT_HOR_RES, TFT_VER_RES, draw_buf, sizeof(draw_buf));
    lv_display_set_rotation(disp, TFT_ROTATION);
```

Also update `#define TFT_ROTATION 0` to `#define TFT_ROTATION LV_DISPLAY_ROTATION_0`

liamHowatt (commented):
Thanks @KlausMu, display rotation would be good to have as part of the example.

examples/arduino/LVGL_Arduino/LVGL_Arduino.ino
KlausMu: Well, I do not fully understand what `lv_display_set_rotation()` is really doing. I none of my tests it actually rotated what I can see on the TFT.
The only way I could get rotated content on the tft, is by setting rotation in `TFT_eSPI`. But the rotation of `TFT_eSPI` is hardcoded in "lv_tft_espi.cpp" to `dsc->tft->setRotation(3)`, which does not work in my case. So I looked for a way to set it.
Maybe we need a way to provide the needed rotation to function `lv_tft_espi_create()` as a parameter.

For my display to work, I have to set
```
#define TFT_HOR_RES   240
#define TFT_VER_RES   320

 ((lv_tft_espi_t *)(lv_display_get_driver_data(disp)))->tft->setRotation(0);
```

KlausMu: Hm, I'm not sure if I can set the rotation of TFT_eSPI by using calls to functions of lvgl. Maybe I misunderstand the concept.
Please see my comment above to what @liamHowatt wrote. I think we need a way to set the rotation in TFT_eSPI. Am I right?

examples/arduino/LVGL_Arduino/LVGL_Arduino.ino
KlausMu: You are right, it was my fault. Sorry

examples/arduino/LVGL_Arduino/LVGL_Arduino.ino
KlausMu: You are right, it was my fault. Sorry

examples/arduino/LVGL_Arduino/LVGL_Arduino.ino
KlausMu: Sure, it is not at all clean to have the redeclaration. I will remove it as soon as we found a better way to set the rotation. To me it is not yet clear how to do it. See comments below. `lv_display_set_rotation` does not work for me. Maybe I don't understand it.

examples/arduino/LVGL_Arduino/LVGL_Arduino.ino
liamHowatt: Aha, based on @kisvegabor's review, I see that `lv_display_set_rotation` is not the whole solution. It seems like adding the event callback he described would do it. So `tft->setRotation` would be invoked, but inside the driver lv_tft_espi.cpp as a result of `lv_display_set_rotation` being called by the user.

Are you up to the task? This is a starting point you can begin testing from.
```diff
diff --git a/src/drivers/display/tft_espi/lv_tft_espi.cpp b/src/drivers/display/tft_espi/lv_tft_espi.cpp
index 64e3ee6ab..adf1ef641 100644
--- a/src/drivers/display/tft_espi/lv_tft_espi.cpp
+++ b/src/drivers/display/tft_espi/lv_tft_espi.cpp
@@ -26,6 +26,7 @@ typedef struct {
  *  STATIC PROTOTYPES
  **********************/
 static void flush_cb(lv_display_t * disp, const lv_area_t * area, uint8_t * px_map);
+static void resolution_changed_event_cb(lv_event_t * e);
 
 /**********************
  *  STATIC VARIABLES
@@ -53,9 +54,10 @@ lv_display_t * lv_tft_espi_create(uint32_t hor_res, uint32_t ver_res, void * buf
 
     dsc->tft = new TFT_eSPI(hor_res, ver_res);
     dsc->tft->begin();          /* TFT init */
-    dsc->tft->setRotation(3);   /* Landscape orientation, flipped */
+    dsc->tft->setRotation(0);
     lv_display_set_driver_data(disp, (void *)dsc);
     lv_display_set_flush_cb(disp, flush_cb);
+    lv_display_add_event_cb(disp, resolution_changed_event_cb, LV_EVENT_RESOLUTION_CHANGED, NULL);
     lv_display_set_buffers(disp, (void *)buf, NULL, buf_size_bytes, LV_DISPLAY_RENDER_MODE_PARTIAL);
     return disp;
 }
@@ -80,4 +82,33 @@ static void flush_cb(lv_display_t * disp, const lv_area_t * area, uint8_t * px_m
 
 }
 
+static void resolution_changed_event_cb(lv_event_t * e)
+{
+    lv_display_t * disp = (lv_display_t *)lv_event_get_target(e);
+    lv_tft_espi_t * dsc = (lv_tft_espi_t *)lv_display_get_driver_data(disp);
+    int32_t hor_res = lv_display_get_horizontal_resolution(disp);
+    int32_t ver_res = lv_display_get_vertical_resolution(disp);
+    lv_display_rotation_t rot = lv_display_get_rotation(disp);
+
+    /* TODO: implement resolution change */
+    LV_UNUSED(hor_res);
+    LV_UNUSED(ver_res);
+
+    /* handle rotation */
+    switch(rot) {
+        case LV_DISPLAY_ROTATION_0:
+            dsc->tft->setRotation(0);   /* Portrait orientation */
+            break;
+        case LV_DISPLAY_ROTATION_90:
+            dsc->tft->setRotation(1);   /* Landscape orientation */
+            break;
+        case LV_DISPLAY_ROTATION_180:
+            dsc->tft->setRotation(2);   /* Portrait orientation, flipped */
+            break;
+        case LV_DISPLAY_ROTATION_270:
+            dsc->tft->setRotation(3);   /* Landscape orientation, flipped */
+            break;
+    }
+}
+
 #endif /*LV_USE_TFT_ESPI*/
```

examples/arduino/LVGL_Arduino/LVGL_Arduino.ino
liamHowatt: No need to apologize for helping 😊️

examples/arduino/LVGL_Arduino/LVGL_Arduino.ino
KlausMu: Ah, ok, now I understand, thanks.
I can confirm that this event handler is working great. With that we can set the rotation.

The TODO for the resolution change is something that cannot be achieved with TFT_eSPI, I think. There are no setters for the resolution, you can only set it in the constructor.

Ok, I think you found a much more elegant solution than I did.
Is there anything you want me to provide? A changed PR? Or do you want to do it on your own? @liamHowatt you already did all the work :-)

liamHowatt: That TODO isn't me asking for an implementation from you, it's just what I copied from lv_lcd_generic_mipi.c and fine to leave unimplemented. Sorry for the confusion.

You have a working TFT setup for testing so you are the best one to sign-off on the changes that worked for you. Also, typically the PR stays in the author's hands except for rare cases. If you can commit your changes that'd be awesome, thanks 🙂️

kisvegabor: It's great that there is an agreement :slightly_smiling_face: 

I think it's better if @KlausMu adds the discussed changes as he has an envirement to test it too.

KlausMu: Ok, I added the changes and comitted them. Please be aware that [this PR](https://github.com/lvgl/lvgl/pull/6001) also needs to be merged, otherwise the example will not even compile (the original example never compiled).

kisvegabor (approved):
Beautiful, thank you :blush:

examples/arduino/LVGL_Arduino/LVGL_Arduino.ino
```diff
@@ -15,9 +15,10 @@
 //#include <examples/lv_examples.h>
 //#include <demos/lv_demos.h>
 
-/*Set to your screen resolution*/
+/*Set to your screen resolution and rotation*/
 #define TFT_HOR_RES   320
 #define TFT_VER_RES   240
```
liamHowatt, line 20: > For my display to work, I have to set
> 
> ```
> #define TFT_HOR_RES   240
> #define TFT_VER_RES   320
> 
>  ((lv_tft_espi_t *)(lv_display_get_driver_data(disp)))->tft->setRotation(0);
> ```

Based on what you said earlier, should these values be swapped in the example? Or is it correct the way it is?

```suggestion
#define TFT_HOR_RES   240
#define TFT_VER_RES   320
```

Another way you could do it is put this line `dsc->tft->setRotation(3); /* Landscape orientation, flipped */` back in tft_espi/lv_tft_espi.cpp

examples/arduino/LVGL_Arduino/LVGL_Arduino.ino
liamHowatt: Assuming you mean this one #5999, it has been merged. Thanks for the heads-up. If you want you can [rebase or merge](https://www.atlassian.com/git/tutorials/merging-vs-rebasing) if you feel like it. Just for your convenience -- totally optional in this case.

examples/arduino/LVGL_Arduino/LVGL_Arduino.ino
KlausMu: I think this depends on the display. I would leave it as it is.

In TFT_eSPI, you can set the horizontal and vertical resolution, and the rotation. All this now can be set also in the "LVGL_Arduino.ino" example. For me this is perfect.

liamHowatt (approved):

XuNeo (approved):

PR #6007, opened 2024-04-02
[Draft] Permanently align/anchor 2 objects (w/o parent-child relationship)

### LINKED ISSUE #5665: Permanently align/anchor 2 objects (w/o parent-child relationship)
### Problem to solve

If a child is aligned to its parent with `lv_obj_align()`, then if the parent is moved or resized the child is moved accordingly. However if the 2 objects are non parent-child and they are aligned/anchored with `lv_obj_align_to()`, then if one of the 2 moves[^1] then the other one remains unchanged.

This is clearly [reported in the official documentation](https://docs.lvgl.io/8.3/overview/coords.html#align), but it would be nice to fix it.

[^1]: for example during an animation.

### Success criteria

When 2 objects are anchored each other with `lv_obj_align_to()` then if one of the two is moved the other must follow it.

### Solution outline

A possible work around for the issue I described below in "Rabbit holes" is that despite of having 2 object which are not parent-child, when we set `lv_obj_align_to()` we also set a sort of priority. I mean, if I align obj1 to obj2, then obj2 become a "sort of child" to obj1 for what concerns position. It's not a real/full parenthood relationship, but it's limited to position (and some other features if it is found that this is interesting also for other purposes).

### Rabbit holes

Unfortunately I never contributed to LVGL yet, but I guess that the difference between parent-child changes and non parent-child ones is that in the first case we have to worry about one possible change (i.e. the parent is changed so the child must be updated) whereas in the latter case there is a double relationship (i.e. if any of the 2 is moved, then the other one should be updated).
Moreover in the 2nd case attention should be paid to the fact that this might create a loop, i.e. I move object 1, so I need to update object 2, but hey object 2 is anchored to object 1 so I need to update it, and so forth.

### Testing

Empty window with 2 objects anchored together + an animation moving one of the two. The 2nd object should just follow the 1st one.

### Teaching

Remove this sentence from the documentation:

> Note that, unlike with lv_obj_align(), lv_obj_align_to() can not realign the object if its coordinates or the reference object's coordinates change.

### Considerations

_No response_

valeriosetti:

When an object is aligned with `align_to()` then if the reference if moved, the aligned object will follow.

This PR also add an example making use of this feature: 1 parent object is created together with 3 followers. Then the parent is moved up & down on the screen and the followers follow it.

I marked as draft because this is my very first PR in LVGL so I would like to know whether this PR is moving in the right direction or not according to what it has been required in #5665.

Resolves #5665

### DIFF
--- a/examples/styles/lv_example_style_16.c
+++ b/examples/styles/lv_example_style_16.c
@@ -1,0 +1,82 @@
+#include "../lv_examples.h"
+#if LV_BUILD_EXAMPLES && LV_USE_LABEL
+
+#define OBJ_HOR_LEN     (LV_HOR_RES/5)
+#define OBJ_VER_LEN     (LV_VER_RES/5)
+#define MAX_ANIM        (LV_VER_RES - OBJ_VER_LEN)
+static lv_obj_t * obj_reference;
+
+/* Animate the reference object to bounce up & down. */
+static void lv_anim_callback(void * obj, int value)
+{
+    (void) obj;
+
+    if(value <= MAX_ANIM) {
+        lv_obj_set_y(obj_reference, value);
+    }
+    else {
+        lv_obj_set_y(obj_reference, 2 * MAX_ANIM - value);
+    }
+}
+
+/* Create the following (i.e. anchored) objects. */
+static void create_follower_object(int index)
+{
+    lv_obj_t * obj;
+    lv_obj_t * label;
+
+    obj = lv_obj_create(lv_scr_act());
+    lv_obj_set_size(obj, OBJ_HOR_LEN, OBJ_VER_LEN);
+    lv_obj_align_to(obj, obj_reference, LV_ALIGN_OUT_RIGHT_MID, OBJ_HOR_LEN * (index - 1), 0);
+    lv_obj_set_style_bg_color(obj, lv_palette_main(LV_PALETTE_BLUE), 0);
+
+    label = lv_label_create(obj);
+    lv_obj_align(label, LV_ALIGN_CENTER, 0, 0);
+    lv_label_set_text_fmt(label, "Follower #%d", index);
+}
+
+/**
+ * Align some objects to a non-parent anchor then move the anchor
+ * and check that the anchored objects follow it.
+ */
+void lv_example_style_16(void)
+{
+    lv_obj_t * label;
+    lv_obj_t * obj_child;
+
+    lv_obj_set_style_pad_all(lv_scr_act(), 0, 0);
+
+    /* Reference object. */
+    obj_reference = lv_obj_create(lv_scr_act());
+    lv_obj_set_size(obj_reference, OBJ_HOR_LEN, OBJ_VER_LEN);
+    lv_obj_align(obj_reference, LV_ALIGN_TOP_LEFT, 0, 0);
+    lv_obj_set_style_bg_color(obj_reference, lv_palette_main(LV_PALETTE_RED), 0);
+
+    /* Normal child of the reference object. */
+    obj_child = lv_obj_create(obj_reference);
+    lv_obj_set_size(obj_child, LV_PCT(100), LV_PCT(100));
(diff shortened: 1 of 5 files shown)


src/core/lv_obj_pos.c
```diff
@@ -319,13 +319,26 @@ void lv_obj_align(lv_obj_t * obj, lv_align_t align, int32_t x_ofs, int32_t y_ofs
     lv_obj_set_pos(obj, x_ofs, y_ofs);
 }
 
+static void add_anchored_children(lv_obj_t * obj, const lv_obj_t * base)
```
liamHowatt, line 322: Please rename to `add_anchored_child` and move to the section in the file called `STATIC FUNCTIONS` and add the function prototype to `STATIC PROTOTYPES`

src/core/lv_obj_pos.c
```diff
@@ -319,13 +319,26 @@ void lv_obj_align(lv_obj_t * obj, lv_align_t align, int32_t x_ofs, int32_t y_ofs
...
+{
+    base->spec_attr->anchored_child_count++;
```
liamHowatt, line 324: Somewhere before we arrive here we need to do one of

- `if(base->spec_attr == NULL) return;`
- `lv_obj_allocate_spec_attr(base);`

May not be in this function, but somewhere, such that the test cases in the CI pass and there's no chance of it being NULL.

examples/styles/lv_example_style_16.c
```diff
@@ -0,0 +1,82 @@
+#include "../lv_examples.h"
+#if LV_BUILD_EXAMPLES && LV_USE_LABEL
+
+#define OBJ_HOR_LEN     (LV_HOR_RES/5)
+#define OBJ_VER_LEN     (LV_VER_RES/5)
+#define MAX_ANIM        (LV_VER_RES - OBJ_VER_LEN)
+static lv_obj_t * obj_reference;
+
+/* Animate the reference object to bounce up & down. */
+static void lv_anim_callback(void * obj, int value)
```
liamHowatt, line 10: `lv_` not needed

examples/styles/lv_example_style_16.c
```diff
@@ -0,0 +1,82 @@
...
+{
+    (void) obj;
```
liamHowatt, line 12:
```suggestion
    LV_UNUSED(obj);
```

liamHowatt (commented):
Thanks very much @valeriosetti for sticking with this feature for so long and opening a PR!

After you respond to this feedback I will look deeper into your implementation.

<br>

Free the `anchored_children` [here](https://github.com/valeriosetti/lvgl/blob/997520a1648eeb522d9908a97a0927a6642cb063/src/core/lv_obj.c#L374). i.e.

```c
if(obj->spec_attr->anchored_children) {
    lv_free(obj->spec_attr->anchored_children);
    obj->spec_attr->anchored_children = NULL;
}
```

<br>

This test is failing because the behavior of `lv_obj_align_to` has changed. If you're able to make this function maintain compatibility, please do. Otherwise if this behavior is correct then update the ref image with the error one to become the new ref image.

tests/ref_imgs/libs/freetype_1.lp64.png
https://github.com/valeriosetti/lvgl/blob/anchor/tests/src/test_cases/libs/test_freetype.c#L450

Original:
freetype_1 lp64
Error:
freetype_1 lp64_err

valeriosetti:
Thanks @liamHowatt for your review! I will start addressing comments.
@kisvegabor thanks for the feedback. I thought that #5665 (linked in the description) was enough for this. Should I open a new one?

kisvegabor:
My closing though in #5665 was
> Sorry, for mentioning only problems, however these are the main concerns why it is not implemented yet. I hope that we can figure out something together.

So I think it requires a deeper investigation to discover all options. Maybe we will find a completely different approach if we dive deeper. 


> Should I open a new one?

Yes, please.

PR #6036, opened 2024-04-07
chore(mem): changed log level to ERROR if memory couldn't be allocated

KlausMu:

If memory couldn't be allocated, this should be logged as an error.

With default log settings (LV_LOG_LEVEL_WARN) currently you are not notified if LVGL is running out of memory. The application simply stops running.

Since LVGL 9 needs approx 33% more memory than LVGL 8, this is now more likely to happen.

Please also see this discussion
https://forum.lvgl.io/t/lvgl-9-higher-memory-usage-and-different-usage-reports-in-lvgl-9-compared-to-8/15308

### DIFF
--- a/src/stdlib/lv_mem.c
+++ b/src/stdlib/lv_mem.c
@@ -70,7 +70,7 @@ void * lv_malloc(size_t size)
     void * alloc = lv_malloc_core(size);
 
     if(alloc == NULL) {
-        LV_LOG_INFO("couldn't allocate memory (%lu bytes)", (unsigned long)size);
+        LV_LOG_ERROR("couldn't allocate memory (%lu bytes)", (unsigned long)size);
 #if LV_LOG_LEVEL <= LV_LOG_LEVEL_INFO
         lv_mem_monitor_t mon;
         lv_mem_monitor(&mon);
@@ -99,7 +99,7 @@ void * lv_malloc_zeroed(size_t size)
 
     void * alloc = lv_malloc_core(size);
     if(alloc == NULL) {
-        LV_LOG_INFO("couldn't allocate memory (%lu bytes)", (unsigned long)size);
+        LV_LOG_ERROR("couldn't allocate memory (%lu bytes)", (unsigned long)size);
 #if LV_LOG_LEVEL <= LV_LOG_LEVEL_INFO
         lv_mem_monitor_t mon;
         lv_mem_monitor(&mon);


kisvegabor:
I few questions to better understand the issue:
1. What kind of widgets do you use the most? (Can you share the whole UI code maybe?)
2. Do you style properties like `opa_layered`, `transform_*`, `belnd_mode`, `clip_corner`?
3. Please share your `lv_conf.h` (attach as a zipped file)
4. Please share some call stacks when the out of memory issue happens.

> With default log settings (LV_LOG_LEVEL_WARN) currently you are not notified if LVGL is running out of memory. The application simply stops running.

I recommend enabling [this assert](https://github.com/lvgl/lvgl/blob/master/lv_conf_template.h#L269) to make the application stop.

KlausMu:
> I few questions to better understand the issue:
> 
> 1. What kind of widgets do you use the most? (Can you share the whole UI code maybe?)

Sure, thanks for your reply.
Here you can find the project: https://github.com/CoretechR/OMOTE

It is an ESP32 PlatformIO project. If you want to, you can run the whole project in the simulator (PlatformIO with Visual Studio Code). No ESP32 needed.

You can simply choose target "linux_64bit" in Platformio.
For Windows, you need MSYS2 to be installed. In Windows, you can use targets "windows_64bit" or "windows_32bit".
I'm sure you know how to run the simulator :-) If not:
https://github.com/CoretechR/OMOTE/wiki/03-Software-simulator-for-fast-creating-and-testing-of-LVGL-GUIs

Main branch is still using lvgl 8.
This feature branch is using lvgl 9: https://github.com/CoretechR/OMOTE/tree/lvgl_9.1

We are mainly using the tabview, labels, buttons, sliders and checkboxes. Nothing complicated (I think) and not too many widgets.

Files creating widgets are:
https://github.com/CoretechR/OMOTE/blob/main/Platformio/src/applicationInternal/gui/guiBase.cpp
https://github.com/CoretechR/OMOTE/blob/main/Platformio/src/applicationInternal/gui/guiMemoryOptimizer.cpp
https://github.com/CoretechR/OMOTE/blob/main/Platformio/src/devices/mediaPlayer/device_appleTV/gui_appleTV.cpp
https://github.com/CoretechR/OMOTE/blob/main/Platformio/src/devices/misc/device_smarthome/gui_smarthome.cpp
https://github.com/CoretechR/OMOTE/blob/main/Platformio/src/guis/gui_irReceiver.cpp
https://github.com/CoretechR/OMOTE/blob/main/Platformio/src/guis/gui_numpad.cpp
https://github.com/CoretechR/OMOTE/blob/main/Platformio/src/guis/gui_sceneSelection.cpp
https://github.com/CoretechR/OMOTE/blob/main/Platformio/src/guis/gui_settings.cpp

> 2. Do you style properties like `opa_layered`, `transform_*`, `belnd_mode`, `clip_corner`?

none of them

> 3. Please share your `lv_conf.h` (attach as a zipped file)

lvgl configuration is in file platformio.ini
https://github.com/CoretechR/OMOTE/blob/main/Platformio/platformio.ini

> 4. Please share some call stacks when the out of memory issue happens.

In lvgl 8, memory set to 32k (ESP32 and windows_32bit) and 53k (linux_64bit and windows_64bit) is completely sufficient. Always 10k or more memory free.

In lvgl 9, with the same memory settings, the application is running out of memory. On third more memory is needed, aproximately (between 30% and 40%, depending on the UI which is currently in memory).

> > With default log settings (LV_LOG_LEVEL_WARN) currently you are not notified if LVGL is running out of memory. The application simply stops running.
> 
> I recommend enabling [LV_ASSERT_MALLOC](https://github.com/lvgl/lvgl/blob/master/lv_conf_template.h#L269) to make the application stop on out-of-memory error.

No matter if I enable it or not, I get 
```
[Error] (4.864, +4864)   lv_anim_start: Asserted at expression: new_anim != NULL (Out of memory) lv_anim.c:91
```

If I spent even less memory, I get a segmentation fault during the animation with this stacktrace:
Screenshot 2024-04-08 103404

Is this something you can work with? Anything else I could provide?
Thanks a lot for your support!

kisvegabor:
Thank you for the details information.
I could run the project on my Linux notebook, but haven't seen anything suspicious at first look. What I know for sure is that in v9 we use `int32_t` for coordinates instead of `lv_coord_t` (which was usually `int16_t`), however I don't think it required 30% extra RAM.

I'm planning to investigate it further, but this week we will be at the Embedded World fair, and I can look into in only next week. :slightly_frowning_face:

KlausMu:
Thanks for having a look - and welcome to Germany :-)

If you are using the simulator with target "linux_64bit" , you should set LV_MEM_SIZE="(53U * 1024U)"
lvgl 8: always at least 10k free
lvgl 9: out of memory


I did another comparison using the example "LVGL_Arduino.ino" included in the lvgl library.
Memory was set to LV_MEM_SIZE="(32U * 1024U)"

This example is really simple. Only one single label, nothing else.
As second test I created 20 labels, but no other widgets.

lvgl 8.4.0
one label: memory size:  32768, used:   4860 (15%), free:  27908 (85%)
20 labels: memory size:  32768, used:   8704 (27%), free:  24064 (73%)
-> 202 bytes/label

lvgl 9.1.0
one label: memory size:  31172, used:   6088 (20%), free:  25084 (80%)
20 labels: memory size:  30716, used:  10088 (33%), free:  20628 (67%)
-> 234 bytes/label (+15%)

The increase for a single label here is only 15%. Not 30%-40% as I experienced whith the whole application, but this might be because of more complex widgets in the application.

BTW: Why is the total memory size in lvgl 9.1.0 not 32768, but 31172 or 30716 or something else, depending on the widgets which were created? Shouldn't be the memory size exactly what was allocated?

kisvegabor:
Quick question: Have you enabled `LV_OBJ_STYLE_CACHE`?

KlausMu:
> Quick question: Have you enabled `LV_OBJ_STYLE_CACHE`?

No, it is not enabled.

KlausMu:
With `LV_OBJ_STYLE_CACHE` activated, it takes about 1.5% more memory.
But as said, I did not activate it.

kisvegabor:
Yes, it increases the memory usage a little. 

So we have 22 extra bytes per label. I made a quick calculation and it seems the +2 bytes from changing from `(u)int16_t` to `(u)int32_t` can add up to 22.

Is it a real issue in your case? I mean I understand that +15% is really a clear increase, but IMO it shouldn't be a game changer, unless you are running your device on the very limits.

KlausMu:
Thanks for having a look.

Well, only the very simple example "LVGL_Arduino.ino" needs only 15% more memory.
The application I'm working on needs 33% more.

For us, it is not impossible to switch to LVGL 9, but I have to save memory at a different place. The application ran with 32k on lvgl 8. This is not possible anymore. Now I have to give LVGL 9 about 10k more.
If you have only about 35k free overall (and you always have to keep about 12k free for WiFi peaks, so in fact there is only about 23k free), then 10k is a lot.

But if you say there is nothing we can do about it, then we have to live with that.

And as you said, ESP32 applications already being at the limit with LVGL 8 will have to find a way to handle the additional memory usage.

Hopfully this will not happen again such drastically with lvgl 10 :-)

Thanks for your support.

kisvegabor:
Ah, I'm sorry, I'll keep this issue in mind as doing further debugging in the future.

kisvegabor:
To comment on the PR itself too: I'm not sure we should change the log level to ERROR. Even LVGL tries to allocate larger memories sometimes, and if it fails it will try it later.

KlausMu:
Ok, I did not know this. Then the PR is not the right solution. I close it.

PR #6047, opened 2024-04-09
fix(evdev): Add missing include for strerror

niklasf:


liamHowatt (approved):

FASTSHIFT (approved):

kisvegabor:
@FASTSHIFT  [mentioned](https://github.com/lvgl/lvgl/pull/6041#issuecomment-2042827899) that `strerror` might not be supported in NuttX. I wonder if `evdev` is supported in NuttX because if so, we might want to avoid using `strerror` here.

niklasf:
Looks like the virtio-input is close, but some more changes would be required. So just keep using it until NuttX support is actually added?

FASTSHIFT:
> @FASTSHIFT [mentioned](https://github.com/lvgl/lvgl/pull/6041#issuecomment-2042827899) that `strerror` might not be supported in NuttX. I wonder if `evdev` is supported in NuttX because if so, we might want to avoid using `strerror` here.

NuttX does not support `evdev`. You can ignore NuttX.

kisvegabor (approved):

PR #6050, opened 2024-04-09
fix include path of lvgl.h

cheops:
### fix compilation error `lvgl/src/font/lv_font_simsun_14_cjk.c:10:14: fatal error: lvgl/lvgl.h: No such file or directory`
```
lvgl/src/font/lv_font_simsun_14_cjk.c:10:14: fatal error: lvgl/lvgl.h: No such file or directory
   10 |     #include "lvgl/lvgl.h"
      |              ^~~~~~~~~~~~~
```

looking at the other font files, the include path in this file is wrong.
Compilation succeeds after the change.
Compiled from https://github.com/kdschlosser/lvgl_micropython

### DIFF
--- a/src/font/lv_font_simsun_14_cjk.c
+++ b/src/font/lv_font_simsun_14_cjk.c
@@ -7,7 +7,7 @@
 #ifdef LV_LVGL_H_INCLUDE_SIMPLE
     #include "lvgl.h"
 #else
-    #include "lvgl/lvgl.h"
+    #include "../../lvgl.h"
 #endif
 
 #ifndef LV_FONT_SIMSUN_14_CJK


liamHowatt (approved):
FYI this is how the correct header is automated. Easy to miss if the font isn't generated with this `generate_all.py` script.

https://github.com/lvgl/lvgl/blob/500e16e96883c8b7a8b010e9f54e8bcfb488c10f/scripts/built_in_font/generate_all.py#L89

FASTSHIFT (approved):

W-Mai (approved):

kisvegabor (approved):
Thank you!

PR #6054, opened 2024-04-10
fix(vg_lite): fix incorrect cache operation

FASTSHIFT:

After the CPU decodes the image, it should call the flush D-Cache operation to completely write the pixel data in the cache to RAM.

The invalidate D-Cache operation is incorrect, which will cause the data in the cache that is not written to RAM to be lost.

Invalidate cache is usually done when the peripheral writes to the memory first and the CPU needs to read it.

### Notes

### DIFF
--- a/src/drivers/nuttx/lv_nuttx_cache.c
+++ b/src/drivers/nuttx/lv_nuttx_cache.c
@@ -27,6 +27,7 @@
  **********************/
 
 static void invalidate_cache(const lv_draw_buf_t * draw_buf, const lv_area_t * area);
+static void flush_cache(const lv_draw_buf_t * draw_buf, const lv_area_t * area);
 
 /**********************
  *  STATIC VARIABLES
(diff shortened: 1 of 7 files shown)


src/drivers/nuttx/lv_nuttx_cache.c
```diff
@@ -44,28 +45,44 @@ void lv_nuttx_cache_init(void)
...
  *   STATIC FUNCTIONS
  **********************/
 
-static void invalidate_cache(const lv_draw_buf_t * draw_buf, const lv_area_t * area)
+static void draw_buf_to_region(
+    const lv_draw_buf_t * draw_buf, const lv_area_t * area,
+    lv_uintptr_t * start, lv_uintptr_t * end)
 {
     LV_ASSERT_NULL(draw_buf);
+    LV_ASSERT_NULL(area);
```
XuNeo, line 60: Set `area` to NULL is a valid parameter to indicate operate on the whole draw buffer.

```c
/**
 * Invalidate the cache of the buffer
 * @param draw_buf     the draw buffer needs to be invalidated
 * @param area         the area to invalidate in the buffer,
 *                     use NULL to invalidate the whole draw buffer address range
 */
void lv_draw_buf_invalidate_cache(const lv_draw_buf_t * draw_buf, const lv_area_t * area);
```

FASTSHIFT: The `area == NULL` situation has been handled in function: `lv_draw_buf_invalidate_cache_user` `lv_draw_buf_flush_cache_user`

```c
lv_area_t full;
if(area == NULL) {
    draw_buf_get_full_area(draw_buf, &full);
    area = &full;
}
```

kisvegabor:
Might it be required e.g. in case of LodePNG too? E.g. we have some image data in cache and a GPU is used to blend the image. If the GPU by passes the cache, it won't see image in the cache.

FASTSHIFT:
> Might it be required e.g. in case of LodePNG too? E.g. we have some image data in cache and a GPU is used to blend the image. If the GPU by passes the cache, it won't see image in the cache.

Yes, we should consider adding a callback function after image decode open to handle cache flush in a unified manner.

kisvegabor:
It seems simple enough, therefore I suggest adding it in the PR.

XuNeo (dismissed):

kisvegabor (changes requested):
>> Yes, we should consider adding a callback function after image decode open to handle cache flush in a unified manner.

> It seems simple enough, therefore I suggest adding it in the PR.

FASTSHIFT:
> > > Yes, we should consider adding a callback function after image decode open to handle cache flush in a unified manner.
> 
> > It seems simple enough, therefore I suggest adding it in the PR.

Done.

XuNeo (approved):

kisvegabor (approved):

PR #6064, opened 2024-04-11
fix(encoder): always fire LV_EVENT_LONG_PRESSED to indev callback

### LINKED ISSUE #5991: v9.1.0 - Encoder does not fire LV_EVENT_LONG_PRESSED
### LVGL version

v9.1.0

### What happened?

I have an encoder indev. It successfully rotates and handles short clicks, but long clicks does not issue LV_EVENT_LONG_PRESSED, although they do issue LV_EVENT_LONG_PRESSED_REPEATED regularly

### How to reproduce?

```c
lv_indev_add_event_cb(indev_encoder, long_press_cb, LV_EVENT_LONG_PRESSED, NULL); // never fires
lv_indev_add_event_cb(indev_encoder, long_press_cb, LV_EVENT_LONG_PRESSED_REPEAT, NULL); // fires regularly when long pressing
lv_indev_add_event_cb(indev_encoder, long_press_cb, LV_EVENT_PRESSED, NULL); // fires on any press

```

aeroniemi:
Fixes #5991 

Currently LV_EVENT_LONG_PRESSED isn't fired by encoder buttons if the current obj is editable or scrollable - this fixes that by ensuring that it runs every time irrespective of the obj state.

### Notes

### DIFF
--- a/src/indev/lv_indev.c
+++ b/src/indev/lv_indev.c
@@ -907,6 +907,10 @@ static void indev_encoder_proc(lv_indev_t * i, lv_indev_data_t * data)
             i->longpr_rep_timestamp = lv_tick_get();
 
             if(data->key == LV_KEY_ENTER) {
+                /* Always send event to indev callbacks*/
+                lv_indev_send_event(indev_act, LV_EVENT_LONG_PRESSED, indev_obj_act);
+                if(indev_reset_check(indev_act)) return;
+
                 bool editable_or_scrollable = lv_obj_is_editable(indev_obj_act) ||
                                               lv_obj_has_flag(indev_obj_act, LV_OBJ_FLAG_SCROLLABLE);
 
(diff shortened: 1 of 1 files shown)


liamHowatt (commented):
Looks good, thanks for creating this!

I'd like to get @kisvegabor to verify whether this one event is okay or whether all of the press events should get this treatment too, in one PR. That way we can point to this commit as "the commit where all press events are emitted for encoder indevs" instead of the transition being spread across many PRs.

Otherwise looks good.

kisvegabor:
Thank you @aeroniemi.

I'm not sure if we need to support all the events here. Adding support for long press makes sens as it's related to the edit/navigation mode change and the user might be curious about it. 

The other events are omitted intentionally and it might be confusing to receive all the events even if they don't do anything. (E.g. `LV_EVENT_SHORT_CLICKED` on entering edit mode)

The only other indev event that I'd add unconditionally is `RELEASE` because it's very basic. 

It feel a little bit arbitrary and I'm not sure this is the right approach. So I'd be curious to hear your opinion.

liamHowatt:
Aha, I see. Long press does make the most sense because the repeat event is always emitted and it just makes sense for long press to also be emitted. I see what you mean about the other events not having as strong of a case to be added, especially short click as you mentioned.

I guess my train of thought was that a user might want their encoder button to always have its events reflect the physical encoder button regardless of the state of the UI. But if that were the case, then they would write their own logic outside LVGL if it was that critical.

I could see release being added but I'd ask if press should also be added. Currently press is only emitted if it causes "editing" mode to switch on.

**TLDR** I think it's fine to leave this PR as-is.

liamHowatt (approved):

kisvegabor:
> TLDR I think it's fine to leave this PR as-is.

Okay, let"s leave it as it is :slightly_smiling_face:

kisvegabor (approved):

FASTSHIFT (approved):

Issue #6011, opened 2024-04-03
API retirement


kisvegabor:
### Problem to solve

Continue #5096

To provide a better developer experience we need to have standards to retire APIs.

### Success criteria

- All the non essential functions are removed from the public API (`lvgl.h`) 
- The retirement policy is documented
- The core API is tested

### Solution outline

Rework the API:
1. The core API should mean the widgets, events, animation, styles, 3rd party libraries, drivers and the essential types (e.g. `lv_area_t`). Review, restructure and clean up `lvgl.h` to not include anything else than these. It can be achieved by adding several `*_private.h` files and moving typedefs to `lv_types.h`. `_lv_*` functions shall be moved to their own `private.h` file. 
2. APIs that shall not be part of the public API: image decoding, text processing, draw functions (except the top level ones, e.g. `lv_draw_rect`), layout APIs, linked list, lru, cache, etc.
3. Not sure about: LVGL's STDIO API

Retirement policy:
1. Keep the API for at least one more major version (e.g. APIs retired in v9 shall be available in v10, and can be removed in v11)
2. Provide backward compatibility in `lv_api_map*.h` files
3. Add `LV_LOG_WARN` in old APIs (we need to have `static inline` functions in `lv_api_map*.h` to add a warning)
4. Update the docs
 

### Rabbit holes

1. Extract the API from [JSON](https://github.com/lvgl/lvgl/pull/5677)
2. Create a `.h` file from the prototypes and compare it with a reference `*.h` file. 

### Testing

Testing:
1. Extract the API from [JSON](https://github.com/lvgl/lvgl/pull/5677)
2. Create a `.h` file from the prototypes and compare it with a reference `*.h` file. 

### Teaching

1. Add warnings in the deprecated functions
2. On each major release communicate the deprecated APIs too

kisvegabor:
@liamHowatt As you have already gained experience with `_lv_text` / `lv_text` update in this field would you mind taking this update?

liamHowatt:
Okay!

kisvegabor:
Not stale.

kisvegabor:
Let's wrap up this issue. We can agree on keeping all deprecated APIs for at least one extra major release.   

Now we have `lv_api_map_v*.h`. This way we can clearly track when the API changes happened. 

We can decide how to inform users about specific API retirements case by case. 

So I close this issue now. Let's reopen it if needed.

PR #6068, opened 2024-04-12
refactor(API): don't expose private symbols in lvgl.h. phase-out "_lv*" names

liamHowatt:

For #6011

- No private functions or struct members can be reached via lvgl.h
- Create *_private.h header files that are included where needed in .c files.
- All *_private.h symbols can be reached through lvgl_private.h
- No more symbols that are named like _lv*. Remove leading underscore and make private.

### Notes


liamHowatt:
Progress report

🚧️ WIP 🚧️

The progress so far has been directed by getting the number of lines output by this command near zero.

```
gcc -std=c11 -E -Wno-incompatible-pointer-types -DLV_LVGL_H_INCLUDE_SIMPLE -DLV_CONF_INCLUDE_SIMPLE -DLV_USE_DEV_VERSION -DPYCPARSER -DLV_CONF_PATH=$PWD/lv_conf.h -I $PWD/../lvgl $PWD/../lvgl/lvgl.h | grep -iE -e '(^|[^a-z0-9_])_lv_' -e '(struct|enum)\s+[a-z0-9_]+\s*\{' -e 'private'
```

Derived from the WIP API JSON generator PR #5677. In particular the lv_conf.h is lv_conf_template.h with [this](https://github.com/lvgl/lvgl/blob/0bbf5a18ebafc7cbebd9c9b2030823d89957e977/scripts/gen_json/gen_json.py#L84-L113) applied and the `gcc` invocation above is the same as what gets invoked [here](https://github.com/lvgl/lvgl/blob/0bbf5a18ebafc7cbebd9c9b2030823d89957e977/scripts/gen_json/gen_json.py#L191). The `-E` flag runs the preprocessor and dumps the preprocessed IR without compiling.

It dumps all symbols visible in the lvgl.h header. `grep` is looking for:

1. symbols that begin with `_lv`
2. `struct`s and `enum`s that have visible members.
3. accidental includes of *_private.h files

<hr>

90% of the _lv symbols in lvgl.h have been hidden so far. The remaining are tricky cases that need to be reasoned about. i.e. `_lv_log_add`.

Some APIs such as `lv_ll` are not going to be exposed anymore.

TODO:

- Compile: Include the _private.h headers where they're needed.
- Finish renaming: not all _lv symbols have been renamed yet even though they're no longer exposed.
- Ensure no APIs were lost in this PR (using #5677).
- Add a CI lint (using #5677).

src/widgets/roller/lv_roller.h
```diff
@@ -33,13 +33,17 @@ extern "C" {
...
+#endif /*DOXYGEN*/
     LV_ROLLER_MODE_NORMAL, /**< Normal mode (roller ends at the end of the options).*/
     LV_ROLLER_MODE_INFINITE, /**< Infinite mode (roller can be scrolled forever).*/
 };
 
 #ifdef DOXYGEN
-typedef _lv_roller_mode_t lv_roller_mode_t;
+typedef enum lv_roller_mode_t lv_roller_mode_t;
 #else
 typedef uint8_t lv_roller_mode_t;
```
kisvegabor: We `typedef` some enums to `uint8_t` to allow defining them as a bitfields (e.g. ` lv_roller_mode_t mode : 1;`)

However this practice is used inconsistently in LVGL now and I'm not sure if it's worth the loss in readability. 
Could you update these like this?
```c
typedef enum {
    LV_ROLLER_MODE_NORMAL, /**< Normal mode (roller ends at the end of the options).*/
    LV_ROLLER_MODE_INFINITE, /**< Infinite mode (roller can be scrolled forever).*/
}lv_roller_mode_t;

...

//In lv_roller_t
uint32_t mode : 1;
```

kisvegabor (commented):
Very nice! 

Can you share some metrics about the original number of symbols and current number of symbols?

kisvegabor:
[API retirement](https://trello.com/c/Rk3jd6k1/1-api-retirement)

examples/libs/tiny_ttf/lv_example_tiny_ttf_3.c
```diff
@@ -1,3 +1,4 @@
+#include "../../../src/others/observer/lv_observer_private.h"
```
kisvegabor: The private header files shouldn't be required to "just use" a features. In this file on the basic observer features are used, so `lv_observer_private.h` shouldn't be needed.

src/others/monkey/lv_monkey_private.h
```diff
@@ -0,0 +1,63 @@
...
+ *      TYPEDEFS
+ **********************/
+
+struct lv_monkey_config_t {
+    /**< Input device type*/
+    lv_indev_type_t type;
+
+    /**< Monkey execution period*/
+    struct {
+        //! @cond Doxygen_Suppress
```
kisvegabor: Why do we need this Doxygen comment?

examples/anim/lv_example_anim_2.c
```diff
@@ -1,3 +1,4 @@
+#include "../../src/misc/lv_anim_private.h"
```
kisvegabor: Same here and probably on a lot of other places too.

kisvegabor (commented):
What a huge update! I couldn't review it in detail yet, so my comments are just an early feedback.

src/others/monkey/lv_monkey_private.h
liamHowatt: It got copy-pasted from the public header by my refactoring automation. I can remove it.

liamHowatt:
Hey @kisvegabor, it looks like losing APIs isn't going to be a problem. After I push a cleaned up commit, what are the things we want to do to get this merged?

Here are all the structs with exposed members. Most of them are in the global singleton, which is why they have to be in this list. Should we add any more? i.e. anim?

- lv_color_t
- lv_color16_t
- lv_color32_t
- lv_color_hsv_t
- lv_color16a_t
- lv_image_dsc_t
- lv_area_t
- lv_draw_buf_t
- lv_point_t
- lv_draw_arc_dsc_t
- lv_draw_line_dsc_t
- lv_draw_label_dsc_t
- lv_draw_rect_dsc_t
- lv_draw_image_dsc_t
- lv_global_t
- lv_style_t
- lv_font_fmt_rle_t
- lv_fs_drv_t
- lv_timer_state_t
- lv_anim_state_t
- lv_tick_state_t
- lv_draw_buf_handlers_t
- lv_draw_global_info_t
- lv_draw_sw_shadow_cache_t
- lv_tlsf_state_t
- lv_sysmon_backend_data_t
- lv_draw_sw_mask_radius_circle_dsc_arr_t
- lv_layout_dsc_t
- lv_font_glyph_dsc_t
- lv_subject_t
- lv_draw_sw_mask_radius_circle_dsc_t
- lv_array_t
- lv_draw_dsc_base_t
- lv_point_precise_t
- lv_image_header_t
- lv_grad_dsc_t
- lv_style_const_prop_t
- lv_gradient_stop_t
- lv_yuv_plane_t
- lv_calendar_date_t
- lv_fs_file_t
- lv_font_fmt_txt_dsc_t
- lv_font_t
- lv_font_fmt_txt_cmap_t
- lv_font_fmt_txt_glyph_dsc_t
- lv_font_fmt_txt_kern_classes_t
- lv_sqrt_res_t

Intentionally hidden APIs. It's just the linked list. I need to add LRU and cache, right?

- ll (linked list)

Protos that are now public

- lv_obj_style_apply_color_filter
- lv_style_prop_lookup_flags
- lv_log_add

Kept inline

- lv_string.h
- lv_style.h
- lv_obj_style.h
- lv_obj_style_gen.h
- lv_style_gen.h

Finally, I attached a file where each row gives a file name and then a private header that the file includes.
[private_includes.txt](https://github.com/lvgl/lvgl/files/15173165/private_includes.txt)

liamHowatt:
I'm not sure why CI isn't running.

FYI here are the generated JSONs

`master` `lvgl.h`: 
`master` `lvgl.h`: 
`master` `lvgl.h`:

kisvegabor:
Thank you Liam!

Can we hide `lv_global_t`? I these shall be hidden to 
```
lv_global_t
lv_font_fmt_rle_t
lv_timer_state_t
lv_anim_state_t
lv_tick_state_t
lv_draw_buf_handlers_t
lv_draw_global_info_t
lv_draw_sw_shadow_cache_t
lv_tlsf_state_t
lv_sysmon_backend_data_t
lv_draw_sw_mask_radius_circle_dsc_arr_t
lv_layout_dsc_t
lv_font_glyph_dsc_t
lv_draw_sw_mask_radius_circle_dsc_t
lv_font_fmt_txt_dsc_t
lv_font_fmt_txt_cmap_t
lv_font_fmt_txt_glyph_dsc_t
lv_font_fmt_txt_kern_classes_t
lv_sqrt_res_t
```
Is it possible?

demos/widgets/lv_demo_widgets.c
```diff
@@ -6,6 +6,12 @@
 /*********************
  *      INCLUDES
  *********************/
+#include "../../src/widgets/chart/lv_chart_private.h"
+#include "../../src/misc/lv_anim_private.h"
+#include "../../src/draw/lv_draw_triangle_private.h"
+#include "../../src/draw/lv_draw_rect_private.h"
+#include "../../src/draw/lv_draw_private.h"
+#include "../../src/core/lv_obj_private.h"
```
kisvegabor: The goal would be to not have any `private.h` includes in demos and examples. It would show that everything is public to use LVGL as a normal user.

kisvegabor (commented):
Sorry, I just noticed that I haven't sent my review...

src/layouts/flex/lv_flex.h
```diff
@@ -22,9 +22,9 @@ extern "C" {
  *      DEFINES
  *********************/
 
-#define _LV_FLEX_COLUMN        (1 << 0)
-#define _LV_FLEX_WRAP       (1 << 2)
-#define _LV_FLEX_REVERSE    (1 << 3)
+#define LV_FLEX_COLUMN        (1 << 0)
```
kisvegabor, line 25: If remove the underscrore let's `#undef` these after `lv_flex_flow_t`.

kisvegabor (commented):
It's a massive and hard to review all the details but in general it looks okay.  It's great that  the idea of `private.h` headers is established now.

I suggest merging it ASAP to avoid conflicts and open follow up PRs if needed.

kisvegabor:
@kdschlosser The failing MP CI probably can be fixed by including `lvgl_private.h` in `lv_mpy.c`. Could check it?

kdschlosser:
the failing CI is more than likely because of `lv_global_t` being made hidden. If you look at the lv_conf.h file for the micropython binding you will see attachment to that.

```c
/*Garbage Collector settings
 *Used if lvgl is bound to higher level language and the memory is managed by that language*/
extern void mp_lv_init_gc();
#define LV_GC_INIT() mp_lv_init_gc()

#define LV_ENABLE_GLOBAL_CUSTOM 1
#if LV_ENABLE_GLOBAL_CUSTOM
    extern void *mp_lv_roots;
    #define LV_GLOBAL_CUSTOM() ((lv_global_t*)mp_lv_roots)
#endif
```

That stuff needs to remain public and not private.

If we include the private headers into the build the size of the firmware is going to become a lot larger. This is due to the number of additional structures that are going to get added resulting in a large number of qstrings being added.

kdschlosser:
the other thing you have to keep a really close eye out for is when you make structures private and they hold callback functions if the type for the function pointer is not defined as having the user_data passed as the last parameter then it tests to see if the first parameter is a structure and if it is then it looks for a user_data field in that structure. If the structure has been made private it is not going to be able to locate the user_data inside the structure.

You will need to modify the callbacks being made to add the user_data as a parameter. The user_data is how the "real" callback is stored and passed in order to get called from the C code. So that MUST be visible as either a parameter or as a field in the structure. There is no way around this.

kdschlosser:
Here is an example as I am sure that this structure has possibly been made private.

```
struct _lv_fs_drv_t;
typedef struct _lv_fs_drv_t lv_fs_drv_t;
struct _lv_fs_drv_t {
    char letter;
    uint32_t cache_size;
    bool (*ready_cb)(lv_fs_drv_t * drv);                                                                                              <========== NO GOOD

    void * (*open_cb)(lv_fs_drv_t * drv, const char * path, lv_fs_mode_t mode);                                  <========== NO GOOD
    lv_fs_res_t (*close_cb)(lv_fs_drv_t * drv, void * file_p);                                                                    <========== NO GOOD
    lv_fs_res_t (*read_cb)(lv_fs_drv_t * drv, void * file_p, void * buf, uint32_t btr, uint32_t * br);             <========= NO GOOD
    lv_fs_res_t (*write_cb)(lv_fs_drv_t * drv, void * file_p, const void * buf, uint32_t btw, uint32_t * bw);             <==== NO GOOD
    lv_fs_res_t (*seek_cb)(lv_fs_drv_t * drv, void * file_p, uint32_t pos, lv_fs_whence_t whence);             <========= NO GOOD
    lv_fs_res_t (*tell_cb)(lv_fs_drv_t * drv, void * file_p, uint32_t * pos_p);                                           <========== NO GOOD

    void * (*dir_open_cb)(lv_fs_drv_t * drv, const char * path);                                                           <========== NO GOOD
    lv_fs_res_t (*dir_read_cb)(lv_fs_drv_t * drv, void * rddir_p, char * fn, uint32_t fn_len);                    <========== NO GOOD
    lv_fs_res_t (*dir_close_cb)(lv_fs_drv_t * drv, void * rddir_p);                                                          <========== NO GOOD

    void * user_data; /**< Custom file user data*/
};
```

There are no functions to set any of those callbacks. The user data is also not passed to any of the callback functions either.

src/core/lv_global.h
```diff
@@ -31,14 +31,25 @@ extern "C" {
...
-#include "../font/lv_font_fmt_txt.h"
+#include "../font/lv_font_fmt_txt_private.h"
 #endif
 
 #include "../tick/lv_tick.h"
 #include "../layouts/lv_layout.h"
 
 #include "../misc/lv_types.h"
 
+#include "../misc/lv_timer_private.h"
```
kdschlosser, line 43: having all of the private headers included into a public header defeats the purpose of making private headers.

src/core/lv_group.h
```diff
@@ -22,7 +22,7 @@ extern "C" {
  *      DEFINES
  *********************/
 /*Predefined keys to control the focused object via lv_group_send(group, c)*/
-enum _lv_key_t {
+enum {
```
kdschlosser: removing the names and making enumerations anonymous is going to break the documentation.

src/core/lv_group_private.h
```diff
@@ -0,0 +1,75 @@
...
+
+/**
+ * Groups can be used to logically hold objects so that they can be individually focused.
+ * They are NOT for laying out objects on a screen (try layouts for that).
+ */
+struct lv_group_t {
+    lv_ll_t obj_ll;        /**< Linked list to store the objects in the group*/
+    lv_obj_t ** obj_focus; /**< The object in focus*/
+
+    lv_group_focus_cb_t focus_cb;              /**< A function to call when a new object is focused (optional)*/
```
kdschlosser, line 35: you need to make sure that these function pointer types have a user_data parameter to them and that the user_data in this structure is getting passed to the callback. You also need to make sure that there are functions to set the callbacks because if there isn't there is not going to be a way to do it if they are private.

src/core/lv_obj.h
```diff
@@ -43,7 +43,7 @@ extern "C" {
  * Possible states of a widget.
  * OR-ed values are possible
  */
-enum _lv_state_t {
+enum {
```
kdschlosser, line 46: gonna mess up the documentation by removing the name

src/core/lv_obj.h
```diff
@@ -68,7 +68,7 @@ enum _lv_state_t {
  * Not all parts are used by every widget
  */
 
-enum _lv_part_t {
+enum {
```
kdschlosser, line 71: gonna mess up the documentation by removing the name

src/core/lv_obj.h
```diff
@@ -89,7 +89,7 @@ enum _lv_part_t {
  * Note: update obj flags corresponding properties below
  * whenever add/remove flags or change bit definition of flags.
  */
-typedef enum {
+enum {
```
kdschlosser: gonna mess up the documentation by removing the name

src/core/lv_obj_class_private.h
```diff
@@ -0,0 +1,72 @@
...
+
+/**********************
+ *      TYPEDEFS
+ **********************/
+
+/**
+ * Describe the common methods of every object.
+ * Similar to a C++ class.
+ */
+struct lv_obj_class_t {
```
kdschlosser, line 31: by making this class private you are removing the ability for people to make custom widgets.

src/core/lv_obj_draw_private.h
```diff
@@ -0,0 +1,48 @@
...
+/**********************
+ * GLOBAL PROTOTYPES
+ **********************/
+
+/**
+ * Get the extended draw area of an object.
+ * @param obj       pointer to an object
+ * @return          the size extended draw area around the real coordinates
+ */
+int32_t lv_obj_get_ext_draw_size(const lv_obj_t * obj);
```
kdschlosser, line 36: this should not be a private function due to possibly using it with a custom widget

src/core/lv_obj_draw_private.h
```diff
@@ -0,0 +1,48 @@
...
+
+lv_layer_type_t lv_obj_get_layer_type(const lv_obj_t * obj);
```
kdschlosser, line 38: this should not be a private function due to possibly using it with a custom widget

src/core/lv_obj_event_private.h
```diff
@@ -0,0 +1,62 @@
...
+ **********************/
+
+/**
+ * Used as the event parameter of ::LV_EVENT_HIT_TEST to check if an `point` can click the object or not.
+ * `res` should be set like this:
+ *   - If already set to `false` an other event wants that point non clickable. If you want to respect it leave it as `false` or set `true` to overwrite it.
+ *   - If already set `true` and `point` shouldn't be clickable set to `false`
+ *   - If already set to `true` you agree that `point` can click the object leave it as `true`
+ */
+struct lv_hit_test_info_t {
```
kdschlosser, line 34: this structure should not be made private as it is needed when performing a hit test

src/core/lv_obj_event_private.h
```diff
@@ -0,0 +1,62 @@
...
+    const lv_point_t * point;   /**< A point relative to screen to check if it can click the object or not*/
+    bool res;                   /**< true: `point` can click the object; false: it cannot*/
+};
+
+/**
+ * Used as the event parameter of ::LV_EVENT_COVER_CHECK to check if an area is covered by the object or not.
+ * In the event use `const lv_area_t * area = lv_event_get_cover_area(e)` to get the area to check
+ * and `lv_event_set_cover_res(e, res)` to set the result.
+ */
+struct lv_cover_check_info_t {
```
kdschlosser, line 44: this should not be private

src/core/lv_obj_style_private.h
```diff
@@ -0,0 +1,95 @@
...
+    uint32_t is_local : 1;
+    uint32_t is_trans : 1;
+};
+
+struct lv_obj_style_transition_dsc_t {
+    uint16_t time;
+    uint16_t delay;
+    lv_style_selector_t selector;
+    lv_style_prop_t prop;
+    lv_anim_path_cb_t path_cb;
```
kdschlosser, line 39: make sure this function pointer type has a user data parameter and also make sure there are functions to set the callback and the user data.

src/core/lv_obj_style_private.h
```diff
@@ -0,0 +1,95 @@
...
+
+/**
+ * Used internally to create a style transition
+ * @param obj
+ * @param part
+ * @param prev_state
+ * @param new_state
+ * @param tr
+ */
+void lv_obj_style_create_transition(lv_obj_t * obj, lv_part_t part, lv_state_t prev_state,
```
kdschlosser, line 68: I do not believe this is a private function

src/display/lv_display.c
```diff
@@ -966,6 +970,36 @@ void lv_display_rotate_area(lv_display_t * disp, lv_area_t * area)
     }
 }
 
+lv_obj_t * lv_screen_active(void)
+{
```
kdschlosser, line 970: all of the function definitions in this header need to be moved into the source file and only the function declarations should be left.

src/draw/lv_draw_buf.h
```diff
@@ -43,11 +43,11 @@ typedef struct {
  * Refine it to suit your needs.
  */
 
-#define _LV_DRAW_BUF_STRIDE(w, cf) \
+#define LV_DRAW_BUF_STRIDE(w, cf) \
```
kdschlosser, line 46: This macro should be a function, macros are not accessible from high level language bindings

src/draw/lv_draw_buf.h
```diff
@@ -43,11 +43,11 @@ typedef struct {
...
     LV_ROUND_UP(((w) * LV_COLOR_FORMAT_GET_BPP(cf) + 7) / 8, LV_DRAW_BUF_STRIDE_ALIGN)
 
 /* Allocate a slightly larger buffer, so we can adjust the start address to meet alignment */
-#define _LV_DRAW_BUF_SIZE(w, h, cf) (_LV_DRAW_BUF_STRIDE(w, cf) * (h) + LV_DRAW_BUF_ALIGN)
+#define LV_DRAW_BUF_SIZE(w, h, cf) (LV_DRAW_BUF_STRIDE(w, cf) * (h) + LV_DRAW_BUF_ALIGN)
```
kdschlosser: This macro should be a function, macros are not accessible from high level language bindings

src/draw/lv_draw_buf_private.h
```diff
@@ -0,0 +1,53 @@
...
+/*********************
+ *      DEFINES
+ *********************/
+
+/**********************
+ *      TYPEDEFS
+ **********************/
+
+struct lv_draw_buf_handlers_t {
+    lv_draw_buf_malloc_cb buf_malloc_cb;
```
kdschlosser, line 28: all of the function pointer type names are incorrect here. they should at a minimum have an `_t` on the end. If there is no `user_data` associated with these then the type names should end in `xcb_t`

src/draw/lv_draw_image.h
```diff
@@ -26,17 +26,7 @@ extern "C" {
...
- **********************/
-
-typedef struct {
-    lv_color_t alpha_color;
-    const lv_color32_t * palette;
-    uint32_t palette_size   : 9;
-} lv_draw_image_sup_t;
-
-typedef struct _lv_draw_image_dsc_t {
+typedef struct lv_draw_image_dsc_t {
```
kdschlosser, line 29: not sure why the struct name and the typedef name are the same or what purpose it serves.

src/draw/lv_draw_mask_private.h
```diff
@@ -0,0 +1,50 @@
...
+ *********************/
+
+/**********************
+ *      TYPEDEFS
+ **********************/
+
+/**********************
+ *      TYPEDEFS
+ **********************/
+struct lv_draw_mask_rect_dsc_t {
```
kdschlosser, line 30: make sure there are functions to set these fields because making it private without having those functions makes the structure unusable to a binding.

src/draw/lv_draw_private.h
```diff
@@ -0,0 +1,141 @@
...
+
+/*********************
+ *      DEFINES
+ *********************/
+
+/**********************
+ *      TYPEDEFS
+ **********************/
+
+struct lv_draw_task_t {
```
kdschlosser, line 27: make sure all of these fields have functions that are able to set them. otherwise it will be unusable to bindings.

src/draw/lv_draw_private.h
```diff
@@ -0,0 +1,141 @@
...
+     * Set to which extent `preferred_draw_unit_id` is good at this task.
+     * 80: means 20% better (faster) than software rendering
+     * 100: the default value
+     * 110: means 10% worse (slower) than software rendering
+     */
+    uint8_t preference_score;
+
+};
+
+struct lv_draw_mask_t {
```
kdschlosser, line 76: same as last review

src/draw/lv_draw_private.h
```diff
@@ -0,0 +1,141 @@
...
+    void * user_data;
+};
+
+struct lv_draw_unit_t {
```
kdschlosser, line 80: same as 2 reviews back

src/draw/lv_draw_vector_private.h
```diff
@@ -0,0 +1,104 @@
...
+#endif
+
+/*********************
+ *      INCLUDES
+ *********************/
+
+#include "lv_draw_vector.h"
+
+#if LV_USE_VECTOR_GRAPHIC
+
```
kdschlosser, line 20: All of the structures in this file need to have functions to set each and every single field otherwise they will not be usable to a binding.

src/draw/lv_image_dsc.h
```diff
@@ -29,7 +29,7 @@ LV_EXPORT_CONST_INT(LV_IMAGE_HEADER_MAGIC);
  *      TYPEDEFS
  **********************/
 
-typedef enum _lv_image_flags_t {
+typedef enum lv_image_flags_t {
```
kdschlosser, line 32: don't need the enum name here, only the typedef name

src/layouts/flex/lv_flex.h
liamHowatt: Hmm, I agree but the .c file depends on these defines. And I can't put them in a private header because this public header needs them to define the enum.

kdschlosser:
It might be better to tackle this animal one header file at a time. do one header and submit a PR for it and make sure that everything is good to go with it then go to the next header. 

I know that's going to be a real huge ball buster to do but it is the only way I think that it will get done properly..

@liamHowatt 

If you would like we can make an addon script that will monkey patch the gen_json script to collect the header file names that the different pieces of LVGL go into and we can have it sort the output based on the filenames.

Sorting it like that we would be able to quickly see what is happening for each file and we will be able to see what is happening with things like the structures and if proper functions are in place in order to move structures into a private header file. it would defiantly make it easier to see the things grouped together like that to be able to determine the work that needs to be done.

we can also parse the JSON output and check the structures and if a structure is not currently being used as a type for a parameter in a function then we can flag it for being private.

kisvegabor:
>     #define LV_GLOBAL_CUSTOM() ((lv_global_t*)mp_lv_roots)

What about `((void*)mp_lv_roots)` 

> If we include the private headers into the build the size of the firmware is going to become a lot larger. This is due to the number of additional structures that are going to get added resulting in a large number of qstrings being added.

The goal of this update to avoid it. We need to include `lvgl_private.h` only in `lv_mpy.c` but not when the binding is generated. 


> the other thing you have to keep a really close eye out for is when you make structures private and they hold callback functions if the type for the function pointer is not defined as having the user_data passed as the last parameter then it tests to see if the first parameter is a structure and if it is then it looks for a user_data field in that structure. If the structure has been made private it is not going to be able to locate the user_data inside the structure.

`lv_display_private.h` and `lv_indev_private.h` was there before this PR too, so I think conceptually it should work.

> There are no functions to set any of those callbacks. The user data is also not passed to any of the callback functions either.

If there is no API to set the fields the struct must remain public. `lv_fs_drv_t` is really public. See [here](https://github.com/lvgl/lvgl/pull/6068/files#diff-f84671453e78244b64722b825fe99c19c2e6145276d520aaf62900ef3252a155R69).

src/core/lv_group.h
kisvegabor: I agree that it should be `typedef enum{} lv_key_t;`

kisvegabor: I agree that it should be `typedef enum{} lv_key_t`;

src/core/lv_group_private.h
kisvegabor: There is a setter for this, but this is unrelated to the changes in this PR.

src/core/lv_obj_class_private.h
kisvegabor: New widgets can be created in C. I'm not sure it's something that we should support in the bindings. In MP with classes new widgets can be handled than in our C++-ish C OOP.

src/core/lv_obj_style_private.h
kisvegabor: It's used only internally. The user shouldn't deal with it.

src/display/lv_display.c
kisvegabor: > all of the function definitions in this header

It's `display.c` so a source file, not header.

src/draw/lv_draw_buf.h
kisvegabor: It's a macro to allow allocating array with its return value in compile time. E.g. `static uint8_t buf[LV_DRAW_BUF_STRIDE(100, LV_COLOR_FORMAT_ARGB8888) * 20]`. I believe we cannot do such a thing in MP.

src/draw/lv_draw_buf_private.h
kisvegabor: True, but it's independent of this PR. Let treat it separately.

src/draw/lv_draw_mask_private.h
kisvegabor: It's internal only.

src/draw/lv_draw_private.h
kisvegabor: We have added the required API.

src/draw/lv_draw_private.h
kisvegabor: Internal only.

src/draw/lv_draw_private.h
kisvegabor: Not needed to created new render units in the binding.

src/draw/lv_draw_buf.h
kisvegabor: Same as above.

src/core/lv_obj_event_private.h
kisvegabor: It's not used directly but only via `lv_event_set_cover_res`

src/core/lv_obj_class_private.h
kisvegabor: Let's discuss creating new MP widgets separately too.

src/core/lv_obj_draw_private.h
kisvegabor: See above.

kisvegabor:
My comments in summary:
- We don't need to solve the issues not related to adding private headers.
- Let's treat creating a new widget type in MP as a separate topic
- All enum-s should have `typedef`
- Invcluding `lvgl_private.h` in `lv_mpy.c` won't change the size, but we really shouldn't generate binding from it.

kdschlosser:
I would really love it if all function pointers were prototyped and had parameter names.

kisvegabor:
I think 2 main things are missing in this PR:
- All enums should be `typedef`-d: @liamHowatt could you take look at it?
- The MP build should work again: I hope we just need to add `lvgl_private.h` in `lv_mpy.c`. @kdschlosser could you try it out?

Anything else (naming fields, new, missing APIs, custom widgets?) shall be discussed in a separate issue.

kdschlosser:
@kisvegabor 


Yeah give me a bit to do it.

liamHowatt:
I typedef'd as many enums as I could, but I left them for a reason, originally.

- `lv_part_t` and `lv_state_t` are typedef'd in lv_types.h to avoid circular dependencies. [See this](https://github.com/lvgl/lvgl/pull/5161#issuecomment-1875056420). lv_obj.h includes many lv_obj_xxx.h. Those lv_obj_xxx.h cannot include lv_obj.h to get the `lv_part_t` type. lv_obj_style.h, for example, says `lv_part_t` is undefined even when I add the include for lv_obj.h (which itself includes lv_obj_style.h). If I understand correctly, it's because the include guards prevent the contents of the header from being expanded more than once per compilation unit. It goes C file -> lv_obj.h -> lv_obj_style.h -> ~~lv_obj.h~~. Therefore lv_obj_style.h cannot see inside lv_obj.h because its guard is already latched.
- `lv_opa_t` is concretely treated as `uint8_t` in many places. [See here](https://github.com/lvgl/lvgl/blob/dfe7bd5563a7cf9a86ab8de33c6ec55d4c363d8b/src/draw/sw/blend/lv_draw_sw_blend_to_rgb565.c#L185). It's treated as an alias for `uint8_t` and buffers of bytes are assigned to `lv_opa_t *` buffers interchangeably with `uint8_t *` buffers.
- `lv_style_prop_t` because of the size optimizations that benefit from it being a `uint8_t` [here](https://github.com/lvgl/lvgl/blob/dfe7bd5563a7cf9a86ab8de33c6ec55d4c363d8b/src/misc/lv_style.c#L242) and [here](https://github.com/lvgl/lvgl/blob/dfe7bd5563a7cf9a86ab8de33c6ec55d4c363d8b/src/misc/lv_style.c#L287). @kisvegabor that looks like your optimization. I can rework it so that it compiles, and we can accept the slight memory increase. I already did something similar in my new commit [here](https://github.com/lvgl/lvgl/pull/6068/commits/efed42e6858e28b3bf9725f3cf2bcec0e6c55b8e#diff-1964caa407fbbf719ca3cd91996e37573dec510d6f48e88f8c22e1183fdc3915) with array overhead going 4X.

src/core/lv_group_private.h
kdschlosser: I was just quickly skimming over and noting things that should be checked to make sure everything is actually complete.

kdschlosser:
The only reason I could see why there would be a forward declaration of an enum is if the enum is defined in a private header. In those cases the enum should not be defined in the private header it should be done in a public one and the private header should be including the public one that has the enum in it.

Then the forward declaration doesn't need to exist at all.


Am I correct in that assessment?

kdschlosser:
One of the big things you need to keep in the forfront of your mind is the documentation and how things are going to link together in it.


So if you have the enums burried behind a private header and there is a forward declaration for the enum that is public what you end up with is documentation that points to an empty enumeration. EVEN if the enum is not defined in a private and it is defined in a public one you still end up with an empty enum that gets linked to in the documentation. doxygen doesn't make the connection between the forward declaration and the definition. For structures it is OK to bury the private ones in the private headers because the fields are not part of the public API. However doing this `void some_function(struct _lv_obj_t);` is not going to link to the typed forward declaration `lv_obj_t`.... You can do that for functions and fields in a structure that are defined in a private header but not in a public one.

kdschlosser:
while forward declarations can help with circular imports they are often times used backwards and that causes either an overly complex documentation or a broken documentation. The full enum should be defined in the header file that it applies to most, Only in the cases where a circular include happens should a forward declaration be used and that forward declaration needs to be in the source or header file where the include is being done that causes the circular issue. Doing that will reduce the number of oddities in the documentation instead of causing it to be something that is widespread across all of the documentation.

liamHowatt:
It's public headers including each other. It's lv_obj.h and lv_obj_style.h including each other.

kdschlosser:
what is in lv_obj.h that would need to include from  lv_obj_style.h??

liamHowatt:
> what is in lv_obj.h that would need to include from lv_obj_style.h??

lv_obj_style.h needs `lv_part_t` which is currently in lv_types.h as `typedef uint32_t lv_part_t;` but I'm trying to remove it and make the enum in lv_obj.h be `typedef enum {...} lv_part_t;`, which is what kisvegabor asked for all enums.

liamHowatt:
Maybe the `lv_part_t` enum AND its members could go in lv_types.h :raised_eyebrow:

kdschlosser:
You can do a forward declaration for it in either one of the headers and the actual enum in the other. The thing you need to check is where is it most commonly used... That is where it should reside. lv_part_t and lv_selector_t should really be apart of the same enum because they are OR'ed together. So with regards to the parts and the selectors it might not be the right thing to do in that case. Maybe creating a typedef for uint32_t that has some kind of a common name where it can be documented as to the 2 other enums being OR'ed together.

kdschlosser:
I am not able to submit a PR for the changes to micropython. It's because of how the binding is set up and it being a fork of the actual micropython repo. I am not able to have the 2 coexist on the same system and use my IDE to edit and submit a PR. The IDE gets all kinds of pissed off at me...


This is the change that needs to be made...


The gen_mpy script needs from line number 725 to line number 763 replaced with the following code.

```python

headers = args.input
for header in headers:
    if 'lvgl.h' in header:
        path, _ = os.path.split(header)
        if path and path != 'lvgl.h':
            path = os.path.join(path, 'src', 'lvgl_private.h')
        else:
            path = 'src/lvgl_private.h'
        
        headers.append(path)
        break


print ("""
/*
 * Auto-Generated file, DO NOT EDIT!
 *
 * Command line:
 * {cmd_line}
 *
 * Preprocessing command:
 * {pp_cmd}
 *
 * Generating Objects: {objs}
 */

/*
 * Mpy includes
 */

#include <stdlib.h>
#include <string.h>
#include "py/obj.h"
#include "py/objint.h"
#include "py/objstr.h"
#include "py/runtime.h"
#include "py/binary.h"
#include "py/objarray.h"
#include "py/objtype.h"
#include "py/objexcept.h"

/*
 * {module_name} includes
 */

{lv_headers}
""".format(
        module_name = module_name,
        cmd_line=' '.join(argv),
        pp_cmd=pp_cmd,
        objs=", ".join(['%s(%s)' % (objname, parent_obj_names[objname]) for objname in obj_names]),
        lv_headers='\n'.join('#include "%s"' % header for header in headers)))

```

That will add the lvgl private header file without exposing the parts deemed private.

kisvegabor:
Thanks Kevin, I've update MicroPython with the suggested changes, made this PR ready to review and fixed the conflicts. 
Now the MicroyPython [CI caught](https://github.com/lvgl/lvgl/actions/runs/9163465771/job/25192613471?pr=6068#step:14:398) really useful issues. 

Regarding the enums, it seems they are mostly okay. Let's accept that a few of them might not be ideal and discuss them separately. It will much easier to see whats going on in a smaller PR.

kisvegabor:
@liamHowatt from [MicroPython's ESP32 CI](https://github.com/lvgl/lvgl/actions/runs/9183573890/job/25254501839?pr=6068#step:10:2594) it seems `lv_draw_sw_rgb565_swap` is not public, but it should be. 

@kdschlosser In the Unix build there is an issue with the screenshot compare. Could you check it out?

kdschlosser:
something changed in LVGL that broke it . add the --DEBUG to the CI for the test and rerun the CI and it will tell us what is broken,

kdschlosser:
I just tested it and the tests are passing. I am not sure why there is an issue here. Try rerunning the CI again and see if it passes.

liamHowatt:
I reproduced it locally. The first 8 lines in the output were an exception instead of hex data
```
(Pdb) p test_data.result[:8]
[b'ERROR START', b'Traceback (most recent call last):', b'  File "<stdin>", line 37, in _wrapper', b'  File "<stdin>", line 185, in create_ui', b"SyntaxError: Cannot convert 'closure' to pointer!", b'', b'ERROR END', b'FRAME START']
```

Failing here: https://github.com/liamHowatt/lvgl/blob/cf7b46ea12d5bf4ce07c4be3781a0a7843bdcf22/tests/micropy_test/micropy.py#L185

Here is the lv_mpy.c generated binding for `add_observer_obj`. I'm not really sure why this is failing now. Does it look correct?
```c
/*
 * lvgl extension definition for:
 * lv_observer_t *lv_subject_add_observer_obj(lv_subject_t *subject, lv_observer_cb_t observer_cb, lv_obj_t *obj, void *user_data)
 */

STATIC mp_obj_t mp_lv_subject_add_observer_obj(size_t mp_n_args, const mp_obj_t *mp_args, void *lv_func_ptr)
{
    void *user_data = mp_to_ptr(mp_args[3]);
    lv_observer_cb_t observer_cb = mp_to_ptr(mp_args[1]);
    lv_obj_t *obj = mp_to_lv(mp_args[2]);
    lv_subject_t *subject = mp_write_ptr_lv_subject_t(mp_args[0]);
    lv_observer_t * _res = ((lv_observer_t *(*)(lv_subject_t *, lv_observer_cb_t, lv_obj_t *, void *))lv_func_ptr)(subject, observer_cb, obj, user_data);
    return mp_read_ptr_lv_observer_t((void*)_res);
}
```

kdschlosser:
Here is the last PR that modified the observer header file. 

https://github.com/liamHowatt/lvgl/blob/cf7b46ea12d5bf4ce07c4be3781a0a7843bdcf22/tests/micropy_test/micropy.py#L185

I suspect it has something to do with changes made in there and in the source file.

kisvegabor:
It really a big update and I would like to be sure everyone is heading in the same direction so that we can work on it effectively. This is the summary of my understanding and vision:
- This is an important PR as it 
   - defines what we consider public API (everything which comes with `lvgl.h`)
   - makes the public API smaller and cleaner
- The concept of having private headers should work in general as we had `lv_display/indev_private.h` in v9.0 too. Probably there are special cases which makes the binding generation fail
- We see that there are issues in Micropython binding (missing arguments, not supported function signatures, etc). The goal in this PR is to make the binding at least as good as it was before. Further improvement can be added in separate PRs to keep things easier to manage.     
- The same applies to the docs. Some `enum`s might be missing, but they were missing previously too. The docs erros shall be fixed in separate PRs.

I don't know much about the binding generator script, and it would be quite slow for me to track the issue. @liamHowatt already has some experience with it, but clearly @kdschlosser is the best in it. 

@liamHowatt mentioned me that he suspects that the issue happens because the `user_data` field of the struct is not visible but the binding relies on it.  If really this the root of the problem, the generator shall be updated based on the new API of LVGL, in which we have `lv_xxx_set/get_user_data()` for each modul and `user_data` doesn't have to be accessed directly. @kdschlosser  Can you confirm this theory?

kdschlosser:
The userdata MUST either be visible as a field in the first parameter of the callback function. So that means the 1st parameter MUST be a structure. OR the user data needs to passed as the last parameter to the callback function. Those are the 2 ways it works.

The reason why the user data needs to be visible is because that is where the python function gets stored. LVGL has no idea of how to make a call to a python function so instead what has to happen is a C function is what gets passed as the callback function and the python function gets stored in the user data so when the c function gets called inside the c function is code to collect the python function from the user data and the code to execute that call to the python function is made. 


As it sits right now I am not able to compile LVGL in MicroPython unless I include the private header files. I get a bucket load of incomplete typedef errors.

liamHowatt:
Can the generator be allowed to see all the private definitions for its own generation needs, and just not generate any Micropython interfaces to private definitions in the output?

kdschlosser:
you need to explain that a little bit more in detail.. 

The JSON generator doesn't care if something is private or not it just reads the header files and outputs the data it is told to. If you do not include the private header files the generator is not able to make a connection between the forward declaration and the definition so what you end up with is something that is empty. as an example if you do a forward declaration of a struct that struct is going to be empty unless the definition of the struct is read from the private header file. This is how PyCParser works. It is only going to process what is available from the compiler preprocess output.

liamHowatt:
For the current issue of callback functions not being able to be generated, can the generator look at the private definitions to figure out how to generate the callback w/ the user data etc. and then just keep track of what came from a private header and not generate bindings for those?

kdschlosser:
You can declare a function prototype more than once if needs be. Say you have a structure that is private that you need a function prototype for and you need to have the same function prototype also used for public functions. If you include the public header into the private header you end up with a circular include for some reason. To handle that declare the prototype in both places.

I believe this is already being done in some places. You have to go and look.

liamHowatt:
Would it be possible to do something like this pseudocode:

```py
public_ast = pycparser("lvgl.h")
all_ast = pycparser("lvgl_all.h")

for symb in public_ast:
    if isinstance(symb, Func):
        gen_func_binding(symb)

def gen_func_binding(func_symb):
    # use data from `all_ast` to generate the symbol
```

Where `lvgl_all.h` is:
```c
#include "lvgl.h"
#include "lvgl_private.h"
```

kdschlosser:
IDK if you have updated the gen_json script with the PR for it. I made a bunch of updates to it including some additional command line parameters.


I believe what you would want to do is create the header file with both lvhl.h and lvgl_private.h included in that header. Then run the gen_json.py script passing the custom header file for it to use. Then run the script again adding the `--filter-private` command line switch. This is going to produce JSON output of ONLY the things that are in the private header files and nothing else. Then what you would do is you iterate over the private stuff and locate the things that have function prototypes that are being used and remove them from that output. Then remove what is remaining from the first JSON output. That will leave only the public stuff and anything that has a function prototype that is in private headers..

kdschlosser:
That is about the easiest way of doing what you are wanting to do. Doing that is going to need a complete rewrite of the entire gen_mpy.py script. Which really needs to be done. The reason why doing that is in a holding pattern is because the gen_json is still up in the air on adding it. so things have remained in limbo on that front. You would also need to get the current maintainer of the binding to do the work that way they know the code.

kdschlosser:
While I believe I have the best grasp on how the binding code generation works I am not the maintainer of it. That task falls one someone else and that is the person that needs to handle updates to the gen_mpy script.

kisvegabor:
@liamHowatt Can you try out your suggestion within a reasonable time?

liamHowatt:
I thought you were the maintainer of the micropython binding for some reason haha. Ah so the JSON generator is supposed to be a step in the right direction for the future of docs generation and helping the micropython binding generation and future bindings.

> @liamHowatt Can you try out your suggestion within a reasonable time?

It's not actually as complex as I thought, but there's a lot of coupling and it will be a big deal to update as @kdschlosser said, now that I'm looking through it. When it goes to get struct information to generate something, it also emits the binding for said struct.

@amirgon, can you grasp what we're trying to do and offer a solution? Temporary solutions are welcome. No pressure. Thanks in advance.

Otherwise I'll keep chipping away and see if I can manage it.

kdschlosser:
I did want to make sure that you name all public enums.

This is NOT naming the enum

```c
typedef enum {
    ...
} some_enum;
```

That is creating a typedef to an anonymous enum. 

This is a named enum

```c
typedef enum _some_enum {
    ...
} some_enum;
```

This is the one use case where having the name prefixed with an underscore should be allowed. naming both the enum and the typedef the same makes it tricky to deal with when using pycparser. It is better to have different names.


You can even forward declare it if needed.

```
enum _some_enum;
typedef _some_enum some_enum;
```


Best practices with enums is to do something along these lines.

```c
typedef enum _lv_enum_example {
    LV_ENUM_EXAMPLE_ITEM1,
    LV_ENUM_EXAMPLE_ITEM2
    LV_ENUM_EXAMPLE_ITEM3
} lv_enum_example;
```

The\ names for the enum items should be prefixed with the typedef or enum name. For the most part this is done, But there are some deviations from it that should be corrected.

kisvegabor:
As these are not introduced in this PR, let's handle these in a follow up PR. 

I think now we should focus on fixing the MP build issue to add more cleanup in smaller incremental steps.

kdschlosser:
OK so I want to bring this up here since you are doing work that is going to end up clashing with some of the work I am doing.

I am in the process of fixing the errors in the documentation build and one of the things I am running into is how things are done with lv_os and with draw...

In order to compile the documentation all of the header files need to be processed to their fullest. so with the draw header files there is the use of macros that are defined that make function calls but when you enable everything so all of the documentation gets built we end up with name collisions because the same macro is defined in more than one header file. The same thing holds true with functions as well.


macros shouldn't be used for this kind of a thing. there should only be a single function declaration and the functions should be defined in the source files and not in header files.

the document build system only reads the header files and not the source files. so I should be able to "turn everything on" and not end up with naming collisions because the macros should control what gets compiled in the source files and not in the header files.

the draw and os stuff needs to be rewritten so it all has a common API. structures can be extended by doing something like the following.

in lv_os_none.h you have this code..
```c
typedef int lv_mutex_t;
typedef int lv_thread_t;
typedef int lv_thread_sync_t;
```

and in lv_os_pthread.h you have this code.

```c
typedef struct {
    pthread_t thread;
    void (*callback)(void *);
    void * user_data;
} lv_thread_t;

typedef pthread_mutex_t lv_mutex_t;

typedef struct {
    pthread_mutex_t mutex;
    pthread_cond_t cond;
    bool v;
} lv_thread_sync_t;
```

This causes a naming collision when building the docs.


If we do the following..

lv_os.h
```c
typedef struct _lv_mutex_t {
    uint8_t id;  // doesn't matter what we do here just need something to declare a non empty structure.
} lv_mutex_t;

typedef struct _lv_thread_t {
    void (*callback)(void *);
    void * user_data;
} lv_thread_t;

typedef struct _lv_thread_sync_t {
    uint8_t id;
} lv_thread_sync_t;
```


and in lv_os_pthread.c we extend the structure like so.

```c
typedef struct {
    uint8_t id;
    pthread_mutex_t mutex;
} lv_mutex_pthread_t;

typedef struct {
    void (*callback)(void *);
    void * user_data;
    pthread_t thread;
} lv_thread_pthread_t;

typedef struct {
    uint8_t id;
    pthread_mutex_t mutex;
    pthread_cond_t cond;
    bool v;
} lv_thread_sync_pthread_t;
```


functions would be de3clared in lv_os.h like so.


```c
lv_result_t lv_thread_init(lv_thread_t * thread, lv_thread_prio_t prio, void (*callback)(void *), size_t stack_size, void * user_data);

lv_result_t lv_thread_delete(lv_thread_t * thread);
```

and this is what you would do in that function definition in lv_os_pthread.c

```c
lv_result_t lv_thread_init(lv_thread_t * thread, lv_thread_prio_t prio, void (*callback)(void *), size_t stack_size, void * user_data)
{
    LV_UNUSED(prio);
    
    lv_thread_pthread_t ret = (lv_thread_pthread_t)calloc(1, sizeof(lv_thread_pthread_t))
    
    pthread_attr_t attr;
    pthread_attr_init(&attr);
    pthread_attr_setstacksize(&attr, stack_size);
    ret.callback = callback;
    ret.user_data = user_data;
    pthread_create(&ret.thread, &attr, generic_callback, &ret);
    *thread = &ret;
    return LV_RESULT_OK;
}

lv_result_t lv_thread_delete(lv_thread_t * thread)
{
    lv_thread_pthread_t  *thrd = (lv_thread_pthread_t  *)thread;
    
    if (thrd->thread != NULL) {
        int ret = pthread_join(thrd->thread, NULL);
        if(ret != 0) {
            LV_LOG_WARN("Error: %d", ret);
            return LV_RESULT_INVALID;
        }
    
        thrd->thread = NULL;
        return LV_RESULT_OK;
    } else {
        return LV_RESULT_INVALID;
    }
}

```

so now everything has a common type that is able to be used so there are no name collisions.


this

```c
void (*callback)(void *)
```
would need to be typed properly and not declared in the structure and the function declarations. But it gives you the basic idea.

kisvegabor:
I see the problem. Let keep in mind for the next pass. For now, please enable only a single OS, and deal with this problem separately. Otherwise I'm afraid we will never never merge anything. 😄

kdschlosser:
I am mentioning it so I don't forget about it. The problem isn't so much enabling every OS because I am not able to do that. The problem stems from how portions of the document build system work and those parts being outside my control. Breathe which is the layer that translates doxygen to sphinx uses doxygen to build XML files and it does this one header file at a time.. There is no preprocessing that gets done when it does this so it takes everything that is seen in the header file exactly how it is. The only way I am able to control is it to not include specific header files from having documentation being generated for it. If the API was written to be common across all of the OS's then there would only be a need to read a single header file and that would work out fine....But because the types of some of the function parameters are not declared in the header file where the functions are declared a whole different set of errors occur because of the types not existing.

kdschlosser:
I also wanted to mention this so that the work that is being done here is done so it doesn't cause more of this kind of an issue in other portions of the LVGL code.

I hate to see work get done and then have to get done again because it caused an issue like the one I mentioned.

kisvegabor (dismissed):
It's passing now! :tada: 

It'd be very hard to review this in detail so please just check the concept from a high level point of view.

kisvegabor:
Ah, 2 conflicts... We need to be quick with reviewing. :slightly_smiling_face:

kisvegabor:
@liamHowatt This PR is getting out of sync very quickly. 😕
Please fix the conflicts and I'll merge after that. In the meantime @XuNeo or @FASTSHIFT please check if you have any principal  objections with this update.

kisvegabor:
@XuNeo @FASTSHIFT now is the time. I'll merge it an hour if there is no objection.

XuNeo (approved):
Thanks! This is a massive work.

I agree to merge it firstly(I haven't review all the changes). But I have concerns about it.

1. Developer's first sense is not to include private headers from third-party lib.
2. This PR definitely breaks lots of application
- API changes should add to map.h
- User may be forced to include private header.
3. Some of the `_` prefix does mean for intermediate usage. I accept them and we can sort them out in following PRs.

kisvegabor (approved):

kisvegabor:
Yes, probably this large change will have unexpected consequences that we can handle in small targeted PRs.

@liamHowatt Thank you so much for the persistence! 💯 
It was a way larger work than we anticipated. Lessons n learned (again) big refactoring should come in smaller chunks if possible.

kisvegabor:
@vwheeler63 FYI, this PR lowers the public symbols, but might mean conflicts in your current Doxygen work too. Please be sure to rebate to master.

kdschlosser:
I know this is already merged but it is still worth mentioning. Most times that I have seen include guards used the name typically begins and ends with either a single or a double underscore.

as an example...

```c
#ifndef __SOME_HEADER_FILE_H__
#define __SOME_HEADER_FILE_H__

#endif /* __SOME_HEADER_FILE_H__ */
```

It's a way of knowing what to ignore

vwheeler63:
> @vwheeler63 FYI, this PR lowers the public symbols, but might mean conflicts in your current Doxygen work too. Please be sure to rebate to master.

Hi, @kisvegabor !  Indeed, it created 2 conflicts.  Before I saw your message here I managed to resolve them the hard way.  \*\*laughing at self\*\*  And more or less accomplished the same results by merging `master` into the branch behind PR#6584.  Next time that happens I'll re-base -- a much "neater" solution.  :-)

Speaking of which, PR#6584 is back in a "ready" state again.

Kind regards,
Vic

PR #6073, opened 2024-04-15
fix(sysmon): fix MicroPython compilation error when system monitor is enabled

kisvegabor:

Continue #5936

Add support for multiple displays.

### Notes

### DIFF
--- a/src/display/lv_display_private.h
+++ b/src/display/lv_display_private.h
@@ -18,7 +18,7 @@ extern "C" {
 #include "../draw/lv_draw.h"
 #include "lv_display.h"
 
-#if defined(LV_USE_PERF_MONITOR) && LV_USE_PERF_MONITOR
+#if LV_USE_PERF_MONITOR
 #include "../others/sysmon/lv_sysmon.h"
 #endif
 
@@ -153,13 +153,13 @@ struct _lv_display_t {
     /** The area being refreshed*/
     lv_area_t refreshed_area;
 
-#if defined(LV_USE_PERF_MONITOR) && LV_USE_PERF_MONITOR
+#if LV_USE_PERF_MONITOR
     lv_obj_t * perf_label;
     lv_sysmon_backend_data_t perf_sysmon_backend;
     lv_sysmon_perf_info_t perf_sysmon_info;
 #endif
 
-#if defined(LV_USE_MEM_MONITOR) && LV_USE_MEM_MONITOR
+#if LV_USE_MEM_MONITOR
     lv_obj_t * mem_label;
 #endif
 
--- a/src/display/lv_display_private.h
+++ b/src/display/lv_display_private.h
@@ -18,7 +18,7 @@ extern "C" {
 #include "../draw/lv_draw.h"
 #include "lv_display.h"
 
-#if LV_USE_PERF_MONITOR
+#if defined(LV_USE_PERF_MONITOR) && LV_USE_PERF_MONITOR
 #include "../others/sysmon/lv_sysmon.h"
 #endif
 
@@ -153,13 +153,13 @@ struct _lv_display_t {
     /** The area being refreshed*/
     lv_area_t refreshed_area;
 
-#if LV_USE_PERF_MONITOR
+#if defined(LV_USE_PERF_MONITOR) && LV_USE_PERF_MONITOR
     lv_obj_t * perf_label;
     lv_sysmon_backend_data_t perf_sysmon_backend;
     lv_sysmon_perf_info_t perf_sysmon_info;
 #endif
 
-#if LV_USE_MEM_MONITOR
+#if defined(LV_USE_MEM_MONITOR) && LV_USE_MEM_MONITOR
     lv_obj_t * mem_label;
 #endif
 
--- a/src/display/lv_display_private.h
+++ b/src/display/lv_display_private.h
@@ -18,6 +18,10 @@ extern "C" {
 #include "../draw/lv_draw.h"
 #include "lv_display.h"
 
+#if LV_USE_PERF_MONITOR
+#include "../others/sysmon/lv_sysmon.h"
+#endif
+
 /*********************
  *      DEFINES
  *********************/
@@ -148,6 +152,17 @@ struct _lv_display_t {
 
     /** The area being refreshed*/
     lv_area_t refreshed_area;
+
+#if LV_USE_PERF_MONITOR
+    lv_obj_t * perf_label;
+    lv_sysmon_backend_data_t perf_sysmon_backend;
+    lv_sysmon_perf_info_t perf_sysmon_info;
+#endif
+
+#if LV_USE_MEM_MONITOR
+    lv_obj_t * mem_label;
+#endif
+
 };
 
 /**********************
--- a/src/display/lv_display_private.h
+++ b/src/display/lv_display_private.h
@@ -18,7 +18,7 @@ extern "C" {
 #include "../draw/lv_draw.h"
 #include "lv_display.h"
 
-#if LV_USE_PERF_MONITOR
+#if LV_USE_SYSMON
 #include "../others/sysmon/lv_sysmon.h"
 #endif
 
--- a/src/osal/lv_windows.c
+++ b/src/osal/lv_windows.c
@@ -197,6 +197,7 @@ lv_result_t lv_thread_sync_delete(lv_thread_sync_t * sync)
     return LV_RESULT_OK;
 }
 
+
 /**********************
  *   STATIC FUNCTIONS
  **********************/
--- a/src/osal/lv_windows.c
+++ b/src/osal/lv_windows.c
@@ -197,7 +197,6 @@ lv_result_t lv_thread_sync_delete(lv_thread_sync_t * sync)
     return LV_RESULT_OK;
 }
 
-
 /**********************
  *   STATIC FUNCTIONS
  **********************/
--- a/src/others/sysmon/lv_sysmon.c
+++ b/src/others/sysmon/lv_sysmon.c
@@ -23,17 +23,8 @@
 
 #define SYSMON_REFR_PERIOD_DEF 300 /* ms */
 
-#if defined(LV_USE_PERF_MONITOR) && LV_USE_PERF_MONITOR
-    #define _USE_PERF_MONITOR   1
-#else
-    #define _USE_PERF_MONITOR   0
-#endif
-
 #if defined(LV_USE_MEM_MONITOR) && LV_USE_MEM_MONITOR
     #define sysmon_mem LV_GLOBAL_DEFAULT()->sysmon_mem
-    #define _USE_MEM_MONITOR   1
-#else
-    #define _USE_MEM_MONITOR   0
 #endif
 
 /**********************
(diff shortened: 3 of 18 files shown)


kdschlosser:
I am not sure how the binding is going to react with lv_display_t being the first parameter type. I don't know what lv_conf.h file is being used but if it gets generated using the template then the sysmon is going to be disabled by default,.. It needs to be turned on. It doesn't need to be shown we just need to make sure it builds the code and compiles properly.

kisvegabor:
Could you try it out with MP to be sure its working? :slightly_smiling_face:

kdschlosser:
How is lv_conf.h built when the CI gets run? If it is turned off then it shouldn't be and it would need to be changed. It needs to be enabled and then code needs to be added to hide the framerate and the memory to the tests.

I do not know if it is turned on or not when the CI runs.... Do you know?

kisvegabor:
I've enabled in in the C CI/CD. See [de366cc](https://github.com/lvgl/lvgl/pull/6073/commits/de366cc4aa3acbebe4585dd50da0a5d2b8ed49c6)

I've hova no idea why the arc test needed to be updated to pass.

kdschlosser:
Force this PR to rerun the tests and see if it passes.

kisvegabor:
I tested the fix locally and forgot to push it...

kdschlosser (approved):

src/others/sysmon/lv_sysmon.c
```diff
@@ -15,16 +15,15 @@
...
  *********************/
-#define MY_CLASS (&lv_sysmon_class)
 
 #define SYSMON_REFR_PERIOD_DEF 300 /* ms */
 
 #if defined(LV_USE_PERF_MONITOR) && LV_USE_PERF_MONITOR
-    #define sysmon_perf LV_GLOBAL_DEFAULT()->sysmon_perf
     #define _USE_PERF_MONITOR   1
 #else
     #define _USE_PERF_MONITOR   0
```
XuNeo: Is it possible to modify `lvgl/scripts/lv_conf_internal_gen.py`,  to always define LV_USE_PERF_MONITOR even if `LV_USE_SYSMON` is 0? 

`LVGL`'s style is to check `#if LV_USE_PERF_MONITOR`, so it'd better define it from the first place.
```c
/*1: Enable system monitor component*/
#ifndef LV_USE_SYSMON
    #ifdef CONFIG_LV_USE_SYSMON
        #define LV_USE_SYSMON CONFIG_LV_USE_SYSMON
    #else
        #define LV_USE_SYSMON   0
    #endif
#endif

//////////////#if LV_USE_SYSMON
    /*1: Show the used memory and the memory fragmentation
     * Requires `LV_USE_STDLIB_MALLOC = LV_STDLIB_BUILTIN`
     * Requires `LV_USE_SYSMON = 1`*/
    #ifndef LV_USE_MEM_MONITOR
        #ifdef CONFIG_LV_USE_MEM_MONITOR
            #define LV_USE_MEM_MONITOR CONFIG_LV_USE_MEM_MONITOR
        #else
            #define LV_USE_MEM_MONITOR 0
        #endif
    #endif
    #if LV_USE_MEM_MONITOR
        #ifndef LV_USE_MEM_MONITOR_POS
            #ifdef CONFIG_LV_USE_MEM_MONITOR_POS
                #define LV_USE_MEM_MONITOR_POS CONFIG_LV_USE_MEM_MONITOR_POS
            #else
                #define LV_USE_MEM_MONITOR_POS LV_ALIGN_BOTTOM_LEFT
            #endif
        #endif
////////////#endif
```

XuNeo (dismissed):

src/others/sysmon/lv_sysmon.c
kdschlosser: having `LV_USE_SYSMON` in the `lv_conf.h` file is pretty useless and it should not be something the user sets. It's a simple test to see if either `LV_USE_MEM_MONITOR`  or `LV_USE_PERF_MONITOR` is set in order to set `LV_USE_SYSMON ` accordingly.


Something along the lines of..


```c
#ifndef LV_USE_SYSMON
    #if LV_USE_MEM_MONITOR || LV_USE_PERF_MONITOR
        #define LV_USE_SYSMON 1
    #else
        #define LV_USE_SYSMON 0
    #endif
#else
    #if LV_USE_MEM_MONITOR || LV_USE_PERF_MONITOR
        #undef LV_USE_SYSMON 
        #define LV_USE_SYSMON 1
    #endif
#endif
```


doing the above would allow the user to define the macro if they wanted to from the command line or they could add it to the config file if they wanted to turn on the system monitors and not have the  performance and memory monitors on.  If the performance or memory monitors are on then by default the system monitor should be enabled.

kdschlosser: I had also thought that when using the macros if the macro is not defined then doing the following would evaluate as `false`. 
```c
#if LV_USE_PERF_MONITOR
```

The only time that a macro needs to be tested to see if it is defined or not is to be used as a guard against setting it a second time which results in a compiler warning or error.


So that line of code could be done as

```
#if LV_USE_PERF_MONITOR
```

instead of 

```
#if defined(LV_USE_PERF_MONITOR) && LV_USE_PERF_MONITOR
```

because they actually evaluate the same.

kisvegabor: As @FASTSHIFT pointed out `sysmon` can be used by the user too, so there should be a way to enable it independently.

I've update `lv_conf_internal_gen.py`.

src/display/lv_display_private.h
```diff
@@ -18,6 +18,10 @@ extern "C" {
 #include "../draw/lv_draw.h"
 #include "lv_display.h"
 
+#if LV_USE_PERF_MONITOR
```
kdschlosser: I believe this should read 


```c
#if LV_USE_SYSMON
```


The reason why I say that is because of either the perf monitor or the mem monitor is enabled then the file sill be included. This is because later on in this file both the memory monitor and also the perf monitor are check to see if they are enabled.

src/others/sysmon/lv_sysmon.c
```diff
@@ -15,26 +15,16 @@
...
 #include "../../stdlib/lv_string.h"
 #include "../../widgets/label/lv_label.h"
+#include "../../display/lv_display_private.h"
 
 /*********************
  *      DEFINES
  *********************/
-#define MY_CLASS (&lv_sysmon_class)
 
 #define SYSMON_REFR_PERIOD_DEF 300 /* ms */
```
kdschlosser: This should be

```c
#ifndef SYSMON_REFR_PERIOD_DEF 
#define SYSMON_REFR_PERIOD_DEF 300 /* ms */
#endif


```

src/others/sysmon/lv_sysmon.c
```diff
@@ -15,26 +15,16 @@
...
 
-#if defined(LV_USE_PERF_MONITOR) && LV_USE_PERF_MONITOR
-    #define sysmon_perf LV_GLOBAL_DEFAULT()->sysmon_perf
-    #define _USE_PERF_MONITOR   1
-#else
-    #define _USE_PERF_MONITOR   0
-#endif
-
 #if defined(LV_USE_MEM_MONITOR) && LV_USE_MEM_MONITOR
```
kdschlosser: This line should read

```c
#if LV_USE_MEM_MONITOR
```

src/others/sysmon/lv_sysmon.c
```diff
@@ -84,19 +70,16 @@ void _lv_sysmon_builtin_init(void)
...
     lv_timer_delete(sysmon_mem.timer);
 #endif
 }
 
-lv_obj_t * lv_sysmon_create(lv_obj_t * parent)
+lv_obj_t * lv_sysmon_create(lv_display_t * disp)
 {
     LV_LOG_INFO("begin");
-    lv_obj_t * label = lv_label_create(parent);
+    if(disp == NULL) disp = lv_display_get_default();
```
kdschlosser, line 82: There needs to be a second check to see if the display is NULL. If it is then an error should be made to the log and the function should exit.

```c
if(disp == NULL) {
    disp = lv_display_get_default();
    if(disp == NULL) {
        // log error
        return NULL;
    }
}
```

I thought that display_get_default can return a NULL pointer if there is no display initialized. If I am right about that then this will need to be done.

src/others/sysmon/lv_sysmon.c
```diff
@@ -105,16 +88,64 @@ lv_obj_t * lv_sysmon_create(lv_obj_t * parent)
     return label;
 }
 
+#if LV_USE_PERF_MONITOR
+
+void lv_sysmon_show_performance(lv_display_t * disp)
+{
+    if(disp == NULL) disp = lv_display_get_default();
```
kdschlosser, line 101: This is the same thing as the last comment with the NULL pointer.

kdschlosser (changes requested):

src/others/sysmon/lv_sysmon.c
kdschlosser: I don't know why it has marked the old code instead of the new code. But it has. That's annoying.

kisvegabor: I wasn't planning planning to expose this to keep lv_conf.h smaller, but we can have the guard here, so if someone looks in to the code, he will see that it can be overwritten.

kisvegabor:
@kdschlosser All are good points. I've pushed fixed for them.

kdschlosser (approved):

kisvegabor:
Passing now. Just need 2 reviews to merge it:slightly_smiling_face:

src/display/lv_display_private.h
XuNeo:
```suggestion
#if LV_USE_SYSMON
```

src/osal/lv_windows.c
```diff
@@ -197,6 +197,7 @@ lv_result_t lv_thread_sync_delete(lv_thread_sync_t * sync)
     return LV_RESULT_OK;
 }
 
+
```
XuNeo:
```suggestion
```

src/display/lv_display_private.h
XuNeo: I think it's true here we should use LV_USE_SYSMON. Please revert if not so. :)

XuNeo (dismissed):

kisvegabor:
Thanks Neo! Although I don't understand how you could approve the PR when you added the last changes. 🤨

src/others/sysmon/lv_sysmon.c
```diff
@@ -105,16 +94,84 @@ lv_obj_t * lv_sysmon_create(lv_obj_t * parent)
...
+    if(disp == NULL) {
+        LV_LOG_WARN("There is no default display");
+        return;
+    }
+
+    disp->perf_label = lv_sysmon_create(disp);
```
FASTSHIFT, line 107: Does calling `lv_sysmon_show_performance` twice cause a memory leak?

src/others/sysmon/lv_sysmon.c
```diff
@@ -105,16 +94,84 @@ lv_obj_t * lv_sysmon_create(lv_obj_t * parent)
...
+
+void lv_sysmon_hide_performance(lv_display_t * disp)
+{
+    if(disp == NULL) disp = lv_display_get_default();
+    if(disp == NULL) {
+        LV_LOG_WARN("There is no default display");
+        return;
+    }
+
+    lv_obj_add_flag(disp->perf_label, LV_OBJ_FLAG_HIDDEN);
```
FASTSHIFT, line 134: if(disp->perf_label == NULL) {
    return;
}

kisvegabor: Fixed

src/others/sysmon/lv_sysmon.c
kisvegabor: Fixed

FASTSHIFT (approved):

PGNetHun (approved):

PR #6093, opened 2024-04-19
feat(chart): patch for adding space between newest data and oldest data in charts

vahidajalluian:

This patch adds space between the newest and oldest data available on the chart in line circular mode.

### DIFF
--- a/src/widgets/chart/lv_chart.c
+++ b/src/widgets/chart/lv_chart.c
@@ -881,8 +881,8 @@ static void draw_series_line(lv_obj_t * obj, lv_layer_t * layer)
                         }
                     }
                     else if(ser->y_points[p_prev] == LV_CHART_POINT_NONE && ser->y_points[p_act] != LV_CHART_POINT_NONE){
-                    	y_min = p2.y;
-                    	y_max = p2.y;
+                    	y_min = line_dsc.p2.y;
+                    	y_max = line_dsc.p2.y;
                     }
                 }
                 else {
--- a/src/widgets/chart/lv_chart.c
+++ b/src/widgets/chart/lv_chart.c
@@ -462,7 +462,7 @@ void lv_chart_set_next_value(lv_obj_t * obj, lv_chart_series_t * ser, int32_t va
 
     lv_chart_t * chart  = (lv_chart_t *)obj;
     ser->y_points[ser->start_point] = value;
-    // If optimization isn't enabled or chart data series aren't updated synchronously, this part makes disturbances.  
+// If optimization isn't enabled or chart data series aren't updated synchronously, this part makes disturbances.  
     for(int32_t i = 1; i < 10; i++) {
         if(ser->start_point + i < chart->point_cnt)
             ser->y_points[ser->start_point + i] = LV_CHART_POINT_NONE;
--- a/src/widgets/chart/lv_chart.c
+++ b/src/widgets/chart/lv_chart.c
@@ -455,7 +455,7 @@ void lv_chart_set_all_value(lv_obj_t * obj, lv_chart_series_t * ser, int32_t val
     lv_chart_refresh(obj);
 }
 
-void lv_chart_set_next_value(lv_obj_t * obj, lv_chart_series_t * ser, int32_t value,int32_t space_len)
+void lv_chart_set_next_value(lv_obj_t * obj, lv_chart_series_t * ser, int32_t value)
 {
     LV_ASSERT_OBJ(obj, MY_CLASS);
     LV_ASSERT_NULL(ser);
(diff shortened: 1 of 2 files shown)


vahidajalluian:
About this talk in the forum 
https://forum.lvgl.io/t/add-space-in-chart-in-circular-mode/6818/12

kisvegabor:
Hello,

Form the [log of the CI](https://github.com/lvgl/lvgl/actions/runs/8750318486/job/24013621689?pr=6093) it seems the widgets demo crashes with these changes. Could you try it out?

vahidajalluian:
@kisvegabor Well, it sends a "Screenshot compare error" and this is to happen. Since now the graph made has different view from past ones. this one does have spaces between new data and old data, however past implementations didn't do this task. So the screenshots should change. and I don't know with what picture or image is this output compared.
/home/runner/work/lvgl/lvgl/tests/src/test_cases/widgets/test_chart.c:183:test_draw_task_hooking:INFO: 
```
Screenshot compare error
  - File: ref_imgs/widgets/chart_line_draw_hook.png
  - Mode: full refresh
  - At x:102, y:120.
  - Expected: 0x003643F4
  - Actual:   0x00FFFFFF
/home/runner/work/lvgl/lvgl/tests/src/test_cases/widgets/test_chart.c:189:test_draw_task_hooking:FAIL: widgets/chart_line_draw_hook.png
```

kisvegabor:
I've noticed the crash [here](https://github.com/lvgl/lvgl/actions/runs/8750318486/job/24013621689?pr=6093#step:6:1275).

vahidajalluian:
@kisvegabor 
I had some changes, added check to do spacing only if it is line chart. Would you please check it again?

vahidajalluian:
@kisvegabor All tests have passed. It is better to change chart example 8 of version 9 now (only omitting that part which is adding NONE data ) since that process is inherent in setting the next data.

src/widgets/chart/lv_chart.c
```diff
@@ -462,8 +462,19 @@ void lv_chart_set_next_value(lv_obj_t * obj, lv_chart_series_t * ser, int32_t va
 
     lv_chart_t * chart  = (lv_chart_t *)obj;
     ser->y_points[ser->start_point] = value;
+    // If optimization isn't enabled or chart data series aren't updated synchronously, this part makes disturbances.
```
kisvegabor:
```suggestion
    /* If optimization isn't enabled or chart data series aren't updated synchronously, this part makes disturbances.*/
```

src/widgets/chart/lv_chart.c
```diff
@@ -462,8 +462,19 @@ void lv_chart_set_next_value(lv_obj_t * obj, lv_chart_series_t * ser, int32_t va
...
+    if(chart->type == LV_CHART_TYPE_LINE && chart->update_mode == LV_CHART_UPDATE_MODE_CIRCULAR) {
+        for(int32_t i = 1; i < 10; i++) {
```
kisvegabor, line 467: We shouldn't do this be default especially not with a hard coded value. 

We could have a dedicated function to set `N` values to an `int` or copy from an array without changing the `start_point` bit is feels too special.

What about adding a function for deleting `N` values in `lv_example_chart_8()`?

kisvegabor (changes requested):

src/widgets/chart/lv_chart.c
vahidajalluian: @kisvegabor Well, you know that I am a newbie. And this change was just a suggestion. You are more familiar with this library. The problem I had encountered is that on the implementation on hardware, if one wants to use this feature it is really hard. Moreover, this chart doesn't provide smooth drawing in crowded mode but only vertical lines. so in case of smooth signals like sinus, it isn't really plotted as smooth as sin (especially when the space or area is not much.). more description is that I had tried different methods, for example to add some points then spaces and then redraw chart (in different functions). This resulted in that if the number of data added is less than the space, then the graph drawing would become piecewise rather being continuous. Of course I know that implementing this widget efficiently is really hard, especially when you want to maintain graph smoothness.

kisvegabor: > Well, you know that I am a newbie. 

No worries. I didn't want to be offensive. Sorry, if it sounded like that.

---

I have 2 ideas for the chart but these should be added in a separate PR:
- Make crowded mode a new type for better control, e.g. `LV_CHART_TYPE_LINE_CORWDED`
- Also add `LV_CHART_TYPE_LINE_SMOOTH`. We can use our build in vector engine and connect the points with bezier curves.

vahidajalluian: This is beyond my knowledge of this library. I will try to dig in more in this library. I think I should start with the implementation of the core. But in this special case, I think we can use the idea behind the qcustomplot of QT or the qtcharts, since they have the smoothing option, and I don't know if they use the bezier or spline method or any other method. I don't know if there is any other open-source chart or graph library available.

kisvegabor: Sorry, I missed your comment :slightly_frowning_face: 

We don't need to deal with easing now. 

I think what we should in this PR:
- Keep the changes in `draw_series_line`
- Add a function for deleting `N` values in `lv_example_chart_8()` (this function can be a reference for the users)

kisvegabor:
I close this PR because there was no activity here for a while.

If you have remarks about this topic feel free to comment here and we can reopen it.

PR #6105, opened 2024-04-22
Updated BIDI text support to include persian and arabic

Maldus512:

I need to add arabic support to a legacy application that uses LVGL v6.0; I realized that the internal function to distinguish between LTR and RTL text does not recognize arabic/persian characters as RTL. I added the feature as it is found in the latest versions.


kisvegabor:
Thank you!

PR #6115, opened 2024-04-23
fix(cmake): add defines for building examples and demos if specified in CMakePresets.json

ghost:
This PR will generate the proper defines for building examples and demos if specified in cmake presets.json


kisvegabor:
Hi,

So does `CONFIG_LV_BUILD_EXAMPLES/DEMOS` come from `CMakePreset.json`? If so, do they need to be added by the user manually?

Ju1He1:
Hey @kisvegabor 
I just noticed that custom.cmake uses the cmake variables

- LV_CONF_BUILD_DISABLE_EXAMPLES
- LV_CONF_BUILD_DISABLE_DEMOS

while esp.cmake uses
- CONFIG_LV_USE_DEMO_WIDGETS
- CONFIG_LV_BUILD_EXAMPLES
- CONFIG_LV_USE_DEMO_MUSIC


IMHO we should use the same. Which way do you prefer? The first  `if NOT ..._DISABLE...`
or the second where the user would have to explicitly activate building demos and examples.


Either way, if somebody would like to build examples and demos the appropriate defines (`LV_USE_DEMO_WIDGETS=1 ` and `LV_BUILD_EXAMPLES=1`)  should be automatically be set

kisvegabor:
In `lv_conf.h` we follow the "positive logic" the use or enable something, instead of disabling something.  So I think we shoul follow the same here too. Just two questions:
1. Can we add backward compatibility for `LV_CONF_BUILD_DISABLE_EXAMPLES/DEMOS`?
2. Could you document these config options [here](https://docs.lvgl.io/master/integration/building/cmake.html)?

CMakePresets.json
```diff
@@ -5,7 +5,9 @@
       "name": "_base",
       "hidden": true,
       "cacheVariables": {
-        "LV_CONF_SKIP": true
+        "LV_CONF_SKIP": true,
```
kisvegabor, line 8: Why do we want to enable it by default?

kisvegabor (commented):
Thank you, it looks good, just one question.

CMakePresets.json
ghost: Hey @kisvegabor thx for your answer this is just for the default setup to be running without any issues. The user will probably remove it.
If i omit this option i cannot build lvgl directly outof the box because
`#include "lv_conf.h"` fails. Since we do not provide dummy config files for windows, linux etc. we must skip the config.
(or is there another way?)

kisvegabor: By default LVGL assumes `lv_conf.h` is placed in the same folder as `lvgl`. It provides a configuration free setting by default. Can we do it with `CMakePresets.json` too?

ghost: sorry there might be an misunderstanding. The goal of this preset is to enable the user to build this project with MSVC without having to make any changes. I can use an existing lv_conf.h but I cannot find any lv_conf.h within the repo. (And I'm guessing there is none since lv_conf.h is application depended)

IMHO if we enforce the user to create an lv_conf.h first before building this would make first steps with lvgl much more complicated. 
=> IMHO goal should be that any MSVC user could clone this repo as it is and build it without having to make any changes. 
Or am I mistaken here?

Obviously when integrating you would like to use your own lv_conf.h.
Usually when working with CmakePresets you will create your own CmakePresets.json within your project which will overwrite the Presets from the lvgl repo. 
=> The presets from the lvgl repo are only intended as a quick start guide and will not be used in user projects.

To make things even more clear usally your project structure looks like this

- my_repo
-      CMakePresets.json
-      CMakeLists.txt
-      main.c
- lvgl
-       CMakepresets.json
-       lvgl sources

To build this via console you need to navigate to my_repo and invoke the presets
```
cd my_repo
cmake --preset my_project_configure_preset
cmake --build --preset my_project_build_preset
```
(or tell your IDE to use the CmakePresets from my_project)

as you can see the CmakePresets.json from the lvgl repo is not invoked at all :)

Did i make things more clear?

kisvegabor: Thank you for the detailed description! The goal is indeed to make getting started as simple as possible. Normally we create ready to use projects like [these](https://github.com/lvgl?q=lv_port_pc&type=all&language=&sort=)

> The goal of this preset is to enable the user to build this project with MSVC without having to make any changes. 

However, user still need to create a `main.c` which is way less trivial than creating `lv_conf.h`.

ghost: hey, alright more clarification:)
You can build the lvgl repo now as a standlone library without having to make any changes or adding a main.c.

However in most cases you would like to build an executable (hence the main.c). 
In fact if you don't want to create an executable you do not need the main.c. You can also create a new wrapper library around lvgl. (In that case you obviously don't need a main.c)

So to sum up, this PR simply enables users to build this repo out of the box without having to make any changes. The provides cmake presets can be easily overwritten by any other cmake project.
For example: If i check out the https://github.com/lvgl/lv_port_pc_visual_studio
the CmakePresets.json within the lvgl repo does not have any effect because the .sln file from lv_port_pc_visual_studio does not invoke cmake :)

kisvegabor: So now the questions is which case is more common, and that one should be the default:
1. Build a library from LVGL
2. Build an executable with LVGL

I think 2) is more common therefore we shouldn't add `"LV_CONF_SKIP": true` by default but let the user add it if they are building a library. Note that even when compiling a library you can choose to use `lv_cong.h`.

I'm afraid that having `"LV_CONF_SKIP": true,` by default will confuse most of the users as they won't know why their changes in `lv_conf.h`have no effect. Having a compiler error about "`lv_conf.h` is missing" is more explicit.

ghost: Hey, I agree the second option is more common.
However in that case the project will not build since the lv_conf.h is not found.
I'd suggest to add a demo executable to this project (i.e. by using cmakes add_executable()). This demo exe can be build with a demo lv_conf.h
Since all is located in a "demo" folder within this project it is clear to every new user that this needs to be adapted when creating an own project. Our cmake presets will simply build this demo exe as well.
Or do you have a better suggestion how to handle the lv_conf.h. We need to provide one at build time, otherwise build will fail :)

However the `LV_SKIP_CONFIG: true` is already in the master branch. To keeps things clear I'd suggest to create a separate PR for this topic. Let's keep things in this PR simple and straight forward :)

kisvegabor: > However the LV_SKIP_CONFIG: true is already in the master branch. To keeps things clear I'd suggest to create a separate PR for this topic. Let's keep things in this PR simple and straight forward :)

Ah, you are right. In this case it's really unrelated to this PR. Let's keep it as it is :smile:

kisvegabor (dismissed):

liamHowatt:
Hey,

CMake Presets is how I tried using LVGL for the very first time and I did not have a positive experience. I don't really understand what someone can do with the limited defaults in the presets, to be honest.

I think your changes make sense in the context of the existing CMake Presets system. I want to continue the discussion though.

After building the shared libs with CMake presets, I made an executable to link those shared libs to run an example that uses vector graphics.

```c
#include <stdint.h>
#include <unistd.h>

void lv_init(void);
typedef struct lv_display_t lv_display_t;
lv_display_t * lv_sdl_window_create(int32_t hor_res, int32_t ver_res);
void lv_example_canvas_8(void);
uint32_t lv_timer_handler(void);

int main()
{
    lv_init();
    lv_sdl_window_create(1000, 600);
    lv_example_canvas_8();

    while(1) {
        uint32_t ms_delay = lv_timer_handler();
        usleep(ms_delay * 1000);
    }
}
```
After the shared libraries have been built:
```sh
$ gcc -Wall -L$PWD/build/linux-base/lib/Debug main.c -llvgl -llvgl_examples -llvgl_thorvg -lSDL2
$ LD_LIBRARY_PATH=$PWD/build/linux-base/lib/Debug ./a.out
```

In order to get it to work, I had to change the CMake presets like this and rebuild (the SDL parts aren't important, ignore):

```diff
diff --git a/CMakePresets.json b/CMakePresets.json
index 9bbced05e..804439ed8 100644
--- a/CMakePresets.json
+++ b/CMakePresets.json
@@ -7,7 +7,8 @@
       "cacheVariables": {
         "LV_CONF_SKIP": true,
         "CONFIG_LV_USE_DEMO_WIDGETS": true,
-        "CONFIG_LV_BUILD_EXAMPLES": true
+        "CONFIG_LV_BUILD_EXAMPLES": true,
+        "CONFIG_LV_USE_SDL": true
       }
     },
     {
diff --git a/env_support/cmake/custom.cmake b/env_support/cmake/custom.cmake
index a7a32a60a..1465d95f3 100644
--- a/env_support/cmake/custom.cmake
+++ b/env_support/cmake/custom.cmake
@@ -66,6 +66,7 @@ if(LV_CONF_BUILD_THORVG_INTERNAL OR (NOT LV_CONF_BUILD_DISABLE_THORVG_INTERNAL))
     target_include_directories(lvgl_thorvg SYSTEM PUBLIC ${LVGL_ROOT_DIR}/src/libs/thorvg)
     target_link_libraries(lvgl_thorvg PUBLIC lvgl)
     target_compile_definitions(lvgl_thorvg PUBLIC LV_USE_THORVG_INTERNAL=1 LV_USE_THORVG_EXTERNAL=0)
+    target_compile_definitions(lvgl PUBLIC LV_USE_VECTOR_GRAPHIC=1 LV_USE_THORVG_INTERNAL=1 LV_USE_THORVG_EXTERNAL=0)
 endif()
 
 if(NOT MSVC)
@@ -79,7 +80,11 @@ if(CONFIG_LV_BUILD_EXAMPLES OR (NOT LV_CONF_BUILD_DISABLE_EXAMPLES))
 
     target_include_directories(lvgl_examples SYSTEM PUBLIC ${LVGL_ROOT_DIR}/examples)
     target_link_libraries(lvgl_examples PUBLIC lvgl)
-    target_compile_definitions(lvgl_examples PUBLIC LV_BUILD_EXAMPLES=1)
+    target_compile_definitions(lvgl_examples PUBLIC LV_BUILD_EXAMPLES=1 LV_USE_CANVAS=1)
+    target_compile_definitions(lvgl PUBLIC LV_USE_CANVAS=1)
+    if(TARGET lvgl_thorvg)
+        target_compile_definitions(lvgl_examples PUBLIC LV_USE_VECTOR_GRAPHIC=1)
+    endif()
 endif()
 
 # Build LVGL demos library (LV_CONF_BUILD_DISABLE_DEMOS ist just kept for backwards compatibility, prefer CONFIG_LV_USE_DEMO_WIDGETS)
@@ -92,6 +97,10 @@ if(CONFIG_LV_USE_DEMO_WIDGETS OR (NOT LV_CONF_BUILD_DISABLE_DEMOS))
     target_compile_definitions(lvgl_demos PUBLIC LV_USE_DEMO_WIDGETS=1)
 endif()
 
+if(CONFIG_LV_USE_SDL)
+    target_compile_definitions(lvgl PUBLIC LV_USE_SDL=1)
+endif()
+
 # Lbrary and headers can be installed to system using make install
 file(GLOB LVGL_PUBLIC_HEADERS "${CMAKE_SOURCE_DIR}/lv_conf.h"
      "${CMAKE_SOURCE_DIR}/lvgl.h")
```

My point is, if thorvg and the examples are enabled, you can't run a vector graphics example until:
- The lvgl lib, thorvg lib, and examples lib have the vector graphics config enabled
- the lvgl lib knows thorvg has been enabled
- the examples lib knows thorvg has been enabled
- the examples lib has canvas (or any other widget) enabled
- the lvgl lib has canvas (or any other widget) enabled

Is there a practical use case of the default presets that I'm overlooking?

<br>

```cmake
if(CONFIG_LV_BUILD_EXAMPLES OR (NOT LV_CONF_BUILD_DISABLE_EXAMPLES))
```

I think this effort preserve backwards-compatibility does not serve the new config option well, or the legacy one.

|               |   | BUILD_DISABLE_EXAMPLES |           |
|---|---|---|---|
|               |   |     false                  |     true     |
|BUILD_EXAMPLES | false | enabled                | disabled  |
|               | true | enabled                | enabled   |

It cannot actually be disabled by setting `"CONFIG_LV_BUILD_EXAMPLES": false` alone. `LV_CONF_BUILD_DISABLE_EXAMPLES` must also explicitly set to `true`.

<br>

Thank you for this change. I wasn't aware of that when I added it.
```
-if(NOT (CMAKE_C_COMPILER_ID STREQUAL "MSVC"))
+if(NOT MSVC)
```

Ju1He1:
thx @liamHowatt  I did completely miss that.
@kisvegabor  I needed to drop backwards compatibility as it did work properly. Hope this is OK with you :)
(I added an error if anybody is still using the old DISABLE_XXX stuff)

@liamHowatt  In out company we are using CMakePresets to build Windows, Linux, arm_none_eabi, MacOS, IOS and Android Apps. It makes cross platform building much more easy that the "traditional" cmake way of adding lots of `-DFOO` to the cmake invocation. However i must admit that we are some cmake power users :P

Regarding your thor_vg issues please create a separate PR. I have not used throrvg, therefore i do not know under which circumstance this will work or not :)

tests/CMakeLists.txt
```diff
@@ -98,6 +98,9 @@ set(LVGL_TEST_OPTIONS_TEST_DEFHEAP
     --coverage
 )
 
+set(CONFIG_LV_BUILD_EXAMPLES ON)
+set(CONFIG_LV_USE_DEMO_WIDGETS ON)
+
```
kisvegabor: It should be added to pass the CI
```cmake
set(LV_CONF_BUILD_THORVG_INTERNAL ON)
```

liamHowatt:
Hey, looks like the merge you did worked out ok? I lost sight of this PR when I changed the `tests/CMakeLists.txt` in this PR: https://github.com/lvgl/lvgl/pull/6264/files#diff-4461c617ceae0c7e0206622aacdf555aedd62eb129c2efb74c84fa1567bcbe0d

Let me know if you're running into an issue with `set(LV_CONF_BUILD_THORVG_INTERNAL ON)` or anything.

Ju1He1:
Hey @liamHowatt  I did set USE_THORVG_INTERNAL to on but I'm getting a compiler error
`#error:  "ThorVG: LV_USE_VECTOR_GRAPHIC is required. Enable it in lv_conf.h"`
In your first post you mentioned that some of the thorvg logic is still wrong.
Can you tell me what I need to change (I haven't used thorvg an have therefore no experience with this topic ;) )

env_support/cmake/custom.cmake
```diff
@@ -46,37 +46,54 @@ if(LV_CONF_SKIP)
...
+  message(FATAL_ERROR "LV_CONF_BUILD_DISABLE_THORVG_INTERNAL is deprecated, use CONFIG_LV_USE_THORVG_INTERNAL instead")
+endif()
 
-if(NOT LV_CONF_BUILD_DISABLE_THORVG_INTERNAL)
+if(LV_CONF_BUILD_THORVG_INTERNAL)
     add_library(lvgl_thorvg ${THORVG_SOURCES})
     add_library(lvgl::thorvg ALIAS lvgl_thorvg)
     target_include_directories(lvgl_thorvg SYSTEM PUBLIC ${LVGL_ROOT_DIR}/src/libs/thorvg)
     target_link_libraries(lvgl_thorvg PUBLIC lvgl)
+    target_compile_definitions(lvgl_thorvg PUBLIC LV_USE_THORVG_INTERNAL=1 LV_USE_THORVG_EXTERNAL=0)
```
liamHowatt:
```suggestion
    target_compile_definitions(lvgl_thorvg PUBLIC LV_USE_VECTOR_GRAPHIC=1 LV_USE_THORVG_INTERNAL=1 LV_USE_THORVG_EXTERNAL=0)
```

It should make the CI pass.

tests/CMakeLists.txt
liamHowatt, line 1: For the CI, please update one more `set (LV_CONF_BUILD_DISABLE_EXAMPLES ON)` on line 139

env_support/cmake/custom.cmake
```diff
@@ -46,37 +46,54 @@ if(LV_CONF_SKIP)
...
 
 # Build LVGL demos library
-if(NOT LV_CONF_BUILD_DISABLE_DEMOS)
+if(CONFIG_LV_USE_DEMO_WIDGETS)
     add_library(lvgl_demos ${DEMO_SOURCES})
     add_library(lvgl::demos ALIAS lvgl_demos)
 
     target_include_directories(lvgl_demos SYSTEM PUBLIC ${LVGL_ROOT_DIR}/demos)
     target_link_libraries(lvgl_demos PUBLIC lvgl)
+    target_compile_definitions(lvgl_demos PUBLIC LV_USE_DEMO_WIDGETS=1)
```
liamHowatt: Unrelated to the CI, I think `CONFIG_LV_USE_DEMO_WIDGETS` should be `CONFIG_LV_BUILD_DEMOS`, and maybe we shouldn't enable the widgets demo in here? `DEMO_WIDGETS` is just one of the `DEMOS`.

env_support/cmake/custom.cmake
```diff
@@ -101,7 +118,7 @@ install(
   PATTERN "*.h")
 
 # install example headers
-if(NOT LV_CONF_BUILD_DISABLE_EXAMPLES)
+if(CONFIG_LV_BUILD_EXAMPLES)
```
liamHowatt, line 136: For all of these, can we do something like `if(NOT DEFINED CONFIG_LV_BUILD_EXAMPLES OR CONFIG_LV_BUILD_EXAMPLES)`? The reason I ask is because users using a CMakeLists.txt like [this](https://gist.github.com/liamHowatt/2b7b101cbcc93ac3a6d687e82f63f816) will have their build break until they add `set(CONFIG_LV_BUILD_EXAMPLES ON)`.

I suppose we are already breaking users of `*_DISABLE_*`

liamHowatt (commented):
> thx @liamHowatt I did completely miss that. @kisvegabor I needed to drop backwards compatibility as it did work properly. Hope this is OK with you :) (I added an error if anybody is still using the old DISABLE_XXX stuff)
> 
> @liamHowatt In out company we are using CMakePresets to build Windows, Linux, arm_none_eabi, MacOS, IOS and Android Apps. It makes cross platform building much more easy that the "traditional" cmake way of adding lots of `-DFOO` to the cmake invocation. However i must admit that we are some cmake power users :P
> 
> Regarding your thor_vg issues please create a separate PR. I have not used throrvg, therefore i do not know under which circumstance this will work or not :)

Gotcha, if someone is using it, then it's good to have!

> Hey @liamHowatt I did set USE_THORVG_INTERNAL to on but I'm getting a compiler error `#error: "ThorVG: LV_USE_VECTOR_GRAPHIC is required. Enable it in lv_conf.h"` In your first post you mentioned that some of the thorvg logic is still wrong. Can you tell me what I need to change (I haven't used thorvg an have therefore no experience with this topic ;) )

I think I was able to get it. See my comments.

Also see my comments for some other thoughts.

env_support/cmake/custom.cmake
ghost: this is a design question ;) @kisvegabor  what do you think :)

tests/CMakeLists.txt
ghost: ah thx :)

Ju1He1:
@liamHowatt  thx for your help. I finally got the CI back to work!

@kisvegabor does this PR work for you?

env_support/cmake/custom.cmake
kisvegabor: I think enabling demos and examples if `CONFIG_LV_BUILD_EXAMPLES` is not defined is a good idea. We need to think about novice users and a good get started experience.   

Early in the main [`CMakeList.txt`](https://github.com/lvgl/lvgl/blob/master/CMakeLists.txt) file we can define these with a default value (if not defined).

env_support/cmake/custom.cmake
ghost: allright replaced CONFIG_LV_USE_DEMO_WIDGETS with CONFIG_LV_BUILD_DEMOS.
If we would like to get rid of CONFIG_LV_BUILD_DEMOS we should remove the define `LV_USE_DEMO_WIDGETS=1`. But lets handle this in a separate PR. This PR is getting far too big :)

env_support/cmake/custom.cmake
ghost: I added a FATAL_ERROR if CONFIG_LV_BUILD_EXAMPLES is set but not CONFIG_LV_BUILD_DEMOS.

I'd like to keep all settings in a single place. 
In this is the CMake Presets.json (or alternatively within the cmake invocation `cmake -G ... -DFOO`) is the only place where config values are set.

If we start setting config values implicitly (in this case the root CmakeLists.txt) thing will get much more complicated .
I hope this is ok for you guys

ghost: ASAN is complaining about a memory leak because i needed to build the examples whithin the test as well.
Does anybody know how to solve this?

kisvegabor (dismissed):
Thank you, LGTM!

env_support/cmake/custom.cmake
kisvegabor: I'm open to where the config options are set, but we shall provide a smooth getting started experience so novice users too. It means `cmake ..; make` should be able to compile the demos and example without reading the docs for extra option. Also note that 

Therefore even if it's more complicated on our side I would prefer adding something like this for demos and examples:
```
if(NOT DEFINED CONFIG_LV_BUILD_EXAMPLES)
    set(CONFIG_LV_BUILD_EXAMPLES ON)
endif()

if(NOT DEFINED CONFIG_LV_BUILD_DEMOS)
    set(CONFIG_LV_BUILD_DEMOS ON)
endif()
```

It's okay to enable ThorVG explicitly as it's not certain that it can be compiled on all systems. And it's considered as an advanced feature.

I've also noticed that the name configs options are structured differenty:
- **CONFIG_LV**_BUILD_EXAMPLES ON
- **CONFIG_LV**_BUILD_DEMOS ON
- **LV_CONF**_BUILD_THORVG_INTERNAL


Probably both the Kconfig names  (e.g. `CONFIG_LV_BUILD_EXAMPLES` for compatibility with ESP and Zephyr) and `lv_conf.h` names `LV_BUILD_EXAMPLES` should work. It can result in a config conflict if the examples are disabled in `lv_conf.h` but enabled in `cmake`, however we cannot do much about it. If you start setting the values via `cmake` you should remove them from `lv_conf.h`.

@liamHowatt before merging we should test if the examples can be enabled/disabled e.g. with ESP with these changes.

kisvegabor (commented):
Oh, sorry, I've missed something.

Ju1He1:
any update on this guys :) Your bot marked this as stale :P

liamHowatt:
Please respond to this thread. Thanks.

> Oh, sorry, I've missed something.

Ju1He1:
> Please respond to this thread. Thanks.
> 
> > Oh, sorry, I've missed something.

Hey maybe there is a misunderstanding on my side or yours :)

From my point of view we should only build examples and demos if someone is actually building this (https://github.com/lvgl/lvgl.git) repo via cmake.
Building examples and demos should be disabled as a default if someone adds this repo to his project via add_subdirectory(), FetchContent() or via a package manager.

At the moment this is the behaviour.

If we add this to some cmake lists _within_ this repo. Examples and Demos will be build for everyone without having an option to turn this off. I don't believe this is a desired behaviour :)

```
if(NOT DEFINED CONFIG_LV_BUILD_EXAMPLES)
    set(CONFIG_LV_BUILD_EXAMPLES ON)
endif()

if(NOT DEFINED CONFIG_LV_BUILD_DEMOS)
    set(CONFIG_LV_BUILD_DEMOS ON)
endif()
```

Just to be clear once again, the presets are purely optional. You can build this repo without presets
(Infact the presets stored in this repo are implicitly discarded when adding this repo to your own project via add_subdirectory() fetchcontent or via package manager). The idea of the presets is just to have an (optional) set of predefined variables with a nice IDE integration.

So correct me if I'm wrong but I'd prefer to build examples and demos ONLY if this repo is build as a stand-alone project, if you embed this repo into your project by default examples and demos are turned off (You are free though to explicitly set CONFIG_LV_BUILD_EXAMPLES and CONFIG_LV_BUILD_DEMOS  in YOUR project).



> I've also noticed that the name configs options are structured differenty:
> 
>     CONFIG_LV_BUILD_EXAMPLES ON
>     CONFIG_LV_BUILD_DEMOS ON
>     LV_CONF_BUILD_THORVG_INTERNAL

I don't care about the names. Just tell we what you would like to rename and I will do it :)

> @liamHowatt before merging we should test if the examples can be enabled/disabled e.g. with ESP with these changes.

This one i cannot do. I'm on Windows. I haven't installed zephyr or the stuff needed for ESP :)

kisvegabor:
> Just to be clear once again, the presets are purely optional. 

Thanks for highlighting it again! It's really important that with the presets we are focusing on the standalone build. However, we still need to consider how the changes of `custom.cmake` affect the `add_subdirectory` cases. 

> If we add this to some cmake lists within this repo. Examples and Demos will be build for everyone without having an option to turn this off. I don't believe this is a desired behaviour :)
>> if(NOT DEFINED CONFIG_LV_BUILD_EXAMPLES)
    set(CONFIG_LV_BUILD_EXAMPLES ON)
endif()
if(NOT DEFINED CONFIG_LV_BUILD_DEMOS)
    set(CONFIG_LV_BUILD_DEMOS ON)
endif()

They can be disabled by adding `set(CONFIG_LV_BUILD_DEMOS OFF)`.

Ju1He1:
> > Just to be clear once again, the presets are purely optional.
> 
> Thanks for highlighting it again! It's really important that with the presets we are focusing on the standalone build. However, we still need to consider how the changes of `custom.cmake` affect the `add_subdirectory` cases.
> 
> > If we add this to some cmake lists within this repo. Examples and Demos will be build for everyone without having an option to turn this off. I don't believe this is a desired behaviour :)
> > > if(NOT DEFINED CONFIG_LV_BUILD_EXAMPLES)
> > > set(CONFIG_LV_BUILD_EXAMPLES ON)
> > > endif()
> > > if(NOT DEFINED CONFIG_LV_BUILD_DEMOS)
> > > set(CONFIG_LV_BUILD_DEMOS ON)
> > > endif()
> 
> They can be disabled by adding `set(CONFIG_LV_BUILD_DEMOS OFF)`.

sorry I do not get it  :(
Can you show me where exactly you would like to add something?

kisvegabor:
I'm not too experienced with CMake so please tell me if I'm talking non-sense. 

> if(NOT DEFINED CONFIG_LV_BUILD_EXAMPLES)
set(CONFIG_LV_BUILD_EXAMPLES ON)
endif()
if(NOT DEFINED CONFIG_LV_BUILD_DEMOS)
set(CONFIG_LV_BUILD_DEMOS ON)
endif()

So if `CONFIG_LV_BUILD_EXAMPLES` is not defined it will be enabled by default. If we want to avoid setting its default we need to define it either
1. In a higher level CMake file by adding `set(CONFIG_LV_BUILD_DEMOS OFF)`
2. In command line with `cmake -D CONFIG_LV_BUILD_EXAMPLES=OFF ..` 

Does it make sense?

Ju1He1:
> I'm not too experienced with CMake so please tell me if I'm talking non-sense.
> 
> > if(NOT DEFINED CONFIG_LV_BUILD_EXAMPLES)
> > set(CONFIG_LV_BUILD_EXAMPLES ON)
> > endif()
> > if(NOT DEFINED CONFIG_LV_BUILD_DEMOS)
> > set(CONFIG_LV_BUILD_DEMOS ON)
> > endif()
> 
> So if `CONFIG_LV_BUILD_EXAMPLES` is not defined it will be enabled by default. If we want to avoid setting its default we need to define it either
> 
>     1. In a higher level CMake file by adding `set(CONFIG_LV_BUILD_EXAMPLES OFF)`
> 
>     2. In command line with `cmake -D CONFIG_LV_BUILD_EXAMPLES=OFF ..`
> 
> 
> Does it make sense?

so basically you would like to build examples and demos by default and disable them optionally?

> 1)In a higher level CMake file by adding set(CONFIG_LV_BUILD_EXAMPLES OFF)
=>this should work

> 2) In command line with cmake -D CONFIG_LV_BUILD_EXAMPLES=OFF ..

basically it should work but specified variables via Command Line and CMake Presets are treated as Cache Variables. This means you cannot change them with a simple `set(CONFIG_LV_BUILD_EXAMPLES ON)` you will need to use 
`set(CONFIG_LV_BUILD_EXAMPLES ON CACHE BOOL "disable examples")`
A "normal" set will not have any effect at all on cache variables. 

If that behaviour is fine for you we could proceed with your proposal.

kisvegabor:
> set(CONFIG_LV_BUILD_EXAMPLES ON CACHE BOOL "disable examples")

If the following are true, I'm okay with the cached version:
- Normal `set(CONFIG_LV_BUILD_EXAMPLES ON)` work in "normal CMake" (not CMake Preset)
- Normal `set(CONFIG_LV_BUILD_EXAMPLES ON)` works if you delete the `build` folder and run `cmake` again
- We document it :slightly_smiling_face:

Ju1He1:
> > set(CONFIG_LV_BUILD_EXAMPLES ON CACHE BOOL "disable examples")
> 
> If the followings are true, I'm okay with the cached version:
> 
>     * Normal `set(CONFIG_LV_BUILD_EXAMPLES ON)` work in "normal CMake" (not CMake Preset)
> 
>     * Normal `set(CONFIG_LV_BUILD_EXAMPLES ON)` works if you delete the `build` folder and run `cmake` again
> 
>     * We document it 🙂

Yes both of your statements are true if no CMakePresets are not used and CONFIG_* are not passed via CLI (`cmake -DCONFIG_*)`


```
- LV_CONF_SKIP (default: true)
    Skip the configuration step. This is useful if you want to use the default configuration.
  - LV_CONF_PATH (default: (not set))
    Path to an optional user defined configuration file. (Obviously you cannot mix it with LV_CONF_SKIP)
  - CONFIG_LV_BUILD_DEMOS (default: true)
    Enable the demo widgets, automatically sets the proper define LV_USE_DEMO_WIDGETS=1
  - CONFIG_LV_BUILD_EXAMPLES (default: true)
    Build the examples, automatically sets the proper define LV_BUILD_EXAMPLES=1
  - LV_CONF_BUILD_THORVG_INTERNAL (default: true)
    Build the thorvg extension, automatically sets the proper define LV_USE_THORVG_INTERNAL=1
```

Documentation already specifies that example and demos are build as default. But it's fine to disable them. Do you wan't to add anything more

Ju1He1:
@kisvegabor  @liamHowatt  nay suggestions how to fix the CI?
I can try to fix the errors but it seams like i need to change stuff which i do not really want to change :)

liamHowatt:
It's because `LV_USE_FREETYPE` and `LV_USE_THORVG` aren't defined (if not already) until lower down in `lv_conf_internal.h`.

How come this PR has these stack size changes? It makes sense but maybe it doesn't belong in this PR anyways.

kisvegabor:
It seems [`lv_test_conf.h`](https://github.com/lvgl/lvgl/blob/master/tests/src/lv_test_conf.h) is not loaded in the tests.

kisvegabor:
Tests require [`LV_CONF_PATH=lv_test_conf.h`](https://github.com/lvgl/lvgl/blob/master/tests/CMakeLists.txt#L172), however seemingly this part is not touched.

PR #6097, opened 2024-04-21
fix(docs): pull version out of lv_version.h

marex:
The version macros moved to lv_version.h , update find_version.sh script.

PR #6122, opened 2024-04-24
fix(docs) use find_version helper in build script

embeddedt:
In #6097 the `scripts/find_version.h` script was updated, but `docs/build.py` was still parsing the version manually. This PR fixes the problem noted in https://github.com/lvgl/lvgl/issues/6111#issuecomment-2072591117 by updating `docs/build.py` to use that script.

### DIFF
--- a/docs/build.py
+++ b/docs/build.py
@@ -207,27 +207,7 @@ else:
 
 
 def get_version():
-    path = os.path.join(project_path, 'lvgl.h')
-    with open(path, 'rb') as f:
-        d = f.read().decode('utf-8')
-
-    d = d.split('#define LVGL_VERSION_MAJOR', 1)[-1]
-    major, d = d.split('\n', 1)
-    d = d.split('#define LVGL_VERSION_MINOR', 1)[-1]
-    minor, d = d.split('\n', 1)
-    # d = d.split('#define LVGL_VERSION_PATCH', 1)[-1]
-    # patch, d = d.split('\n', 1)
-
-    ver = '{0}.{1}'.format(major.strip(), minor.strip())
-
-    # ver = '{0}.{1}.{2}'.format(major.strip(), minor.strip(), patch.strip())
-
-    # if '#define LVGL_VERSION_INFO' in d:
-    #     d = d.split('#define LVGL_VERSION_INFO', 1)[-1]
-    #     info, d = d.split('\n', 1)
-    #     info = info.strip().replace('"', '')
-    #     ver += '-' + info
-
+    _, ver = subprocess.getstatusoutput("../scripts/find_version.sh")
     return ver
 
 cmd('sphinx-build -b html "{src}" "{dst}" -D version="{version}" -E -j {cpu}'.format(


kisvegabor (approved):
Thank you so much for the quick fix! I'm merging it right now.

kdschlosser:
This fix is no good. It limits the documentation to only being able to be compiled on unix.

kisvegabor:
It's good enough to make the docs work again. :slightly_smiling_face: 

However, ideas are welcome. I added a [`backup` branch](https://github.com/lvgl/docs/tree/backup) as we almost lost all the docs from the previous versions. So it's a safer playground now.

kdschlosser:
All that needed to be done is pointing the build.py script for the docs to the correct header file in LVGL. While version 9.0 was being developed at some point the version information got moved from lvgl.h to lvgl_version.h. That is all that needs to be changed is the file name.

Need to change this code..
```
def get_version():
    path = os.path.join(project_path, 'lvgl.h')
    with open(path, 'rb') as f:
        d = f.read().decode('utf-8')

    d = d.split('#define LVGL_VERSION_MAJOR', 1)[-1]
    major, d = d.split('\n', 1)
    d = d.split('#define LVGL_VERSION_MINOR', 1)[-1]
    minor, d = d.split('\n', 1)
    # d = d.split('#define LVGL_VERSION_PATCH', 1)[-1]
    # patch, d = d.split('\n', 1)

    ver = '{0}.{1}'.format(major.strip(), minor.strip())

    # ver = '{0}.{1}.{2}'.format(major.strip(), minor.strip(), patch.strip())

    # if '#define LVGL_VERSION_INFO' in d:
    #     d = d.split('#define LVGL_VERSION_INFO', 1)[-1]
    #     info, d = d.split('\n', 1)
    #     info = info.strip().replace('"', '')
    #     ver += '-' + info

    return ver
```


To this
```
def get_version():
    path = os.path.join(project_path, 'lv_version.h')
    with open(path, 'rb') as f:
        d = f.read().decode('utf-8')

    d = d.split('#define LVGL_VERSION_MAJOR', 1)[-1]
    major, d = d.split('\n', 1)
    d = d.split('#define LVGL_VERSION_MINOR', 1)[-1]
    minor, d = d.split('\n', 1)
    # d = d.split('#define LVGL_VERSION_PATCH', 1)[-1]
    # patch, d = d.split('\n', 1)

    ver = '{0}.{1}'.format(major.strip(), minor.strip())

    # ver = '{0}.{1}.{2}'.format(major.strip(), minor.strip(), patch.strip())

    # if '#define LVGL_VERSION_INFO' in d:
    #     d = d.split('#define LVGL_VERSION_INFO', 1)[-1]
    #     info, d = d.split('\n', 1)
    #     info = info.strip().replace('"', '')
    #     ver += '-' + info

    return ver
```

and it's a done deal and is able to compile on Windows or a unix system.

embeddedt:
I agree with your proposed change but IMO we should then move it to a standalone script, import that script from the docs builder, and update `find_version.sh` to call that script behind the scenes. The main reason I changed it this way rather than doing the simple one-line fix was to avoid the same issue happening again in the future, where we have two distinct implementations of the same thing.

kdschlosser:
we can definitely do that no worries. Until 5 days ago the only thing that used that script was the documentation build system until I updated the system a year ago.  5 days ago there was a change made in the cmake build system to use that script which is also not the right thing to do because the easy way of compiling under windows is to use cmake which is now not possible unless WSL is installed.
 
 Python comes pre installed into a lot of linux distros especially Python 2. If we want to go that route I can write the code so it is both python 2 and 3 compliant so there is no worries about it running there. We need to make sure to add the lines of code into the cmake build so it will do a check to make sure python is installed on the computer, whether it be unix or windows doesn't matter. I am pretty sure that this is able to be easily done in cmake.

kdschlosser:
we can just add the python script into the scripts folder and do away with the bash script all together. I can import the script from anywhere in the LVGL library so that's not a problem at all. The extra layer of having the bash script run the python script is just adding another layer to it which make it more complex. 


what are all of the places where that script is used?? I only know of 2 places cmake and the doc build system...

An actual better way to handle this whole thing would be to use the release numbers assigned to the branches and use git to collect that branch name and generate lv_version using that. If the mater branch is what is being used the version can be set to whatever default is wanted. It could be some off the wall number with `dev` in it.

kdschlosser:
when a version is going to be made at that point in time a new branch should be made for it. Once that version is considered as released a PR from that new version branch should be made to the master branch and the code gets merged. Active development should not be done in the master branch. The master branch should have the latest release and that goes untouched until another release is made.

kdschlosser:
The way to handle this is to use something to create/collect the version that is not platform specific. I suggest using python to do it because python is available for most OS's and the same script would run in the same manner no matter what OS it is running on.

Using cmake or make to build the version file is not a wise idea because cmake and make are not used to compile the documentation. so if the documentation was to be compiled it would fail because of the version file not existing. It would require using make or cmake to do the compilation prior to being able to build documentation which seems silly to me.

That being said. build commands can be added to cmake to build the documentation if that is the way things are wanting to go. It is not a horrible idea to do that as it would streamline everything into 2 build systems which would make things easier at the CI level. 

I can write the python script that can build the version file. Then that code would be in a single place and it would get called from make or cmake. If this is done using the build system (cmake and make) would require 2 different sets of code be maintained instead of one. Having a python script run from inside of cmake or make is easy to do. and wouldn't require any maintenance.

Also using a python script would allow the document build system to collect the versions once again using the same code that is used to generate the version file.

I am only pointing out that the current fix is not cross platform and it locks everything to OS's that are able to run bash. Prior to making this change everything was able to be compiled on just about any OS. Now it has been limited. I personally feel that going this route is not the correct thing to do. But.. I am not the one who makes those decisions. So whatever it is that is decided then that is what will be done. If the use of the bash script is what is chosen this is going to limit what I am able to fix because I use Windows.

kisvegabor:
@marex have already opened #6135. I've seen that first and commented there about using a generic python script. :slightly_smiling_face: 

I wouldn't jump into the reworking of the docs' build system, as it's okay as it is. I think just adding `find_version.py` as a central script to get LVGL's version would be simple but great step forward.

PR #6128, opened 2024-04-25
fix(fsdrv/fatfs): support FF_DIR and FATFS_DIR typedef in ff.h

### LINKED ISSUE #6078: Compilation error for FATFS
### LVGL version

v9.1.0

### What happened?

Enabled LV_USE_FS_FATFS in lv_conf,h and got the following error:
.pio/libdeps/esp32-s3-devkitc-1/lvgl/src/libs/fsdrv/lv_fs_fatfs.c:228:5: error: unknown type name 'DIR'; did you mean 'DDR'?

### How to reproduce?

#define LV_USE_FS_FATFS 1
#if LV_USE_FS_FATFS
    #define LV_FS_FATFS_LETTER 'F'     /*Set an upper cased letter on which the drive will accessible (e.g. 'A')*/
    #define LV_FS_FATFS_CACHE_SIZE 0    /*>0 to cache this number of bytes in lv_fs_read()*/
#endif

liamHowatt:

Fixes #6078

See the issue for more details.

### Notes

### DIFF
--- a/src/libs/fsdrv/lv_fs_fatfs.c
+++ b/src/libs/fsdrv/lv_fs_fatfs.c
@@ -17,7 +17,7 @@
  *********************/
 
 #ifdef ESP_PLATFORM
-#define DIR FF_DIR  /* ESP IDF typedefs `DIR` as `FF_DIR` in its version of ff.h. Use `FF_DIR` in LVGL too */
+    #define DIR FF_DIR  /* ESP IDF typedefs `DIR` as `FF_DIR` in its version of ff.h. Use `FF_DIR` in LVGL too */
 #endif
 
 #if LV_FS_FATFS_LETTER == '\0'
--- a/src/libs/fsdrv/lv_fs_fatfs.c
+++ b/src/libs/fsdrv/lv_fs_fatfs.c
@@ -9,6 +9,9 @@
 #include "../../../lvgl.h"
 
 #if LV_USE_FS_FATFS
+
+#define FF_DIR DIR    /* ESP IDF typedefs `DIR` as `FF_DIR`. Rename to `DIR` for use with this source file */
+#define FATFS_DIR DIR /* Mbed OS typedefs `DIR` as `FATFS_DIR`. Rename to `DIR` for use with this source file */
 #include "ff.h"
 
 #include "../../core/lv_global.h"
--- a/src/libs/fsdrv/lv_fs_fatfs.c
+++ b/src/libs/fsdrv/lv_fs_fatfs.c
@@ -9,15 +9,16 @@
 #include "../../../lvgl.h"
 
 #if LV_USE_FS_FATFS
-
-#define FF_DIR DIR    /* ESP IDF typedefs `DIR` as `FF_DIR`. Rename to `DIR` for use with this source file */
-#define FATFS_DIR DIR /* Mbed OS typedefs `DIR` as `FATFS_DIR`. Rename to `DIR` for use with this source file */
 #include "ff.h"
 
 #include "../../core/lv_global.h"
 /*********************
  *      DEFINES
  *********************/
+
+#ifdef ESP_PLATFORM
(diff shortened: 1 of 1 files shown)


kisvegabor:
Can't we add the defines below `#include "ff.h"` to be sure that the new names won't mess up anything in `ff.h`?

liamHowatt:
Manipulating the contents of ff.h is the point of putting the defines before the include. To convey that I should `undef` it right after the include.

Ideally we'd detect what the platform has renamed `DIR` to and put the define after the header.

Or something like this pseudocode:
```c
#incude "ff.h"

#if (defined(ESP32) && !defined(ARDUINO)) || defined(ESP_IDF)
#define DIR FF_DIR
#elif defined(MBED_MAJOR_VERSION)
#define DIR FATFS_DIR
#endif

**Comment by liamHowatt** [comment]
Actually, IIRC Mbed users would not use lv_fs_fatfs. They'd use lv_fs_stdio which maps to Mbed's own FATFS adapter. So maybe we can ignore FATFS_DIR and only worry about FF_DIR. But the problem still stands, how to we conditionally define DIR as FF_DIR when on ESP IDF?

**Comment by kisvegabor** [comment]
Still don't get it. ESP's `"ff.h"` is working internally, and the problem is that it defines `FF_DIR` but we use `DIR` in `lv_fs_fatfs.c`. Is that correct? 

If so, all we need to do in `lv_fs_fatfs.c` is telling  that `FF_DIR` also means `DIR` by adding `#define FF_DIR DIR` after `ff.h`.

**Comment by liamHowatt** [comment]
> `"ff.h"` is working internally, and the problem is that it defines FF_DIR but we use DIR in lv_fs_fatfs.c. Is that correct?

Correct.

> If so, all we need to do in `lv_fs_fatfs.c` is telling that `FF_DIR` also means `DIR` by adding `#define FF_DIR DIR` after `ff.h`.

I'm afraid `#define FF_DIR DIR` won't work. That define would mean replace any instance of the token `FF_DIR` with `DIR` after this define. There are no more `FF_DIR` tokens _after_ the include of `"ff.h"` was already done. Look at what worked for @brandon-kim https://github.com/lvgl/lvgl/issues/6078#issuecomment-2063024388. They did `#define DIR FF_DIR` which replaces all `DIR` in `lv_fs_fatfs.c` with `FF_DIR` to match how `"ff.h"` has defined it. That works only if the `"ff.h"` uses `FF_DIR`. If it used `DIR`, we don't want that define.

So my idea was to manipulate the names inside `"ff.h"` to be `DIR` in case they're not.

**Comment by kisvegabor** [comment]
> I'm afraid #define FF_DIR DIR won't work.

Oh, sorry, you are right. It should be `#define DIR FF_DIR` after `#include "ff.h"`. If there is no objection let's use it not affect `ff.h` by any means.

**Comment by liamHowatt** [comment]
Ah but you see, if they're using the variant of `"ff.h"` that defines the directory type as `DIR` (i.e. Arduino), then that `#define` would change all the `DIR` in `lv_fs_fatfs.c` to be `FF_DIR` which is incompatible with that variant of `"ff.h"`.

**Comment by kisvegabor** [comment]
I mean adding the define with a guard. Sorry for being vague. So
```c
#include "ff.h"
#ifdef ESP_PLATFORM
#define DIR FF_DIR
#endif
```

Can it work?

liamHowatt:
So there is a preprocessor define for ESP. I think that is the best solution. [Reference](https://docs.espressif.com/projects/esp-idf/en/stable/esp32/api-guides/build-system.html#preprocessor-definitions)

@mgaman @brandon-kim @anujdeshpande thanks for helping. Would someone be able to validate the solution in their ESP IDF environment?

adcr8iv:
> Would someone be able to validate the solution in their ESP IDF environment?

Tested - works as expected @liamHowatt

anujdeshpande:
Tested - works as expected [@liamHowatt](https://github.com/liamHowatt)

kisvegabor (approved):
Thank you all! :+1:

FASTSHIFT (approved):

PR #6080, opened 2024-04-16
feat(tools): Python script for automatically applying version (#6012)

becseya:
Fixes #6012

PR #6133, opened 2024-04-26
docs: simplify the find_version script and fix shellcheck report in it

marex:
Use sed one-liner to extract the version without any temp files.

Fix the following shellcheck report. The cd builtin may fail, handle it gracefully.
```
In scripts/find_version.sh line 3:
SCRIPTPATH="$( cd -- "$(dirname "$0")" >/dev/null 2>&1 ; pwd -P )"
               ^-- SC2164 (warning): Use 'cd ... || exit' or 'cd ... || return' in case cd fails.

Did you mean:
SCRIPTPATH="$( cd -- "$(dirname "$0")" >/dev/null 2>&1 || exit ; pwd -P )"

For more information:
  https://www.shellcheck.net/wiki/SC2164 -- Use 'cd ... || exit' or 'cd ... |...
```

### DIFF
--- a/scripts/find_version.sh
+++ b/scripts/find_version.sh
@@ -1,4 +1,4 @@
 #!/bin/sh
 
-SCRIPTPATH="$( cd -- "$(dirname "$0")" >/dev/null 2>&1 ; pwd -P )"
+SCRIPTPATH="$( cd -- "$(dirname "$0")" >/dev/null 2>&1 || exit ; pwd -P )"
 sed -n '/LVGL_VERSION_MAJOR/ {N;s@#define \+LVGL_VERSION_M.... \+@@g;s@\n@.@p}' "$SCRIPTPATH/../lv_version.h"
--- a/scripts/find_version.sh
+++ b/scripts/find_version.sh
@@ -1,8 +1,4 @@
-#!/bin/bash
-# Credit: https://stackoverflow.com/a/4774063
+#!/bin/sh
+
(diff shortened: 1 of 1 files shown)


kisvegabor:
Hey,

I've fixed it and it works well. Thank you!

kisvegabor (approved):

marex:
> Hey,
> 
> I've ~fixed~ tested it and it works well. Thank you!

@kisvegabor thank you.

I have noticed the MRs are all over the place, so let me try to create some sort of order in it:
- #6133 - probably useful in short term, can be deferred
- #6135 - useful as a stopgap measure to fix builds on systems without bourne shell
- #6080 - seems like the ultimate final solution

Do I have it right ?

FASTSHIFT (approved):

kisvegabor:
Yes, this issue have many branches now. 

I'll merge this PR as a hotfix for non-Linux but https://github.com/lvgl/lvgl/pull/6135 should be superior.

I thought https://github.com/lvgl/lvgl/pull/6080 is not really related as it should help only in releasing to update the versions on all the places. However, with the help this scrip we can hardcode the version in the docs too, and update it with a script before releases. 

cc @kdschlosser @becseya

kisvegabor:
And we have #6143 too. 

Finally https://github.com/lvgl/lvgl/pull/6080 should solve all the issues if the version id hardcoded in the docs.

PR #6135, opened 2024-04-26
fix(docs): pull version out of lv_version.h in python

marex:
Pull version out of lv_version.h in python instead of calling
a shell script, to retain compatibility with various platforms.

Paging @embeddedt and @kisvegabor . This revert is not great, sorry. I suspect this will fix breakage for @kdschlosser .

### DIFF
--- a/docs/build.py
+++ b/docs/build.py
@@ -207,7 +207,7 @@ else:
 
 
 def get_version():
-    path = os.path.join(project_path, 'lvgl.h')
+    path = os.path.join(project_path, 'lv_version.h')
     with open(path, 'rb') as f:
         d = f.read().decode('utf-8')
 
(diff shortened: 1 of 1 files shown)


kisvegabor:
It would be nice to use the same script on multiple places to simplify maintenance. What about converting `find_version.sh` to `find_version.py`?

kdschlosser:
I will make the changed needed and submit a PR for it.

marex:
> It would be nice to use the same script on multiple places to simplify maintenance. What about converting `find_version.sh` to `find_version.py`?

@kisvegabor I assume #6080 is already implementing that ?

becseya:
I'd suggest closing this PR as #6143 contains the (functionally) same change

kisvegabor:
Closing in favor of https://github.com/lvgl/lvgl/pull/6143#issuecomment-2098879768

Issue #6145, opened 2024-04-29
Arabic and Hebrew percent sign display issues under RTL layout


qinshijing:
### LVGL version

v9.1.1

### What happened?

Both Arabic and Hebrew have RTL layouts. The percent sign in Hebrew is on the right side of the number, and the percent sign in Arabic is on the left side of the number. The RTL layout will force the percent sign to the left side of the number. as follows:
![image](https://github.com/lvgl/lvgl/assets/51692568/2c79e466-e2b9-4f61-a942-5e029f41d0ec)
![image](https://github.com/lvgl/lvgl/assets/51692568/87b3b181-eb1b-4e0d-ac5a-059d5116ef2c)

```C
const char* Arab_string =
"زيادة 56%";


const char* Hebrew_string = 
"עלייה של 56%.";

void lv_freetype_demo(void)
{
    /*Create a font*/
  lv_font_t* font = lv_freetype_font_create("./NotoSansArabic-Bold.ttf",
                                            LV_FREETYPE_FONT_RENDER_MODE_BITMAP,
                                            24, LV_FREETYPE_FONT_STYLE_NORMAL);

  lv_font_t* font2 = lv_freetype_font_create("./NotoSansHebrew-Bold.ttf",
                                            LV_FREETYPE_FONT_RENDER_MODE_BITMAP,
                                            24, LV_FREETYPE_FONT_STYLE_NORMAL);

  if (!font || !font2) {
    LV_LOG_ERROR("freetype font create failed.");
    return;
    }

    font->fallback = font2;

    lv_obj_set_style_base_dir(lv_screen_active(), LV_BASE_DIR_RTL, 0);

    /*Create style with the new font*/
    static lv_style_t style;
    lv_style_init(&style);
    lv_style_set_text_font(&style, font);
    lv_style_set_text_align(&style, LV_TEXT_ALIGN_CENTER);


    /*Create a label with the new style*/
    lv_obj_t * label = lv_label_create(lv_screen_active());
    lv_obj_add_style(label, &style, 0);
    lv_label_set_text(label, Arab_string);
    lv_label_set_long_mode(label, LV_LABEL_LONG_WRAP);
    lv_obj_set_width(label, 400);


    label = lv_label_create(lv_screen_active());
    lv_obj_add_style(label, &style, 0);
    lv_label_set_text(label, Hebrew_string);
    lv_label_set_long_mode(label, LV_LABEL_LONG_WRAP);
[...]

PR #6146, opened 2024-04-29
feat(bidi): support set neutral string

qinshijing:

When switching languages ​​externally, take the initiative to reset the neutral string. https://github.com/lvgl/lvgl/issues/6145

### Notes

### DIFF
--- a/src/misc/lv_bidi.c
+++ b/src/misc/lv_bidi.c
@@ -64,6 +64,7 @@ static uint32_t get_txt_len(const char * txt, uint32_t max_len);
  **********************/
 static const uint8_t bracket_left[] = {"<({["};
 static const uint8_t bracket_right[] = {">)}]"};
+static const char * custom_neutrals = NULL;
 
 /**********************
  *      MACROS
@@ -281,6 +282,11 @@ void lv_bidi_calculate_align(lv_text_align_t * align, lv_base_dir_t * base_dir, 
     }
 }
 
+void lv_bidi_set_custom_neutrals_static(const char * neutrals)
+{
+    custom_neutrals = neutrals;
+}
+
 /**********************
  *   STATIC FUNCTIONS
  **********************/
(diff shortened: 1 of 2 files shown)


src/misc/lv_bidi.c
```diff
@@ -364,7 +370,11 @@ static bool lv_bidi_letter_is_rtl(uint32_t letter)
 static bool lv_bidi_letter_is_neutral(uint32_t letter)
 {
     uint16_t i;
-    static const char neutrals[] = " \t\n\r.,:;'\"`!?%/\\-=()[]{}<>@#&$|";
+    const char * neutrals = " \t\n\r.,:;'\"`!?%/\\-=()[]{}<>@#&$|";
+    if(custom_neutrals) {
+        neutrals = custom_neutrals;
+    }
+
```
kisvegabor, line 377: This way we cannot easily revert to the default.  I suggest
```suggestion
    const char * default_neutrals = " \t\n\r.,:;'\"`!?%/\\-=()[]{}<>@#&$|";
    if(custom_neutrals) {
        neutrals = custom_neutrals;
    } else  {
       neutrals =  default_neutrals;
    }

```

kisvegabor (commented):
Thank you for the PR, but first I would like to understand what the root of the issue is. 
See [here](https://github.com/lvgl/lvgl/issues/6145#issuecomment-2084798028).

src/misc/lv_bidi.c
qinshijing: The current method is that as long as custom_neutrals is set to NULL, the default neutrals will be used.

kisvegabor: In Line 375 the original `neutrals` are overwritten with `custom_neutrals`, so there is no way to get back the default settings anymore.

qinshijing: > neutrals

I don't quite understand, neutrals is no longer a static pointer, why there is no way to get back the default settings anymore?

kisvegabor: I was stupid, sorry :smile: Somehow I thought it's  `static const char * neutrals = ...`.

kisvegabor:
I made some more research an it turned out that it's really language specific. So let's go with your suggestion.

kisvegabor (approved):

FASTSHIFT (approved):

PR #6169, opened 2024-05-02
fix(event): fix LV_EVENT_PREPROCESS flag was not processed correctly on VisualStudio

ymh-kotaro-terada:

Related to #2985 #3003 

Fix #3003 was not working properly on VisualStudio.

On VisualStudio, the type size of lv_event_code_t was not the size expected by #3003.
Therefore, it was fixed to explicitly limit the size.

This problem has already been reported in the following thread.
https://github.com/lvgl/lvgl/pull/3003#issuecomment-1941396211

### DIFF
--- a/src/core/lv_event.c
+++ b/src/core/lv_event.c
@@ -20,7 +20,7 @@
 typedef struct _lv_event_dsc_t {
     lv_event_cb_t cb;
     void * user_data;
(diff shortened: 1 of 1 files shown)


kisvegabor:
Thank you, I've cherry-picked is to release/v8.4 too.

PR #6170, opened 2024-05-02
feat(draw_sw): implemented radial gradient background

zjanosy:

Implemented radial gradient background. Work in progress. For internal testing only.

### Notes

Added example lv_example_styles_16.c. To test you need to add
`#define LV_DRAW_SW_COMPLEX_GRADIENTS    1`
to lv_conf.h, and add
```extern "C" { void lv_example_style_16(void); }```
to main.

### CODE REVIEW

File: examples/styles/lv_example_style_16.c [hunk 1]
```diff
@@ -0,0 +1,86 @@
...
+void lv_example_style_16(void)
+{
+    static lv_style_t style;
+    lv_style_init(&style);
+    lv_style_set_radius(&style, 5);
+
+    /*Make a radial gradient*/
+    lv_style_set_bg_opa(&style, LV_OPA_COVER);
+    static lv_grad_dsc_t grad;
+#ifdef USE_RADIAL_GRADIENT
```
kisvegabor: Please make these cases separate examples.

src/draw/sw/lv_draw_sw_gradient.c
```diff
@@ -25,12 +25,48 @@
...
+ **********************/
+
+#if LV_USE_DRAW_SW_COMPLEX_GRADIENTS
+
+typedef struct {
+    /* w = (-b(xp, yp) + sqrt(sqr(b(xp, yp)) - 4 * a * c(xp, yp))) / (2 * a) */
+    int32_t x0;         /* center of the start circle */
+    int32_t y0;         /* center of the start circle */
+    int32_t r0;         /* radius of the start circle */
+    int32_t inv_dr;     /* 1 / (r1 - r0) */
```
kisvegabor, line 40: `r1` is not defined here.

src/draw/sw/lv_draw_sw_gradient.c
```diff
@@ -137,4 +184,319 @@ void lv_gradient_cleanup(lv_grad_t * grad)
...
+#elif 1
+
+#endif
+
+static inline int32_t sqr32(int32_t x)
+{
+    return x * x;
+}
+
+void lv_gradient_radial_setup(lv_grad_dsc_t * dsc)
```
kisvegabor: Please fix these trivial build errors.

kisvegabor (commented):
Thank you, just a few minor comments at first look.

src/draw/sw/lv_draw_sw_gradient.c
zjanosy: r1 is not needed during the rendering.

src/draw/sw/lv_draw_sw_gradient.c
zjanosy: Hmmm... I don't see why this is generated.

zjanosy: Ah, because I used #ifdef instead of #if.

examples/styles/lv_example_style_17.c
```diff
@@ -0,0 +1,56 @@
...
+
+    /*Define radial gradient by the start and end limit circles. Here we use two limit circles with the start circle being within the end circle, but at an offset, and repeating the gradient outside the end circle (all coordinates are relative to the object)*/
+    grad.dir = LV_GRAD_DIR_RADIAL;
+    grad.r.start.x = width / 2;                     /*start circle center x position*/
+    grad.r.start.y = height / 2;                    /*start circle center y position*/
+    grad.r.start_radius = 10;                       /*start circle radius*/
+    grad.r.end.x = grad.r.start.x - 20;             /*end circle center x position*/
+    grad.r.end.y = grad.r.start.y - 20;             /*end circle center y position*/
+    grad.r.end_radius = 70;                         /*end circle radius*/
+    grad.extend = LV_GRAD_EXTEND_REPEAT;            /*repeat color pattern outside the end circle*/
```
kisvegabor: Please also an example for using multiple gradient.

src/misc/lv_style.h
```diff
@@ -152,10 +168,39 @@ typedef struct {
...
-    uint8_t              stops_count;                  /**< The number of used stops in the array */
-    lv_grad_dir_t        dir : 3;                      /**< The gradient direction.
-                                                        * Any of LV_GRAD_DIR_HOR, LV_GRAD_DIR_VER, LV_GRAD_DIR_NONE */
+    lv_gradient_stop_t   stops[LV_GRADIENT_MAX_STOPS];  /**< A gradient stop array */
+    uint8_t              stops_count;                   /**< The number of used stops in the array */
+    lv_grad_dir_t        dir : 3;                       /**< The gradient direction.
+                                                         * Any of LV_GRAD_DIR_NONE, LV_GRAD_DIR_VER, LV_GRAD_DIR_HOR,
+                                                           LV_GRAD_TYPE_LINEAR, LV_GRAD_TYPE_RADIAL, LV_GRAD_TYPE_CONICAL */
+    lv_grad_extend_t     extend : 2;                    /**< Behaviour outside the defined range.
+                                                         * LV_GRAD_EXTEND_NONE, LV_GRAD_EXTEND_PAD, LV_GRAD_EXTEND_REPEAT, LV_GRAD_EXTEND_REFLECT */
```
kisvegabor, line 177: Do we really support all these?

src/misc/lv_style.h
```diff
@@ -152,10 +168,39 @@ typedef struct {
...
+#if LV_USE_DRAW_SW_COMPLEX_GRADIENTS
+    union {
+        /*Linear gradient parameters*/
+        struct {
+            lv_point_t  start;                          /**< Linear gradient vector start point */
+            lv_point_t  end;                            /**< Linear gradient vector end point */
+        } l;
```
kisvegabor: Abbreviations shall be avoided.

src/draw/sw/lv_draw_sw_gradient.c
```diff
@@ -55,6 +98,120 @@ static lv_grad_t * allocate_item(const lv_grad_dsc_t * g, int32_t w, int32_t h)
     return item;
 }
 
+#if LV_USE_DRAW_SW_COMPLEX_GRADIENTS
+
+/*
+// Integer Square Root function
```
kisvegabor:
```suggestion
 Integer Square Root function
```

src/draw/sw/lv_draw_sw_gradient.c
```diff
@@ -55,6 +98,120 @@ static lv_grad_t * allocate_item(const lv_grad_dsc_t * g, int32_t w, int32_t h)
...
+{
+    if(w < 0)
+        return 0;
+    if(extend == LV_GRAD_EXTEND_PAD) {                  /**< Repeat the same color*/
+        return LV_MIN(w, 255);
+    }
+    if(extend == LV_GRAD_EXTEND_REPEAT) {       /**< Repeat the pattern*/
+        return w & 255;
+    }
+    w &= 511;
```
kisvegabor, line 132: Is this `LV_GRAD_EXTEND_REPEAT`?

zjanosy: I don't understand the question.

src/misc/lv_style.h
zjanosy: Well, that's the plan. Linear at an angle and conical are not yet supported, but I wanted to be prepared for those, too.

src/misc/lv_style.h
zjanosy: What do you mean?

kisvegabor: I mean
```suggestion
        } linear;
```

src/draw/sw/lv_draw_sw_gradient.c
kisvegabor: Sorry, I meant `LV_GRAD_EXTEND_REFLECT`. So is this part of the code implements `LV_GRAD_EXTEND_REFLECT`? If so, I think we should add it in it's own `if` for better readability.

zjanosy: This was a minor performance optimization. First I've implemented this function with a switch(), but the overhead seemed to be significant, compared to this implementation.

kisvegabor:
@zjanosy 
It seems incredibly powerful! Please add some tests so that we can also see the whole power of this feature and also fix the CI failures. :slightly_smiling_face:

src/draw/sw/lv_draw_sw_gradient.h
```diff
@@ -25,6 +25,12 @@ extern "C" {
 #error LVGL needs at least 2 stops for gradients. Please increase the LV_GRADIENT_MAX_STOPS
 #endif
 
+#define LV_GRAD_LEFT    LV_PCT(0)
+#define LV_GRAD_RIGHT   LV_PCT(100)
+#define LV_GRAD_TOP     LV_PCT(0)
+#define LV_GRAD_BOTTOM  LV_PCT(100)
+#define LV_GRAD_CENTER  LV_PCT(50)
```
kisvegabor, line 32: IMO these are not required. Using `lv_pct()` directly in the examples better expresses what is happening and it's more informative for the users too.

src/draw/sw/lv_draw_sw_gradient.h
```diff
@@ -58,6 +64,140 @@ lv_grad_t * lv_gradient_get(const lv_grad_dsc_t * gradient, int32_t w, int32_t h
...
+/**
+ * Initialize gradient color map from a table
+ * @param grad      pointer to a gradient descriptor
+ * @param colors    color array
+ * @param fracs     position array (0..255): if NULL, then colors are distributed evenly
+ * @param opa       opacity array: if NULL, then LV_OPA_COVER is assumed
+ * @param num_stops number of gradient stops (1..LV_GRADIENT_MAX_STOPS)
+ */
+void lv_gradient_init_stops(lv_grad_dsc_t * grad, const lv_color_t colors[], const uint8_t fracs[],
+                            const lv_opa_t opa[], int num_stops);
```
kisvegabor: Having the color and opa next to each other seems more logical
```suggestion
void lv_gradient_init_stops(lv_grad_dsc_t * grad, const lv_color_t colors[], const lv_opa_t opa[], 
                              const uint8_t fracs[], int num_stops);
```

src/draw/sw/lv_draw_sw_gradient.h
zjanosy: This was modeled after the CSS keywords.

src/draw/sw/lv_draw_sw_gradient.h
zjanosy: My reasoning was that for complex gradient color maps specifying the stop positions may be more common than using opacity, but certainly both ways make sense.

kisvegabor: I see it like `array1, array_len, array2`, and `array1, array2, array_len` is cleaner to me. On other APIs we put `len` to the end too, so let's follow this here as well.

src/draw/sw/lv_draw_sw_gradient.h
kisvegabor: Okay, let's keep them.

src/draw/sw/lv_draw_sw_gradient.h
zjanosy: Currently it is `array1, array2, array3, array_len`, i.e., the array length is the last argument. Your original comment was about swapping `array2` (fracs) and `array3` (opa). Now I don't understand what is wrong. Shall I swap `fracs` and `opa`?

kisvegabor: Oh, I'm sorry. I got confused. Originally I just meant that the color and opa should be togother.

kisvegabor (dismissed):

src/draw/sw/lv_draw_sw_gradient.c
```diff
@@ -137,4 +219,450 @@ void lv_gradient_cleanup(lv_grad_t * grad)
     lv_free(grad);
 }
 
+void lv_gradient_init_stops(lv_grad_dsc_t * grad, const lv_color_t colors[], const uint8_t fracs[],
+                            const lv_opa_t opa[], int num_stops)
```
liamHowatt: Match the header prototype. I caught this because the render demo broke.
```suggestion
void lv_gradient_init_stops(lv_grad_dsc_t * grad, const lv_color_t colors[], const lv_opa_t opa[],
                            const uint8_t fracs[], int num_stops)
```

demos/render/lv_demo_render.c
```diff
@@ -818,6 +818,238 @@ static void blend_mode_cb(lv_obj_t * parent)
...
+    lv_obj_set_grid_dsc_array(parent, grid_cols, grid_rows);
+
+    const char * opa_txt[] = { "no opa", "no opa round", "stop opa", "stop opa round" };
+    int32_t radius_values[] = { 0, 20, 0, 20 };
+    bool opa_map_values[] = { false, false, true, true };
+
+    const char * offs_txt[] = { "map", "repeat", "reflect" };
+    int32_t x1_values[] = { lv_pct(100), lv_pct(15), lv_pct(30)};
+    int32_t y1_values[] = { lv_pct(100), lv_pct(30), lv_pct(15) };
+    lv_grad_extend_t extend_values[] = { LV_GRAD_EXTEND_PAD, LV_GRAD_EXTEND_REPEAT, LV_GRAD_EXTEND_REFLECT };
```
liamHowatt: Should `"map"` be `"pad"`? Or should `LV_GRAD_EXTEND_PAD` be `LV_GRAD_EXTEND_MAP`?

src/misc/lv_style.h
```diff
@@ -152,10 +168,38 @@ typedef struct {
...
+        /*Conical gradient parameters*/
+        struct {
+            lv_point_t  center;                         /**< Conical gradient center point */
+            int16_t     start_angle;                    /**< Start angle 0..3600 */
+            int16_t     end_angle;                      /**< End angle 0..3600 */
+        } conical;
+    } params;
+    void * state;
+#endif
+    struct _lv_grad_dsc_t * next;                       /**< Next gradient to render more gradients*/
```
liamHowatt: Is this going to be used?

liamHowatt (commented):
Not done reviewing yet but here are some comments so far.

src/misc/lv_style.h
zjanosy: Adding this linked list pointer was originally suggested by Gabor. It is not yet implemented, but it would be a useful feature. Maybe we can remove it for now.

demos/render/lv_demo_render.c
zjanosy: Yes, actually the string should be `"pad"`. The #define is correct. I always mix those.

liamHowatt:
@kisvegabor, FYI, another instance of CI failing due to a memory error in the RLE decompressor. This time on 32 bit x86. Before it was 64 bit ARM.

kisvegabor (approved):

liamHowatt (approved):

kisvegabor:
Let's have something merged first and we can improve it later as we will have some feedback :slightly_smiling_face:

kisvegabor:
Thank you @zjanosy! :blush:

PR #6176, opened 2024-05-04
fix(lv_msgbox): Automatically adjust msgbox's content height.

TridentTD:

image

When msgbox is resized, such as:

```
lv_obj_set_size(msgbox, lv_pct(80), lv_pct(80));
```

the height of the msgbox's content will automatically adjust,
and the scrollable feature will be removed from msgbox's buttons.

### Notes


src/widgets/msgbox/lv_msgbox.c
```diff
@@ -145,6 +146,7 @@ lv_obj_t * lv_msgbox_add_title(lv_obj_t * obj, const char * title)
         lv_obj_set_size(mbox->header, lv_pct(100), lv_display_get_dpi(lv_obj_get_display(obj)) / 3);
         lv_obj_set_flex_flow(mbox->header, LV_FLEX_FLOW_ROW);
         lv_obj_set_flex_align(mbox->header, LV_FLEX_ALIGN_START, LV_FLEX_ALIGN_CENTER, LV_FLEX_ALIGN_CENTER);
+        lv_obj_remove_flag(mbox->header, LV_OBJ_FLAG_SCROLLABLE);
```
PierreRambaud, line 150: I don't think scroll for header and footer should be disable by default :thinking:

TridentTD: Thank you for the review.

If the header of the msgbox is scrollable, 
I noticed that when accidentally touched directly on the header, 
it will have a similar appearance to this clip.

https://www.veed.io/view/047f1414-a8e6-4d1b-952d-c507a0349ff0?panel=quality-survey

I still can't recall of an example of a case 
where the header (or titlebar) of msgbox has a scrollbar. 
Could you provide an example of a titlebar/header with a scrollbar?

PierreRambaud: I was just thinking about a scrollbar on a little screen with a lot of buttons (bad interface, but still capable of).
Since elements can be retrieved by using `lv_msgbox_get_header`, maybe the scrollbar flag can be removed by the user himself.

Maybe Core reviewers have something different in mind but even the `lv_obj_set_flex_grow(mbox->content, 1);` can be resolved by doing:

```c
lv_obj_t* mbox = lv_msgbox_create(NULL);
lv_obj_set_flex_grow(lv_msgbox_get_content(mbox), 1);
```

tests/src/test_cases/widgets/test_msgbox.c
```diff
@@ -44,7 +44,7 @@ void test_msgbox_creation_successful_with_close_button(void)
 
     TEST_ASSERT_NOT_NULL(msgbox);
 
-    TEST_ASSERT_EQUAL_SCREENSHOT("widgets/msgbox_ok_with_close_btn.png");
+    TEST_ASSERT_EQUAL_SCREENSHOT("ref_imgs/widgets/msgbox_ok_with_close_btn.png");
```
kisvegabor: "ref_imgs/" is appended automatically [here](https://github.com/lvgl/lvgl/blob/master/tests/unity/unity_support.c#L141) so the original path should be correct.

tests/src/test_cases/widgets/test_msgbox.c
```diff
@@ -144,4 +144,38 @@ void test_msgbox_close_async_modal(void)
     TEST_ASSERT_NOT_NULL(msgbox);
 }
 
+void test_msgbox_content_auto_height(void)
+{
+    // If parent is NULL the message box will be modal
```
kisvegabor: We use `/**/` comments:
```suggestion
    /* If parent is NULL the message box will be modal */
```

tests/src/test_cases/widgets/test_msgbox.c
```diff
@@ -144,4 +144,38 @@ void test_msgbox_close_async_modal(void)
...
+    int32_t h_content = lv_obj_get_height(content);
+
+    int32_t h_obj_content = lv_obj_get_content_height(msgbox);
+    int32_t h_msgbox_element_sum  = h_header + h_footer + h_content;
+
+    TEST_ASSERT_NOT_NULL(msgbox);
+
+    TEST_ASSERT_EQUAL(h_obj_content, h_msgbox_element_sum);
+
+    // Since msgbox has no parent, it won´t be clean up at tearDown()
```
kisvegabor:
```suggestion
    /* Since msgbox has no parent, it won´t be clean up at tearDown() */
```

kisvegabor (commented):
Hi,

Thanks for the suggestion.

I've just tested it, but with this update `lv_example_msgbox_1` is not working as the default `LV_SIZE_CONTENT` height of the message box and `flex_grow` on an element is conflicting. (It's a circular dependency)

I would like to keep the current set-height-to-the-content behaviour, but I also see the value in your suggestion. What about creating a new example, where some controls are added to the message box (slider, text input, etc) and the height is set manually?

I agree with removing the  SCROLLABLE flag.

tests/src/test_cases/widgets/test_msgbox.c
TridentTD: image

Now , support  both  msgbox's default size (by LV_SIZE_CONTENT)  and  msgbox' manual size.

However, for `test_msgbox.c` in the `test_cases` folder, 
when I tested whether the msgbox's height is set to LV_SIZE_CONTENT or not, 
it's OK. 

image

But when the command 
```
lv_obj_update_content(msgbox); 
```
is added, an error occurs. 

image

I don't know what caused the error in `test_msgbox.c`.

kisvegabor:
Thank you for adding the test. I fixed the crash and few other minor things in [17c9fdc](https://github.com/lvgl/lvgl/pull/6176/commits/17c9fdca60e1a45412ba5d6e8fd29cdc2b6844b9)

TridentTD:
Thank you for fixing the crash.

kisvegabor:
No problem, I've reset it to the `687691a`.

TridentTD:
Thank you.

liamHowatt (approved):

PierreRambaud (approved):

kisvegabor (approved):

XuNeo (approved):

Issue #6059, opened 2024-04-11
close the frame, break!


xinsuinizhuan:
### LVGL version

v9.1.1

### What happened?

in windows，when close the pop window，break:
![图片](https://github.com/lvgl/lvgl/assets/40679769/7a1b6a65-f096-4c1b-bb91-13bd278936d9)
because the disp_refr_timer is NULL in lv_sysmon.c

xinsuinizhuan:
to add a judget:
![图片](https://github.com/lvgl/lvgl/assets/40679769/b18669c5-9377-4858-9ba2-85d9ace47315)

liamHowatt:
Hi, your default display is not set for some reason. Did you create one? If so, did you use `lv_sdl_window_create` or `lv_windows_create_display`?

liamHowatt:
Also, since you appear to have a debugging environment, could you please check to see if you reach this line and whether the returned timer is NULL?

https://github.com/lvgl/lvgl/blob/eb62ddf4ea8e84ea76cc89d932ce20f0066b8618/src/display/lv_display.c#L95

xinsuinizhuan:
> Hi, your default display is not set for some reason. Did you create one? If so, did you use `lv_sdl_window_create` or `lv_windows_create_display`?

 i set #define LV_USE_OS   LV_OS_WINDOWS in lv_conf.h and i use lv_windows_create_display:
![图片](https://github.com/lvgl/lvgl/assets/40679769/a2afcf44-b6f1-4720-9ce9-a935d6ea3cbf)

xinsuinizhuan:
> disp->refr_timer = lv_timer_create

it not NULL:
![图片](https://github.com/lvgl/lvgl/assets/40679769/43604c09-3239-4468-80a3-103c7ee1ae92)

liamHowatt:
Strange. I am going to use Windows soon and will look at this. You seem comfortable with debugging. I welcome you to investigate the issue if you want.

liamHowatt:
Yes! I reproduced the problem.

[...]

PR #6182, opened 2024-05-05
feat(demos): add demo for the OSAL

### LINKED ISSUE #6049: Demo for the OSAL
### Problem to solve

During development of LVGL integrations one task with the new release is the support of the OSAL. It would be very useful to have tests of some sort that integrators could refer to to verify their implementation. 

### Success criteria

LVGL integrators have a convenient way of verifying their OSAL implementation

### Solution outline

I would propose this to be a demo rather then a feature though, because there might be customer preferences as to what framework to use. A demo could be vendor neutral way to provide the results by using print statements and `LV_ASSERT` etc.

### Rabbit holes

None that I know of.

### Testing

Running integrations tests (at least for a linux pthread osal) within the pipeline.

### Teaching

Adding a note to the OS porting section and of course to the demos.

### Considerations

_No response_

faxe1008:
Adds a demo demonstrating the use of the OSAL.

Resolves: #6049

### Notes


demos/osal/lv_demo_osal.c
```diff
@@ -0,0 +1,88 @@
...
+}
+
+void increment_thread_entry(void * user_data)
+{
+    lv_obj_t * counter_label;
+    uint32_t press_count = 0;
+
+    counter_label = lv_label_create(lv_scr_act());
+    lv_obj_align(counter_label, LV_ALIGN_CENTER, 0, 0);
+    lv_label_set_text_fmt("Pressed %u times", press_count);
```
kisvegabor:
```suggestion
    lv_label_set_text_fmt(counter_label, "Pressed %u times", press_count);
```

demos/osal/lv_demo_osal.c
```diff
@@ -0,0 +1,88 @@
...
+void counter_button_event_cb(lv_event_t * e)
+{
+    if(lv_thread_sync_signal(&press_sync) != LV_RESULT_OK) {
+        LV_LOG_ERROR("Error signaling thread sync");
+    }
+}
+
+void increment_thread_entry(void * user_data)
+{
+    lv_obj_t * counter_label;
```
kisvegabor: Add `LV_UNUSED(user_data);`

demos/osal/lv_demo_osal.c
```diff
@@ -0,0 +1,88 @@
...
+    counter_label = lv_label_create(lv_scr_act());
+    lv_obj_align(counter_label, LV_ALIGN_CENTER, 0, 0);
+    lv_label_set_text_fmt("Pressed %u times", press_count);
+
+    while(true) {
+        if(lv_thread_sync_wait(&press_sync) != LV_RESULT_OK) {
+            LV_LOG_ERROR("Error awaiting thread sync");
+        }
+        press_count += 1;
+        lv_label_set_text_fmt("Pressed %u times", press_count);
```
kisvegabor:
```suggestion
        lv_label_set_text_fmt(counter_label, "Pressed %u times", press_count);
```

kisvegabor (commented):
Thank you Fabian! 

[Here](https://github.com/lvgl/lvgl/pull/6182/files) you can see some errors reported by CI. Fortunately, they are easy to fix.

I feel like this update could be rather an example. Usually we add complex eye catching things in the `demos` folder. So would you mind moving it to the `examples` folder?

demos/osal/lv_demo_osal.c
```diff
@@ -0,0 +1,88 @@
...
+    }
+}
+
+void increment_thread_entry(void * user_data)
+{
+    lv_obj_t * counter_label;
+    uint32_t press_count = 0;
+
+    counter_label = lv_label_create(lv_scr_act());
+    lv_obj_align(counter_label, LV_ALIGN_CENTER, 0, 0);
```
kisvegabor: Don't we need a mutex here? What if this thread is running concurrently with the thread of `lv_timer_handler`?

faxe1008:
Moved this to `examples/porting` as it felt the most appropriate. How can I make sure that this is build during CI?

demos/osal/lv_demo_osal.c
faxe1008: True, but this would mean that the `lv_timer_handler` thread would also need to take this lock. I do not understand how this can be achieved though.

examples/porting/osal/lv_example_osal.c
```diff
@@ -0,0 +1,89 @@
+/**
+ * @file lv_example_osal.c
+ *
+ */
+
+/*********************
+ *      INCLUDES
+ *********************/
+#include "lv_example_osal.h"
```
liamHowatt:
```suggestion
#include "../../lv_examples.h"
```

examples/porting/osal/lv_example_osal.c
```diff
@@ -0,0 +1,89 @@
...
+    }
+}
+
+/**********************
+ *   STATIC FUNCTIONS
+ **********************/
+
+void counter_button_event_cb(lv_event_t * e)
+{
+    if(lv_thread_sync_signal(&press_sync) != LV_RESULT_OK) {
```
liamHowatt:
```suggestion
{
    LV_UNUSED(e);
    if(lv_thread_sync_signal(&press_sync) != LV_RESULT_OK) {
```

liamHowatt (commented):
Add
```c
#include "porting/osal/lv_example_osal.h"
```
to lv_examples.h

demos/osal/lv_demo_osal.c
liamHowatt: Correct. See the latest comments in #6059

.devcontainer/__lv_conf.h__
```diff
@@ -938,6 +938,9 @@
 
 /*Vector graphic demo*/
 #define LV_USE_DEMO_VECTOR_GRAPHIC  0
+
+/*Demonstrate OSAL*/
+#define LV_USE_DEMO_OSAL          1
```
kisvegabor:
```suggestion
```

kisvegabor (commented):
If we want to run it with the CI we need to add it [here](https://github.com/lvgl/lvgl/tree/master/tests/src/test_cases). However now the CI is running only with `pthread`. Do you think it still makes sens to run it as a test?

kisvegabor:
`lv_conf_internal.h` is generated by `scripts/lv_conf_internal_gen.py` and shouldn't be edited by manually.

demos/osal/lv_demo_osal.c
faxe1008: So if I followed the disussion correctly there will be some sort of global mutex that may be taken on the PR for that is ready, right?

liamHowatt: Yes, after that PR, there will be a new API you can use in this PR to acquire the global mutex.

kisvegabor:
The original example was failing for me so I added a little bit update. With the help `lv_lock` and `lv_unlock` it's easier to create a thread safe app. 

What do you think?

src/osal/lv_os.h
```diff
@@ -143,6 +143,28 @@ lv_result_t lv_thread_sync_signal(lv_thread_sync_t * sync);
...
+ * different than lv_timer_handler's thread. It doesn't need to be called in LVGL events because
+ * they are called from lv_timer_handler().
+ * It is called internally in lv_timer_handler().
+ */
+void lv_lock(void);
+
+/**
+ * Same as `lv_lock()` but can be called from an interrupt.
+ */
+void lv_lock_isr(void);
```
liamHowatt:
```suggestion
lv_result_t lv_lock_isr(void);
```

Please see my review.

liamHowatt (commented):
Nice @kisvegabor, I think yes, this PR is actually a good place to add the global lock, so this PR doesn't have to wait for another PR. After this, we just have to update the Windows driver. Thanks!!

It surprised me that we have `lv_mutex_lock_isr`. If it has to block, it will deadlock, because the OS can't suspend the ISR to switch to the thread holding it the lock.

We implement it with `xSemaphoreTakeFromISR` in our FreeRTOS OSAL. Here's the description from the FreeRTOS docs ([link](https://www.freertos.org/xSemaphoreTakeFromISR.html)):

> A version of [xSemaphoreTake()](https://www.freertos.org/a00122.html) that can be called from
an ISR.  Unlike xSemaphoreTake(), xSemaphoreTakeFromISR() does not permit a block
time to be specified.

So it's okay. It's just fallible. I think we can remove the `LV_LOG_ERROR` since it's a possible and expected failure.

However for cmsis_rtos2 we call `osMutexAcquire` which according to the docs, always fails in an ISR ([link](https://arm-software.github.io/CMSIS_5/latest/RTOS2/html/group__CMSIS__RTOS__MutexMgmt.html#gabc54686ea0fc281823b1763422d2a924)):

> osErrorISR: cannot be called from interrupt service routines.

> Mutex management functions cannot be called from Interrupt Service Routines (ISR), unlike a binary semaphore that can be released from an ISR.

All that to say, my feedback for your changes for now is to make the new function `lv_lock_isr` return `lv_result_t` and we can merge it.

src/osal/lv_os.h
kisvegabor: Okay, done!

liamHowatt (approved):
Thank you faxe1008 and kisvegabor!

faxe1008:
Thanks for moving the PR forward and making LVGL thread safe @kisvegabor, this is very good stuff 👍

FASTSHIFT (approved):

PR #6187, opened 2024-05-06
feat(indev): detect double and triple click (closes #6020)

### LINKED ISSUE #6020: Double and triple click support
### Problem to solve

Continue #5351.
The users get used to the double and triple click as they are commonly used on PC. 

### Success criteria

1. Double and triple click are are recognized by LVGL internally. 
2. It works for all input device type

### Solution outline

In `lv_indev.c` a generic method can be implemented to handle press, click, release, short click, long press, long press repeat, double click and triple click. This method can be applied to all input devices. 

### Rabbit holes

Nothing for now.

### Testing

Once it's implemented deep testing shall be added for all detected "gestures" and input device types.

### Teaching

Update the docs.

### Considerations

I proposed to support complex cases like the second press after a click, or the long press after a double click, but thinking about it more I think they are not common, so we shouldn't focus on them.

niklasf:
This is a first attempt at detecting double and triple clicks (#6020). It's basically working but needs documentation and testing. I have some questions before proceding:

* Should `LV_EVENT_SHORT_CLICKED`, `LV_EVENT_DOUBLE_CLICKED`, and `LV_EVENT_TRIPLE_CLICKED` be separate events, or should there be something like `int lv_event_get_short_click_streak(lv_event_t*)`?
* Should short/double/triple click be mutually exclusive, or is every double/triple click also a short click?

### DIFF
--- a/src/indev/lv_indev.c
+++ b/src/indev/lv_indev.c
@@ -1419,12 +1419,12 @@ static lv_result_t indev_proc_short_click(lv_indev_t * indev)
     /*Update streak for clicks within small distance and short time*/
     indev->pointer.short_click_streak++;
     if(lv_tick_elaps(indev->pointer.last_short_click_timestamp) > indev->long_press_time) {
-        indev->pointer.short_click_streak = 0;
+        indev->pointer.short_click_streak = 1;
     }
     else if(indev->type == LV_INDEV_TYPE_POINTER || indev->type == LV_INDEV_TYPE_BUTTON) {
         int32_t dx = indev->pointer.last_short_click_point.x - indev->pointer.act_point.x;
         int32_t dy = indev->pointer.last_short_click_point.y - indev->pointer.act_point.y;
-        if(dx * dx + dy * dy > indev->scroll_limit * indev->scroll_limit) indev->pointer.short_click_streak = 0;
+        if(dx * dx + dy * dy > indev->scroll_limit * indev->scroll_limit) indev->pointer.short_click_streak = 1;
     }
 
     indev->pointer.last_short_click_timestamp = lv_tick_get();
@@ -1437,7 +1437,7 @@ static lv_result_t indev_proc_short_click(lv_indev_t * indev)
     }
 
     /*Cycle through single/double/triple click*/
-    switch(indev->pointer.short_click_streak % 3) {
+    switch((indev->pointer.short_click_streak - 1) % 3) {
         case 0:
             return send_event(LV_EVENT_SINGLE_CLICKED, indev_act);
         case 1:
--- a/src/indev/lv_indev.c
+++ b/src/indev/lv_indev.c
@@ -71,6 +71,7 @@ static void indev_encoder_proc(lv_indev_t * i, lv_indev_data_t * data);
 static void indev_button_proc(lv_indev_t * i, lv_indev_data_t * data);
 static void indev_proc_press(lv_indev_t * indev);
 static void indev_proc_release(lv_indev_t * indev);
+static lv_result_t indev_proc_short_click(lv_indev_t * indev);
 static void indev_proc_pointer_diff(lv_indev_t * indev);
 static lv_obj_t * pointer_search_obj(lv_display_t * disp, lv_point_t * p);
 static void indev_proc_reset_query_handler(lv_indev_t * indev);
@@ -485,6 +486,11 @@ uint32_t lv_indev_get_key(const lv_indev_t * indev)
     return key;
 }
 
+uint8_t lv_indev_get_short_click_streak(const lv_indev_t * indev)
+{
+    return indev->pointer.short_click_streak;
+}
+
 lv_dir_t lv_indev_get_scroll_dir(const lv_indev_t * indev)
 {
     if(indev == NULL) return false;
@@ -852,7 +858,7 @@ static void indev_keypad_proc(lv_indev_t * i, lv_indev_data_t * data)
             if(send_event(LV_EVENT_RELEASED, indev_act) == LV_RESULT_INVALID) return;
 
             if(i->long_pr_sent == 0) {
-                if(send_event(LV_EVENT_SHORT_CLICKED, indev_act) == LV_RESULT_INVALID) return;
+                if(indev_proc_short_click(i) == LV_RESULT_INVALID) return;
             }
 
             if(send_event(LV_EVENT_CLICKED, indev_act) == LV_RESULT_INVALID) return;
@@ -863,6 +869,8 @@ static void indev_keypad_proc(lv_indev_t * i, lv_indev_data_t * data)
     }
     indev_obj_act = NULL;
 }
+
+
 
 /**
  * Process a new point from LV_INDEV_TYPE_ENCODER input device
@@ -1015,7 +1023,7 @@ static void indev_encoder_proc(lv_indev_t * i, lv_indev_data_t * data)
                 }
 
                 if(i->long_pr_sent == 0 && is_enabled) {
-                    if(send_event(LV_EVENT_SHORT_CLICKED, indev_act) == LV_RESULT_INVALID) return;
+                    if(indev_proc_short_click(i) == LV_RESULT_INVALID) return;
                 }
 
                 if(is_enabled) {
@@ -1029,7 +1037,7 @@ static void indev_encoder_proc(lv_indev_t * i, lv_indev_data_t * data)
                 if(!i->long_pr_sent || lv_group_get_obj_count(g) <= 1) {
                     if(is_enabled) {
                         if(send_event(LV_EVENT_RELEASED, indev_act) == LV_RESULT_INVALID) return;
-                        if(send_event(LV_EVENT_SHORT_CLICKED, indev_act) == LV_RESULT_INVALID) return;
+                        if(indev_proc_short_click(i) == LV_RESULT_INVALID) return;
                         if(send_event(LV_EVENT_CLICKED, indev_act) == LV_RESULT_INVALID) return;
                     }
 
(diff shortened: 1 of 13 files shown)


kisvegabor:
> The clicked event should be fired the first time of a double or triple click. The second and third click can be eliminated because now you know for sure it's a double click.
However, following with standard web behavior is also recommended.

I think adding `LV_EVENT_SINGLE_CLICKED`, and firing `LV_EVENT_CLICKED` three times on triple click would be a good solution.

niklasf:
Not stale, just momentarily busy with other things.

niklasf:
Additionally firing `LV_EVENT_SINGLE_CLICKED` on the first short click is a nice way to do this while keeping backwards compatibility. Implemented that.

Ready for review, now.

kisvegabor:
Thank you! 

Do you still have time for adding a test too?

src/indev/lv_indev.c
```diff
@@ -1362,6 +1365,35 @@ static void indev_proc_release(lv_indev_t * indev)
...
+
+    /*Update streak for clicks within small distance and short time*/
+    int32_t dx = indev->pointer.last_short_click_point.x - indev->pointer.act_point.x;
+    int32_t dy = indev->pointer.last_short_click_point.y - indev->pointer.act_point.y;
+    if(dx * dx + dy * dy > indev->scroll_limit * indev->scroll_limit ||
+       lv_tick_elaps(indev->pointer.last_short_click_timestamp) > indev->long_press_time) {
+        indev->pointer.short_click_streak = 0;
+    }
+    else {
+        indev->pointer.short_click_streak = (indev->pointer.short_click_streak + 1) % 3;
```
XuNeo: For fourth and coming more clicks, we should saturate the click_streak here instead of reset it to zero.

Considering user click a button for multiple times, the single, double, triple click should be triggered only once, until user stops clicking and timer expires when we can start next round of detection.

src/indev/lv_indev_private.h
```diff
@@ -89,7 +89,10 @@ struct _lv_indev_t {
         lv_area_t scroll_area;
         lv_point_t gesture_sum; /*Count the gesture pixels to check LV_INDEV_DEF_GESTURE_LIMIT*/
         int32_t diff;
-
+        /*Short click streaks*/
+        uint8_t short_click_streak : 2;
+        lv_point_t last_short_click_point;
+        uint32_t last_short_click_timestamp;
```
XuNeo: Let's keep the bitfields close to each other.
```suggestion

        lv_point_t last_short_click_point;
        uint32_t last_short_click_timestamp;        
        /*Short click streaks*/
        uint8_t short_click_streak : 2;
```

src/indev/lv_indev.c
```diff
@@ -1362,6 +1365,35 @@ static void indev_proc_release(lv_indev_t * indev)
...
+    lv_result_t res = send_event(LV_EVENT_SHORT_CLICKED, indev_act);
+    if(res == LV_RESULT_INVALID) {
+        return res;
+    }
+
+    /*Update streak for clicks within small distance and short time*/
+    int32_t dx = indev->pointer.last_short_click_point.x - indev->pointer.act_point.x;
+    int32_t dy = indev->pointer.last_short_click_point.y - indev->pointer.act_point.y;
+    if(dx * dx + dy * dy > indev->scroll_limit * indev->scroll_limit ||
+       lv_tick_elaps(indev->pointer.last_short_click_timestamp) > indev->long_press_time) {
```
XuNeo: Check time is cheaper than distance, it can optimize the processing time for usual short clicks .

```suggestion
    if(lv_tick_elaps(indev->pointer.last_short_click_timestamp) > indev->long_press_time ||
       dx * dx + dy * dy > indev->scroll_limit * indev->scroll_limit
       ) {
```

src/indev/lv_indev.c
```diff
@@ -1362,6 +1365,35 @@ static void indev_proc_release(lv_indev_t * indev)
...
+    else {
+        indev->pointer.short_click_streak = (indev->pointer.short_click_streak + 1) % 3;
+    }
+    indev->pointer.last_short_click_timestamp = lv_tick_get();
+    indev->pointer.last_short_click_point = indev->pointer.act_point;
+
+    /*Fire single/double/triple click*/
+    if(indev->pointer.short_click_streak == 0) send_event(LV_EVENT_SINGLE_CLICKED, indev_act);
+    if(indev->pointer.short_click_streak == 1) send_event(LV_EVENT_DOUBLE_CLICKED, indev_act);
+    if(indev->pointer.short_click_streak == 2) send_event(LV_EVENT_TRIPLE_CLICKED, indev_act);
```
XuNeo:
```suggestion
    else if(indev->pointer.short_click_streak == 1) send_event(LV_EVENT_DOUBLE_CLICKED, indev_act);
    else if(indev->pointer.short_click_streak == 2) send_event(LV_EVENT_TRIPLE_CLICKED, indev_act);
```

src/indev/lv_indev.c
```diff
@@ -1362,6 +1365,35 @@ static void indev_proc_release(lv_indev_t * indev)
...
+static lv_result_t indev_proc_short_click(lv_indev_t * indev)
+{
+    /*Simple short click*/
+    lv_result_t res = send_event(LV_EVENT_SHORT_CLICKED, indev_act);
+    if(res == LV_RESULT_INVALID) {
+        return res;
+    }
+
+    /*Update streak for clicks within small distance and short time*/
+    int32_t dx = indev->pointer.last_short_click_point.x - indev->pointer.act_point.x;
```
XuNeo: Let's check if the `indev` is actually a pointer device before using it.

For encoder, keypad, we can simply ignore the check if distance.
Question:
For keypad, which obj should be used to receive the event?

src/indev/lv_indev.c
niklasf: Saturate such that triple click is fired repeatedly, or such that no further click streak events are fired?

XuNeo: I think no further click streak events should be fired.

kisvegabor: On PC it's restarted: 
clicks

XuNeo: On PC, it could be a value to notify application click counts, instead of resending the single click event.

Resend the single, double click is not what actually happens, thus I think we should ignore futhure clicks for now.

Here's a reference to what web will do 
https://stackoverflow.com/questions/6480060/how-do-i-listen-for-triple-clicks-in-javascript

kisvegabor: It would be really the most flexible if we incremented `short_click_streak` infinitely and added an API to get its count. As the `CLICKED` event is sent [after incrementing](https://github.com/lvgl/lvgl/pull/6187/files#diff-35a05ebbf5595318ed115bcabc7f70e6ae0b4f1f3ace3629e154a7434fe680f1R818-R821) the user can do any custom thing there.

kisvegabor:
Hi @niklasf,

Only a little is missing to merge this PR. Can you get back to this in the near future? :crossed_fingers:

koendv:
A question. Would it make sense for LV_EVENT_DOUBLE_CLICK to default to being clicked twice? 
Else you might click twice, and nothing happens. I don't know, just asking.

```
--- a/src/core/lv_obj.c
+++ b/src/core/lv_obj.c
@@ -729,6 +729,10 @@ static void lv_obj_event(const lv_obj_class_t * class_p, lv_event_t * e)
     if(code == LV_EVENT_PRESSED) {
         lv_obj_add_state(obj, LV_STATE_PRESSED);
     }
+    else if(code == LV_EVENT_DOUBLE_CLICKED) {
+        lv_obj_send_event(obj, LV_EVENT_CLICKED, NULL);
+        lv_obj_send_event(obj, LV_EVENT_CLICKED, NULL);
+    }
     else if(code == LV_EVENT_RELEASED) {
         lv_obj_remove_state(obj, LV_STATE_PRESSED);
         void * param = lv_event_get_param(e);
```
At the very least this PR shows there is a case for adding LV_EVENT_DOUBLE_CLICKED to lv_event_code_t.

kisvegabor:
See [here](https://github.com/lvgl/lvgl/pull/6187/files#diff-35a05ebbf5595318ed115bcabc7f70e6ae0b4f1f3ace3629e154a7434fe680f1R818-R821). The CLICKED event is sent independently of SINGLE/DOUBLE/TRIPLE clicks.

kisvegabor:
@niklasf Are you still interested in working on it?

If you are busy with something else we can pick it up. The implementation is almost ready, however
- Some comments still need to be implemented
- Docs need to be added
- Examples need to be added
- Tests need to be added

src/indev/lv_indev_private.h
niklasf: With the proposed getter for `short_click_streak` that can be used instead of cycling, this can no longer be a bitfield.

src/indev/lv_indev.c
niklasf: Added `lv_indev_get_short_click_streak()` that can be used from short clicked handlers, and kept the cycling for the simple single/double/triple click events.

src/indev/lv_indev.c
niklasf: Now checking the type before the distance.

Not sure about keypads. I assume whichever object used to receive `LV_EVENT_SHORT_CLICKED` should also receive `LV_EVENT_SINGLE_CLICKED` and so on?

niklasf:
Thanks for the review and sorry for the delay. I have now tried to address the feedback, added an example, and expanded the test to include the new `lv_indev_get_short_click_streak()`. Ready for the next round of review.

src/indev/lv_indev.h
```diff
@@ -285,6 +285,15 @@ lv_dir_t lv_indev_get_gesture_dir(const lv_indev_t * indev);
  */
 uint32_t lv_indev_get_key(const lv_indev_t * indev);
 
+
+/**
+ * Get the counter for consecutive clicks within a short distance and time.
+ * The counter is updated before LV_EVENT_SHORT_CLICKED is fired.
+ * @param indev pointer to an input device
+ * @return 0 based short click streak counter
```
kisvegabor:
```suggestion
 * @param indev     pointer to an input device
 * @return          based short click streak counter
```

I suggest making it 1 based because it seems more intuitive.

kisvegabor (dismissed):
Thank you!

kisvegabor (approved):

liamHowatt (approved):
Looks great! Nicely done.

Issue #5350, opened 2024-01-16
Improve the animation API 


kisvegabor:
### Introduce the problem

> This is feature is the pioneer of the [new shaping-pitching](https://docs.lvgl.io/master/CONTRIBUTING.html) project management process. Please, leave your feedback about how effective you find this method, so that we can improve it. 
In this issue we can discuss what we want or don't want to add to this feature. Based on that we will create a [pitch](https://github.com/lvgl/lvgl/blob/pitches/__pitch_template.md) and someone can start the development considering the set boundaries.

As described in #4920 the "create" API of animations doesn't follow the other modules as we have `lv_anim_init` instead of `lv_anim_create`. Without the `create` function the life cycle management of the animations is much harder.

### Proposal

For backward compatibility we should keep the current `init` API too and `lv_anim_create` can set a flag to tell if the animation was malloc-ed in it.

Besides the animations are lacking some more controls too:
- pause
- restart
- seek to time
- auto delete on ready

The idea of "open-ended" animations  also came up. It means we specify from where to start, a speed, and the speed changes according to the animation path. The main issue I see with it is it depends on how often the animation handler is called (as we can move the object in the animation handler without knowing any absolute values). For this we can rather have a helper which calculates the start/end values from some special input parameters. But it's not critical for now.

Of course tests, and examples needs to added and documentation needs to be updated as well.

hyansuper:
minor problem: `ready_cb` sounds like "ready to start" but actually it's called when anim finishes, `complete_cb` maybe more appropriate

kisvegabor:
> minor problem: ready_cb sounds like "ready to start" but actually it's called when anim finishes, complete_cb maybe more appropriate

It was mentioned by others as well.  I've opened a #5378 for renaming ready_cb to completed_cb

ggoraa:
By the way, I had an idea of an "animation library" for LVGL in the back of my head for several weeks by now, basically something with keyframes, orchestration of different animations together, etc

kisvegabor:
Do you mean something like the [Timeline animation](https://docs.lvgl.io/master/overview/animations.html#timeline)?

ggoraa:
Hot damn, I didn't realise LVGL already had allat

kisvegabor:
Not stale

kisvegabor:
cc @XuNeo

kdschlosser:
[...]

PR #6197, opened 2024-05-08
[Draft] Animation API (#5350)

becseya:
This PR implements #5350

### DIFF
--- a/examples/anim/lv_example_anim_4.c
+++ b/examples/anim/lv_example_anim_4.c
@@ -45,7 +45,7 @@ static void on_animation_resume(lv_anim_t * a)
 static void on_slider_value_change(lv_event_t * e)
 {
     LV_UNUSED(e);
-    lv_anim_set_act_time(g.animation, lv_slider_get_value(g.slider));
+    lv_anim_set_current_time(g.animation, lv_slider_get_value(g.slider));
 }
 
 static void on_slider_pressed(lv_event_t * e)
--- a/examples/anim/lv_example_anim_4.c
+++ b/examples/anim/lv_example_anim_4.c
@@ -1,0 +1,105 @@
+#include "../lv_examples.h"
+#if LV_BUILD_EXAMPLES && LV_USE_SLIDER && LV_USE_BUTTON && LV_USE_SWITCH
+
+static struct { // globals
+    lv_anim_t * animation;
+    lv_obj_t * button_label;
+    lv_obj_t * slider;
+} g;
+
+static void anim_y_cb(void * var, int32_t v)
+{
+    lv_obj_set_y(var, v);
+}
+
+static int32_t lv_anim_path_bouncing(const lv_anim_t * a)
+{
+    /* ease in during reverse */
+    if(a->playback_now)
+        return lv_anim_path_cubic_bezier(a, LV_BEZIER_VAL_FLOAT(0), LV_BEZIER_VAL_FLOAT(0.2),
+                                         LV_BEZIER_VAL_FLOAT(0.5), LV_BEZIER_VAL_FLOAT(1));
+    /* ease out during forward */
+    else
+        return lv_anim_path_cubic_bezier(a, LV_BEZIER_VAL_FLOAT(0.2), LV_BEZIER_VAL_FLOAT(0),
+                                         LV_BEZIER_VAL_FLOAT(1), LV_BEZIER_VAL_FLOAT(0.5));
+}
+
+static void btn_start_event_handler(lv_event_t * a)
+{
+    LV_UNUSED(a);
+    lv_anim_toggle_running(g.animation);
+}
+
+static void on_animation_pause(lv_anim_t * a)
+{
+    LV_UNUSED(a);
+    lv_label_set_text(g.button_label, "resume");
+}
+
+static void on_animation_resume(lv_anim_t * a)
+{
+    LV_UNUSED(a);
+    lv_label_set_text(g.button_label, "pause");
+}
+
+static void on_slider_value_change(lv_event_t * e)
+{
+    LV_UNUSED(e);
+    lv_anim_set_act_time(g.animation, lv_slider_get_value(g.slider));
+}
+
+static void on_slider_pressed(lv_event_t * e)
+{
+    LV_UNUSED(e);
+    lv_anim_pause(g.animation);
+}
+
+static void on_slider_press_lost(lv_event_t * e)
+{
+    LV_UNUSED(e);
+    lv_anim_resume(g.animation);
+}
+
+/**
+ * Demonstrate pause, resume, seek
+ */
+void lv_example_anim_4(void)
+{
+    lv_obj_t * red_dot = lv_obj_create(lv_screen_active());
+    lv_obj_set_style_bg_color(red_dot, lv_palette_main(LV_PALETTE_RED), 0);
+    lv_obj_set_style_radius(red_dot, LV_RADIUS_CIRCLE, 0);
+    lv_obj_set_size(red_dot, 15, 15);
+    lv_obj_align(red_dot, LV_ALIGN_TOP_LEFT, 50, 10);
+
+    g.animation = lv_anim_create();
+    lv_anim_set_var(g.animation, red_dot);
+    lv_anim_set_values(g.animation, 10, 220);
+    lv_anim_set_duration(g.animation, 400);
+    lv_anim_set_playback_duration(g.animation, 400);
+    lv_anim_set_repeat_count(g.animation, LV_ANIM_REPEAT_INFINITE);
+    lv_anim_set_path_cb(g.animation, lv_anim_path_bouncing);
+    lv_anim_set_exec_cb(g.animation, anim_y_cb);
+    lv_anim_set_pause_cb(g.animation, on_animation_pause);
+    lv_anim_set_resume_cb(g.animation, on_animation_resume);
+    lv_anim_trigger(g.animation);
+
+    lv_obj_t * btn_start = lv_button_create(lv_screen_active());
+    lv_obj_add_event_cb(btn_start, btn_start_event_handler, LV_EVENT_PRESSED, NULL);
+    lv_obj_add_flag(btn_start, LV_OBJ_FLAG_IGNORE_LAYOUT);
+    lv_obj_align(btn_start, LV_ALIGN_BOTTOM_LEFT, 10, -10);
+
+    g.button_label = lv_label_create(btn_start);
+    lv_label_set_text(g.button_label, "pause");
+    lv_obj_center(g.button_label);
+
+    g.slider = lv_slider_create(lv_screen_active());
+    lv_obj_add_event_cb(g.slider, on_slider_value_change, LV_EVENT_VALUE_CHANGED, NULL);
+    lv_obj_add_event_cb(g.slider, on_slider_pressed, LV_EVENT_PRESSED, NULL);
+    lv_obj_add_event_cb(g.slider, on_slider_press_lost, LV_EVENT_RELEASED, NULL);
+    lv_obj_add_flag(g.slider, LV_OBJ_FLAG_IGNORE_LAYOUT);
+    lv_obj_align(g.slider, LV_ALIGN_BOTTOM_LEFT, 120, -20);
+    lv_obj_set_size(g.slider, 150, 10);
+    lv_slider_set_range(g.slider, 0, 400);
+}
+
+#endif
--- a/src/misc/lv_anim.c
+++ b/src/misc/lv_anim.c
@@ -155,17 +155,17 @@ bool lv_anim_toggle_running(lv_anim_t * a)
     return a->is_running;
 }
 
-void lv_anim_set_act_time(lv_anim_t * a, uint32_t act_time)
+void lv_anim_set_current_time(lv_anim_t * a, uint32_t current_time)
 {
     int32_t value = 0;
 
     if(a->playback_now) anim_toggle_playback_state(a);
 
-    if(act_time <= (uint32_t)a->duration) {
-        a->act_time = act_time;
+    if(current_time <= (uint32_t)a->duration) {
+        a->current_time = current_time;
         value = a->path_cb(a);
     }
-    else { // act_time > duration
+    else { // current_time > duration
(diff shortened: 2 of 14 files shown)


kisvegabor:
@XuNeo 
We have discussed this changes with @becseya offline, so I like the proposed changes. However, I know that this topic is important for you too. Could take a look an tell your opinion?

@

src/misc/lv_anim.c
```diff
@@ -73,60 +77,102 @@ void _lv_anim_core_deinit(void)
...
-            new_anim->custom_exec_cb(new_anim, new_anim->start_value);
-        }
-    }
+    return new_anim;
+}
 
-    /*Creating an animation changed the linked list.
-     *It's important if it happens in a ready callback. (see `anim_timer`)*/
-    anim_mark_list_change();
+lv_anim_t * lv_anim_clone(const lv_anim_t * other)
```
XuNeo: We have a word `dup` used in draw buffer. How about `lv_anim_dup` ? Please go ahead if clone is more suitable.

src/misc/lv_anim.c
```diff
@@ -73,60 +77,102 @@ void _lv_anim_core_deinit(void)
...
+    if(new_anim == NULL) return NULL;
+
+    /* Initialize with the other animation */
+    lv_memcpy(new_anim, other, sizeof(lv_anim_t));
 
-    LV_TRACE_ANIM("finished");
     return new_anim;
 }
 
+void lv_anim_trigger(lv_anim_t * a)
```
XuNeo, line 126: Can we support trigger same animation again and again?

src/misc/lv_anim.c
```diff
@@ -73,60 +77,102 @@ void _lv_anim_core_deinit(void)
...
+    }
+    else {
+        if(a->resume_cb != NULL) a->resume_cb(a);
+        a->is_running = true;
+    }
+
+    return a->is_running;
+}
+
+void lv_anim_set_act_time(lv_anim_t * a, uint32_t act_time)
```
XuNeo:
```suggestion
void lv_anim_set_duration(lv_anim_t * a, uint32_t duration)
```

src/misc/lv_anim.c
```diff
@@ -550,3 +600,75 @@ static bool remove_concurrent_anims(lv_anim_t * a_current)
...
+
+static lv_anim_t * anim_list_allocate(void)
+{
+    lv_anim_t * new_anim = _lv_ll_node_allocate(anim_ll_p);
+    LV_ASSERT_MALLOC(new_anim);
+
+    return new_anim;
+}
+
+static void anim_list_remove(lv_anim_t * a, bool do_call_completed)
```
XuNeo:
```suggestion
static void anim_list_remove(lv_anim_t * a, bool call_completed)
```

src/misc/lv_anim.c
```diff
@@ -550,3 +600,75 @@ static bool remove_concurrent_anims(lv_anim_t * a_current)
...
+{
+    /*Delete the animation from the list first.
+     * This way callbacks will see the animations like it's animation is already deleted*/
+    _lv_ll_remove(anim_ll_p, a);
+
+    /*Flag that the list has changed*/
+    anim_mark_list_change();
+
+    /*Call the callback function at the end*/
+    if((a->completed_cb != NULL) && do_call_completed) a->completed_cb(a);
```
XuNeo:
```suggestion
    if(call_completed && a->completed_cb) a->completed_cb(a);
```

src/misc/lv_anim.h
```diff
@@ -174,16 +183,44 @@ void _lv_anim_core_init(void);
...
+
+/**
+ * Start a dynamically allocated animation.
+ * E.g.:
+ * lv_anim_t* a = lv_anim_create();
+ * lv_anim_set_...(a);
+ * lv_anim_trigger(a);
+ * @param a     pointer to an `lv_anim_t` variable, created by `lv_anim_create()`.
+ */
+void lv_anim_trigger(lv_anim_t * a);
```
XuNeo, line 222: Can we use `lv_anim_start`?

src/misc/lv_ll.c
```diff
@@ -54,27 +54,50 @@ void _lv_ll_init(lv_ll_t * ll_p, uint32_t node_size)
     ll_p->n_size = node_size;
 }
 
-void * _lv_ll_ins_head(lv_ll_t * ll_p)
+void * _lv_ll_node_allocate(lv_ll_t * ll_p)
```
XuNeo, line 57: I think lvgl intends to make the alloc inside `insert` operation. Let's keep it or rewrite it following https://github.com/torvalds/linux/blob/master/include/linux/list.h(which is implemented also in many rtos(https://github.com/apache/nuttx/blob/master/include/nuttx/list.h))

src/misc/lv_anim.h
```diff
@@ -460,6 +540,12 @@ static inline void * lv_anim_get_user_data(const lv_anim_t * a)
  */
 bool lv_anim_delete(void * var, lv_anim_exec_xcb_t exec_cb);
 
+/**
+ * Delete a dynamically allocated animation
+ * @param   a pointer to an `lv_anim_t` variable, created by `lv_anim_create()`.
+ */
+void lv_anim_delete_created(lv_anim_t * a);
```
XuNeo: lv_anim_destroy, or reuse the name lv_anim_delete.

src/misc/lv_anim.h
```diff
@@ -133,6 +139,8 @@ struct _lv_anim_t {
     lv_anim_custom_exec_cb_t custom_exec_cb;/**< Function to execute to animate,
                                                  same purpose as exec_cb but different parameters*/
     lv_anim_start_cb_t start_cb;         /**< Call it when the animation is starts (considering `delay`)*/
+    lv_anim_pause_cb_t pause_cb;    /**< Call it when the animation is paused */
+    lv_anim_resume_cb_t resume_cb;  /**< Call it when the animation is resumed */
     lv_anim_completed_cb_t completed_cb; /**< Call it when the animation is fully completed*/
     lv_anim_deleted_cb_t deleted_cb;     /**< Call it when the animation is deleted*/
```
XuNeo, line 145: Since the memory is allocated by user now, let's avoid free the memory users allocates.

examples/anim/lv_example_anim_4.c
```diff
@@ -0,0 +1,105 @@
+#include "../lv_examples.h"
+#if LV_BUILD_EXAMPLES && LV_USE_SLIDER && LV_USE_BUTTON && LV_USE_SWITCH
+
+struct { // globals
```
XuNeo:
```suggestion
static struct { // globals
```

src/misc/lv_anim.c
```diff
@@ -550,3 +600,75 @@ static bool remove_concurrent_anims(lv_anim_t * a_current)
...
+        if(a->get_value_cb) {
+            int32_t v_ofs = a->get_value_cb(a);
+            a->start_value += v_ofs;
+            a->end_value += v_ofs;
+
+        }
+
+        resolve_time(a);
+
+        /*Do not let two animations for the same 'var' with the same 'exec_cb'*/
```
XuNeo, line 597: Please, why we can't have two animations with same var to same exec callback?

src/misc/lv_anim.c
kisvegabor: I think `dup` is not the best keyword. `copy` or `clone` could be better. So we should rename `lv_draw_buf_dup` too to `lv_draw_buf_copy/clone` and add a define in `lv_api_map_v9.h` for it.

src/misc/lv_anim.c
kisvegabor: I think `act_time` is correct here as this function seeks in the animation. However, `act_time` itself is a wrong a name. It's  coming from my bad English from many years ago. :smile: 

As `anim->act_time` is used only internally we can safely rename it to `current_time`.

src/misc/lv_ll.c
kisvegabor: It's shown oddly in the diff, but in `lv_ll.c` we still have `_lv_ll_ins_head` and `_lv_ll_node_allocate` is added just as an option to allocate a "floating" node.

src/misc/lv_anim.h
kisvegabor: In most of the cases the users want the animation to be deleted automatically on competition. However I do understand for better life cycle management you might need the animation pointers. Can you mention a use case when the animation is not deleted automatically with the widgets they are animating, but you are managing the animation separately?

XuNeo: > a use case when the animation is not deleted automatically

What about a animation when users clicks screen over and over? Reuse the previous animation already setup sounds reasonable.

The complete life cycle management actually comes from the binding language, where the memory should be free'd at certain time. For example, `lua` code may still hold a reference to `anim` handler while it's already automatically deleted.

src/misc/lv_anim.c
becseya: I've renamed `lv_draw_buf_dup`. Please check if the api map is correct

src/misc/lv_anim.c
becseya: It's supported by:
``` c
// configure once
lv_anim_t* a_reusable = lv_anim_create();
// lv_anim_set_*(a_reusable, ...

// call again and again
lv_anim_trigger(lv_anim_clone(a_reusable));
```
---
A @kisvegabor mentioned below, we are having trouble fitting the "user managed memory" paradigm with a consistent API. I can imagine 2 main scenarios:
- Keep the current - stack allocated, then internally copied - API and extend it with "advanced" features, like:
    - add a `tag` field to each animation, then call `lv_anim_delete_by_tag()` to free a subset of animations
    - "smart pointers" returned by "lv_anim_start()", which indicate if the underlying memory has been freed in the meantime
  
  **Also we might need to change our mindset. The fact that `anim` does not have `create` method is not a inconsistency in the API, it's a notification for users that the `create`/`delete` pattern does not apply here due to the internal mechanism**
- Introduce the new dynamic allocation API (as seen in the PR). The issue here is to keep the default - and in most cases very useful - behavior of freeing upon completion. (Another issue is `anim_timeline` module, which creates internal copies too, but never triggers / starts them). Maybe changing to
    - `lv_anim_tigger_single()` **with** freeing upon completion
    - `lv_anim_tigger_reusable()` **without** freeing upon completion
    would make sense

src/misc/lv_anim.c
becseya: renamed

src/misc/lv_anim.h
becseya: We dropped the idea of using the same name due to following reasons:
- Easy to mix with resume / pause
- Notify users of the API change with "deprecated" log messages in `lv_anim_start()`

src/misc/lv_anim.h
becseya: Reusing the animation is possible as of the PR, but see my comment making it more streamlined above.

---

I don't clearly see the relevance of the problems raised by the binding language.
- Freeing at certain time is safe and possible by `lv_anim_delete()` (Right now `lv_anim_delete_by_ptr()` can lead to double freeing if the animation has finished beforehand, `lv_anim_tigger_reusable()` would solve that.)
- Is "holding a reference to deleted handlers" truly animation specific? I'd imagine the same thing would happen with a widget that is being closed, or any other lvgl object, that is being freed by internal mechanisms, like patent-child "garbage collection".

I'm not so familiar with lua. Would you mind creating an example script where this issue is demonstrated? It would be a big help.

src/misc/lv_anim.c
becseya: It would lead to concurrency (who got registered to the list first) and undefined behavior.

Imagine 2 animations setting the y offset of the same ball with a time delay between them. Which one is right? Where the ball should really be?

src/misc/lv_anim.c
becseya: Turns out `dup` is [kind of a standard abbreviation in C (strdup) and POSIX](https://stackoverflow.com/a/252802). 

I'd go with `lv_anim_dup` and addig it to `CODE_STYLE.rst`.
@kisvegabor

kisvegabor:
Not stale

XuNeo:
Is it possible to add `lv_anim_dsc_t` specifically for parameters of an animation? So we can separate it from run time animation handler `lv_anim_t`.

In this way, we can describe an anim in style more appropriately. 
E.g.

`lv_obj_set_style_anim(obj, anim_dsc, 0);`

kisvegabor:
I think it's a good point. Now the root of some confusion is that the animation descriptor and created animation are not separated correctly. 

Do you think we still need `lv_anim_t * lv_anim_create(void);`? 

How can we address your requirements about life cycle management? My idea is that if have an API to control the animations we can restart a created animation multiple times if needed. The only things missing now is an `auto_delete` flag for convenience.

becseya:
I also like the idea of `lv_anim_dsc_t`, it fits better with `timeline` and half of the methods are just modifying the descriptor anyway.

`lv_anim_dsc_t * lv_anim_desc_create(void);` can be used in the case of re-usable animation and stack based allocation can be used for "single-shot".

The question becomes life cycle management. In the case of `lv_anim_dsc_t` it can be manual, so `cb_delete` can be removed, When triggering the animation, lifecycle management should be hidden, and some new mechanism (id, tag) should be added for pause / resume etc,

kisvegabor:
With `lv_anim_dsc_t` life cycle management is more straightforward: 
- The user is responsible for the life cycle of `lv_anim_dsc_t`
- LVGL is responsible for the life cycle of `lv_anim_t` (created internally when an animation starts)

I more concerned about backward compatibility. It seems we need to duplicate basically all the setters and getters to set the values of `lv_abim_dsc_t`. E.g.
```c
lv_anim_set_delay(lv_anim_t * a, int32_t v); //Modify the delay of an already running animation. 
lv_anim_dsc_set_delay(lv_anim_dsc_t * dsc, int32_t v); //Modify the delay of a descriptor

lv_anim_t * lv_anim_create(lv_anim_dsc_t * dsc); //Create an animation from a descriptor
```

kisvegabor:
Closing it in lack of interest. Let's warm it up later if needed.

PR #5927, opened 2024-03-20
feat(stdlib): add lv_strncat and refactor strcat uses

liamHowatt:
### Description of the feature or fix

Implement `lv_strncat` and replace `(lv_)strcat` with `(lv_)strncat` where possible.

A goal of this PR was to replace `strcat` with `strncat` where possible. In practice, that refactor is only helpful if the `src` param is is potentially unterminated. In most cases, `src` is known to be terminated. What needs to be protected in most cases is overrunning `dst`, but `strncat` is not the right tool for that.

This is the kind of refactor that would improve most uses of `strcat`.

https://github.com/lvgl/lvgl/blob/905de3d7c81367608577e1e38d032831a7871c16/examples/others/file_explorer/lv_example_file_explorer_1.c#L53-L58

```c
    const char * envvar = "HOME";
    char home_dir[LV_FS_MAX_PATH_LENGTH];
    /* get the user's home directory from the HOME environment variable*/
    int res = lv_snprintf(home_dir, LV_FS_MAX_PATH_LENGTH, "A:%s", getenv(envvar));
    if (res < 0 || res >= GPIO_MAX_NAME_SIZE) {
        /* optionally handle error */
    }
    LV_LOG_USER("home_dir: %s\n", home_dir);
```

#### Re `strcpy`:

The current builtin `lv_strncpy` is subtly incompatible with the C spec behavior of `strncpy`. A future PR could be opened to change the behavior of the builtin to match the standard.

For that reason, the builtin `lv_strncat` was not implemented in terms of the builtin `lv_strncpy` in this PR since the builtin `lv_strncpy` could change at some point. ~~In a `lv_strncpy` PR, the builtin `lv_strncat` can be written in terms of the builtin `lv_strncpy`.~~ (Edit: The current, subtly flawed, implementation of `lv_strncpy` could be used to implement `lv_strncat` right now, but once it gets fixed, it couldn't be used to implement `lv_strncat` anymore. 😅️)

#### Appendix: Standard behavior of `strncat` and `strncpy`

`strncat` adds a null terminator to `dst` no matter what. (implemented in builtin `lv_strncat` in this PR)

`strncpy` will not add a null terminator if `dest_size` chars are copied before the null terminator. (not the current behavior of the builtin `lv_strncpy`)

<br>

### Notes
- Update the [Documentation](https://github.com/lvgl/lvgl/tree/master/docs) if needed.
- Add [Examples](https://github.com/lvgl/lvgl/tree/master/examples) if relevant.
[...]

PR #6204, opened 2024-05-08
feat(stdlib): strncpy consistency and add strlcpy

liamHowatt:

Following up with the discussion about `strncpy` in #5927

Make `lv_strncpy` consistent with the standard spec.

The standard spec of `strncpy` isn't necessarily better than the one that's implemented in LVGL `lv_string_builtin.c`. The concern is that the behavior of `lv_strncpy` changes depending on the configured `LV_USE_STDLIB_STRING`. We wrap the libc `strncpy` with the modified behavior to match the unique builtin behavior, however we don't wrap the rt-thread implementation and the [rt-thread implementation](https://github.com/RT-Thread/rt-thread/blob/dbf8a26e6c9a8770b2cf708a24e080a81db4f07c/src/klibc/kstring.c#L348) matches the standard spec, i.e. the libc behavior.

I updated the uses of `lv_strncpy` where it's called to ensure a null terminator is added where it's expected.

See the [manpage for strncpy](https://linux.die.net/man/3/strncpy).

### Notes

### DIFF
--- a/src/others/file_explorer/lv_file_explorer.c
+++ b/src/others/file_explorer/lv_file_explorer.c
@@ -599,8 +599,7 @@ static void show_dir(lv_obj_t * obj, const char * path)
     lv_obj_scroll_to_y(explorer->file_table, 0, LV_ANIM_OFF);
 
     lv_memzero(explorer->current_path, sizeof(explorer->current_path));
-    lv_strncpy(explorer->current_path, path, sizeof(explorer->current_path) - 1);
-    explorer->current_path[sizeof(explorer->current_path) - 1] = '\0';
+    lv_strlcpy(explorer->current_path, path, sizeof(explorer->current_path));
     lv_label_set_text_fmt(explorer->path_label, LV_SYMBOL_EYE_OPEN" %s", path);
 
     size_t current_path_len = lv_strlen(explorer->current_path);
--- a/src/others/file_explorer/lv_file_explorer.c
+++ b/src/others/file_explorer/lv_file_explorer.c
@@ -598,7 +598,6 @@ static void show_dir(lv_obj_t * obj, const char * path)
     /*Move the table to the top*/
     lv_obj_scroll_to_y(explorer->file_table, 0, LV_ANIM_OFF);
 
-    lv_memzero(explorer->current_path, sizeof(explorer->current_path));
     lv_strlcpy(explorer->current_path, path, sizeof(explorer->current_path));
     lv_label_set_text_fmt(explorer->path_label, LV_SYMBOL_EYE_OPEN" %s", path);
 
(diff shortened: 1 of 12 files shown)


XuNeo:
> I feel like our behavior for `lv_strncpy` is bette

You are looking for `strlcpy`. Let's use `lv_strlcpy` then. The standard is fixed, we cannot do much about it now.

kisvegabor:
I didn't know `strlcpy` either, but it's great! I agree to use `lv_strlcpy`. @liamHowatt if you also agree could you update this PR accordingly?

liamHowatt:
I agree. I'll add `strlcpy` and use it in most places `strncpy` is.

liamHowatt:
Updated.

So it looks like the `strlcpy` is a BSD thing. You need libbsd installed. So I just implemented the clib one in terms of `strlen` and `memcpy` so people using clib string don't need libbsd.

FYI `strlcpy` returns the length of `src`. Something to note.

src/others/file_explorer/lv_file_explorer.c
```diff
@@ -599,7 +599,7 @@ static void show_dir(lv_obj_t * obj, const char * path)
     lv_obj_scroll_to_y(explorer->file_table, 0, LV_ANIM_OFF);
 
     lv_memzero(explorer->current_path, sizeof(explorer->current_path));
```
XuNeo: Remove `lv_memzero(explorer->current_path, sizeof(explorer->current_path));`

XuNeo (dismissed):

XuNeo (approved):
Thank you!

kisvegabor (commented):
Thank you! 

I'm not sure we need to keep `lv_strncpy` but let's leave it for backward compatibility.

kisvegabor (approved):

kdschlosser:
@kisvegabor 

I think this is the one that broke the MicroPython CI.

kdschlosser:
it looks like this function is wrong. It strips off the trailing `'\0'`.

```c
size_t lv_strlcpy(char * dst, const char * src, size_t dst_size)
{
    size_t i = 0;
    if(dst_size > 0) {
        for(; i < dst_size - 1 && src[i]; i++) {
            dst[i] = src[i];
        }
        dst[i] = '\0';
    }
    while(src[i]) i++;
    return i;
}

```

liamHowatt:
That CI was run before this got merged.

kdschlosser:
There are only 2 things that got merged in that time frame and this is the only one that would effect the micropython CI

Maybe something got changed over in the micropython binding?

kdschlosser:
nothing changed over there either. Gonna have to turn on debugging for the test to see what is happening.

PR #6206, opened 2024-05-09
fix(animimage): add NULL pointer check

yinntian:

When dsc sets an incorrect NULL value, it can cause the animation's callback function to incorrectly take an array subscript value for the NULL value and cause the program to crash, which should prompt an error message and return.

### Notes

### DIFF
--- a/src/widgets/animimage/lv_animimage.c
+++ b/src/widgets/animimage/lv_animimage.c
@@ -36,7 +36,7 @@
 /**********************
  *  STATIC PROTOTYPES
  **********************/
-static void index_change(lv_obj_t * obj, int32_t index);
+static void index_change(lv_obj_t * obj, int32_t idx);
 static void lv_animimg_constructor(const lv_obj_class_t * class_p, lv_obj_t * obj);
 
 /**********************
@@ -156,9 +156,8 @@ static void lv_animimg_constructor(const lv_obj_class_t * class_p, lv_obj_t * ob
     lv_anim_set_repeat_count(&animimg->anim, LV_ANIM_REPEAT_INFINITE);
 }
 
-static void index_change(lv_obj_t * obj, int32_t index)
+static void index_change(lv_obj_t * obj, int32_t idx)
 {
-    int32_t idx;
     lv_animimg_t * animimg = (lv_animimg_t *)obj;
 
     if(animimg->dsc == NULL) {
@@ -166,7 +165,7 @@ static void index_change(lv_obj_t * obj, int32_t index)
         return;
     }
 
-    idx = index % animimg->pic_count;
+    if(idx >= animimg->pic_count) idx =  animimg->pic_count - 1;
 
     lv_image_set_src(obj, animimg->dsc[idx]);
 }
--- a/src/widgets/animimage/lv_animimage.c
+++ b/src/widgets/animimage/lv_animimage.c
@@ -162,7 +162,7 @@ static void index_change(lv_obj_t * obj, int32_t index)
     lv_animimg_t * animimg = (lv_animimg_t *)obj;
 
     if(animimg->dsc == NULL) {
-        LV_LOG_ERROR("dsc is null");
+        LV_LOG_WARN("dsc is null");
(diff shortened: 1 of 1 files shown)


XuNeo (commented):
Could you also add a commit to fix the value error in `lv_animimg_set_src`?

I think it should be `    lv_anim_set_values(&animimg->anim, 0, (int32_t)num);`

https://github.com/lvgl/lvgl/pull/6206/files#diff-e02308a3e1e1189a9fb9cc6f0e8408f740eb84b5685ee8c7e5bc17b6c72a2543L69-L76

XuNeo (dismissed):

W-Mai (approved):

liamHowatt (dismissed):
@XuNeo the `lv_example_animimg_1` now looks more correct after removing the ` - 1`.

kisvegabor:
If you modify `lv_example_animimg_1` like this:
```c
lv_animimg_set_repeat_count(animimg0, 1);
```

On the last frame it will set `image[0]`  again due to this in `index_change`:
```c
static void index_change(lv_obj_t * obj, int32_t index)
{
    int32_t idx;
    lv_animimg_t * animimg = (lv_animimg_t *)obj;

    idx = index % animimg->pic_count;
```

For good timing we should really remove the `-1` but use this in `index_change`:
```c

if(index >= animimg->pic_count) index =  animimg->pic_count - 1; 
```

src/widgets/animimage/lv_animimage.c
```diff
@@ -161,6 +161,11 @@ static void index_change(lv_obj_t * obj, int32_t index)
     int32_t idx;
     lv_animimg_t * animimg = (lv_animimg_t *)obj;
 
+    if(animimg->dsc == NULL) {
+        LV_LOG_ERROR("dsc is null");
```
kisvegabor: ERROR is used for crash or close-crash situations. I think it should be just a warning.

yinntian: Yes, it should be warning

XuNeo:
> if(index >= animimg->pic_count) index =  animimg->pic_count - 1; 

This is better. The anim value is mapped directly to the image index, thus `lv_anim_set_values(&animimg->anim, 0, (int32_t)num - 1); ` sounds correct. `if(index >= animimg->pic_count) index =  animimg->pic_count - 1; ` makes sure if animation has overshoot, the index is still in range.

@yinntian Could you please update the PR accordingly?

yinntian:
@XuNeo 
Now here's the state of my attempts:
```c
lv_anim_set_values(&animimg->anim, 0, (int32_t)num);
```
```c
static void index_change(lv_obj_t * obj, int32_t index)
{
    int32_t idx;
    lv_animimg_t * animimg = (lv_animimg_t *)obj;

    if(animimg->dsc == NULL){
        return ;
    }
    
    if(index >= animimg->pic_count){
        index = animimg->pic_count - 1;
    }
    idx = index % animimg->pic_count;
    
    // debug code -------
    static int debug_count = -2;
    printf("%d ", idx);
    debug_count++;
    if(debug_count % 4 == 0){
        printf("\n");
    }
    //-------

    lv_image_set_src(obj, animimg->dsc[idx]);
}
```
Looks like the animation is normal

动画2

But by the output message, the change of 'idx' is a cycle of 0, 1, 2, 2, but the ‘2’ output will immediately output ‘0’, is it possible that there is something wrong with anim in this behavior?

kisvegabor:
`   idx = index % animimg->pic_count;` is not required. It looks like this for a 3 second animation with 3 images:
```
  0s        1s          2s         3s (stop)
  img0      img1        img2       img2
   ---------- ----------- ----------
       1s         1s          1s
``` 

As you can see this way each image were shown for 1 sec.

kisvegabor:
I've added the fix proposed above. Now it looks good with `lv_example_animimg_1();`

liamHowatt (approved):
Nice

XuNeo (approved):

PR #6219, opened 2024-05-13
feat(micropython): improve mem core micropython

Carglglz:
Allow using `MICROPY_MALLOC_USES_ALLOCATED_SIZE` option required to build `lv_bindings_micropython` with upstream MicroPython.

Required for ~https://github.com/lvgl/lv_binding_micropython/pull/242~
https://github.com/lvgl/lv_binding_micropython/pull/341

### DIFF
--- a/src/stdlib/micropython/lv_mem_core_micropython.c
+++ b/src/stdlib/micropython/lv_mem_core_micropython.c
@@ -62,17 +62,32 @@ void lv_mem_remove_pool(lv_mem_pool_t pool)
 
 void * lv_malloc_core(size_t size)
 {
+#if MICROPY_MALLOC_USES_ALLOCATED_SIZE
+    return gc_alloc(size, true);
+#else
     return m_malloc(size);
+#endif
 }
 
 void * lv_realloc_core(void * p, size_t new_size)
 {
+
+#if MICROPY_MALLOC_USES_ALLOCATED_SIZE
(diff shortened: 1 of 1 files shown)


kisvegabor:
cc @kdschlosser

kisvegabor (approved):
Thank you! I haven't tested it, but I assume that it's really required.

FASTSHIFT (approved):

kdschlosser:
I thought I had actually made a comment on this, maybe I didn't press the comment button. 

These functions should not even exist in LVGL. This is something that is able to be done using macros and it should be done at the binding level and not in LVGL.

This is the exact reason as to why there shouldn't be anything binding specific n LVGL because Now you have micropython specific code spread across multiple repositories. It can lead to a tangled mess when trying to locate where things are being done at.

Another option is to declare functions as extern and then let them get populated using a source file that is compiled in the binding.


The LV_STDLIB_MICROPYTHON macro doesn't need to exist in LVGL as that is able to be defined in the binding as well if the source file is moved. this would stop the other source files from getting compiled.

In the lv_mem_core_micropython.c file it is actually including a file that is from the micropython binding. that's kind of a goofy setup. not to mention confusing too.

In the binding I am working on I went about handling this issue in a different manner but this PR would make it easier to deal with that's for sure. I think this entire file needs to be located in the binding and not in LVGL.

PR #6221, opened 2024-05-14
Renesas GLCDC dirver

becseya:
Add display driver to support Renesas GLCDC peripheral.


becseya:
`tests/src/lv_test_conf_full.h` was not extended with the new config macro, as build would fail, due to missing FSP libraries. I'm assuming this is preferred over mocking everything. @kisvegabor

kisvegabor:
> I'm assuming this is preferred over mocking everything. 

Correct, normally we don't compile the hardware specific code.

docs/integration/driver/display/renesas_glcdc.rst
```diff
@@ -0,0 +1,11 @@
+=============================
+Renesas GLCDC
+=============================
+
+Overview
```
kisvegabor, line 7: The decoration should be as long as the text:
```suggestion
Overview
--------
```

src/drivers/display/renesas_glcdc/lv_renesas_glcdc.c
```diff
@@ -0,0 +1,213 @@
...
+
+static void glcdc_init(void)
+{
+    /* Fill the Frame buffer with a colour, to zero out info from previous execution runs */
+    uint32_t count;
+    uint16_t * p = (uint16_t *)&fb_background[0][0];
+
+    for(count = 0; count < sizeof(fb_background) / 2; count++) {
+        *p++ = RGB_565_BLACK;
+    }
```
kisvegabor:
```suggestion
    lv_memzero(fb_background, sizeof(fb_background));
```

src/drivers/display/renesas_glcdc/lv_renesas_glcdc.c
```diff
@@ -0,0 +1,213 @@
...
+}
+
+/*This function is declared in and being used by FSP generated code modules*/
+void glcdc_callback(display_callback_args_t * p_args)
+{
+    if(DISPLAY_EVENT_LINE_DETECTION == p_args->event) {
+#if BSP_CFG_RTOS == 2               // FreeRTOS
+        BaseType_t context_switch;
+
+        //
```
kisvegabor: Please use `/**/` comments

src/drivers/display/renesas_glcdc/lv_renesas_glcdc.c
```diff
@@ -0,0 +1,213 @@
...
+
+#include "LVGL_thread.h"
+#include <stdbool.h>
+#include "../../../display/lv_display_private.h"
+
+/*********************
+ *      DEFINES
+ *********************/
+
+#define RGB_565_BLACK  (0)
```
kisvegabor:
```suggestion
#define RGB_565_BLACK  (0)
```

src/drivers/display/renesas_glcdc/lv_renesas_glcdc.c
```diff
@@ -0,0 +1,213 @@
...
+
+}
+
+static void flush_partial(lv_display_t * display, const lv_area_t * area, uint8_t * px_map)
+{
+    LV_UNUSED(display);
+
+    int32_t w = lv_area_get_width(area);
+    int32_t h = lv_area_get_height(area);
+    SCB_CleanInvalidateDCache_by_Addr(px_map, w * h * 2);
```
kisvegabor: It should be called in line 204 for the frame buffer.
```c
SCB_CleanInvalidateDCache_by_Addr(fb, DISPLAY_HSIZE_INPUT0 * DISPLAY_VSIZE_INPUT0 * 2);
```

src/drivers/display/renesas_glcdc/lv_renesas_glcdc.h
```diff
@@ -0,0 +1,57 @@
...
+
+/**********************
+ * GLOBAL PROTOTYPES
+ **********************/
+
+/**
+ * Create a display using Renesas' GLCDC peripherial in DIRECT render mode
+ * @return  pointer to the created display
+ */
+lv_display_t * lv_renesas_glcdc_create_direct(void);
```
kisvegabor: Usually `create` is the last keyword in function names:
```suggestion
lv_display_t * lv_renesas_glcdc_direct_create(void);
```

src/drivers/display/renesas_glcdc/lv_renesas_glcdc.h
```diff
@@ -0,0 +1,57 @@
...
+
+/**
+ * Create a display using Renesas' GLCDC peripherial in PARTIAL render mode
+ * @param buf1      first buffer
+ * @param buf2      second buffer (can be `NULL`)
+ * @param buf_size  buffer size in byte
+ * @return          pointer to the created display
+ */
+lv_display_t * lv_renesas_glcdc_create_partial(void * buf1, void * buf2, uint32_t buf_size);
```
kisvegabor:
```suggestion
lv_display_t * lv_renesas_glcdc_partial_create(void * buf1, void * buf2, uint32_t buf_size);
```

kisvegabor:
```suggestion
lv_display_t * lv_renesas_glcdc_create_partial(void * buf1, void * buf2, size_t buf_size);
```

kisvegabor (dismissed):

kisvegabor (commented):
Looks good in general, just a few comments.

src/drivers/display/renesas_glcdc/lv_renesas_glcdc.c
becseya: replacing 191 or keep both? and why call for the whole buffer in partial mode?

kisvegabor: Only at line 191. 
It would be more optimal to invalidate on the changed area. It can be added now or in a later PR as an optimization.

becseya: The suggested line does not work, as `fb` is being incremented in the loop, and pointing to a bad location by line 204.
This works* at the end of function:
```c
SCB_CleanInvalidateDCache_by_Addr(fb_background[1], sizeof(fb_background[1]));
```
Or this in the loop:
```c
SCB_CleanInvalidateDCache_by_Addr(fb, w * 2);
```

However it only works for the artifacts seen at switch transition. Pressing the "Log out" button also creates artifacts and they are not resolved by cache invalidation. It might be a different issue, as the stripes are inside the button, with white color. At no point the animation should be that, so it's not a "slow transition" related to cache emptying out.

image

This is not present in direct mode, however a scroll bar appears after pressing the button once. The bar does not appear in the PC emulator with matching screen resolution.

kisvegabor (approved):
Looks good! We can fix the issue of the partial rendering in a separate PR. 

The failing  CI is unrelated and being discussed [here](https://github.com/lvgl/lvgl/pull/6081#issuecomment-2124498696).

src/drivers/display/renesas_glcdc/lv_renesas_glcdc.c
```diff
@@ -0,0 +1,192 @@
...
+}
+
+static void flush_wait_direct(lv_display_t * display)
+{
+    if(!lv_display_flush_is_last(display)) return;
+
+#if BSP_CFG_RTOS == 2              /*FreeRTOS*/
+    /*If Vsync semaphore has already been set, clear it then wait to avoid tearing*/
+    if(uxSemaphoreGetCount(_SemaphoreVsync)) {
+        xSemaphoreTake(_SemaphoreVsync, 10);
```
liamHowatt, line 153: Based on your comment, should this line give/clear the semaphore instead of take it? Otherwise everything LGTM!

kisvegabor: It was copied from the original code, and it worked, but I don't understand something. 

The idea is to take the semaphore twice:
1. Take it normally (nothing happens)
2. Take it again to yield and wait until [VSYNC gives it](https://github.com/lvgl/lvgl/pull/6221/files/1563b8df398e10cf5d6dca371a37dc62285ed7d0#diff-faec9062bf810d2b190e1ce87e4165dfa0b63423d074de2645fb851e50b64fddR79)

But I can't see how `if(uxSemaphoreGetCount(_SemaphoreVsync))` can be `true` as the semaphore is is always given at the end. And if it's not taken twice this mechanism is not working.  

@becseya can you take a deeper look on this part? What happens if you just add
```c
xSemaphoreTake(_SemaphoreVsync, portMAX_DELAY);
xSemaphoreTake(_SemaphoreVsync, portMAX_DELAY);
```
here?

liamHowatt: kisvegabor's comment helped me understand how this logic works. I think it's actually okay. The code comment wording confused me because `set` means `give` and `clear` means `take`, in the code comment.

Also, regarding
> But I can't see how `if(uxSemaphoreGetCount(_SemaphoreVsync))` can be `true` as the semaphore is is always given at the end.

`GetCount` returns 1 if it's _available_ [according to the docs](https://www.freertos.org/uxSemaphoreGetCount.html)

So it should work as a signalling mechanism, as intended. I'm going to approve.

liamHowatt (approved):

kisvegabor:
Merged, thank you Akos!

PR #6244, opened 2024-05-20
fix(drivers): drm driver not initialising with small screens

timstableford:

The DRM display driver would not initialise with screens smaller than the 800x480 passed to lv_display_create because it called lv_display_set_buffers with buffers that were too small for the expected resolution when doing direct rendering.


kisvegabor (approved):
Nice, thank you!

FASTSHIFT (approved):

PR #6249, opened 2024-05-22
feat(obj): add check null pointer

XuNeo:

Check null pointer and return early.

### Notes

### DIFF
--- a/src/core/lv_obj_class.c
+++ b/src/core/lv_obj_class.c
@@ -66,8 +66,12 @@ lv_obj_t * lv_obj_class_create_obj(const lv_obj_class_t * class_p, lv_obj_t * pa
             disp->screen_cnt = 0;
         }
 
+        lv_obj_t ** screens = lv_realloc(disp->screens, sizeof(lv_obj_t *) * (disp->screen_cnt + 1));
+        LV_ASSERT_NULL(disp->screens);
+        if(screens == NULL) return NULL;
+
         disp->screen_cnt++;
-        disp->screens = lv_realloc(disp->screens, sizeof(lv_obj_t *) * disp->screen_cnt);
+        disp->screens = screens;
         disp->screens[disp->screen_cnt - 1] = obj;
 
         /*Set coordinates to full screen size*/
--- a/src/core/lv_obj_class.c
+++ b/src/core/lv_obj_class.c
@@ -67,7 +67,7 @@ lv_obj_t * lv_obj_class_create_obj(const lv_obj_class_t * class_p, lv_obj_t * pa
         }
 
         lv_obj_t ** screens = lv_realloc(disp->screens, sizeof(lv_obj_t *) * (disp->screen_cnt + 1));
-        LV_ASSERT_NULL(disp->screens);
+        LV_ASSERT_NULL(screens);
         if(screens == NULL) return NULL;
 
         disp->screen_cnt++;
--- a/src/core/lv_obj.c
+++ b/src/core/lv_obj.c
@@ -98,6 +98,8 @@ lv_obj_t * lv_obj_create(lv_obj_t * parent)
 {
     LV_LOG_INFO("begin");
     lv_obj_t * obj = lv_obj_class_create_obj(MY_CLASS, parent);
+    LV_ASSERT_NULL(obj);
+    if(obj == NULL) return NULL;
     lv_obj_class_init_obj(obj);
     return obj;
 }


kisvegabor:
We have so many weird issues in CI recently that I suspect some alien activity or a string sun flare. :smile:

@kdschlosser do you have any ideas why the MicrtoPython test fails?

kdschlosser:
add --DEBUG to the CI for the test then rerun the CI and then check the output. It will tell you what is going on.

XuNeo:
> add --DEBUG to the CI for the test then rerun the CI and then check the output. It will tell you what is going on.

Could you explain where to add `--DEBUG` ? I can add it to CI so we always the the verbose logs.

kdschlosser:
> Could you explain where to add --DEBUG ? I can add it to CI so we always the the verbose logs.

You don't always want the verbose logs. It will spit out an absolute massive amount of data. All of the data that passes between the testing script and micropython is done using stdout of the micropython process. the frame buffer data gets passed as well and for an RGB888 at something like 800 x 400 resolution that's a huge amount of data that ends up getting output.


I am going to make a PR that will have better handling when this kind of an error comes up so it will automatically output the data so we can see. I didn't think about it having an issue where it's having the issue and didn't add any exception catching for it. I will fix this so in the future the errors will show up.

src/core/lv_obj_class.c
```diff
@@ -66,8 +66,12 @@ lv_obj_t * lv_obj_class_create_obj(const lv_obj_class_t * class_p, lv_obj_t * pa
             disp->screen_cnt = 0;
         }
 
+        lv_obj_t ** screens = lv_realloc(disp->screens, sizeof(lv_obj_t *) * (disp->screen_cnt + 1));
+        LV_ASSERT_NULL(disp->screens);
```
XuNeo:
```suggestion
        LV_ASSERT_NULL(screens);
```

kdschlosser:
simple solution to a complex problem?

Is that last change what fixed the build?

kdschlosser:
I am going to update the test script so it is more informative. Give me a day or so to do that.

XuNeo:
> Is that last change what fixed the build?

Yes. It's a stupid typo.

> I am going to update the test script so it is more informative. Give me a day or so to do that.

Thank you!

kdschlosser:
The PR is in place for the update to the test script to provide more information if an error happens. That should now be all set and hopefully will make it easier to locate where a problem is.

kdschlosser:
Isn't this code 
c
```
       LV_ASSERT_NULL(screens);
        if(screens == NULL) return NULL;
```

kind of self defeating? There really isn't a purpose of having the assert if there is a check being done to see if it is NULL..

XuNeo:
Unlike the `assert()` macro, `LV_ASSERT_NULL` can be optionally turned off. We want to make sure upper layer catch the error instead of continue to use the null pointer. 

Similar code can be found here too.

lvgl:

https://github.com/lvgl/lvgl/blob/f2e81d80b37214eb72a2b21efd658817d928ac1e/src/lv_init.c#L78-L85


NuttX

https://github.com/apache/nuttx/blob/07f0e0c166b8bf3c65561531c8ea7f87a1bb9b53/fs/inode/fs_files.c#L99-L104

FASTSHIFT (commented):
There are too many functions related to obj. If judgment is added to all of them, the code will be bloated and will also affect performance. We should focus on solving the problem of insufficient memory on the system. Because if the system cannot allocate even one `lv_obj` memory, it is almost in an unusable state.

XuNeo:
Do you mean the check for `lv_obj_create`? 
```c
    lv_obj_t * obj = lv_obj_class_create_obj(MY_CLASS, parent);
    LV_ASSERT_NULL(obj);
    if(obj == NULL) return NULL;
```

The issue I actually met is the `lv_display_get_default` returns NULL. And for this particular system, address 0 is SRAM. So the code continues to execute.

As for the concern of performance, I think creating obj is not bottle necked by the NULL pointer check.

liamHowatt (approved):
`lv_obj_class_create_obj` is called in 65 places. Should we just `LV_ASSERT_MALLOC` in there? Or is there some reason that function should be allowed to fail without crashing the app?

Currently:
```c
lv_obj_t * lv_obj_class_create_obj(const lv_obj_class_t * class_p, lv_obj_t * parent)
{
    LV_TRACE_OBJ_CREATE("Creating object with %p class on %p parent", (void *)class_p, (void *)parent);
    uint32_t s = get_instance_size(class_p);
    lv_obj_t * obj = lv_malloc_zeroed(s);
    if(obj == NULL) return NULL;
    obj->class_p = class_p;
    obj->parent = parent;
[...]
```

> Because if the system cannot allocate even one `lv_obj` memory, it is almost in an unusable state.

I have often thought that :wink:

XuNeo:
> Should we just LV_ASSERT_MALLOC in there?

`lv_display_get_default` also could return NULL, thus put it outside of `lv_obj_class_create_obj` is more straighforward.

> Or is there some reason that function should be allowed to fail without crashing the app?

It should not be allowed. The issue is that LV_ASSERT_NULL can be optionally turned off. Normally we could simply turn it off because access address 0 normally would crash.

kisvegabor (approved):

PR #6254, opened 2024-05-22
feat(opengles): add basic driver for opengles

danyyliu:

add basic driver for opengles.

### Notes


kisvegabor:
Thank you Dany, we will review it in detail, but I have a quick question to @GorgonMeducer: Shall we add the OpenGL related code/configs to CMSIS-Pack?

src/drivers/opengles/lv_opengles_window.c
```diff
@@ -0,0 +1,438 @@
...
+        return;
+    }
+    lv_opengles_window_t * dsc = lv_display_get_driver_data(disp);
+    dsc->ignore_size_chg = 1;
+    lv_display_set_resolution(disp, width / dsc->zoom, height / dsc->zoom);
+    dsc->ignore_size_chg = 0;
+    lv_refr_now(disp);
+}
+
+static void texture_resize(lv_display_t * disp)
```
becseya: Resizing the window can cause the following artifact or crash the application. If it's reproducible on other machines, we should investigate it, or drop the feature.
image

src/drivers/opengles/lv_opengles_window.c
```diff
@@ -0,0 +1,438 @@
...
+    if(dsc->fb1) {
+        LV_LOG_INFO("fb1 pixel: %d\n", ((uint16_t *)(dsc->fb1))[0]);
+    }
+}
+
+static void res_chg_event_cb(lv_event_t * e)
+{
+    lv_display_t * disp = lv_event_get_current_target(e);
+
+    int32_t hor_res = lv_display_get_horizontal_resolution(disp);
```
becseya: Unused variable `hor_res`. I'd suggest turning on `-Wall` and `-Werror` compiler flags

Kconfig
```diff
@@ -1628,6 +1628,15 @@ menu "LVGL configuration"
 		config LV_USE_WINDOWS
 			bool "Use LVGL Windows backend"
 			default n
+
+		config LV_USE_OPENGLES
+			bool "Use GLFW and OpenGL to open window on PC and handle mouse and keyboard"
+			default n
+
+		config LV_USE_OPENGLES_DEBUG
```
becseya, line 1646: `LV_USE_OPENGLES_DEBUG` is missing from conf header files, We should add it, or leave it as a manually settable option.

docs/integration/driver/opengles.rst
```diff
@@ -0,0 +1,65 @@
+===============================
+OpenGL ES Display/Inputs driver
+===============================
+
+Overview
+-------------
```
becseya: The underline should match the text, like:
```
Overview
--------
```

docs/integration/driver/opengles.rst
```diff
@@ -0,0 +1,65 @@
...
+
+Configure OpenGL driver
+-----------------------
+
+Enable the OpenGL driver support in lv_conf.h, by cmake compiler define or by KConfig
+    .. code:: c
+
+        #define LV_USE_OPENGLES  1
+
+Usage
```
becseya, line 29: Information about required linked libraries should be added to this section.

It was `-lGL` `-lGLEW` `-lglfw` for me on Linux.

docs/integration/driver/opengles.rst
```diff
@@ -0,0 +1,65 @@
...
+| It is an alternative to **Wayland**, **XCB**, **SDL** or **Qt**.
+
+The main purpose for this driver is for testing/debugging the LVGL application in an **OpenGL** simulation window.
+
+Prerequisites
+-------------
+
+The OpenGL driver uses GLEW GLFW to access the OpenGL window manager.
+
+1. Install GLEW: ``sudo apt-get install libglew-dev``
```
becseya: `SUGGESTION`
Required packages are usually installed via a single command:
```markdown
Install them with you package manager: `sudo apt-get install libglew-dev libglfw3-dev`
```

src/drivers/opengles/lv_opengles_common.h
```diff
@@ -0,0 +1,39 @@
+/**
+ * @file lv_opengles_common.h
```
becseya: *"common"* is not the best fitting phase, as it's being used only in `lv_opengles_driver.c`. I'd suggest `lv_opengles_debug.h` as it fits better with usage.

src/drivers/opengles/lv_opengles_common.h
```diff
@@ -0,0 +1,39 @@
...
+#include <stdbool.h>
+
+#if LV_USE_OPENGLES
+
+#include <GL/glew.h>
+#include <GLFW/glfw3.h>
+
+#if LV_USE_OPENGLES_DEBUG
+#define ASSERT(x) if(!(x)) __builtin_trap();
+#define GLCall(x) GLClearError();\
```
becseya: Macros should be all caps: `GL_CALL`

src/drivers/opengles/lv_opengles_driver.h
```diff
@@ -0,0 +1,79 @@
...
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+void lv_opengles_vertex_buffer_init(const void * data, unsigned int size);
+
+void lv_opengles_vertex_buffer_deinit();
+
+void lv_opengles_vertex_buffer_bind();
```
becseya: This is actually a static function (being used only in `lv_opengles_driver.c`), so the declaration should be moved there. There are multiple cases of this in this file.

src/drivers/opengles/lv_opengles_driver.h
```diff
@@ -0,0 +1,79 @@
...
+#ifndef LV_OPENGLES_DRIVER_H
+#define LV_OPENGLES_DRIVER_H
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+void lv_opengles_vertex_buffer_init(const void * data, unsigned int size);
+
+void lv_opengles_vertex_buffer_deinit();
```
becseya: This function is never being called, just like multiple other `*_deinit()` methods.

src/drivers/opengles/lv_opengles_driver.c
```diff
@@ -0,0 +1,322 @@
...
+
+#if LV_USE_OPENGLES
+
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <stdint.h>
+#include "lv_opengles_common.h"
+
+void GLClearError()
```
becseya: `SUGGESTION`
If we rename `lv_opengles_common.h`, the following 2 definitions could move to `lv_opengles_debug.c`.

src/drivers/opengles/lv_opengles_driver.c
```diff
@@ -0,0 +1,322 @@
...
+{
+    GLCall(glBindBuffer(GL_ARRAY_BUFFER, m_vertex_buffer_id));
+}
+
+void lv_opengles_vertex_buffer_unbind()
+{
+    GLCall(glBindBuffer(GL_ARRAY_BUFFER, 0));
+}
+
+static unsigned int m_vertex_array_id = 0;
```
becseya: Please follow the pattern for macros, declarations, static and global functions provided by [the template](https://github.com/lvgl/lvgl/blob/master/src/misc/lv_templ.c).

src/drivers/opengles/lv_opengles_driver.c
```diff
@@ -0,0 +1,322 @@
...
+
+    GLCall(glGenTextures(1, &m_texture_id));
+    GLCall(glBindTexture(GL_TEXTURE_2D, m_texture_id));
+
+    GLCall(glTexParameteri(GL_TEXTURE_2D, GL_TEXTURE_MIN_FILTER, GL_LINEAR));
+    GLCall(glTexParameteri(GL_TEXTURE_2D, GL_TEXTURE_MAG_FILTER, GL_LINEAR));
+    GLCall(glTexParameteri(GL_TEXTURE_2D, GL_TEXTURE_WRAP_S, GL_CLAMP_TO_EDGE));
+    GLCall(glTexParameteri(GL_TEXTURE_2D, GL_TEXTURE_WRAP_T, GL_CLAMP_TO_EDGE));
+
+    //GLCall(glTexImage2D(GL_TEXTURE_2D, 0, GL_RGBA8, width, height, 0, GL_RGBA, GL_UNSIGNED_BYTE, buffer));
```
becseya: Commented out source code is hard to maintain. It should be:
- removed
- turned into a written text comment, explaining the possible usage of `GL_RGBA8`
- in the case of log messages, the level can be lowered to debug, or macro guarding can be used (like in `VGLITE_LOG_TRACE`)

src/drivers/opengles/lv_opengles_driver.c
```diff
@@ -0,0 +1,322 @@
...
+        return false;
+    }
+    return true;
+}
+
+static unsigned int m_vertex_buffer_id = 0;
+
+void lv_opengles_vertex_buffer_init(const void * data, unsigned int size)
+{
+    GLCall(glGenBuffers(1, &m_vertex_buffer_id));//create an ID :buffer
```
becseya: Please use `/**/` comment format for compatibility with code-base.

src/drivers/opengles/lv_opengles_mouse.c
```diff
@@ -0,0 +1,142 @@
...
+ *  STATIC PROTOTYPES
+ **********************/
+static void opengles_mouse_read(lv_indev_t * indev, lv_indev_data_t * data);
+static void release_indev_cb(lv_event_t * e);
+
+/**********************
+ *  STATIC VARIABLES
+ **********************/
+
+typedef struct {
```
becseya: `typedef`s should be in the `DEFINES` template section at line 18.

src/drivers/opengles/lv_opengles_mouse.c
```diff
@@ -0,0 +1,142 @@
...
+    lv_opengles_mouse_t * dsc = lv_indev_get_driver_data(indev);
+    if(dsc) {
+        lv_indev_set_driver_data(indev, NULL);
+        lv_indev_set_read_cb(indev, NULL);
+        lv_free(dsc);
+        LV_LOG_INFO("done");
+    }
+}
+
+void _lv_opengles_mouse_move_handler(lv_display_t * disp, int x, int y)
```
becseya: Missing `static` keyword and declarations in `STATIC PROTOTYPES` section at line 28

GorgonMeducer:
@kisvegabor If you want, we can. As you might noticed, I have already added PC/Windows/Linux related content to cmsis-pack. 
I see no special reason to reject any source code as long as you want them.  LVGL owns the cmsis-pack. I am only the maintainer.

src/drivers/opengles/lv_opengles_mouse.c
```diff
@@ -0,0 +1,142 @@
...
+    lv_indev_t * indev = lv_indev_get_next(NULL);
+    while(indev) {
+        if(lv_indev_get_display(indev) == disp && lv_indev_get_type(indev) == LV_INDEV_TYPE_POINTER) {
+            break;
+        }
+        indev = lv_indev_get_next(indev);
+    }
+
+    if(indev == NULL) return;
+    lv_opengles_mouse_t * indev_dev = lv_indev_get_driver_data(indev);
```
becseya: `SUGGESTION`
*"Find a suitable indev"* is being used across the code-base multiple times. It could be extracted to a standalone function in the `display` module, to prevent code duplication.

src/drivers/opengles/lv_opengles_window.h
```diff
@@ -0,0 +1,48 @@
...
+ *      TYPEDEFS
+ **********************/
+
+/**********************
+ * GLOBAL PROTOTYPES
+ **********************/
+
+lv_display_t * lv_opengles_window_create(int32_t hor_res, int32_t ver_res);
+
+void lv_opengles_quit();
```
becseya: This is actually a `static` function.

src/drivers/opengles/lv_opengles_window.c
```diff
@@ -0,0 +1,438 @@
...
+    if(disp == NULL) {
+        lv_free(dsc);
+        return NULL;
+    }
+
+    lv_display_add_event_cb(disp, release_disp_cb, LV_EVENT_DELETE, disp);
+
+    lv_display_set_driver_data(disp, dsc);
+
+    window_create(disp);
```
becseya: Return value is being ignored

src/drivers/opengles/lv_opengles_window.c
```diff
@@ -0,0 +1,438 @@
...
+static int window_create(lv_display_t * disp)
+{
+    lv_opengles_window_t * dsc = lv_display_get_driver_data(disp);
+    dsc->zoom = 1;
+
+    glfwSetErrorCallback(glfw_error_cb);
+    if(!glfwInit())
+        return 1;
+
+    const char * glsl_version = "#version 130";
```
becseya: Unused variable

src/drivers/opengles/lv_opengles_window.c
```diff
@@ -0,0 +1,438 @@
...
+    uint8_t * fb2;
+    uint8_t * fb_act;
+    uint8_t * buf1;
+    uint8_t * buf2;
+    uint8_t * rotated_buf;
+    size_t rotated_buf_size;
+
+    uint8_t zoom;
+    uint8_t ignore_size_chg;
+} lv_opengles_window_t;
```
becseya: `SUGGESTION`
The phase *"window"* is a bit misleading, as the struct itself contains a window member. Suggesting `gl_disp_data_t`

src/drivers/opengles/lv_opengles_window.c
```diff
@@ -0,0 +1,438 @@
...
+            // LV_LOG_INFO("Mouse position: (%f, %f)\n", xpos, ypos);
+            _lv_opengles_mouse_btn_handler(disp, 1);
+        }
+        else if(action == GLFW_RELEASE) {
+            _lv_opengles_mouse_btn_handler(disp, 0);
+        }
+    }
+}
+
+void _lv_opengles_mouse_move_handler(lv_display_t * disp, int x, int y);
```
becseya: I think manually defining an external functions instead of using the the other module's header has a negative effect on the readability API over time. The `*_private.h` pattern can be used if we want to hide some functions from the user. But I also seen it's being used elsewhere. @kisvegabor what do you think?

src/drivers/opengles/lv_opengles_window.c
```diff
@@ -0,0 +1,438 @@
...
+
+    const char * glsl_version = "#version 130";
+    glfwWindowHint(GLFW_CONTEXT_VERSION_MAJOR, 3);
+    glfwWindowHint(GLFW_CONTEXT_VERSION_MINOR, 0);
+
+    int32_t hor_res = disp->hor_res;
+    int32_t ver_res = disp->ver_res;
+
+    // Create window with graphics context
+    dsc->window = glfwCreateWindow(hor_res * dsc->zoom, ver_res * dsc->zoom, "hellow world", NULL, NULL);
```
becseya: Setting the window title to `LVGL Simulator` would be more professional.

becseya (changes requested):
I'm excited to see one more driver option for PC, especially since demo was running out of the box. There are no show-stoppers in the PR, however the code style should be aligned with the rest of the code-base at a few points.

Comments marked with `SUGGESTION` are not requested changes on my side, but @kisvegabor  please check them.

src/drivers/opengles/lv_opengles_window.c
danyyliu: It will help a lot if you can provide more info about this bug, since I cannot reproduce it on my machine. I did find another issue of memory, but maybe it has no relationship with this bug.

Kconfig
danyyliu: I run scripts/lv_conf_internal_gen.py, but I cannot get this part generated in lv_conf_internal.h:
#ifndef LV_USE_OPENGLES_DEBUG
    #ifdef CONFIG_LV_USE_OPENGLES_DEBUG
        #define LV_USE_OPENGLES_DEBUG CONFIG_LV_USE_OPENGLES_DEBUG
    #else
        #define LV_USE_OPENGLES_DEBUG       0
    #endif
#endif
I tried to add it manually, but CI would fail. I tried to run scripts/lv_conf_internal_gen.py on several machines, but it didn't work. Maybe there is some actions I missed, please let me know. @kisvegabor @becseya

src/drivers/opengles/lv_opengles_window.c
kisvegabor: You can see if you change the size of window with your mouse.

Kconfig
kisvegabor: `lv_conf_internal.h` is generated from `lv_conf_template.h` and `LV_USE_OPENGLES_DEBUG` is not present in `lv_conf_template.h`.

docs/integration/driver/opengles.rst
kisvegabor: Confirmed and I agree.

src/drivers/opengles/lv_opengles_window.c
kisvegabor: I agree. Instead of `_lv` function, please create `lv_opengles_..._private.c/h` files if needed.

kisvegabor:
See the [CI log](https://github.com/lvgl/lvgl/actions/runs/9229195105/job/25394878914?pr=6254) for warning (treated as errors).

falkTX:
> @falkTX I find it important to have a standalone OpenGL ES driver for LVGL because in the practice it seems vendors add better support for OpenGL ES than for SDL. In some projects that I'm working only OpenGL ES is available but SDL is not.

Well if the glfw3 layer would be separate it would allow for GLES in both glfw3 and SDL drivers.

kisvegabor:
> Well if the glfw3 layer would be separate it would allow for GLES in both glfw3 and SDL drivers.

If we keep both then it makes sense. I guess it will allow adding OpenGL stuff to SDL. In this or a new PR, it would be great to have a function like `lv_opengles_bind_to_texture()` (or so) to make LVGL draw on any textures without a window. I don't know much about OpenGL but it feel related to your suggestion.

src/drivers/opengles/lv_opengles_window.c
danyyliu: I use an ubuntu 22 in virtualbox, and I can change the size of window with mouse:
image
image

I will check the resize function, to see what is not done correctly. Maybe you can let me know which system you are using, so that I can check if it has any relationship with different system.

danyyliu:
> > Well if the glfw3 layer would be separate it would allow for GLES in both glfw3 and SDL drivers.
> 
> If we keep both then it makes sense. I guess it will allow adding OpenGL stuff to SDL. In this or a new PR, it would be great to have a function like `lv_opengles_bind_to_texture()` (or so) to make LVGL draw on any textures without a window. I don't know much about OpenGL but it feel related to your suggestion.

I totally agree with @falkTX and you, the code in this PR contains two parts, as we talked in issue 6017:
The code in opengl driver directory can be split into two parts: GLFW code for window/mouse/etc; OpenGL code for texture management. The GLFW part should stay in this directory, while the texture management should be moved to draw module.

Actually, the API of opengl texture(i.e. lv_opengles_driver.c), is not good either. Too much detail of opengl code is exposed to GLFW code. I will take some time to modify it.

src/drivers/opengles/lv_opengles_window.c
kisvegabor: I'm on Linux Mint 21.3

danyyliu: Please check if this issue is fixed with the latest code. I can produce it in a real windows system, it doesn't crash, while it does show same picture which becseya shown above. It seems system in virtualbox has a different implementation of opengl, or a fake gpu, so the issue can only be reproduced in real system with a real gpu. The reason I see is the data pass to gpu is not aligned correctly, so the picture will not shown correctly, and function glTexImage2D will crash. @becseya @kisvegabor

danyyliu:
I am thinking it's better to rename lv_opengles_windows.c/lv_opengles_mouse.c to lv_glfw_***.c, because these files have no relationship with opengles directly now. If we do this, this opengles directory in driver should be renamed to glfw too. We can leave lv_opengles_driver.c in this directory for now, and move it out after we have a render of opengles. What do you think? @kisvegabor

src/drivers/opengles/lv_opengles_window.c
danyyliu: I changed it to lv_glfw_window_t, this name follows the lv_sdl_window_t. Maybe we can leave it as is, or find a better name for these kinds of structure.

kisvegabor (approved):
Using "glfw" name is okay for me. I suggest merging this PR now and start a new one which will allow using any OpenGL ES textures as LVGL display. A nice example code for that would be to 
1. Create a display with `lv_glfw_window_create()`
2. Create a smaller texture for an LVGL display, e.g. `lv_opengles_texture_create(parent, width, height)`. The `parent_texture` parameter could be get like `lv_glfw_window_get_texture(disp)` or set to any texture.

The gist is to allow using the OpenGL ES driver where there is no window, just OpenGL ES connected to Wayland for example.

The above is only my idea without being an OpenGL expert, so feel free to suggest a better flow :slightly_smiling_face:

liamHowatt (approved):
I was doing a pedantic review but, given that graceful close+reopen / multiple displays does not seem to be in scope for this PR, all of my suggestions are inapplicable.

Also, I can see that some parts are left to be filled in by other PRs, like using `lv_glfw_window_t` members `fb2`, `buf1`, `buf2` if modes other than direct get added.

danyyliu:
> I was doing a pedantic review but, given that graceful close+reopen / multiple displays does not seem to be in scope for this PR, all of my suggestions are inapplicable.
> 
> Also, I can see that some parts are left to be filled in by other PRs, like using `lv_glfw_window_t` members `fb2`, `buf1`, `buf2` if modes other than direct get added.

I agree with you. In this PR, the multiple displays are not supported, the opengl driver APIs don't smell good either. In my mind, this PR is just like a demo, which can just be used as a proof of concept, not a real feature can be used in a real product.

I will go on improving it in the next days, but it will take some time, since I need to go deep into the details of lvgl draw module.

kisvegabor:
Merging, thank you! 

@danyyliu as the next step please focus on making it work with an external texture without a window. It's very important for a customer project. 

After that we can go for OpenGL ES based rendering (similar to [lv_draw_sdl.c](https://github.com/lvgl/lvgl/blob/master/src/draw/sdl/lv_draw_sdl.c)).

PR #6255, opened 2024-05-22
fix(layouts): fix rounding for fr in grid layout

niklasf:
A grid layout containing `LV_GRID_FR(...)` should always cover all available space. But note the red edges produced by the test case:

grid_fr_err

The current approach to deal with rounding errors does not work because the rounding in `(free_w * (col_fr_cnt - last_fr_x)) / col_fr_cnt` (large dividend) does not necessarily match the rounding in `(free_h * f) / row_fr_cnt` (small dividend).

This patch:
* Fixes the calculation to use all available space.
* Previously all space accumulated by rounding errors was assigned to the last fr column. Now differences due to rounding are distributed more evenly (at most 1px difference from ideal size).

Fixed two unrelated warnings in separate commits, so that I could run the tests locally (runner uses `-Werror`).


src/libs/thorvg/tvgSwMemPool.cpp
```diff
@@ -84,7 +84,7 @@ SwMpool* mpoolInit(uint32_t threads)
 {
     auto allocSize = threads + 1;
 
-    auto mpool = static_cast<SwMpool*>(calloc(sizeof(SwMpool), 1));
+    auto mpool = static_cast<SwMpool*>(calloc(1, sizeof(SwMpool)));
```
liamHowatt:
```suggestion
    auto mpool = static_cast<SwMpool*>(calloc(sizeof(SwMpool), 1));
```

liamHowatt (changes requested):
Thanks! Awesome logic and great test! It all makes sense to me.

My only comment is the change to the thorvg file. We overwrite that directory whenever we upgrade the version of thorvg so your fix will get lost. Please submit a PR to https://github.com/thorvg/thorvg for that fix. Otherwise LGTM.

src/libs/thorvg/tvgSwMemPool.cpp
niklasf: Removed that commit and submitted https://github.com/thorvg/thorvg/pull/2321.

liamHowatt (approved):

src/libs/thorvg/tvgSwMemPool.cpp
liamHowatt: Thanks!

kisvegabor (approved):
Thank you!

kisvegabor:
I should have rebased it instead of squash... Sorry...

PR #6256, opened 2024-05-23
feat(refr): updates LVGL's refresh when double buffering is being used.

kdschlosser:
There was no reason to have the wait for the buffer flushing called more than a single time during the refresh. I move it so it is being called right away.  

I also removed the else statement in the wait_for_flushing function that would do the spinning wheels and the automatic setting of disp-> flushing to zero if the the flush_wait_cb has been set. The user is able to call the flush ready function from inside of the callback should they want to notify LVGL that the buffer has actually flushed. That is the reason why I removed the else statement that checks to see if the `disp->flushing` flag has been set. This allows the user to use the callback to complete other tasks while a buffer is flushing. 

It leaves it up to the user to decide if that callback is to be used as a way to stall LVGL without having a spinning wheels while the buffer is transferring or if they want to use that callback to do something else like checking pin states on an MCU in the event there is a wait that has to take place.  

The other thing I added was an exit from doing the refresh if using double buffering instead of sitting there spinning it's wheels. for the most part double buffering and DMA memory use go hand in hand. The benefit it to be able to do other things while the buffer is still being transmitted. LVGL causing a stall is wasted processor time. So when is done is the timer duration gets reduced to a 1 so next time lv_task_handler gets called the refresh will check if the buffer has finished and if it has then it will perform the refresh. This allows other things to get processed instead of wasting processor time. This can be enabled or disabled using the LV_RETURN_FROM_FLUSH_WAIT which is in the config template. 

I also changed the refr_now function so it will inform the user if the refresh has completed or not. It will only return false if double buffer is enabled and LV_RETURN_FROM_FLUSH_WAIT is set to 1 and the refresh doesn't complete. Otherwise it will always return true

I also removed the the pausing of the refresh timer as this wasn't doing anything except going to cause an issue because it was backwards. It was set to pause when the performance monitor was running and it was supposed to be the opposite of that. But seeing as how time spent doing a refresh is time spent it should not be paused to begin with. LVGL function fine without pausing it so there is no need to have it there to begin with. I was also unable to locate where the timer was being unpaused as well so if the performance monitor was running it would have caused issues with the refresh occurring when it should be.

### Notes

### DIFF
--- a/lv_conf_template.h
+++ b/lv_conf_template.h
@@ -74,6 +74,16 @@
 /*Default display refresh, input device read and animation step period.*/
 #define LV_DEF_REFR_PERIOD  33      /*[ms]*/
 
+/* This is used to control what LVGL does if it has to wait for a buffer
+ * to finish flushing. Instead of just spinning the wheels this sets the
+ * refresh timer to 1 millisecond and then exist the refresh should it have to
+ * wait for the buffer to finish flushing. This only happen with double
+ * buffering. If DMA memory is being used it allows the processor to complete
+ * ther tasks instead of just sitting there spinning it's wheels.
+ */
+#define LV_RETURN_FROM_FLUSH_WAIT 0
+
+
 /*Default Dot Per Inch. Used to initialize default sizes such as widgets sized, style paddings.
  *(Not so important, you can adjust it to modify default sizes and spaces)*/
 #define LV_DPI_DEF 130     /*[px/inch]*/
--- a/src/core/lv_refr.c
+++ b/src/core/lv_refr.c
@@ -74,21 +74,33 @@ void _lv_refr_deinit(void)
 {
 }
 
-void lv_refr_now(lv_display_t * disp)
+bool lv_refr_now(lv_display_t * disp)
 {
     lv_anim_refr_now();
-
+    bool ret = true;
     if(disp) {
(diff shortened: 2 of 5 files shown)


kdschlosser:
@kisvegabor , @XuNeo , @PGNetHun 

You guys want to have a look see at this PR. give it some performance tests when using double buffering with the buffers set into DMA memory. The idea is to do some work in the loop between the calls to the task_handler. This should boost performance by allowing other things to be done instead of sitting there spinning the wheels if the buffer is still being transmitted.

kdschlosser:
The current code formatting failure is actually incorrect. The formatting should be exactly how it currently is.

kdschlosser:
I am not exactly sure why the arc test is failing.

It seems that the callback is not being made. I do not know what the inner workings of the tests are to be able to see what is happening and what would cause the failure.

kisvegabor:
Hi Kevin,

I have some concerns with settings the timer period to 1 ms:
1. It allows the user to run any long code between flushes, but in partial mode it can easily happen that the screen is refreshed in 10 parts. If we lose 1 ms on each it's +10 rendering time. 
2. It mimics an OS-like behavior. The original idea is the use `flush_wait_cb` to yield the current task, and return back to it when flushing is ready.  Are there any issues with it?

PR #6265, opened 2024-05-24
feat(env_support): add build-time dependencies for custom cmake

falkTX:
When using cmake and enabling certain features via `lv_conf.h` we need a way to also tell the build system to include and link against those external libs.
This way we are able to use lvgl through cmake in a very simple manner:

```
set(LV_CONF_PATH "${PROJECT_SOURCE_DIR}/lv_conf.h")
set(LV_CONF_USE_LIBINPUT TRUE)
set(LV_CONF_USE_LINUX_DRM TRUE)
add_subdirectory(lvgl)
```

Marking the PR as a draft as I am not sure if I am not missing some other better way of handling this...

### DIFF
--- a/env_support/cmake/custom.cmake
+++ b/env_support/cmake/custom.cmake
@@ -68,6 +68,25 @@ if(NOT (CMAKE_C_COMPILER_ID STREQUAL "MSVC"))
   set_source_files_properties(${LVGL_ROOT_DIR}/src/others/vg_lite_tvg/vg_lite_tvg.cpp PROPERTIES COMPILE_FLAGS -Wunused-parameter)
 endif()
 
+# Add build-time dependencies
+if(LV_CONF_USE_LIBINPUT)
+    find_package(PkgConfig REQUIRED)
+    pkg_check_modules(libinput REQUIRED IMPORTED_TARGET libinput)
+    target_link_libraries(lvgl PUBLIC PkgConfig::libinput)
+endif()
+
+if(LV_CONF_USE_LINUX_DRM)
+    find_package(PkgConfig REQUIRED)
+    pkg_check_modules(libdrm REQUIRED IMPORTED_TARGET libdrm)
+    target_link_libraries(lvgl PUBLIC PkgConfig::libdrm)
+endif()
+
+if(LV_CONF_USE_SDL2)
+    find_package(PkgConfig REQUIRED)
+    pkg_check_modules(sdl2 REQUIRED IMPORTED_TARGET sdl2)
+    target_link_libraries(lvgl PUBLIC PkgConfig::sdl2)
+endif()
+
 # Build LVGL example library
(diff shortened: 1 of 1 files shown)


kisvegabor:
I'm okay with the approach you have suggested, however it shall be documented [here](https://docs.lvgl.io/master/integration/building/cmake.html). Could you add a  few lines about these new options?

kisvegabor:
@falkTX will you have time in the near future to update the docs? No worried if not, just please us know.

falkTX:
> @falkTX will you have time in the near future to update the docs? No worried if not, just please us know.

Hi sorry for late reply, I looked into it quickly but couldnt really tell the best way to handle it.
Should I just add a new chapter on the doc file? There arent any other cmake build options in there yet, so this would make it the first.
If you show me an example diff it would help a lot, I can then fill in the technical details about these new options.

kisvegabor:
I've imagined something like this:
```diff
diff --git a/docs/integration/building/cmake.rst b/docs/integration/building/cmake.rst
index f26497ed6..c8b77a036 100644
--- a/docs/integration/building/cmake.rst
+++ b/docs/integration/building/cmake.rst
@@ -16,8 +16,14 @@ You need to install
 - CMake
 - Ninja (for Linux builds). Be sure to Add ninja to your PATH!
 
-How to build this project using cmake
--------------------------------------
+
+Build variables
+***************
+
+- ``LV_CONF_USE_LINUX_DRM``: Set to ``1`` to enabled DRM. 
+
+How to build LVGL using cmake
+*****************************
 
 The recommended way to build this project is to use the provided CMakePresets.json. This file contains 2 configurations
 
@@ -37,9 +43,9 @@ The recommend way for consuming CMakePresets is a CMakePresets aware IDE such as
 
 Simply load this project into your IDE and select your desired preset and you are good to go.
 
-
 Build with CMake GUI
 --------------------
+
 Open this project with CMake GUI and select your desired preset.
 When hitting the generate button, CMake will create solution files (for VS) or Ninja Files (for Linux Ninja Build)
 
@@ -61,11 +67,10 @@ If you use cmake to install lvgl 3 folders will be created.
 Build with Command line
 -----------------------
 
-You can also build your project using the command line.
-Run the follwoing commands
+You can also build your project using command line.
+Run the follwoing commands:
 
 - cmake --preset windows-base
 - cmake --build --preset windows-base_dbg
 - ctest --preset windows-base_dbg

```

kisvegabor:
@falkTX do you plan to finalize this PR?

falkTX:
yes it is! need it for building locally here so will try to clean it this week.

kisvegabor:
Can we add FreeType, FFMpeg and ThorVG the same way?

kisvegabor:
@falkTX I know you are busy now, but I wonder if you will have time to finish this PR soon. It would be great to make it part of LVGL v9.2.

PR #6277, opened 2024-05-25
fix(display): cancelled screen animation may block input indefinitely

niklasf:
When calling `lv_screen_load_anim()` without `auto_del` while the previous screen change is still ongoing, input may end up being blocked indefinitely:

* Input is blocked while `d->prev_scr` is set
* `scr_load_anim_start` sets `d->prev_scr`
* `scr_anim_completed` unsets `d->prev_src`, but this does not happen if the animation is cancelled

### DIFF
--- a/src/display/lv_display.c
+++ b/src/display/lv_display.c
@@ -602,10 +602,8 @@ void lv_screen_load_anim(lv_obj_t * new_scr, lv_screen_load_anim_t anim_type, ui
 
     d->scr_to_load = new_scr;
 
-    if(d->prev_scr && d->del_prev) {
-        lv_obj_delete(d->prev_scr);
-        d->prev_scr = NULL;
-    }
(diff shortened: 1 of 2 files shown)


liamHowatt (approved):
Nice! :100:

Example for other reviewers' reference:
```c
static void timer_cb(lv_timer_t * tim)
{
    lv_screen_load(lv_timer_get_user_data(tim));
}
static void screen_anim_cancel_test(void)
{
    lv_obj_t * orig_screen = lv_screen_active();
    lv_roller_create(orig_screen);
    lv_obj_t * new_scr = lv_obj_create(NULL);
    lv_roller_create(new_scr);
    lv_screen_load_anim(new_scr, LV_SCR_LOAD_ANIM_MOVE_LEFT, 5000, 5000, false);
    lv_timer_create(timer_cb, 7500, orig_screen);
}
```

kisvegabor (approved):
Thank you!

PR #6278, opened 2024-05-25
feat(obj): add API to set/get object ID.

XuNeo:

Add macro `LV_USE_OBJ_ID_ASSIGN`(default ON) to automatically assign ID to object after it's created.
Add API `lv_obj_[get|set]_id`.
Add API `lv_obj_get_child_by_id `. It could be useful for binding language.

### Notes

### DIFF
--- a/docs/others/obj_id.rst
+++ b/docs/others/obj_id.rst
@@ -14,19 +14,20 @@ Usage
 -----
 
 Enable this feature by setting :c:macro:`LV_USE_OBJ_ID` to `1` in ``lv_conf.h``.
-Use the builtin obj ID generator by setting :c:macro:`LV_USE_OBJ_ID_BUILTIN` to `1`.
-Otherwise provide your own custom implementation.
 
-The ID is automatically generated and assigned to :cpp:expr:`obj->id` during obj's
-construction by calling API :cpp:expr:`lv_obj_assign_id(obj)` from :cpp:func:`lv_obj_constructor`.
+Enable :c:macro:`LV_USE_OBJ_ID_ASSIGN` to automatically assign an ID to object when it's created.
+It's done by calling function :cpp:func:`lv_obj_assign_id` from :cpp:func:`lv_obj_constructor`.
 
-You can directly access the ID by :cpp:expr:`obj->id` or use API :cpp:expr:`lv_obj_stringify_id(obj, buf, len)`
+You can either use you own ID generator by defining the function :cpp:func:`lv_obj_assign_id` or you can utilize the built-in one.
+To use the builtin ID generator, set :c:macro:`LV_USE_OBJ_ID_BUILTIN` to `1`.
+
+You can directly access the ID by :cpp:expr:`lv_obj_get_id(obj)` or use API :cpp:expr:`lv_obj_stringify_id(obj, buf, len)`
 to get a string representation of the ID.
 
 Use custom ID generator
 ~~~~~~~~~~~~~~~~~~~~~~~
 
-Set :c:macro:`LV_USE_OBJ_ID_BUILTIN` to `0` in ``lv_conf.h``. 
+Set :c:macro:`LV_USE_OBJ_ID_BUILTIN` to `0` in ``lv_conf.h``.
 
 Below APIs needed to be implemented and linked to lvgl.
 
@@ -55,3 +56,10 @@ This is useful to debug UI crash. From log we can rebuilt UI the moment before c
 For example, if the obj is stored to a :cpp:expr:`timer->user_data`, but obj is deleted when timer expired.
 Timer callback will crash because of accessing wild pointer.
 From the dump log we can clearly see that the obj does not exist.
+
+Find child by ID
+~~~~~~~~~~~~~~~~
+
+Use API :cpp:expr:`lv_obj_t * lv_obj_get_child_by_id(const lv_obj_t * obj, void * id, bool (compare)(lv_obj_t *, void * id));` to find a child by ID.
+It will walk through all children and return the first child with the given ID.
+If the ``compare`` callback is NULL, the ID is determined solely by its pointer. Alternatively, you can provide your own callback function to customize the evaluation.
--- a/docs/others/obj_id.rst
+++ b/docs/others/obj_id.rst
@@ -36,6 +36,7 @@ Below APIs needed to be implemented and linked to lvgl.
     void lv_obj_assign_id(const lv_obj_class_t * class_p, lv_obj_t * obj);
     void lv_obj_free_id(lv_obj_t * obj);
     const char * lv_obj_stringify_id(lv_obj_t * obj, char * buf, uint32_t len);
+    int lv_obj_id_compare(void * id1, void * id2);
 
 
 :cpp:func:`lv_obj_assign_id` is called when an object is created. The object final class is passed from
@@ -60,6 +61,5 @@ From the dump log we can clearly see that the obj does not exist.
 Find child by ID
 ~~~~~~~~~~~~~~~~
 
-Use API :cpp:expr:`lv_obj_t * lv_obj_get_child_by_id(const lv_obj_t * obj, void * id, bool (compare)(lv_obj_t *, void * id));` to find a child by ID.
+Use API :cpp:expr:`lv_obj_t * lv_obj_get_child_by_id(const lv_obj_t * obj, void * id);` to find a child by ID.
 It will walk through all children and return the first child with the given ID.
-If the ``compare`` callback is NULL, the ID is determined solely by its pointer. Alternatively, you can provide your own callback function to customize the evaluation.
--- a/docs/others/obj_id.rst
+++ b/docs/others/obj_id.rst
@@ -15,10 +15,10 @@ Usage
 
 Enable this feature by setting :c:macro:`LV_USE_OBJ_ID` to `1` in ``lv_conf.h``.
 
-Enable :c:macro:`LV_USE_OBJ_ID_ASSIGN` to automatically assign an ID to object when it's created.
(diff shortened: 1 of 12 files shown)


src/core/lv_obj.c
```diff
@@ -309,10 +313,55 @@ bool lv_obj_is_valid(const lv_obj_t * obj)
...
+{
+    obj->id = id;
+}
+
+void * lv_obj_get_id(const lv_obj_t * obj)
+{
+    return obj->id;
+}
+
+lv_obj_t * lv_obj_get_child_by_id(const lv_obj_t * obj, void * id, bool (compare)(lv_obj_t *, void * id))
```
kisvegabor: IMO we shouldn't pass `compare` in each call as it's the same in the whole app. Instead we can expect the user to either 
1. implement it in the app with a  fixed name (e.g. `lv_obj_id_compare`)
2. set it as a global callback
 
As an alternative to use can create a wrapper like:
```
lv_obj_t * my_lv_obj_get_child_by_id(const lv_obj_t * obj, void * id)
{
   return lv_obj_get_child_by_id(obj, id, my_compare);
}
```

lv_conf_template.h
```diff
@@ -333,8 +333,11 @@
 /* Add `id` field to `lv_obj_t` */
 #define LV_USE_OBJ_ID           0
 
+/* Automatically assign an ID when obj is created */
+#define LV_USE_OBJ_ID_ASSIGN    1
```
```suggestion
# if LV_USE_OBJ_ID != 0
/* Automatically assign an ID when obj is created. */
#define LV_OBJ_ID_AUTO_ASSIGN    1
```

lv_conf_template.h
```diff
@@ -333,8 +333,11 @@
...
+
 /* Use lvgl builtin method for obj ID */
-#define LV_USE_OBJ_ID_BUILTIN   0
+#define LV_USE_OBJ_ID_BUILTIN   1
```
kisvegabor, line 345: For the comment:
```
/*Use the built obj ID handler functions:
* - lv_obj_assign_id:  Called when a widget is created. Use a separate counter for each widget class as an ID.
* - lv_obj_id_compare: Compare the counters to decide if it matches with a requested value.
* - lv_obj_stringify_id:  Return e.g. "button3"
* - lv_obj_free_id: Does nothing, as there is no memory allocation  for the ID.
* When disabled these functions needs to be implemented by the user.*/
```

docs/others/obj_id.rst
```diff
@@ -14,19 +14,20 @@ Usage
...
-Use the builtin obj ID generator by setting :c:macro:`LV_USE_OBJ_ID_BUILTIN` to `1`.
-Otherwise provide your own custom implementation.
 
-The ID is automatically generated and assigned to :cpp:expr:`obj->id` during obj's
-construction by calling API :cpp:expr:`lv_obj_assign_id(obj)` from :cpp:func:`lv_obj_constructor`.
+Enable :c:macro:`LV_USE_OBJ_ID_ASSIGN` to automatically assign an ID to object when it's created.
+It's done by calling function :cpp:func:`lv_obj_assign_id` from :cpp:func:`lv_obj_constructor`.
 
-You can directly access the ID by :cpp:expr:`obj->id` or use API :cpp:expr:`lv_obj_stringify_id(obj, buf, len)`
+You can either use you own ID generator by defining the function :cpp:func:`lv_obj_assign_id` or you can utilize the built-in one.
```
kisvegabor:
```suggestion
You can either use your own ID generator by defining the function :cpp:func:`lv_obj_assign_id` or you can utilize the built-in one.
```

lv_conf_template.h
XuNeo: I found out we can use `#define LV_OBJ_ID_AUTO_ASSIGN    LV_USE_OBJ_ID ` So it's automatically enabled and always defined. 

Do you think it's acceptable?

lv_conf_template.h
XuNeo: Changed to `* - lv_obj_id_compare:      Compare the ID to decide if it matches with a requested value.` since the ID could also not be a number.

src/core/lv_obj.c
```diff
@@ -309,6 +310,40 @@ bool lv_obj_is_valid(const lv_obj_t * obj)
     return false;
 }
 
+#if LV_USE_OBJ_ID
+void lv_obj_set_id(lv_obj_t * obj, void * id)
+{
+    obj->id = id;
```
FASTSHIFT, line 317: LV_ASSERT_NULL(obj);

src/core/lv_obj.c
```diff
@@ -309,6 +310,40 @@ bool lv_obj_is_valid(const lv_obj_t * obj)
...
+}
+
+void * lv_obj_get_id(const lv_obj_t * obj)
+{
+    return obj->id;
```
FASTSHIFT, line 323: LV_ASSERT_NULL(obj);

src/core/lv_obj.h
```diff
@@ -405,23 +405,64 @@ const lv_obj_class_t * lv_obj_get_class(const lv_obj_t * obj);
...
+ */
+void * lv_obj_get_id(const lv_obj_t * obj);
 
 /**
- * Assign an id to an object if not previously assigned
- * Set `LV_USE_OBJ_ID_BUILTIN` to 1 to use builtin method to generate object ID.
- * Otherwise, these functions including `lv_obj_[assign|free|stringify]_id` should be implemented externally.
+ * Get the child object by its id.
+ * It will check children and grandchildren recursively.
+ * Compare is done by pointer comparison if `match` is NULL.
```
liamHowatt: Line no longer applies

src/core/lv_obj.h
```diff
@@ -405,23 +405,64 @@ const lv_obj_class_t * lv_obj_get_class(const lv_obj_t * obj);
...
+ *
+ * @param obj       pointer to an object
+ * @param id        the id of the child object
+ * @return          pointer to the child object or NULL if not found
+ */
+lv_obj_t * lv_obj_get_child_by_id(const lv_obj_t * obj, void * id);
+
+#if LV_OBJ_ID_AUTO_ASSIGN
+/**
+ * Assign d to object if not previously assigned.
```
liamHowatt:
```suggestion
 * Assign id to object if not previously assigned.
```

src/core/lv_obj.h
```diff
@@ -405,23 +405,64 @@ const lv_obj_class_t * lv_obj_get_class(const lv_obj_t * obj);
...
+ * It will check children and grandchildren recursively.
+ * Compare is done by pointer comparison if `match` is NULL.
+ *
+ * @param obj       pointer to an object
+ * @param id        the id of the child object
+ * @return          pointer to the child object or NULL if not found
+ */
+lv_obj_t * lv_obj_get_child_by_id(const lv_obj_t * obj, void * id);
+
+#if LV_OBJ_ID_AUTO_ASSIGN
```
liamHowatt: Shouldn't these prototypes be defined even if `LV_OBJ_ID_AUTO_ASSIGN` is `0`?

FASTSHIFT (approved):

lv_conf_template.h
liamHowatt: I do think kisvegabor's suggestion matches the existing pattern in the config file. Configs that depend on another config are inside an `#if ... #endif` and indented.

W-Mai (approved):

lv_conf_template.h
XuNeo: The generated conf.h looks like this.
```c
/* Automatically assign an ID when obj is created */
#ifndef LV_OBJ_ID_AUTO_ASSIGN
    #ifdef CONFIG_LV_OBJ_ID_AUTO_ASSIGN
        #define LV_OBJ_ID_AUTO_ASSIGN CONFIG_LV_OBJ_ID_AUTO_ASSIGN
    #else
        #define LV_OBJ_ID_AUTO_ASSIGN   LV_USE_OBJ_ID
    #endif
#endif
```

So LV_OBJ_ID_AUTO_ASSIGN is always defined, and default to 1 if `LV_USE_OBJ_ID` is enabled.

This is what we want(always define the macro with correct value).

liamHowatt: I understand that you want to use `LV_OBJ_ID_AUTO_ASSIGN` as a sort of compound macro that, when `1`, means `LV_USE_OBJ_ID && (!defined(LV_OBJ_ID_AUTO_ASSIGN) || (defined(LV_OBJ_ID_AUTO_ASSIGN) && LV_OBJ_ID_AUTO_ASSIGN != 0))`.

Until now, using defines like `LV_DEMO_MUSIC_LARGE` when `LV_USE_DEMO_MUSIC=0` is "undefined behavior". Code that uses `LV_DEMO_MUSIC_LARGE` first checks that `LV_USE_DEMO_MUSIC == 1` before considering the value of `LV_DEMO_MUSIC_LARGE`. In your changes, you use `LV_OBJ_ID_AUTO_ASSIGN` as a shortcut.

It makes sense, but I cannot say whether it's okay to break the old pattern for the first time. @kisvegabor?

XuNeo: You are correct. This breaks the old pattern.
For system using Kconfig, it normally uses `#ifdef` instead of `#if `. 

LVGL follow the latter style, thus it makes sense to discuss whether we should always define those macros, either enabled as 1 or disabled as 0.

I can follow the old pattern in this PR and raise another discussion if that's more appropriate.

kisvegabor (approved):

kisvegabor:
@liamHowatt please review and merge it if all look good.

lv_conf_template.h
kisvegabor: @liamHowatt just pointed out this unresolved thread.

If we use `#if` it would really make more sense to **not** add guards around the child configs.  Let's keep it as it is in this feature, and discuss it a new issue.

liamHowatt (approved):

PR #6287, opened 2024-05-27
fix(obj) search child object using dfs(depth-first search)

jaeho2021:
Change-Id: I9a8b3dd664e7fb2701feaeb170e77283c28a5ff2

To support the object dump function, all children of an object are searched using a depth-first search method. however, When the first leaf node is encountered, the for statement is exited using a break statement and the sibling node is not searched.

### DIFF
--- a/src/core/lv_obj_tree.c
+++ b/src/core/lv_obj_tree.c
@@ -623,7 +623,7 @@ static lv_obj_tree_walk_res_t dump_tree_core(lv_obj_t * obj, int32_t depth)
         for(uint32_t i = 0; i < obj->spec_attr->child_cnt; i++) {
             res = dump_tree_core(lv_obj_get_child(obj, i), depth + 1);
             if(res == LV_OBJ_TREE_WALK_END)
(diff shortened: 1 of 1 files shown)


src/core/lv_obj_tree.c
```diff
@@ -623,7 +623,7 @@ static lv_obj_tree_walk_res_t dump_tree_core(lv_obj_t * obj, int32_t depth)
         for(uint32_t i = 0; i < obj->spec_attr->child_cnt; i++) {
             res = dump_tree_core(lv_obj_get_child(obj, i), depth + 1);
             if(res == LV_OBJ_TREE_WALK_END)
```
XuNeo: Let's remove `res` and make `dump_tree_core` returns `void`.

jaeho2021: okay, it looks work well. please review second patch.

XuNeo (approved):

kisvegabor (approved):

PR #6291, opened 2024-05-28
feat(drivers): GLCDC support for RX72 family

becseya:
This PR extends the GLCDC driver for the RX72 family


src/drivers/display/renesas_glcdc/lv_renesas_glcdc.c
```diff
@@ -3,14 +3,38 @@
...
+ *********************/
+
+#ifdef _RENESAS_RA_
+    #define USE_FREE_RTOS (BSP_CFG_RTOS == 2)
+#else // RX with SMC code generation
+    #ifndef _RENESAS_RX_
+        #define _RENESAS_RX_ 1
+    #endif
+    #define USE_FREE_RTOS 1
+    #define BYTES_PER_PIXEL 2
```
liamHowatt: Maybe define this in all cases and use it where you have `w * 2`?

liamHowatt (dismissed):
Nice! I really couldn't spot anything. Driver code is allowed to be a bit flexible. I understand this is high-ish priority so I approve.

docs/requirements.txt
```diff
@@ -1,4 +1,5 @@
 Sphinx
+alabaster
```
liamHowatt: Please also add `sphinx-design`.

liamHowatt (commented):
Great docs! When I try building the docs from a fresh venv, I needed to add `sphinx-design` to the `requirements.txt` so please add that. Soon we might add a check to the PR CI so we catch this sort of thing before merging :P

Also, if you try building the docs locally, you'll want to merge/rebase master to get the fix for a docs building issue.

docs/requirements.txt
becseya: Nice catch. `alabaster` is not needed actually, it probably got mixed up while looking proper package. Did a check with a clean venv.

I was under the assumption the CI does build the docs.

liamHowatt (approved):

liamHowatt:
Thanks!

kisvegabor (approved):
Thank you!

PR #6298, opened 2024-05-29
feat(nxp/pxp): add Zephyr Support

faxe1008:

Reopening of [https://github.com/lvgl/lvgl/pull/6159](https://github.com/lvgl/lvgl/pull/6159) since it got merged and reverted the original PR does not track changes anymore (@nicusorcitu ). 

### Notes

### DIFF
--- a/src/draw/nxp/pxp/lv_pxp_osa.c
+++ b/src/draw/nxp/pxp/lv_pxp_osa.c
@@ -25,6 +25,11 @@
     #include "semphr.h"
 #endif
 
+#if defined(__ZEPHYR__)
+    #include <zephyr/kernel.h>
+    #include <zephyr/irq.h>
+#endif
+
 /*********************
  *      DEFINES
  *********************/
@@ -64,6 +69,9 @@ static void _pxp_wait(void);
 #if defined(SDK_OS_FREE_RTOS)
     static SemaphoreHandle_t xPXPIdleSemaphore;
 #endif
+#if defined(__ZEPHYR__)
+    static K_SEM_DEFINE(s_pxpIdleSem, 0, 1);
+#endif
 static volatile bool ucPXPIdle;
 
 static pxp_cfg_t _pxp_default_cfg = {
@@ -97,11 +105,20 @@ void PXP_IRQHandler(void)
         priority task.  The macro used for this purpose is dependent on the port in
         use and may be called portEND_SWITCHING_ISR(). */
         portYIELD_FROM_ISR(xHigherPriorityTaskWoken);
+#elif defined(__ZEPHYR__)
+        k_sem_give(&s_pxpIdleSem);
 #else
         ucPXPIdle = true;
 #endif
     }
 }
+
+#if defined(__ZEPHYR__)
+static void PXP_ZephyrIRQHandler(void *)
+{
+    PXP_IRQHandler();
+}
+#endif
 
 pxp_cfg_t * pxp_get_default_cfg(void)
 {
@@ -119,17 +136,22 @@ static void _pxp_interrupt_init(void)
     PXP_ASSERT_MSG(xPXPIdleSemaphore, "xSemaphoreCreateBinary failed!");
 
     NVIC_SetPriority(PXP_IRQ_ID, configLIBRARY_MAX_SYSCALL_INTERRUPT_PRIORITY + 1);
+    NVIC_EnableIRQ(PXP_IRQ_ID);
+#elif defined(__ZEPHYR__)
+    IRQ_CONNECT(DT_IRQN(DT_NODELABEL(pxp)), CONFIG_LV_Z_PXP_INTERRUPT_PRIORITY, PXP_ZephyrIRQHandler, NULL, 0);
+    irq_enable(DT_IRQN(DT_NODELABEL(pxp)));
 #endif
     ucPXPIdle = true;
-
-    NVIC_EnableIRQ(PXP_IRQ_ID);
 }
 
 static void _pxp_interrupt_deinit(void)
 {
-    NVIC_DisableIRQ(PXP_IRQ_ID);
 #if defined(SDK_OS_FREE_RTOS)
+    NVIC_DisableIRQ(PXP_IRQ_ID);
     vSemaphoreDelete(xPXPIdleSemaphore);
+#elif defined(__ZEPHYR__)
+    irq_disable(DT_IRQN(DT_NODELABEL(pxp)));
+    k_sem_reset(&s_pxpIdleSem);
 #endif
 }
 
@@ -156,6 +178,9 @@ static void _pxp_wait(void)
 
     if(xSemaphoreTake(xPXPIdleSemaphore, portMAX_DELAY) == pdTRUE)
         ucPXPIdle = true;
+#elif defined(__ZEPHYR__)
+    if(k_sem_take(&s_pxpIdleSem, K_FOREVER) == 0)
+        s_pxpIdle = true;
 #else
     while(ucPXPIdle == false) {
     }
--- a/src/draw/nxp/pxp/lv_pxp_osa.c
+++ b/src/draw/nxp/pxp/lv_pxp_osa.c
@@ -18,11 +18,11 @@
 #if LV_USE_DRAW_PXP
 #include "lv_pxp_utils.h"
 #include "../../../misc/lv_log.h"
+#include "../../../osal/lv_os.h"
 #include "fsl_pxp.h"
 
 #if defined(SDK_OS_FREE_RTOS)
     #include "FreeRTOS.h"
-    #include "semphr.h"
 #endif
 
 #if defined(__ZEPHYR__)
@@ -62,15 +62,19 @@ static void _pxp_run(void);
  */
 static void _pxp_wait(void);
 
+#if defined(__ZEPHYR__)
+    /**
+    * Interrupt handler for Zephyr IRQ
+    */
+    static void _pxp_zephyr_irq_handler(void *);
+#endif
+
 /**********************
  *  STATIC VARIABLES
  **********************/
 
-#if defined(SDK_OS_FREE_RTOS)
-    static SemaphoreHandle_t xPXPIdleSemaphore;
-#endif
-#if defined(__ZEPHYR__)
-    static K_SEM_DEFINE(s_pxpIdleSem, 0, 1);
+#if LV_USE_OS
+    static lv_thread_sync_t pxp_sync;
 #endif
 static volatile bool ucPXPIdle;
 
@@ -91,34 +95,15 @@ static pxp_cfg_t _pxp_default_cfg = {
 
 void PXP_IRQHandler(void)
 {
-#if defined(SDK_OS_FREE_RTOS)
-    BaseType_t xHigherPriorityTaskWoken = pdFALSE;
-#endif
-
     if(kPXP_CompleteFlag & PXP_GetStatusFlags(PXP_ID)) {
         PXP_ClearStatusFlags(PXP_ID, kPXP_CompleteFlag);
-#if defined(SDK_OS_FREE_RTOS)
-        xSemaphoreGiveFromISR(xPXPIdleSemaphore, &xHigherPriorityTaskWoken);
-
-        /* If xHigherPriorityTaskWoken is now set to pdTRUE then a context switch
-        should be performed to ensure the interrupt returns directly to the highest
-        priority task.  The macro used for this purpose is dependent on the port in
-        use and may be called portEND_SWITCHING_ISR(). */
-        portYIELD_FROM_ISR(xHigherPriorityTaskWoken);
-#elif defined(__ZEPHYR__)
-        k_sem_give(&s_pxpIdleSem);
+#if LV_USE_OS
+        lv_thread_sync_signal_isr(&pxp_sync);
 #else
         ucPXPIdle = true;
 #endif
     }
 }
-
-#if defined(__ZEPHYR__)
-static void PXP_ZephyrIRQHandler(void *)
-{
-    PXP_IRQHandler();
-}
-#endif
 
 pxp_cfg_t * pxp_get_default_cfg(void)
 {
@@ -129,29 +114,45 @@ pxp_cfg_t * pxp_get_default_cfg(void)
  *   STATIC FUNCTIONS
  **********************/
 
+#if defined(__ZEPHYR__)
+static void _pxp_zephyr_irq_handler(void *)
+{
+    PXP_IRQHandler();
+}
+#endif
+
 static void _pxp_interrupt_init(void)
 {
-#if defined(SDK_OS_FREE_RTOS)
-    xPXPIdleSemaphore = xSemaphoreCreateBinary();
-    PXP_ASSERT_MSG(xPXPIdleSemaphore, "xSemaphoreCreateBinary failed!");
-
-    NVIC_SetPriority(PXP_IRQ_ID, configLIBRARY_MAX_SYSCALL_INTERRUPT_PRIORITY + 1);
-    NVIC_EnableIRQ(PXP_IRQ_ID);
-#elif defined(__ZEPHYR__)
-    IRQ_CONNECT(DT_IRQN(DT_NODELABEL(pxp)), CONFIG_LV_Z_PXP_INTERRUPT_PRIORITY, PXP_ZephyrIRQHandler, NULL, 0);
-    irq_enable(DT_IRQN(DT_NODELABEL(pxp)));
+#if LV_USE_OS
+    if(lv_thread_sync_init(&pxp_sync) != LV_RESULT_OK) {
+        PXP_ASSERT_MSG(false, "Failed to init thread_sync.");
+    }
 #endif
+
+#if defined(__ZEPHYR__)
+    IRQ_CONNECT(DT_IRQN(DT_NODELABEL(pxp)), CONFIG_LV_Z_PXP_INTERRUPT_PRIORITY, _pxp_zephyr_irq_handler, NULL, 0);
+    irq_enable(DT_IRQN(DT_NODELABEL(pxp)));
+#elif defined(SDK_OS_FREE_RTOS)
+    NVIC_SetPriority(PXP_IRQ_ID, configLIBRARY_MAX_SYSCALL_INTERRUPT_PRIORITY + 1);
+#endif
+
+#if !defined(__ZEPHYR__)
+    NVIC_EnableIRQ(PXP_IRQ_ID);
+#endif
+
     ucPXPIdle = true;
 }
 
 static void _pxp_interrupt_deinit(void)
 {
-#if defined(SDK_OS_FREE_RTOS)
-    NVIC_DisableIRQ(PXP_IRQ_ID);
-    vSemaphoreDelete(xPXPIdleSemaphore);
-#elif defined(__ZEPHYR__)
+#if defined(__ZEPHYR__)
     irq_disable(DT_IRQN(DT_NODELABEL(pxp)));
-    k_sem_reset(&s_pxpIdleSem);
+#else
+    NVIC_DisableIRQ(PXP_IRQ_ID);
+#endif
+
+#if LV_USE_OS
+    lv_thread_sync_delete(&pxp_sync);
 #endif
 }
 
@@ -171,16 +172,11 @@ static void _pxp_run(void)
  */
 static void _pxp_wait(void)
 {
-#if defined(SDK_OS_FREE_RTOS)
-    /* Return if PXP was never started, otherwise the semaphore will lock forever. */
     if(ucPXPIdle == true)
         return;
-
-    if(xSemaphoreTake(xPXPIdleSemaphore, portMAX_DELAY) == pdTRUE)
+#if LV_USE_OS
+    if(lv_thread_sync_wait(&pxp_sync) == LV_RESULT_OK)
         ucPXPIdle = true;
-#elif defined(__ZEPHYR__)
-    if(k_sem_take(&s_pxpIdleSem, K_FOREVER) == 0)
-        s_pxpIdle = true;
 #else
     while(ucPXPIdle == false) {
     }
--- a/src/osal/lv_freertos.c
+++ b/src/osal/lv_freertos.c
@@ -329,40 +329,19 @@ lv_result_t lv_thread_sync_signal_isr(lv_thread_sync_t * pxCond)
     /* If the cond is uninitialized, perform initialization. */
     prvCheckCondInit(pxCond);
 
-    /* Acquire the mutex. */
-    if(pdFALSE == xSemaphoreTakeFromISR(pxCond->xSyncMutex, NULL)) {
-        return LV_RESULT_INVALID;
-    }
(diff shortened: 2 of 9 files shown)


nicusorcitu:
@cristian-stoica

src/draw/nxp/pxp/lv_pxp_osa.c
```diff
@@ -18,11 +18,13 @@
...
 #include "lv_pxp_utils.h"
 #include "../../../misc/lv_log.h"
+#include "../../../osal/lv_os.h"
 #include "fsl_pxp.h"
 
-#if defined(SDK_OS_FREE_RTOS)
-    #include "FreeRTOS.h"
-    #include "semphr.h"
+
+#if defined(__ZEPHYR__)
```
nicusorcitu, line 28: Those should not be included - they will get over lv_os.h inclusion

src/draw/nxp/pxp/lv_pxp_osa.c
```diff
@@ -114,22 +113,32 @@ pxp_cfg_t * pxp_get_default_cfg(void)
...
+    if(lv_thread_sync_init(&pxp_sync) != LV_RESULT_OK) {
+        PXP_ASSERT_MSG(false, "Failed to init thread_sync.");
+    }
+#endif
 
+#if defined(SDK_OS_FREE_RTOS)
     NVIC_SetPriority(PXP_IRQ_ID, configLIBRARY_MAX_SYSCALL_INTERRUPT_PRIORITY + 1);
+    NVIC_EnableIRQ(PXP_IRQ_ID);
+#elif defined(__ZEPHYR__)
+    IRQ_CONNECT(DT_IRQN(DT_NODELABEL(pxp)), CONFIG_LV_Z_PXP_INTERRUPT_PRIORITY, PXP_ZephyrIRQHandler, NULL, 0);
```
nicusorcitu: I understand now, you included the ZEPHYR header just for this `CONFIG_LV_Z_PXP_INTERRUPT_PRIORITY`.

We should not mix any longer this with lv_osal support.

src/draw/nxp/pxp/lv_pxp_osa.c
```diff
@@ -119,17 +136,22 @@ static void _pxp_interrupt_init(void)
     PXP_ASSERT_MSG(xPXPIdleSemaphore, "xSemaphoreCreateBinary failed!");
 
     NVIC_SetPriority(PXP_IRQ_ID, configLIBRARY_MAX_SYSCALL_INTERRUPT_PRIORITY + 1);
```
cristian-stoica, line 136: With PXP=1 and OS_NONE the build fails on MCUXpresso SDK projects. This is a common build configuration for us. The build fails because SDK_OS_FREE_RTOS is defined (by MCUX) but configLIBRARY_MAX_SYSCALL_INTERRUPT_PRIORITY is no longer found.

Since you similarly need CONFIG_LV_Z_PXP_INTERRUPT_PRIORITY for __ZEPHYR__ we could probably use a common abstraction for interrupt priory that will resolve the build for both cases. What do you think we should do about this?

nicusorcitu: hi @faxe1008 
Can you please address this issue? Try to not mix anylonger the `SDK_OS_FREE_RTOS` /`__ZEPHYR__` with `LV_USE_OS`.
We can fix by ourself the other problem with multithreading mentioned [here](https://github.com/lvgl/lvgl/pull/6159#discussion_r1593353421) but please fixup this PR before merging.

src/draw/nxp/pxp/lv_pxp_osa.c
faxe1008: The includes are needed for the entire interrupt logic (retrieving nodes from DT, hooking up the IRQ).

src/draw/nxp/pxp/lv_pxp_osa.c
faxe1008: @nicusorcitu I am not quite sure if I understood everything correctly but I ammended the commit with a change that hopefully fixes this.

nicusorcitu: Actually it makes it worst because now there is no distinction at all between FreeRTOS and Bare Metal for PXP.

What I men was to have kind of an abstraction os function like "lv_os_interrupt_enable()" or "lv_os_interrupt_disable()" but I figure out that this is basically something platform dependent and not OS dependent.
@cristian-stoica suggested a simpler way to have only an abstracted define in lv osal:
```
  #define LV_OS_INTERRUPT_PRIORITY configLIBRARY_MAX_SYSCALL_INTERRUPT_PRIORITY +1
#elif LV_USE_OS == LV_OS_ZEPHYR
  #define LV_OS_INTERRUPT_PRIORITY CONFIG_LV_Z_PXP_INTERRUPT_PRIORITY
#endif
```
But we still do not get rid of the extra include headers for FreeRTOS or Zephyr, correct?

To keep things simple and conclude on this PR:
1. please revert your last changes from [here](https://github.com/lvgl/lvgl/compare/fd1e2466a5133625483e819ee3bf08738d80b9f2..158a93acf415225cac7c74f0d1a0536a76b793fa)
2. add back the inclusion necessary for freertos irq priority (for freertos)
```
#if defined(SDK_OS_FREE_RTOS)
    #include "FreeRTOS.h"
#endif
```

faxe1008: Performed the suggested changes :^)

nicusorcitu: I do not see them. Can you please point a link?

cristian-stoica:
-#if defined(SDK_OS_FREE_RTOS)
-    #include "FreeRTOS.h"
-    #include "semphr.h"
+
+#if defined(__ZEPHYR__)
+    #include <zephyr/kernel.h>
+    #include <zephyr/irq.h>
 #endif
 
 I have a problem with the code above. It is not exactly OSAL as it claims and it also breaks our build. There has to be a better way

src/draw/nxp/pxp/lv_pxp_osa.c
faxe1008: I forgot to push the changes and was out of office for the week. Pushed it right now :^)

faxe1008:
> ```
> 
> -#if defined(SDK_OS_FREE_RTOS)
> -    #include "FreeRTOS.h"
> -    #include "semphr.h"
> +
> +#if defined(__ZEPHYR__)
> +    #include <zephyr/kernel.h>
> +    #include <zephyr/irq.h>
> #endif
> ```
> 
> I have a problem with the code above. It is not exactly OSAL as it claims and it also breaks our build. There has to be a better way

Yeah, these includes are needed for the interrupts. Personally I think that an OSAL is maybe not the right place for an abstraction for this, the OSAL should be kept as thin as possible. Maybe it would be an option somehow provide the user a way to give LVGL a `setup` and `teardown` method for any of the drawing engines so that this can be handled by each OS?

src/draw/nxp/pxp/lv_pxp_osa.c
```diff
@@ -83,26 +88,23 @@ static pxp_cfg_t _pxp_default_cfg = {
...
-#if defined(SDK_OS_FREE_RTOS)
-        xSemaphoreGiveFromISR(xPXPIdleSemaphore, &xHigherPriorityTaskWoken);
-
-        /* If xHigherPriorityTaskWoken is now set to pdTRUE then a context switch
-        should be performed to ensure the interrupt returns directly to the highest
-        priority task.  The macro used for this purpose is dependent on the port in
-        use and may be called portEND_SWITCHING_ISR(). */
-        portYIELD_FROM_ISR(xHigherPriorityTaskWoken);
+#if LV_USE_OS
+        lv_thread_sync_signal(&pxp_sync);
```
nicusorcitu: The is still one issue:
`lv_thread_sync_signal()` was not ment to work work from ISR.

The implementation from ISR should be using FreeRTOS `vTaskNotifyGiveFromISR()` instead of `xTaskNotifyGive()`.

nicusorcitu: We should add a new osal API in FreeRTOS (and same for Zephyr) like `lv_thread_sync_sygnal_from_isr()`.

faxe1008: I added this method to the OSAL, however I would need help from experts of all other plattforms if I have done it correctly.

nicusorcitu: The API does not look good on FreeRTOS either.
1. You should use `vTaskNotifyGiveFromISR()`
2. From documentation:
```
vTaskNotifyGiveFromISR() will set *pxHigherPriorityTaskWoken to pdTRUE if sending the notification caused a task to unblock, and the unblocked task has a priority higher than the currently running task.

If vTaskNotifyGiveFromISR() sets this value to pdTRUE then a context switch should be requested before the interrupt is exited. See the example below. 
```


```
/* The transmit end interrupt. */
void vTransmitEndISR( void )
{
BaseType_t xHigherPriorityTaskWoken = pdFALSE;

    /* At this point xTaskToNotify should not be NULL as a transmission was
    in progress. */
    configASSERT( xTaskToNotify != NULL );

    /* Notify the task that the transmission is complete. */
    vTaskNotifyGiveIndexedFromISR( xTaskToNotify, 0, &xHigherPriorityTaskWoken );

    /* There are no transmissions in progress, so no tasks to notify. */
    xTaskToNotify = NULL;

    /* If xHigherPriorityTaskWoken is now set to pdTRUE then a context switch
    should be performed to ensure the interrupt returns directly to the highest
    priority task.  The macro used for this purpose is dependent on the port in
    use and may be called portEND_SWITCHING_ISR(). */
    portYIELD_FROM_ISR( xHigherPriorityTaskWoken );
}
```

src/osal/lv_freertos.c
```diff
@@ -318,6 +318,52 @@ lv_result_t lv_thread_sync_delete(lv_thread_sync_t * pxCond)
     return LV_RESULT_OK;
 }
 
+lv_result_t lv_thread_sync_signal_isr(lv_thread_sync_t * sync)
+{
+#if USE_FREERTOS_TASK_NOTIFY
+    /* Send a notification to the task waiting. */
+    vTaskNotifyGiveFromISR(pxCond->xTaskToNotify, NULL);
```
nicusorcitu: Please don't use NULL.
Take care of xHigherPriorityTaskWoken --> I already copy-paste an example on how to use it.

src/osal/lv_freertos.c
```diff
@@ -318,6 +318,52 @@ lv_result_t lv_thread_sync_delete(lv_thread_sync_t * pxCond)
     return LV_RESULT_OK;
 }
 
+lv_result_t lv_thread_sync_signal_isr(lv_thread_sync_t * sync)
```
nicusorcitu: This API need to be tested with USE_FREERTOS_TASK_NOTIFY(task_notify implementation) or without USE_FREERTOS_TASK_NOTIFY (semaphores implementation).

kisvegabor:
We got a lot request about updating Zephyr to LVGL v9 so I wonder if we can help to make it happen soon. @becseya is interesting Zephyr, so he can help with LVGL related issues. We also have several NXP boards so we can help with PXP as well.

src/osal/lv_freertos.c
faxe1008: Applied the suggestion, also for the other branch. Hopefully correct since I am not an Free RTOS expert :^)

kisvegabor:
>   155 | lv_result_t lv_thread_sync_signal_isr(lv_thread_sync_t * sync)

Please add `LV_UNUSED(sync);`

faxe1008:
The function var had to be renamed to `pxCond` :^)

kisvegabor:
> The function var had to be renamed to pxCond

Which one? :slightly_smiling_face:

nicusorcitu:
image

As a general rule, inside the FreeRTOS files I have used FreeRTOS style while naming variables, functions...
Thanks for keeping it on the same manner.

nicusorcitu (approved):

kisvegabor:
Please add `lv_thread_sync_signal_isr` to `lv_os_none.c` too.

faxe1008:
@kisvegabor added it to `lv_os_none` :^)

kisvegabor (dismissed):

src/draw/nxp/pxp/lv_pxp_osa.c
```diff
@@ -83,26 +88,23 @@ static pxp_cfg_t _pxp_default_cfg = {
...
 #endif
     }
 }
 
+#if defined(__ZEPHYR__)
+static void PXP_ZephyrIRQHandler(void *)
+{
+    PXP_IRQHandler();
+}
+#endif
```
liamHowatt: Please put this in the static functions section and forward declare it. Maybe also change the name to make it seem less like a platform specific global function definition (unless it is).

src/osal/lv_freertos.c
```diff
@@ -318,6 +318,57 @@ lv_result_t lv_thread_sync_delete(lv_thread_sync_t * pxCond)
...
+    vTaskNotifyGiveFromISR(pxCond->xTaskToNotify, &xHigherPriorityTaskWoken);
+    portYIELD_FROM_ISR(xHigherPriorityTaskWoken);
+#else
+    /* If the cond is uninitialized, perform initialization. */
+    prvCheckCondInit(pxCond);
+
+    /* Acquire the mutex. */
+    if(pdFALSE == xSemaphoreTakeFromISR(pxCond->xSyncMutex, NULL)) {
+        return LV_RESULT_INVALID;
+    }
```
liamHowatt: If this fails because `lv_thread_sync_wait` has the mutex, will the signal never be sent and a deadlock would occur?

liamHowatt (commented):
Nice atomic CAS loop!

In another PR we should correct the implementation of `lv_mutex_lock_isr` for cmsis_rtos2. See [this comment](https://github.com/lvgl/lvgl/pull/6182#pullrequestreview-2084004168).

src/draw/nxp/pxp/lv_pxp_osa.c
faxe1008: Ah it kinda is - I do not know how many other RTOS give the possibility to add userdata to an ISR.

src/osal/lv_freertos.c
faxe1008: True - what would be the best solution for this? Not using the mutex at all and disabling interrupts instead so we can not be preempted by another ISR?

faxe1008: See my last commit, if the change is ok from your side, I'll squash it :^)

src/draw/nxp/pxp/lv_pxp_osa.c
liamHowatt: What I mean is the name implies that it's overriding a `weak` definition in the OS like a systick handler that gets called by the OS, but it's really just a local function.

Also, does `PXP_IRQHandler` need to be global? Does it get called by anything external?

src/osal/lv_freertos.c
liamHowatt: Yes, I think that should work, nice. Is the CAS loop no longer needed now?

src/draw/nxp/pxp/lv_pxp_osa.c
faxe1008: Ah, no its really just a wrapper to accept the `void *` that the Zephyr impl provides. 
`PXP_IRQHandler` does not need to be global - but that I did not touch. Should I make it a static one as well?

src/osal/lv_freertos.c
faxe1008: Correct, I removed the CAS loop.

liamHowatt: Did you? :wink:

faxe1008: Now I did :'D

liamHowatt: Oh, you pushed right after I sent that :laughing:

src/draw/nxp/pxp/lv_pxp_osa.c
liamHowatt: Yeah, it's best to not add to the global namespace if it can be avoided. For now, just make `PXP_IRQHandler` static and move this heading above `PXP_IRQHandler` please, and it all looks good :slightly_smiling_face:

```c
/**********************
 *   STATIC FUNCTIONS
 **********************/
```

src/osal/lv_freertos.c
kisvegabor: @liamHowatt please approve if all looks good.

kisvegabor (dismissed):

src/draw/nxp/pxp/lv_pxp_osa.c
faxe1008: Performed the change and renamed the handler to match the other non namespace scoped functions.

liamHowatt (dismissed):
Thank you!

src/draw/nxp/pxp/lv_pxp_osa.c
```diff
@@ -81,27 +98,12 @@ static pxp_cfg_t _pxp_default_cfg = {
  *   GLOBAL FUNCTIONS
  **********************/
 
-void PXP_IRQHandler(void)
+#if defined(__ZEPHYR__)
```
nicusorcitu: Please do not remove the PXP_IRQHandler(). @faxe1008

faxe1008: Within LVGL this is not used outside of the osa file - is this something that needs to be callable from the outside?

nicusorcitu: `PXP_IRQHandler()` is what the name suggest - the handler of the PXP IRQ.

It is similar to what you already added `PXP_ZephyrIRQHandler` except that in case of our platform (w/ or w/o FreeRTOS) the PXP handler is mapped in the vector table outside of the lvgl code.

liamHowatt: Okay, so that function _is_ called from outside? I wasn't clear on that.

Even though it wasn't perfect, I approved yesterday because it was harmless and close enough. But the naming is still confusing because `PXP_IRQHandler` is actually called from outside and the name is important, because it maps to something in the symbol table as you say, while `PXP_ZephyrIRQHandler` is not special yet has a name that makes it seem like it maps to something in the symbol table.

nicusorcitu: In this case we can rename the `PXP_ZephyrIRQHandler` as this is a static function.
But `PXP_IRQHandler` need to stay in place with exactly the same name.

Thanks.

faxe1008: Reverted this part of the change moved the zephyr handler to the static section.

src/draw/nxp/pxp/lv_pxp_osa.c
```diff
@@ -81,22 +93,12 @@ static pxp_cfg_t _pxp_default_cfg = {
  *   GLOBAL FUNCTIONS
  **********************/
 
-void PXP_IRQHandler(void)
+void PXP_IRQHandler(void);
```
faxe1008: Gotta remove the semicolon :^)

src/draw/nxp/pxp/lv_pxp_osa.c
```diff
@@ -112,24 +114,45 @@ pxp_cfg_t * pxp_get_default_cfg(void)
...
+#elif defined(SDK_OS_FREE_RTOS)
     NVIC_SetPriority(PXP_IRQ_ID, configLIBRARY_MAX_SYSCALL_INTERRUPT_PRIORITY + 1);
 #endif
-    ucPXPIdle = true;
 
+#if !defined(__ZEPHYR__)
     NVIC_EnableIRQ(PXP_IRQ_ID);
+#endif
+
+    ucPXPIdle = true;
```
nicusorcitu, line 143: Please move this line before NVIC_EnableIRQ() like before.

src/draw/nxp/pxp/lv_pxp_osa.c
```diff
@@ -57,12 +62,19 @@ static void _pxp_run(void);
...
 /**********************
  *  STATIC VARIABLES
  **********************/
 
-#if defined(SDK_OS_FREE_RTOS)
-    static SemaphoreHandle_t xPXPIdleSemaphore;
+#if LV_USE_OS
+    static lv_thread_sync_t pxp_sync;
 #endif
 static volatile bool ucPXPIdle;
```
nicusorcitu, line 79: As this file no longer refer to FreeRTOS only, please rename this variable to "pxp_idle". Has nothing to do directly with your changes but please include this change as well - if not big deal. Thanks.

src/draw/nxp/pxp/lv_pxp_osa.c
```diff
@@ -112,24 +114,45 @@ pxp_cfg_t * pxp_get_default_cfg(void)
  *   STATIC FUNCTIONS
  **********************/
```
nicusorcitu, line 116: Why do you ever need the static `_pxp_zephyr_irq_handler`? You can use directly `PXP_IRQHandler()` for zephyr as well.

faxe1008: I can not use it, the `IRQ_CONNECT` macro needs a function with a `(void)(void*)` signature:
[https://docs.zephyrproject.org/apidoc/latest/group__isr__apis.html#ga131739d1faf501a15590053817aba984](https://docs.zephyrproject.org/apidoc/latest/group__isr__apis.html#ga131739d1faf501a15590053817aba984)

kisvegabor:
Hi,

What is missing to finalize this PR and for LVGL v9 Zephyr support in general? I'd like to be sure all are added before releasing v9.2.

faxe1008:
@kisvegabor at least from my side there are no more blockers to this PR. I guess some testing from the other RTOSes might be helpful to avoid regressions.

As for the requirements of updating zephyr to v9.2 the only thing left for me is to figure out what to do about STM32 DMA2D support, as we have added it in 8.4 but I am unsure if it can be used in 9.1.

nicusorcitu:
I have already approved, no more issues from my side.

nicusorcitu (approved):

liamHowatt (approved):

kisvegabor (approved):

kisvegabor:
Great! Is the missing DMA2D is blocking to have LVGL v9.2 in Zephyr? We got closer to ST, so now we can add DMA2D support but it might take some time.

faxe1008:
> Great! Is the missing DMA2D is blocking to have LVGL v9.2 in Zephyr? We got closer to ST, so now we can add DMA2D support but it might take some time.

People using the new version on STM32 boards will experience a performance hit, but imo that should be accetable to some degree. Btw. I started drafting this implementation over here:
https://github.com/faxe1008/lvgl/commit/6497837e396fb2b91804158f7e1aed667a69f0a6
There is still a lot of stuff missing, I used the old implementation as a kind of guide line and some datasheets, but I am by no means an expert on ChromArt.

kisvegabor:
@liamHowatt is also looking into it now. I'd like to have at least something basic in v9.2.

PR #6313, opened 2024-05-31
feat(draw): add a configuration to reduce code size

zjanosy:

Added config options to selectively disable color format support in order to reduce code size. By default all formats are enabled. 

NOTE: some features use various color formats internally, e.g., gradients use RGB888, transparency uses ARGB8888. Use at your own risk!

For example, on the ESP32-S3 platform using an RGB565 display enabling only RGB565 and ARGB8888 (for transparent bitmaps) reduces the size of the LVGL code by 32 kB, i.e., about 20% (156 kB -> 124 kB). Disabling ARGB8888 as well results in an additional 6 kB reduction.


FASTSHIFT (approved):

kisvegabor (approved):
Thank you Zoltan!

TimJTi:
@zjanosy @kisvegabor 
Hi - I'm finding that if I use Master, this commit is causing nothing to be displayed on my LCD and the new Kconfig entries are not showing when I run make menuconfig, despite, obviously, `LV_USE_DRAW_SW` being set
If I revert the change, it works OK.

Probably something daft I'm doing as this is the first time I have used Master (as I want to submit a PR myself) but if you have any suggestions I'd be grateful!

kisvegabor:
> Vaguely remember some NuttX discussion about this - so the solution is imperfect but I understand it now.

Yes, that's the case. I think at least we should document it [here](https://docs.lvgl.io/master/integration/os/nuttx.html). Do you have any ideas to detect the not updated Kconfig? Maybe test `LV_VERSION_MAJOR/MINOR` somehow?
cc @XuNeo

XuNeo:
> Maybe test LV_VERSION_MAJOR/MINOR somehow?

I was thinking the same. Let's open a new issue to track it. https://github.com/lvgl/lvgl/issues/6576

PR #6327, opened 2024-06-03
perf(draw): improve execute time of transforming widget

jaeho2021:
Change-Id: I7fb91f884dc23c01f38e074bf13a666ca39d7293

I think,
if obj doesn't have any transform, there is no benefit to execute lv_obj_get_transformed_area(), however it result in increasing  execute time. 
also, eventually, transform_point function check that obj has any transform and  if there isn't any transform, just return, do nothing. 

In order to avoid performing unnecessary lv_obj_transform_point functions, the condition for determining whether a transform has been applied to an object has been moved a little forward.

Please review whether there are any parts that may cause problems when the condition that determines whether or not to perform the transform is moved to the front.

in my case, 
after applying this patch, improve execute time of lv_obj_get_transformed_area function from 147.4us to 13.9us
Although the difference in the execution time of the function is minimal, it is believed that there will be an improvement effect if the function is performed repeatedly.

### DIFF
--- a/src/core/lv_obj_pos.c
+++ b/src/core/lv_obj_pos.c
@@ -781,6 +781,12 @@ void lv_obj_transform_point(const lv_obj_t * obj, lv_point_t * p, bool recursive
 void lv_obj_get_transformed_area(const lv_obj_t * obj, lv_area_t * area, bool recursive,
                                  bool inv)
 {
+    int32_t angle = lv_obj_get_style_transform_rotation(obj, 0);
+    int32_t scale_x = lv_obj_get_style_transform_scale_x_safe(obj, 0);
+    int32_t scale_y = lv_obj_get_style_transform_scale_y_safe(obj, 0);
+
+    if(angle == 0 && scale_x == LV_SCALE_NONE && scale_y == LV_SCALE_NONE) return;
+
     lv_point_t p[4] = {
         {area->x1, area->y1},
         {area->x1, area->y2},


kisvegabor:
Hey,

Thank you for opening this PR however I think we cannot apply such a simplification. Later in this function `lv_obj_transform_point` is called which applies the transformation from the parent too (if any). Therefore we cannot just check the values in the starting widget but we need to check them on the parents too.

jaeho2021:
thank you for reviewing  my opinion. 
I have a question about you mentioned. 

the lv_obj_area_is_visible function calls the lv_obj_get_transformed_area function several times (if object have parent). 
At that time, the lv_obj_get_transformed_area() function is called with obj first time, and then the lv_obj_get_transformed_area() function is called while following the parent of obj until it is not null.
Although there are some conditional statements, (lv_obj_flag_hidden, lv_obj_flag_overflow_visible), it seems that the parent's properties you mentioned can all be covered through that statement.
If I misunderstood, please comment.

jaeho2021:
hi kisvegabor. 
I understood what you mentioned. 
probably, I think you want to tell me below situation

lv_obj_t* container = lv_obj_create();
lv_obj_t* btn = lv_btn_create(container);
lv_obj_t* label = lv_label_create(btn);

if container has transformation properties, 
in this case, button/label object should do transformation because of parent's transformation properties. 

is it right what I understood?

kisvegabor:
> if container has transformation properties,
in this case, button/label object should do transformation because of parent's transformation properties.
is it right what I understood?

Exactly! :slightly_smiling_face:

PR #6329, opened 2024-06-03
feat(property): add property name

XuNeo:

Binding language like `lua` or `js` prefer to use name to set property.
This patch is one of the series to make bindings easier to set widget property.

Added a table to convert property name to ID. The name table is generated automatically though `properties.py` script, based on the property ID definitions.

If the name is not provided for specified widget, then recursively looking up its base classes. `bsearch` is used to match up strings.

The test has passed.

This API is experimental and is subject to change if binding language requires differently.

### Notes


src/widgets/property/lv_style_properties.c
```diff
@@ -0,0 +1,130 @@
...
+#include "../../core/lv_obj_property.h"
+
+#if LV_USE_OBJ_PROPERTY && LV_USE_OBJ_PROPERTY_NAME
+
+
+/**
+ * Style property names, name must be in order.
+ * Generated code from properties.py
+ */
+const lv_property_name_t style_property_names[111] = {
```
FASTSHIFT: lv_style_property_names

scripts/properties.py
```diff
@@ -0,0 +1,153 @@
+#!/usr/bin/env python3
```
FASTSHIFT, line 1: How do we prevent users from forgetting to call properties.py to update this file after adding a style?

XuNeo: I'm planning to add property.py to CI similar to lv_conf_generate.py. Let's firstly review if the proposed method is the right way.

W-Mai (approved):

W-Mai:
LGTM

docs/others/obj_property.rst
```diff
@@ -35,6 +37,8 @@ Two APIs are provided to get/set widget properties. It can be enabled by setting
     lv_result_t lv_obj_set_property(lv_obj_t * obj, const lv_property_t * value);
     lv_property_t lv_obj_get_property(lv_obj_t * obj, lv_prop_id_t id);
 
+    lv_prop_id_t lv_obj_property_name2id(const lv_obj_class_t * clz, const char * name);
```
kisvegabor:
```suggestion
    lv_prop_id_t lv_obj_get_property_id(const lv_obj_class_t * clz, const char * name);
```

kisvegabor:
It seems to be very useful for us with the editor too.  It looks good to me, but @coevin @liamHowatt and @richardgazdik could you also take look at it?

docs/others/obj_property.rst
XuNeo: I recall that micropython has a naming constraint that the first parameter for lv_obj_xxx APIs should be `lv_obj_t *`.
Is it still true?

`lv_obj_get_` sounds like to get something from an object. But here it intends to convert a property name to ID based on its class. How about `lv_obj_property_get_id`?

src/widgets/property/lv_obj_property_names.h
```diff
@@ -0,0 +1,18 @@
...
+#define LV_OBJ_PROPERTY_NAMES_H
+
+#include "../../misc/lv_types.h"
+
+#if LV_USE_OBJ_PROPERTY && LV_USE_OBJ_PROPERTY_NAME
+
+    extern const lv_property_name_t lv_style_property_names[111];
+    extern const lv_property_name_t lv_image_property_names[11];
+    extern const lv_property_name_t lv_obj_property_names[48];
+    /* *INDENT-ON* */
```
XuNeo: Remove.

coevin (approved):
Looks good to me!

docs/others/obj_property.rst
kisvegabor: I've just copied the name from the [code](https://github.com/lvgl/lvgl/pull/6329/files#diff-c9ad3a7d72566d008dd5214f92aae68ce51e57b0aa2de961cc373e448698f9aaR174). :slightly_smiling_face: 

But you are right, the `lv_obj_property` prefix seems more appropriate.

src/core/lv_obj_property.c
```diff
@@ -165,23 +207,37 @@ static lv_result_t obj_property(lv_obj_t * obj, lv_prop_id_t id, lv_property_t *
...
+                    }
+                case LV_PROPERTY_TYPE_COLOR: {
+                        if(set)((lv_property_set_color_t)prop->setter)(obj, value->color);
+                        else value->color = ((lv_property_get_color_t)(prop->getter))(obj);
+                        break;
+                    }
+                case LV_PROPERTY_TYPE_POINT: {
+                        lv_point_t * point = &value->point;
+                        if(set)((lv_property_set_pointer_t)(prop->setter))(obj, point);
+                        else *point = ((lv_property_get_point_t)(prop->getter))(obj);
```
liamHowatt: Should `lv_property_set_pointer_t` be `lv_property_set_point_t`? You declared `lv_property_set_point_t` at the top.

liamHowatt (commented):
Cool metaprogramming and use of bsearch :slightly_smiling_face:

src/widgets/property/lv_obj_properties.c
```diff
@@ -0,0 +1,68 @@
...
+ * Obj widget property names, name must be in order.
+ * Generated code from properties.py
+ */
+/* *INDENT-OFF* */
+const lv_property_name_t lv_obj_property_names[48] = {
+    {"flag_adv_hittest",       LV_PROPERTY_OBJ_FLAG_ADV_HITTEST,},
+    {"flag_checkable",         LV_PROPERTY_OBJ_FLAG_CHECKABLE,},
+    {"flag_click_focusable",   LV_PROPERTY_OBJ_FLAG_CLICK_FOCUSABLE,},
+    {"flag_clickable",         LV_PROPERTY_OBJ_FLAG_CLICKABLE,},
+    {"flag_end",               LV_PROPERTY_OBJ_FLAG_END,},
```
kisvegabor, line 22: Can we avoid adding it?

XuNeo: Are you concerned about the flash size? `flag` is treated as a property so it's automatically included here. Maybe we should remove the `flag` property totally?

liamHowatt (approved):

FASTSHIFT (approved):

Issue #6305, opened 2024-05-30
Tiny TTF should be caching stbtt_FindGlyphIndex


codewitch-honey-crisis:
### LVGL version

9.1

### What happened?

TinyTTF currently does not cache the results of stbtt_FindGlyphIndex but it really really should, and should be using the cache in the font glyph desc retrieval callback. stbtt_FindGlyphIndex is probably the most expensive call in the lib outside of rendering.

I can't imagine the cache is incredibly effective right now.

MacDragon:
I can confirm it really really hurts current performance, I've not had time to go back and try to do a 'real' cache implementation, but I quickly hacked a local one into our project and it increased performance well over an order of magnitude for a busy text screen. The quick hack I did was mentioned in one of the lvgl 9 closed cache threads the other day, and a suggestion was given to copy ideas from freetype implementatino to tiny ttf.

codewitch-honey-crisis:
Thanks for you feedback. I'm the original contributor of this library which I adapted from some public domain code for game engines :), and I implemented the cache quickly as it was a requirement for LVGL. I need to do some legwork if I was to upgrade this code, because I do not yet fully understand the LVGL cache system yet. I learned just enough of it to be dangerous.

MacDragon:
With the additional quick 'n dirty cacheing in place, it now works pretty well for us, difference in speed vs plain bitmap fonts is negligible now, at least in environment with plenty enough RAM to cache everything - I did also increase the max amount of elements allowed into the cache quite a bit, though I haven't tried to verify how much difference that made.

codewitch-honey-crisis:
I appreciate this. It's all very helpful. I'll tinker with the caching once I solve some unrelated issues I'm having with the new LVGL 9+

codewitch-honey-crisis:
@kisvegabor can I get some guidance on using the cache? I'm not clear on how the search key is used below:
```cpp
const lv_font_t * font = g_dsc->resolved_font;
tiny_ttf_cache_data_t search_key = {
    .font = (lv_font_t *)font,
    .unicode = unicode_letter,
    .size = font->line_height,
    //.glyph_index
    //.draw_buf
};```
I've added glyph_index. Like draw_buf it's to be fetched FROM the cache, not used to search for the data. Does that make sense? How does `lv_cache_acquire_or_create(tiny_ttf_cache, &search_key, (void *)font->dsc);` know to leave draw_buf out of the search criteria? I need the same for glyph_index.

[...]

Issue #6271, opened 2024-05-24
tiny_ttf ignores cache_size in lv_tiny_ttf_create_file_ex function


cybekRT:
### LVGL version

v9.1.0, master

### What happened?

The documentation ( https://docs.lvgl.io/master/libs/tiny_ttf.html ) states that the default cache size is 4kB and can be modified by calling the function with `_ex` suffix. However in the code ( https://github.com/lvgl/lvgl/blob/master/src/libs/tiny_ttf/lv_tiny_ttf.c#L273 ) the cache_size parameter is unused.

kisvegabor:
Hi,

@W-Mai is our cache expert. :slightly_smiling_face: 
Could you take a look at it?

W-Mai:
> Hi,
> 
> @W-Mai is our cache expert. 🙂 Could you take a look at it?

Ok, let me take a look in it. 😁

cybekRT:
Hi, thank you very much for your effort! I've checked and it looks like the cache is properly created and used now :)
Unfortunately, looks like the stb_truetype_htcw doesn't use cache, so in case when the font is in flash (file api), it still does many uncached reads. But even so, now it is definitely faster than previously, so thanks again for your work!

W-Mai:
> Hi, thank you very much for your effort! I've checked and it looks like the cache is properly created and used now :) Unfortunately, looks like the stb_truetype_htcw doesn't use cache, so in case when the font is in flash (file api), it still does many uncached reads. But even so, now it is definitely faster than previously, so thanks again for your work!
> 
> One thing worth noting, the max cache size parameter is applied to two caches, so if user requests limit of 32kB of cache, it will use up to 64kB, 32kB for glyphs and 32kB for bitmap. There is probably no better way to provide one limit for two caches, but maybe worth noting in the documentation or somewhere.

I don't know if you know that lvgl's file system has cache function. You can take a look at this related api.
- LV_FS_XXXX_CACHE_SIZE

cybekRT:
> I don't know if you know that lvgl's file system has cache function. You can take a look at this related api.

I know, but if I understand correctly, it caches the linear space of file, and since the ttf library is doing many seeks, the cache size must be huge enough to work. It would be nice to have a ttf tables cached, but I understand that it is huge task and I do not complain. I am just doing some project on the ESP32E without spiram, and I was surprised that rendering short text from ttf is taking like about 30 seconds (it was much more before the caching added to tinyttf :) ).

[...]

Issue #6308, opened 2024-05-31
Line 11 of src/libs/tiny_ttf.c declared incorrectly


codewitch-honey-crisis:
### LVGL version

9.0

### What happened?

All recent versions with Tiny TTF
Line 11 of the C file
```cpp
#if LV_USE_TINY_TTF
```
should be
```cpp
#if LV_USE_TINY_TTF != 0
```

I would fix it but I'm not set up to make a PR right now.

kisvegabor:
I asked @W-Mai to fix it in an open PR.

W-Mai:
> I asked @W-Mai to fix it in an open PR.

Done!

PR #6333, opened 2024-06-05
refactor(tiny_fft): refactor tiny_ttf cache

W-Mai:

Tiny_ttf has been reconstructed:
1. cache of uincode(letter) -> glyph_dsc  @codewitch-honey-crisis @MacDragon
2. lv_tiny_ttf_create_xxxx_ex really affect the cache size @cybekRT
3. all tiny_ttf use a same cache -> each font use separate caches  

Theoretically, it accelerates the creation speed of tiny_ttf text.

- https://github.com/lvgl/lvgl/issues/4835#issuecomment-2134459861
- #6305 
- #6271 

cc @kisvegabor 

Could you @MacDragon help test whether there is a performance improvement?

### Notes

### DIFF
--- a/src/libs/tiny_ttf/lv_tiny_ttf.h
+++ b/src/libs/tiny_ttf/lv_tiny_ttf.h
@@ -37,10 +37,6 @@ lv_font_t * lv_tiny_ttf_create_file(const char * path, int32_t font_size);
 lv_font_t * lv_tiny_ttf_create_file_ex(const char * path, int32_t font_size, size_t cache_size);
 #endif
 
-void lv_tiny_ttf_init(void);
-
-void lv_tiny_ttf_deinit(void);
-
 /* create a font from the specified data pointer with the specified line height.*/
 lv_font_t * lv_tiny_ttf_create_data(const void * data, size_t data_size, int32_t font_size);
 
--- a/Kconfig
+++ b/Kconfig
@@ -1187,6 +1187,10 @@ menu "LVGL configuration"
 			bool "Enable loading Tiny TTF data from files"
 			default n
 			depends on LV_USE_TINY_TTF
+		config LV_TINY_TTF_CACHE_GLYPH_CNT
+			bool "Tiny ttf cache entries count"
+			default 256
+			depends on LV_USE_TINY_TTF
 
 		config LV_USE_RLOTTIE
 			bool "Lottie library"
--- a/lv_conf_template.h
+++ b/lv_conf_template.h
@@ -754,6 +754,7 @@
 #if LV_USE_TINY_TTF
     /* Enable loading TTF data from files */
     #define LV_TINY_TTF_FILE_SUPPORT 0
+    #define LV_TINY_TTF_CACHE_GLYPH_CNT 256
 #endif
 
 /*Rlottie library*/
--- a/src/core/lv_global.h
+++ b/src/core/lv_global.h
@@ -174,10 +174,6 @@ typedef struct _lv_global_t {
     struct _lv_freetype_context_t * ft_context;
 #endif
 
-#if LV_USE_TINY_TTF
-    lv_cache_t * tiny_ttf_cache;
-#endif
-
 #if LV_USE_FONT_COMPRESSED
     lv_font_fmt_rle_t font_fmt_rle;
 #endif
--- a/src/libs/tiny_ttf/lv_tiny_ttf.c
+++ b/src/libs/tiny_ttf/lv_tiny_ttf.c
@@ -8,7 +8,7 @@
  *********************/
 #include "../../lvgl.h"
 
-#if LV_USE_TINY_TTF
+#if LV_USE_TINY_TTF != 0
 
 #define font_draw_buf_handlers &(LV_GLOBAL_DEFAULT()->font_draw_buf_handlers)
 
@@ -458,4 +458,4 @@ static lv_cache_compare_res_t tiny_ttf_draw_data_cache_compare_cb(const tiny_ttf
     return 0;
 }
 
-#endif
+#endif // LV_USE_TINY_TTF != 0
--- a/src/libs/tiny_ttf/lv_tiny_ttf.c
+++ b/src/libs/tiny_ttf/lv_tiny_ttf.c
@@ -308,7 +308,7 @@ static lv_font_t * lv_tiny_ttf_create(const char * path, const void * data, size
 #endif
 
     /*Init cache*/
-    dsc->glyph_cache = lv_cache_create(& lv_cache_class_lru_rb_count, sizeof(tiny_ttf_cache_data_t), cache_size,
+    dsc->glyph_cache = lv_cache_create(&lv_cache_class_lru_rb_count, sizeof(tiny_ttf_glyph_cache_data_t), cache_size,
     (lv_cache_ops_t) {
         .compare_cb = (lv_cache_compare_cb_t)tiny_ttf_glyph_cache_compare_cb,
         .create_cb = (lv_cache_create_cb_t)tiny_ttf_glyph_cache_create_cb,
@@ -316,7 +316,7 @@ static lv_font_t * lv_tiny_ttf_create(const char * path, const void * data, size
     });
     lv_cache_set_name(dsc->glyph_cache, "TINY_TTF_GLYPH");
 
-    dsc->draw_data_cache = lv_cache_create(& lv_cache_class_lru_rb_count, sizeof(tiny_ttf_cache_data_t), cache_size,
+    dsc->draw_data_cache = lv_cache_create(&lv_cache_class_lru_rb_count, sizeof(tiny_ttf_cache_data_t), cache_size,
     (lv_cache_ops_t) {
         .compare_cb = (lv_cache_compare_cb_t)tiny_ttf_draw_data_cache_compare_cb,
         .create_cb = (lv_cache_create_cb_t)tiny_ttf_draw_data_cache_create_cb,
--- a/src/libs/tiny_ttf/lv_tiny_ttf.c
+++ b/src/libs/tiny_ttf/lv_tiny_ttf.c
@@ -10,15 +10,11 @@
 
 #if LV_USE_TINY_TTF
 
-#include "../../core/lv_global.h"
-
 #define font_draw_buf_handlers &(LV_GLOBAL_DEFAULT()->font_draw_buf_handlers)
 
 /*********************
  *      DEFINES
  *********************/
-
-#define CACHE_NAME  "TINY_TTF"
 
 #define STB_RECT_PACK_IMPLEMENTATION
 #define STBRP_STATIC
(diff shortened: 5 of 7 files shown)


MacDragon:
I'll check it when current time critical stuff is done, hopefully friday. I'll see if I can import latest master tree into a working branch to run comparisons on, as @codewitch-honey-crisis mentioned that there was already noticeable rendering speed improvements without since 9.1 release due to other changes it's probably best to try and run the different options against newer tree.

W-Mai:
It's a strange bug. Why do the coordinates move up a pixel? 🤔

W-Mai:
fixed, it's stupid mistake 😄💦

image

C47D:
> fixed, it's a stupid mistake 😄💦
> 
> image
> 
> 

Nice font, which is it?

W-Mai:
> > fixed, it's a stupid mistake 😄💦
> > image
> 
> Nice catch and also nice font, which is it?

Pixeloid Mono 😁
I'm glad you also like this font.

- https://www.dafont.com/pixeloid-mono.font

src/libs/tiny_ttf/lv_tiny_ttf.h
```diff
@@ -37,10 +37,6 @@ lv_font_t * lv_tiny_ttf_create_file(const char * path, int32_t font_size);
 lv_font_t * lv_tiny_ttf_create_file_ex(const char * path, int32_t font_size, size_t cache_size);
```
kisvegabor, line 37: Please fix the API comments of these functions.

kisvegabor (commented):
Thanks @W-Mai! It looks good in general. I cannot test the performance difference  now, but if someone verifies it, we are good to go. 

Just one comment:

kisvegabor:
Could you also take care of this?
https://github.com/lvgl/lvgl/issues/6308

kisvegabor:
And a good catch here: #6340

MacDragon:
Today ended up a fair bit busier than I had anticipated, but monday I should  be clear to do some comparison trialing.

MacDragon:
Need to do a little more testing, but currently I do not appear to see any difference between master branch and applying the cache update atop:

Master branch:

 24.06.10 12:02.42      4.033 flush done  5:38 -> 315:190 81us 47583
 24.06.10 12:02.42      4.034 render done 284942us
redraw:
 24.06.10 12:05.28    170.187 flush done  5:38 -> 315:190 93us 47583
 24.06.10 12:05.28    170.187 render done 263750us
 
 
Cache branch:
 24.06.10 12:13.26      3.276 flush done  5:38 -> 315:190 82us 47583
 24.06.10 12:13.26      3.276 render done 286044us
 redraw:
 24.06.10 12:13.52     29.059 flush done  5:38 -> 315:190 93us 47583
 24.06.10 12:13.52     29.059 render done 264225us

Need to verify it's actually active and then do another comparison tomorrow morning.

MacDragon:
Ok, I'm further investigating this today, I think I possibly have some improvements - I'm first fixing up my hacky implementation to be more correct, and then seeing if I can apply some fixes here and see if they help.

kisvegabor:
Great, thank you!

MacDragon:
still investigating, it looks like the really expensive part of the call for me is actually the kerning lookup, and my original implementation had completely ignored this as unecessary in my use case - as I was using a monospaced font, so there should be no kerning.

It might be that need an option to disable kerning on font creation, when know don't need it - kerning lookups can probably also be cached in some form too.

codewitch-honey-crisis:
I'd just like to add here, if it's helpful, that anything Tiny TTF has to do a table lookup for is going to be necessarily somewhat slow, and slower the larger the table. The reason being is that it has to do a lookup over the TTF/OTF file as it exists on disk/flash. It puts memory at a premium and so it does not create indexed lookup tables for this stuff although it will use a binary search if the TTF/OTF file keeps a particular table in sorted order. Might be helpful to bear in mind, as it affects kerning lookup, glyph lookup, etc. This is actually I think the best case scenario, because we're effectively punting faster lookups to the cache which is better managed than if TinyTTF was creating those tables itself. Also @MacDragon thanks for your help on this. I simply don't have time presently to make improvements to the code, so I appreciate all your hard work.

codewitch-honey-crisis:
After thinking about your comment some more @MacDragon it occurs to me that you can't actually technically disable the kerning for a font from a TTF/OTF even if monospaced. The issue is that each glyph has an "advance" property that tells the font how many pixels to advance along the X axis. That's a per glyph lookup even for monospaced fonts. At best, you could cache that for monospaced fonts, but frankly, I think it would be easier on the user and more flexible to integrate the advance factor into the cache.

MacDragon:
Yes, that's effectively what I'm looking to do, cache it and use default spacing for char if no next char or kerning flagged off.

MacDragon:
Ok, in my current case rendering has now improved to:

24.06.12 12:01.31     10.702 flush done  5:38 -> 315:190 142us 47583
 24.06.12 12:01.31     10.702 render done 10806us

Though this is a pretty idealistic state with I think effectively no kerning really happening - need to find some test case that actually triggers real kerning need. But yes, with a few tweaks this is already getting much better now for my present use case at least.

W-Mai:
> Ok, in my current case rendering has now improved to:
> 
> 24.06.12 12:01.31 10.702 flush done 5:38 -> 315:190 142us 47583 24.06.12 12:01.31 10.702 render done 10806us
> 
> Though this is a pretty idealistic state with I think effectively no kerning really happening - need to find some test case that actually triggers real kerning need. But yes, with a few tweaks this is already getting much better now for my present use case at least.

Although kerning cannot be completely banned, can you also test the code of the cache branch to remove the calculation of kerning, will it reduce the time consumption?

MacDragon:
Yes, there is definitely gains to be had by a few less floating point calculations.

MacDragon:
With kerning on, but no actual kerning required to be.
 24.06.12 13:03.12      2.664 flush done  5:38 -> 315:190 130us 47583
 24.06.12 13:03.12      2.664 render done 16588us
after cached and redrawn:
 24.06.12 13:03.23     13.459 flush done  5:38 -> 315:190 131us 47583
 24.06.12 13:03.23     13.459 render done 9366us

In current test I seem to be seeing effectively no difference between kerning on and off - though I don't think the fon't I'm testing with ever triggers real kerning so I think it's a bad test case.. I'll try to find some case tomorrow morning that does, but optimising by adding no kerning calculation to the cache creation seems to have cut another 1ms off my case after cacheing done, though this seems to be about the same whether kerning enabled or not. I'll continue tomorrow morning for a bit and try and find a different test case.

MacDragon:
Also of note, I am testing on an MPU with hardware floating point, the gains may well be more notable on software only. Once I've got something that seems reasonable to test further what's best way to submit?

MacDragon:
Currently working on this again, I don't seem to be reliably getting unicode_letter_next for kerning checks in lv_label?

MacDragon:
Ok, now I think the kerning enable is working properly:
no kern first draw:

 24.06.13 08:12.28      6.428 flush done  5:38 -> 315:190 131us 47583
 24.06.13 08:12.28      6.428 render done 19351us

kern first draw: 
24.06.13 08:11.28      5.533 flush done  5:38 -> 315:190 94us 47583
 24.06.13 08:11.28      5.533 render done 238175us

Kern pairing data should still probably be cached in some way also, but even so the kerned version is still running in 2/3 the time it used to be.

MacDragon:
Yes, that helped quite a bit more:

 24.06.13 08:44.30      5.356 flush done  5:38 -> 315:190 87us 47583
 24.06.13 08:44.30      5.356 render done 143749us

I think this is pro

MacDragon:
[https://github.com/MacDragon/lvgl/tree/%F0%9F%90%BB](url) Has what I've currentlly been trialling to look at.

It should disable kerning automatically if no tables are found, _ex has option to disable on request - which speeds it up a lot if don't need it. Otherwise I'm sure a better way of cacheing the kern pairs could still be found, it didn't seem to speed up a lot when I changed the cache size so I'm not sure if it's working right or not for sure.

kisvegabor:
I've tested your changes on PC and it `kern caching` really makes a difference! I can test it on embedded hardware on Monday. Until that @W-Mai could you review the changes of @MacDragon and if they look good integrate into this PR?

MacDragon:
Taken a little more look, and I'm not sure if the actual kerning lookups are working properly.. so I definitely wouldn't pull quite yet, I'm investigating a bit further. - I only seem to be getting 0 returns on copy of helvetica I was just now testing with.

MacDragon:
Yes, it looks like the fonts I've tried support a newer kerning table type than tiny ttf currently supports, hence the lookup time was constant.. I'm adding a check for invalid table type to save pointless lookup, and trying to find an actually supported font to verify and test.

MacDragon:
Ok, ixed it, so the current version actually works - and now disables kerning if incompatible table format.

Will do performance testing a bit later.

MacDragon:
At first glance, performance appears to be excellent now - with actual working kerning cache.

MacDragon:
Another thought occurs, technically the kerning cache could be shared between different instances of same font to save memory, not sure how tricky that would be to implement.

kisvegabor:
I've just tested @MacDragon's version.

**Ubuntu font** (Kerning is not supported):
- Kerning: None
  - Cache size: 1
     - Render time: 195ms
   - Cache size: 256
      - Render time: 39ms
- Kerning: Normal
  - No change


**Arial font** (Kerning is supported)
  - No kerning: 40ms
  - Kern cache 0:   50ms
  - Kern cache 1:   87ms
  - Kern cache 256: 52ms
  - Kern cache 4096: 51ms

Kern cache 0 (no cache) is very strange. Why is it faster than cache 1?

MacDragon:
Hmm, that does indeed sound odd - is it actually skipping the kerning lookups on 0 and just always assuming 0?

kisvegabor:
With 0 kern cache size it used `stbtt_GetGlyphKernAdvance` which returned `-186`. 
I've used this font: [Arialn.ttf.zip](https://github.com/user-attachments/files/15904368/Arialn.ttf.zip)

MacDragon:
True, it should be running the command directly if cache lookup fails which with 0 cache it should I suppose.. I'll see if I can run more few more tests of own for comparison, it certainly sped up my original test rendering when I added the kern cache vs only glyph desc lookups.

MacDragon:
Finally got some time to look at this again, and I can confirm your odd looking results appear to replicate here, will investigate further as something obviously not going as expected.

    //testfont = lv_tiny_ttf_create_data_ex2(Arialn_ttf, sizeof Arialn_ttf, 16, LV_FONT_KERNING_NORMAL, 1024, 0);
    //                    1.368 render done 14223us
    //                    2.369 render done 7238us

    //testfont = lv_tiny_ttf_create_data_ex2(Arialn_ttf, sizeof Arialn_ttf, 16, LV_FONT_KERNING_NORMAL, 1024, 1);
    //                    1.375 render done 26620us
    //                    2.362 render done 11299us

    //testfont = lv_tiny_ttf_create_data_ex2(Arialn_ttf, sizeof Arialn_ttf, 16, LV_FONT_KERNING_NORMAL, 1024, 256);
    //                    1.360 render done 15643us
    //                    2.360 render done 7695us

first number is first render, second number is second cached render. They appear to bear out your observations that 0 is somehow faster.

MacDragon:
I think it may even be that the actual kerning lookup is slower cached, and most of the speedup was the second glyph cache lookup. Which would actually be good as that would save extra cache - I implemented both parts at same time with assumption it would be faster.

MacDragon:
Yes, it looks like the cache overhead is bigger than the direct lookup speed in normal cases.

I suspect it would be more beneficial in the newer more complex kerning lookups - but it looks like it is indeed best to at least default kern cacheing itself to zero for now, and only use the glyph lookups.

MacDragon:
Tidied away the kern cache for now and cleaned up formatting so it passes verifier:

https://github.com/MacDragon/lvgl/tree/tiny_ttf_cache

kisvegabor:
Awesome! @MacDragon as you version is a superset of this one, could you open a new PR?

kisvegabor:
Let's continue in #6441

PR #6345, opened 2024-06-08
feat(draw/sw): added support for LV_COLOR_FORMAT_I1

faxe1008:

This adds support for the I1 color format which I assume to be a bit pixel format for monochrome displays. The zephyr integration needs this to finally be able to move towards the v9.X version, so I thought I might take a stab at this. Currently missing are all the tests and refrence images for this, if anyone wants to chime in and help me carry this over the finish line I'd be more then happy :^)

### Notes


src/draw/sw/blend/lv_draw_sw_blend_to_rgb565.c
```diff
@@ -358,6 +379,116 @@ void LV_ATTRIBUTE_FAST_MEM lv_draw_sw_blend_image_to_rgb565(_lv_draw_sw_blend_im
...
+                    dest_buf_u16 = drawbuf_next_row(dest_buf_u16, dest_stride);
+                    src_buf_i1 += src_stride;
+                }
+            }
+        }
+        else if(mask_buf == NULL && opa < LV_OPA_MAX) {
+            if(LV_RESULT_INVALID == LV_DRAW_SW_I1_BLEND_NORMAL_TO_RGB565_WITH_OPA(dsc)) {
+                for(y = 0; y < h; y++) {
+                    for(dest_x = 0, src_x = 0; dest_x < w; dest_x++, src_x++) {
+                        dest_buf_u16[dest_x] = lv_color_8_16_mix(src_buf_i1[src_x], dest_buf_u16[dest_x], opa);
```
kisvegabor: Is correct `src_buf_i1[src_x]` correct? I think it should be read bit-by-bit not byte-by-byte.

tests/src/test_cases/draw/test_render_to_i1.c
```diff
@@ -0,0 +1,37 @@
...
+    uint32_t opa;
+    for(opa = 0; opa < 2; opa++) {
+        uint32_t i;
+        for(i = 0; i < _LV_DEMO_RENDER_SCENE_NUM; i++) {
+            lv_demo_render(i, opa_values[opa]);
+
+            char buf[128];
+            lv_snprintf(buf, sizeof(buf), "draw/render/i1/demo_render_%s_opa_%d.png",
+                        lv_demo_render_get_scene_name(i), opa_values[opa]);
+            TEST_ASSERT_EQUAL_SCREENSHOT(buf);
```
kisvegabor, line 33: Please push the new reference images too. They are created automatically as you run the tests locally.

kisvegabor (commented):
Thank you! I was dreaming for long about this features. 

Just to be user I understand correctly: this update supports the case when the bits are mapped horizontally. (Some display controllers supports vertical mapping too, but I hope horizontal is supported in all)

kisvegabor:
Please also add guards similar to https://github.com/lvgl/lvgl/pull/6313.

faxe1008:
> Thank you! I was dreaming for long about this features. 
> 
> Just to be user I understand correctly: this update supports the case when the bits are mapped horizontally. (Some display controllers supports vertical mapping too, but I hope horizontal is supported in all)

Good catch @kisvegabor, does the display driver have any means of keeping track of the vtiled property already?

kisvegabor:
> Good catch @kisvegabor, does the display driver have any means of keeping track of the vtiled property already?

Nope, the display are not aware of this. However, I think having the horizontal mapping already covers most of the use cases. 

I'm not sure if you already use it, but the display's `LV_EVENT_INVALIDATE_AREA` event shall be used to be sure that all the invalidated areas are 8px (byte boundary) aligned.

src/draw/sw/blend/lv_draw_sw_blend_to_rgb565.c
faxe1008: You are correct, I hopefully got all of the I1 -> X blending functions sorted.

kisvegabor:
I've just added the reference images and fixed a few things in the CI. Now the test are passing, however on the reference image it's clear that some calculations are not correct yet. It's quite obvious on [the text tests](https://github.com/lvgl/lvgl/pull/6345/files#diff-32b8876d15938d02c606943726c210d8f14c7fb28ee2e1d12e1ac6ec22057e80).

faxe1008:
Awesome thank you, never would have found this 👍 . The I1 -> X cases look fine, the X -> I1 rendering still has issues.

kisvegabor:
I found the problem and fixed it one case. See
```c
  /*Create a buffer for the canvas*/
    LV_DRAW_BUF_DEFINE_STATIC(draw_buf, CANVAS_WIDTH, CANVAS_HEIGHT, LV_COLOR_FORMAT_I1);
    LV_DRAW_BUF_INIT_STATIC(draw_buf);
    lv_draw_buf_set_palette(&draw_buf, 0, lv_color32_make(0, 0, 0, 0xff));
    lv_draw_buf_set_palette(&draw_buf, 1, lv_color32_make(0xff, 0, 0, 0xff));

    /*Create a canvas and initialize its palette*/
    lv_obj_t * canvas = lv_canvas_create(lv_screen_active());
    lv_canvas_set_draw_buf(canvas, &draw_buf);
    lv_canvas_fill_bg(canvas, lv_color_hex3(0xccc), LV_OPA_COVER);
    lv_obj_center(canvas);

    lv_layer_t layer;
    lv_canvas_init_layer(canvas, &layer);

    lv_draw_label_dsc_t dsc;
    lv_draw_label_dsc_init(&dsc);
    dsc.color = lv_color_white();
    dsc.font = &lv_font_montserrat_18;
    dsc.decor = LV_TEXT_DECOR_UNDERLINE;
    dsc.text = "Hello";

    lv_area_t coords = {10, 10, 130, 60};

    lv_draw_label(&layer, &dsc, &coords);

    lv_canvas_finish_layer(canvas, &layer);
```
image

The problem was that when LVGL tried to move to the first pixel of the destination buffer ([here](https://github.com/lvgl/lvgl/blob/master/src/draw/sw/blend/lv_draw_sw_blend.c#L76)) the "bit information" was lost.
For example when the goal was to jump to x=12 (1.5 byte) it just jumped only 1 byte.

To resolve it I have saved the target area from which we can calculate the bit-information before blending. See my fix in d57f81e

Can you run LVGL's unit test locally? 
```sh
cd tests
./main.py test
```

faxe1008:
Hey @kisvegabor thanks for the help yet again :^)

Tested it locally and the text rendering seems to be a lot better now:
image

There are still some artifacts in the image blending:
image

src/draw/sw/blend/lv_draw_sw_blend.h
```diff
@@ -53,6 +53,7 @@ typedef struct {
     int32_t mask_stride;
     lv_color_t color;
     lv_opa_t opa;
+    lv_area_t realitive_area;
```
C47D: @kisvegabor should be `relative_area` instead?

kisvegabor: Yes :sweat_smile:

kisvegabor:
Cool! 
image
Something is still not okay here. :slightly_frowning_face:

kisvegabor:
I haven't applied the fox on all the places just in one case as a test.

src/draw/sw/blend/lv_draw_sw_blend.h
faxe1008: I'll change it :^)

ceggers-arri:
> Thank you! I was dreaming for long about this features.
> 
> Just to be user I understand correctly: this update supports the case when the bits are mapped horizontally. (Some display controllers supports vertical mapping too, but I hope horizontal is supported in all)

I am currently working on a driver for the [Sitronix](https://www.orientdisplay.com/wp-content/uploads/2022/08/ST7539-G4_V1.1.pdf) monochrome display controller. On this controller, each byte addresses a row of 8 pixels (vertical mapping?).

Without knowing about this pull request, I did the conversion from L8 to monochrome inside my ```push_cb```. A can try to rebase my driver on this branch (and create another pull request) as soon as support for I1 is finished.

kisvegabor:
@ceggers-arri 
It seems you are right
image
> Data is written to the 8 lines of each page directly through D0 to D7. 

image

So we should support column based addressing too. Let's handle it in a separate PR when row based addressing works.

faxe1008:
@kisvegabor cleaned up the commits a bit and will now try to fixup the remaining issues :^)

faxe1008:
@kisvegabor found a few other issues, going into the right direction now. Still some odd things for which I can not find the cause:

For example here the artifacts where the image seems to be doubled?
demo_render_blend_mode_opa_128

Those "random" stripes within the shadows:
demo_render_box_shadow_opa_128

The fact that some image rendering seem to have opacity issues?
demo_render_image_normal_1_opa_255

About your fix and the fact that you said you only adressed one spot for testing purposes - As I did not quite understand the underlying issue I would need some help for this. Does the same thing have to be done with the img_descriptor? Adding this `relative_area` and using it to offset when setting or clearing a bit?

kisvegabor:
Yes, these need to be added on all the other places too. I can do it later this week (probably on Thursday)

kisvegabor:
I added a fix. Color blending related things are working but images are not perfect yet.

faxe1008:
Thanks @kisvegabor, I found a few more issues and fixed them. Now only the L8 rendering for transformer images still has opacity issues.

kisvegabor:
We are getting pretty close :slightly_smiling_face: 

> Now only the L8 rendering for transformed images still has opacity issues.

I see similar issues for the [background of rotated RGB565 on I1](https://github.com/lvgl/lvgl/pull/6345/files#diff-ae893aebdcb30287ce60bd5f26f018a38078a993bd544b49618bc5f6a348e805).

I can take look at it only next week. Will have time earlier to investigate it?

faxe1008:
@kisvegabor I did not yet bother to update the reference images, locally with the current HEAD of this PR I can only see defects for L8. I will do a commit cleanup etc. once everything is done 😀. 

I will probably have some cycles to look at this before next week :^)

kisvegabor:
> I will probably have some cycles to look at this before next week :^)

Fantastic! :heart_eyes:

kisvegabor:
There is a  general Zephyr-LVGL board support related topic that I'd like to discuss. Can you tell your email address or write me at gabor@lvgl.io?

src/draw/sw/blend/lv_draw_sw_blend.h
```diff
@@ -67,6 +68,8 @@ typedef struct {
     lv_color_format_t src_color_format;
     lv_opa_t opa;
     lv_blend_mode_t blend_mode;
+    lv_area_t relative_area;    /**< The blend area relative to the layer's buffer area. */
+    lv_area_t src_area;             /**< The original src area. */
```
faxe1008, line 72: TODO: is this needed?

src/draw/sw/blend/lv_draw_sw_blend_to_argb8888.c
```diff
@@ -334,8 +359,101 @@ void LV_ATTRIBUTE_FAST_MEM lv_draw_sw_blend_image_to_argb8888(_lv_draw_sw_blend_
...
+    int32_t dest_x;
+    int32_t src_x;
+    int32_t y;
+
+    if(dsc->blend_mode == LV_BLEND_MODE_NORMAL) {
+        if(mask_buf == NULL && opa >= LV_OPA_MAX) {
+            if(LV_RESULT_INVALID == LV_DRAW_SW_I1_BLEND_NORMAL_TO_ARGB8888(dsc)) {
+                for(y = 0; y < h; y++) {
+                    for(dest_x = 0, src_x = 0; src_x < w; dest_x++, src_x++) {
+                        uint8_t pixel = src_buf_i1[src_x / 8] & (1 << (7 - (src_x % 8)));
```
faxe1008: TODO: maybe move the get_pixel to some place to use it here as well?

src/draw/sw/blend/lv_draw_sw_blend_to_i1.c
```diff
@@ -0,0 +1,1132 @@
...
+void LV_ATTRIBUTE_FAST_MEM lv_draw_sw_blend_color_to_i1(_lv_draw_sw_blend_fill_dsc_t * dsc)
+{
+    int32_t w = dsc->dest_w;
+    int32_t h = dsc->dest_h;
+    lv_opa_t opa = dsc->opa;
+    const lv_opa_t * mask = dsc->mask_buf;
+    int32_t mask_stride = dsc->mask_stride;
+    int32_t dest_stride = dsc->dest_stride;
+
+    uint8_t color1 = (lv_color_luminance(dsc->color) >= I1_LUM_THRESHOLD) ? 1 : 0;
```
faxe1008: Nit: naming is a bit unfortunate

src/draw/sw/blend/lv_draw_sw_blend_to_i1.c
```diff
@@ -0,0 +1,1132 @@
...
+        }
+    }
+    else {
+        lv_color32_t src_argb;
+        for(y = 0; y < h; y++) {
+            for(dest_x = 0, src_x = 0; src_x < w; dest_x++, src_x++) {
+                src_argb.red = get_bit(src_buf_i1, src_x) * 255;
+                src_argb.green = src_argb.red;
+                src_argb.blue = src_argb.red;
+                if(mask_buf == NULL) {
```
faxe1008: Nit: Inconsitent indent with other occurences of this pattern

src/draw/sw/blend/lv_draw_sw_blend_to_i1.c
```diff
@@ -0,0 +1,1132 @@
...
+    int32_t y;
+
+    int32_t bit_ofs = dsc->relative_area.x1 % 8;
+
+    if(dsc->blend_mode == LV_BLEND_MODE_NORMAL) {
+        if(mask_buf == NULL && opa >= LV_OPA_MAX) {
+            if(LV_RESULT_INVALID == LV_DRAW_SW_L8_BLEND_NORMAL_TO_I1(dsc)) {
+                for(y = 0; y < h; y++) {
+                    for(dest_x = 0, src_x = 0; src_x < w; dest_x++, src_x++) {
+                        uint8_t src_luminance = src_buf_l8[src_x];
```
faxe1008: Nit: remove additional var

src/draw/sw/blend/lv_draw_sw_blend_to_i1.c
```diff
@@ -0,0 +1,1132 @@
...
+                    dest_buf_i1 = drawbuf_next_row(dest_buf_i1, dest_stride);
+                    src_buf_l8 = drawbuf_next_row(src_buf_l8, src_stride);
+                }
+            }
+        }
+        else if(mask_buf == NULL && opa < LV_OPA_MAX) {
+            if(LV_RESULT_INVALID == LV_DRAW_SW_L8_BLEND_NORMAL_TO_I1_WITH_OPA(dsc)) {
+                for(y = 0; y < h; y++) {
+                    for(dest_x = 0, src_x = 0; src_x < w; dest_x++, src_x++) {
+                        uint8_t src_luminance = src_buf_l8[src_x];
```
faxe1008: Nit: remove additional var.

src/draw/sw/blend/lv_draw_sw_blend_to_i1.c
```diff
@@ -0,0 +1,1132 @@
...
+                    src_buf_l8 = drawbuf_next_row(src_buf_l8, src_stride);
+                }
+            }
+        }
+        else if(mask_buf && opa >= LV_OPA_MAX) {
+            if(LV_RESULT_INVALID == LV_DRAW_SW_L8_BLEND_NORMAL_TO_I1_WITH_MASK(dsc)) {
+                for(y = 0; y < h; y++) {
+                    for(dest_x = 0, src_x = 0; src_x < w; dest_x++, src_x++) {
+                        uint8_t src_luminance = src_buf_l8[src_x];
+                        uint8_t mask_val = mask_buf[src_x];
```
faxe1008: Nit: remove additional var.

src/draw/sw/blend/lv_draw_sw_blend_to_i1.c
```diff
@@ -0,0 +1,1132 @@
...
+                    mask_buf += mask_stride;
+                }
+            }
+        }
+        else if(mask_buf && opa < LV_OPA_MAX) {
+            if(LV_RESULT_INVALID == LV_DRAW_SW_L8_BLEND_NORMAL_TO_I1_MIX_MASK_OPA(dsc)) {
+                for(y = 0; y < h; y++) {
+                    for(dest_x = 0, src_x = 0; src_x < w; dest_x++, src_x++) {
+                        uint8_t src_luminance = src_buf_l8[src_x];
+                        uint8_t mask_val = mask_buf[src_x];
```
faxe1008: Nit: remove additional var.

src/draw/sw/blend/lv_draw_sw_blend_to_i1.c
```diff
@@ -0,0 +1,1132 @@
...
+                    mask_buf += mask_stride;
+                }
+            }
+        }
+        else if(mask_buf && opa < LV_OPA_MAX) {
+            if(LV_RESULT_INVALID == LV_DRAW_SW_AL88_BLEND_NORMAL_TO_I1_MIX_MASK_OPA(dsc)) {
+                for(y = 0; y < h; y++) {
+                    for(dest_x = 0, src_x = 0; src_x < w; dest_x++, src_x++) {
+                        uint8_t dest_val = get_bit(dest_buf_i1, dest_x + bit_ofs) * 255;
+                        uint8_t mask_val = mask_buf[src_x];
```
faxe1008: Nit: remove additional var.

src/draw/sw/blend/lv_draw_sw_blend_to_i1.c
```diff
@@ -0,0 +1,1132 @@
...
+    const lv_color32_t * src_buf_c32 = dsc->src_buf;
+    int32_t src_stride = dsc->src_stride;
+    const lv_opa_t * mask_buf = dsc->mask_buf;
+    int32_t mask_stride = dsc->mask_stride;
+
+    int32_t x;
+    int32_t y;
+
+    int32_t bit_ofs = dsc->relative_area.x1 % 8;
+    int32_t src_bit_ofs = 0; //(dsc->relative_area.x1 - dsc->src_area.x1) % 8;
```
faxe1008: TODO: Remove this, as it is 0 anyways.

src/draw/sw/blend/lv_draw_sw_blend_to_i1.c
```diff
@@ -0,0 +1,1132 @@
...
+                    mask_buf += mask_stride;
+                }
+            }
+        }
+    }
+    else {
+        for(y = 0; y < h; y++) {
+            for(x = 0; x < w; x++) {
+                lv_color32_t color_argb = src_buf_c32[x + src_bit_ofs];
+                if(mask_buf == NULL) {
```
faxe1008: Nit: Inconsistent style.

src/draw/sw/blend/lv_draw_sw_blend_to_i1.c
```diff
@@ -0,0 +1,1132 @@
...
+#if LV_DRAW_SW_SUPPORT_RGB888
+static void LV_ATTRIBUTE_FAST_MEM rgb888_image_blend(_lv_draw_sw_blend_image_dsc_t * dsc,
+                                                     const uint8_t src_px_size)
+{
+    int32_t w = dsc->dest_w;
+    int32_t h = dsc->dest_h;
+    lv_opa_t opa = dsc->opa;
+    uint8_t * dest_buf_i1 = dsc->dest_buf;
+    int32_t dest_stride = dsc->dest_stride;
+    const uint8_t * src_buf_u8 = dsc->src_buf;
```
faxe1008: Nit: unfortunate name

src/draw/sw/blend/lv_draw_sw_blend_to_i1.c
```diff
@@ -0,0 +1,1132 @@
...
+                        uint8_t src = lv_color24_luminance(&src_buf_u8[src_x]);
+                        if(src > I1_LUM_THRESHOLD) {
+                            set_bit(dest_buf_i1, dest_x + bit_ofs);
+                        }
+                        else {
+                            clear_bit(dest_buf_i1, dest_x + bit_ofs);
+                        }
+                    }
+                    dest_buf_i1 += dest_stride;
+                    src_buf_u8 += src_stride;
```
faxe1008: Use the draw_buf_next_row instead.

src/draw/sw/blend/lv_draw_sw_blend_to_i1.c
```diff
@@ -0,0 +1,1132 @@
...
+    }
+    else {
+        clear_bit(dest_buf, dest_x);
+    }
+}
+
+static inline void LV_ATTRIBUTE_FAST_MEM lv_color_8_8_mix(const uint8_t src, uint8_t * dest, uint8_t mix)
+{
+
+    if(mix == 0) return;
```
faxe1008, line 1086: TODO: is a branchless version maybe faster?

faxe1008:
Found the last issue with L8, was caused by an error in transparency handling when blending an al88 buffer. Added some comments about stuff that I want to cleanup, otherwise the latest reference images are inside the last commit. I can not spot any more defects @kisvegabor 😀

src/draw/sw/blend/lv_draw_sw_blend_to_i1.c
faxe1008: Nope, not really:
grafik

kisvegabor:
Very nice! :tada: I couldn't spot any issues either. The CI is failing probably because some images are updated. 

---

Just two more idea to make the user's life simpler:
- Document what is happening with I1 rendering. [This section](https://docs.lvgl.io/master/porting/display.html#color-format) is a good place to add the basics, e.g. light colors will `1`, dark will be `0`, bits are mapped horizontally, etc.
- Now we need `LV_EVENT_INVALIDATE_AREA` to round the invalidated area but we could do it internally for I1 [here](https://github.com/lvgl/lvgl/blob/master/src/core/lv_refr.c#L295). 

What do you think?

faxe1008:
I rebased the branch and also reran the update script, but CI is still failing. 
Regarding your other points:
- Documentation: I added a short explanation for the color format. However I left out the part of what an active bit actually means as this is hardware dependant. On traditional EPDs an active bit means black color. I do not know if this would make the explanation more confusing.
- Regarding the rounding callback: I am kinda split on this one because the alignment might be display controller specific. Leaving it to the user is fine imo. Maybe adding a paragraph on this would be neat though. (For refrence this is the Zephyr implementation: https://github.com/zephyrproject-rtos/zephyr/blob/main/modules/lvgl/lvgl_display.c#L49)

ceggers-arri:
> However I left out the part of what an active bit actually means as this is hardware dependant. 

Does this mean that the rendering algorithm can be configured to use '1' for dark or light pixels? My expectation was that the developer of the display driver probably needs to invert all pixels (according to the display hardware). In this case, the driver developer needs to know how to interpret the pixel values provided to ```flush_cb?```.

faxe1008:
@ceggers-arri Currently this is not the case. For convenience sake this could be added as an option :^).

faxe1008:
@ceggers-arri I added a flag in the latest commit `LV_COLOR_FORMAT_I1_INVERT_LUMINANCE` which can be used for this purpose. Now the images can be rendered "inverted":
image

kisvegabor:
Instead of `LV_COLOR_FORMAT_I1_INVERT_LUMINANCE` can we have a function like `lv_draw_sw_i1_invert(buf, buf_size)`? It can just invert `uint32_t` numbers, like `buf_u32[i] = ~buf_u32[i]`;

This way you have have multiple displays and each can be inverted separately.

src/draw/sw/lv_draw_sw.c
```diff
@@ -226,6 +226,15 @@ void lv_draw_sw_rgb565_swap(void * buf, uint32_t buf_size_px)
 
 }
 
+void lv_draw_sw_i1_invert(void * buf, uint32_t buf_size) {
+    if(buf == NULL) return;
+
+    uint8_t *byte_buf = (uint8_t *)buf;
+    for(uint32_t i = 0; i < buf_size; i++) {
```
kisvegabor: Inverting uint32_t could be faster, but I guess it's okay for small monochrome screens

kisvegabor (dismissed):
Thank you! It looks good to me! 

It's a pity that I could try it out only with a canvas as SDL doesn't support I1.

kisvegabor (dismissed):
Fixed code formatting

src/draw/sw/lv_draw_sw.c
faxe1008: True, did a few benchmarks:
https://quick-bench.com/q/ovlk-VymfR3x-TIdjkPhh6IiRNo
There is a slight perfomance to be gained here :^)

src/draw/sw/lv_draw_sw.c
```diff
@@ -226,6 +226,26 @@ void lv_draw_sw_rgb565_swap(void * buf, uint32_t buf_size_px)
 
 }
 
+void lv_draw_sw_i1_invert(void * buf, uint32_t buf_size)
+{
+    if(buf == NULL) return;
+
+    uint32_t * word_buf = (uint32_t *)buf;
```
ceggers-arri: As ```buf``` is an array of `uint8_t`, I think that there is no alignment guarantee.
Performing 16 or 32 bit wide operations on non aligned addresses does not
work reliable on non x86 platforms.

You may use the implementation of [memset](https://github.com/eblot/newlib/blob/master/newlib/libc/string/memset.c) as an example how this could be implemented.

faxe1008: True, did not consider this. Added a version inspired by memset :^)

kisvegabor:
I've just added a test for invert too.

Once you have removed the `TOO_SMALL` macro, the other could be easily removed too. So there no macros anymore.

kisvegabor:
@FASTSHIFT I've removed the VG-Lite reference images for A8, AL88 and I1 renerings, and also skipped these tests with VG-Lite. See [here](https://github.com/lvgl/lvgl/pull/6345/files#diff-a996ce8bbf3b69f490b811d285e9984e68282765ef3c3ca413446ac5fa3b937f).

kisvegabor (approved):
Finally passing :slightly_smiling_face:

faxe1008:
The history has gotten a bit messy, should I clean this up?

FASTSHIFT (approved):

kisvegabor:
Merged! :tada: 

>  The history has gotten a bit messy, should I clean this up?

Not required, usually we squash PRs.

PR #6349, opened 2024-06-09
feat(libs): added JPEGDEC optimized codec

bitbank2:

This pull request adds an optimized (faster speed and smaller memory requirements) codec for JPEG decoding. This codec has been in wide use for several years on Arduino, but are written in platform-independent C code. The main advantages are that it can decode faster and deliver the designed pixel format without needing intermediate conversion steps. No additional tests or documentation is needed since they behave the same as the codecs they replace (as far as user code is concerned).

### DIFF
--- a/src/drivers/nuttx/lv_nuttx_image_cache.c
+++ b/src/drivers/nuttx/lv_nuttx_image_cache.c
@@ -136,6 +136,19 @@ static bool defer_init(void)
     return true;
 }
 
+static void heap_memdump(struct mm_heap_s * heap)
+{
+    struct mm_memdump_s dump = {
+        PID_MM_ALLOC,
+#if CONFIG_MM_BACKTRACE >= 0
+        0,
+        ULONG_MAX
+#endif
+    };
+
+    mm_memdump(heap, &dump);
+}
+
 static void * malloc_cb(size_t size_bytes, lv_color_format_t color_format)
 {
     LV_UNUSED(color_format);
@@ -163,6 +176,7 @@ static void * malloc_cb(size_t size_bytes, lv_color_format_t color_format)
         bool evict_res = lv_cache_evict_one(img_cache_p, NULL);
         if(evict_res == false) {
             LV_LOG_ERROR("failed to evict one cache entry");
+            heap_memdump(ctx->heap);
             return NULL;
         }
     }


src/libs/jpegdec/jpegdec.h
```diff
@@ -0,0 +1,308 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//===========================================================================
+//
+#ifndef __JPEGDEC__
+#define __JPEGDEC__
+// fixed to Linux for LVGL
+#define __LINUX__
+#if defined( __MACH__ ) || defined( __LINUX__ ) || defined( __MCUXPRESSO ) || defined( ESP_PLATFORM )
+    #include <stdlib.h>
```
kisvegabor: Why do we need `stdlib.h`?

src/lv_init.c
```diff
@@ -285,6 +286,10 @@ void lv_init(void)
     lv_fs_arduino_sd_init();
 #endif
 
+#if LV_USE_JPEGDEC
+    lv_jpegdec_init();
```
PGNetHun: Please add also `lv_jpegdec_deinit()` in the `deinit()` function

PGNetHun (changes requested):

src/lv_init.c
bitbank2: I can add that @PGNetHun , but none of the other imaging codecs call deinit(). Why is that?

kisvegabor:
I've fixed a few warning in the PR and tested it on my Linux notebook but got these errors in the log:
```
[Warn]	(0.098, +98)	 decode_jpeg_data: JPEG_openRAM failed
 lv_jpegdec.c:293
[Warn]	(0.098, +0)	 decoder_open: Error decoding JPEG
 lv_jpegdec.c:230
[Error]	(0.098, +0)	 _lv_draw_image_normal_helper: Failed to open image lv_draw_image.c:161
```
and no image was displayed.

The test code is:
```c
  lv_obj_t * wp = lv_image_create(lv_screen_active());
  lv_image_set_src(wp, "A:lvgl/examples/libs/tjpgd/img_lvgl_logo.jpg");
  lv_obj_center(wp);
```

src/libs/jpegdec/lv_jpegdec.c
```diff
@@ -0,0 +1,342 @@
...
+    /*If it's a JPEG file...*/
+    if(src_type == LV_IMAGE_SRC_FILE) {
+        const char * fn = src;
+        /* Read the width and height from the file */
+        res = lv_fs_open(&f, fn, LV_FS_MODE_RD);
+        lv_fs_seek(&f, 0, LV_FS_SEEK_END);
+        lv_fs_tell(&f, &rn);
+        lv_fs_seek(&f, 0, LV_FS_SEEK_SET);
+        iFileSize = (int)rn;
+        if(res != LV_FS_RES_OK) return LV_RESULT_INVALID;
```
kisvegabor: Please move this after `lv_fs_open` and add a log entry. E.g. `LV_LOG_WARN("Can't open %s, result: %d", fn, res);`

bitbank2: Updated with both suggested changes

kisvegabor:
I've tested the latest changes and with the [code above](https://github.com/lvgl/lvgl/pull/6349#issuecomment-2189738206) and now see the output, however it's still scrambled a little:
image

bitbank2:
> I've tested the latest changes and with the [code above](https://github.com/lvgl/lvgl/pull/6349#issuecomment-2189738206) and now see the output, however it's still scrambled a little: image

thanks for letting me know; I'll fix it.

bitbank2:
fixed

src/libs/jpegdec/jpegdec.c
```diff
@@ -4351,30 +4357,30 @@ static int DecodeJPEG(JPEGIMAGE * pJPEG)
...
+            cy = (pJPEG->iCropY + pJPEG->iCropCY + 7) >> 3;
             iCr = MCU2;
             iCb = MCU3;
             mcuCX = 16;
             mcuCY = 8;
             break;
         case 0x22:
             cx = (pJPEG->iWidth + 15) >> 4;  // number of MCU blocks
-            cy = (pJPEG->iCropY + pJPEG->iCropCY) >> 4;
+            cy = (pJPEG->iCropY + pJPEG->iCropCY + 15) >> 4;
```
C47D: Can we add a comment or replace the magic numbers with a symbol?

bitbank2:
Removed Desktop CPU SIMD code

kisvegabor:
I've tested it on embedded hardware and it worked well. This the benchmark with 18 LVGL logo placed next to each other and continuously invalidating the screen:

| Name              | Max memory usage | Render time |
|-------------------|------------------|-------------|
| Empty screen      | 7kB              | 7ms         |
| TJPGD no cache   | 13.7kB           | 39ms        |
| TJPGD cache      | 13.7kB           | 39ms        |
| JPEGDEC no cache  | 36.9kB           | 26ms        |
| JPEGDEC cache     | 36.9kB           | 13ms        |

So JPEGDEC is faster but uses more memory. Probably it's because TJEGD decoded the image title by tile, while JPGDEC decodes the whole image.  It also explains why the cache is has no affect with TJPGDEC.

bitbank2:
> I've tested it on embedded hardware and it worked well. This the benchmark with 18 LVGL logo placed next to each other and continuously invalidating the screen:
> 
> Name	Max memory usage	Render time
> Empty screen	7kB	7ms
> TJPGD no cache	13.7kB	39ms
> TJPGD cache	13.7kB	39ms
> JPEGDEC no cache	36.9kB	26ms
> JPEGDEC cache	36.9kB	13ms
> So JPEGDEC is faster but uses more memory. Probably it's because TJEGD decoded the image title by tile, while JPGDEC decodes the whole image. It also explains why the cache is has no affect with TJPGDEC.

I can immediately remove 4K of that by getting rid of the output pixel buffer (not needed when decoding the frame to a full frame buffer). The rest is needed because of my fast Huffman decode tables.

kisvegabor:
Could you rebase on `master`? In the meantime our logo was updated. :slightly_smiling_face: 

> I can immediately remove 4K of that by getting rid of the output pixel buffer (not needed when decoding the frame to a full frame buffer). The rest is needed because of my fast Huffman decode tables.

I thought that it's decompressing the whole image as there is no `get_area_cb` in the decoder.

bitbank2:
It is decompressing the whole image. I think I can remove the temporary output buffer I used for block by block decoding. I can also try to incorporate the optimized cropping feature I recently added to speed up partial decodes.

kisvegabor:
If there are any updates, could you push them?

bitbank2:
pushed

kisvegabor:
Hi Larry,

I've tested with this simple code, but the result is not so good.
```c
LV_IMAGE_DECLARE(test_img_lvgl_logo_jpg);

lv_obj_set_flex_flow(lv_screen_active(), LV_FLEX_FLOW_ROW_WRAP);

for(int i = 0; i < 10; i++) {
    wp = lv_image_create(lv_screen_active());
    /* Assuming a File system is attached to letter 'A'
     * E.g. set LV_USE_FS_STDIO 'A' in lv_conf.h */
    lv_image_set_src(wp, &test_img_lvgl_logo_jpg);
    lv_obj_center(wp);
}
```


The rendering time varies highly. Something it's above 1 sec. Does it work well for you on an embedded hardware?

kisvegabor:
Hi Larry, Any updates on this?

bitbank2:
Working on it today -

bitbank2:
@kisvegabor The loop you created is reading the source file each pass through. This can kill the performance on almost any system. Is there a way to avoid this activity when using the same image file more than once?

kisvegabor:
The file should be cached so creating it 10 times shouldn't be an issue. An it was fast with TJPGD.

PR #6362, opened 2024-06-12
fix(label): do not break last line for LV_LABEL_LONG_DOT (#5606)

### LINKED ISSUE #5606: Bug of label with DOT mode
### LVGL version

V8.2

### What happened?

When the content of label is conbined with chinese char and number, there will be a problem, only showing “…”. As the following picture shown

image

### How to reproduce?

#!/opt/bin/lv_micropython -i

import lvgl as lv
import display_driver
import fs_driver

fs_drv = lv.fs_drv_t()
fs_driver.fs_register(fs_drv, 'S')
font_simsun_16_cjk = lv.font_load("S:../../assets/font/lv_font_simsun_16_cjk.fnt")

label1 = lv.label(lv.scr_act())
label = label1
label.set_style_text_font(font_simsun_16_cjk, 0)
label.set_long_mode(lv.label.LONG.DOT)     
label.set_text("机123456")
label.set_size(70, 20)                
label.align(lv.ALIGN.TOP_MID, 0, 20)
label.set_style_bg_color(lv.color_hex(0xFF00FF), 0)
label.set_style_bg_opa(100, 0)

label2 = lv.label(lv.scr_act())
label = label2
label.set_style_text_font(font_simsun_16_cjk, 0)
label.set_long_mode(lv.label.LONG.DOT)     
label.set_text("机1234567")
label.set_size(70, 20)                
label.align(lv.ALIGN.TOP_MID, 0, 60)
label.set_style_bg_color(lv.color_hex(0x00FFFF), 0)
label.set_style_bg_opa(100, 0)

niklasf:
Fixes #5606 by implementing and using a new flag `LV_TEXT_FLAG_BREAK_ALL` (like [`word-break: break-all` in CSS](https://developer.mozilla.org/en-US/docs/Web/CSS/word-break)). The private `lv_text_get_next_word()` already had a `force` parameter, which is replaced by the new public flag.

Testing: Covered by existing reference images. Note how the last line is no longer cut off too early.


kisvegabor:
Looks good! Thank you for resolving this long lasting issue. :+1: 

Just one note: could you set some background color for the labels in the test to better see where the boundaries are?

niklasf:
Sure. Background added.

kisvegabor (approved):
Thank you!

niklasf:
Not stale. Waiting for merge.

kisvegabor:
@liamHowatt @FASTSHIFT @XuNeo Could you review this PR?

liamHowatt (approved):
Nice!

W-Mai:
image

Is it reasonable that one word is missing in this place?

XuNeo (approved):

XuNeo (changes requested):
The original rule is to ensure that a whole word can be displayed. If `...` cannot fit, clip the previous word. Looks like this PR changes the rule.

kisvegabor:
> Is it reasonable that one word is missing in this place?

> The original rule is to ensure that a whole word can be displayed. If ... cannot fit, clip the previous word. Looks like this PR changes the rule.

Actually this is the issue being foxed by this PR. Example (clip to 2 lines):

Original:
```
s asd asd adsasd  |
asda              |
01234567890123    |
```

Previously:
```
s asd asd adsasd  |
a...              |
```

Fixed:
```
s asd asd adsasd  |
asda 012345678... |
```

XuNeo (approved):
I may have made the bad experiments. Could @W-Mai  help to check again the original behavior of what's wrong?

liamHowatt (approved):
The correct appearance of kisvegabor's example would look like this according XuNeo's reasoning:
```
s asd asd adsasd  |
asda...           |
```

W-Mai:
> The correct appearance of kisvegabor's example would look like this according XuNeo's reasoning:
> 
> ```
> s asd asd adsasd  |
> asda...           |
> ```

This is the correct one!

image

kisvegabor:
> This is the correct one!

This is the bug we are about to fix. IMO the expected behavior is to fill the whole line with characters.

W-Mai:
> fill the whole line with characters.

Of course, it is a better-looking choice and a simpler way of writing, but we can see that the default behavior in html is like that.

kisvegabor:
> Of course, it is a better-looking choice and a simpler way of writing, but we can see that the default behavior in html is like that.

Maybe it's buggy too. 😀
Anyway, even if it diverges from HTML, I think the users expect this (fixed) behaviour.

PR #6418, opened 2024-06-24
chore(lvgl_private): remove unnecessary private header file includes

FASTSHIFT:

Hi, we upgraded LVGL on our system. Our CI encountered the problem that the upper-level application based on LVGL reported that `ft2build.h` (freetype compilation configuration file) could not be found. 

After investigation, we found that `lv_freetype_private.h` was also exposed to the public space in `lvgl_private.h` (see this PR: https://github.com/lvgl/lvgl/pull/5913). The original design was that `lv_freetype_private.h` was only used internally in freetype porting layer, so only the directory where `ft2build.h` was located was exposed to LVGL. If `lv_freetype_private.h` was exposed to the public space, the compilation system would need to open the `ft2build.h` directory to all applications that use `lvgl_private.h` (and these programs originally only wanted to access display and indev), which is very likely to cause naming pollution and compilation conflicts!!!

I think `lvgl_private.h` should only contain necessary modules, rather than adding all `*_private.h` files for the sake of convenience. For new special requirements, we should package them into carefully designed public APIs, and eventually guide applications to gradually move away from `lvgl_private.h`.

### Notes

### DIFF
--- a/src/lvgl_private.h
+++ b/src/lvgl_private.h
@@ -16,6 +16,8 @@ extern "C" {
 #include "display/lv_display_private.h"
 #include "indev/lv_indev_private.h"
 #include "misc/lv_text_private.h"
+#include "misc/cache/lv_cache_entry_private.h"
+#include "misc/cache/lv_cache_private.h"
 
 /*********************
  *      DEFINES
--- a/src/lvgl_private.h
+++ b/src/lvgl_private.h
@@ -16,10 +16,6 @@ extern "C" {
 #include "display/lv_display_private.h"
 #include "indev/lv_indev_private.h"
 #include "misc/lv_text_private.h"
-#include "libs/freetype/lv_freetype_private.h"
-#include "misc/cache/lv_cache_entry_private.h"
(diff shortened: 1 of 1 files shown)


FASTSHIFT:
> We are preparing to take the private convention much further in #6068.
> 
> I can appreciate removing Free type and Windows, but cache too?
> 
> CC @kisvegabor

I suggest deleting it first, and then adding it if the user really needs it. 
cc @W-Mai

kisvegabor:
I see the problem, and agree to remove the private FreeType and Windows header.  However, the cache is an internal LVGL thing. If the user wants all the LVGL stuff they can just include `lvgl_private.h`.

FASTSHIFT:
> I see the problem, and agree to remove the private FreeType and Windows header. However, the cache is an internal LVGL thing. If the user wants all the LVGL stuff they can just include `lvgl_private.h`.

OK, let's keep the cache private API.

liamHowatt (approved):

XuNeo (approved):

PR #6427, opened 2024-06-26
feat(draw_buf): make draw buf API more OOP style

XuNeo:

This is important because once we require to use correct APIs for different kinds of draw buffers, it can be easily misused.
Specially for create/destroy API, misuse it can lead to memory leak if different heap is used.

Add a handler pointer to lv_draw_buf_t so later operations for it can use same set of draw buf APIs. 4bytes compared to image data should be acceptable.

Rename the `_user` suffix to `_ex`, which means an extended sets of API.

### Notes

### DIFF
--- a/src/draw/lv_draw_buf.c
+++ b/src/draw/lv_draw_buf.c
@@ -85,11 +85,11 @@ lv_draw_buf_handlers_t * lv_draw_buf_get_handlers(void)
 
 uint32_t lv_draw_buf_width_to_stride(uint32_t w, lv_color_format_t color_format)
 {
-    return lv_draw_buf_width_to_stride_user(&default_handlers, w, color_format);
+    return lv_draw_buf_width_to_stride_ex(&default_handlers, w, color_format);
 }
 
-uint32_t lv_draw_buf_width_to_stride_user(const lv_draw_buf_handlers_t * handlers, uint32_t w,
-                                          lv_color_format_t color_format)
+uint32_t lv_draw_buf_width_to_stride_ex(const lv_draw_buf_handlers_t * handlers, uint32_t w,
+                                        lv_color_format_t color_format)
 {
     if(handlers->width_to_stride_cb) return handlers->width_to_stride_cb(w, color_format);
     else return 0;
@@ -97,10 +97,10 @@ uint32_t lv_draw_buf_width_to_stride_user(const lv_draw_buf_handlers_t * handler
 
 void * lv_draw_buf_align(void * data, lv_color_format_t color_format)
 {
-    return lv_draw_buf_align_user(&default_handlers, data, color_format);
+    return lv_draw_buf_align_ex(&default_handlers, data, color_format);
 }
 
-void * lv_draw_buf_align_user(const lv_draw_buf_handlers_t * handlers, void * data, lv_color_format_t color_format)
+void * lv_draw_buf_align_ex(const lv_draw_buf_handlers_t * handlers, void * data, lv_color_format_t color_format)
 {
     if(handlers->align_pointer_cb) return handlers->align_pointer_cb(data, color_format);
     else return NULL;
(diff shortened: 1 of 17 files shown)


src/draw/lv_draw_buf.c
```diff
@@ -84,37 +84,32 @@ lv_draw_buf_handlers_t * lv_draw_buf_get_handlers(void)
 
 uint32_t lv_draw_buf_width_to_stride(uint32_t w, lv_color_format_t color_format)
 {
-    return lv_draw_buf_width_to_stride_user(&default_handlers, w, color_format);
+    return lv_draw_buf_width_to_stride_ex(&default_handlers, w, color_format);
```
W-Mai, line 88: renaming stuff I think we can open a new pr to process
here just change the code logic

XuNeo: I have separated the changes to two commits.

src/draw/lv_draw_buf.c
```diff
@@ -19,6 +19,8 @@
 #define font_draw_buf_handlers LV_GLOBAL_DEFAULT()->font_draw_buf_handlers
 #define image_cache_draw_buf_handlers LV_GLOBAL_DEFAULT()->image_cache_draw_buf_handlers
 
+#define draw_buf_get_handler(drawbuf) ((drawbuf)->handlers ? (drawbuf)->handlers : &default_handlers)
```
liamHowatt: Seems like `->handlers` can never be `NULL`.

liamHowatt (commented):
Nice

src/draw/lv_draw_buf.c
kisvegabor: IMO if it's `NULL` it's rather an error that shouldn't be silently replaced by the defaults.

kisvegabor (commented):
Good idea!

src/draw/lv_draw_buf.c
XuNeo: There's hacky code assembles `lv_draw_buf_t` manually and could have NULL handlers. I'll fix them and add assert directly.

XuNeo:
The failing CI cannot be reproduced locally. Need some time to check whats' wrong.

XuNeo:
Turns out the issue is from `lv_draw_buf_from_image`. VG-Lite CI can correctly detect this difference with premultiplied images that generated by ThorVG.

```patch
 /**
  * Initialize a draw buf with the given buffer and parameters. Clear draw buffer flag to zero.
@@ -336,6 +336,7 @@ static inline void lv_draw_buf_from_image(lv_draw_buf_t * buf, const lv_image_ds
 {
     lv_draw_buf_init(buf, img->header.w, img->header.h, img->header.cf, img->header.stride,
                      (void *)img->data, img->data_size);
+    buf->header.flags = img->header.flags;
 }
```

kisvegabor:
I've re-based this PR on master to see if the MicroPython CI issue is resolved.

kisvegabor:
The MicroPython CI issue seems to be fixed! :tada:

kisvegabor (dismissed):

tests/main.py
```diff
@@ -29,9 +29,9 @@
     build_only_options['OPTIONS_SDL'] = 'SDL simulator with full config, 32 bit color depth'
 
 test_options = {
+    'OPTIONS_TEST_VG_LITE': 'VG-Lite simulator with full config, 32 bit color depth',
```
FASTSHIFT: No need change this.

src/draw/lv_draw_buf.c
```diff
@@ -258,22 +250,23 @@ lv_result_t lv_draw_buf_init(lv_draw_buf_t * draw_buf, uint32_t w, uint32_t h, l
     header->flags = 0;
     header->magic = LV_IMAGE_HEADER_MAGIC;
 
-    draw_buf->data = lv_draw_buf_align(data, cf);
```
FASTSHIFT, line 266: why change this?

XuNeo: lv_draw_buf_init is normally used in `lv_bin_decoder` to convert a const image array to draw_buf struct. 

Instead of align the data secretly, which could lead to global buffer out of range access and wrong draw result if data is not aligned already, I think we should add warning below instead and keep the data unchanged.

The log message could be ERROR, or we can add assert here. What do you think?

FASTSHIFT: @XuNeo 
I think it is enough to print the error log and let the bin decoder do the copy + alignment.

kisvegabor:
According to the CI there are some out of bounds issues.

src/draw/lv_draw_buf.c
XuNeo: I have changed it back. Current image post process won't check address alignement and I don't think it should.
If the C-array image data start address is not aligned, error is reported and user should adjust the image source.

kisvegabor (dismissed):

src/draw/lv_draw_buf.h
```diff
@@ -91,6 +85,14 @@ typedef void (*lv_draw_buf_cache_operation_cb)(const lv_draw_buf_t * draw_buf, c
 
 typedef uint32_t (*lv_draw_buf_width_to_stride_cb)(uint32_t w, lv_color_format_t color_format);
 
+struct _lv_draw_buf_t {
```
liamHowatt: We are trying to avoid `_lv_*` now

XuNeo: You are right. Should I change it to `struct lv_draw_buf_t`?

liamHowatt (approved):

src/draw/lv_draw_buf.h
liamHowatt: Looks good now, thanks!

kisvegabor (approved):

Issue #5379, opened 2024-01-18
`lv_array` auto resize capability


XuNeo:
### Introduce the problem

It was raised in https://github.com/lvgl/lvgl/pull/5300.
For adding elements to array, it's clearly now that `push_back` will automatically resize as need. Same for `array_concat`.

Now we need to decide how to shrink the array. We can define a threshold, so after removing elements the array should shrink when empty space is larger than threshold .

Let's discuss it here.

cc @PGNetHun

kisvegabor:
What functions do we have remove elements from the array? I've found and `lv_array_remove`, `lv_array_erase`. Are there anything else?

PGNetHun:
Well, I think those code lines needs to be checked that call these functions: whether the callers repeatedly remove and add new items, or just remove occasionally.
I doubt that in current LVGL codebase there are many calls to remove and then adding items to same array continously.

XuNeo:
Not stale.

kisvegabor:
Not stale

XuNeo:
I'm thinking of adding SIZE_POLICY_DEFAULT, and SIZE_POLICY_DOUBLE. Not in hurry to implement though.

PGNetHun:
> > I doubt that in current LVGL codebase there are many calls (use cases) to remove and then adding items to same array continously.
> 
> I agree with @PGNetHun. Probably it doesn't worth the extra config option as (at least I believe) we don't add items to the arrays on performance critical places.

Great, then we could close this ticket.
I think there are other places that could be optimized, not this one.

PR #6431, opened 2024-06-26
perf(array): optimize array push back function performance

onecoolx:

### Notes

### DIFF
--- a/src/misc/lv_array.h
+++ b/src/misc/lv_array.h
@@ -18,7 +18,7 @@ extern "C" {
 /*********************
  *      DEFINES
  *********************/
-#define LV_ARRAY_DEFAULT_CAPACITY 8
+#define LV_ARRAY_DEFAULT_CAPACITY 4
 
 /**********************
  *      TYPEDEFS
--- a/src/misc/lv_array.h
+++ b/src/misc/lv_array.h
@@ -18,7 +18,10 @@ extern "C" {
 /*********************
  *      DEFINES
  *********************/
-#define LV_ARRAY_DEFAULT_CAPACITY 4
+
+#ifndef LV_ARRAY_DEFAULT_CAPACITY
+#define LV_ARRAY_DEFAULT_CAPACITY  4
+#endif
 
 /**********************
  *      TYPEDEFS
--- a/src/misc/lv_array.c
+++ b/src/misc/lv_array.c
@@ -143,7 +143,7 @@ lv_result_t lv_array_push_back(lv_array_t * array, const void * element)
 
     if(array->size == array->capacity) {
         /*array is full*/
-        lv_array_resize(array, array->capacity + 1);
+        lv_array_resize(array, array->capacity + LV_ARRAY_DEFAULT_CAPACITY);
     }
 
     uint8_t * data = array->data + array->size * array->element_size;
--- a/tests/src/test_cases/test_array.c
+++ b/tests/src/test_cases/test_array.c
@@ -40,7 +40,7 @@ void test_array_size(void)
         lv_array_push_back(&array, &i);
     }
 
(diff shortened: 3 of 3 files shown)


FASTSHIFT (dismissed):

onecoolx:
Decrease the number of memory allocations when array push back.
cc @XuNeo @FASTSHIFT

kisvegabor (dismissed):

kisvegabor:
I'm not sure if it's a good trade off. When a single event is added to a widget 4x8=32 bytes will be allocated instead of 4.

onecoolx:
> Previously, we have discussed to add array runtime memory expanding options such as double the memory or linear expanding. But agreed in the end this is not necessary because array is not used in performance critical path.
> 
> #5379

This is not true. When you add points to a path in the vector api, resizing the memory each time will cause severe performance degradation and memory fragmentation. Strict memory growth is not the purpose of designing array container.

onecoolx:
> I'm not sure if it's a good trade off. When a single event is added to a widget 4x8=32 bytes will be allocated instead of 4.

I think we can find a balance between memory growth and performance, that's why it's not growing by two times or more, but by a limited growth; this gives better performance than resizing every time you add an element, while the memory growth is manageable.

kisvegabor:
I consider memory fragmentation a bigger issue than the performance. 


The 100 widgets need ~20kB.  

In case of the events, if we have these 100 widgets with 1 event, with the current method we will use 4 x 7 x 100 = 2.8kB extra RAM. (+14%) 
If we change `LV_ARRAY_DEFAULT_CAPACITY` to `4` it's only 1.2kB. (+6%) It seems like a good trade off to me.

onecoolx:
> I consider memory fragmentation a bigger issue than the performance.
> 
> The 100 widgets need ~20kB.
> 
> In case of the events, if we have these 100 widgets with 1 event, with the current method we will use 4 x 7 x 100 = 2.8kB extra RAM. (+14%) If we change `LV_ARRAY_DEFAULT_CAPACITY` to `4` it's only 1.2kB. (+6%) It seems like a good trade off to me.

We could set this value become a build config , but as an array, when you add some values, you will most likely add more again.   For example  bezier curve need three points one time， when you create a complex path, memory will be realloc many times.

kisvegabor:
Changing `LV_ARRAY_DEFAULT_CAPACITY` to `4`. I think it wouldn't be a good config option as it has too small impact. We should be able to find a good fixed value instead.

FASTSHIFT:
We can set `LV_ARRAY_DEFAULT_CAPACITY` to 1 by default, which is more inclined to save memory. If the user has enough memory and wants to pursue speed, the user can increase the value of `LV_ARRAY_DEFAULT_CAPACITY` by himself

kisvegabor:
I think 4 is a good default value a no need for bloating `lv_conf.h` even further. Although we can define it like
```
#ifndef LV_ARRAY_DEFAULT_CAPACITY
#define LV_ARRAY_DEFAULT_CAPACITY  4
#endif
```

So if someone really needs this level of fine tuning, it's still possible.

src/misc/lv_array.h
```diff
@@ -18,7 +18,10 @@ extern "C" {
 /*********************
  *      DEFINES
  *********************/
-#define LV_ARRAY_DEFAULT_CAPACITY 8
+
+#ifndef LV_ARRAY_DEFAULT_CAPACITY
+#define LV_ARRAY_DEFAULT_CAPACITY  4
```
FASTSHIFT, line 23: Can we move this into `lv_conf`? That way it's configurable.

FASTSHIFT (approved):

kisvegabor (approved):

Issue #6340, opened 2024-06-07
Tiny TTF lib: lv_tiny_ttf_set_size() should be removed or clear the cache


codewitch-honey-crisis:
### LVGL version

9.1

### What happened?

If you set the size of the font using the `lv_tiny_ttf_set_size(...)` routine it does not clear the glyph cache. This could lead to stale caches of the bitmaps at the wrong size.

It should either be removed, or it should clear the cache

kisvegabor:
Thanks! Mentioned in #6333

W-Mai:
Let me open a new pr to remove this

MacDragon:
I added a grlyphe/bitmap cache delete and recreate on resize in my cacheing modifications, though I didn't get around to testing it - kern caching not removed as this is not size dependant.

codewitch-honey-crisis:
I made a little performance tester for a TTGO T1 Display since they're so common. It works under Arduino or the ESP-IDF
To use it, copy the root of the latest branch into libs/ and upload w/ Platform IO
https://github.com/codewitch-honey-crisis/tiny_ttf_improvements

Results - 6/1 build - 187 iterations per second under Arduino! (i think the caching was flawed). 18 under the ESP-IDF
Results 6/14 build 30 iterations per second under Arduino, 30 under the ESP-IDF.

So discounting the 187 iterations result, it looks like an improvement from 18 to 30, cautiously.

I'd be happy to test the latest branch whenever it's committed.

MacDragon:
The forked branch linked on my GH in the main cacheing thread ( https://github.com/MacDragon/lvgl/tree/🐻 ) has where I left it at currently, and appeared to be working pretty well for me - comparing kerning on and off times would be useful too, assuming working with a font that has kerning tables it can actually use. the font kerning flag gets turned off if unusable table found.

codewitch-honey-crisis:
I'm getting weird results with your forked branch. Once again Arduino is about 10 times as fast as the IDF. I'm no longer going to discount that.

Arduino - 1010 iterations per second. ESP-IDF 104 iterations per second.

Either way, that's a significant improvement but I'm confused about the results.

MacDragon:
A difference that big I'd have thought could be true hw floating point vs software or something?

codewitch-honey-crisis:
[...]

PR #6441, opened 2024-06-28
refactor(tiny_fft): refactor tiny_ttf cache with updated improvements

MacDragon:

Tiny_ttf has been reconstructed:

cache of uincode(letter) -> glyph_dsc @codewitch-honey-crisis @MacDragon
lv_tiny_ttf_create_xxxx_ex really affect the cache size @cybekRT
all tiny_ttf use a same cache -> each font use separate caches
kerning support is automatically checked and disabled, or disableable via ex create.
Kerning lookups now use glyph desc cache to find glyph id
greatly speeds up the usage of tiny_ttf text.

https://github.com/lvgl/lvgl/issues/4835#issuecomment-2134459861
https://github.com/lvgl/lvgl/issues/6305
https://github.com/lvgl/lvgl/issues/6271
https://github.com/lvgl/lvgl/issues/6340
cc @kisvegabor

### Notes

### DIFF
--- a/docs/libs/tiny_ttf.rst
+++ b/docs/libs/tiny_ttf.rst
@@ -29,11 +29,14 @@ entire time the font is being used.
 After a font is created, you can change the font size in pixels by using
 :cpp:expr:`lv_tiny_ttf_set_size(font, font_size)`.
 
-By default, a font will use up to 4KB of cache to speed up rendering
-glyphs. This maximum can be changed by using
-:cpp:expr:`lv_tiny_ttf_create_data_ex(data, data_size, font_size, cache_size)`
-or :cpp:expr:`lv_tiny_ttf_create_file_ex(path, font_size, cache_size)` (when
-available). The cache size is indicated in bytes.
+By default, a font will cache data for upto 256 glyphs elements to speed up rendering.
+This maximum can be changed by using
+:cpp:expr:`lv_tiny_ttf_create_data_ex(data, data_size, font_size, kerning, cache_size, )`
+or :cpp:expr:`lv_tiny_ttf_create_file_ex(path, font_size, kerning, cache_size)` (when
+available). The cache size is indicated in number of entries. Kerning is whether to allow
+if supported, or disable.
+
+
 
 .. _tiny_ttf_example:
 
--- a/src/libs/tiny_ttf/lv_tiny_ttf.c
+++ b/src/libs/tiny_ttf/lv_tiny_ttf.c
@@ -51,6 +51,7 @@ static void ttf_cb_stream_seek(ttf_cb_stream_t * stream, size_t position);
 /**********************
  *      TYPEDEFS
  **********************/
+
 typedef struct ttf_font_desc {
     lv_fs_file_t file;
 #if LV_TINY_TTF_FILE_SUPPORT != 0
@@ -66,11 +67,8 @@ typedef struct ttf_font_desc {
     lv_font_kerning_t kerning;
 
     int cache_size;
-    int kernpair_cache_size;
-
     lv_cache_t * glyph_cache;
     lv_cache_t * draw_data_cache;
-    lv_cache_t * kernpair_cache;
 } ttf_font_desc_t;
 
 typedef struct _tiny_ttf_glyph_cache_data_t {
@@ -78,12 +76,6 @@ typedef struct _tiny_ttf_glyph_cache_data_t {
     int adv_w;
     lv_font_glyph_dsc_t glyph_dsc;
 } tiny_ttf_glyph_cache_data_t;
-
-typedef struct _tiny_ttf_kernpair_cache_data_t {
-    uint32_t g1;
-    uint32_t g2;
-    int k;
-} tiny_ttf_kernpair_cache_data_t;
 
 typedef struct _lv_tiny_ttf_cache_data_t {
     uint32_t glyph_index;
@@ -99,7 +91,7 @@ static const void * ttf_get_glyph_bitmap_cb(lv_font_glyph_dsc_t * g_dsc, lv_draw
 static void ttf_release_glyph_cb(const lv_font_t * font, lv_font_glyph_dsc_t * g_dsc);
 static lv_font_t * lv_tiny_ttf_create(const char * path, const void * data, size_t data_size,
                                       int32_t font_size, bool kerning,
-                                      size_t cache_size, size_t kernpair_cache_size );
+                                      size_t cache_size);
 
 static bool tiny_ttf_glyph_cache_create_cb(tiny_ttf_glyph_cache_data_t * node, void * user_data);
 static void tiny_ttf_glyph_cache_free_cb(tiny_ttf_glyph_cache_data_t * node, void * user_data);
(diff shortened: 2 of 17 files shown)


MacDragon:
Fixed the couple of unused vars causing check fail

MacDragon:
Added note to documentation that cache value is elements, not memory size, and mentioned kerning option.

src/libs/tiny_ttf/lv_tiny_ttf.c
```diff
@@ -82,13 +90,20 @@ static bool ttf_get_glyph_dsc_cb(const lv_font_t * font, lv_font_glyph_dsc_t * d
 static const void * ttf_get_glyph_bitmap_cb(lv_font_glyph_dsc_t * g_dsc, lv_draw_buf_t * draw_buf);
 static void ttf_release_glyph_cb(const lv_font_t * font, lv_font_glyph_dsc_t * g_dsc);
 static lv_font_t * lv_tiny_ttf_create(const char * path, const void * data, size_t data_size,
-                                      int32_t font_size,
+                                      int32_t font_size, bool kerning,
```
kisvegabor:
```suggestion
                                      int32_t font_size, lv_font_kerning_t kerning,
```

src/libs/tiny_ttf/lv_tiny_ttf.c
```diff
@@ -117,6 +132,20 @@ void lv_tiny_ttf_set_size(lv_font_t * font, int32_t font_size)
     stbtt_GetFontVMetrics(&dsc->info, &dsc->ascent, &dsc->descent, &line_gap);
     font->line_height = (int32_t)(dsc->scale * (dsc->ascent - dsc->descent + line_gap));
     font->base_line = (int32_t)(dsc->scale * (line_gap - dsc->descent));
+
+    // size change means cache needs to be invalidated.
```
kisvegabor: Please use `/* */` comments.

kisvegabor (commented):
Looks good to me, thank you. Just 2 minor notes.

MacDragon:
Fixed, parameter definition evidently slipped past in .c when fixing type usage.

kisvegabor (approved):

kisvegabor:
Thank you!

docs/libs/tiny_ttf.rst
```diff
@@ -29,11 +29,14 @@ entire time the font is being used.
...
 :cpp:expr:`lv_tiny_ttf_set_size(font, font_size)`.
 
-By default, a font will use up to 4KB of cache to speed up rendering
-glyphs. This maximum can be changed by using
-:cpp:expr:`lv_tiny_ttf_create_data_ex(data, data_size, font_size, cache_size)`
-or :cpp:expr:`lv_tiny_ttf_create_file_ex(path, font_size, cache_size)` (when
-available). The cache size is indicated in bytes.
+By default, a font will cache data for upto 256 glyphs elements to speed up rendering.
+This maximum can be changed by using
+:cpp:expr:`lv_tiny_ttf_create_data_ex(data, data_size, font_size, kerning, cache_size, )`
```
liamHowatt, line 34:
```suggestion
:cpp:expr:`lv_tiny_ttf_create_data_ex(data, data_size, font_size, kerning, cache_size)`
```

src/libs/tiny_ttf/lv_tiny_ttf.c
liamHowatt, line 1: Please remove
```c
#define tiny_ttf_cache LV_GLOBAL_DEFAULT()->tiny_ttf_cache
```
now that the global has been removed.

liamHowatt, line 1: A couple more `//` comments.

src/libs/tiny_ttf/lv_tiny_ttf.c
```diff
@@ -205,31 +218,51 @@ static bool ttf_get_glyph_dsc_cb(const lv_font_t * font, lv_font_glyph_dsc_t * d
...
+    if(font->kerning == LV_FONT_KERNING_NORMAL &&
+       unicode_letter_next != 0) { // check if we need to do any kerning calculations
+        uint32_t g1 = data->glyph_dsc.gid.index;
+
+        int g2 = 0;
+        search_key.unicode = unicode_letter_next; // reuse search key
+        lv_cache_entry_t * entry_next = lv_cache_acquire_or_create(dsc->glyph_cache, &search_key, (void *)dsc);
+
+        if(entry_next == NULL)
+            g2 = stbtt_FindGlyphIndex(&dsc->info, (int)unicode_letter_next);
```
liamHowatt, line 248: Would this ever return nonzero?

FASTSHIFT (approved):

liamHowatt (commented):
Awesome PR. Glad to see it got merged. I was partway through a review but I think most points are unimportant. Can maybe be disregarded.

Does `.devcontainer/__lv_conf.h__` need the new config too? @kisvegabor?

src/libs/tiny_ttf/lv_tiny_ttf.c
MacDragon: I didn't actually check what behaviour would be if cache size is zero - my assumption was that could return NULL then, as the code I started atop was doing a NULL check on acquired entry.

kisvegabor:
@MacDragon could you send a follow PR to address these minor issues pointed out by @liamHowatt?

MacDragon:
Sure, need to double check the behaviour on zero cache if the null check is pointless or not - but the formatting I'll at least get a PR up later.

src/libs/tiny_ttf/lv_tiny_ttf.c
MacDragon: Hm, yes, in the current implementation this check is pointless - if no cache is given then font rendering fails entirely, if cache is given, then this call shouldn't fail. - It should I suppose ideally work with cache size 0 still, though performance would be as bad of course.

For some reason my SDL compile is failing to init, which is currently complicating debug.

liamHowatt: Oh of course, if the cache size is zero then `lv_cache_acquire_or_create` will always fail and this line is what creates it, in that case.

You answered my question: yes it can return non-zero.

Thanks

MacDragon: More or less I think so, though the first lookup will already fail if cache size is zero which is an issue I think needs a workaround to allow non cached operation if desired. - that part I hadn't modified from the start position. Though I'm not sure how often anyone would want to use ttf with no cacheing as the performance drops so much, but it should probably be supported properly for completeness.

kisvegabor:
Thank you!

MacDragon:
still working on when got a bit of time, I just need to get a new reliable test bench setup on spare pi pico so I can verify timings whilst on holiday as desktop setup is too variable to rely on, but I think I found more potential speed up still whilst enabling non cached calls, will investigate further.

MacDragon:
Finally more or less got new test bench setup last night enough to do some testing whilst verifying output looks correct, and done some changes  so works with cache size 0 or 1 properly - though right now it's a tiny bit slower when using cache than existin branch version that I still need to investigate cause of, though it's close.

kisvegabor:
Great, thank you for the update!

PR #6443, opened 2024-06-28
fix(flex): Added null check around flex box pointer

aharms795:
Added null check around flex grow dsc pointer. This will stop showing 'p is not a known buffer' errors while using flex containers that aren't set to grow.

The better fix would be to add a null check to return immediately inside lv_mem_buf_release, but there are a lot more files that touch that function.


kisvegabor:
Thanks!

PR #6448, opened 2024-06-30
feat(display): add globle function lv_display_get_render_state

spbgzh:
When I change screen really fast, I need to avoid ERROR Asserted at expression: !disp->rendering_in_progress (Invalidate area is not allowed during rendering.), I need to check rendering_in_progress state, but disp its in lv_display_private.h, if I want to access it without c hearder "lvgl/src/display/lv_display_private.h", it will show error: invalid use of incomplete.
so in this case I suggest to set a globle function 'lv_display_get_render_state' to get rendering_in_progress state

```
/**
 * Get display render state
 * @param disp              pointer to a display
 * @return                  state of rendering_in_progress
 */
uint32_t lv_display_get_render_state(lv_display_t * disp);
```

```
uint32_t lv_display_get_render_state(lv_display_t * disp)
{
    return disp->rendering_in_progress;
}
```

### DIFF
--- a/src/display/lv_display.c
+++ b/src/display/lv_display.c
@@ -443,6 +443,12 @@ void lv_display_set_render_mode(lv_display_t * disp, lv_display_render_mode_t re
     disp->render_mode = render_mode;
 }
 
+uint32_t lv_display_get_render_state(lv_display_t * disp);
+{
+    return disp->rendering_in_progress;
+}
+
+
 void lv_display_set_flush_cb(lv_display_t * disp, lv_display_flush_cb_t flush_cb)
 {
     if(disp == NULL) disp = lv_display_get_default();
--- a/src/display/lv_display.c
+++ b/src/display/lv_display.c
@@ -443,7 +443,7 @@ void lv_display_set_render_mode(lv_display_t * disp, lv_display_render_mode_t re
     disp->render_mode = render_mode;
 }
 
-uint32_t lv_display_get_render_state(lv_display_t * disp);
+uint32_t lv_display_get_render_state(lv_display_t * disp)
 {
     return disp->rendering_in_progress;
 }
--- a/src/display/lv_display.c
+++ b/src/display/lv_display.c
@@ -448,7 +448,6 @@ uint32_t lv_display_get_render_state(lv_display_t * disp);
     return disp->rendering_in_progress;
 }
 
-
 void lv_display_set_flush_cb(lv_display_t * disp, lv_display_flush_cb_t flush_cb)
 {
     if(disp == NULL) disp = lv_display_get_default();
--- a/src/display/lv_display.h
+++ b/src/display/lv_display.h
@@ -261,7 +261,6 @@ void lv_display_set_render_mode(lv_display_t * disp, lv_display_render_mode_t re
  */
 uint32_t lv_display_get_render_state(lv_display_t * disp);
 
-
 /**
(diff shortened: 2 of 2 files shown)


spbgzh:
If you don't consider this function, you can add a check for d->rendering_in_progress in the lv_screen_load_anim function to prevent errors.

FASTSHIFT:
@spbgzh 
This problem is usually caused by users using the lvgl API across threads.
For example, lv_timer_handler works in the main loop, but lv_screen_xxx is called in another thread or in an interrupt function, which causes the lvgl's core data to be changed during reading. An error occurred that caused the thread to be unsafe. 

So please use lvgl single thread, or lock lvgl thread, refer to: https://docs.lvgl.io/master/porting/os.html

spbgzh:
@FASTSHIFT In fact, I did not use threads. I used the serial port to control the screen switching. When the screen switching is not completed, the serial port sends the command to switch the screen, which will cause the lvgl system to crash. For this reason, I added a check at the top of the serial port control command to determine whether the screen rendering is completed.

spbgzh:
This is my serial port log, where 01 means the operation is successful, and 00 means the operation fails. You can see that the process of controlling the switching screen includes the failed instructions.

[20:35:04.446] 01 A0 20 
[20:35:04.597] 00 A0 20 
[20:35:04.748] 01 A0 20 
[20:35:04.946] 00 A0 20 
[20:35:05.097] 01 A0 20 
[20:35:05.246] 01 A0 20 
[20:35:05.395] 00 A0 20 
[20:35:05.596] 01 A0 20 
[20:35:05.746] 01 A0 20 
[20:35:08.596] 01 A0 20 
[20:35:08.747] 01 A0 20 
[20:35:08.898] 00 A0 20 
[20:35:09.048] 00 A0 20 
[20:35:09.197] 01 A0 20 
[20:35:09.300] 00 A0 20 
[20:35:09.447] 00 A0 20 
[20:35:09.599] 01 A0 20

FASTSHIFT:
@spbgzh 
In a bare metal environment, the main loop and the serial interrupt can also be considered two threads, where the interrupt handler takes precedence over the main loop (because the interrupt handler can interrupt the main loop at any time).
There are many shared resources inside LVGL, such as heap and global variables, which do not need to be locked in a single-threaded environment because functions do not interrupt each other's execution.

If multi-threading is used, it will cause problems, for example, lv_timer_handler parses an obj to do the preparation work before rendering, but because of the interrupt, the user deletes the obj directly in the interrupt callback function. When the interrupt exits, the main loop accesses the data of the obj that has just been deleted, which leads to execution errors.

Adding the `rendering_in_progress` flag check is just a means to detect the problem, bypassing the error does not really solve the problem, because there are still many shared resources that are not protected.

The correct solution is to set a flag in the serial port interrupt callback to identify the page to jump to. Check the flag in the LVGL main loop (lv_timer is recommended), call lv_screen_load_anim to load the page based on the flag, and clear the flag when the call is complete, so there are no thread-safety issues.

FASTSHIFT:
I added a new code comment to illustrate: https://github.com/lvgl/lvgl/pull/6450

kisvegabor:
I think adding  `lv_display_get_render_state` would foster bad habits, like work-arounding multi threading issue.

kisvegabor:
Closing in light of [this](https://github.com/lvgl/lvgl/pull/6448#issuecomment-2200142117). Thank you for taking your time to contribute anyway! :+1:

PR #6467, opened 2024-07-04
feat(textarea): make textarea editable

massoudsoltanian:
Although currently the textarea is editable, the flag is by default set to LV_OBJ_CLASS_EDITABLE_FALSE.  the problem is that we have a weird (horizontal in our experience) scroll jump in textarea when we move cursor one character to the left or right. This weird behavior comes from the fact that it enters to this block of code:  https://github.com/lvgl/lvgl/blob/c3d4510f843b40d74e2b42ca9ba790892084288e/src/core/lv_obj.c#L652C8-L652C26 which in my opinion should not.


C47D (approved):

kisvegabor (approved):
Great catch!

FASTSHIFT (approved):

PR #6486, opened 2024-07-08
feat(tests): add performance report action cleanup and refactor

AnthoJack:

- [x] The job ID is used by the action to retrieve the execution time data from the last successful run and isn't the same in the forked and original repo so it had to be changed
- [x]  GCC features a compilation flag that instruments function calls to call other functions at entry and exit. Using it (if possible) would mean fewer caveats (ARM64 only, limited arguments) and more control over instrumentation functions features compared to current home-made solution 

### Notes

- [x] Update Documentation
- [x] Update Examples
- [x] Fix code style errors

### DIFF
--- a/tests/perf/instr_func.c
+++ b/tests/perf/instr_func.c
@@ -1,0 +1,22 @@
+#include <stdlib.h>
+#include <stdio.h>
+#include <stdint.h>
+#include <time.h>
+
+void __cyg_profile_func_enter (void *this_fn, void *call_site)
+{
+    struct timespec ts;
+	// Get timestamp
+	clock_gettime(CLOCK_REALTIME, &ts);
+	// Write timestamp with identifier
+	printf("prof > %lx: %llu.%llu\n", call_site, ts.tv_sec, ts.tv_nsec);
+}
+
+void __cyg_profile_func_exit  (void *this_fn, void *call_site)
+{
+    struct timespec ts;
+	// Get timestamp
+	clock_gettime(CLOCK_REALTIME, &ts);
+	// Write timestamp with identifier
+	printf("prof < %lx: %llu.%llu\n", call_site, ts.tv_sec, ts.tv_nsec);
+}
--- a/tests/perf/instr_func.c
+++ b/tests/perf/instr_func.c
@@ -3,6 +3,15 @@
 #include <stdint.h>
 #include <time.h>
 
+/**
+ * Entry function for function instrumentation
+ * This functions will be called at the start of an instrumented function's execution 
+ * (see gcc's "-finstrument functions" flag) and provided with the address
+ * of the instrumented function (this_fn) and the address where this function
+ * was called from in the parent function (call_site)
+ * 
+ * Prints a message to timestamp the entry time of the instrumented function
+ */
 void __cyg_profile_func_enter(void * this_fn, void * call_site)
 {
     struct timespec ts;
@@ -12,6 +21,15 @@ void __cyg_profile_func_enter(void * this_fn, void * call_site)
     printf("prof > %lx: %llu.%llu\n", call_site, ts.tv_sec, ts.tv_nsec);
 }
 
+/**
+ * Exit function for function instrumentation
+ * This functions will be called at the end of an instrumented function's execution
+ * (see gcc's "-finstrument functions" flag) and provided with the address
+ * of the instrumented function (this_fn) and the address where this function
+ * was called from in the parent function (call_site)
+ * 
+ * Prints a message to timestamp the exit time of the instrumented function
+ */
 void __cyg_profile_func_exit(void * this_fn, void * call_site)
 {
     struct timespec ts;
--- a/tests/perf/instr_func.c
+++ b/tests/perf/instr_func.c
@@ -5,11 +5,11 @@
 
 /**
  * Entry function for function instrumentation
- * This functions will be called at the start of an instrumented function's execution 
+ * This functions will be called at the start of an instrumented function's execution
  * (see gcc's "-finstrument functions" flag) and provided with the address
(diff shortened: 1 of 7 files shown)


AnthoJack:
Hi @kisvegabor 
I've tried using the script you mentioned to fix the code style errors but when I do I get:

> $ python3 scripts/code-format.py 
>
> Formatting demos
> sh: 1: astyle: not found
>
> Formatting examples
> sh: 1: astyle: not found
> Formatting src
> sh: 1: astyle: not found
>
> Formatting tests
> sh: 1: astyle: not found

I must admit I'm not very familiar with this type of script so I don't know about any dependency I should be providing. If you could give me a few pointers it'd be very appreciated

kisvegabor:
Hi, 

You can install it with:
```sh
git clone https://github.com/szepeviktor/astyle.git
cd astyle 
git checkout v3.4.12
cd astyle/build/gcc/
make
sudo make install
astyle --version
```

AnthoJack:
Currently, the only way to run it would be to reproduce what the action is doing:

1. Download the Dockerfile.lvgl using the link in the action
2. Put it somewhere were there's an lvgl repository clone
3. Build the docker image
4. Run the image, the result file should be in the same folder
5. Alternatively, you can run the docker image interactively (override the default command with bash) and execute "./st" yourself

If you want to be able to build and launch the test application only (and without using docker) to check if it works (without function instrumentation), It is necessary to modify the build pipeline in lvgl to add the perf test application build

kisvegabor:
> If you want to be able to build and launch the test application only (and without using docker) to check if it works (without function instrumentation), It is necessary to modify the build pipeline in lvgl to add the perf test application build

I don't have a particular issue with Docker, but it's an other layer of complexity and place for potential failures.  Does "add the perf test application build" mean adding a new CMakeFile or so? If yes, I vote for it.

AnthoJack:
There already is a CMakeFile.txt in the folder so it could be that all that's needed is to add something like "add_subdirectory(perf)" in the CMakeFile.txt of the parent folder to build it with all the other tests but I have no idea how your cmake architecture works so I can't guarantee anything and to be honest, I don't know cmake well enough to understand it quickly

kisvegabor:
> I don't know cmake well enough to understand it quickly

Neither am I :smile: I'm working on the preparation of the v9.2 release and get back to this update only a little later. :slightly_frowning_face:

kisvegabor:
Not stale

kisvegabor:
@liamHowatt Next week if you will have a few free hours could you take a look at this one?

kisvegabor:
Not stale

tests/perf/instr_func.c
```diff
@@ -0,0 +1,40 @@
...
+ * of the instrumented function (this_fn) and the address where this function
+ * was called from in the parent function (call_site)
+ *
+ * Prints a message to timestamp the entry time of the instrumented function
+ */
+void __cyg_profile_func_enter(void * this_fn, void * call_site)
+{
+    struct timespec ts;
+    // Get timestamp
+    clock_gettime(CLOCK_REALTIME, &ts);
```
liamHowatt, line 19: Do you know if a clock like `CLOCK_PROCESS_CPUTIME_ID` could produce more consistent/reproducible results? I think a clock that measures only process/thread time like this would help ignore time incurred by other loads on the system.

tests/perf/instr_func.c
```diff
@@ -0,0 +1,40 @@
...
+    // Write timestamp with identifier
+    printf("prof > %lx: %llu.%llu\n", call_site, ts.tv_sec, ts.tv_nsec);
```
liamHowatt, line 21: Could you use `%p` instead of `%lx`? If it's only run on aarch64 or x86-64 Linux then it's fine.

liamHowatt (commented):
I ran the dockerfile successfully. After running, it produced a `.txt`. I had to run the container interactively to get the `.log` file where I see the `prof > ...` lines! The strange thing is that those `prof > ...` lines are there regardless of whether I use the lvgl with your PR checked out or the current `ci/perf_test`. I don't think it's a docker caching thing. I tried using `--no-cache`.

```
...

Profiling LVGL
	Profiling lv_bezier3
		Cubic bezier curve with parameters (0;50;954;1024), t = 256
		Without profiling: res = 171
prof > 1ca8: 0.590511824
prof > 7918: 0.590529328
prof < 7918: 0.590546976
prof < 1ca8: 0.590564480
		With profiling: res = 171
	Profiling _lv_area_intersect
		Area a1 (0;0)->(4;3)
		Area a2 (3;1)->(6;4)
		Calculating intersection
		Without profiling: res = (3;1)->(4;3)
prof > 1d78: 0.590663856
prof < 1d78: 0.590681408
		With profiling: res = (3;1)->(4;3)
```

Fyi the CI for the action on your branch seems to not not find the previous successful run: https://github.com/daniel-rossier/lvgl/actions/runs/9956413477/job/27506268212
Maybe it's in a different repo? [lvgl](https://github.com/lvgl/lvgl/actions/workflows/53635437/runs), [daniel-rossier](https://github.com/daniel-rossier/lvgl/actions/workflows/53635437/runs)

tests/perf/instr_func.c
liamHowatt: Also, should nsecs be zero-padded? That way 50,000,000 nsecs can't be mistaken for 500,000,000 nsecs. You should be able to do `%09llu` for 9 zeros.

kissa96 (approved):

kisvegabor:
@daniel-rossier @AnthoJack what are the plans with this PR? 

It would be great if we could add just a single test to measure the 
- creation time of `lv_demo_widgets()`
- the first rendering time (some layout calculation are happening here)
- invalidate screen and measure the second rendering time.

PR #6494, opened 2024-07-09
fix(display): delete previous screen instead of current

lorenzo-arena:

The current implementation of the `lv_scr_load_anim` checks if an animation is in progress; if so it correctly tries to immediately stop the animations so that the new ones can be started. However, after stopping the animation, the current screen is deleted instead of the previous one (if the `auto_del` flag was set) and this results in segmentation faults down the road when the current screen is accessed.

### DIFF
--- a/src/display/lv_display.c
+++ b/src/display/lv_display.c
@@ -592,10 +592,8 @@ void lv_screen_load_anim(lv_obj_t * new_scr, lv_screen_load_anim_t anim_type, ui
         lv_obj_set_pos(d->scr_to_load, 0, 0);
         lv_obj_remove_local_style_prop(d->scr_to_load, LV_STYLE_OPA, 0);
 
-        if(d->del_prev) {
-            lv_obj_delete(act_scr);
-        }
-        act_scr = d->act_scr; /*Active screen changed.*/
+        d->prev_scr = d->act_scr;
+        act_scr = d->scr_to_load; /*Active screen changed.*/
 
         scr_load_internal(d->scr_to_load);
     }
--- a/tests/src/test_cases/test_screen_load.c
+++ b/tests/src/test_cases/test_screen_load.c
@@ -18,4 +18,117 @@ void test_screen_load_no_crash(void)
     lv_screen_load_anim(screen_with_anim_2, LV_SCR_LOAD_ANIM_OVER_RIGHT, 1000, 500, false);
 }
 
+void test_screen_load_with_delete_no_crash(void)
+{
+    /*Delete active screen and load new screen should not crash*/
+    lv_obj_t * screen = lv_screen_active();
+    lv_obj_del(screen);
+    screen = lv_obj_create(NULL);
+    lv_scr_load(screen);
+
+    /*Consecutively loading multiple screens (while deleting one) with transition animations should not crash*/
+    lv_obj_t * screen_with_anim_1 = lv_obj_create(NULL);
+    lv_obj_t * screen_with_anim_2 = lv_obj_create(NULL);
+    lv_obj_t * screen_with_anim_3 = lv_obj_create(NULL);
+
+    lv_scr_load_anim(screen_with_anim_1, LV_SCR_LOAD_ANIM_OVER_LEFT, 0, 0, false);
+    lv_scr_load_anim(screen_with_anim_2, LV_SCR_LOAD_ANIM_OVER_RIGHT, 1000, 0, true);
+
+    /*Wait to trigger the animation start callbacks*/
+    lv_test_wait(100);
+
+    lv_scr_load_anim(screen_with_anim_3, LV_SCR_LOAD_ANIM_OVER_LEFT, 200, 0, true);
+
+    /*The active screen should be immediately replaced*/
+    TEST_ASSERT_EQUAL(lv_screen_active(), screen_with_anim_2);
+
+    lv_test_wait(400);
+
+    /*Check for the screens status after the transition*/
+    TEST_ASSERT_EQUAL(lv_obj_is_valid(screen_with_anim_1), false);
+    TEST_ASSERT_EQUAL(lv_obj_is_valid(screen_with_anim_2), false);
+    TEST_ASSERT_EQUAL(lv_obj_is_valid(screen_with_anim_3), true);
+}
+
+void test_screen_load_with_delete_no_crash2(void)
+{
+    /*Delete active screen and load new screen should not crash*/
+    lv_obj_t * screen = lv_screen_active();
+    lv_obj_del(screen);
+    screen = lv_obj_create(NULL);
+    lv_scr_load(screen);
+
+    /*Consecutively loading multiple screens (while deleting one) with transition animations should not crash*/
+    lv_obj_t * screen_with_anim_1 = lv_obj_create(NULL);
+    lv_obj_t * screen_with_anim_2 = lv_obj_create(NULL);
+    lv_obj_t * screen_with_anim_3 = lv_obj_create(NULL);
+    lv_obj_t * screen_with_anim_4 = lv_obj_create(NULL);
+
+    lv_scr_load_anim(screen_with_anim_1, LV_SCR_LOAD_ANIM_OVER_LEFT, 0, 0, false);
+    lv_scr_load_anim(screen_with_anim_2, LV_SCR_LOAD_ANIM_OVER_RIGHT, 1000, 0, true);
+    lv_scr_load_anim(screen_with_anim_3, LV_SCR_LOAD_ANIM_OVER_LEFT, 0, 0, true);
+
+    /*Wait to trigger the animation start callbacks*/
+    lv_test_wait(100);
+
+    lv_scr_load_anim(screen_with_anim_4, LV_SCR_LOAD_ANIM_OVER_LEFT, 200, 0, true);
+
+    /*The active screen should be immediately replaced*/
+    TEST_ASSERT_EQUAL(lv_screen_active(), screen_with_anim_3);
+
+    lv_test_wait(400);
+
+    /*Check for the screens status after the transition*/
+    TEST_ASSERT_EQUAL(lv_obj_is_valid(screen_with_anim_1), false);
+    TEST_ASSERT_EQUAL(lv_obj_is_valid(screen_with_anim_2), false);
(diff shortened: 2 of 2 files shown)


lorenzo-arena:
I reworked the PR a little bit; I didn't notice that the `prev_scr` was already removed after the block which checks for animations in progress, so I just remove the "wrong" delete of the current screen and instead swapped the active screen with the one which was meant to load in the previous `lv_screen_load_anim` call (the one for which the animation is currently running).

kisvegabor:
Thanks for investigating and fixing it :100: 

Could you also add test for it to be sure that it won't be broken again?

lorenzo-arena:
> Could you also add test for it to be sure that it won't be broken again?

I'll set the PR as a draft while I implement one

lorenzo-arena:
I tried adding a test, and while it fails in the v8.4 branch (the one which I'm currently using in production) it doesn't in the master branch. If I'm understanding it correctly, this is due to the fact that the check for transition in progress is done with the following check in the master branch

```c
/*If another screen load animation is in progress
 *make target screen loaded immediately. */
if(d->scr_to_load && act_scr != d->scr_to_load) {
```

and with the following one in the v8.4 branch

```c
/*If an other screen load animation is in progress
 *make target screen loaded immediately. */
if(d->scr_to_load) {
```

The master branch is free from this bug due the `act_scr != d->scr_to_load` check, which never seems to be true. So, while I'll add the same test for the https://github.com/lvgl/lvgl/pull/6495 backport, I'm starting to question if this block is necessary at all

kisvegabor:
Thank you forr adding test, Lorenzo. 

 > The master branch is free from this bug due the act_scr != d->scr_to_load check, which never seems to be true. So, while I'll add the same test for the https://github.com/lvgl/lvgl/pull/6495 backport, I'm starting to question if this block is necessary at all

I think the whole "screen load with animation" section cab be backported to v8.4 (including this `if` statement as well), as this part hasn't really changed.

lorenzo-arena:
Should I drop e637c8fb2fce44370bec7e543392894d5103d805 then, and leave the screen load with animation block as is?

kisvegabor:
Isn't e637c8fb2fce44370bec7e543392894d5103d805 required to fix it? 

What I mean is that we can backport the whole v9 logic to v8.

lorenzo-arena:
To recap, what I did (and did not) understand is the following:

* the test doesn't fail on the v9 branch, even without the patch
* this is due to the fact that the `act_scr != d->scr_to_load` condition prevents the if block to be executed and the same for the `lv_obj_delete(act_scr);`
* I don't know if that delete would be "good" if those condition are met, and thus I'm not so sure about removing it; if you think that the delete is indeed a bug I'll keep that commit in the PR
* the same test does indeed fail on the v8.4 branch, however by adding the same condition to the `if` block the situation becomes the same that is true for the v9 branch

lorenzo-arena:
I removed the draft status, I'll also backport the if condition to the v8.4 branch

kisvegabor (dismissed):
Thanks! 

It's a little bit hard to think of all the possible use cases, however I've tested it locally and it worked well with reasonable combinations.

So please apply the same to v8.4.

lorenzo-arena:
https://github.com/lvgl/lvgl/pull/6495 is updated with the new check, as well as the "removed" delete and the new screen load test

liamHowatt:
Thanks!

I found a case where a screen won't be deleted.

`screen_with_anim_1` does not get deleted in this slightly modified version of the test case. The test fails at `TEST_ASSERT_EQUAL(lv_obj_is_valid(screen_with_anim_1), false);`.

Sidenote: I made sure it wasn't because of the sometimes inaccurate results of `lv_obj_is_valid`.

```c
void test_screen_load_with_delete_no_crash(void)
{
    /*Delete active screen and load new screen should not crash*/
    lv_obj_t * screen = lv_screen_active();
    lv_obj_del(screen);
    screen = lv_obj_create(NULL);
    lv_scr_load(screen);

    /*Consecutively loading multiple screens (while deleting one) with transition animations should not crash*/
    lv_obj_t * screen_with_anim_1 = lv_obj_create(NULL);
    lv_obj_t * screen_with_anim_2 = lv_obj_create(NULL);
    lv_obj_t * screen_with_anim_3 = lv_obj_create(NULL);
    lv_obj_t * screen_with_anim_4 = lv_obj_create(NULL);

    lv_scr_load_anim(screen_with_anim_1, LV_SCR_LOAD_ANIM_OVER_LEFT, 0, 0, false);
    lv_scr_load_anim(screen_with_anim_2, LV_SCR_LOAD_ANIM_OVER_RIGHT, 1000, 0, true);
    lv_scr_load_anim(screen_with_anim_3, LV_SCR_LOAD_ANIM_OVER_LEFT, 0, 0, true);

    /*Wait to trigger the animation start callbacks*/
    lv_test_wait(100);

    lv_scr_load_anim(screen_with_anim_4, LV_SCR_LOAD_ANIM_OVER_LEFT, 200, 0, true);

    /*The active screen should be immediately replaced*/
    TEST_ASSERT_EQUAL(lv_screen_active(), screen_with_anim_3);

    lv_test_wait(400);

    /*Check for the screens status after the transition*/
    TEST_ASSERT_EQUAL(lv_obj_is_valid(screen_with_anim_1), false);
    TEST_ASSERT_EQUAL(lv_obj_is_valid(screen_with_anim_2), false);
    TEST_ASSERT_EQUAL(lv_obj_is_valid(screen_with_anim_3), false);
    TEST_ASSERT_EQUAL(lv_obj_is_valid(screen_with_anim_4), true);
}
```

I fixed it by adding back

```c
        if(d->del_prev) {
            lv_obj_delete(act_scr);
        }
```

I think you chainging `act_scr = d->act_scr;` to `act_scr = d->scr_to_load;` was a very important fix, however.

I see your concern that `if(d->del_prev) ...` deletes the wrong screen. Are you able to give an example where the wrong screen is deleted? Thanks!

lorenzo-arena:
> I see your concern that if(d->del_prev) ... deletes the wrong screen. Are you able to give an example where the wrong screen is deleted? Thanks!

I first got the issue in the v8.4 branch, where the `if(d->del_prev) ...` deletes the wrong screen since the if check before that was a little bit different, and the test which I added was failing. However, that was fixed by either removing the `if(d->del_prev) ...` of backporting the new `if(d->scr_to_load && act_scr != d->scr_to_load)` condition before that.

> I found a case where a screen won't be deleted.

That is failing indeed however I have another proposal: what if instead of restoring that delete we replace it with `d->prev_scr = d->act_scr;`? I tested it and with that both tests pass (the delete is performed by the `if(d->prev_scr && d->del_prev) lv_obj_delete(d->prev_scr);` instruction shortly after the if block. That would also be more aligned with what is done in the `scr_load_anim_start`.

liamHowatt:
Oh yeah, sounds solid! Feel free to push so it's easier to review. Would `d->prev_scr` need to be checked and deleted first before being reassigned to `d->act_scr`? Also, I wonder if everywhere `act_screen` is being used, that should actually be `d->prev_scr`.

lorenzo-arena:
> Feel free to push so it's easier to review.

Done, I did a couple of fixups on the current commits; I'll also backport everything to the v8.4 branch.

> Would `d->prev_scr` need to be checked and deleted first before being reassigned to `d->act_scr`?

I think at this point there are 4 possible cases, dictated by if the load is requested with no time and no delay (i.e. `(time == 0 && delay == 0)`), or with some animation time or delay, and if there was a previous animation in progress:
- no time/no delay + no previous animation in progress is correctly covered by the `/*Shortcut for immediate load*/`
- no time/no delay + a previous animation is still correctly covered since the previous screen will be deleted after the animation check and the screen which was being loaded will be deleted as well in the `/*Shortcut for immediate load*/`
- some anim time or delay + no previous animation will be covered by the usual animation path with the `scr_load_anim_start` and `scr_anim_completed` callbacks
- some anim time or delay + a previous animation will be covered since the previous animation gets immediately loaded, the screen which was active gets moved to the previous and deleted, then the usual animation path kicks in (also if the previous animation was for the first screen (so `d->act_scr` was `NULL`) the delete would be skipped since it's checked in `if(d->prev_scr && d->del_prev) lv_obj_delete(d->prev_scr);`.

The only corner case is the one which came up with the last test but it should be correctly handled both if the animation got started or not, with something like:

```c
/* Here the loads are made one after the other and the scr_load_anim_start doesn't trigger
 * after loading screen_with_anim_3; the deletion of screen_with_anim_1 will
 * be correctly handled by setting it to prev_scr (as would be done in the scr_load_anim_start)
 */
lv_scr_load_anim(screen_with_anim_1, LV_SCR_LOAD_ANIM_OVER_LEFT, 0, 0, false);
lv_scr_load_anim(screen_with_anim_2, LV_SCR_LOAD_ANIM_OVER_RIGHT, 1000, 0, true);
lv_scr_load_anim(screen_with_anim_3, LV_SCR_LOAD_ANIM_OVER_LEFT, 0, 0, true);

/* Here scr_load_anim_start is executed after loading screen_with_anim_2, and while
 * loading screen_with_anim_3 the new "d->prev_scr = d->act_scr;" assignment would
 * be skipped since the "if(d->scr_to_load && act_scr != d->scr_to_load)" condition
 * would not be true (both scr_to_load and act_scr would point to screen_with_anim_2);
 * prev_scr would already point to screen_with_anim_1 (as set in the scr_load_anim_start)
 * and the usual "if(d->prev_scr && d->del_prev) lv_obj_delete(d->prev_scr);" instruction
 * would take care of its deletion
 */
lv_scr_load_anim(screen_with_anim_1, LV_SCR_LOAD_ANIM_OVER_LEFT, 0, 0, false);
lv_scr_load_anim(screen_with_anim_2, LV_SCR_LOAD_ANIM_OVER_RIGHT, 1000, 0, true);
lv_test_wait(100);
lv_scr_load_anim(screen_with_anim_3, LV_SCR_LOAD_ANIM_OVER_LEFT, 0, 0, true);
```

> Also, I wonder if everywhere `act_screen` is being used, that should actually be `d->prev_scr`

I think `prev_scr` is probably only needed for corner cases like the one before but still needed nonetheless.

liamHowatt (approved):
Nice, it's amazing that you solved my edge case and the original cases at the same time. It is still tricky to reason about but you clearly thought it through and it's a huge net improvement.

FASTSHIFT (approved):

PR #6499, opened 2024-07-11
fix(refr): eliminate side effect in assert

### LINKED ISSUE #6497: Side effects inside assert statement
### Introduce the problem

In addition to the special-type `LV_ASSERT_FOO` macros enabled by the `LV_USE_ASSERT_FOO` macros in `lv_conf.h`, there is an underlying LV_ASSERT macro which appears to be used 'bare' in handful of places.

Unlike the `LV_ASSERT_FOO` macros, this one cannot be disabled for two reasons. First, there appears to be no mechanism for doing so. Second, disabling it (by defining it to nothing) would cause problems because in at least one place it wraps a function call.

https://github.com/lvgl/lvgl/blob/ca70164c6b409ecfc09b407bd3d85c734fdfe67e/src/core/lv_refr.c#L587

### Proposal

Instead of
```
LV_ASSERT(lv_draw_buf_reshape( ... ) != NULL);
```
it would be better to use
```
lv_draw_buf_t *ret = lv_draw_buf_reshape( ... );
LV_ASSERT(ret != NULL);
```
Also, since this is a check for not null, `LV_ASSERT_NULL` can be used instead of `LV_ASSERT` here.

In general, expressions inside assertions should never have side effects, so that they can be safely eliminated from non-debug builds.

mdkendall:
fixes #6497

### Description of the fix

Refactor `layer_reshape_draw_buf` to move the call to `lv_draw_buf_reshape` outside the `ASSERT` macro, so it can be safely eliminated from non-debug builds.

### Notes

### DIFF
--- a/src/core/lv_refr.c
+++ b/src/core/lv_refr.c
@@ -590,7 +590,7 @@ static void layer_reshape_draw_buf(lv_layer_t * layer)
                               lv_area_get_width(&layer->buf_area),
                               lv_area_get_height(&layer->buf_area),
                               0);
-    (void)ret; /* Unused if LV_USE_ASSERT_NULL is disabled */
+    LV_UNUSED(ret);
     LV_ASSERT_NULL(ret);
 }
 
--- a/src/core/lv_refr.c
+++ b/src/core/lv_refr.c
@@ -584,13 +584,13 @@ static void refr_invalid_areas(void)
  */
 static void layer_reshape_draw_buf(lv_layer_t * layer)
 {
-    LV_ASSERT(lv_draw_buf_reshape(
-                  layer->draw_buf,
-                  layer->color_format,
-                  lv_area_get_width(&layer->buf_area),
-                  lv_area_get_height(&layer->buf_area),
-                  0)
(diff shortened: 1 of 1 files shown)


src/core/lv_refr.c
```diff
@@ -584,13 +584,14 @@ static void refr_invalid_areas(void)
...
-                  lv_area_get_height(&layer->buf_area),
-                  0)
-              != NULL);
+    lv_draw_buf_t * ret = lv_draw_buf_reshape(
+                              layer->draw_buf,
+                              layer->color_format,
+                              lv_area_get_width(&layer->buf_area),
+                              lv_area_get_height(&layer->buf_area),
+                              0);
+    (void)ret; /* Unused if LV_USE_ASSERT_NULL is disabled */
```
FASTSHIFT: LV_UNUSED(ret);

FASTSHIFT (changes requested):

FASTSHIFT (approved):

kisvegabor (approved):
Thanks!

PR #6503, opened 2024-07-12
fix(obj_tree): fix incorrect return value of function lv_obj_get_sibling_by_type()

wy1107412334:
### fix(obj_tree): incorrect return value of function lv_obj_get_sibling_by_type()

The class of the object returned by lv_obj_get_sibling_by_type() is not always class_p.


liamHowatt (commented):
Thank you, but I don't think the previous code was wrong. I think the `idx` parameter may be causing an issue for you. If you call `lv_obj_get_sibling_by_type` with the `idx` parameter as `0`, does it work in your application as-is?

kisvegabor:
I think this fixed is really appropriate. If `parent` has mixed buttons and labels, and you are interested only in the buttons `return lv_obj_get_child(parent, sibling_idx);` will ignore the type you have requested.

liamHowatt (approved):
Gotcha. So the `idx` gets the Nth child of class `class_p`.

kisvegabor (approved):

PR #6509, opened 2024-07-15
fix(test): fix filter option dot escape by setting regexp string

lorenzo-arena:

While running tests on Ubuntu 24.04 a warning is emitted:

```bash
./main.py:142: SyntaxWarning: invalid escape sequence '\.'
  '--html-title', 'LVGL Test Coverage', '--filter', '../src/.*/lv_.*\.c']
```

This is due to the fact that Ubuntu 24.04 includes Python 3.12 and from this version [a SyntaxWarning is emitted](https://docs.python.org/dev/whatsnew/3.12.html#other-language-changes) (it is also state that it will eventually become a SyntaxError) for invalid escape sequences.


liamHowatt (approved):
Nice

kisvegabor (approved):
Thanks!

PR #6543, opened 2024-07-22
feat(event): Add check for unexpected event list modifications

XuNeo:

Introduce an option to detect if the event list is altered (e.g., events added or removed) during user callbacks, which could lead to array resizing and potentially cause crashes. Given the high frequency of lv_event_send calls, this check is disabled by default to avoid performance impacts.

```c
static void clicked_cb(lv_event_t *e)
{
  printf("clicked\n");
  lv_obj_remove_event_cb(e->current_target, clicked_cb);
}

void test_obj_event(void)
{
  lv_obj_t *obj = lv_obj_create(lv_screen_active());
  lv_obj_center(obj);
  lv_obj_add_event_cb(obj, clicked_cb, LV_EVENT_CLICKED, NULL);
}
```
### Notes

### DIFF
--- a/.devcontainer/__lv_conf.h__
+++ b/.devcontainer/__lv_conf.h__
@@ -234,6 +234,7 @@
 #define LV_USE_ASSERT_NULL          1   /*Check if the parameter is NULL. (Very fast, recommended)*/
 #define LV_USE_ASSERT_MALLOC        1   /*Checks is the memory is successfully allocated or no. (Very fast, recommended)*/
 #define LV_USE_ASSERT_STYLE         0   /*Check if the styles are properly initialized. (Very fast, recommended)*/
+#define LV_USE_ASSERT_EVENT         0   /*Check if the event list changed while iterating it.*/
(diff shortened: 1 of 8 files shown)


XuNeo:
Will provide a better way to detect event array list changes during iterating.

Idea is to add a flag to indicate iterating. If event is removed during iteration, simply mark the event->cb to NULL but do not remove it from array. After iteration, remove those holes.

Event can be added during iteration.

kisvegabor:
> Idea is to add a flag to indicate iterating. If

Good idea! Note that in rare cases "iteration" can be nested. E.g. from a click event  you can send a refresh or key event manually. It can be resolved by using a counter to indicate iterations instead of a flag.

kisvegabor:
We can also traverse `event_head` in `lv_obj_event.c` when an event is removed. We can see if the widget whose event shall be removed is active (i.e. can be found in `event_head`'s linked list) we mark the event_cb as NULL as you have proposed. 

In most of the cases event removal happens when the events are not running so it will be fast.

PR #6544, opened 2024-07-22
perf(array): optimize array remove / erase function performance

onecoolx:

### Notes

### DIFF
--- a/src/misc/lv_array.c
+++ b/src/misc/lv_array.c
@@ -75,7 +75,10 @@ lv_result_t lv_array_remove(lv_array_t * array, uint32_t index)
 
     /*Shortcut*/
     if(index == array->size - 1) {
-        lv_array_resize(array, array->size - 1);
+        array->size--;
+        if(array->size <= array->capacity >> 1) {
+            lv_array_resize(array, array->size);
+        }
         return LV_RESULT_OK;
     }
 
@@ -83,7 +86,10 @@ lv_result_t lv_array_remove(lv_array_t * array, uint32_t index)
     uint8_t * remaining = start + array->element_size;
     uint32_t remaining_size = (array->size - index - 1) * array->element_size;
     lv_memmove(start, remaining, remaining_size);
-    lv_array_resize(array, array->size - 1);
+    array->size--;
+    if(array->size <= array->capacity >> 1) {
+        lv_array_resize(array, array->size);
+    }
     return LV_RESULT_OK;
 }
 
@@ -99,7 +105,10 @@ lv_result_t lv_array_erase(lv_array_t * array, uint32_t start, uint32_t end)
 
     /*Shortcut*/
     if(end == array->size) {
-        lv_array_resize(array, start);
+        array->size = start;
+        if(array->size <= array->capacity >> 1) {
+            lv_array_resize(array, array->size);
+        }
         return LV_RESULT_OK;
     }
 
@@ -107,7 +116,10 @@ lv_result_t lv_array_erase(lv_array_t * array, uint32_t start, uint32_t end)
     uint8_t * remaining = start_p + (end - start) * array->element_size;
     uint32_t remaining_size = (array->size - end) * array->element_size;
     lv_memcpy(start_p, remaining, remaining_size);
-    lv_array_resize(array, array->size - (end - start));
+    array->size -= (end - start);
+    if(array->size <= array->capacity >> 1) {
+        lv_array_resize(array, array->size);
+    }
     return LV_RESULT_OK;
 }
 
--- a/src/misc/lv_array.c
+++ b/src/misc/lv_array.c
@@ -67,6 +67,13 @@ void lv_array_copy(lv_array_t * target, const lv_array_t * source)
     target->size = source->size;
 }
 
+void lv_array_shrink(lv_array_t * array)
+{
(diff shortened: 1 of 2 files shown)


onecoolx:
optimize array remove / erase function performance
@XuNeo @FASTSHIFT @kisvegabor

src/misc/lv_array.c
```diff
@@ -75,15 +75,21 @@ lv_result_t lv_array_remove(lv_array_t * array, uint32_t index)
 
     /*Shortcut*/
     if(index == array->size - 1) {
-        lv_array_resize(array, array->size - 1);
+        array->size--;
```
XuNeo, line 85: Let's add a static function and use it elsewhere.
```c
lv_array_shrink(array, capacity) {
   if (capacity <= array->capacity / 2) {
      lv_array_resize(array, capacity);
   }
}
```

PGNetHun: I agree that this new functionality should be in a separate function.


... and I wonder if it should be behind a toggle: 
- by default resize array to half when fill ratio decreases to 50%
- but on devices with much less memory keep the original resize logic (AKA resize immediately), because keeping many of these "blown up" arrays in memory without decreasing them could "eat up memory".
This is only an assumption if there are too many `lv_array`s used... but we would see clearer picture if it's really measured (number of arrays, bytes used / capacity (fill ratio) through time, etc)

onecoolx: > I agree that this new functionality should be in a separate function.
> 
> ... and I wonder if it should be behind a toggle:
> 
> * by default resize array to half when fill ratio decreases to 50%
> * but on devices with much less memory keep the original resize logic (AKA resize immediately), because keeping many of these "blown up" arrays in memory without decreasing them could "eat up memory".
>   This is only an assumption if there are too many `lv_array`s used... but we would see clearer picture if it's really measured (number of arrays, bytes used / capacity (fill ratio) through time, etc)

 I disagree with this perspective. Even on devices with limited memory, it is not advisable to frequently use the resize function. This can result in higher memory fragmentation, leading to faster memory depletion and significantly impacting overall performance.
I agree that the size of this resize could be made configurable to quickly free up memory on lower-end devices, for example config it to 30% or 20 %.

PGNetHun (approved):

src/misc/lv_array.h
```diff
@@ -23,6 +23,10 @@ extern "C" {
 #define LV_ARRAY_DEFAULT_CAPACITY  4
 #endif
 
+#ifndef LV_ARRAY_DEFAULT_SHRINK_RATIO
+#define LV_ARRAY_DEFAULT_SHRINK_RATIO 2
```
kisvegabor, line 27: I'm not sure if we shall define it as a ratio. Having and absolute number, e.g. `#define LV_ARRAY_MAX_FREE_CAPACITY` would be more predictable.

onecoolx: > I'm not sure if we shall define it as a ratio. Having and absolute number, e.g. `#define LV_ARRAY_MAX_FREE_CAPACITY` would be more predictable.

I believe that defining the memory release size as an absolute value may not always be suitable. For instance, if it is set to 8, frequent resizing may still be necessary for large lists exceeding 256 elements, while small lists comprising less than 8 elements may no chance to be released

kisvegabor: I think we need to pick the strategy to our use cases. Now we are using arrays for 
- **events**: it's rare the you have >4 events for a widget, but there are many widgets. Removing events is quite rare too. Here immediate shrink would be a better choice. 
-   **VGLite pending lists**: Usually small too (4, or 16) but they might change more frequently. cc @FASTSHIFT 

In light of that (for small lists) an absolute shrink value (e.g. 4) seems like a better fit.

onecoolx: > I think we need to pick the strategy to our use cases. Now we are using arrays for
> 
> * **events**: it's rare the you have >4 events for a widget, but there are many widgets. Removing events is quite rare too. Here immediate shrink would be a better choice.
> * **VGLite pending lists**: Usually small too (4, or 16) but they might change more frequently. cc @FASTSHIFT
> 
> In light of that (for small lists) an absolute shrink value (e.g. 4) seems like a better fit.

I think a ratio based posting rule would also be good for small lists, it would handle this more intelligently.

kisvegabor: This how it looks like when starting from 16 and keep reducing by 1.

Reduce is better at 12..9 but with smaller number ratio is better. Actually I'm convinced now that ratio is better in our case. Thank you for not giving up :slightly_smiling_face: 

| Size | Ratio (2) | Reduce (4) |
|------|-----------|------------|
| 16   | 16        | 16         |
| 15   | 16        | 16         |
| 14   | 16        | 16         |
| 13   | 16        | 16         |
| 12   | 16        | 12         |
| 11   | 16        | 12         |
| 10   | 16        | 12         |
| 9    | 16        | 12         |
| 8    | 8         | 8          |
| 7    | 8         | 8          |
| 6    | 8         | 8          |
| 5    | 8         | 8          |
| 4    | 4         | 4          |
| 3    | 4         | 4          |
| 2    | 2         | 4          |
| 1    | 1         | 4          |
| 0    | 0         | 0          |

kisvegabor (approved):

Issue #6389, opened 2024-06-18
NuttX + FB device (SAMA5D2) - rendering problem/tearing


TimJTi:
### LVGL version

9.1.0

### What happened?

Having moved from LVGL V8 to V9 I am seeing issues with rendering basic graphics to an 800x480 framebuffer using NuttX.

I have created a very simple screen with just a spinner, but the leading/trailing edges of the spinner "tear" with horizontal lines much of the time.

https://github.com/lvgl/lvgl/assets/56726697/2e24582b-11a1-49e2-8078-c3567bbabb77

I have played around with every rendering option I can find (via Kconfig) but it is stubbornly consistent, and occurs for ARGB888 and RGB565. I am at a loss and am thinking it might be a problem with the LVGL+NuttX framebuffer implementation?

Neither V8 nor V9 use double buffering as far as I can tell, but is looks like it is almost like the rendering being done to the buffer at the same time as it is being written to the LCD. Or

LVGL demos suffer the same issue, but they didn't with V8 so I don't think its my (custom) board nor the NuttX RTOS itself (as opposed to the NuttX LVGL port)?

I have run out of ideas but am very happy to try things - I really don't want to go back to LVGL V8 given all the function name changes etc as that is a PITA!!

TimJTi:
For completeness here is (nearly) the same spinner in LVGL V8.3.3:


https://github.com/lvgl/lvgl/assets/56726697/323e11e4-60b0-42d3-8a2f-47b7e655f85b

liyang5945:
check your  lv_port_disp_init function, has a lv_display_set_buffers... function call, wha't the last param? if LV_DISPLAY_RENDER_MODE_FULL or LV_DISPLAY_RENDER_MODE_DIRECT change it to LV_DISPLAY_RENDER_MODE_PARTIAL

jianglianfang:
It happend to me when I enbled CONFIG_FB_UPDATE with single framebuffer mode. U can try this https://github.com/lvgl/lvgl/pull/6397. it may help you.

TimJTi:
> check your lv_port_disp_init function, has a lv_display_set_buffers... function call, wha't the last param? if LV_DISPLAY_RENDER_MODE_FULL or LV_DISPLAY_RENDER_MODE_DIRECT change it to LV_DISPLAY_RENDER_MODE_PARTIAL

@liyang5945 - if I change this, the display does not update at all unfortunately. The spinner is drawn once but there's then no animation of it.

[...]

PR #6548, opened 2024-07-22
Fix artefacts caused by NuttX framebuffer driver

TimJTi:
I have found that the current NuttX framebuffer driver causes artefacts, as reported [here](https://github.com/lvgl/lvgl/issues/6389)

I won't profess to fully understand why, and although a few people tried to help, there was no magic fix.

What I believe is that:

- The existing driver allows the rendering to be DIRECT to the hardware framebuffer not to an intermediate buffer. But I am not 100% sure of that as I can find very little documentation that actually helps explain the difference between the various ways buffers are set up (`lv_display_set_buffers` vs `lv_display_set_draw_buffers` for example).
- Using the lvgl linux framebuufer driver as a reference I managed to fix the artefacts but was left with a different set of rendering artefacts and some areas (e·g. frame/cpu usage monitor) not updated. This, I found to be because the 'flush_cb' does an early exit if it isn't the last flush so, of course, not everything was rendered.

This PR is based on V9.1 release as "master" results in my custom board's display not working. I trfied a bisect but it failed to help me pinpoint why so I decided to stick at v9.1

It is quite probably I have completely missed the point of something coded in the original driver but I have a half hunch that there is a muddling of render buffers and display planes. Perhaps the buffer is writtedn to cope with a dual plane framebuufer but I find no examples of that back over at the NuttX repo.

I am happy to take on board any explanations, suggestions, whatever - but the bottom line is that this fraembuffer driver now works nicely on my board (SAMA5D7 + 800x480 32bpp TFT).

### DIFF
--- a/README.md
+++ b/README.md
@@ -102,7 +102,6 @@ LVGL is available as:
 - CMSIS-Pack
 - [RIOT OS package](https://doc.riot-os.org/group__pkg__lvgl.html#details)
 
-
 ## :robot: Examples
 
 See some examples of creating widgets, using layouts and applying styles. You will find C and MicroPython code, and links to try out or edit the examples in an online MicroPython editor.
--- a/docs/CHANGELOG.rst
+++ b/docs/CHANGELOG.rst
@@ -3,150 +3,596 @@
 Changelog
 =========
 
-v9.0.0
-~~~~~~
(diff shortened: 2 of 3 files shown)


TimJTi:
Might have a version issue here...will delete for now and return later!

PR #6549, opened 2024-07-22
feat(driver): import wayland driver

etag4048:


src/drivers/wayland/lv_wayland.h
```diff
@@ -0,0 +1,75 @@
...
+ *      TYPEDEFS
+ **********************/
+
+typedef bool (*lv_wayland_display_close_f_t)(lv_display_t * disp);
+
+/**********************
+ * GLOBAL PROTOTYPES
+ **********************/
+
+void lv_wayland_init(void);
```
kisvegabor: To docs is great and very detailed but please add a basic API docs here too.

src/drivers/wayland/lv_wayland_smm.h
```diff
@@ -0,0 +1,67 @@
+/**
+ * @file lv_wayland_smm.h
+ *
+ */
+#ifndef LV_WAYLAND_SMM_H
```
kisvegabor, line 5: Please follow `lv_templ.h`'s format and add `#if LV_USE_WAYLAND` in the H files too.

src/drivers/wayland/lv_wayland_smm.h
```diff
@@ -0,0 +1,67 @@
...
+#define LV_WAYLAND_SMM_H
+#include <stddef.h>
+#include <stdbool.h>
```
kisvegabor:
```suggestion
#include LV_STDDEF_INCLUDE
#include LV_STDBOOL_INCLUDE
```
to fix the uefi CI issue.

src/drivers/wayland/lv_wayland.h
```diff
@@ -0,0 +1,75 @@
...
+
+/*********************
+ *      INCLUDES
+ *********************/
+
+#include "../../display/lv_display.h"
+#include "../../indev/lv_indev.h"
+
+#if LV_USE_WAYLAND
+
```
kisvegabor, line 32:
```suggestion

```

kisvegabor (commented):
Thank you! This driver was requested many times already.

In order to make it built automatically by our CI please
- update [install-prerequisites.sh](https://github.com/lvgl/lvgl/blob/master/scripts/install-prerequisites.sh) and [install-prerequisites.bat](https://github.com/lvgl/lvgl/blob/master/scripts/install-prerequisites.bat)
- enable WAYLAND in [`lv_test_conf_full.h`](https://github.com/lvgl/lvgl/blob/master/tests/src/lv_test_conf_full.h#L121)

etag4048:
@kisvegabor If wayland is enabled in ``lv_test_conf_full.h``  an extra build step is required, it must run ``wayland-scanner`` to generate the header file and source file for the XDG shell. Does that mean adding an extra cmake file in ``env_support/cmake``  that is specific to wayland ? Also I don't see how it will be possible to build wayland on windows..

src/drivers/wayland/lv_wayland_smm.h
ghost: Thanks for the help, this one wasn't so obvious.

kisvegabor:
We can disable Wayland in case of Windows build by adding `#ifndef _WIN32`.

`wayland-scanner` is new to me. Could you add the required steps in the docs so that I can also understand it better and recommend a solution?

etag4048:
@kisvegabor No problems, will add ``#ifndef _WIN32`` and also document the steps to generate sources from the wayland protocol definitions.

etag4048:
@kisvegabor added a section on wayland-scanner, please see the CMakeLists.txt in this PR https://github.com/lvgl/lv_port_linux/pull/51/files

kisvegabor:
I see now thanks! Any change that we can add a template files for `wayland_xdg_shell.h/c` just to make it compile out of the box?

etag4048:
@kisvegabor This is problematic because not all distributions are on the same version of the protocol. If the version are mismatching it won't work. And will result in the following error(s) ``interface 'xdg_toplevel' has no event 3 ``

etag4048:
@kisvegabor Or do you mean stubs, so that the compiler is happy ?

kisvegabor:
> Or do you mean stubs, so that the compiler is happy ?

Yes, just to make the compiler happy. :slightly_smiling_face:

etag4048:
@kisvegabor I will see what I can do, I am worried that when a user will attempt to compile wayland for a legitimate use the stubs are going to get pulled in, and result in multiple definitions of symbols. But will investigate and get back to you.

kisvegabor:
Can we assume that the stubs will be overwritten?

etag4048:
@kisvegabor Could be a solution, if they are overwritten by the CMake script I showed you earlier.

kisvegabor:
I have an idea! Let's add the stubs to [here](https://github.com/lvgl/lvgl/tree/master/tests/src)  in a `test_wayland` folder, add an `#if LV_BUILD_TEST` guard to them, and also update the include paths [here](https://github.com/lvgl/lvgl/blob/master/tests/CMakeLists.txt#L267) if needed. 

This way a normal user will get an error for sure because these generated files are missing.

etag4048:
> I have an idea! Let's add the stubs to [here](https://github.com/lvgl/lvgl/tree/master/tests/src) in a `test_wayland` folder, add an `#if LV_BUILD_TEST` guard to them, and also update the include paths [here](https://github.com/lvgl/lvgl/blob/master/tests/CMakeLists.txt#L267) if needed.
> 
> This way a normal user will get an error for sure because these generated files are missing.

Sounds like a plan, that's much more elegant.

etag4048:
Nearly there, the compiler is pendantic

src/drivers/wayland/lv_wayland.h
```diff
@@ -0,0 +1,153 @@
...
+/**
+ * @brief Obtains the input device of the touch screen
+ * @note It is used to create an input group on application start
+ * @param disp Reference to the LVGL display associated to the window
+ * @return The input device
+ */
+lv_indev_t * lv_wayland_get_touchscreen(lv_display_t * disp);
+
+/**
+ * @brief Wrapper around lv_timer_handler
```
kisvegabor: Please remove the `@brief` tags for consistency with other docs
```suggestion
 * Wrapper around lv_timer_handler
```

kisvegabor (commented):
I hope you enjoyed the weekend and didn't think too much about work :slightly_smiling_face: 

Just a minor note, but other than that it looks good to me. 

It seems the stubs files to the test folder are not added yet.

src/drivers/wayland/lv_wayland.h
ghost: Ok will remove the ``@brief``, for the stubs since we added extra dependencies to ``install-prerequistes.sh`` I had an idea to generate those files anyways, they are quite large and might change and break the CI several times per year it will probably annoy everyone. So I've added the generation step to ``tests/main.py`` it generates the files and adds guards to them using the ``sed`` utility. (On the weekend, I prefer to avoid Cmake :) ) It's a weekend after all.

ghost: https://github.com/lvgl/lvgl/pull/6549/commits/bd39614dbdb6617e7efa44ffbbfac6f3b05b386f

kisvegabor: Nice! It's great that the test uses `wayland-scanner`.

etag4048:
@kisvegabor CI passes!

kisvegabor:
Fantastic! Just one more question: Would it be possible to eliminate the `memcpy` [here](https://github.com/lvgl/lvgl/pull/6549/files#diff-f340380b3d81957f0e512e3d197ac629819a563c203cfee5a280d7c73f6d2cb0R2176) by using FULL or DIRECT buffering mode and using Wayland's buffer(s) directly?

etag4048:
@kisvegabor yes, I am planning to remove the PARTIAL rendering mode, or make it optional. But this really needs to be done separately from the import in my opinion. It's a big undertaking and a re-write is necessary.

kisvegabor (dismissed):
Okay, let's proceed gradually! 

Thank you Erik!

.gitignore
```diff
@@ -24,6 +24,7 @@ test_screenshot_error.h
 build/
 tests/build_*/
 tests/report/
+tests/wayland_protocols/
```
liamHowatt, line 27:
```suggestion
```

docs/integration/driver/wayland.rst
```diff
@@ -0,0 +1,182 @@
...
+
+.. code:: c
+
+   #define LV_WAYLAND_WINDOW_DECORATIONS 1
+
+- Enable support for the deprecated 'wl_shell', Only useful when the BSP on the target has weston ``9.x``
+
+.. code:: c
+
+   #define LV_WAYLAND_WL_SHELL  1 /*preferred default*/
```
liamHowatt: Is it actually the preferred default?

docs/integration/driver/wayland.rst
```diff
@@ -0,0 +1,182 @@
...
+
+Usage
+-----
+
+#. In ``main.c`` ``#incude "lv_drivers/wayland/wayland.h"``
+#. Enable the Wayland driver in ``lv_conf.h`` with ``LV_USE_WAYLAND 1``
+
+#. ``LV_COLOR_DEPTH`` should be set either to ``32`` or ``16`` in ``lv_conf.h``
+
+#. After ``lv_init()`` call ``lv_wayland_init()``.
```
liamHowatt: Would it make sense to have `lv_wayland_init()` happen in lv_init.c when `LV_USE_WAYLAND` is `1`?

src/drivers/wayland/lv_wayland.h
```diff
@@ -0,0 +1,153 @@
...
+                                        lv_wayland_display_close_f_t close_cb);
+
+/**
+ * Closes the window programmatically
+ * @param disp Reference to the LVGL display associated to the window
+ */
+void lv_wayland_window_close(lv_display_t * disp);
+
+/**
+ * Closes the window programmatically
```
liamHowatt: I think you accidentally forgot to update this

src/drivers/wayland/lv_wayland.h
```diff
@@ -0,0 +1,153 @@
...
+int lv_wayland_get_fd(void);
+
+/**
+ * Creates a window
+ * @param hor_res The width of the window in pixels
+ * @param ver_res The height of the window in pixels
+ * @param title The title of the window
+ * @param close_cb The callback that will be execute when the user closes the window
+ */
+lv_display_t * lv_wayland_window_create(uint32_t hor_res, uint32_t ver_res, char * title,
```
liamHowatt: Please add a `@return`

src/drivers/wayland/lv_wayland.h
```diff
@@ -0,0 +1,153 @@
+/*******************************************************************
+ *
+ * lv_wayland.h - Public functions of the LVGL Wayland client
+ *
+ * Based on the original file from the repository.
+ *
+ * Porting to LVGL 9.1
+ * Copyright (c) 2024 EDGEMTech Ltd.
+ *
+ * See LICENSE.txt for details
```
liamHowatt: I don't think LVGL has a LICENCE.txt. Should there be a link to a EDGEMTech repo?

liamHowatt (commented):
Nice detailed docs!

I didn't look too in-depth into lv_wayland.c

.gitignore
ghost: This entry in the gitignore file is necessary, the ``tests/wayland_protocols`` directory contains generated code for wayland's XDG shell. it's generated by ``tests/main.py`` those generated files should not be added to the index.

docs/integration/driver/wayland.rst
ghost: Good catch, nope it's not it's left over from the old docs.

docs/integration/driver/wayland.rst
ghost: Done, it totally makes sense if the other drivers are doing it this way

src/drivers/wayland/lv_wayland.h
ghost: It has it, but the spelling is different. https://github.com/lvgl/lvgl/blob/master/LICENCE.txt

src/drivers/wayland/lv_wayland.h
ghost: Good catch

src/drivers/wayland/lv_wayland.h
kisvegabor: I'm not entirely sure what it means to have a copyright in an MIT licensed repo, where the copyright refers to a company other than the actual owner. What are the practical implications of having this copyright?

ghost: @kisvegabor EDGEMTech directives, if a change is made to a file the header must explain the change(s) and then mention the author of the changes, company + email. This change is Copyright 2024 but under the MIT license. This common practice in the Linux kernel or Zephyr. https://github.com/torvalds/linux/blob/master/kernel/irq/manage.c

kisvegabor: Hm, okay then.

liamHowatt: > It has it, but the spelling is different. https://github.com/lvgl/lvgl/blob/master/LICENCE.txt

Totally missed that

docs/integration/driver/wayland.rst
liamHowatt: Thanks! For consistency with the new patterns could you please move the function prototypes for `lv_wayland_init` and `lv_wayland_deinit` to a new file `lv_wayland_private.h` (and include that header in `lv_init.c` and `lv_wayland.c`)? Either that, or perform the init on the first call to `lv_wayland_window_create` if you prefer.

kisvegabor:
@etag4048 
We will release LVGL v9.2 in one and a half weeks. It would be great to add the Wayland driver in it too. Do you have time to finalize it this week? From my side, only the above mentioned mostly formatting related comments should be resolved.

docs/integration/driver/wayland.rst
ghost: @liamHowatt Ok, moved init/deinit declarations to a private header

etag4048:
> @etag4048 We will release LVGL v9.2 in one and a half weeks. It would be great to add the Wayland driver in it too. Do you have time to finalize it this week? From my side, only the above mentioned mostly formatting related comments should be resolved.

I've removed some ``@brief`` from comments, been careful to not add extra new lines in headers.

docs/integration/driver/wayland.rst
jefdriesen: Calling the `lv_wayland_init()` function directly in the `lv_init()` function removes the ability to compile lvgl with multiple drivers enabled (e.g. sdl, fbdev, drm and wayland) in the config file:
```
#define LV_USE_SDL 1
#define LV_USE_LINUX_FBDEV 1
#define LV_USE_LINUX_DRM 1
#define LV_USE_WAYLAND 1
```
And then let the application choose the driver by calling the corresponding function to create the window (e.g. `lv_sdl_window_create`, `lv_linux_fbdev_create`, `lv_linux_drm_create` or `lv_wayland_window_create`).

This can be avoided by either requiring the developer to the call the initialization function manually again, or letting the window creation function take care of calling it internally.

ghost: @jefdriesen This is a valid point,  I looked at ``lv_init.c`` and only the windows driver is initialized from there. FBDEV, SDL and DRM aren't. So I will move the call to ``lv_init`` into  ``lv_wayland_window_create``

kisvegabor (dismissed):
Thank you Erik! 

I really would like to have Wayland part of v9.2 so I approve it now. 

`lv_wayland_smm.c/h` are mysteries to me. Are these written by us? If so they should follow `lv_templ.c/h` format.  Anyway, these are not part of the public API and we can improve them later.

liamHowatt:
Ignore the mention. I linked the wrong issue/PR number.

etag4048:
> Thank you Erik!
> 
> I really would like to have Wayland part of v9 so I approve it now.
> 
> `lv_wayland_smm.c/h` are mysteries to me. Are these written by us? If so they should follow `lv_templ.c/h` format. Anyway, these are not part of the public API and we can improve them later.

Hey @kisvegabor, I've added the missing comment blocks to ``lv_wayland_smm.h`` The SMM code was added by a previous contributor. It's an abstraction layer on top of ``mmap(2)`` ``munmap(2)`` and ``ftruncate(2)`` libc shared memory manipulation functions. Shared memory is used to pass the framebuffer to the wayland compositor. And it's internal to the wayland driver. 

@jefdriesen @liamHowatt I've also made wayland_init and deinit internal to the driver.

There is however a curious crash with the current HEAD... 

```
Process 1081838 launched: '/home/etag/lvgl/src/test/lv_port_linux/bin/lvglsim' (x86_64)
[Warn]	(262287.362, +262287362)	 lv_draw_buf_init: Data size too small, required: 3600, provided: 0 lv_draw_buf.c:245
Process 1081838 stopped
* thread #1, name = 'lvglsim', stop reason = signal SIGFPE: integer divide by zero
    frame #0: 0x00005555555d5c50 lvglsim`img_draw_core(draw_unit=0x0000555555656d30, draw_dsc=0x0000555555679310, decoder_dsc=0x00007fffffffc360, sup=0x00007fffffffc2d0, img_coords=0x00005555556f194c, clipped_img_area=0x00007fffffffc350) at lv_draw_sw_img.c:318:34
   315 	        }
```

Will investigate when I get back.. It seems like it fails to allocate a buffer, the client is using LIBC for memory allocation.

jefdriesen:
I also noticed the "Data size too small" problem, but it appears to be fixed with commit 78d29d28531ccd1e5e9f90c3764e3500a1ea553f. For me the demo application didn't crash, but became unresponsive instead.

kisvegabor:
Thank you Erik!

Which demo do you use? @jefdriesen which driver do you use?

I can try reproducing it in the afternoon (in 8 hours)

jefdriesen:
I used the built-in lvgl widget demo. The problem shows up in the "Analytics" tab when scrolling to the circular graphs at the bottom. I could reproduce it with all three 3 linux drivers (fbdev, drm and wayland). After applying commit 78d29d28531ccd1e5e9f90c3764e3500a1ea553f I can no longer reproduce it.

kisvegabor:
It seems the issue is resolved than. Thank you for reporting and looking into it.

kisvegabor (approved):

docs/integration/driver/wayland.rst
liamHowatt: The way it is now looks good. I think maybe my suggestion to do it it lv_init.c was bad advice.

Thanks!

Just a question: it says multiple window support is experimental for now. Do you know what happens if a second window is created? Will it re-initialize Wayland and put the first window into an erroneous state?

liamHowatt (approved):
Wondering if instead of a new timer handler, the driver could just handle the `LV_EVENT_REFR_START` event to run the actions that come before `lv_timer_handler` and also handle the `LV_EVENT_REFR_READY` event for actions to run after `lv_timer_handler`.

It's working now so I am approving.

kisvegabor:
Let's fine tune in a new PR. :slightly_smiling_face: 

Thank you Erik!

PR #6561, opened 2024-07-23
fix(nuttx): fix framebuffer driver render artefacts

TimJTi:
### Fix Artefacts caused by NuttX framebuffer driver that currently allows rendering direct to hardware framebuffer

I found I was getting artefacts using LVGL V9 with a SAMA5D27 processor using the Nuttx framebuffer driver. This is becuase the LVGL driver passes the hardware MMAP'd framebuffer address as the address of the render buffer, whereas is should - I believe - be a MALLOC's local buffer, withe th flush routine copying the color_p buffer to the hardware framebuffer in flush_cb.

This amended code works, with no render artefacts, and a notable performance increase.

The driver (still) has a function `fbdev_init_mem2` the purpose of which is unclear to me. I seems perhaps to be related to a second video plane of some NuttX framebuffer device driver but I can't work it out.  I am wondering if it is an erroneous attempt at the usual LVGL double render buffer method? 

I have put code in for a second render buffer, but am not 100% sure if that "conflicts" in anyway with the purpose of `fbdev_init_mem2`? With Kconfig setup too.

My first LVGL commit - be kind, but I am expecting feedback and requests for change!!!

### DIFF
--- a/Kconfig
+++ b/Kconfig
@@ -1625,7 +1625,7 @@ menu "LVGL configuration"
 
 		choice
 			prompt "NuttX LCD buffer size"
-			depends on LV_USE_NUTTX_LCD
+			depends on LV_USE_NUTTX_LCD && LV_USE_NUTTX
 			default LV_NUTTX_LCD_SINGLE_BUFFER
 
 			config LV_NUTTX_LCD_SINGLE_BUFFER
@@ -1641,15 +1641,39 @@ menu "LVGL configuration"
 
 		config LV_NUTTX_LCD_BUFFER_COUNT
 			int
-			depends on LV_USE_NUTTX_LCD
+			depends on LV_USE_NUTTX_LCD && LV_USE_NUTTX
 			default 0 if LV_NUTTX_LCD_CUSTOM_BUFFER
 			default 1 if LV_NUTTX_LCD_SINGLE_BUFFER
 			default 2 if LV_NUTTX_LCD_DOUBLE_BUFFER
 
 		config LV_NUTTX_LCD_BUFFER_SIZE
 			int "Custom partial buffer size (in number of rows)"
-			depends on LV_USE_NUTTX_LCD && LV_NUTTX_LCD_CUSTOM_BUFFER
+			depends on LV_USE_NUTTX_LCD && LV_NUTTX_LCD_CUSTOM_BUFFER && LV_USE_NUTTX
 			default 60
+
+		config LV_USE_NUTTX_FBDEV
+			bool "Use NuttX LCD Framebuffer"
+			depends on LV_USE_NUTTX
+			default n
+
+		choice
+			prompt "Framebuffer size"
+			depends on LV_USE_NUTTX && LV_USE_NUTTX_FBDEV
+			default LV_NUTTX_FBDEV_SINGLE_BUFFER
+
+			config LV_NUTTX_FBDEV_SINGLE_BUFFER
+				bool "One screen-sized buffer"
+
+			config LV_NUTTX_FBDEV_DOUBLE_BUFFER
+				bool "Two screen-sized buffer"
+
+		endchoice
+
+		config LV_NUTTX_FBDEV_BUFFER_COUNT
+			int
+			depends on LV_USE_NUTTX_FBDEV
+			default 1 if LV_NUTTX_FBDEV_SINGLE_BUFFER
+			default 2 if LV_NUTTX_FBDEV_DOUBLE_BUFFER
 
 		config LV_USE_NUTTX_TOUCHSCREEN
 			bool "Use NuttX touchscreen driver"
--- a/src/drivers/nuttx/lv_nuttx_fbdev.c
+++ b/src/drivers/nuttx/lv_nuttx_fbdev.c
@@ -139,20 +139,35 @@ int lv_nuttx_fbdev_set_file(lv_display_t * disp, const char * file)
     uint32_t h = dsc->vinfo.yres;
     uint32_t stride = dsc->pinfo.stride;
     uint32_t data_size = h * stride;
-    lv_draw_buf_init(&dsc->buf1, w, h, color_format, stride, dsc->mem, data_size);
+    uint8_t * buf1 = NULL;
+    uint8_t * buf2 = NULL;
+
+    buf1 = malloc(data_size);
+    if(buf1 == NULL) {
+        LV_LOG_ERROR("display draw buffer malloc failed");
+        goto errout;
+    }
+
+    lv_draw_buf_init(&dsc->buf1, w, h, color_format, stride, buf1, data_size);
+
+    if(LV_NUTTX_FBDEV_BUFFER_COUNT == 2) {
+        buf2 = malloc(data_size);
+        if(buf2 == NULL) {
+            LV_LOG_ERROR("display draw buffer2 malloc failed");
+        }
+        else {
+            lv_draw_buf_init(&dsc->buf2, w, h, color_format, stride, buf2, data_size);
+        }
+    }
 
     /* double buffer mode */
 
     bool double_buffer = dsc->pinfo.yres_virtual == (dsc->vinfo.yres * 2);
     if(double_buffer) {
-        if((ret = fbdev_init_mem2(dsc)) < 0) {
-            goto errout;
-        }
-
-        lv_draw_buf_init(&dsc->buf2, w, h, color_format, stride, dsc->mem2, data_size);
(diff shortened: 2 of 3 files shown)


TimJTi:
This relates to [this](https://github.com/lvgl/lvgl/issues/6389) issue

liamHowatt:
Thanks for taking the initiative and digging for a good solution for some time. I see there's a lot that's led up to it.

The idea of using the hardware mapped framebuffer address as the direct draw buffer is that extra copying can be avoided. Some of the logic such as `FBIO_UPDATE` and `FBIOPAN_DISPLAY` (vsync) only make sense when the draw buffer is the framebuffer.

Before your double buffer config, it used to check if the underlying fb driver was double buffered to determine whether it should be double buffered. Configs like [this](https://github.com/apache/nuttx/blob/a83a569652da88b73f04c4cc1c33879ee72476f9/arch/xtensa/src/esp32s3/Kconfig#L2340) configure drivers like [this](https://github.com/apache/nuttx/blob/master/arch/xtensa/src/esp32s3/esp32s3_lcd.c).

I think what could be done is to add a config and logic for double+single buffered `PARTIAL` mode and your buffer copying code can be used in that situation.

TimJTi:
> Thanks for taking the initiative and digging for a good solution for some time. I see there's a lot that's led up to it.
> 
> The idea of using the hardware mapped framebuffer address as the direct draw buffer is that extra copying can be avoided. Some of the logic such as `FBIO_UPDATE` and `FBIOPAN_DISPLAY` (vsync) only make sense when the draw buffer is the framebuffer.
> 
> Before your double buffer config, it used to check if the underlying fb driver was double buffered to determine whether it should be double buffered. Configs like [this](https://github.com/apache/nuttx/blob/a83a569652da88b73f04c4cc1c33879ee72476f9/arch/xtensa/src/esp32s3/Kconfig#L2340) configure drivers like [this](https://github.com/apache/nuttx/blob/master/arch/xtensa/src/esp32s3/esp32s3_lcd.c).
> 
> I think what could be done is to add a config and logic for double+single buffered `PARTIAL` mode and your buffer copying code can be used in that situation.

Hi @liamHowatt - thank you for taking a look and providing much needed feedback!

I only really worked on a single buffer solution so the double buffer was just en embellishment. The proble was very simply that if the multiple calls to flush_cb are done direct to the HW framebuffer we get artefacts. It can only be flushed - in this case - once all calls are done. I played around with check if it was the last call to flush, but couldn't get it to behave.

I'll take a look at the configs etc - thanks! - and see if I can extend the logic to cover more options but, bottom line, my particular processor - whether double or single buffered - has to use intermediate buffering.

liamHowatt:
I'm not sure if this is helpful but I had to make these https://github.com/liamHowatt/nuttx/commit/3de20f7d09f665875777f4efa5ad2ffd591695c8 changes to this ESP driver to add synchronized double buffering to get rid of the inevitable artifacts of single buffering. (I think artifact-free synchronized single buffering is possible but it was just easier to add double buffering for me)

FASTSHIFT:
Letting LVGL go directly to the framebuffer is to improve rendering efficiency, because this can reduce buffer copying, but this usually requires the display to support switching the framebuffer of the previous and next frames. However, this method does not support the driver of the display that only refreshes a single frame in video mode, because the screen must transfer the content of the video memory to the display at a certain period (usually 60Hz), which will cause the unfinished drawing to be displayed and cause flickering.

There are two ways to solve this problem:
1. Modify the screen display driver to support dual buffer switching, which can bring the best display efficiency.
2. Add a configuration to let the fbdev interface layer apply for an independent drawing buffer and copy it every time flush_cb, which is suitable for the driver of a single frame framebuffer

FASTSHIFT:
`fbdev_init_mem2` is designed to solve the problem that the device cannot provide two framebuffers with consecutive addresses due to SRAM space limitations. LVGL may try to request the driver for the second frame framebuffer with non-consecutive addresses, and then calculate `yoffset` based on the offset of the two addresses for switching.

TimJTi:
> Letting LVGL go directly to the framebuffer is to improve rendering efficiency, because this can reduce buffer copying, but this usually requires the display to support switching the framebuffer of the previous and next frames. However, this method does not support the driver of the display that only refreshes a single frame in video mode, because the screen must transfer the content of the video memory to the display at a certain period (usually 60Hz), which will cause the unfinished drawing to be displayed and cause flickering.
> 
> There are two ways to solve this problem:
> 
> 1. Modify the screen display driver to support dual buffer switching, which can bring the best display efficiency.
> 2. Add a configuration to let the fbdev interface layer apply for an independent drawing buffer and copy it every time flush_cb, which is suitable for the driver of a single frame framebuffer.

I did try adding a second buffer to the NuttX driver for the SAMA5D2 family but couldn't make it work - I'm sure it could be and I just did something wrong. But...

...if we want to keep portability in mind, LVGL probably shouldn't be relying on the OS to deal with the fact that only partially completed renders to the actual /dev/fb0 device will be put on the LCD. Maybe the principle should be "What you write to the framebuffer is what you see". In my case, what you see is partial-render artefacts.

The way I see it is that partial renders should **_never_** be copied to the OS driver layer but that is what the current LVGL NuttX FB driver does. It just seems wrong.

But if there are processors or whatever that can work like that, and it's efficient, and you want to keep it, then it can be made an LVGL option via Kconfig? I can rework my code to include this, perhaps called `DIRECT_FB_WRITE` or something?

Or I just add the usual LVGL partial buffer mode as well and then it is all as-expected and flexible, following the traditional LVGL methodologies?

FASTSHIFT:
> But if there are processors or whatever that can work like that, and it's efficient, and you want to keep it, then it can be made an LVGL option via Kconfig? I can rework my code to include this, perhaps called `DIRECT_FB_WRITE` or something?

I agree to add a Kconfig option to choose partial mode or direct mode. By default, direct mode is selected to ensure forward compatibility.

TimJTi:
Perhaps the answer is obvious (except to a newcomer i.e. me!!):

- `LV_DISPLAY_RENDER_MODE_PARTIAL` is not there at present but could be added.
- `LV_DISPLAY_RENDER_MODE_DIRECT` is what was there and is what I have changed. I should "revert" it and use the old method for this mode
- `LV_DISPLAY_RENDER_MODE_FULL` is actually what I have implemented but under the wrong name

Is that right?

TimJTi:
> * LV_DISPLAY_RENDER_MODE_FULL

Umm...no, it's not right...should read the descriptions more thoroughly and in less of a rush. `LV_DISPLAY_RENDER_MODE_FULL` will always redraw everything, every pixel, which will kill performance. Not advised IMHO.

So it's a variant of `LV_DISPLAY_RENDER_MODE_DIRECT` and we need two modes:

My requirement could be sorted using `LV_DISPLAY_RENDER_MODE_PARTIAL` since there's no reason not to configure  a full sized frame buffer (or 2)?

Right?

If so I can re-look next week and will add some notes in Kconfig help to explain the potential pitfalls of the DIRECT method as it writes direct to the hardware framebuffer so may suffer artefacts.

TimJTi:
Done a little more digging on `FBIOPAN_DISPLAY` and I think it is a misuse of it for the generic `LV_DISPLAY_RENDER_MODE_DIRECT`. I believe my rework is the correct, default, method for this.

I suspect it might be "goldfish" specific?

The "NuttX" framebuffer support here is a non-standard case and should not be the default, in my opinion.

Can we come up with an agreed KConfig/LV_???? name for the special case and I will revert the old code, add it as an option that can be enabled if required, and leave mine as the generic.

Is that OK?

FASTSHIFT:
I have pushed an implementation I envisioned: https://github.com/lvgl/lvgl/pull/6580. Can you help test it?
It can automatically detect single/double buffering, no need to add additional configuration options, and is forward compatible.

TimJTi:
> @TimJTi I have pushed an implementation I envisioned: #6580. Can you help test it? It can automatically detect single/double buffering, no need to add additional configuration options, and is forward compatible.

I will try it for sure - thanks! Busy at the moment but will have time next week and will report back. Obviously like the idea of anything automatic :-)

TimJTi:
@FASTSHIFT I tested just now and [your fix](https://github.com/lvgl/lvgl/pull/6580) is good. Thanks.
Might revist this and look at adding a partial redraw feature as I now have 2 framebuffers taking up 3Mbytes of memory when only 1 and a smaller draw buffer might be all it needs - but I am not short of memory at the moment.

TimJTi:
I have just restored the branch for this PR as a temporary measure - I am seeing very poor performance from the nuttx framebuffer driver, and want to see what I did last year to give me some clues!

PR #6562, opened 2024-07-23
fix(kconfig): Do not set LV_CONF_SKIP by default

faxe1008:
Due to the way that Kconfig symbols work, this can not be turned off if it is defaulted to true.

While rebasing the Zephyr integration I encountered the issue that the `lv_conf.h` that is used to point to the `malloc_core` etc. functions is not included correctly in the respective headers. Investigating the issue shows that since this symbol is defined with `default y` I can not turn it  off from my "third party" code. Having an unconditional `default y` in Kconfig essentially always enables the option since it can not be turned of. This should be avoided.

### Notes

### DIFF
--- a/docs/porting/project.rst
+++ b/docs/porting/project.rst
@@ -90,7 +90,8 @@ will attempt to include ``lv_conf.h`` simply with
 
 You can even use a different name for ``lv_conf.h``. The custom path can
 be set via the :c:macro:`LV_CONF_PATH` define. For example
-``-DLV_CONF_PATH="/home/joe/my_project/my_custom_conf.h"``
+``-DLV_CONF_PATH="/home/joe/my_project/my_custom_conf.h"``. If this define
+is set :c:macro:`LV_CONF_SKIP` is assumed to be ``0``.
 
 If :c:macro:`LV_CONF_SKIP` is defined, LVGL will not try to include
 ``lv_conf.h``. Instead you can pass the config defines using build
--- a/scripts/lv_conf_internal_gen.py
+++ b/scripts/lv_conf_internal_gen.py
@@ -69,7 +69,7 @@ fout.write(
 #endif
 
 /*If lv_conf.h is not skipped include it*/
-#ifndef LV_CONF_SKIP
+#if !defined(LV_CONF_SKIP) || defined(LV_CONF_PATH)
     #ifdef LV_CONF_PATH                           /*If there is a path defined for lv_conf.h use it*/
         #define __LV_TO_STR_AUX(x) #x
         #define __LV_TO_STR(x) __LV_TO_STR_AUX(x)
--- a/src/lv_conf_internal.h
+++ b/src/lv_conf_internal.h
@@ -47,7 +47,7 @@
 #endif
 
 /*If lv_conf.h is not skipped include it*/
-#ifndef LV_CONF_SKIP
+#if !defined(LV_CONF_SKIP) || defined(LV_CONF_PATH)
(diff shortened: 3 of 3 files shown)


FASTSHIFT:
For custom memory allocation functions, you only need to implement `lv_malloc_core(size_t size)` externally and configure `LV_USE_CUSTOM_MALLOC` to y in menuconfig. Why do we need `lv_conf.h`?

faxe1008:
I need to set `LV_STDLIB_INCLUDE` because otherwise the lv_malloc_core symbol can not be found. Also I need to set a custom OSAL implementation. I thought that this was the purpose of `lv_conf.h` in the first place?

kisvegabor:
I see the problem, however if we disable `LV_CONF_SKIP` by default, all platforms need to add at least an empty `lv_conf.h` file.

Can we workaround it by making it a `choice` similar to [LV_COLOR_DEPTH](https://github.com/lvgl/lvgl/blob/master/Kconfig#L16-L40)?

Having a Zephyr stdlib and OSAL interface built-in to LVGL would also solve the issue! :slightly_smiling_face:  See [RT-Thread](https://github.com/lvgl/lvgl/tree/master/src/stdlib/rtthread) as an example.

faxe1008:
> I see the problem, however if we disable `LV_CONF_SKIP` by default, all platforms need to add at least an empty `lv_conf.h` file.
> 
> Can we workaround it by making it a `choice` similar to [LV_COLOR_DEPTH](https://github.com/lvgl/lvgl/blob/master/Kconfig#L16-L40)?
> 
> Having a Zephyr stdlib and OSAL interface built-in to LVGL would also solve the issue! 🙂 See [RT-Thread](https://github.com/lvgl/lvgl/tree/master/src/stdlib/rtthread) as an example.

Unfortunately this will not work with a choice sybol either. Roughly speaking this is the build process on Zephyr side:
```
1. Parse LVGLs Kconfig file
2. Parse the Kconfig file of Zephyr: Here we have additional options regarding memory management, additional shells etc.
3. Parse the users overlay file and overwrite
```
Overwriting any defaulted value in step 2. does not work as for any Kconfig symbol with a default value the first occurence of the declaration will determine the value (which is set in step 1 by LVGLs Kconfig). 

You can in theory overwrite this in the overlay file which can be provided by the user. But its kind of odd to push this responsibility to the user to enable a config that "configures the build" and needs to be unset every time no matter the project really.

Anyways if you are not ok with removing the default I found a solution that works and should not break any other users of Kconfig:
```
diff --git a/src/lv_conf_internal.h b/src/lv_conf_internal.h
index 2c1986092..6336284eb 100644
--- a/src/lv_conf_internal.h
+++ b/src/lv_conf_internal.h
@@ -47,7 +47,7 @@
 #endif

 /*If lv_conf.h is not skipped include it*/
-#ifndef LV_CONF_SKIP
+#if !defined(LV_CONF_SKIP) || defined(LV_CONF_PATH)
     #ifdef LV_CONF_PATH                           /*If there is a path defined for lv_conf.h use it*/
         #define __LV_TO_STR_AUX(x) #x
         #define __LV_TO_STR(x) __LV_TO_STR_AUX(x)
```

Effectively assume that the conf should not be skipped if `LV_CONF_PATH` has a value. Since `LV_CONF_PATH` is not controlled via Kconfig Zephyr can simply use a preprocessor define for this. If you think about it, it also makes sense semantically since why would you set this macro if you do not intend to use an "outside" lv_conf.h?

What do you think @kisvegabor ?

kisvegabor (approved):
Great!

FASTSHIFT (approved):

PR #6575, opened 2024-07-25
feat(label): add property

XuNeo:

### Notes

### DIFF
--- a/src/core/lv_obj_property.h
+++ b/src/core/lv_obj_property.h
@@ -62,6 +62,7 @@ enum {
     /*Define the property ID for every widget here. */
     LV_PROPERTY_OBJ_START       = 0x0100, /* lv_obj.c */
     LV_PROPERTY_IMAGE_START     = 0x0200, /* lv_image.c */
+    LV_PROPERTY_LABEL_START     = 0x0300, /* lv_label.c */
 
     /*Special ID, use it to extend ID and make sure it's unique and compile time determinant*/
     LV_PROPERTY_ID_BUILTIN_LAST = 0xffff, /*ID of 0x10000 ~ 0xfffffff is reserved for user*/
--- a/src/widgets/label/lv_label.c
+++ b/src/widgets/label/lv_label.c
@@ -60,6 +60,31 @@ static void calculate_x_coordinate(int32_t * x, const lv_text_align_t align, con
 /**********************
  *  STATIC VARIABLES
  **********************/
+#if LV_USE_OBJ_PROPERTY
+static const lv_property_ops_t properties[] = {
+    {
+        .id = LV_PROPERTY_LABEL_TEXT,
+        .setter = lv_label_set_text,
+        .getter = lv_label_get_text,
+    },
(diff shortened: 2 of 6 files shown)


src/core/lv_obj_property.h
```diff
@@ -56,15 +56,16 @@ enum {
     LV_PROPERTY_ID_INVALID      = 0,
 
     /*ID 0x01 to 0xff are style ID, check lv_style_prop_t*/
```
W-Mai:
```suggestion
    /*ID 0x00 to 0xff are style ID, check lv_style_prop_t*/
```

W-Mai (changes requested):

src/core/lv_obj_property.h
XuNeo: It's actually `0x01`. See
```c
enum _lv_style_prop_t {
    LV_STYLE_PROP_INV               = 0,

    /*Group 0*/
    LV_STYLE_WIDTH                  = 1,
```

We use `LV_PROPERTY_STYLE_START` 0 as enum start. It's a bit counter intuitive.

kisvegabor (approved):
Great, thanks!

I wonder if we can generate these automatically somehow. I thin for most of the properties it should work, but I don't know how to handle 
- objects created by `lv_*_add_*` 
  - chart series
  - scale section
  - messagebox title, footer, etc (probably ok, as these are `lv_obj`s)
  - tabview tabs (`lv_obj`s too)
  - window buttons and titles (`lv_obj`s too)
- more complex setters 
  - buttonmatrix maps 
  - line points
  - table cell data
  - spans

liamHowatt (approved):

XuNeo:
> I wonder if we can generate these automatically somehow. I thin for most of the properties it should work, but I don't know how to handle

The property can only work with simply value. If it's too complex, like an array of points, pointers, it should be handled case by case.

> more complex setters

The original purpose of property API is to unify and simplify binding language. If it's not unified, then it's not achievable through property API. 

E.g.
```c
void lv_roller_set_options(lv_obj_t * obj, const char * options, lv_roller_mode_t mode);
void lv_roller_set_selected(lv_obj_t * obj, uint32_t sel_opt, lv_anim_enable_t anim);

```

kisvegabor:
For the editor we should have a parseable and simple API for all the widgets which maps to properties in a declarative language. So ideally each 
- create function should have a single argument: the parent
- setter should have a two argument: the target and the value

Still considering if it's possible in all cases.

XuNeo:
For luavgl, I provide a method to override LVGL's default property API, extend those special cases.

Given a provided property name

1. Try luavgl's widget property table
2. Try lvgl's property table
3. Try lvgl's obj style.

https://github.com/XuNeo/luavgl/pull/50/commits/2c3fc0ef32e46ff6a577c5170058e615a6f3db4c#diff-7e3a522db40999cfa49a42764aef9413eeac47ecaf8e21d5f02582c09820b9dbR716-R750

PR #6583, opened 2024-07-26
Update QNX documentation as the build directory changed

elahav:

The last commit in #6507 changed moved the QNX build directory under `env_support`. Update the documentation to reflect that.

### Notes

### DIFF
--- a/docs/integration/os/qnx.rst
+++ b/docs/integration/os/qnx.rst
@@ -35,7 +35,7 @@ architectures. To build all libraries, simply invoke `make` in this directory:
 
 .. code:: shell
 
-    # cd $(LVGL_ROOT)/qnx
+    # cd $(LVGL_ROOT)/env_support/qnx
     # make
 
 If you prefer to build for a specific architecture and variant, go to the
@@ -44,15 +44,16 @@ library for ARMv8:
 
 .. code:: shell
 
-    # cd $(LVGL_ROOT)/qnx/aarch64/so.le
+    # cd $(LVGL_ROOT)/env_support/qnx/aarch64/so.le
     # make
 
 As a general rule, if you only want to have one LVGL application in your system
 then it is better to use a static library. If you have more than one, and
 especially if they run concurrently, it is better to use the shared library.
 
-Before building the library, you may wish to edit `$(LVGL_ROOT)/qnx/lv_conf.h`,
-e.g. to enable double-buffering.
+Before building the library, you may wish to edit
+`$(LVGL_ROOT)/env_support/qnx/lv_conf.h`, e.g. to add fonts or disable
+double-buffering.
 
 Writing a LVGL Application
 ~~~~~~~~~~~~~~~~~~~~~~~~~~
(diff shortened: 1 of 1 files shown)


jarxon:
Looks good to me.  I don't think I have approval permissions though.

kisvegabor:
Thanks!

I've just noticed that a new [`lv_conf.h`](https://github.com/lvgl/lvgl/blob/master/env_support/qnx/lv_conf.h) file was added in the qnx folder. We should remove that if possible. The recommended workflow is that the user copies `lv_conf_template.h` next to the `lvgl` folder or to the path specified by `LV_CONF_PATH`. There there shouldn't be "ready-to-use" config files in the driver.

An other note: on LVGL v9 have an operating system abstraction layer for some OS-es. Would it make sense to add it for QNX too? See an example for [pthread](https://github.com/lvgl/lvgl/blob/master/src/osal/lv_pthread.c).

elahav:
@kisvegabor the reason I added the `lv_conf.h` file is that the workflow I enabled via the build directory is slightly different than the normal one. You can still follow the standard LVGL way of adding the source to your project, setting up your own `lv_conf.h` file and building the application and LVGL as one executable. The build directory, on the other hand, enables a ready-to-use library (static and shared), which is why it needs a pre-populated conf header.

If that doesn't make sense or doesn't fit with the way LVGL does things I can certainly revisit.

kisvegabor:
Having a dedicated `lv_conf.h` makes maintenance more difficult as we need to continuously update that `lv_conf.h`. If people would like to build a standalone library they can still copy `lv_conf_template.h` next to the `lvgl` folder and set it up themselves. 
Could it really work?

elahav:
That's a very good point about maintenance. I've created a new pull request #6596 to generate the header at build time from the template. What do you think?
I will update the documentation to make it clear that you can use LVGL for a QNX project in two ways:
1. The standard LVGL way of including the source with your project
2. As a library

elahav:
> An other note: on LVGL v9 have an operating system abstraction layer for some OS-es. Would it make sense to add it for QNX too? See an example for [pthread](https://github.com/lvgl/lvgl/blob/master/src/osal/lv_pthread.c).

`LV_OS_PTHREAD` would work on QNX

kisvegabor (dismissed):
Looks good, thank you!

liamHowatt (commented):
I think I spotted one more near the bottom
```make
LDFLAGS=-lscreen -llvgl -L$(LVGL_DIR)/qnx/aarch64/so.le
```

elahav:
> I think I spotted one more near the bottom
> 
> ```makefile
> LDFLAGS=-lscreen -llvgl -L$(LVGL_DIR)/qnx/aarch64/so.le
> ```

Fixed, thanks. I also renamed `LVGL_DIR` in the mekafile example to `LVGL_ROOT` to be more consistent with the rest of the document.

liamHowatt (approved):
Thanks!

kisvegabor (approved):

PR #6626, opened 2024-08-02
fix(png): use color rounding like img conv tool

### LINKED ISSUE #6491: Images converted using lvgl's online image converter exhibit color differences.
### LVGL version

v8.3.10

### What happened?

I need to convert a solid-color background image into a C array using LVGL's online converter. I've selected the CF_TRUE_COLOR_ALPHA color format, and I'm using a 16-bit color depth.

However, I have noticed a slight difference between the C array generated by the online converter and the C array I obtain by reading the same file directly using the png decoder.

Here are the details:
When I use the online converter, the converted value for each pixel in ARGB565 format is 0xE7 0x00 0xFF.
But when I read the file directly, the corresponding pixel value is 0xE6 0x00 0xFF.

For reference, the original ARGB888 value for that pixel in the C array is 0x36 0x1c 0x00 0xff. Converting this to ARGB565 results in the value 0xE6 0x00 0xFF.

I need to use the solid-color background image along with another image. When viewed using an image viewer software on Windows, there is no color difference between this image and the background.

Below is the solid-color background image.
bg

However, after converting the background images using the online Image Converter, there is a significant color difference.

When the background image is stored in Flash and displayed using a PNG decoder, the color difference issue does not arise. Alternatively, manually modifying the c-array after conversion in Image Converter to 0xE6 00 also eliminates the color difference.

Therefore, I would like to inquire if there is a potential issue with the online Image Converter.

### How to reproduce?

_No response_

liamHowatt:

Fixes #6491

### Notes

### DIFF
--- a/src/extra/libs/png/lv_png.c
+++ b/src/extra/libs/png/lv_png.c
@@ -276,16 +276,9 @@ static void convert_color_depth(uint8_t * img, uint32_t px_cnt)
 static inline lv_color_t lv_color_make_rounding(uint8_t r, uint8_t g, uint8_t b)
 {
 #if LV_COLOR_DEPTH == 16
-    /* if the high bits are not all set and the highest low bit is set     */
-    /* then the high bits should be rounded up so add one to the high bits */
-    /* if((r & 0b11111000) != 0b11111000 && (r & 0b100)) r += 0b1000       */
-    if((r & 0xf8) != 0xf8 && (r & 0x04)) r += 0x08;
-    if((g & 0xfc) != 0xfc && (g & 0x02)) g += 0x04;
-    if((b & 0xf8) != 0xf8 && (b & 0x04)) b += 0x08;
-#elif LV_COLOR_DEPTH == 8
-    if((r & 0xe0) != 0xe0 && (r & 0x10)) r += 0x20;
-    if((g & 0xe0) != 0xe0 && (g & 0x10)) g += 0x20;
-    if((b & 0xc0) != 0xc0 && (b & 0x20)) b += 0x40;
+    if(r <= 251) r += 4;
+    if(g <= 253) g += 2;
+    if(b <= 251) b += 4;
 #endif
 
     return lv_color_make(r, g, b);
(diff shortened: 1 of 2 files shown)


liamHowatt:
Status: 16 bit working, 8 bit not working.

src/extra/libs/png/lv_png.c
```diff
@@ -272,4 +273,22 @@ static void convert_color_depth(uint8_t * img, uint32_t px_cnt)
 #endif
 }
 
+static inline lv_color_t lv_color_make_rounding(uint8_t r, uint8_t g, uint8_t b)
+{
+#if LV_COLOR_DEPTH == 16
+    /* if the high bits are not all set and the highest low bit is set     */
```
kisvegabor: In case of red 3 bits will be dropped. In value it's 0..7. Can't we just add 3 (half of 7) to round up the value after clipping in lv_color_make? (Handle overflow too, if course)

liamHowatt:
Not stale

liamHowatt:
> In case of red 3 bits will be dropped. In value it's 0..7. Can't we just add 3 (half of 7) to round up the value after clipping in lv_color_make? (Handle overflow too, if course)

Yes, thank you. It's much simpler. 4 was the number to use for R and B. 4 is 0b100. If the 3rd bit is set in the color, it's more than halfway to the next value in (& 0b1111_1000), so adding 4 (0b100) will overflow the 3rd bit, causing a carry which increases the 4th bit, rounding up.

Regarding 8 bit depth, I've spent a long time on it and I'm seeing two issues. The JavaScript converter has R and B swapped. For RGB332, R gets 3 bits and B gets 2, but it's the other way around on the converter. It went unnoticed because RGB332 is the only color format where R is not the same number of bits as B, and then it gets swapped back at some point. The other issue is that I cannot get a PNG to decode and draw as RGB332 without a buffer overflow in LVGL.

- Should we fix the JS converter?
- Should we fix LVGL crashing when trying to decode a PNG in 8 bit RGB332 mode?

For now, 16 bit should now be fixed in this PR.

FYI, the 8 bit part would look like this, if it were included:
```c
    if(r <= 239) r += 16;
    if(g <= 239) g += 16;
    if(b <= 223) b += 32;
```

liamHowatt:
Correction: the JS converter has the color channels the right way around. It's `lv_png.c` which has the channels reversed.

kisvegabor:
Great! we can ignore the 8bit part. The RGB565 part is way more important.

kisvegabor (approved):

XuNeo (approved):

PR #6592, opened 2024-07-29
feat(event): add event delay deletion

xiaoleipeng:
### Description of the feature or fix

In the event (event list) callback, adding or deleting events will cause the program to crash. For this reason, I added a flag to the event list to mark whether there is a deletion operation during the traversal of the event list. If a remove is detected, During operation, I will uniformly adjust the structure of the event list after the traversal is completed to achieve safe deletion of the event list to ensure the stability of the program.
I conducted a 48-hour stress test. After the modification, I can ensure that my program runs normally.

```
void ui_event_btn_start(lv_event_t* e)
{
    lv_event_code_t event_code = lv_event_get_code(e);

    if (event_code == LV_EVENT_CLICKED) {
        LV_LOG_USER("Start button clicked");
        lv_obj_remove_event_cb(ui_btn_start, ui_event_btn_delete);
    }
}

void ui_anim_screen_init(void)
{
    lv_obj_add_event_cb(ui_btn_start, ui_event_btn_delete, LV_EVENT_DELETE, NULL);
    lv_obj_add_event_cb(ui_btn_start, ui_event_btn_start, LV_EVENT_CLICKED, NULL);
}
```

- Before modification, the resulting crash stack information

```
==51694==ERROR: AddressSanitizer: heap-use-after-free on address 0x60400002da60 at pc 0x5555560d570a bp 0x7fffffffcfe0 sp 0x7fffffffcfd0
READ of size 8 at 0x60400002da60 thread T0
    #0 0x5555560d5709 in lv_event_send /lvgl/src/misc/lv_event.c:71
    #1 0x5555560072e2 in event_send_core /lvgl/src/core/lv_obj_event.c:364
    #2 0x555556004826 in lv_obj_send_event /lvgl/src/core/lv_obj_event.c:64
    #3 0x5555560b3094 in send_event /lvgl/src/indev/lv_indev.c:1628
    #4 0x5555560ae6bf in indev_proc_release /lvgl/src/indev/lv_indev.c:1320
    #5 0x5555560a4959 in indev_pointer_proc /lvgl/src/indev/lv_indev.c:680
    #6 0x55555609fb0e in lv_indev_read /lvgl/src/indev/lv_indev.c:245
    #7 0x555556097905 in lv_sdl_mouse_handler /lvgl/src/drivers/sdl/lv_sdl_mouse.c:201
    #8 0x55555609a0b9 in sdl_event_handler /lvgl/src/drivers/sdl/lv_sdl_window.c:306
[...]

PR #6634, opened 2024-08-05
feat(event): cache removed event

FASTSHIFT:

This modification is similar to https://github.com/lvgl/lvgl/pull/6592. Both add a mark to the deleted event. This PR marks the event slot as pending reuse. This solution should be the most efficient because it saves the overhead introduced by the `realloc` operation of `lv_array`, and the changes to the current code are minimal. The disadvantage is that the insertion order cannot be guaranteed, which may affect the current APP that relies on the order of event sending.
The design of https://github.com/lvgl/lvgl/pull/6592 is to mark the deleted events as pending deletion and clean them up uniformly after the events are sent. The advantage is that it will not affect the event order and can ensure forward compatibility. The disadvantage is that it is necessary to introduce additional member variables to record whether event_list has been deleted.
Which option do you prefer?

### Notes

### DIFF
--- a/src/misc/lv_event.c
+++ b/src/misc/lv_event.c
@@ -88,17 +88,33 @@ lv_result_t lv_event_send(lv_event_list_t * list, lv_event_t * e, bool preproces
 lv_event_dsc_t * lv_event_add(lv_event_list_t * list, lv_event_cb_t cb, lv_event_code_t filter,
                               void * user_data)
 {
+    if(lv_array_size(list) == 0) {
+        /*event list hasn't been initialized.*/
+        lv_array_init(list, 1, sizeof(lv_event_dsc_t *));
+        LV_TRACE_EVENT("Event list initialized");
+    }
+    else {
+        uint32_t size = lv_array_size(list);
+        lv_event_dsc_t ** events = lv_array_front(list);
+        for(uint32_t i = 0; i < size; i++) {
+            /*Search for a free slot*/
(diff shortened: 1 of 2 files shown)


XuNeo:
> The disadvantage is that the insertion order cannot be guaranteed, which may affect the current APP that relies on the order of event sending

If there's application rely on this feature, can we provide a log to assist debug after this PR is merged?

FASTSHIFT:
> > The disadvantage is that the insertion order cannot be guaranteed, which may affect the current APP that relies on the order of event sending
> 
> If there's application rely on this feature, can we provide a log to assist debug after this PR is merged?

Done

kisvegabor (changes requested):
The order of events is very important and should be fully determined and reproducible. Otherwise it will cause very hard to debug anomalies. 


Is even removal so common in your use cases that it's worth optimizing for it?

FASTSHIFT:
> The order of events is very important and should be fully determined and reproducible. Otherwise it will cause very hard to debug anomalies.
> 
> Is even removal so common in your use cases that it's worth optimizing for it?

There are no scenarios where events are frequently added or deleted.

kisvegabor:
In this case I wouldn't risk adding events in a non-defined order.

FASTSHIFT:
> In this case I wouldn't risk adding events in a non-defined order.

OK，let's continue here https://github.com/lvgl/lvgl/pull/6592.

Issue #24, opened 2017-09-03
password field not being updated correctly when lv_ta_del called


scopa90:
There looks to be an issue with a password text area field.  The following:

 ```
 lv_obj_t *pin_inp = lv_ta_create(lv_scr_act(), NULL);
  lv_ta_set_one_line(pin_inp, true);
  lv_ta_set_cursor_show(pin_inp, true);
  lv_ta_set_pwd_mode(pin_inp, true);
  lv_ta_set_text(pin_inp, "");
  lv_ta_set_cursor_pos(pin_inp, LV_TA_CUR_LAST);
  lv_ta_add_char(pin_inp, '1');
  lv_ta_set_cursor_pos(pin_inp, LV_TA_CUR_LAST);
  lv_ta_del(pin_inp);
  lv_ta_set_cursor_pos(pin_inp, LV_TA_CUR_LAST);
  lv_ta_add_char(pin_inp, '2');
  char buff[3];
  strcpy(buff, lv_ta_get_txt(pin_inp));

```
Will leave "21" in buff instead of "2"

kisvegabor:
Hi,
Thank you for reporting the issue. I forget lv_ta_del() when fixed lv_ta_add_char/text()

I pushed the patch into the `beta` branch.

scopa90:
Fixed now. Thanks and enjoy your holiday.

kisvegabor:
Thanks :) Actually it will start tomorrow.

Issue #23, opened 2017-09-01
Sorted data in list


scopa90:
Would it be possible to add a sort function to the lv_list_add function or create a new lv_list_add_sorted function?

kisvegabor:
Sure! I will be on a holiday for a week. After that I can deal with it!

kisvegabor:
Recently I was thinking about this suggestion and found that it's very complicated to solve inside LittelvGL. 
It seems much easier and flexible to do it outside of the library using `lv_lsit_clean` and adding the buttons again.

Now, I close this issue but feel free to comment here and reopen it if you have any remarks.

PR #6638, opened 2024-08-05
perf(draw): improve execute time of obj's transforming

jaeho2021:
Change-Id: I943b26504251d26305b0d1fa97bdb11b0331ead2

I had a problem such as conflict commit.
and I'm not familiar with how to use GitHub, thus delete previous commit and made new commit based on lastest code. 

please review this change.

I try to resolve the problems of first PR (https://github.com/lvgl/lvgl/pull/6327)
As I explain before, I want to reduce execute time of lv_obj_get_transformed_area().

the key of this solution, mark(increase) to object's transform_changed variable when parent has transform property, , object's childrens as well.

LV_STYLE_TRANSFORM_ROTATION
LV_STYLE_TRANSFORM_SCALE_X
LV_STYLE_TRANSFORM_SCALE_Y
If the transform_changed value of an object is not 0, it means that the transform property (scale, angle) of the parent of the object has changed.
That's why you need to change the object's scale or angle according to the scale and angle values ​​of the parent.

for example,
null
|
a ( a object's transform_changed is 0 )
|
b ( b has scale and angle property, b object's transform_changed is 1 )
|
c ( c has angle property, c object's transform_changed is 2 )
|
d ( object's transform_changed is 2 )

object d's transform_changed is 2. that means d's parent has transform property (scale, angle).
so call lv_obj_transform_point() to check parent's transform property( scale, angle )
object c, d's parent, also is set to 2, so check c's parent one more time
object b, c's parent, is set to 1. however, parent(a)'s transform_changed set to 0.
This means that there is no need to check the parent's attributes anymore.

In this way we can improve execution speed by reducing unnecessary recursive function calls.

### Notes

### DIFF
--- a/src/core/lv_obj_pos.c
+++ b/src/core/lv_obj_pos.c
@@ -783,12 +783,14 @@ void lv_obj_transform_point_array(const lv_obj_t * obj, lv_point_t points[], siz
         bool recursive = flags & LV_OBJ_POINT_TRANSFORM_FLAG_RECURSIVE;
         bool inverse = flags & LV_OBJ_POINT_TRANSFORM_FLAG_INVERSE;
         if(inverse) {
-            if(recursive) lv_obj_transform_point_array(lv_obj_get_parent(obj), points, count, flags);
+            if(recursive &&
+               obj->parent->transform_changed) lv_obj_transform_point_array(lv_obj_get_parent(obj), points, count, flags);
             if(do_tranf) transform_point_array(obj, points, count, inverse);
         }
         else {
             if(do_tranf) transform_point_array(obj, points, count, inverse);
-            if(recursive) lv_obj_transform_point_array(lv_obj_get_parent(obj), points, count, flags);
+            if(recursive &&
+               obj->parent->transform_changed) lv_obj_transform_point_array(lv_obj_get_parent(obj), points, count, flags);
         }
     }
 }
@@ -878,7 +880,8 @@ bool lv_obj_area_is_visible(const lv_obj_t * obj, lv_area_t * area)
     /*The area is not on the object*/
     if(!lv_area_intersect(area, area, &obj_coords)) return false;
 
-    lv_obj_get_transformed_area(obj, area, LV_OBJ_POINT_TRANSFORM_FLAG_RECURSIVE);
+    if(obj->transform_changed)
+        lv_obj_get_transformed_area(obj, area, LV_OBJ_POINT_TRANSFORM_FLAG_RECURSIVE);
 
     /*Truncate recursively to the parents*/
     lv_obj_t * parent = lv_obj_get_parent(obj);
@@ -893,7 +896,9 @@ bool lv_obj_area_is_visible(const lv_obj_t * obj, lv_area_t * area)
             lv_area_increase(&parent_coords, parent_ext_size, parent_ext_size);
         }
 
-        lv_obj_get_transformed_area(parent, &parent_coords, LV_OBJ_POINT_TRANSFORM_FLAG_RECURSIVE);
+        if(parent->transform_changed)
+            lv_obj_get_transformed_area(parent, &parent_coords, LV_OBJ_POINT_TRANSFORM_FLAG_RECURSIVE);
+
         if(!lv_area_intersect(area, area, &parent_coords)) return false;
 
         parent = lv_obj_get_parent(parent);
--- a/src/core/lv_obj_pos.c
+++ b/src/core/lv_obj_pos.c
@@ -854,6 +854,15 @@ void lv_obj_invalidate(const lv_obj_t * obj)
     lv_obj_invalidate_area(obj, &obj_coords);
 }
 
+bool is_transformed(const lv_obj_t * obj)
+{
+    while(obj) {
+        if(obj->spec_attr && obj->spec_attr->layer_type == LV_LAYER_TYPE_TRANSFORM) return true;
+        obj = obj->parent;
+    }
+    return false;
+}
+
 bool lv_obj_area_is_visible(const lv_obj_t * obj, lv_area_t * area)
 {
     if(lv_obj_has_flag(obj, LV_OBJ_FLAG_HIDDEN)) return false;
@@ -878,7 +887,12 @@ bool lv_obj_area_is_visible(const lv_obj_t * obj, lv_area_t * area)
     /*The area is not on the object*/
     if(!lv_area_intersect(area, area, &obj_coords)) return false;
 
-    lv_obj_get_transformed_area(obj, area, LV_OBJ_POINT_TRANSFORM_FLAG_RECURSIVE);
+    if(!is_transformed(obj)) {
+        *area = obj->coords;
+    }
+    else {
+        lv_obj_get_transformed_area(obj, area, LV_OBJ_POINT_TRANSFORM_FLAG_RECURSIVE);
+    }
 
     /*Truncate recursively to the parents*/
     lv_obj_t * parent = lv_obj_get_parent(obj);
@@ -893,7 +907,12 @@ bool lv_obj_area_is_visible(const lv_obj_t * obj, lv_area_t * area)
             lv_area_increase(&parent_coords, parent_ext_size, parent_ext_size);
         }
 
-        lv_obj_get_transformed_area(parent, &parent_coords, LV_OBJ_POINT_TRANSFORM_FLAG_RECURSIVE);
+        if(!is_transformed(parent)) {
+            parent_coords = parent->coords;
+        }
+        else {
+            lv_obj_get_transformed_area(parent, &parent_coords, LV_OBJ_POINT_TRANSFORM_FLAG_RECURSIVE);
+        }
         if(!lv_area_intersect(area, area, &parent_coords)) return false;
 
         parent = lv_obj_get_parent(parent);
--- a/src/core/lv_obj_pos.c
+++ b/src/core/lv_obj_pos.c
@@ -783,14 +783,12 @@ void lv_obj_transform_point_array(const lv_obj_t * obj, lv_point_t points[], siz
         bool recursive = flags & LV_OBJ_POINT_TRANSFORM_FLAG_RECURSIVE;
         bool inverse = flags & LV_OBJ_POINT_TRANSFORM_FLAG_INVERSE;
         if(inverse) {
-            if(recursive &&
-               obj->parent->transform_changed) lv_obj_transform_point_array(lv_obj_get_parent(obj), points, count, flags);
+            if(recursive) lv_obj_transform_point_array(lv_obj_get_parent(obj), points, count, flags);
             if(do_tranf) transform_point_array(obj, points, count, inverse);
         }
         else {
             if(do_tranf) transform_point_array(obj, points, count, inverse);
-            if(recursive &&
-               obj->parent->transform_changed) lv_obj_transform_point_array(lv_obj_get_parent(obj), points, count, flags);
+            if(recursive) lv_obj_transform_point_array(lv_obj_get_parent(obj), points, count, flags);
         }
     }
 }
@@ -880,8 +878,7 @@ bool lv_obj_area_is_visible(const lv_obj_t * obj, lv_area_t * area)
     /*The area is not on the object*/
     if(!lv_area_intersect(area, area, &obj_coords)) return false;
 
-    if(obj->transform_changed)
-        lv_obj_get_transformed_area(obj, area, LV_OBJ_POINT_TRANSFORM_FLAG_RECURSIVE);
+    lv_obj_get_transformed_area(obj, area, LV_OBJ_POINT_TRANSFORM_FLAG_RECURSIVE);
 
     /*Truncate recursively to the parents*/
     lv_obj_t * parent = lv_obj_get_parent(obj);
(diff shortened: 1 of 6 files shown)


src/core/lv_obj_style.c
```diff
@@ -286,6 +298,9 @@ void lv_obj_refresh_style(lv_obj_t * obj, lv_style_selector_t selector, lv_style
     bool is_ext_draw = lv_style_prop_has_flag(prop, LV_STYLE_PROP_FLAG_EXT_DRAW_UPDATE);
     bool is_inheritable = lv_style_prop_has_flag(prop, LV_STYLE_PROP_FLAG_INHERITABLE);
     bool is_layer_refr = lv_style_prop_has_flag(prop, LV_STYLE_PROP_FLAG_LAYER_UPDATE);
+    bool is_transform_rotation = lv_style_prop_has_flag(prop, LV_STYLE_TRANSFORM_ROTATION);
```
kisvegabor: `LV_STYLE_TRANSFORM_ROTATION` is not a flag. Did you mean `prop == LV_STYLE_TRANSFORM_ROTATION`?

kisvegabor (commented):
Hey,

Sorry for the late reply. Have you measured the performance change?

jaeho2021:
actually, I have measure data based on v8.3.11. as you know, It depends on object hierarchy and tranform property
in my case, object hierarchy is like below and no transform property

measure tool : SEGGER system view

before : about 830ms 
after :  about 570ms 

image

src/core/lv_obj_style.c
jaeho2021: oops!. correct. I upload new patch. however, I got build error in C/C++ CI / Run tests with 64bit build (pull_request). 
let me check furthur

jaeho2021: I can't find any dependency between 2nd patch and the fail reason below,
could someone help me? 


C/C++ CI / Run tests with 64bit build (pull_request) 

20/92 Test #23: test_libpng ......................***Failed    0.29 sec
/home/runner/work/lvgl/lvgl/tests/src/test_cases/libs/test_libpng.c:73:test_libpng_1:FAIL: Expected 48 to be less than or equal to 32

-----------------------
1 Tests 1 Failures 0 Ignored 
FAIL

      Start 24: test_lodepng
21/92 Test #24: test_lodepng .....................***Failed    0.31 sec
/home/runner/work/lvgl/lvgl/tests/src/test_cases/libs/test_lodepng.c:73:test_lodepng_1:FAIL: Expected 56 to be less than or equal to 40

-----------------------
1 Tests 1 Failures 0 Ignored 
FAIL

jaeho2021: I'm not sure, but it seems like the problem came from adding the transform_changed variable to the "struct lv_obj_t" structure.

kisvegabor:
Hey, 
Sorry for the delay. Finally, v9.2 is released and I can get back to reviewing PR.  

The CI fails because memory leakage is detected. Please increase `20` to `100` [here](https://github.com/lvgl/lvgl/blob/master/tests/src/test_cases/libs/test_lodepng.c#L64) to see if the leakage also increases. Note that due to memory fragmentation, it's normal that the memory usage is not same even if you delete all the widgets.

jaeho2021:
when I've test as you comment, there is no memory leakage further except first error we had met. 
As a above result, I guess, we can change test pass criteria.

test_libpng_1()
from  TEST_ASSERT_MEM_LEAK_LESS_THAN(mem_before, 32);
to TEST_ASSERT_MEM_LEAK_LESS_THAN(mem_before, 48);

test_lodepng_1()
from TEST_ASSERT_MEM_LEAK_LESS_THAN(mem_before, 40);
to TEST_ASSERT_MEM_LEAK_LESS_THAN(mem_before, 56);

please comment you opinion

kisvegabor:
Okay, please increase the limits as you have recommended.

jaeho2021:
thanks for your feedback. I've increased the limits and All test passed.

src/core/lv_obj_private.h
```diff
@@ -70,6 +70,7 @@ struct lv_obj_t {
     uint16_t h_layout   : 1;
     uint16_t w_layout   : 1;
     uint16_t is_deleting : 1;
+    uint16_t transform_changed;
```
kisvegabor:
```suggestion
    /** Count how many ancestors are transformed. The widget itself counts as well. 
      * Used to skip unnecessary area transformations if there are no transformations. */
    uint16_t transformed_ancestor_count; 
```

src/core/lv_obj_style.c
```diff
@@ -23,6 +23,8 @@
 #define style_trans_ll_p &(LV_GLOBAL_DEFAULT()->style_trans_ll)
 #define _style_custom_prop_flag_lookup_table LV_GLOBAL_DEFAULT()->style_custom_prop_flag_lookup_table
 #define STYLE_PROP_SHIFTED(prop) ((uint32_t)1 << ((prop) >> 3))
+#define INCREASE_CNT  1
+#define DECREASE_CNT -1
```
kisvegabor: In my opinion these are too trivial and the values can be used directly.

src/core/lv_obj_style.c
```diff
@@ -286,6 +298,9 @@ void lv_obj_refresh_style(lv_obj_t * obj, lv_style_selector_t selector, lv_style
     bool is_ext_draw = lv_style_prop_has_flag(prop, LV_STYLE_PROP_FLAG_EXT_DRAW_UPDATE);
     bool is_inheritable = lv_style_prop_has_flag(prop, LV_STYLE_PROP_FLAG_INHERITABLE);
     bool is_layer_refr = lv_style_prop_has_flag(prop, LV_STYLE_PROP_FLAG_LAYER_UPDATE);
+    bool is_transform_rotation = (prop == LV_STYLE_TRANSFORM_ROTATION) ? true : false;
```
```suggestion
    bool is_transform_rotation = (prop == LV_STYLE_PROP_ANY || prop == LV_STYLE_TRANSFORM_ROTATION) ? true : false;
```
We need t handle `LV_STYLE_PROP_ANY` but in this case `transform_changed` will be always incremented.

kisvegabor (commented):
I wonder how safe this approach is. E.g. what if a widget's parent changes from a non transformed to transformed one. In this case the `transform_changed` count is not updated. 

I guess the goal is to optimize the most common case when nothing is transformed. I think a function like 
```c
bool is_transformed(lv_obj_t * obj)
{
  lv_obj_t * parent = obj;
  while(parent) {
    if(obj->spec_attr && obj->spec_attr->layer_type == LV_LAYER_TYPE_TRANSFORM) return true;
    parent = lv_obj_get_parent(obj);
  }
 return false;
}
```
could tell very quickly if it's worth running `lv_obj_transform_point_array/point` or we can use the area as its. 

This approach seems less fragile to me and also quite fast. What do you think?

jaeho2021:
if our goal is to optimize the common case that does not use transformed, I guess we can cover it with following code:
Although it might be difficult to optimize if the parent has a transform property, I think this is the best way. 
In my case, is_transformed() function takes about 5us. 

```c
bool lv_obj_area_is_visible(const lv_obj_t * obj, lv_area_t * area) {
    ....
    if(is_transformed(obj)) {                                  <<<<<<<  add here!
        lv_obj_get_transformed_area(obj, area, LV_OBJ_POINT_TRANSFORM_FLAG_RECURSIVE);
    }


    /*Truncate recursively to the parents*/
    while(parent != NULL) {
        ...
        if(is_transformed(parent))
            lv_obj_get_transformed_area(parent, &parent_coords, LV_OBJ_POINT_TRANSFORM_FLAG_RECURSIVE);
        ...
    }
``` 


On the other hand, if we use this function (is_transformed()) to determine whether it is worth calling the lv_obj_transform_point_array/point function, it may become an unnecessary function in the case where the actual transformation is performed, this is because the is_transformed() function is called frequently and in my case, it takes about 5us per call. 

```c
void lv_obj_transform_point_array(const lv_obj_t * obj, lv_point_t points[], size_t count,
                                  lv_obj_point_transform_flag_t flags)
{
    if(obj) {
        lv_layer_type_t layer_type = lv_obj_get_layer_type(obj);
        bool do_tranf = layer_type == LV_LAYER_TYPE_TRANSFORM;
        bool recursive = flags & LV_OBJ_POINT_TRANSFORM_FLAG_RECURSIVE;
        bool inverse = flags & LV_OBJ_POINT_TRANSFORM_FLAG_INVERSE;
        if(inverse) {
            if(recursive && is_transformed(obj) ) lv_obj_transform_point_array(lv_obj_get_parent(obj), points, count, flags);
            if(do_tranf) transform_point_array(obj, points, count, inverse);
        }
        else {
            if(do_tranf) transform_point_array(obj, points, count, inverse);
            if(recursive && is_transformed(obj) ) lv_obj_transform_point_array(lv_obj_get_parent(obj), points, count, flags);
        }
    }
}

```

src/core/lv_obj_style.c
jaeho2021: this may be a basic question, 
could you let me know why we handle LV_STYLE_PROP_ANY as well?

jaeho2021:
Not stale

src/core/lv_obj_style.c
kisvegabor: `LV_STYLE_PROP_ANY` means we don't know the property. E.g. because a style was added with multiple properties.

kisvegabor:
Sorry for the delay. :slightly_smiling_face: 

If we assume that the most common case is the no-transformation and if there is transformation is will slow anyway, we can say:
```c
bool lv_obj_area_is_visible(const lv_obj_t * obj, lv_area_t * area) {
     if(!is_transformed(obj) { 
        *area = obj->coords;   //Happy path
     } else {
         complex calculation  //Long calculation but not typical
     }
}
```

It can help to speed up `is_transformed`:
```c
bool is_transformed(lv_obj_t * obj)
{
  lv_obj_t * parent = obj;
  while(parent) {
    if(obj->spec_attr && obj->spec_attr->layer_type == LV_LAYER_TYPE_TRANSFORM) return true;
    parent = obj->parent;  //Direct access the parent
  }
 return false;
}
```

jaeho2021:
I did rebase on lastest code.
there was no code that I initially intended to reflect 😂,
but I think the above code will improve performance for cases without transform.

kisvegabor:
Could you measure the performance change? Is it still noticeable?

src/core/lv_obj_pos.h
```diff
@@ -440,6 +440,13 @@ int32_t lv_clamp_width(int32_t width, int32_t min_width, int32_t max_width, int3
  */
 int32_t lv_clamp_height(int32_t height, int32_t min_height, int32_t max_height, int32_t ref_height);
 
+/**
+ * Checks whether the given object or any of its parent objects has been transformed.
+ * @param           obj Pointer to a constant lv_obj_t object.
+ * @return          true if the object or any of its parents is transformed, false otherwise.
+ */
+bool is_transformed(const lv_obj_t * obj);
```
kisvegabor:
```suggestion
static bool is_transformed(const lv_obj_t * obj);
```
And please move it to the C file

kisvegabor (dismissed):

src/core/lv_obj_pos.c
```diff
@@ -854,6 +855,15 @@ void lv_obj_invalidate(const lv_obj_t * obj)
     lv_obj_invalidate_area(obj, &obj_coords);
 }
 
+bool is_transformed(const lv_obj_t * obj)
+{
+    while(obj) {
+        if(obj->spec_attr && obj->spec_attr->layer_type == LV_LAYER_TYPE_TRANSFORM) return true;
+        obj = obj->parent;
```
kissa96: Unreachable code due to previous return at 861. Please remove

jaeho2021:
I think, still pretty good. 
image

src/core/lv_obj_pos.c
jaeho2021: the purpose of this code is to check object and its parent has transform property.

kissa96: ack

kisvegabor (commented):
Great, it's 30% with a few lines of code. I like it. :slightly_smiling_face:

src/core/lv_obj_pos.c
```diff
@@ -854,6 +855,15 @@ void lv_obj_invalidate(const lv_obj_t * obj)
     lv_obj_invalidate_area(obj, &obj_coords);
 }
 
+bool is_transformed(const lv_obj_t * obj)
```
liamHowatt:
```suggestion
static bool is_transformed(const lv_obj_t * obj)
```

And move to the `STATIC FUNCTIONS` section at the bottom.

src/core/lv_obj_pos.c
```diff
@@ -878,7 +888,12 @@ bool lv_obj_area_is_visible(const lv_obj_t * obj, lv_area_t * area)
     /*The area is not on the object*/
     if(!lv_area_intersect(area, area, &obj_coords)) return false;
 
-    lv_obj_get_transformed_area(obj, area, LV_OBJ_POINT_TRANSFORM_FLAG_RECURSIVE);
+    if(!is_transformed(obj)) {
+        *area = obj->coords;
+    }
+    else {
+        lv_obj_get_transformed_area(obj, area, LV_OBJ_POINT_TRANSFORM_FLAG_RECURSIVE);
+    }
```
liamHowatt, line 887: Is the condition inverted by mistake? Should the `!` be removed?

jaeho2021: if obj and its parent has transform property, is_transformed() will be return true.
This means that the lv_obj_get_transformed_area() call is necessary.

on the other way, return false (obj and its parent doesn't have any transform property) means that we don't need to transform object's size

src/core/lv_obj_pos.c
jaeho2021: Good!. done.

src/core/lv_obj_pos.c
liamHowatt: Ok, thanks for clarifying

src/core/lv_obj_pos.c
liamHowatt: Thanks. Could you please also move the function body below the

```c
/**********************
 *   STATIC FUNCTIONS
 **********************/
```

section?

jaeho2021: yes. thanks for checking

kisvegabor (approved):

liamHowatt (approved):
Thanks!

Issue #78, opened 2018-01-17
sparse font support


hotpeperoncino:
Far east font like Chinese, Japanese font has mostly sparse data. (valid unicode values are sparse)
Current font implementation looks expecting that the font data is not sparse.

I guess such logic.
1. if sparse data
       map unicode value to index in fontdata.
   if not spase data
      unicode value is used as index
2. index is used to retrieve font data.

please prepare such mechanism to handle sparse font in the library

kisvegabor:
Hi,

I'm planning to implement this feature in 5.1 release.

hotpeperoncino:
i'll wait for it

xiaoleizii:
![20180118175156](https://user-images.githubusercontent.com/11530558/35091733-4ffb6942-fc78-11e7-9f60-0b3bba16c07e.jpg)

xiaoleizii:
I'm sorry my English is not good, this passage is translated by Google translation.
The number of commonly used Chinese characters in Chinese characters is huge (Chinese common characters are 20902 words, Unicode encoding range is 0x4E00-0x9FA5), which means that I need a large external FLASH to store this continuous encoded font lattice data. But in general, the embedded system needs a limited number of text, usually 200-500 words, so a solution is needed to make 200-500 words as a font file, the Unicode encoding of the characters in the font is not continuous , So that you can use the Chinese font without increasing hardware costs.
Currently I've implemented the functionality I need by adjusting the lv_font_get_bitmap () and lv_font_get_width () functions in the lv_font.c file, but this is not universally applicable and I hope you can consider adding support for this requirement.
I personally feel the easiest way is to create the font, the text corresponding to the Unicode encoding value can also be stored in an array, read from the array to find the corresponding Unicode encoding value, and then return from the dot matrix data corresponding Text lattice.

kisvegabor:
Hi,
Thanks you for the image! 
Support for Chinese character will be added in the next release! 

My idea is the following:
Now `lv_font_t` describes the fonts with arrays. I'm planning to extend it with 2 functions:
1. **read_bitmap(unicode_letter, buf)** which fills `buf` with bit-based image of the letter. You can load it from anywhere, even from external flash. It depends on your implementation.
2. **read_width(unicode_letter)** get the width of a character (can return a constant for Chinese letters)

You can create a font (in the next release) this way:
```c
static lv_font_t custom_font;
custom_font.read_bitmap = my_font_30_read_bitmap;
[...]

PR #75, opened 2018-01-16
fix warning/error with lv_conf.h option

Zaltora:
I found some warning and i fix it when user change some settings in lv_conf.h

I discovered that this function seem do not be used (line 72):
`static void (*px_fp)(lv_coord_t x, lv_coord_t y, const lv_area_t * mask, lv_color_t color, lv_opa_t opa) = lv_rpx;`

It is WIP, don't merge now

Issue #67, opened 2018-01-09
compilation failure if LV_INDEV_READ_PERIOD = 0


hreichert:
Defined in `lv_conf.h`:
```
/*Input device settings*/
#define LV_INDEV_READ_PERIOD 0
```

Compilation error:
> [Error] lv_indev.c@60,25: too few arguments to function 'lv_task_create'                                                                          
> [ERROR] ./libs/littlevgl/lvgl/lv_core/lv_indev.c: In function 'lv_indev_init':                                                                    
> ./libs/littlevgl/lvgl/lv_core/lv_indev.c:60:25: error: too few arguments to function 'lv_task_create'

Temporary fix to make it work locally:
(added NULL as fourth argument)
```
diff --git a/lv_core/lv_indev.c b/lv_core/lv_indev.c
index 25a4157..c180478 100644
--- a/lv_core/lv_indev.c
+++ b/lv_core/lv_indev.c
@@ -57,7 +57,7 @@ void lv_indev_init(void)
 #if LV_INDEV_READ_PERIOD != 0
     indev_proc_task_p = lv_task_create(indev_proc_task, LV_INDEV_READ_PERIOD, LV_TASK_PRIO_MID, NULL);
 #else
-    indev_proc_task_p = lv_task_create(indev_proc_task, 1, LV_TASK_PRIO_OFF); /*Not use lv_indev_proc*/
+    indev_proc_task_p = lv_task_create(indev_proc_task, 1, LV_TASK_PRIO_OFF, NULL); /*Not use lv_indev_proc*/
 #endif
 
     lv_indev_reset(NULL);   /*Reset all input devices*/
```

kisvegabor:
Thank you!
I fixed in the `beta` branch. Now `LV_INDEV_READ_PERIOD 0` also disabled a lot of functions in lv_indev.c to make the code smaller with this configuration.

hreichert:
Thank you very much! Really appreciate your efforts.

PR #6646, opened 2024-08-06
perf(nxp): v9.1.0 upstream vglite optimization

nicusorcitu:

- Align the new lvgl logo to stride 64 bytes
- Add draw_unit count
**- For systems with just one draw unit available, during dispatching tasks we don't have to wait for fill screen to complete or to wait for task dependencies - task dependencies are sent in order anyway.
E.G. in case of a button with a label on it, first task queued is the rectangle then is the label. Label task can be dispatched right away to the drawer even if rectangle is not yet complete.**
- Update VGLite task queue to support up to 100 tasks to be flushed just once to the GPU
- VGLite draw unit can run as a separate thread and draw asynchronous a bunch of tasks from the queue. Once the GPU will complete those tasks, it will report to the dispatcher the ready status
- Add the capability to signal the draw thread to complete the pending tasks - via "wait_for_finish_cb"


nicusorcitu:
A comparison before and after have been made on RT1170:
[vglite_perf_optimization.xlsx](https://github.com/user-attachments/files/16508206/vglite_perf_optimization.xlsx)

nicusorcitu:
Hi @kisvegabor @XuNeo 
I see this test failing:
```
/home/nxf68195/bitbucket/lvgl_mcu/lvgl/tests/src/test_cases/widgets/test_bar.c:190:test_bar_render_corner:INFO: 
Screenshot compare error
  - File: ref_imgs/widgets/bar_corner_2.png
  - Mode: full refresh
  - At x:30, y:121.
  - Expected: 0x00000000003DBD4A
  - Actual:   0x0000000000B646C3
  - Tolerance: 0
/home/nxf68195/bitbucket/lvgl_mcu/lvgl/tests/src/test_cases/widgets/test_bar.c:391:test_bar_render_corner:FAIL: widgets/bar_corner_2.png
```
Is it saved somewhere the test output before the comparation with the reference from widgets/bar_corner_2.png?

nicusorcitu:
#60
**ref**
image
**out**
image

#67
**ref**
image
**out**
image

#75
**ref**
image
**out**
image

#78
**ref**
image
**out**
image

nicusorcitu:
I believe the root cause is that tasks are not queued in dependency order. Most of the tasks are related to rectangle draws (background, shadow, border, outline).

src/draw/lv_draw.c
```diff
@@ -430,6 +449,9 @@ static bool is_independent(lv_layer_t * layer, lv_draw_task_t * t_check)
     LV_PROFILER_BEGIN;
     lv_draw_task_t * t = layer->draw_task_head;
 
+    if(_draw_info.unit_cnt <= 1)
```
nicusorcitu: Removing this line would fix the problem, but it would be a woarkaround. The question is why the tasks are put in queue out of order?

src/draw/lv_draw.c
```diff
@@ -154,19 +155,32 @@ void lv_draw_finalize_task_creation(lv_layer_t * layer, lv_draw_task_t * t)
...
+        u = u->next;
+    }
+#endif
+}
+
 void lv_draw_dispatch(void)
 {
     LV_PROFILER_BEGIN;
-    bool render_running = false;
+    bool task_dispatched = false;
```
kisvegabor, line 177: It will be also `true` if no dispatch happened now, but there is an already running rendering.

src/draw/lv_draw.c
kisvegabor: It's a great idea!

They shouldn't be put into the queue out of order. They are added as LVGL traverses the widgets.

See this:
image
It seems multiple CPU cores are trying to draw the 2 rectangles at the same time. There is a little bit of scrollbar then the white fill will be stronger. 

However I've checked that in all test cases there is only 1 draw unit. Which is mistake and earlier we had tests with more draw units, but somehow it was removed. I've tried and the test are failing with more draw unit. I'll investigate tomorrow.

Anyway, I can't see how a draw unit can mess up with itself.

src/draw/lv_draw_private.h
```diff
@@ -111,6 +111,13 @@ struct lv_draw_unit_t {
      */
     int32_t (*evaluate_cb)(lv_draw_unit_t * draw_unit, lv_draw_task_t * task);
 
+    /**
+     *
```
kisvegabor, line 157: Please fill this

src/draw/lv_draw_private.h
```diff
@@ -111,6 +111,13 @@ struct lv_draw_unit_t {
...
+     * @param draw_unit
+     * @return
+     */
+    int32_t (*wait_for_finish_cb)(lv_draw_unit_t * draw_unit);
```
kisvegabor, line 161: @FASTSHIFT I believe it might be useful for you too

src/draw/lv_draw.c
nicusorcitu: Yes, the logic did not changed.
I rename it just because it can mean: unit running (busy), or task is unsupported by the unit, or no more tasks in the queue.

So the name `task_dispatched` looks more representative.

src/draw/lv_draw_private.h
nicusorcitu:
```
LVGL thread			VGLite thread		GPU HW
task1				queued			IDLE
				vg_lite_flush()		task1 is flushed to GPU
task2				queued			BUSY (with task1)

task3				queued			BUSY (still with task1)
				
task4				qeueud			IDLE (task1 ready now)
				vg_lite_flush()		task2, task3, task4 are flushed to GPU
				task1->state = READY
no more tasks in the queue
finish_cb()			vg_lite_finish()	GPU have to complete all tasks previously flushed
				task2->state = READY
				task3->state = READY
				task4->state = READY
```

nicusorcitu: I will rename the `wait_for_finish_cb` to just `finish_cb' because there is no waiting at all at the LVGL thread, it will be just a signal sent to VGLite thread to finish all pending tasks and send the ready status back.

nicusorcitu:
@cosmindanielradu19

src/draw/lv_draw.c
```diff
@@ -154,19 +155,32 @@ void lv_draw_finalize_task_creation(lv_layer_t * layer, lv_draw_task_t * t)
     LV_PROFILER_END;
 }
 
+void lv_draw_wait_for_finish()
```
FASTSHIFT: void lv_draw_wait_for_finish(void)

src/draw/lv_draw.c
nicusorcitu: If there are 2 CPU cores but just one SW draw unit (LV_DRAW_SW _DRAW_UNIT_CNT1), then is should be just one draw thread - hence no race condition.
If there are 2 CPU cores and 2 SW draw units (LV_DRAW_SW_DRAW_UNIT_CNT 2), then it should not pass the if condition.

In both situations this failure should not happen.

kisvegabor: I'm suspecting what's happening. It's related to the runtime created internal `lv_layer`s. When LVGL needs to create a layer (for a rotated widget, clip_corner, etc). In this case, each layer has its own list of draw tasks and somehow it's being messed up. 

I couldn't figure out the exact reason, therefore I suggest reverting this line and merge this PR for v9.2. We can fix it later in v9.3.

nicusorcitu: Done: [69543bf](https://github.com/lvgl/lvgl/pull/6646/commits/69543bf570a0a42a593cdb9e6d542f116135fbcc)

nicusorcitu:
Hi @kisvegabor 
I add the copyright modification patch [e158fd8](https://github.com/lvgl/lvgl/pull/6646/commits/e158fd85447112e4f9d6c12bc4fd43c3a9db9642)
The NXP policy is that every file updated consistently needs to have any of these model of copyright in place:
`Modified by NXP on <this date>`
or
`Modifications Copyright <year range> NXP`

I picked the second one, but please let me know if the first one looks better to you.

src/draw/lv_draw.c
```diff
@@ -3,6 +3,10 @@
  *
  */
 
+/**
+ * Modified by NXP on 2024
```
kisvegabor:
```suggestion
 * Modified by NXP in 2024
```

(Same for the others)

kisvegabor (dismissed):

src/draw/lv_draw.c
```diff
@@ -3,6 +3,10 @@
  *
  */
 
+/**
+ * Modified by NXP in 2024
+ */
```
liamHowatt, line 8: @vwheeler63 will the `/**` trigger unwanted Doxygen behavior?

src/draw/nxp/vglite/lv_draw_vglite.c
```diff
@@ -282,9 +288,26 @@ static int32_t _vglite_dispatch(lv_draw_unit_t * draw_unit, lv_layer_t * layer)
...
+        /* Fake unsupported tasks as ready. */
+        if(t->preferred_draw_unit_id != DRAW_UNIT_ID_VGLITE) {
+            t->state = LV_DRAW_TASK_STATE_READY;
+
+            /* Request a new dispatching as it can get a new task. */
+            lv_draw_dispatch_request();
+
+            return 1;
+        }
+    }
```
liamHowatt, line 309: I'm still learning how draw dispatching works. For the `else` case (one unit), wouldn't this consume a SW unit and skip it being drawn by any renderer?

src/draw/nxp/vglite/lv_draw_vglite.c
```diff
@@ -311,9 +334,22 @@ static int32_t _vglite_dispatch(lv_draw_unit_t * draw_unit, lv_layer_t * layer)
...
 }
 
+#if LV_USE_VGLITE_DRAW_ASYNC
+static int32_t _vglite_wait_for_finish(lv_draw_unit_t * draw_unit)
+{
+    lv_draw_vglite_unit_t * draw_vglite_unit = (lv_draw_vglite_unit_t *) draw_unit;
+    draw_vglite_unit->wait_for_finish = true;
+
+    if(draw_vglite_unit->inited)
+        lv_thread_sync_signal(&draw_vglite_unit->sync);
```
liamHowatt, line 344: Should this be a wait?

src/draw/nxp/vglite/lv_draw_vglite.c
```diff
@@ -442,41 +478,57 @@ static void _vglite_execute_drawing(lv_draw_vglite_unit_t * u)
...
+static inline void _vglite_signal_task_ready(lv_draw_task_t * task)
+{
+    /* Signal the ready state to dispatcher. */
+    task->state = LV_DRAW_TASK_STATE_READY;
+    _head = (_head + 1) % VGLITE_TASK_BUF_SIZE;
+
+    /* No need to cleanup the tasks in buffer as we advance with the _head. */
+}
+
+static inline void _vglite_signal_all_task_ready(void)
```
liamHowatt, line 499: Please forward declare the new static functions.

src/draw/sw/lv_draw_sw_private.h
```diff
@@ -38,6 +38,7 @@ struct lv_draw_sw_unit_t {
     lv_thread_t thread;
     volatile bool inited;
     volatile bool exit_status;
+    volatile bool wait_for_finish;
```
liamHowatt: I see why you put this member in the sw_unit struct: because `typedef lv_draw_sw_unit_t lv_draw_vglite_unit_t`. Maybe `lv_draw_vglite_unit_t` should be its own struct now?

src/widgets/canvas/lv_canvas.c
```diff
@@ -386,9 +390,17 @@ void lv_canvas_init_layer(lv_obj_t * obj, lv_layer_t * layer)
...
+    bool task_dispatched;
+
     while(layer->draw_task_head) {
         lv_draw_dispatch_wait_for_request();
-        lv_draw_dispatch_layer(lv_obj_get_display(canvas), layer);
+        task_dispatched = lv_draw_dispatch_layer(lv_obj_get_display(canvas), layer);
+
+        if(!task_dispatched) {
+            lv_draw_wait_for_finish();
+            lv_draw_dispatch_request();
```
liamHowatt, line 403: Is this allowed to be called outside of a draw driver like this?

src/draw/nxp/vglite/lv_draw_vglite.c
nicusorcitu: No, it has to be sync.

This API gets called by the LVGL thread and will signal the VGLite thread to run and wait for completion for all pending tasks.

src/draw/sw/lv_draw_sw_private.h
nicusorcitu: I put it here because some other units might use it.

src/widgets/canvas/lv_canvas.c
nicusorcitu: What do you mean?

src/draw/nxp/vglite/lv_draw_vglite.c
nicusorcitu: Is this mandatory?
Those small static functions (helpers) are implemented before their use, just to have the code cleaner.

All the callback functions used statically like `_vglite_dispatch`, `_vglite_delete`, `_vglite_execute`.... are declared on top of the file.

src/draw/nxp/vglite/lv_draw_vglite.c
nicusorcitu: I think I understand your concern...

The LVGL main thread calls `lv_draw_wait_for_finish();`
This means that the draw unit gets signaled to wait for completing the pending queue. The draw unit thread will actually wait for completion and signal back the tasks ready status.

src/draw/nxp/vglite/lv_draw_vglite.c
nicusorcitu: Each unit dispatcher can have its own implementation.

In NXP VGLite case, before this else implementation, unsupported tasks were given to the SW draw unit.
But that means that LVGL should have had one main LVGL thread + VGLite draw uni thread + Sw draw unit thread.
We disabled the SW thread because it was kept just to cover "few" unsupported tasks. Those "few" unsupported tasks are not so common (like the box shadow). We can implement them in the future if really need it.

But the performance is much better when we have only one GPU draw unit to process everything because we do not have to wait for any task dependency and getting rid of one freertos thread (SW draw unit) also saves the CPU.

src/draw/lv_draw.c
nicusorcitu: In LVGL I see a lot of comments style.
```
/**
 * ... text ...
 */
```
or
```
/*
 ... text ...
*/
```
or
```
/*********************************************
 *  ... text
 *********************************************/
```
It was no problem so far. But hence, what is the accepted comment style? All of them?

vwheeler63: > @vwheeler63 will the `/**` trigger unwanted Doxygen behavior?

Hi, Liam!

Technically no, but what it _is_ causing is probably not intentional:  it causes Doxygen to parse that comment, but then Doxygen will look for the code member it should assign that documentation to, and it won't find anything, so it will be ignored.  So it should either be:

1. If Gabor is okay with that comment being part of the FILE's documentation, then I would move it into the top comment after the `@file` command.
2. Otherwise, it should only have one asterisk:  `/*`, even though it won't hurt anything as explained.

src/draw/lv_draw.c
```diff
@@ -285,11 +303,16 @@ void lv_draw_dispatch_request(void)
...
+{
+    return _draw_info.unit_cnt;
+}
+
 lv_draw_task_t * lv_draw_get_next_available_task(lv_layer_t * layer, lv_draw_task_t * t_prev, uint8_t draw_unit_id)
 {
     LV_PROFILER_BEGIN;
     /*If the first task is screen sized, there cannot be independent areas*/
-    if(layer->draw_task_head) {
+    if(_draw_info.unit_cnt > 1 && layer->draw_task_head) {
```
kisvegabor: As the other `if(_draw_info.unit_cnt > 1)` is removed, it should be removed too.

src/widgets/canvas/lv_canvas.c
kisvegabor: Yes it's valid. When `lv_canvas_finish_layer` is called drawing is running and it's okay to handle dispatching here.

src/draw/sw/lv_draw_sw_private.h
kisvegabor: I agree to move it to VG-Lite rather. If others will need it they can also create their own types.

src/draw/nxp/vglite/lv_draw_vglite.c
kisvegabor: In these very specific files I'm not too picky about it, but we should follow `lv_templ.c/h` where the static functions are moved at the bottom.

We can leave them here now.

src/draw/nxp/vglite/lv_draw_vglite.c
```diff
@@ -282,9 +288,26 @@ static int32_t _vglite_dispatch(lv_draw_unit_t * draw_unit, lv_layer_t * layer)
...
+        if(t->preferred_draw_unit_id != DRAW_UNIT_ID_VGLITE)
+            return LV_DRAW_UNIT_IDLE;
+    }
+    else {
+        /* Fake unsupported tasks as ready. */
+        if(t->preferred_draw_unit_id != DRAW_UNIT_ID_VGLITE) {
+            t->state = LV_DRAW_TASK_STATE_READY;
+
+            /* Request a new dispatching as it can get a new task. */
+            lv_draw_dispatch_request();
```
kisvegabor, line 305: Having a `while` loop here to get as many tasks are possible might be faster. (Not critical now)

src/draw/lv_draw.c
```diff
@@ -154,19 +159,32 @@ void lv_draw_finalize_task_creation(lv_layer_t * layer, lv_draw_task_t * t)
...
         lv_layer_t * layer = disp->layer_head;
         while(layer) {
             if(lv_draw_dispatch_layer(disp, layer))
-                render_running = true;
+                task_dispatched = true;
             layer = layer->next;
         }
-        if(!render_running) {
+        if(!task_dispatched) {
+            lv_draw_wait_for_finish();
```
kisvegabor, line 188: Please confirm if it works like this:
1. VG-Lite queues 5 tasks
2. The 6th task shall be rendered by SW but it depends on the previous 5
3. Other VG-Lite tasks depend on the 6th SW task so VG-Lite cannot queue more tasks
4. As none of the draw units could dispatch new tasks `task_dispatched` will be `false` and `lv_draw_wait_for_finish` will be called
5. The first 5 draw task will be finished
6. SW can draw the 6th

So VG-Lite can queue the tasks until there is a blocking one.

kisvegabor (dismissed):

liamHowatt (dismissed):
Approving because each concern was addressed and none were critical.

src/draw/lv_draw.c
liamHowatt: > Otherwise, it should only have one asterisk: `/*`, even though it won't hurt anything as explained.

That's what I was wondering just because of [this thread](https://github.com/lvgl/lvgl/pull/6584#discussion_r1706008049). Glad it's harmless regardless.

vwheeler63: > Glad it's harmless regardless.

**thumbs up in agreement**

kisvegabor:
I'd like to wait a reply to [this](https://github.com/lvgl/lvgl/pull/6646#discussion_r1729508158). Other than we should be good to go.

src/draw/nxp/vglite/lv_draw_vglite.c
nicusorcitu: The static functions are all at the buttom of the file. `_vglite_signal_all_task_ready` is a static function called by another static function :) This is why I called it helper and marked as inline.

src/draw/sw/lv_draw_sw_private.h
nicusorcitu: Ok, do you want to change it now or later on in another PR after the 9.2 gets released?

src/draw/lv_draw.c
nicusorcitu: I believe we can leave it, this also saves some time.
Fill screen is the most consuming task. As all the tasks are handled by one single unit, I see no reason to wait for.

src/draw/lv_draw.c
nicusorcitu: So there are two possible configs:
1. VGLite draw unit only (best performance):
```
#define LV_USE_VGLITE_DRAW_THREAD 1
#define LV_DRAW_SW_DRAW_UNIT_CNT 0
```
E.g. `_draw_info.unit_cnt = 1`
This means that there are no dependency tasks at all - tasks should be queued and as long as the GPU is idle the tasks should be flushed to GPU.
The explanation is [here](https://github.com/lvgl/lvgl/pull/6646/files#r1706509069)

2. If both VGLite draw unit + SW draw unit are available (less performance):
```
#define LV_USE_VGLITE_DRAW_THREAD 1
#define LV_DRAW_SW_DRAW_UNIT_CNT 1
```
E.g. `_draw_info.unit_cnt = 2`
Then is should work as you described.

src/draw/lv_draw.c
kisvegabor: With the current code if there is only 1 draw unit but 100 draw tasks are queued this `if` won't run. and all the 100 draw tasks will be checked below.  

We should return in all cases if a screen-sized area is queued. Even if I'm mistaken it's too risky to add such a change on the day of the release.

src/draw/lv_draw.c
kisvegabor: Thanks!

src/draw/sw/lv_draw_sw_private.h
kisvegabor: Yes, please change it now. Otherwise people might start using it and it will be more difficult to change after the release.

src/draw/lv_draw.c
nicusorcitu: Ok, If we revert also this change then basically all the performance code is gone :)
But we can keep them locally at NXP and create another PR for LVGL 9.3.

The performance was relying on those two ifs - first one already reverted due to SW tests failures.

kisvegabor: Ah, I see the problem. Let's keep it local if possible. I'm afraid that it will negatively affect pure SW rendering performance but don't have time to test it correctly now.

nicusorcitu: Ok, I reverted all the dependency tasks and let them as before, done a separate commit to evidence them: 938e854266cb6c0367bd2e3130fa9ae697e98e30

We need to address it on the LVGL 9.3 as the difference on CPU usage is evident without the fix:
image

src/draw/sw/lv_draw_sw_private.h
nicusorcitu: Done.

src/draw/lv_draw.c
kisvegabor: I can't see any difference. It seems the `if`s aren't doing much difference.

kisvegabor:
Please note the conflict.

src/draw/lv_draw.c
nicusorcitu: I shrinked the image now, please look at the CPU usage.

kisvegabor: Okay, see now. So that `if` really helps lowering the performance.

kisvegabor:
Thank you! Merging now.

src/draw/lv_draw.c
nicusorcitu: Both `if`s from commit https://github.com/lvgl/lvgl/commit/938e854266cb6c0367bd2e3130fa9ae697e98e30 are lowering the CPU usage, so would like to address them in LVGl 9.3.
Thanks!

PR #6648, opened 2024-08-06
fix(arc): add missing private include

Pavi92:
include of ../../core/lv_obj_private.h was missing

### DIFF
--- a/src/core/lv_obj.c
+++ b/src/core/lv_obj.c
@@ -6,12 +6,12 @@
 /*********************
  *      INCLUDES
  *********************/
+#include "lv_obj_private.h"
 #include "../misc/lv_event_private.h"
 #include "../misc/lv_area_private.h"
 #include "lv_obj_style_private.h"
 #include "lv_obj_event_private.h"
 #include "lv_obj_class_private.h"
-#include "lv_obj_private.h"
 #include "../indev/lv_indev.h"
 #include "../indev/lv_indev_private.h"
 #include "lv_refr.h"
--- a/src/core/lv_obj_draw.c
+++ b/src/core/lv_obj_draw.c
@@ -12,6 +12,7 @@
 #include "../display/lv_display.h"
 #include "../indev/lv_indev.h"
 #include "../stdlib/lv_string.h"
+#include "../draw/lv_draw_arc.h"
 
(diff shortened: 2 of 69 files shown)


kisvegabor:
I wonder how could the CI compile it then? @liamHowatt could you take a deeper look?

Pavi92:
I think its because of the include order in lv_arc.c 

lv_obj_private.h is included before lv_arc_private.h and therefore lv_obj_t is known.

Pavi92:
> I wonder how could the CI compile it then? @liamHowatt could you take a deeper look?
I think its because of the include order in lv_arc.c

lv_obj_private.h is included before lv_arc_private.h and therefore lv_obj_t is known.

But when includig lv_arc.h in a user source file then the definition of lv_obj_t is not known because the include is missing in lv_arc_private.h

liamHowatt (dismissed):
Yes, that should be the cause. Thank you. I think it would be a good idea to:

1. All lv_x.c files with an lv_x_private.h should have the lv_x_private.h as the first include.
2. After building with the changes from 1, all the issues like the one you found should be revealed in the build errors.

I will do it after this PR, or you are welcome to do it in this PR but only if you want to.

<hr>

Also, for your change, I think the new include you added should go after `#include "lv_arc.h"` but it's not critical.

liamHowatt:
Wow, thanks for doing it!

After your changes, it was confirmed by the build system that widgets other than `arc` also had the missing `lv_obj_private.h` so I added all of those.

There was also a nasty circular dependency issue with `lv_draw.h`. It's because `lv_draw.h` includes `lv_draw_*.h` which include `lv_draw.h`.

Thank you very much!

@kisvegabor, you can see what I did in my commit to fix the circular dependency. Do you think it'd be better to have a `lv_draw_all.h` which includes all the draw primitive headers? Similar to `lv_types.h`.

Pavi92:
You're welcome

kisvegabor (commented):
I think it's fine as it is. It's unlikely that we add new draw task types in the near future. 

Thank you for both of you! 🙂

kisvegabor (approved):

FASTSHIFT (approved):

PR #6654, opened 2024-08-07
feat(cmake): Make LVGL version available for all CMake environments

tore-espressif:
LVGL version was available only if `custom.cmake` file was used.

The version variables were extracted to a separate file `version.cmake` that is then included in all environments .cmake files

**EDIT:** The LVGL_VERSION variable is also exported as env variable


tore-espressif:
cc @espzav @kisvegabor

FASTSHIFT (approved):

espzav (approved):
LGTM

XuNeo (approved):

PR #6664, opened 2024-08-09
feat(indev): add long press time setter

### LINKED ISSUE #6633: Let the user change long_press_time easily
### Problem to solve

Currently, the only way to change `long_press_time` is to edit `LV_INDEV_DEF_LONG_PRESS_TIME` [here](https://github.com/lvgl/lvgl/blob/master/src/indev/lv_indev.c#L38) (because `lv_indev_t` structure is private) or to rewrite the time logic like @kisvegabor suggested in #5821 

### Success criteria

The user must be able to change `long_press_time` without editing source code or rewriting time logic.

### Solution outline

1. As suggested in #5821, a `lv_obj_set_long_pressed_time` will allow to change `long_press_time` at runtime for each input device.

2. Another solution might be to move the define to the `lv_conf.h` which then do not increase API size but we can only change it for each device nor during execution (which is fine for me).

### Rabbit holes

None

### Testing

Write a unit test

### Teaching

Update the documentation

### Considerations

I will love it 😊

avilleret:
fix https://github.com/lvgl/lvgl/issues/6633

Add lv_indev_set_long_press_time method to set long press time to input device

### DIFF
--- a/src/indev/lv_indev.c
+++ b/src/indev/lv_indev.c
@@ -362,6 +362,8 @@ void lv_indev_set_display(lv_indev_t * indev, lv_display_t * disp)
 
 void lv_indev_set_long_press_time(lv_indev_t * indev, uint16_t long_press_time)
 {
+    if(indev == NULL) return;
+
     indev->long_press_time = long_press_time;
 }
 
--- a/src/indev/lv_indev.c
+++ b/src/indev/lv_indev.c
@@ -360,6 +360,11 @@ void lv_indev_set_display(lv_indev_t * indev, lv_display_t * disp)
     indev->disp = disp;
 }
 
+void lv_indev_set_long_press_time(lv_indev_t * indev, uint16_t long_press_time)
+{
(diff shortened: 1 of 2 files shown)


src/indev/lv_indev.c
```diff
@@ -360,6 +360,11 @@ void lv_indev_set_display(lv_indev_t * indev, lv_display_t * disp)
     indev->disp = disp;
 }
 
+void lv_indev_set_long_press_time(lv_indev_t * indev, uint16_t long_press_time)
+{
```
XuNeo, line 364: Other APIs have check of NULL pointer. We may add it here too?
```
    if(indev == NULL) return;
```

FASTSHIFT (approved):

src/indev/lv_indev.c
avilleret: done !

liamHowatt (approved):
Thank you!

FASTSHIFT (approved):

PR #6667, opened 2024-08-10
fix(rtthread): display driver hang

koendv:
The lvgl 9.1 st7789 driver hangs on rt-thread. 
The reason is that the lv_lcd_generic_mipi display driver uses lv_delay_ms() before clock ticks are running.
Solution is moving lv_tick_set_cb() before lv_port_disp_init().

Also, update argument list of logger to lvgl 9.1.


kisvegabor (approved):
Makes sense, thank you!

liamHowatt (approved):

PR #6675, opened 2024-08-13
fix(env_support/cmake): If LV_CONF_PATH is set, install the indicated config instead of the default one.

NicGrimpe:

When using CMake as build system, if `LV_CONF_PATH` is set, install the one pointed as public header instead of trying to install the config from the default location.


kisvegabor (approved):
Great catch! Thank you!

XuNeo (approved):

PR #6682, opened 2024-08-14
fix(draw/neon): fix build break

ghost:
Currently porting our software stack to LVGL v9.1.0. I noticed some problems that I couldn't trace back to bugs in our own code, so I switched to LVGL Git master. We are building LVGL as shared library for Linux.

The bump to Git master results in a build error when building the lib. The error was apparently introduced with refactoring in 1d14386b99cdff731cc3367d1d8b5e96ffb8fcea.

Since the reason for the refactoring was symbol visibility I tried to fix the build error, while keeping symbol visibility low.

More details in the commit message.

### Notes

### DIFF
--- a/src/draw/sw/blend/neon/lv_blend_neon.S
+++ b/src/draw/sw/blend/neon/lv_blend_neon.S
@@ -641,6 +641,7 @@ TMP_Q1      .qn  q14
 .thumb_func
 .func name
 .global name
+.hidden name
 name&:
     blender     src_bpp, dst_bpp, mask, opa, mode
 .endfunc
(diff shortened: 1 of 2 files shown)


kisvegabor:
Hi,

Thank you for working on it! As `lv_blend_neon.h` is not included only in C files, visibility is not an issue. So the `HIDDEN`-related changes can be removed. 

Could you measure the performance impact of neon? (I recommend using `lv_demo_benchmark`)

tobiasjakobi-compleo:
> Hi,
> 
> Thank you for working on it! As `lv_blend_neon.h` is not included only in C files, visibility is not an issue. So the `HIDDEN`-related changes can be removed.
> 
> Could you measure the performance impact of neon? (I recommend using `lv_demo_benchmark`)

Hi Gabor,

thanks for your reply. This is not correct though, the header is included in the various `src/draw/sw/blend/lv_draw_sw_blend_to_xyz.c` files. In particular, when omitting setting `.hidden` for the symbols, these internals then pop up as global symbols (because that's how they're created at the moment) in the shared library I build (can be verified either via nm or objdump).

As I understood the refactoring MR, this is undesirable since these symbols are internal to LVGL, and should not be called by a user.

With best wishes,
Tobias

kisvegabor:
That's strange. Okay, let's leave it as it is than. Just please fix the code formatting issue.

tobiasjakobi-compleo:
> That's strange. Okay, let's leave it as it is than. Just please fix the code formatting issue.

Should be fixed done. I've also removed the `Draft` prefix.

kisvegabor (approved):

kisvegabor:
Thank you! Do you have any info about the performance change?

tobiasjakobi-compleo:
> Thank you! Do you have any info about the performance change?

You mean performances changes introduces by these changes? I don't think there are any.

Or do you mean performance SW vs. NEON? I have yet to test our application on a real charging station. I'm currently running everything inside our simulated environment on our x86-64 development systems. I plan to do first tests on real hw once things look good in the simulator (which they sadly don't at the moment :sob: )

liamHowatt (approved):
The refactoring PR removed the leading underscore from the `static inline` functions which caused them to collide with the assembly symbol names. Sorry for missing that.

You fixed it by adding an underscore to the assembly names.

The visibility change you did basically has the same effect as `static` because it makes the symbols not accessible in an ELF/SO, however unlike `static` they are still visible between compilation units before linking. Right?

kisvegabor:
> I plan to do first tests on real hw once things look good in the simulator (which they sadly don't at the moment 😭 )

Are these due to LVGL issues or just need more time to implement the UI?

tobiasjakobi-compleo:
> The visibility change you did basically has the same effect as `static` because it makes the symbols not accessible in an ELF/SO, however unlike `static` they are still visible between compilation units before linking. Right?

Correct. My understand of the refactoring MR is that neither the symbols generated by the assembler code, nor the helper wrappers are public API, and hence should both be hidden.

Since there is no way to tell the assembler "this is supposed to be a static function", we need to go through the symbol visibility here. Static would not make any sense here, since the static keyword in C applies to one single translation unit. If I were to apply this concept to the assembler source file, when I would never be able to call anything from that assembler source file.

tobiasjakobi-compleo:
> > I plan to do first tests on real hw once things look good in the simulator (which they sadly don't at the moment 😭 )
> 
> Are these due to LVGL issues or just need more time to implement the UI?

Yeah, these two are currently blockers for us:
* https://github.com/lvgl/lvgl/issues/6685
* https://github.com/lvgl/lvgl/issues/6684

I need to recheck these with a fresh Git master if I find the time. Currently working on a different project again.

liamHowatt:
> Correct. My understand of the refactoring MR is that neither the symbols generated by the assembler code, nor the helper wrappers are public API, and hence should both be hidden.
> 
> Since there is no way to tell the assembler "this is supposed to be a static function", we need to go through the symbol visibility here. Static would not make any sense here in the first place, since the static keyword in C applies to one single translation unit. If I were to apply this concept to the assembler source file, when I would never be able to call anything from that assembler source file.

Right, static is not the right tool. There are already many functions in LVGL that should never be called from an application but they are not hidden like these assembly routines have been. I think we should not apply `hidden` to too many future things unless we decide that's a pattern we want to standardize.

kisvegabor:
>  unless we decide that's a pattern we want to standardize.

I'm thinking the same way. We should be careful with introducing new patterns. 

We are still new to assembly embedding and learning to potential issues. I'm sure that soon we will iron out these kind of issues and have stable reference for the new implementation.

PR #6686, opened 2024-08-14
feat(freetype): add colored glyphs support

Moschn:
### Freetype: Support Colored Bitmap Glyphs

Fonts often come with bitmap glyphs, e.g. for emojis, and Freetype already supports these glyphs. This PR enables LVGL to receive colored bitmaps from Freetype and draw them. 

Since this is my first contribution to LVGL, I am probably making many mistakes. I am open for feedback. 

Notes:

- The drawing of ARGB888 glyphs is not yet define guarded. This should probably be guarded, but I was not sure which one to choose. Also, the code is currently copy pasted from `LV_FONT_GLYPH_FORMAT_IMAGE`. Ideally, we would not duplicate code, but I did not figure out a way to set the `glyph_draw_dsc->format` correctly in `lv_freetype_image.c`. Let me know if you know of a way to correctly tag a glyph in `lv_freetype_image.c` with `LV_FONT_GLYPH_FORMAT_IMAGE`.
- Documentation could also be updated to help users use this new feature. I added an example already, but was not sure if I should also add documentation (and where).
- Ideally, LVGL would also support drawing SVG colored glyphs as those can easily be scaled. I believe many emoji fonts such as twemoji and newer versions of NotoColoredEmoji use scalable glyphs in this way. Adding such a feature would require quite a bit of knowledge in LVGL and Freetype though.


src/libs/freetype/lv_freetype_glyph.c
```diff
@@ -136,8 +136,19 @@ static bool freetype_glyph_create_cb(lv_freetype_glyph_cache_data_t * data, void
...
+    else {
+        LV_LOG_WARN("font is not scalable, selecting available size\n");
+        error = FT_Select_Size(face, 0);
+    }
+    if(error) {
+        FT_ERROR_MSG("FT_Set_Pixel_Sizes", error);
+        lv_mutex_unlock(&dsc->cache_node->face_lock);
+        return false;
+    }
+    error = FT_Load_Glyph(face, glyph_index,  FT_LOAD_COMPUTE_METRICS | FT_LOAD_NO_AUTOHINT);
```
W-Mai: You can change the g_dsc->format = LV_FONT_GLYPH_FORMAT_IMAGE

W-Mai:
image

Sorry but I can't reproduce your code. I got a error message like this:

image

image

So, what configuration information did I miss?

W-Mai:
Well, when I open this option, it works well

image

image

examples/libs/freetype/lv_example_freetype_2.c
```diff
@@ -0,0 +1,62 @@
...
+    static lv_style_t style;
+    lv_style_init(&style);
+    lv_style_set_text_font(&style, font);
+    lv_style_set_text_align(&style, LV_TEXT_ALIGN_CENTER);
+
+    /*Create a label with the new style*/
+    lv_obj_t * label = lv_label_create(lv_screen_active());
+    lv_obj_add_style(label, &style, 0);
+    lv_label_set_text(label, "Hello world\nI'm a font created with FreeType 😀");
+    lv_obj_center(label);
```
kisvegabor, line 47: Please add something similar [as a test](https://github.com/lvgl/lvgl/blob/master/tests/src/test_cases/libs/test_freetype.c) as well.

kisvegabor (commented):
It's a great update! :rocket: 

@W-Mai can comment more on the technical details, I can point out some high-level things:
- Please add a test
- Please add a paragraph about it [in the docs](https://github.com/lvgl/lvgl/blob/master/docs/libs/freetype.rst)

src/libs/freetype/lv_freetype_glyph.c
Moschn: Thanks for the hint. So the format can be selected when loading the glyph in `lv_freetype_glyph`, but at that point, I am not sure if we know yet if it is a colored bitmap. 

TBH, I did not yet fully understand the flow. First the glyph is loaded:
https://github.com/lvgl/lvgl/blob/d7b82c79668d7b31d52f6535360eff15396eade5/src/libs/freetype/lv_freetype_glyph.c#L140
but then the glyph is loaded again but with different flags?
https://github.com/lvgl/lvgl/blob/d7b82c79668d7b31d52f6535360eff15396eade5/src/libs/freetype/lv_freetype_image.c#L131
Adding prints or using GDB shows that both paths are being hit when rendering a freetype font.

Moschn:
> Well, when I open this option, it works well
> image image

Oh very interesting. I compiled freetype2 from upstream but I did not change any of its configuration. I will add a log message that points to this flag such that users can find it more easily.

@kisvegabor Thanks for the pointers. I will add a test and documentation.

Moschn:
I added some documentation, and added test coverage.

I now just tagged all bitmap glyphs from FreeType to be `LV_FONT_GLYPH_FORMAT_IMAGE`. This seems to work.

Drawing a letter as an image is guarded by `LV_USE_IMGFONT`. I guess the feature in this PR should work even when `LV_USE_IMGFONT=0`. Therefore, I removed the define guard in `src/draw/sw/lv_draw_sw_letter.c`. I could also add `|| LV_FREETYPE` to the define guard. What do you prefer?

tests/src/test_files/fonts/noto/NotoColorEmoji-32.subset.ttf
W-Mai, line 1: One copy is ok, other path can be loaded from a relative path

Moschn:
I removed the duplicated font files. Any further comments?

tests/src/test_cases/libs/test_freetype.c
```diff
@@ -456,6 +466,16 @@ void test_freetype_bitmap_rendering_test(void)
     lv_label_set_text(label2, UNIVERSAL_DECLARATION_OF_HUMAN_RIGHTS_JP);
     lv_obj_align_to(label2, label1, LV_ALIGN_OUT_BOTTOM_MID, 0, 10);
 
+    // test emoji rendering
+    // emoji font does not contain normal characters, use fallback to render them
```
kisvegabor:
```suggestion
    /* test emoji rendering
     * emoji font does not contain normal characters, use fallback to render them */
```

examples/libs/freetype/lv_example_freetype_2.c
```diff
@@ -0,0 +1,62 @@
...
+{
+    /*Create a font*/
+    lv_font_t * font = lv_freetype_font_create(PATH_PREFIX "lvgl/examples/libs/freetype/Lato-Regular.ttf",
+                                               LV_FREETYPE_FONT_RENDER_MODE_BITMAP,
+                                               24,
+                                               LV_FREETYPE_FONT_STYLE_NORMAL);
+
+    // this font is created from a downscaled NotoColorEmoji to 34x32px
+    // Subset containing only a single emoji was created using fonttools:
+    // Command: fonttools subset NotoColorEmoji.ttf --text=😀
```
kisvegabor:
```suggestion
    /* this font is created from a downscaled NotoColorEmoji to 34x32px
     * Subset containing only a single emoji was created using fonttools:
     * Command: fonttools subset NotoColorEmoji.ttf --text=😀 */
```

kisvegabor (dismissed):
It looks good to me, thank you! 

I've just changed `//` to  `/* */`

src/libs/freetype/lv_freetype.c
```diff
@@ -161,7 +161,18 @@ lv_font_t * lv_freetype_font_create(const char * pathname, lv_freetype_font_rend
...
+        error = FT_Set_Pixel_Sizes(face, 0, size);
+    }
+    else {
+        LV_LOG_WARN("font is not scalable, selecting available size\n");
+        error = FT_Select_Size(face, 0);
+    }
+    if(error) {
+        FT_ERROR_MSG("FT_Set_Pixel_Sizes", error);
+        return false;
+    }
```
W-Mai, line 175:
```suggestion
    else {
        LV_LOG_WARN("font is not scalable, selecting available size");
        error = FT_Select_Size(face, 0);
    }
    if(error) {
        FT_ERROR_MSG("FT_Set_Pixel_Sizes", error);
        return NULL;
    }
```

W-Mai:
```suggestion
    else {
        LV_LOG_WARN("font is not scalable, selecting available size");
        error = FT_Select_Size(face, 0);
    }
    if(error) {
        FT_ERROR_MSG("FT_Set_Pixel_Sizes", error);
        return NULL;
    }
```

W-Mai (changes requested):

src/libs/freetype/lv_freetype_glyph.c
```diff
@@ -136,8 +136,18 @@ static bool freetype_glyph_create_cb(lv_freetype_glyph_cache_data_t * data, void
...
+    }
+    else {
+        error = FT_Select_Size(face, 0);
+    }
+    if(error) {
+        FT_ERROR_MSG("FT_Set_Pixel_Sizes", error);
+        lv_mutex_unlock(&dsc->cache_node->face_lock);
+        return false;
+    }
+    error = FT_Load_Glyph(face, glyph_index,  FT_LOAD_COMPUTE_METRICS | FT_LOAD_NO_AUTOHINT);
```
W-Mai: I'm not sure FT_LOAD_NO_BITMAP will affect the load efficiency.

I suggest to keep original codes and add a new `if` branch to handle bitmap render case.

W-Mai:
```cpp
typedef enum {
    LV_FREETYPE_FONT_RENDER_MODE_BITMAP = 0,
    LV_FREETYPE_FONT_RENDER_MODE_OUTLINE = 1,
} lv_freetype_font_render_mode_t;
```

You can add this enum as you wish. Like `LV_FREETYPE_FONT_RENDER_MODE_IMAGE`.

Do like this will keep some branch more clearer I think.

Moschn:
Thanks for the feedback. I added the changes that you suggested. 

I did not see how adding an extra entry into `lv_freetype_font_render_mode_t` would help simplify some branches/code. But if you have a particular branch, I would be happy to refactor that with a new entry in `lv_freetype_font_render_mode_t`.

W-Mai:
> Thanks for the feedback. I added the changes that you suggested.
> 
> I did not see how adding an extra entry into `lv_freetype_font_render_mode_t` would help simplify some branches/code. But if you have a particular branch, I would be happy to refactor that with a new entry in `lv_freetype_font_render_mode_t`.

Now is perfect 😁

W-Mai (approved):

kisvegabor (approved):

liamHowatt (approved):
Nice! Looks great to me!

PR #6691, opened 2024-08-16
feat(dma2d): add basic support

liamHowatt:

### Notes


faxe1008:
@liamHowatt i started working on this as well some time ago:
https://github.com/faxe1008/lvgl/commit/4893a282d530f6474a7e59b372c8fafc78768879
But I do not own a STM board with the ChromArt engine and also did not have much time to make progress. Feel free to take stuff from there, would love to see this getting in :^)

src/draw/dma2d/lv_draw_dma2d.c
```diff
@@ -0,0 +1,569 @@
...
+#include "lv_draw_dma2d.h"
+#if LV_USE_DRAW_DMA2D
+
+#include "../lv_draw_private.h"
+#include "../sw/lv_draw_sw.h"
+#include "../../misc/lv_area_private.h"
+
+#include "stm32h7xx_hal.h"
+#include "stm32h7xx_hal_dma2d.h"
+#include "stm32h7b3i_discovery_lcd.h"
```
kisvegabor: We need more generic header here. E.g. `F7`, `F4` and many other also have DMA2D. Dev board related headers (like `stm32h7b3i_discovery_lcd.h`) should be also avoided.

src/draw/dma2d/lv_draw_dma2d.c
```diff
@@ -0,0 +1,569 @@
...
+} lv_draw_dma2d_unit_t;
+
+/**********************
+ *  STATIC PROTOTYPES
+ **********************/
+
+static int32_t evaluate_cb(lv_draw_unit_t * draw_unit, lv_draw_task_t * task);
+static int32_t dispatch_cb(lv_draw_unit_t * draw_unit, lv_layer_t * layer);
+static int32_t delete_cb(lv_draw_unit_t * draw_unit);
+static void opaque_fill(void * first_pixel, int32_t w, int32_t h, int32_t stride, lv_color_t color,
```
kisvegabor: Probably just personal preference, but I'd add `lv_draw_dma2d_unit_t * u` as the first parameter.

src/draw/dma2d/lv_draw_dma2d.c
```diff
@@ -0,0 +1,569 @@
...
+    g_unit = draw_dma2d_unit;
+
+#if LV_USE_DRAW_DMA2D_INTERRUPT
+    lv_result_t res = lv_thread_sync_init(&draw_dma2d_unit->interrupt_signal);
+    LV_ASSERT(res == LV_RESULT_OK);
+
+    res = lv_thread_init(&draw_dma2d_unit->thread, LV_THREAD_PRIO_HIGH, thread_cb, 2 * 1024, draw_dma2d_unit);
+    LV_ASSERT(res == LV_RESULT_OK);
+#else
+    lv_result_t res = lv_thread_sync_init(&draw_dma2d_unit->thread_begin_polling_signal);
```
kisvegabor: I suggest following [this pattern](https://github.com/lvgl/lvgl/blob/d7b82c79668d7b31d52f6535360eff15396eade5/src/draw/renesas/dave2d/lv_draw_dave2d.c#L404-L412) as it's already proven to be work in many environments. (Even if it's also not perfect)

src/draw/dma2d/lv_draw_dma2d.c
```diff
@@ -0,0 +1,569 @@
...
+
+    task->preferred_draw_unit_id = DRAW_UNIT_ID_DMA2D;
+    task->preference_score = 0;
+
+    return 0;
+}
+
+static int32_t dispatch_cb(lv_draw_unit_t * draw_unit, lv_layer_t * layer)
+{
+    lv_draw_dma2d_unit_t * draw_dma2d_unit = (lv_draw_dma2d_unit_t *) draw_unit;
```
kisvegabor: We should check if the `draw_unit` is busy. See [here](https://github.com/lvgl/lvgl/blob/d7b82c79668d7b31d52f6535360eff15396eade5/src/draw/renesas/dave2d/lv_draw_dave2d.c#L344)

src/draw/dma2d/lv_draw_dma2d.c
```diff
@@ -0,0 +1,569 @@
...
+                | DMA2D_CR_TCIE
+#endif
+                ;
+
+#if !LV_USE_DRAW_DMA2D_INTERRUPT
+#if LV_USE_OS
+    lv_result_t res = lv_thread_sync_signal(&u->thread_begin_polling_signal);
+    LV_ASSERT(res == LV_RESULT_OK);
+#else
+    await_transfer_completion();
```
kisvegabor:
```suggestion
    /*Blocking wait for DMA2D completion*/
    await_transfer_completion();
```

src/draw/dma2d/lv_draw_dma2d.c
```diff
@@ -0,0 +1,569 @@
...
+{
+    lv_draw_dma2d_unit_t * u = arg;
+
+    while(1) {
+
+#if LV_USE_DRAW_DMA2D_INTERRUPT
+        lv_thread_sync_wait(&u->interrupt_signal);
+#else
+        lv_thread_sync_wait(&u->thread_begin_polling_signal);
+        await_transfer_completion();
```
kisvegabor: Is this scenario makes sense? It will spin a wile loop and we are the mercy of the OS to run other tasks while we are waiting here. Although it's a [low priority thread](https://github.com/lvgl/lvgl/pull/6691/files?diff=unified&w=0#diff-4a9425813cbac916944f0efcf3738f379cf07d3b44e1ed51558125096db3b61cR167) I wouldn't play gamble.

kisvegabor (commented):
I feel like the OS/interrupt combinations are a little bit messy. So we have these cases:
1. **No interrupt, no OS**: we can start DMA2D, return immediately, wait for completion at the beginning of `dispatch_cb`. This way at least LVGL can prepare the next task or do some SW rendering. 
2. **No Interrupt, OS**: It's strange to me because of the blocking `while` loop in the thread. I'd use case 1) here too as it's more predictable
3. **Interrupt, no OS**: We can't do much here, fallback to 1)
4. **Interrupt, OS**: It's really useful. Your implementation works here.

So as I see it can be this simple:
- Has OS **and** interrupt? If yes, do 4)  
- Otherwise go for 1)

Please also add a brief docs about it [here](https://docs.lvgl.io/master/integration/chip/stm32.html). Just 1-2 paragraph about the options. 

Please also add support for image belnding.

**Interrupts**: cc @faxe1008 
Is [this](https://github.com/lvgl/lvgl/pull/6691/files?diff=unified&w=0#diff-0be6707528460cdc1dd90dad35b2597054e790c6b811c3ed5f9d3ce41c105ed0R249-R250) doable in Zephyr? In general what we can assume about the interrupt config if Zephyr used?

src/draw/dma2d/lv_draw_dma2d.c
liamHowatt: If `lv_os.h` had something like `lv_os_sched_yield` or similar, we could check once and yield to other tasks. Scheduler yielding is available in most [RT]OSs, [even Linux](https://man7.org/linux/man-pages/man2/sched_yield.2.html) (although, unfortunately that isn't under the pthread umbrella :confused:).

I know the low priority isn't a reliable way to do it either. See my next comment in the main thread.

liamHowatt:
>    1. **No interrupt, no OS**: we can start DMA2D, return immediately, wait for completion at the beginning of `dispatch_cb`. This way at least LVGL can prepare the next task or do some SW rendering.

I didn't realize that was possible. I shall update it. I see how this captures 2 and 3 :+1:

liamHowatt:
Image blending is WIP.

liamHowatt:
I still need to fix the HAL header also.

liamHowatt:
Image support is working well @kisvegabor. Coordinates and blending are correct for `lv_demo_benchmark`. I see that there is an issue with `lv_demo_music`. I don't imagine it will be a complex fix.

All feedback is addressed. The OS/interrupt combinations are down to two possibilities, there are docs, and the HAL header is generalized.

I have not tested with OS enabled. If you're able to validate it then great. If there is an issue I will use OS and debug.

I'm testing with an STM32H7 on this branch: https://github.com/liamHowatt/lv_port_stm32h7b3i_disco/tree/feat/v9-dma2d

lhdjply:
Isn't it too messy to have everything in one C file? Can you split it into lv_draw_dma2d_fill.c, lv_draw_dma2d_img.c, etc.?

liamHowatt:
> Isn't it too messy to have everything in one C file? Can you split it into lv_draw_dma2d_fill.c, lv_draw_dma2d_img.c, etc.?

The [v8.4 DMA2D driver](https://github.com/lvgl/lvgl/tree/release/v8.4/src/draw/stm32_dma2d) implemented everything in one file but I agree that it's good to set a good example for other draw drivers so I have now moved the fill and image tasks into their own files :+1:

Unrelated: DMA2D can support L8. Maybe that's something we can add support for in a future PR.

lhdjply:
If possible, could you change 'image' to 'img'? This would maintain naming consistency with the other drivers.

kissa96:
DMA2D feature has been tested. These are the benchmark results, comparing them with basic sw.

| Name                      | Software [ms] | DMA2D [ms] |
|---------------------------|----------|-------|
| All                       | 6        | 5     |
| Empty screen              | 2        | 2     |
| Moving wallpaper          | 4        | 2     |
| Single rectangle          | 0        | 0     |
| Multiple rectangles       | 2        | 2     |
| Multiple RGB images       | 0        | 0     |
| Multiple ARGB images      | 3        | 1     |
| Rotated ARGB images       | 31       | 30    |
| Multiple labels           | 1        | 1     |
| Screen sized text         | 25       | 25    |
| Multiple arcs             | 0        | 0     |
| Containers                | 2        | 1     |
| Containers with overlay   | 7        | 4     |
| Containers with opa       | 3        | 2     |
| Containers with opa_layer | 6        | 5     |
| Containers with scrolling | 8        | 6     |
| Widgets demo              | 10       | 9     |

src/draw/dma2d/lv_draw_dma2d.c
```diff
@@ -0,0 +1,461 @@
...
+{
+    lv_draw_dma2d_unit_t * draw_dma2d_unit = (lv_draw_dma2d_unit_t *) draw_unit;
+
+    if(draw_dma2d_unit->task_act) {
+#if LV_DRAW_DMA2D_ASYNC
+        /*Return immediately if it's busy with draw task*/
+        return 0;
+#else
+        if(!check_transfer_completion()) {
+            return -1;
```
kisvegabor:
```suggestion
            return LV_DRAW_UNIT_IDLE;
```

src/draw/dma2d/lv_draw_dma2d_private.h
```diff
@@ -0,0 +1,155 @@
...
+ *      DEFINES
+ *********************/
+
+#if LV_USE_DRAW_DMA2D_INTERRUPT && LV_USE_OS
+#define LV_DRAW_DMA2D_ASYNC 1
+#else
+#define LV_DRAW_DMA2D_ASYNC 0
+#endif
+
+#if defined(__CORTEX_M) && (__CORTEX_M == 7)
```
kisvegabor, line 33: It's not future proof but okay for now.

kisvegabor (commented):
I was here with @kissa96 and it looks good to me as well. We haven't tested it with FrreRTOS though.

src/draw/dma2d/lv_draw_dma2d_private.h
liamHowatt: Would `#ifdef __SCB_DCACHE_LINE_SIZE` be better?

kisvegabor: I have no idea to be honest

liamHowatt:
I fixed the music demo issue. It was a case when the image header stride is `0`.

I tried and failed to use FreeRTOS with lv_port_stm32h7b3i_disco, so OS is still untested.

kisvegabor (approved):
Thank you!

I'm still not sure about how to handle the interrupt hook, but we can improve it later based on feedback. (before v9.3)

kisvegabor:
@XuNeo @FASTSHIFT 
We have tested this updated and all worked well. Could you also review and approve it?

GPTKEY:
Regarding the adjustment of DMA2D acceleration, it may be related to various factors, one is the block size of this update, if it is too small, the interrupt method of enabling DMA2D will play a reverse role, so it is necessary to set the minimum size of a single update elsewhere in order to get the optimal solution, or make a judgement before calling DMA2D, if the amount of data is more than the threshold and then enable DMA2D.

FASTSHIFT (approved):

kisvegabor:
Let's merge it without this optimization and improve it in a next step.

To find the optimal size we can so some experiments, like changing 100 x 1px, 10 x 10px, 1 x 100px and measure the execution time.

tdjastrzebski:
I just started another project using 1024x600 display where use of DMA2D makes even more sense and I am willing to use LVGL rev 9.x. After switching DMA2D on (`#define LV_USE_DRAW_DMA2D 1`) I get errors like below. Probably some files are malformed.

image

kisvegabor:
I think it's because you have an include in lv_conf.h. Does it work if you remove it?

tdjastrzebski:
Ach, that could be. My build system hates things like
```C
#define LV_DRAW_DMA2D_HAL_INCLUDE "stm32u5xx_hal.h"`
```
since it makes static dependency analysis much harder

liamHowatt:
I was thinking of using `__has_include` in lv_draw_dma2d.c to detect which header is available. Not all compilers support it, but what that matters is that all compilers that target STM32 support it. Looks like the IAR compiler does not have it. LVGL depends on `__has_include` for a couple of things that aren't critical.

Maybe it could try to use `__has_include`, but if it's not available, require the `LV_DRAW_DMA2D_HAL_INCLUDE` to be defined?

Open to ideas.

tdjastrzebski:
Why "stm32h7xx_hal.h" is hardcoded in `lv_conf_internal.h` and `lv_conf_cmsis.h` as well?
I am not saying it is a bug, just trying to understand the new 9.x setup.
This may seem too radical, but how about just including "main.h"?

liamHowatt:
Oh, `lv_conf_internal.h` just applies defaults in case they aren't defined in your `lv_conf.h`. Set it in your `lv_conf.h`. If your `lv_conf.h` doesn't have those configs yet, copy them from `lv_conf_template.h`. See [here](https://github.com/lvgl/lvgl/blob/559618d9d2c66069f61ba673b7bce6bf3e7eb036/lv_conf_template.h#L270-L277).

tdjastrzebski:
I think have correct config, it all builds with `LV_USE_DRAW_DMA2D 1` and even works.
Does `LV_USE_DRAW_DMA2D_INTERRUPT` require some RTOS?
Is it possible for the rendering pipeline to simply `await` and continue on DMA2D interrupt (DMA2D_IRQHandler)
I mean, sort of using APM async pattern.

tdjastrzebski:
Btw ST is going to release some GPU2D docs. Read [here](https://community.st.com/t5/stm32-mcus-touchgfx-and-gui/neochrom-gpu2d-registers-documentation/m-p/618737#M35451). Is there any work in progress aimed on LVGL impl?

tdjastrzebski:
One more observation for today. With DMA2D enabled LVGL 9.3 appears to be **MUCH** slower than 8.4.
I cannot say exactly how much, but 9.3 perf test stops at "John Smith A DYI enthusiast" after 45 s and does not move any further.
8.4 perf test with DMA2D shows weighted 28 FPS and it is 1024x600 display. I would say, not a bad result.
It seems that 9.3 uses DMA2D in a very limited fashion.
IMG_9457

kisvegabor:
To compare v8 and v9 please use this branch fro v8: https://github.com/lvgl/lvgl/tree/v8.3-with-v9-benchmark
It's the v9 benchmark backported to v8.

Please also compare the performance with pure SW rendering first.

tdjastrzebski:
With 8.3 rev new demo_benchmark stops at the same stage: "containers with opa_layer" after 35 s vs 45 s in case of rev 9.3.
Unlikely it is a memory constraint problem, excluding 1800k LTDC buffer defined as a separate linker script section, STM32U5G9 has 1208k SRAM left.

kisvegabor:
Please set LV_MEM_SIZE to 150kB

tdjastrzebski:
Thanks, increasing LV_MEM_SIZE helped. Here are the results.
Btw, is there any standard way to 'printf' those numbers upon test finish?
8.3 + DMA2D
IMG_9461
IMG_9462
8.3 without DMA2D
IMG_9463
IMG_9464
9.3 + DMA2D
IMG_9465
IMG_9466
9.3 without DMA2D
IMG_9467
IMG_9468

kisvegabor:
Wooow! These tell a lot
- v8.3 no DMA2D 
   - Empty screen: 45+10 ms
   - Movin wallpaper: 137+8 ms
- v9 no DMA2D 
   - Empty screen: 76+8 ms
   - Movin wallpaper: 422+10 ms

These are the simplest task, that we can't really mess up. Just a memfill and a memcpy. If these are slow usually the bottle neck is the memory bandwidth. So is there any change in the 
- framebuffer location
- LVGL's rendering mode (e.g. partial vs full)
-  compiler settings
-  lv_conf settings (e.g. color depth or asserts)?

You can increase the length of the Moving wallpaper scene and pause the debugger multiple times. Do you see that it's usually spinning in a "blend" function?

> Btw, is there any standard way to 'printf' those numbers upon test finish?

If logging is enabled these should be printed too.

tdjastrzebski:
Nope, buffer location, mode, sizes and method used to transfer pixels from lvgl buffer to ltdc buffer is exactly the same.
Compiler settings are the same as well. I will keep investigating

tdjastrzebski:
Something tells me the problem is not memcpy() because for 9.3 "empty screen" DMA2D improves performance by 100%, but render time it is still twice longer than with 8.3.

<html xmlns:v="urn:schemas-microsoft-com:vml"
xmlns:o="urn:schemas-microsoft-com:office:office"
xmlns:x="urn:schemas-microsoft-com:office:excel"
xmlns="http://www.w3.org/TR/REC-html40">

<head>

<meta name=ProgId content=Excel.Sheet>
<meta name=Generator content="Microsoft Excel 15">
<link id=Main-File rel=Main-File
href="file:///C:/Users/Thomas/AppData/Local/Temp/msohtmlclip1/01/clip.htm">
<link rel=File-List
href="file:///C:/Users/Thomas/AppData/Local/Temp/msohtmlclip1/01/clip_filelist.xml">
<style>

</style>
</head>

<body link="#0563C1" vlink="#954F72">


8.3 -   DMA2D |   |   |   |   |  
-- | -- | -- | -- | -- | --
Name | Avg. CPU | Avg. FPS | Avg. time | render time | flush time
Empty screen | 99% | 20 | 55 | 46 | 9
Moving wallpaper | 100% | 8 | 146 | 136 | 10
Single rectangle | 35% | 33 | 6 | 6 | 0
Multiple rectangles | 92% | 20 | 46 | 42 | 4
Multiple RGB images | 100% | 10 | 112 | 104 | 8
Multiple ARGB images | 100% | 6 | 239 | 230 | 9
Rotated ARGB images | 100% | 3 | 1139 | 1122 | 17
Multiple labels | 89% | 13 | 159 | 152 | 7
Screen sized text | 15% | 33 | 18 | 18 | 0
Multiple arcs | 99% | 6 | 176 | 168 | 8
Containers | 100% | 6 | 169 | 166 | 3
Containers with overlay | 100% | 3 | 308 | 301 | 7
Containers with opa | 100% | 3 | 292 | 288 | 4
Containers with opa_layer | 74% | 32 | 17 | 15 | 2
Containers with scrolling | 98% | 3 | 303 | 293 | 10
Widgets demo | 97% | 4 | 277 | 271 | 6
All scenes avg. | 87% | 12 | 215 | 209 | 6
  |   |   |   |   |  
8.3 + DMA2D |   |   |   |   |  
Name | Avg. CPU | Avg. FPS | Avg. time | render time | flush time
Empty screen | 77% | 33 | 24 | 16 | 8
Moving wallpaper | 95% | 20 | 55 | 47 | 8
Single rectangle | 11% | 33 | 0 | 0 | 0
Multiple rectangles | 48% | 33 | 15 | 6 | 9
Multiple RGB images | 97% | 16 | 50 | 44 | 6
Multiple ARGB images | 100% | 16 | 51 | 41 | 10
Rotated ARGB images | 100% | 3 | 1094 | 1082 | 12
Multiple labels | 85% | 17 | 91 | 82 | 9
Screen sized text | 11% | 33 | 7 | 7 | 0
Multiple arcs | 96% | 7 | 131 | 124 | 7
Containers | 100% | 10 | 84 | 81 | 3
Containers with overlay | 100% | 10 | 93 | 90 | 3
Containers with opa | 100% | 10 | 92 | 88 | 4
Containers with opa_layer | 51% | 34 | 4 | 3 | 1
Containers with scrolling | 98% | 6 | 171 | 164 | 7
Widgets demo | 96% | 7 | 158 | 151 | 7
All scenes avg. | 79% | 18 | 131 | 126 | 5
  |   |   |   |   |  
9.3 - DMA2D |   |   |   |   |  
Name | Avg. CPU | Avg. FPS | Avg. time | render time | flush time
Empty screen | 99% | 11 | 84 | 76 | 8
Moving wallpaper | 100% | 2 | 431 | 427 | 4
Single rectangle | 43% | 27 | 12 | 12 | 0
Multiple rectangles | 94% | 12 | 72 | 68 | 4
Multiple RGB images | 100% | 2 | 342 | 330 | 12
Multiple ARGB images | 100% | 2 | 331 | 318 | 13
Rotated ARGB images | 100% | 0 | 1337 | 1320 | 17
Multiple labels | 95% | 6 | 203 | 195 | 8
Screen sized text | 23% | 26 | 98 | 97 | 1
Multiple arcs | 100% | 3 | 282 | 276 | 6
Containers | 100% | 2 | 268 | 264 | 4
Containers with overlay | 100% | 1 | 691 | 680 | 11
Containers with opa | 100% | 1 | 519 | 510 | 9
Containers with opa_layer | 100% | 1 | 613 | 602 | 11
Containers with scrolling | 100% | 1 | 543 | 532 | 11
Widgets demo | 100% | 1 | 563 | 555 | 8
All scenes avg. | 90% | 6 | 398 | 391 | 7
  |   |   |   |   |  
9.3 + DMA2D |   |   |   |   |  
Name | Avg. CPU | Avg. FPS | Avg. time | render time | flush time
Empty screen | 99% | 22 | 41 | 32 | 9
Moving wallpaper | 100% | 2 | 389 | 375 | 14
Single rectangle | 35% | 28 | 5 | 5 | 0
Multiple rectangles | 69% | 26 | 30 | 26 | 4
Multiple RGB images | 98% | 7 | 120 | 112 | 8
Multiple ARGB images | 100% | 7 | 121 | 115 | 6
Rotated ARGB images | 100% | 0 | 1304 | 1297 | 7
Multiple labels | 95% | 7 | 138 | 132 | 6
Screen sized text | 21% | 26 | 89 | 88 | 1
Multiple arcs | 91% | 3 | 221 | 213 | 8
Containers | 100% | 3 | 203 | 198 | 5
Containers with overlay | 100% | 3 | 214 | 212 | 2
Containers with opa | 100% | 2 | 421 | 416 | 5
Containers with opa_layer | 100% | 1 | 547 | 541 | 6
Containers with scrolling | 100% | 2 | 454 | 444 | 10
Widgets demo | 100% | 1 | 542 | 534 | 8
All scenes avg. | 88% | 8 | 302 | 296 | 6



</body>

</html>

kisvegabor:
Something is not good here:
- v8 no DMA
  - Screen sized text	15% CPU: 18ms -> Seems too fast. Here a screen sized text should be scrolled
- v9 no DMA
  - Screen sized text	23% CPU: 97ms -> More realistic, but long test rendering is 2-3 times faster in v9 (according to other measurement) and the CPU usage should be 100% here.

How does this screen look like on device? Is the text long enough for scrolling? How do you call the `lv_tick_inc()`

Please enable `LV_USE_REFR_DEBUG` it will draw semi-transparent rectangles on the invalidated area. Due to this performance will be way worse, but you can compare if v8 and v9 are updating the same areas.  (especially in case of the "Screen sized text")

tdjastrzebski:
@kisvegabor I responded under #6709 since this performance problem likely has nothing to do with DMA2D.
see: https://github.com/lvgl/lvgl/issues/6709#issuecomment-2354921070

PR #6692, opened 2024-08-17
fix(freertos): notifiy-based sync can be initialized from any task

### LINKED ISSUE #6414: lv_timer_handler hangs when running on separate thread on ESP32 using FREERTOS using Arduino framework
### LVGL version

9.1.0

### What happened?

When I call lv_timer_handler from the main loop (with or without a mutex protection everything works.

When I create a separate thread using xTaskCreatePinnedToCore (on either core) with or without a mutex protection and this task calls the method lv_timer_handler this method never returns (it actually hangs in: refr_invalid_areas(); which again hangs in: lv_thread_sync_wait(&_draw_info.sync); in the lv_draw.c file.

### How to reproduce?

//This is the task definition:
void ProcessDisplay(void *arg)
{
    ESP_LOGD(TAG, "Starting LVGL task");

    uint32_t task_delay_ms = LVGL_PORT_TASK_MAX_DELAY_MS;
    while (1) {
        lv_mutex_lock(&lvgl_mutex);
        task_delay_ms = lv_timer_handler(); 
        lv_mutex_unlock(&lvgl_mutex);
        if (task_delay_ms > LVGL_PORT_TASK_MAX_DELAY_MS) {
            task_delay_ms = LVGL_PORT_TASK_MAX_DELAY_MS;
        } else if (task_delay_ms < LVGL_PORT_TASK_MIN_DELAY_MS) {
            task_delay_ms = LVGL_PORT_TASK_MIN_DELAY_MS;
        }
        vTaskDelay(pdMS_TO_TICKS(task_delay_ms));
    }
}

// Mutex is created like this:
lv_mutex_init(&lvgl_mutex);                         // Make lvgl thread safe

// This is how the task is created
BaseType_t ret = xTaskCreatePinnedToCore(ProcessDisplay, "lvgl", LVGL_PORT_TASK_STACK_SIZE, NULL,
                                                                               LVGL_PORT_TASK_PRIORITY, &lvgl_task_handle, core_id);
// where stack size is 6k bytes, the piority is 2

Note that this only happens if I define:
#define LV_USE_OS   LV_OS_FREERTOS
in lv_conf.h

When I define this as LV_OS_NONE it all works.

Should I not set this LV_USE_OS to LV_OS_FREERTOS when using ESP32 on framwork Arduino which is running on top of the FreeRTOS???

liamHowatt:

Fixes #6414

`lv_thread_sync_wait` can be called from a thread other than the one that called `lv_thread_sync_init`.

@nicusorcitu

### Notes

### DIFF
--- a/src/osal/lv_freertos.c
+++ b/src/osal/lv_freertos.c
@@ -179,8 +179,8 @@ lv_result_t lv_mutex_delete(lv_mutex_t * pxMutex)
 lv_result_t lv_thread_sync_init(lv_thread_sync_t * pxCond)
 {
 #if USE_FREERTOS_TASK_NOTIFY
-    /* Store the handle of the calling task. */
-    pxCond->xTaskToNotify = xTaskGetCurrentTaskHandle();
+    pxCond->xTaskToNotify = NULL;
+    pxCond->xSignalSent = pdFALSE;
 #else
     /* If the cond is uninitialized, perform initialization. */
     prvCheckCondInit(pxCond);
@@ -194,10 +194,21 @@ lv_result_t lv_thread_sync_wait(lv_thread_sync_t * pxCond)
     lv_result_t lvRes = LV_RESULT_OK;
 
 #if USE_FREERTOS_TASK_NOTIFY
-    LV_UNUSED(pxCond);
+    TaskHandle_t current_task_handle = xTaskGetCurrentTaskHandle();
 
-    /* Wait for other task to notify this task. */
-    ulTaskNotifyTake(pdTRUE, portMAX_DELAY);
+    _enter_critical();
+    BaseType_t signal_sent = pxCond->xSignalSent;
+    pxCond->xSignalSent = pdFALSE;
+    if(signal_sent == pdFALSE) {
+        /* The signal hasn't been sent yet. Tell the sender to notify this task */
+        pxCond->xTaskToNotify = current_task_handle;
+    }
+    _exit_critical();
+
+    if(signal_sent == pdFALSE) {
+        /* Wait for other task to notify this task. */
+        ulTaskNotifyTake(pdTRUE, portMAX_DELAY);
+    }
 #else
     uint32_t ulLocalWaitingThreads;
 
@@ -261,8 +272,19 @@ lv_result_t lv_thread_sync_wait(lv_thread_sync_t * pxCond)
 lv_result_t lv_thread_sync_signal(lv_thread_sync_t * pxCond)
 {
 #if USE_FREERTOS_TASK_NOTIFY
-    /* Send a notification to the task waiting. */
-    xTaskNotifyGive(pxCond->xTaskToNotify);
+    _enter_critical();
+    TaskHandle_t task_to_notify = pxCond->xTaskToNotify;
+    pxCond->xTaskToNotify = NULL;
+    if(task_to_notify == NULL) {
+        /* No task waiting to be notified. Send this signal for later */
+        pxCond->xSignalSent = pdTRUE;
+    }
+    _exit_critical();
+
+    if(task_to_notify != NULL) {
+        /* There is a task waiting. Send a notification to it */
+        xTaskNotifyGive(task_to_notify);
+    }
 #else
     /* If the cond is uninitialized, perform initialization. */
     prvCheckCondInit(pxCond);
@@ -322,9 +344,20 @@ lv_result_t lv_thread_sync_signal_isr(lv_thread_sync_t * pxCond)
 {
     BaseType_t xHigherPriorityTaskWoken = pdFALSE;
 #if USE_FREERTOS_TASK_NOTIFY
-    /* Send a notification to the task waiting. */
-    vTaskNotifyGiveFromISR(pxCond->xTaskToNotify, &xHigherPriorityTaskWoken);
-    portYIELD_FROM_ISR(xHigherPriorityTaskWoken);
+    _enter_critical();
+    TaskHandle_t task_to_notify = pxCond->xTaskToNotify;
+    pxCond->xTaskToNotify = NULL;
+    if(task_to_notify == NULL) {
+        /* No task waiting to be notified. Send this signal for later */
+        pxCond->xSignalSent = pdTRUE;
+    }
+    _exit_critical();
+
+    if(task_to_notify != NULL) {
+        /* There is a task waiting. Send a notification to it */
+        vTaskNotifyGiveFromISR(task_to_notify, &xHigherPriorityTaskWoken);
+        portYIELD_FROM_ISR(xHigherPriorityTaskWoken);
+    }
 #else
     /* If the cond is uninitialized, perform initialization. */
     prvCheckCondInit(pxCond);
--- a/src/osal/lv_freertos.c
+++ b/src/osal/lv_freertos.c
@@ -46,11 +46,11 @@ static void prvMutexInit(lv_mutex_t * pxMutex);
 
 static void prvCheckMutexInit(lv_mutex_t * pxMutex);
 
-#if !USE_FREERTOS_TASK_NOTIFY
 static void prvCondInit(lv_thread_sync_t * pxCond);
 
 static void prvCheckCondInit(lv_thread_sync_t * pxCond);
 
+#if !USE_FREERTOS_TASK_NOTIFY
 static void prvTestAndDecrement(lv_thread_sync_t * pxCond,
                                 uint32_t ulLocalWaitingThreads);
 #endif
@@ -83,7 +83,7 @@ lv_result_t lv_thread_init(lv_thread_t * pxThread, lv_thread_prio_t xSchedPriori
                            void (*pvStartRoutine)(void *), size_t usStackSize,
                            void * xAttr)
 {
-    pxThread->xTaskArg = xAttr;
+    pxThread->pTaskArg = xAttr;
     pxThread->pvStartRoutine = pvStartRoutine;
 
     BaseType_t xTaskCreateStatus = xTaskCreate(
@@ -178,13 +178,8 @@ lv_result_t lv_mutex_delete(lv_mutex_t * pxMutex)
 
 lv_result_t lv_thread_sync_init(lv_thread_sync_t * pxCond)
 {
-#if USE_FREERTOS_TASK_NOTIFY
-    pxCond->xTaskToNotify = NULL;
-    pxCond->xSignalSent = pdFALSE;
-#else
     /* If the cond is uninitialized, perform initialization. */
     prvCheckCondInit(pxCond);
-#endif
 
     return LV_RESULT_OK;
 }
@@ -192,6 +187,9 @@ lv_result_t lv_thread_sync_init(lv_thread_sync_t * pxCond)
 lv_result_t lv_thread_sync_wait(lv_thread_sync_t * pxCond)
 {
     lv_result_t lvRes = LV_RESULT_OK;
+
+    /* If the cond is uninitialized, perform initialization. */
+    prvCheckCondInit(pxCond);
 
 #if USE_FREERTOS_TASK_NOTIFY
     TaskHandle_t current_task_handle = xTaskGetCurrentTaskHandle();
@@ -211,9 +209,6 @@ lv_result_t lv_thread_sync_wait(lv_thread_sync_t * pxCond)
     }
 #else
     uint32_t ulLocalWaitingThreads;
-
-    /* If the cond is uninitialized, perform initialization. */
-    prvCheckCondInit(pxCond);
 
     /* Acquire the mutex. */
     xSemaphoreTake(pxCond->xSyncMutex, portMAX_DELAY);
@@ -271,6 +266,9 @@ lv_result_t lv_thread_sync_wait(lv_thread_sync_t * pxCond)
 
 lv_result_t lv_thread_sync_signal(lv_thread_sync_t * pxCond)
 {
+    /* If the cond is uninitialized, perform initialization. */
+    prvCheckCondInit(pxCond);
+
 #if USE_FREERTOS_TASK_NOTIFY
     _enter_critical();
     TaskHandle_t task_to_notify = pxCond->xTaskToNotify;
@@ -286,9 +284,6 @@ lv_result_t lv_thread_sync_signal(lv_thread_sync_t * pxCond)
         xTaskNotifyGive(task_to_notify);
     }
 #else
-    /* If the cond is uninitialized, perform initialization. */
-    prvCheckCondInit(pxCond);
-
     /* Acquire the mutex. */
     xSemaphoreTake(pxCond->xSyncMutex, portMAX_DELAY);
 
@@ -326,16 +321,14 @@ lv_result_t lv_thread_sync_signal(lv_thread_sync_t * pxCond)
 
 lv_result_t lv_thread_sync_delete(lv_thread_sync_t * pxCond)
 {
-#if USE_FREERTOS_TASK_NOTIFY
-    LV_UNUSED(pxCond);
-#else
+#if !USE_FREERTOS_TASK_NOTIFY
     /* Cleanup all resources used by the cond. */
     vSemaphoreDelete(pxCond->xCondWaitSemaphore);
     vSemaphoreDelete(pxCond->xSyncMutex);
     pxCond->ulWaitingThreads = 0;
     pxCond->xSyncSignal = pdFALSE;
-    pxCond->xIsInitialized = pdFALSE;
 #endif
+    pxCond->xIsInitialized = pdFALSE;
 
     return LV_RESULT_OK;
 }
@@ -343,6 +336,10 @@ lv_result_t lv_thread_sync_delete(lv_thread_sync_t * pxCond)
 lv_result_t lv_thread_sync_signal_isr(lv_thread_sync_t * pxCond)
 {
     BaseType_t xHigherPriorityTaskWoken = pdFALSE;
+
+    /* If the cond is uninitialized, perform initialization. */
+    prvCheckCondInit(pxCond);
+
 #if USE_FREERTOS_TASK_NOTIFY
     _enter_critical();
     TaskHandle_t task_to_notify = pxCond->xTaskToNotify;
@@ -359,9 +356,6 @@ lv_result_t lv_thread_sync_signal_isr(lv_thread_sync_t * pxCond)
         portYIELD_FROM_ISR(xHigherPriorityTaskWoken);
     }
 #else
-    /* If the cond is uninitialized, perform initialization. */
-    prvCheckCondInit(pxCond);
-
     /* Enter critical section to prevent preemption. */
     _enter_critical();
 
@@ -390,7 +384,7 @@ static void prvRunThread(void * pxArg)
     lv_thread_t * pxThread = (lv_thread_t *)pxArg;
 
     /* Run the thread routine. */
-    pxThread->pvStartRoutine((void *)pxThread->xTaskArg);
+    pxThread->pvStartRoutine((void *)pxThread->pTaskArg);
 
     vTaskDelete(NULL);
 }
@@ -429,9 +423,12 @@ static void prvCheckMutexInit(lv_mutex_t * pxMutex)
     }
 }
 
-#if !USE_FREERTOS_TASK_NOTIFY
 static void prvCondInit(lv_thread_sync_t * pxCond)
 {
+#if USE_FREERTOS_TASK_NOTIFY
+    pxCond->xTaskToNotify = NULL;
+    pxCond->xSignalSent = pdFALSE;
+#else
     pxCond->xCondWaitSemaphore = xSemaphoreCreateCounting(ulMAX_COUNT, 0U);
 
     /* Ensure that the FreeRTOS semaphore was successfully created. */
@@ -453,13 +450,13 @@ static void prvCondInit(lv_thread_sync_t * pxCond)
     /* Condition variable successfully created. */
     pxCond->ulWaitingThreads = 0;
     pxCond->xSyncSignal = pdFALSE;
+#endif
+
     pxCond->xIsInitialized = pdTRUE;
 }
 
 static void prvCheckCondInit(lv_thread_sync_t * pxCond)
 {
-    BaseType_t xSemCreateStatus = pdTRUE;
-
     /* Check if the condition variable needs to be initialized. */
     if(pxCond->xIsInitialized == pdFALSE) {
         /* Cond initialization must be in a critical section to prevent two
@@ -478,6 +475,7 @@ static void prvCheckCondInit(lv_thread_sync_t * pxCond)
     }
 }
 
+#if !USE_FREERTOS_TASK_NOTIFY
 static void prvTestAndDecrement(lv_thread_sync_t * pxCond,
                                 uint32_t ulLocalWaitingThreads)
 {
--- a/src/osal/lv_freertos.c
+++ b/src/osal/lv_freertos.c
@@ -195,18 +195,20 @@ lv_result_t lv_thread_sync_wait(lv_thread_sync_t * pxCond)
     TaskHandle_t current_task_handle = xTaskGetCurrentTaskHandle();
 
     _enter_critical();
-    BaseType_t signal_sent = pxCond->xSignalSent;
-    pxCond->xSignalSent = pdFALSE;
+    BaseType_t signal_sent = pxCond->xSyncSignal;
+    pxCond->xSyncSignal = pdFALSE;
     if(signal_sent == pdFALSE) {
         /* The signal hasn't been sent yet. Tell the sender to notify this task */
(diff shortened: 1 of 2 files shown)


kisvegabor:
cc @espzav @Jvsjvs

kisvegabor (dismissed):
Thank you Liam! 

I wonder if other operating systems don't have the same issue. 

Anyway, I've looked into LVGL and couldn't see an obvious way to restructure it to make it work even without this fix. So I suggest merging this fix now and see if the issue shows up for others too.

Jvsjvs:
Thanks for picking it up. 

If there is a performance hit, then I believe it is better to simply enforce the users to use the same thread with proper documentation. 

This restriction is easy to address by the users, maybe you can propose to the users to use a similar setup as I did and documented in the original issue.

If it works now without performance hit then this is of course perfect!!

src/osal/lv_freertos.c
```diff
@@ -261,8 +272,19 @@ lv_result_t lv_thread_sync_wait(lv_thread_sync_t * pxCond)
...
+    _enter_critical();
+    TaskHandle_t task_to_notify = pxCond->xTaskToNotify;
+    pxCond->xTaskToNotify = NULL;
+    if(task_to_notify == NULL) {
+        /* No task waiting to be notified. Send this signal for later */
+        pxCond->xSignalSent = pdTRUE;
+    }
+    _exit_critical();
+
+    if(task_to_notify != NULL) {
```
nicusorcitu, line 284: Please use the freertos naming convention: rename `task_to_notify` to 'xTaskToNotify`:

https://www.freertos.org/Documentation/02-Kernel/06-Coding-guidelines/02-FreeRTOS-Coding-Standard-and-Style-Guide#naming-conventions

src/osal/lv_freertos.c
```diff
@@ -194,10 +194,21 @@ lv_result_t lv_thread_sync_wait(lv_thread_sync_t * pxCond)
     lv_result_t lvRes = LV_RESULT_OK;
 
 #if USE_FREERTOS_TASK_NOTIFY
-    LV_UNUSED(pxCond);
+    TaskHandle_t current_task_handle = xTaskGetCurrentTaskHandle();
+
+    _enter_critical();
+    BaseType_t signal_sent = pxCond->xSignalSent;
```
nicusorcitu, line 200: Please use the freertos naming convention: rename `signal_sent` to 'xSignalSent`

nicusorcitu:
@liamHowatt thanks for this PR.

nicusorcitu (approved):

src/osal/lv_freertos.c
```diff
@@ -194,10 +194,21 @@ lv_result_t lv_thread_sync_wait(lv_thread_sync_t * pxCond)
     lv_result_t lvRes = LV_RESULT_OK;
 
 #if USE_FREERTOS_TASK_NOTIFY
-    LV_UNUSED(pxCond);
+    TaskHandle_t current_task_handle = xTaskGetCurrentTaskHandle();
```
nicusorcitu, line 197: `xCurrentTaskHandle`

src/osal/lv_freertos.c
```diff
@@ -261,8 +272,19 @@ lv_result_t lv_thread_sync_wait(lv_thread_sync_t * pxCond)
 lv_result_t lv_thread_sync_signal(lv_thread_sync_t * pxCond)
 {
 #if USE_FREERTOS_TASK_NOTIFY
-    /* Send a notification to the task waiting. */
-    xTaskNotifyGive(pxCond->xTaskToNotify);
+    _enter_critical();
+    TaskHandle_t task_to_notify = pxCond->xTaskToNotify;
```
nicusorcitu, line 276: Use the new member `BaseType_t    xIsInitialized; ` to verify if the task to notify is valid.

src/osal/lv_freertos.h
```diff
@@ -63,6 +63,7 @@ typedef struct {
 typedef struct {
 #if USE_FREERTOS_TASK_NOTIFY
     TaskHandle_t xTaskToNotify;
+    BaseType_t xSignalSent;
```
nicusorcitu: Please use the freertos naming convention all over in this file:
https://www.freertos.org/Documentation/02-Kernel/06-Coding-guidelines/02-FreeRTOS-Coding-Standard-and-Style-Guide#naming-conventions

At a better look, I think you can use a member like 
`BaseType_t    xIsInitialized; `
similar as it is done on semaphore implementation.

E.G. no signal can be done until the initialization is done.

src/osal/lv_freertos.c
```diff
@@ -179,8 +179,8 @@ lv_result_t lv_mutex_delete(lv_mutex_t * pxMutex)
 lv_result_t lv_thread_sync_init(lv_thread_sync_t * pxCond)
 {
 #if USE_FREERTOS_TASK_NOTIFY
-    /* Store the handle of the calling task. */
-    pxCond->xTaskToNotify = xTaskGetCurrentTaskHandle();
+    pxCond->xTaskToNotify = NULL;
```
nicusorcitu: Please don't remove the scope of the init API.

Add a local function here like:
```
    /* If the cond is uninitialized, perform initialization. */
    prvCheckCondInit(pxCond);
```
and let the initialization happen if it get called explicitly.

Then make use of the same static function `prvCheckCondInit(pxCond);` in the `lv_thread_sync_wait()`.

nicusorcitu (changes requested):

liamHowatt:
Thanks for the review!

src/osal/lv_freertos.h
```diff
@@ -51,7 +51,7 @@ extern "C" {
 
 typedef struct {
     void (*pvStartRoutine)(void *);       /**< Application thread function. */
-    void * xTaskArg;                      /**< Arguments for application thread function. */
+    void * pTaskArg;                      /**< Arguments for application thread function. */
```
nicusorcitu, line 54: `pvTaskArg`

src/osal/lv_freertos.h
```diff
@@ -61,11 +61,12 @@ typedef struct {
 } lv_mutex_t;
 
 typedef struct {
+    BaseType_t
+    xIsInitialized;                       /**< Set to pdTRUE if this condition variable is initialized, pdFALSE otherwise. */
 #if USE_FREERTOS_TASK_NOTIFY
     TaskHandle_t xTaskToNotify;
+    BaseType_t xSignalSent;
```
nicusorcitu: No need for `xSignalSent`, so it can be removed, correct? We make use of the `pxCond->xSyncSignal` now.

src/osal/lv_freertos.c
```diff
@@ -396,9 +423,12 @@ static void prvCheckMutexInit(lv_mutex_t * pxMutex)
     }
 }
 
-#if !USE_FREERTOS_TASK_NOTIFY
 static void prvCondInit(lv_thread_sync_t * pxCond)
 {
+#if USE_FREERTOS_TASK_NOTIFY
+    pxCond->xTaskToNotify = NULL;
```
nicusorcitu, line 438: Implement this, similar as it is done for semaphores.
I see it like:
```
#if USE_FREERTOS_TASK_NOTIFY
    pxCond->xTaskToNotify = xTaskGetCurrentTaskHandle();
    pxCond->xIsInitialized = pdTRUE;
    pxCond->xSyncSignal = pdFALSE;
#else
```
You might have the last two common for semaphores and sygnal thread.

src/osal/lv_freertos.c
```diff
@@ -396,9 +423,12 @@ static void prvCheckMutexInit(lv_mutex_t * pxMutex)
...
+    pxCond->xSignalSent = pdFALSE;
```
nicusorcitu, line 430: Use `pxCond->xSyncSignal = pdFALSE;` instead

src/osal/lv_freertos.c
```diff
@@ -260,13 +266,24 @@ lv_result_t lv_thread_sync_wait(lv_thread_sync_t * pxCond)
...
 {
-#if USE_FREERTOS_TASK_NOTIFY
-    /* Send a notification to the task waiting. */
-    xTaskNotifyGive(pxCond->xTaskToNotify);
-#else
     /* If the cond is uninitialized, perform initialization. */
     prvCheckCondInit(pxCond);
 
+#if USE_FREERTOS_TASK_NOTIFY
+    _enter_critical();
```
nicusorcitu, line 275: Replace all new code with:
```
#if USE_FREERTOS_TASK_NOTIFY
    xTaskNotifyGive(pxCond->xTaskToNotify);
    pxCond->xSyncSignal = pdTRUE;
#else
```

src/osal/lv_freertos.c
```diff
@@ -193,17 +188,28 @@ lv_result_t lv_thread_sync_wait(lv_thread_sync_t * pxCond)
 {
     lv_result_t lvRes = LV_RESULT_OK;
 
+    /* If the cond is uninitialized, perform initialization. */
+    prvCheckCondInit(pxCond);
```
nicusorcitu, line 192: If wait gets called prior of any init, this call will do the initialization.

src/osal/lv_freertos.c
```diff
@@ -193,17 +188,28 @@ lv_result_t lv_thread_sync_wait(lv_thread_sync_t * pxCond)
...
+
 #if USE_FREERTOS_TASK_NOTIFY
-    LV_UNUSED(pxCond);
+    TaskHandle_t current_task_handle = xTaskGetCurrentTaskHandle();
```
nicusorcitu, line 195: Replace all new code with:
```
#if USE_FREERTOS_TASK_NOTIFY
    while(!pxCond->xSyncSignal) {
        ulTaskNotifyTake(pdTRUE, portMAX_DELAY);
    }
    pxCond->xSyncSignal = pdFALSE;
#else
````

File: src/osal/lv_freertos.c [hunk 14]
```diff
@@ -304,31 +321,41 @@ lv_result_t lv_thread_sync_signal(lv_thread_sync_t * pxCond)
...
-    /* Send a notification to the task waiting. */
-    vTaskNotifyGiveFromISR(pxCond->xTaskToNotify, &xHigherPriorityTaskWoken);
-    portYIELD_FROM_ISR(xHigherPriorityTaskWoken);
-#else
+
     /* If the cond is uninitialized, perform initialization. */
     prvCheckCondInit(pxCond);
 
+#if USE_FREERTOS_TASK_NOTIFY
+    _enter_critical();
```
nicusorcitu, line 348: Similar as done for sync_signal.

nicusorcitu (changes requested):

src/osal/lv_freertos.h
liamHowatt: In the coding standard naming conventions you linked, the "v" seems to only apply for function names that return `void` (nothing), not void pointer variables.

src/osal/lv_freertos.h
liamHowatt: Not currently, but I can make that variable common between both implementations :+1:

src/osal/lv_freertos.c
liamHowatt: Correct. I applied your feedback after your last review to make it work that way. Is there something you want me to change?

src/osal/lv_freertos.c
liamHowatt: I'm not completely clear on this. Is this a different solution?

nicusorcitu: I mean to replace all the code under `#if USE_FREERTOS_TASK_NOTIFY` with the above suggestion and see if it works to you.

src/osal/lv_freertos.c
nicusorcitu: I mean to replace all the code under `#if USE_FREERTOS_TASK_NOTIFY` with the above suggestion and see if it works to you.

src/osal/lv_freertos.c
nicusorcitu: I mean to replace all the code under `#if USE_FREERTOS_TASK_NOTIFY` with the above suggestion and see if it works to you.

liamHowatt:
> I mean to replace all the code under `#if USE_FREERTOS_TASK_NOTIFY` with the above suggestion and see if it works to you.

Gotcha, I will have time to try tomorrow.

liamHowatt:
Yes, it is currently working.

kisvegabor (approved):

kisvegabor:
Thanks, merging then to make sure that it's fixed in v9.2.

src/osal/lv_freertos.c
liamHowatt: > Replace all new code with:
> 
> ```
> #if USE_FREERTOS_TASK_NOTIFY
>     xTaskNotifyGive(pxCond->xTaskToNotify);
>     pxCond->xSyncSignal = pdTRUE;
> #else
> ```

`pxCond->xTaskToNotify` may be `NULL` which is why there is the current logic to determine which task to notify, if any.

liamHowatt (commented):
I think you have a simpler way to do it than what I've done. Would you mind opening a PR? :slightly_smiling_face:

nicusorcitu:
`prvCheckCondInit(pxCond);` should always make sure that `pxCond->xTaskToNotify` is not null, (e.g. the conditional wait is always initialized).

src/osal/lv_freertos.c
```diff
@@ -304,31 +325,43 @@ lv_result_t lv_thread_sync_signal(lv_thread_sync_t * pxCond)
...
```
nicusorcitu, line 348: `_enter_critical()` does not work for ISR, it will hang.
You would have to add new defines:
```
#define _enter_critical_isr() taskENTER_CRITICAL_FROM_ISR();
```

nicusorcitu:
> I think you have a simpler way to do it than what I've done. Would you mind opening a PR? 🙂

We have a strict way of pushing commits to upstream, need to pass some additional checks.

For this particular issue we might send a commit with the proposal, will open a PR soon.
@cosmindanielradu19

src/osal/lv_freertos.c
liamHowatt: Very good catch! :open_mouth: In all of the environments I've seen so far, the same critical section API is used inside and outside of ISRs. I tested on ESP32-S3 and it worked for me. Maybe only some ports of FreeRTOS implement the ISR version with important extra steps.

nicusorcitu:
I have done a lot of tests and investigated the new code which does not seem to work anymore on FreeRTOS.
The whole story start from this sentence:
> lv_thread_sync_wait can be called from a thread other than the one that called lv_thread_sync_init.
>
But why? Task notification is different than semaphore implementation.
On semaphores there is a single sync that can be used in both directions.
On task notification you can use the sync only in a single direction. If you want to be both ways you will need two synchronizations, hence two init calls.

I would revert the changes due to this commit and revise it from the application level. It had more sense as it was before.

liamHowatt:
Even after using the `taskENTER_CRITICAL_FROM_ISR` fix? Do you know what usage pattern causes the failure you're seeing? I would like to understand why these changes aren't working.

Please see the issue this PR is linked to https://github.com/lvgl/lvgl/issues/6414. It's non-obvious to users that the task that calls `lv_timer_handler` must be the same task that called `lv_init`. Vendor port code also made the mistake.

It seems like the task notify vs semaphore choice is not very configurable. It's a define in a header file, not part of the config. I think the task notify implementation has the potential to deprecate the semaphore implementation at some point, but only if it doesn't have too many tradeoffs. If both implementations stay, I think it should be easier for a user to choose which implementation they want.

nicusorcitu:
> Even after using the taskENTER_CRITICAL_FROM_ISR fix?
>
Yes, that would fix only the hang from ISR.
Let's say Task1 is the LVGL and Task2 is the GPU unit.
Task1 have `_draw_info.sync` to know when it has to dispatch again.
Task2 have `u.sync` to know when to process a task.

As I mentioned in the previous comment, the task notification means that Task1 waits on `_draw_info.sync` and Task2 signals the Task1 Handle. So it works only in one direction.
Task2 waits on `u.sync` until a new task is passed to the unit and signaled from Task1.

1. In this PR you changed the behavior of what sync means. E.G. if Task2 signals before Task1 waits, then Task1 wait will just pass waiting.
And is fine, this case happens case Task2 have higher priority than Task1.

2. But, in dispatcher in Task1, while there is a need for Task2 to `wait_for_completion` it will signal Task2 and will return to dispatcher.
image
But during dispatching, there will be no wait at all because of the previous signal:
image
Hence, Task1 will get blocked on the steps: signal, wait, signal, wait, signal, wait.... on its own thread handle. So Task1 will never wait (because will just pass waiting as previously signaled), so Task1 will never sleep so never returned to scheduler to give the possibility to Task2 to wake up and complete his job.

I really suggest to revert back the change and I will add the config `LV_USE_TASK_NITOFICATION` under `LV_USE_FREETOS` to let it configurable by user.
If used want, he can switch back to the semaphores implementation.
If user want to stay with task notification then he need to be aware that task notification is unidirectional. It is what the name says.

liamHowatt:
I'm looking at the code you're talking about. I think `lv_draw_dispatch_wait_for_request` will block the next time it's called. It should block until the next `lv_draw_dispatch_request`.

I don't think `lv_draw_dispatch_wait_for_request` will return immediately every time while being called in a loop. If the signal was already sent, yes, it will return immediately. But the next time it will block at `ulTaskNotifyTake` until another `lv_draw_dispatch_request` is called from Task2 (the GPU).

nicusorcitu:
Exactly that is the case, the sync goes first then wait will return immediately. So It will create a loop in dispatcher.

All the user need to do is to keep in mind that the sync point means that the thread handle `pxCond->xTaskToNotify` is the task that waits. Any other task can signal it to unblock. This already is functioning on LVGL, there are 2 sync points for signaling in both direction:
> Let's say Task1 is the LVGL and Task2 is the GPU unit.
> Task1 have `_draw_info.sync` to know when it has to dispatch again.
> Task2 have `u.sync` to know when to process a task.

Please also see the example in FreeRTOS: https://www.freertos.org/Documentation/02-Kernel/04-API-references/05-Direct-to-task-notifications/01-xTaskNotifyGive

I will prepare a patch and will push it to github on PR https://github.com/lvgl/lvgl/pull/6793

liamHowatt:
> Exactly that is the case, the sync goes first then wait will return immediately. So It will create a loop in dispatcher.

It would block at the next call, though. No? `xSyncSignal` is reset to `pdFALSE` so at most it can only return immediately once and then block the next time. How does it become and infinite loop?

```c
lv_result_t lv_thread_sync_wait(lv_thread_sync_t * pxCond)
{
    lv_result_t lvRes = LV_RESULT_OK;

    /* If the cond is uninitialized, perform initialization. */
    prvCheckCondInit(pxCond);

#if USE_FREERTOS_TASK_NOTIFY
    TaskHandle_t current_task_handle = xTaskGetCurrentTaskHandle();

    _enter_critical();
    BaseType_t signal_sent = pxCond->xSyncSignal;
    pxCond->xSyncSignal = pdFALSE;
    if(signal_sent == pdFALSE) {
        /* The signal hasn't been sent yet. Tell the sender to notify this task */
        pxCond->xTaskToNotify = current_task_handle;
    }
    /* If we have a signal from the other task, we should not ask to be notified */
    _exit_critical();

    if(signal_sent == pdFALSE) {
        /* Wait for other task to notify this task. */
        ulTaskNotifyTake(pdTRUE, portMAX_DELAY);
    }
    /* If the signal was received, no wait needs to be done */
#else
    /* ... */
#endif

    return lvRes;
}
```

nicusorcitu:
I will loop because the signal will always come first in this order: signal, wait, signal, wait, signal, wait....

liamHowatt:
So it's not spinning in an infinite loop in one task. It's switching between tasks. What's wrong with signal, wait, signal, wait, signal, wait?

liamHowatt:
It looks like draw drivers are supposed to provide a `wait_for_finish_cb` which is called in `lv_draw_wait_for_finish` to trigger the context switch if they have their own threads. Looks like vglite is the only one that has a `wait_for_finish_cb`. draw_sw doesn't have a `wait_for_finish_cb`. I think it's a bit strange that we are relying on the signal to always context switch even if the thread is signalling itself.

How did the old implementation work? How does a high priority task trigger a context switch to a lower priority task by sending a signal to itself?

src/osal/lv_freertos.c
```diff
@@ -193,17 +188,30 @@ lv_result_t lv_thread_sync_wait(lv_thread_sync_t * pxCond)
...
+    BaseType_t signal_sent = pxCond->xSyncSignal;
+    pxCond->xSyncSignal = pdFALSE;
+    if(signal_sent == pdFALSE) {
+        /* The signal hasn't been sent yet. Tell the sender to notify this task */
+        pxCond->xTaskToNotify = current_task_handle;
+    }
+    /* If we have a signal from the other task, we should not ask to be notified */
+    _exit_critical();
+
```
nicusorcitu, line 206: After exiting from critical section the `pxCond->xSyncSignal` might get changed, so this is not thread safe.

src/osal/lv_freertos.c
```diff
@@ -260,13 +268,26 @@ lv_result_t lv_thread_sync_wait(lv_thread_sync_t * pxCond)
...
+    _enter_critical();
+    TaskHandle_t task_to_notify = pxCond->xTaskToNotify;
+    pxCond->xTaskToNotify = NULL;
+    if(task_to_notify == NULL) {
+        /* No task waiting to be notified. Send this signal for later */
+        pxCond->xSyncSignal = pdTRUE;
+    }
+    /* If a task is already waiting, there is no need to set the sync signal */
+    _exit_critical();
+
```
nicusorcitu, line 284: Same here, you exit from critical section but still rely on the `task_to_notify`. Not thread safe.

nicusorcitu:
@cosmindanielradu19 @cristian-stoica 
This PR generated another task synchronization issue with both VGLite and PXP screen rotation enabled.
I still believe we should get back to the original implementation where a single task can wait and a single task can signal.
>>I really suggest to revert back the change and I will add the config LV_USE_TASK_NOTIFICATION under LV_USE_FREETOS to let it configurable by user.
If used want, he can switch back to the semaphores implementation.
If user want to stay with task notification then he need to be aware that task notification is unidirectional. It is what the name says.
>>

kisvegabor:
@nicusorcitu Please open a PR to see the proposed changes. @liamHowatt, you were more involved into this issue, please take a look.

nicusorcitu:
I do not have a fix yet, neither time to look at it right now. But I will try a patch till 20Dec. I will let you know both about it.

liamHowatt:
I would like to understand what's wrong with the fix I did. You raised some concerns in the PR during review. Can you reiterate the ones that you still believe are potential causes of instability? I truly believe the parts you flagged as thread-unsafe are thread-safe. Can we begin reviewing those changes, starting with a specific bug?

cristian-stoica:
@liamHowatt I'm looking at a situation when running lv_demo_benchmark with both VGLITE and PXP enabled (SW draw unit = 0). Tests run as expected until rotated_argb_image_cb when the screen turns black and app stops. Debugging shows the code is still running in a dispatch-execute loop without any progress at all.  I'm not sure what is actually going on. What I've also tried is disabling TASK_NOTIFY and the application runs. I can reproduce the issue only by running only that test under MCUX from flash (LVGL 9.2.1).
It's strange because I've had reproducibility issues when running the project built with different compilers IAR, MCUX and running under debugger (no breakpoints) or without it.

nicusorcitu:
The issue we see works just fine with semaphore implementation. It hangs only if we use `LV_USE_FREERTOS_TASK_NOTIFY`.

>> I truly believe the parts you flagged as thread-unsafe are thread-safe. Can we begin reviewing those changes, starting with a specific bug?
>>

```
lv_result_t lv_thread_sync_wait(lv_thread_sync_t * pxCond)
{
    lv_result_t lvRes = LV_RESULT_OK;

    /* If the cond is uninitialized, perform initialization. */
    prvCheckCondInit(pxCond);

#if LV_USE_FREERTOS_TASK_NOTIFY
    TaskHandle_t xCurrentTaskHandle = xTaskGetCurrentTaskHandle();

    _enter_critical();
    BaseType_t xSyncSygnal = pxCond->xSyncSignal; ---> This might be TRUE
    pxCond->xSyncSignal = pdFALSE; ---> Becoming FALSE
    if(xSyncSygnal == pdFALSE) {
        /* The signal hasn't been sent yet. Tell the sender to notify this task */
        pxCond->xTaskToNotify = xCurrentTaskHandle;
    }
    /* If we have a signal from the other task, we should not ask to be notified */
    _exit_critical();

    if(xSyncSygnal == pdFALSE) { ---> Another Thread might just set pxCond->xSyncSignal = pdTRUE so this check might get wrong
        /* Wait for other task to notify this task. */
        ulTaskNotifyTake(pdTRUE, portMAX_DELAY);
    }
    /* If the signal was received, no wait needs to be done */
#else
```

I will try to do some debugs but can't have right now, I will let you know when I have something.

PR #6714, opened 2024-08-21
fix(i1): fix compiler and runtime issues with I1 rendering

### LINKED ISSUE #6701: Support for LV_COLOR_FORMAT_I1 is incomplete
### LVGL version

Commit 790aa5a529d1ac37afbb0a18c33a9ecf4d64bff8

### What happened?

Commit 790aa5a529d1ac37afbb0a18c33a9ecf4d64bff8 from 19-Jul-2024 added support for 1-bit color depth. I tried to use it for my 296x128 EPaper display. However, I encountered several problems.

1. Defining `LV_COLOR_DEPTH 1` in lv_conf.h resulted in compile errors. This was fixed (I think) by modifying lv_color.h:
grafik
2. I provided an update buffer of size 296*128/8=4736 bytes to `lv_display_set_buffers()`, which is enough for a full update of the screen at 1 bit/pixel. However each call of `flush_cb()` would only update an area of 296x4 pixels.
3. The first 8 bytes of `pxmap` in `flush_cb` were unassigned and must be skipped (bytes AA below). The correct pixel data begins afterwards.
grafik

The problem still exists in recent commit c52aa1012898bca217e35f4d5e632e6f73667d25.

### How to reproduce?

1. In `lv_conf.h`, `#define LV_COLOR_DEPTH 1`.
2. In `lv_color.h`, apply the modification described above.
3. Build and run with `main.c` below.
    - The terminal output shows that in each call of `flush_cb()`, only 296x4 pixels are updated.
    - View `screenout.bin` in a hex editor. On each call of `flush_cb()`, the first 8 bytes are AA, which is the initial buffer value. Therefore, these bytes are never written by LVGL.
4. Uncomment the line `pxMap += 8;` in main.c and repeat. Now the pixel data in `screenout.bin` is correct.

```
#include "lvgl/lvgl.h"
#include <time.h>
#include <stdint.h>
#include <string.h>
#include <stdio.h>

#if LV_COLOR_DEPTH!=1
    #error Set LV_COLOR_DEPTH=1 in lv_conf.h
#endif

// --------------------------------------
// Local const and variables
// --------------------------------------
static const uint16_t DISP_WIDTH  = 296u;
static const uint16_t DISP_HEIGHT = 128u;
static const uint32_t BUFFER_SIZE_BYTES = DISP_WIDTH*DISP_HEIGHT / 8u;

// --------------------------------------
// Local function declarations
// --------------------------------------
static void hal_init(void);
static void hello_world(void);
static void flush_cb(lv_display_t* display, const lv_area_t* area, uint8_t* px_map);

// --------------------------------------
// Function definitions
// --------------------------------------
static void hal_init(void)
{
    static uint8_t m_buffer[296*128/8];
    (void)memset(m_buffer, 0xAA, sizeof(m_buffer));

    lv_group_set_default(lv_group_create());
    lv_display_t* const display = lv_display_create(DISP_WIDTH, DISP_HEIGHT);
    lv_display_set_buffers(display, m_buffer, NULL, sizeof(m_buffer), LV_DISP_RENDER_MODE_PARTIAL);
    lv_display_set_flush_cb(display, &flush_cb);
    lv_display_set_default(display);
}

static void hello_world(void)
{
    lv_obj_t* const label = lv_label_create(lv_screen_active());
    lv_label_set_text(label, "Hello world!");
    lv_obj_set_style_text_color(lv_screen_active(), lv_color_hex(0x000000), LV_PART_MAIN);
    lv_obj_align(label, LV_ALIGN_CENTER, 0, 0);
}

static void flush_cb(lv_display_t* display, const lv_area_t* area, uint8_t* px_map)
{
    static uint8_t m_fileBuffer[296*128/8];

    printf("Area: %d %3d %3d %3d\n", area->x1, area->x2, area->y1, area->y2);

    // Skip 8 bytes that are unassigned
    // px_map += 8;
    for (int32_t y=area->y1; y<=area->y2; y++)
    {
        const uint32_t offset = ((y*DISP_WIDTH) + area->x1) / 8u;
        uint8_t* pBuffer = &m_fileBuffer[offset];
        for (int32_t x=area->x1; x<=area->x2; x+=8u)
        {
            *pBuffer = *px_map;
            ++pBuffer;
            ++px_map;
        }
    }

    if (lv_display_flush_is_last(display))
    {
        FILE* const fp = fopen("screenout.bin", "wb");
        if (fp!=NULL)
        {
            (void)fwrite(m_fileBuffer, 1u, BUFFER_SIZE_BYTES, fp);
            (void)fclose(fp);
        }
    }
    lv_display_flush_ready(display);
}

int main(void)
{
    lv_init();
    hal_init();
    hello_world();

    while (1)
    {
        const struct timespec time_5ms = {0u, 5000000u};
        lv_timer_handler();
        (void)nanosleep(&time_5ms, NULL);
    }

    return 0;
}
```

kisvegabor:

Fixes #6701

### Notes


kisvegabor:
cc @faxe1008

src/core/lv_refr.c
```diff
@@ -291,6 +291,12 @@ void lv_inv_area(lv_display_t * disp, const lv_area_t * area_p)
     suc = lv_area_intersect(&com_area, area_p, &scr_area);
     if(suc == false)  return; /*Out of the screen*/
 
+    if(disp->color_format == LV_COLOR_FORMAT_I1) {
+        com_area.x1 &= ~0x7; /*Round down*/
+        com_area.x2 += 7;    /*Round up*/
+        com_area.x2 &= ~0x7;
+    }
```
faxe1008, line 299: True, the rounder callback was missing.

src/misc/lv_color.h
```diff
@@ -171,7 +171,10 @@ typedef enum {
     LV_COLOR_FORMAT_YUV_END           = LV_COLOR_FORMAT_UYVY,
 
     /*Color formats in which LVGL can render*/
-#if LV_COLOR_DEPTH == 8
+#if LV_COLOR_DEPTH == 1
+    LV_COLOR_FORMAT_NATIVE            = LV_COLOR_FORMAT_I1,
+    LV_COLOR_FORMAT_NATIVE_WITH_ALPHA = LV_COLOR_FORMAT_I1,
+#elif LV_COLOR_DEPTH == 8
```
faxe1008, line 177: is an else case with an build_time assert maybe sensible?

faxe1008 (approved):

XuNeo (dismissed):

liamHowatt (commented):
LGTM after a response to the feedback.

Would it be a good idea to add one sentence somewhere in the docs to mention that the palette is at the start of `px_map` in the flush callback? It was a concern in the issue.

src/core/lv_refr.c
amanita-citrina: Is x2 "the first unused position" (comparison `foo<x2`) or "the last used position" (comparison `foo<=x2`)? In the latter case, lines 296 and 297 must be exchanged for correct rounding.

kisvegabor:
- Docs is added
- Rounding is fixed
- Assert is added in `lv_color.h`

Please review again.

liamHowatt (dismissed):
Thanks!

> Rounding is fixed

Was it? I'm not sure what the concern was.

docs/porting/display.rst
```diff
@@ -205,6 +205,29 @@ to
...
+- End on an ``Nx8 - 1`` coordinate
+
+When setting up the buffers for rendering (:cpp:func:`lv_display_set_buffers`), make the buffer 8 bytes larger.
+This is necessary because LVGL skips 8 bytes in the buffer, as these are assumed to be used as a palette.
+
+For example, a 128 x 64 monochrome buffer has 8192 pixels, which requires 8192 / 8 = 1024 bytes. To accommodate the palette, the buffer size should be 1024 + 8 = 1032 bytes.
+
+To skip the palette, include the following line in your ``flush_cb`` function: ``px_map += 8``.
+
+As usual, monochrome displays support partial, full and dirent rendering modes as well.
```
liamHowatt:
```suggestion
As usual, monochrome displays support partial, full and direct rendering modes as well.
```

liamHowatt (dismissed):

src/core/lv_refr.c
kisvegabor: Should be fixed now

kisvegabor:
>> Rounding is fixed
> Was it? I'm not sure what the concern was.

Huh, it's great that you have noticed it. Probably I haven't saved the file before commiting. Should be fixed now.

liamHowatt (dismissed):

amanita-citrina:
Unfortunately, the latest commit 22b15da0ba51a5b950ac77c06b5244107cb8e7cd does not work at all.

`lv_draw_buf.c`, function `lv_draw_buf_reshape()`, line 338ff:
```
    if(size > draw_buf->data_size) {
        LV_LOG_TRACE("Draw buf too small for new shape");
        return NULL;
    }
```
The comparison is always true because `size` is `draw_buf->data_size + 8` (the color palette size). Subseqeuently, the code fails the assertion in line 607 of `layer_reshape_draw_buf()` and remains in the endless loop therein.

kisvegabor:
I don't have a monochrome display at hand so can play only with the simulator, but it's not freezing for me:
```c
#define BUF_SIZE  128 * 96 / 8 + 8
  static uint8_t buf[BUF_SIZE];

  lv_display_t * disp = lv_display_create(128, 96);
  lv_display_set_color_format(disp, LV_COLOR_FORMAT_I1);
  lv_display_set_buffers(disp, buf, NULL, BUF_SIZE, LV_COLOR_FORMAT_I1);
```

amanita-citrina:
I used buffer size 296*128/8 and /16 (without +8). It should work for arbitrary sizes; IIRC, the instructions suggest WxH/10.

24.08.2024 11:21:07 Gabor Kiss-Vamosi ***@***.***>:

kisvegabor:
In `LV_DISPLAY_RENDER_MODE_FULL` and `LV_DISPLAY_RENDER_DIRECT` the buffers should be large enough for the whole screen. In this case it's `w x h / 8 + 8`. In `LV_DISPLAY_RENDER_PARTIAL` any size (larger than one line) should work.

docs/porting/display.rst
```diff
@@ -205,6 +205,29 @@ to
...
+- End on an ``Nx8 - 1`` coordinate
+
+When setting up the buffers for rendering (:cpp:func:`lv_display_set_buffers`), make the buffer 8 bytes larger.
+This is necessary because LVGL skips 8 bytes in the buffer, as these are assumed to be used as a palette.
+
+For example, a 128 x 64 monochrome buffer has 8192 pixels, which requires 8192 / 8 = 1024 bytes. To accommodate the palette, the buffer size should be 1024 + 8 = 1032 bytes.
+
+To skip the palette, include the following line in your ``flush_cb`` function: ``px_map += 8``.
+
+As usual, monochrome displays support partial, full and direct rendering modes as well.
```
kisvegabor:
```suggestion
As usual, monochrome displays support partial, full and direct rendering modes as well. In full and direct mode the buffer size should be large enough for the whole screen (``w x h / 8 + 8`` bytes)
```

amanita-citrina:
I run the `main.c` provided in the issue description, which does use `LV_DISPLAY_RENDER_PARTIAL`. 

Just checked again:
- WxH/8 + 8 -> OK
- WxH/16 + 8 -> OK
- WxH/8 -> FAIL
- WxH/16 -> FAIL

kisvegabor:
I could reproduce it and pushed a fix.

kisvegabor:
Fantastic! The docs are updated as you have suggested. Thank you for testing it even on the weekend.

xlla:
hi @kisvegabor , even so it is work, I think it should draw precisely using bitwise operation for I1 format to keep outside pixel untouched, for tiny device, align to Byte will lost a lot of usable space.

such as 128x64 display, only 8 position can put a button(widget).
where is the actual drawing operation occurred, I'd like to review on it.

kisvegabor:
Drawing always happens precisely. LVGL just draws a larger area. 

E.g. if a single pixel need to be updated, you will get 8 pixels to let you send 1 full byte to the display controller. Otherwise you wouldn't know the other 7 bits.

xlla:
I'v pull the latest pr-6714 and test on device, it works without random freeze , soft reset or out of memory exception.
lv-monochrom-test

another suggestion for monochrome wiki,  since the code consume row buffer by stride ( it is round up to bytes) , so for buffer size allocation, it should base on this theory too


> meaning ``(horizontal_resolution x vertical_resolution / 8) + 8`` bytes.

the precise size = ((horizontal_resolution + 7 ) // 8 x vertical_resolution) + 8

amanita-citrina:
Maybe you want to update `lv_conf_template.h` to mention I1. Now it is (line29):
`/*Color depth: 8 (A8), 16 (RGB565), 24 (RGB888), 32 (XRGB8888)*/`
but should probably become:
`/*Color depth: 1 (I1), 8 (A8), 16 (RGB565), 24 (RGB888), 32 (XRGB8888)*/`

kisvegabor:
Updated.

kisvegabor:
This PR should be included in v9.2 for sure, and it was already reviews and tested by multiple people, so I merge it now.

amanita-citrina (approved):

PR #6743, opened 2024-08-27
feat(stdlib): add lv_calloc function

onecoolx:

### Notes

### DIFF
--- a/src/stdlib/lv_mem.c
+++ b/src/stdlib/lv_mem.c
@@ -116,6 +116,12 @@ void * lv_malloc_zeroed(size_t size)
     return alloc;
 }
 
+void * lv_calloc(size_t num, size_t size)
+{
+    LV_TRACE_MEM("allocating number of %lu each %lu bytes", (unsigned long)num, (unsigned long)size);
+    return lv_malloc_zeroed(num * size);
+}
+
 void lv_free(void * data)
 {
     LV_TRACE_MEM("freeing %p", data);
--- a/src/stdlib/lv_mem.c
+++ b/src/stdlib/lv_mem.c
@@ -122,6 +122,11 @@ void * lv_calloc(size_t num, size_t size)
     return lv_malloc_zeroed(num * size);
 }
 
+void * lv_zalloc(size_t size)
+{
+    return lv_malloc_zeroed(size);
+}
+
 void lv_free(void * data)
 {
     LV_TRACE_MEM("freeing %p", data);
--- a/src/stdlib/lv_mem.c
+++ b/src/stdlib/lv_mem.c
@@ -118,7 +118,7 @@ void * lv_malloc_zeroed(size_t size)
 
 void * lv_calloc(size_t num, size_t size)
 {
-    LV_TRACE_MEM("allocating number of %lu each %lu bytes", (unsigned long)num, (unsigned long)size);
+    LV_TRACE_MEM("allocating number of %zu each %zu bytes", num, size);
     return lv_malloc_zeroed(num * size);
 }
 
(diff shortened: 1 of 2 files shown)


onecoolx:
cc @XuNeo @FASTSHIFT

src/stdlib/lv_mem.c
```diff
@@ -116,6 +116,12 @@ void * lv_malloc_zeroed(size_t size)
     return alloc;
 }
 
+void * lv_calloc(size_t num, size_t size)
+{
+    LV_TRACE_MEM("allocating number of %lu each %lu bytes", (unsigned long)num, (unsigned long)size);
```
XuNeo:
```suggestion
    LV_TRACE_MEM("allocating number of %zu each %zu bytes", num, size);
```

FASTSHIFT:
What is the purpose of adding this API?

onecoolx:
> What is the purpose of adding this API?

`calloc` is a C standard library function, adding `lv_calloc` facilitates the porting of some standard algorithms and code to LVGL

src/stdlib/lv_mem.c
onecoolx: > 
Consistent with other code, and not all compilers support `% z`

XuNeo: It's already used for example in line 77. And it's a standard from C99 according to https://cplusplus.com/reference/cstdio/printf/.

XuNeo:
`lv_calloc` and `lv_zalloc` is prefered than `lv_malloc_zeroed` in my opinion.
Maybe we should add `lv_zalloc` too?

src/stdlib/lv_mem.c
XuNeo: It's already used for example in line 77. And it's a standard from C99 according to https://cplusplus.com/reference/cstdio/printf/.

onecoolx:
> `lv_calloc` and `lv_zalloc` is prefered than `lv_malloc_zeroed` in my opinion.
> 
> Maybe we should add `lv_zalloc` too?

I think it is a good idea.

kisvegabor:
I also vote for `lv_zalloc` to keep the parameter list similar to `lv_malloc`

kissa96:
This change seems nesecarry and useful. Tested it manually as well, everything works.
I support the implementation of an lv_zalloc too.

liamHowatt (commented):
From the manpage:

> If the multiplication of nmemb and size would result in integer overflow, then calloc() returns an error. By contrast, an integer overflow would not be detected in the following call to malloc(), with the result that an incorrectly sized block of memory would be allocated: malloc(nmemb * size);

Would this be good to have? Or no?

kisvegabor:
> Would this be good to have? Or no?

Well, the overflow happens when we are about to allocate 4GB on an 32 bit MCU. Shouldn't be an issue :smile:

src/stdlib/lv_mem.c
```diff
@@ -116,6 +116,17 @@ void * lv_malloc_zeroed(size_t size)
     return alloc;
 }
 
+void * lv_calloc(size_t num, size_t size)
```
kisvegabor, line 119: Do we need `calloc` too? For me
```c
calloc(12, 30);
zalloc(12 * 30);
```

are very similar, so I wonder why we need a new function to use `(..., ...)` instead of `(... * ...)`

XuNeo (approved):
A possible use case could be for third party libs to use lvgl memory.

We can add -Dcalloc=lv_calloc to CFLAGS

kisvegabor:
It feels a little hacky, but if there is a use case, I'm ok.
@kissa96 also pointed out that it can be useful during debugging to see the count and size separately.

kisvegabor (approved):

PR #6752, opened 2024-08-28
fix(ime_pinyin): fix letter count wrong when using some dictionary

squi2rel:
When I'm using custom dictionary, I noticed some letter just can't match anything. I check the built-in dictionary and found
image
and my dictionary
image
the algorithm is depends on the next letter, so I got this:
image
and I can't type all word start with 'h'.

The [Documentation](https://github.com/lvgl/lvgl/blob/master/docs/others/ime_pinyin.rst) was not mentioned about this, so maybe use the real letter to calculate index is a better way?

### Notes

### DIFF
--- a/src/others/ime/lv_ime_pinyin.c
+++ b/src/others/ime/lv_ime_pinyin.c
@@ -877,8 +877,8 @@ static void init_pinyin_dict(lv_obj_t * obj, const lv_pinyin_dict_t * dict)
         }
         else {
             headletter = dict[i].py[0];
+            pinyin_ime->py_num[letter_calc] = offset_count;
             letter_calc = headletter - 'a';
-            pinyin_ime->py_num[letter_calc - 1] = offset_count;
             offset_sum += offset_count;
             pinyin_ime->py_pos[letter_calc] = offset_sum;
 


liamHowatt (approved):
Interesting! Letters with no phrases have an empty string entry as a placeholder to bridge the next letter. I confirmed that "h..." phrases don't work if the "i" placeholder is removed, and your fix works.

I tested in a debugger to see if there are any consequences for leaving parts of `py_num` and `py_pos` as the initial zero value and it looks totally fine.

Thank you!

Are you able to confirm if this part is correct? A dictionary will never have i, u, or v?
```c
    if(*py_str == '\0')    return NULL;
    if(*py_str == 'i')     return NULL;
    if(*py_str == 'u')     return NULL;
    if(*py_str == 'v')     return NULL;
    if(*py_str == ' ')     return NULL;
```

squi2rel:
> Interesting! Letters with no phrases have an empty string entry as a placeholder to bridge the next letter. I confirmed that "h..." phrases don't work if the "i" placeholder is removed, and your fix works.
> 
> I tested in a debugger to see if there are any consequences for leaving parts of `py_num` and `py_pos` as the initial zero value and it looks totally fine.
> 
> Thank you!
> 
> Are you able to confirm if this part is correct? A dictionary will never have i, u, or v?
> 
> ```c
>     if(*py_str == '\0')    return NULL;
>     if(*py_str == 'i')     return NULL;
>     if(*py_str == 'u')     return NULL;
>     if(*py_str == 'v')     return NULL;
>     if(*py_str == ' ')     return NULL;
> ```

These part are right. In fact, there are no Chinese characters starting with 'i', 'u' or 'v'.

kisvegabor (approved):

Issue #4576, opened 2023-09-20
Arc : clickable area outside of the draw 


spilz87:
### Perform all steps below and tick them with [x]
- [x] Read the [FAQ](https://docs.lvgl.io/master/intro/index.html#faq)
- [x] Check the related part of the [Documentation](https://docs.lvgl.io/)
- [x] Update lvgl to the latest version
- [ ] Reproduce the issue in a [Simulator](https://docs.lvgl.io/latest/en/html/get-started/pc-simulator.html)

### Describe the bug

With a partial arc (120°), when I click outside of these 120° but still in the ring, it changes the value.
which can be a big issue when it controls volume for example and it put it at 100% :(

### To Reproduce

```
// Volume
  lv_obj_t *volume_label = lv_label_create(lv_scr_act());
  lv_label_set_text(volume_label, "Value");

  /*Create Volume Arc*/
  volume_arc = lv_arc_create(lv_scr_act());
  lv_obj_set_size(volume_arc, 225, 225);
  // lv_arc_set_rotation(volume_arc, 135);
  lv_arc_set_rotation(volume_arc, 210);
  lv_obj_set_style_arc_width(volume_arc, 5, LV_PART_MAIN); // Changes background arc width
  lv_obj_set_style_arc_width(volume_arc, 6, LV_PART_INDICATOR); // Changes set part width 
  // lv_arc_set_bg_angles(volume_arc, 0, 270);
  lv_arc_set_bg_angles(volume_arc, 0, 120);
  lv_arc_set_value(volume_arc, 10);
  lv_obj_remove_style(volume_arc, NULL, LV_PART_KNOB);   /*Be sure the knob is not displayed*/
  lv_obj_center(volume_arc);
  lv_obj_add_event_cb(volume_arc, value_changed_event_cb, LV_EVENT_VALUE_CHANGED, volume_label);
```
and
```
static void value_changed_event_cb(lv_event_t * e)
{
    lv_obj_t * arc = lv_event_get_target(e);
    lv_obj_t * label = (lv_obj_t *)lv_event_get_user_data(e);

    lv_label_set_text_fmt(label, "%d%%", lv_arc_get_value(arc));
}
```

### Expected behavior

I would expect it change only when I click on the drown part of the arc, not on the full ring
### Screenshots or video

None

thanks in advance for your help

C47D:
I think this patch make it work as you expected @spilz87 

Does it look good @kisvegabor ?
```patch
[...]

PR #6753, opened 2024-08-28
fix(arc): ignore hits that are outside drawn background arc

triktron:
### Summary:
When two arcs share the same center and radius but have non-overlapping background angles, the lower arc is not interactable. The current implementation checks whether the interaction is at the correct distance from the center but does not verify if it falls within the correct arc range.

### Changes:
Added an arc range check to ensure that the interaction occurs within the specified angles of the arc.

### Implementation Details:
The code checks the angular range of the arcs, in addition to the existing distance check, to determine if the interaction should be processed. This logic is based on similar checks already implemented in the LV_EVENT_PRESSING event, as referenced in https://github.com/lvgl/lvgl/issues/4576 .

### Example Usage:
```c
lv_obj_t *arc1 = lv_arc_create(scr);
lv_arc_set_bg_angles(arc1, 5, 175);
lv_arc_set_value(arc1, 10);
lv_obj_set_size(arc1, 380, 380);
lv_obj_center(arc1);
lv_obj_add_flag(arc1, LV_OBJ_FLAG_ADV_HITTEST);

lv_obj_t *arc2 = lv_arc_create(scr);
lv_arc_set_bg_angles(arc2, 185, 355);
lv_arc_set_value(arc2, 10);
lv_obj_set_size(arc2, 380, 380);
lv_obj_center(arc2);
lv_obj_add_flag(arc2, LV_OBJ_FLAG_ADV_HITTEST);
```

### Testing:
Tested with overlapping and non-overlapping arcs with the same center and radius.
Verified that interactions are processed correctly based on the angular range.

### Notes:
This change ensures that arcs with distinct background angles remain independently interactable even when they share the same center and radius.

### Notes:


src/widgets/arc/lv_arc.c
```diff
@@ -655,6 +655,24 @@ static void lv_arc_event(const lv_obj_class_t * class_p, lv_event_t * e)
             return;
         }
 
+        /*Calculate the angle of the pressed point*/
+        lv_value_precise_t angle = lv_atan2(info->point->y - p.y, info->point->x - p.x);
+        angle -= arc->rotation;
+        angle -= arc->bg_angle_start;  /*Make the angle relative to the start angle*/
+
+        /* If we click near the bg_angle_start the angle will be close to 360° instead of a small angle */
+        if(angle < 0) angle += 360;
```
liamHowatt:
```suggestion
        while(angle < 0) angle += 360;
```

I was able to break your example by adding
```c
    lv_arc_set_rotation(arc1, 355);
    lv_arc_set_rotation(arc2, 355);
```

Maybe also change it in the other place this logic is.

A "better" (but less readable) alternative would be `angle = (angle % 360 + 360) % 360;` according to ChatGPT.

liamHowatt (commented):
Awesome!

Nice application of `LV_OBJ_FLAG_ADV_HITTEST`.

kisvegabor:
Nice! Could you add a [test](https://github.com/lvgl/lvgl/tree/master/tests#readme) as well?

src/widgets/arc/lv_arc.c
triktron: Thanks for catching that!

I've pushed a new commit to ensure the angle remains between 0 and 360 in several instances. After re-testing, it seems to have resolved the issue when adding large positive or negative angles.

There may be other places where this could occur, such as in lv_arc_set_start_angle. However, since I'm still relatively new to LVGL, I'm not entirely sure how it internally handles angles outside the 0-360 range.

Edit:
I just noticed it failed some tests, so I'll look into fixing those when I have the time.

liamHowatt:
Fyi the CI is failing because `lv_value_precise_t` can be a float depending on the configuration and `%` cannot directly be used on floats. I don't know if this is the best fix but maybe we could do something like this? If there's a better idea I'm open to it. Or maybe the `while` idea is better. Not sure.

In the defines section at the top
```c
#if LV_USE_FLOAT
#define VALUE_PRECISE_MOD(x, y) fmodf((x), (y))
#else
#define VALUE_PRECISE_MOD(x, y) (x) % (y)
#endif
```

And then you can do
```c
angle = VALUE_PRECISE_MOD(VALUE_PRECISE_MOD(angle, 360) + 360, 360);
```

triktron:
I had a similar idea in mind, but was unsure of it would cause trouble including the math.h lib.
I ended up replacing it locally with this:
```c
while (angle >= 360) angle -= 360;
while (angle < 0) angle += 360;
```
This works without having to add any dependency, but feels wasteful to me, not that i any test either one would be faster.
This approach would be more readable in my opinion.

I haven't pushed this yet as I'm first trying to set up the unit tests for this case.

triktron:
> Nice! Could you add a [test](https://github.com/lvgl/lvgl/tree/master/tests#readme) and mention this behavior in the docs as well?

I wasn't 100% sure what to do for the documentation. would changing the paragraph https://github.com/lvgl/lvgl/blob/2c021b69dcc68ee4a6d443bc3c903de1b878dd93/docs/widgets/arc.rst?plain=1#L108-L116 to the following suffice?  

```
Advanced hit test
-----------------

If the :cpp:enumerator:`LV_OBJ_FLAG_ADV_HITTEST` flag is enabled the arc can be
clicked through in the middle. Now, clicks are recognized only when they occur
on the ring of the arc and within the defined start and end angles. The end angles
are extended by :cpp:expr:`lv_dpx(50U)` to allow for a slightly larger interactive area.
Additionally, :cpp:func:`lv_obj_set_ext_click_size` can be used to further increase
the sensitive area both inside and outside the arc by a specified number of pixels.
```

kisvegabor:
I also vote for `while(angle < 0) angle += 360;` because it's simpler and equally good. 

Regarding the docs: based on the latest code I think this is happening:

```rst
Interactive area
----------------

By default :cpp:enumerator:`LV_OBJ_FLAG_ADV_HITTEST` is disabled which 
means the arc's whole area is interactive. As usual :cpp:func:`lv_obj_set_ext_click_size` can be used to increase
the sensitive area outside the arc by a specified number of pixels.

If :cpp:enumerator:`LV_OBJ_FLAG_ADV_HITTEST` is enabled the arc will be sensitive only in the range of start and end background angles and on the arc itself (not inside the arc). In this case ``ext_click_size`` makes the sensitive area ticker both inward and outward. 
```

Please check if it's corect.

triktron:
Sounds good, i think it explains it nicely! I've added a small part to include the extended arc ranges.

kisvegabor (approved):
Great, thank you!

liamHowatt (approved):
Thank you!

PR #6758, opened 2024-08-29
Revert "perf(draw): skip empty draw tasks (#6720)"

FASTSHIFT:
This reverts commit ed4cb1992628f750053f0641435f064a94ee6971.

After stress testing, it was found that `lv_draw_dispatch_layer` cannot be simply executed by checking whether `layer->draw_task_head` is empty.

After the task type is `LV_DRAW_TASK_TYPE_LAYER`, it may be necessary to execute `lv_draw_dispatch_layer` once more to switch it from `LV_DRAW_TASK_STATE_WAITING` state to `LV_DRAW_TASK_STATE_QUEUED` state, see function `lv_draw_dispatch_layer`:

```c
bool lv_draw_dispatch_layer(lv_display_t * disp, lv_layer_t * layer)
{
    ...
    /*This layer is ready, enable blending its buffer*/
    if(layer->parent && layer->all_tasks_added && layer->draw_task_head == NULL) {
        /*Find a draw task with TYPE_LAYER in the layer where the src is this layer*/
        lv_draw_task_t * t_src = layer->parent->draw_task_head;
        while(t_src) {
            if(t_src->type == LV_DRAW_TASK_TYPE_LAYER && t_src->state == LV_DRAW_TASK_STATE_WAITING) {
                lv_draw_image_dsc_t * draw_dsc = t_src->draw_dsc;
                if(draw_dsc->src == layer) {
                    t_src->state = LV_DRAW_TASK_STATE_QUEUED;
                    lv_draw_dispatch_request();
                    break;
                }
            }
            t_src = t_src->next;
        }
    }
    ...
}
```

Otherwise, the `lv_draw_dispatch` function execution may cause a busy loop. Because there is a task of `LV_DRAW_TASK_TYPE_LAYER` left in the drawing queue, and this task is always in the `WAITING` state.

No case that can reproduce the problem independently has been found yet, so revert is performed first.

Can this be modified and merged into release v9.2? @kisvegabor 

### Notes

### DIFF
--- a/src/draw/lv_draw.c
+++ b/src/draw/lv_draw.c
@@ -179,8 +179,7 @@ void lv_draw_dispatch(void)
     while(disp) {
         lv_layer_t * layer = disp->layer_head;
         while(layer) {
-            /* If there are no tasks in the layer, skip it */
-            if(layer->draw_task_head && lv_draw_dispatch_layer(disp, layer))
+            if(lv_draw_dispatch_layer(disp, layer))
                 task_dispatched = true;
             layer = layer->next;
         }


FASTSHIFT:
Is there any way to determine whether the current display needs to execute lv_draw_dispatch_layer? When there are multiple displays registered, skip the drawing of the display without redraw area.
If this is not done, vg_lite_finish will be called in each drawing task, resulting in performance degradation.
The correct situation should be to execute vg_lite_finish after the last drawing task call is completed to ensure that both the GPU and the CPU are in an efficient state.

kisvegabor:
I understand why it needs to be reverted and why it's freezing now, but I'm not sure about VG-Lite part. 

> If this is not done, vg_lite_finish will be called in each drawing task, resulting in performance degradation.

Is the problem that if there are no more draw tasks on that layer the dispatcher still will be called [here](https://github.com/lvgl/lvgl/blob/master/src/draw/lv_draw.c#L273-L280) casing the call of `vg_lite_finish`? If so, we can just change [this line](https://github.com/lvgl/lvgl/blob/master/src/draw/lv_draw.c#L272) to `else if(layer->draw_task_head)`.

XuNeo (approved):

FASTSHIFT:
> Is the problem that if there are no more draw tasks on that layer the dispatcher still will be called [here](https://github.com/lvgl/lvgl/blob/master/src/draw/lv_draw.c#L273-L280) casing the call of vg_lite_finish?

Yes.

> If so, we can just change [this line](https://github.com/lvgl/lvgl/blob/master/src/draw/lv_draw.c#L272) to else if(layer->draw_task_head).

OK, I will try it. To be on the safe side, I suggest merge this revert first to avoid affecting other people's development.
In addition, this performance optimization PR is already in release v9.2, does it also need to be reverted?

kisvegabor:
> In addition, this performance optimization PR is already in release v9.2, does it also need to be reverted?

Let's release v9.2.1 in a week.

kisvegabor (approved):

XuNeo:
> Let's release v9.2.1 in a week.

Hi @kisvegabor , which branch v9.2.1 will be used? Is it release/v9.2 or master?

kisvegabor:
Well it wasn't released in a week. :smile: 

Yes, we can cherry-pick commit to `release/v9.2`

Issue #6690, opened 2024-08-15
Ability to permanently disable back button on all menus


conor-third:
### Introduce the problem

Currently you can disable (not visible) the root back button for the root menu page but nested menus where the depth is >= 2 will always draw a back button. I have a particular use case where the navigation of the menus is done elsewhere and the back button is not required.

### Proposal

Potentially change the following line in lv_menu.c:

`if(menu->cur_depth >= 2 || menu->mode_root_back_btn == LV_MENU_ROOT_BACK_BUTTON_ENABLED)`

to use '&&' instead of '||'. The caveat being that you will have to call "lv_menu_set_mode_root_back_button" if you do want to have the built in back button for nested menus/pages. There potentially could be other side effects to this but I haven't gave it extensive thought.

kisvegabor:
Makes sense to me. Could you send a Pull request with suggested change?

kisvegabor:
Finally the PR is not merged as it would break backward compatibility. Thank you for looking into it anyways! :+1:

PR #6783, opened 2024-09-03
feat(menu): disable root back button on sub menus by default

conor-third:
See issue #6690

### DIFF
--- a/src/widgets/menu/lv_menu.c
+++ b/src/widgets/menu/lv_menu.c
@@ -256,7 +256,7 @@ void lv_menu_set_page(lv_obj_t * obj, lv_obj_t * page)
     }
     else {
         /* With sidebar disabled */
-        if(menu->cur_depth >= 2 || menu->mode_root_back_btn == LV_MENU_ROOT_BACK_BUTTON_ENABLED) {
+        if(menu->cur_depth >= 2 && menu->mode_root_back_btn == LV_MENU_ROOT_BACK_BUTTON_ENABLED) {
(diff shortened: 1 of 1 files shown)


kisvegabor:
Thank you! 

It's a breaking change as now the back button never will be displayed on the main page. Is that correct?

conor-third:
What about removing the cur_depth check so the root back button mode applies to all menus? It's probably not an ideal solution as setting the root back button mode should only apply to the root back button. Alternatively, I've adjusted my code to disable the back button on menu page loaded events. I'm happy to close this PR since this is a niche use case, let me know if you think otherwise.

kisvegabor:
Using `LV_MENU_ROOT_BACK_BUTTON_ENABLED` for non-root menus is counterintuitive. 

To stay on the safe side, let's close this PR for now, but please add your workaround in a comment for future readers.

conor-third:
As a workaround I register a callback with the root menu object:

`lv_obj_add_event_cb(root_menu, PageLoadedHandler, LV_EVENT_VALUE_CHANGED, root_menu);`

And in the callback:

`lv_obj_add_flag(root_menu->main_header_back_btn, LV_OBJ_FLAG_HIDDEN);`
`lv_obj_remove_flag(root_menu->main_header_back_btn, LV_OBJ_FLAG_CLICKABLE);`

Not an ideal solution as its called every time a page in the menu is loaded but it does the job.

kisvegabor:
Thank you!

PR #6789, opened 2024-09-03
chore(Kconfig): add version info to Kconfig file to check mismatch

### LINKED ISSUE #6576: Add method to detect not synced Kconfig, specially for NuttX
>              OK - it's a NuttX thing. It seems that NuttX takes a copy of LVGLs Kconfig file and uses it locally rather than using the LVGL one directly. Vaguely remember some NuttX discussion about this - so the solution is imperfect but I understand it now.

_Originally posted by @TimJTi in https://github.com/lvgl/lvgl/issues/6313#issuecomment-2245290554_

Since the build system of NuttX has limitation of using third-party Kconfig that downloaded during build. NuttX copies the Kconfig file to work around it. But when migrating between different LVGL version, it's likely happen the Kconfig is out of sync with LVGL's Kconfig. We can use version number to detect it and report error if so.

XuNeo:
For NuttX, lvgl/Kconfig file is copied out, add version info in Kconfig file to detect if the Kconfig is out of sync with source code.

Fix #6576

### Notes

### DIFF
--- a/Kconfig
+++ b/Kconfig
@@ -1527,15 +1527,25 @@ menu "LVGL configuration"
 			default y
 			help
 				This can save some memory, but not much. After the quick access bar is created, it can be hidden by clicking the button at the top left corner of the browsing area, which is very useful for small screen devices.
-	
+
 		config LV_USE_FONT_MANAGER
 			bool "Enable freetype font manager"
 			depends on LV_USE_FREETYPE
-			default n			
+			default n
 		config LV_FONT_MANAGER_NAME_MAX_LEN
 			int "Font manager name max length"
 			depends on LV_USE_FONT_MANAGER
 			default 32
+
+		config LVGL_VERSION_MAJOR
+			int
+			default 9 # LVGL_VERSION_MAJOR
+		config LVGL_VERSION_MINOR
+			int
+			default 2 # LVGL_VERSION_MINOR
+		config LVGL_VERSION_PATCH
+			int
+			default 0 # LVGL_VERSION_PATCH
 	endmenu
 
 	menu "Devices"
--- a/src/lv_conf_kconfig.h
+++ b/src/lv_conf_kconfig.h
@@ -18,6 +18,16 @@ extern "C" {
 
 #  ifdef __NuttX__
 #    include <nuttx/config.h>
+/*
+ * Make sure version number in Kconfig file is correctly set.
+ * Mismatch can happen when user manually copy lvgl/Kconfig file to their project, like what NuttX does.
+ */
+#    include "../lv_version.h"
+
+#    if CONFIG_LVGL_VERSION_MAJOR != LVGL_VERSION_MAJOR || CONFIG_LVGL_VERSION_MINOR != LVGL_VERSION_MINOR \
+        || CONFIG_LVGL_VERSION_PATCH != LVGL_VERSION_PATCH
+#        warning "Version mismatch between Kconfig and lvgl/lv_version.h"
+#    endif
 #  elif defined(__RTTHREAD__)
 #    define LV_CONF_INCLUDE_SIMPLE
 #    include <lv_rt_thread_conf.h>
--- a/scripts/update_version.py
+++ b/scripts/update_version.py
@@ -122,6 +122,35 @@ class CmakeReplacer(MacroReplacer):
     def getReplacement(self, val: str):
         return r'\g<1>' + val + r'\g<3>'
 
+class KconfigReplacer(RepoFileVersionReplacer):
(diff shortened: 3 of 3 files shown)


src/lv_conf_kconfig.h
```diff
@@ -18,6 +18,15 @@ extern "C" {
...
 #  ifdef __NuttX__
 #    include <nuttx/config.h>
+/*
+ * Make sure version number in Kconfig file is correctly set.
+ * Mismatch can happen when user manually copy lvgl/Kconfig file to their project, like what NuttX does.
+ */
+#    include "../lvgl/lv_version.h"
+
+#if CONFIG_LVGL_VERSION_MAJOR != LVGL_VERSION_MAJOR || CONFIG_LVGL_VERSION_MINOR != LVGL_VERSION_MINOR
+#    error "Version mismatch between Kconfig and lvgl/lv_version.h"
```
FASTSHIFT: It is recommended to use the warning level, and the mismatched configuration has a default value

Kconfig
```diff
@@ -1473,6 +1473,16 @@ menu "LVGL configuration"
 			int "Font manager name max length"
 			depends on LV_USE_FONT_MANAGER
 			default 32
+
+		config LVGL_VERSION_MAJOR
+			int "Major version"
```
kisvegabor: Can someone change it by accident? If so I suggest 

```suggestion
			int "Major version (used internally to check the version, do not change it)"
```

kisvegabor (commented):
Please add these to [update_version.py](https://github.com/lvgl/lvgl/blob/master/scripts/update_version.py) as well.

XuNeo:
> Please add these to [update_version.py](https://github.com/lvgl/lvgl/blob/master/scripts/update_version.py) as well.

@becseya  Could you help on this with a separate PR? Current update_version.py is based on line, while Kconfig needs to match several lines.

```patch
++class KconfigReplacer(MacroReplacer):
++    """Replace version info in Kconfig file"""
++
++    def getPattern(self, key: str):
++        # Match the version fields in Kconfig file
++        return rf'(^\s*config\s+{key}\n\s*int\s*\n.*\n\s*default\s+)(\d+)'
++
++    def getReplacement(self, val: str):
++        # Replace the version value
++        return r'\g<1>' + val
++
```

becseya:
@XuNeo @kisvegabor  Sure, I can help with that in a separate PR.

Kconfig
```diff
@@ -1473,6 +1473,16 @@ menu "LVGL configuration"
...
+			default 9
+		config LVGL_VERSION_MINOR
+			int "Minor version"
+			default 2
```
kisvegabor: What about to match a single line?
```suggestion
			default 2  #LVGL minor version
```

Kconfig
XuNeo: Added.

src/lv_conf_kconfig.h
```diff
@@ -18,6 +18,15 @@ extern "C" {
 
 #  ifdef __NuttX__
 #    include <nuttx/config.h>
+/*
+ * Make sure version number in Kconfig file is correctly set.
+ * Mismatch can happen when user manually copy lvgl/Kconfig file to their project, like what NuttX does.
+ */
+#    include "../lvgl/lv_version.h"
+
+#    if CONFIG_LVGL_VERSION_MAJOR != LVGL_VERSION_MAJOR || CONFIG_LVGL_VERSION_MINOR != LVGL_VERSION_MINOR
```
XuNeo: I did not compare PATCH version. Please suggest if is needed.

FASTSHIFT (dismissed):

kisvegabor (commented):
My review wasn't sent, sorry...

src/lv_conf_kconfig.h
kisvegabor: I think matching major and minor should solve most of the issues. Theoretically, it can happen that we fix a Kconfig entry in a patch release but it's very rare. It would happen more often that people need to update Kconfig for nothing.

XuNeo: Added PATCH version check to be sure.

XuNeo:
I have updated suggests from @kisvegabor , now the update_version.py can cover this case too.
I'll pick to release/v9.2 once it's reviewed and merged.

liamHowatt (approved):

kisvegabor (approved):

kisvegabor:
> I'll pick to release/v9.2 once it's reviewed and merged.

I've discussed it with @nicusorcitu that we can release v9.2.1 at the beginning of October.

PR #6793, opened 2024-09-04
fix(freertos) sync signal from isr fixed

cosmindanielradu19:
Fixes [#6692](https://github.com/lvgl/lvgl/pull/6692)
`lv_thread_sync_signal_isr` needed to call `_enter_critical_isr()` instead of `_enter_critical()` because it was hanging

### Notes

### DIFF
--- a/Kconfig
+++ b/Kconfig
@@ -162,6 +162,15 @@ menu "LVGL configuration"
 			string "Custom OS include header"
 			default "stdint.h"
 			depends on LV_OS_CUSTOM
+
+        config LV_USE_FREERTOS_TASK_NOTIFY
+            bool "Use RTOS task with a direct notification for synchronization"
+            default y
+            depends on LV_OS_FREERTOS
+        help
+            Unblocking an RTOS task with a direct notification is 45% faster and uses less RAM
+            than unblocking a task using an intermediary object such as a binary semaphore.
+            RTOS task notifications can only be used when there is only one task that can be the recipient of the event.
 	endmenu
 
 	menu "Rendering Configuration"
--- a/src/osal/lv_freertos.c
+++ b/src/osal/lv_freertos.c
@@ -15,11 +15,7 @@
 #include "lv_os.h"
 #if LV_USE_OS == LV_OS_FREERTOS
 
-#if (ESP_PLATFORM)
-    #include "freertos/atomic.h"
-#else
-    #include "atomic.h"
-#endif
+#include "atomic.h"
 
 #include "../tick/lv_tick.h"
 #include "../misc/lv_log.h"
--- a/src/osal/lv_freertos.c
+++ b/src/osal/lv_freertos.c
@@ -54,6 +54,8 @@ static void prvCondInit(lv_thread_sync_t * pxCond);
 
 static void prvCheckCondInit(lv_thread_sync_t * pxCond);
 
+static void prvCheckCondInitIsr(lv_thread_sync_t * pxCond);
+
 #if !USE_FREERTOS_TASK_NOTIFY
 static void prvTestAndDecrement(lv_thread_sync_t * pxCond,
                                 uint32_t ulLocalWaitingThreads);
@@ -74,9 +76,13 @@ static void prvTestAndDecrement(lv_thread_sync_t * pxCond,
 #if (ESP_PLATFORM)
     #define _enter_critical()   taskENTER_CRITICAL(&critSectionMux);
     #define _exit_critical()    taskEXIT_CRITICAL(&critSectionMux);
+    #define _enter_critical_isr() taskENTER_CRITICAL_FROM_ISR();
+    #define _exit_critical_isr(x) taskEXIT_CRITICAL_FROM_ISR(x);
 #else
     #define _enter_critical()   taskENTER_CRITICAL();
     #define _exit_critical()    taskEXIT_CRITICAL();
+    #define _enter_critical_isr() taskENTER_CRITICAL_FROM_ISR();
+    #define _exit_critical_isr(x) taskEXIT_CRITICAL_FROM_ISR(x);
 #endif
 
 /**********************
@@ -196,19 +202,19 @@ lv_result_t lv_thread_sync_wait(lv_thread_sync_t * pxCond)
     prvCheckCondInit(pxCond);
 
 #if USE_FREERTOS_TASK_NOTIFY
-    TaskHandle_t current_task_handle = xTaskGetCurrentTaskHandle();
+    TaskHandle_t xCurrentTaskHandle = xTaskGetCurrentTaskHandle();
 
     _enter_critical();
-    BaseType_t signal_sent = pxCond->xSyncSignal;
+    BaseType_t xSyncSygnal = pxCond->xSyncSignal;
     pxCond->xSyncSignal = pdFALSE;
-    if(signal_sent == pdFALSE) {
+    if(xSyncSygnal == pdFALSE) {
         /* The signal hasn't been sent yet. Tell the sender to notify this task */
-        pxCond->xTaskToNotify = current_task_handle;
+        pxCond->xTaskToNotify = xCurrentTaskHandle;
     }
     /* If we have a signal from the other task, we should not ask to be notified */
     _exit_critical();
 
-    if(signal_sent == pdFALSE) {
+    if(xSyncSygnal == pdFALSE) {
         /* Wait for other task to notify this task. */
         ulTaskNotifyTake(pdTRUE, portMAX_DELAY);
     }
@@ -277,18 +283,18 @@ lv_result_t lv_thread_sync_signal(lv_thread_sync_t * pxCond)
 
 #if USE_FREERTOS_TASK_NOTIFY
     _enter_critical();
-    TaskHandle_t task_to_notify = pxCond->xTaskToNotify;
+    TaskHandle_t xTaskToNotify = pxCond->xTaskToNotify;
     pxCond->xTaskToNotify = NULL;
-    if(task_to_notify == NULL) {
+    if(xTaskToNotify == NULL) {
         /* No task waiting to be notified. Send this signal for later */
         pxCond->xSyncSignal = pdTRUE;
     }
     /* If a task is already waiting, there is no need to set the sync signal */
     _exit_critical();
 
-    if(task_to_notify != NULL) {
+    if(xTaskToNotify != NULL) {
         /* There is a task waiting. Send a notification to it */
-        xTaskNotifyGive(task_to_notify);
+        xTaskNotifyGive(xTaskToNotify);
     }
     /* If there was no task waiting to be notified, we sent a signal for it to see later. */
 #else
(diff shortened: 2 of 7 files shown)


kisvegabor (dismissed):
Thank you!

nicusorcitu:
@liamHowatt

src/osal/lv_freertos.c
```diff
@@ -74,9 +74,13 @@ static void prvTestAndDecrement(lv_thread_sync_t * pxCond,
 #if (ESP_PLATFORM)
     #define _enter_critical()   taskENTER_CRITICAL(&critSectionMux);
     #define _exit_critical()    taskEXIT_CRITICAL(&critSectionMux);
+    #define _enter_critical_isr() _enter_critical()
+    #define _exit_critical_isr()  _exit_critical()
```
liamHowatt:
```suggestion
    #define _enter_critical_isr() taskENTER_CRITICAL_FROM_ISR()
    #define _exit_critical_isr(x)  taskEXIT_CRITICAL_FROM_ISR(x)
```

I tested on ESP32 and that API is available. I think it would be good to use it here too.

liamHowatt (commented):
Thank you!

src/osal/lv_freertos.c
kschwan: It has to be this or FreeRTOS will assert in `vPortExitCritical` - at least on the CM4/7 port

kschwan:
I am seeing this issue after upgrading from v9.1 to v9.2. Let me know if I can help testing. Currently working with an i.MX RT1170 CM7

nicusorcitu:
hi @liamHowatt 
I reverted the PR https://github.com/lvgl/lvgl/pull/6692 and let the notification sync as it was before. More details are on the PR https://github.com/lvgl/lvgl/pull/6692.

We can work together to fix any issue on your side.

kschwan:
> Thanks @kschwan.
> 
> > Currently working with an i.MX RT1170 CM7
> 
> So it is working now with this fix?

I'll check tomorrow! I believe it will because I made a very similar change myself before coming here

kschwan:
OK, so I checked out your branch, updated my lv_conf.h accordingly. Something's not working though. The display isn't refreshing.

I use PXP to rotate image buffers.

I attached a debugger and it looks like every thread is blocked in `lv_thread_sync_wait()` (`ulTaskNotifyTake()` @ lv_freertos.c:200)
image

kschwan:
A note: I had to patch lv_draw_pxp.h to `#include "../../sw/lv_draw_sw_private.h"`

kschwan:
If I switch to master (651f69f) and apply the following patch, it seems to run correctly (though I have done any testing otherwise)

```
diff --git a/src/draw/nxp/pxp/lv_draw_pxp.h b/src/draw/nxp/pxp/lv_draw_pxp.h
index d5e98d791..eb0f69e17 100644
--- a/src/draw/nxp/pxp/lv_draw_pxp.h
+++ b/src/draw/nxp/pxp/lv_draw_pxp.h
@@ -24,6 +24,7 @@ extern "C" {

 #if LV_USE_DRAW_PXP
 #include "../../sw/lv_draw_sw.h"
+#include "../../sw/lv_draw_sw_private.h"

 /*********************
  *      DEFINES
diff --git a/src/osal/lv_freertos.c b/src/osal/lv_freertos.c
index 1c81a8e05..86eac9a82 100644
--- a/src/osal/lv_freertos.c
+++ b/src/osal/lv_freertos.c
@@ -349,7 +349,7 @@ lv_result_t lv_thread_sync_signal_isr(lv_thread_sync_t * pxCond)
     prvCheckCondInit(pxCond);

 #if USE_FREERTOS_TASK_NOTIFY
-    _enter_critical();
+    UBaseType_t interrupt_status = taskENTER_CRITICAL_FROM_ISR();
     TaskHandle_t task_to_notify = pxCond->xTaskToNotify;
     pxCond->xTaskToNotify = NULL;
     if(task_to_notify == NULL) {
@@ -357,7 +357,7 @@ lv_result_t lv_thread_sync_signal_isr(lv_thread_sync_t * pxCond)
         pxCond->xSyncSignal = pdTRUE;
     }
     /* If a task is already waiting, there is no need to set the sync signal */
-    _exit_critical();
+    taskEXIT_CRITICAL_FROM_ISR(interrupt_status);

     if(task_to_notify != NULL) {
         /* There is a task waiting. Send a notification to it */
```

nicusorcitu:
I also see this blocking status with PXP draw + SW draw enabled.
The commit that caused it is that PXP interrupt now uses sync due to a PR https://github.com/lvgl/lvgl/pull/6298

So now, the PXP interrupt and LVGL main thread both uses the same thread handle to be signaled. So after the PXP IRQ comes, a different thread is being woke up.
The solution would be to initialize the PXP interrupt from the PXP thread and not from LVGL thread. I will get back with a fix for this.

nicusorcitu:
Have to solve that problem too on PXP.
PXP wait can be called from two threads now: from PXP draw thread (for drawing) and from LVGL main thread (for screen rotation).
Need to find a solution to make it work with thread synchronization because it is a single task handle that can be signaled.
The PR with changes on interrupt from semaphores to task sync should have not been merged either https://github.com/lvgl/lvgl/pull/6298

kisvegabor:
I'm reading #6692 and these comments and trying to understand the root of the issue, but I'm not quite there yet. 

The original issue was to solve calling `lv_init` and `lv_timer_handler` from different threads. If they are called from the same thread does it work well? 

---

There is some thing which might be an issue in the current logic.   In [lv_draw_dispatch()](https://github.com/lvgl/lvgl/blob/master/src/draw/lv_draw.c#L186-L189) we have this:
```c
        if(!task_dispatched) {
            lv_draw_wait_for_finish();
            lv_draw_dispatch_request();
        }
```
Note that
 - `task_dispatched==true` means that at least on draw unit is working on something. Maybe it was dispatched earlier, maybe now, doesn't matter. 
 - `if(!task_dispatched)` will be true when all the draw units are idle, they don't do anything. 
 - Then why should we `wait_for_finish`? By definition there is nothing to finish.

nicusorcitu:
`task_dispatched` is true only if dispatcher dispatched something to draw units.
So it can be false in those cases: unit does not support the task or no more tasks in the queue. If task_dispatched is false it does not necessarily mean the units are idle. Some units can be still busy processing.
Even if it is the idle case, e.g. all tasks are completed, it should work. There will be just one single wait for finish triggered which is harmless.
I put here an example on how `lv_draw_wait_for_finish()` works:
https://github.com/lvgl/lvgl/blob/dc9e42360613e8506b13d8bfc3b28b1207da10b1/src/draw/lv_draw_private.h#L118

kisvegabor:
> So it can be false in those cases: unit does not support the task or no more tasks in the queue. If task_dispatched is false it does not necessarily mean the units are idle. 

If there is no supported draw task for a unit or the list is empty, the draw unit will be idle.  Just to be sure we understand it the same way: if the draw unit it busy and cannot take new tasks, the dispatcher should return `0` (meaning 0 tasks were taken)

> Some units can be still busy processing.

The return value of all draw unit dispatches are combined into a single result [here](https://github.com/lvgl/lvgl/blob/master/src/draw/lv_draw.c#L277).

> There will be just one single wait for finish triggered which is harmless.

If it should work without this wait too, let's remove it to see if we understand what's happening correctly.

---

I'm not familiar with the FreeRTOS part, so @liamHowatt will comment on it.

nicusorcitu:
@kisvegabor 

> > Some units can be still busy processing.
> >
> The return value of all draw unit dispatches are combined into a single result [here](https://github.com/lvgl/lvgl/blob/master/src/draw/lv_draw.c#L277).

VGLite dispatcher will never return:
 0  - busy to handle the previous task (e.g. not yet queued to the GPU).
But will not return busy to signal the GPU HW is busy. That is different.
 1  - for task dispatched
-1 - otherwise (task not supported or task list empty)
`LV_DRAW_UNIT_IDLE` define need to be renamed or at least let's agree on a return code convention and make the changes appropriate on all draw units.

So, if task is not supported or task list is empty, the GPU HW might be busy, so we can call a wait for finish which will make sure the unit is done processing all tasks.
I have done a separate patch that actually does wait for completion, and it works.
I will push the patch once with these freertos changes even my intend was to send it separately together with other vglite fixes.

kisvegabor:
Why can't VGLite return 0 when:
1. Some tasks are already queued
2. There is no available draw task at this moment

`0` would indicate that I couldn't take a new task, but I'm doing something.

I think the missing information for me is the new flow of VGLite. When will VGLite render the queued tasks? Only on  `lv_draw_wait_for_finish` or are there any other cases?

nicusorcitu:
If GPU is still busy doing some other previous tasks, the tasks will get queued (max 100 tasks).
Tasks are queued and will be processed by GPU as soon as the GPU will become idle.
I tried to explain that here: https://github.com/lvgl/lvgl/blob/87f6e9d691f20703dd47d4d1454c568e552af116/src/draw/lv_draw_private.h#L141

> When will VGLite render the queued tasks? Only on lv_draw_wait_for_finish or are there any other cases?
>
The GPU will render the queued tasks:
1. When a new task is dispatched to VGLite and GPU is IDLE - GPU will process the incoming task + the previous queued tasks
2. When `lv_draw_wait_for_finish()` gets called
The second case is needed just to be sure GPU completes his work given at the end of frame.

kisvegabor:
Okay, thank you! Let's keep logic as it is :slightly_smiling_face:

nicusorcitu:
I have not changed any logic in dispatcher.
The only callback was this `lv_draw_wait_for_finish()` which can be NULL on the other dispatchers. So any other unit is not affected.

kschwan:
> Hi @liamHowatt @kschwan please let me know if this solution works for you. It works for me on all combinations: SW + VGLite + PXP SW + PXP SW + VGLite VGLite + PXP
> 
> VGLite still needs an extra patch but I will send it separately on a different PR. I intend to do it soon to fixup the remaining issues on lvgl 9.2.0

Thanks! I'll give it a spin tomorrow.

kschwan:
I'm seeing a weird issue: The display is not refreshing if I disable `LV_USE_PERF_MONITOR`. With `LV_USE_PERF_MONITOR` enabled, it seems to be working correctly

nicusorcitu:
Please open a separate thread for that issue.

kschwan:
> Please open a separate thread for that issue.

Thanks, will do

kschwan:
Looks like it's working correctly on my end :)

kisvegabor (dismissed):
Okay for me in general, but @liamHowatt can add more meaningful comments.

Kconfig
```diff
@@ -162,6 +162,14 @@ menu "LVGL configuration"
 			string "Custom OS include header"
 			default "stdint.h"
 			depends on LV_OS_CUSTOM
+
+        config LV_USE_FREERTOS_TASK_NOTIFY
+            string "Use RTOS task with a direct notification for synchronization"
+            default 1
```
liamHowatt:
```suggestion
            bool "Use RTOS task with a direct notification for synchronization"
            default y
            depends on LV_OS_FREERTOS
```

src/osal/lv_freertos.c
```diff
@@ -15,11 +15,7 @@
 #include "lv_os.h"
 #if LV_USE_OS == LV_OS_FREERTOS
 
-#if (ESP_PLATFORM)
-    #include "freertos/atomic.h"
-#else
-    #include "atomic.h"
-#endif
+#include "atomic.h"
```
liamHowatt, line 18: With ESP IDF I get `fatal error: atomic.h: No such file or directory`

liamHowatt (commented):
It works great on ESP IDF. There are a few nice catches you made.

Just two things.

src/osal/lv_freertos.c
nicusorcitu: I removed that because the file name is the same, only the path is different.
The suggestion how to fix it is in the commit message, can you please try it out?
[fix(freertos) Use the proper include path("C_INCLUDE_PATH" or "GCC -I") instead.](https://github.com/lvgl/lvgl/pull/6793/commits/3237397062e163148b7700119890a129bea06ab3)

liamHowatt: I got it to work by adding

```cmake
if(CONFIG_FREERTOS_SMP)
    target_include_directories(${COMPONENT_LIB} PRIVATE "${IDF_PATH}/components/freertos/FreeRTOS-Kernel-SMP/include/freertos/")
else()
    target_include_directories(${COMPONENT_LIB} PRIVATE "${IDF_PATH}/components/freertos/FreeRTOS-Kernel/include/freertos/")
endif()
```

to the bottom of `env_support/cmake/esp.cmake`. Please add it.

nicusorcitu: Done: [fb975da](https://github.com/lvgl/lvgl/pull/6793/commits/fb975daae68d0e1699bf03226f99a465d0b3614f)

liamHowatt (approved):
Thanks

FASTSHIFT (approved):

PR #6799, opened 2024-09-04
fix(sdl): check against NULL before using the driver data of a display

Viatorus:
We are migrating from LVGL 8 to 9.2.

When we simulate our device's GUI, we embed the device GUI as a secondary display (draw-buffer-only) in an SDL display. The SDL display shows some hardware buttons around the device GUI. So it is a display in display.

In LVGL 9, all displays are assumed to be an SDL display. But this isn't the case. The simplest workaround that works for us is to check if the drive data is not null. 

It would be even better to check what type of display it is. Unfortunately, I don't see such an option yet.


kisvegabor (approved):
Maybe we could
- storing unique IDs for the drivers (or string names): makes sense to me
- introduce a similar class system that we have for widgets: seems overkill

Anyway, it'd be a larger change. Let's merge it as a hotfix.

XuNeo (approved):

PR #6835, opened 2024-09-12
feat(sdl): speed up rotation

### LINKED ISSUE #6827: rotate180_rgb565 code optimization to yield >= 25% increase in speed
### Introduce the problem

Here is the existing function that is located in `src/draw/sw/lv_draw_sw.c`

```c
static void rotate180_rgb565(const uint16_t * src, uint16_t * dst, int32_t width, int32_t height, int32_t src_stride,
                             int32_t dest_stride)
{
    if(LV_RESULT_OK == LV_DRAW_SW_ROTATE180_RGB565(src, dst, srcWidth, srcHeight, srcStride, dstStride)) {
        return ;
    }

    src_stride /= sizeof(uint16_t);
    dest_stride /= sizeof(uint16_t);

    for(int32_t y = 0; y < height; ++y) {
        int32_t dstIndex = (height - y - 1) * dest_stride;
        int32_t srcIndex = y * src_stride;
        for(int32_t x = 0; x < width; ++x) {
            dst[dstIndex + width - x - 1] = src[srcIndex + x];
        }
    }
}
```

Now the first thing is I am not even sure how this is even managing to compile without pitching a fit about the `srcWidth`, `srcHeight`, `srcStride` and `dstStride` variables not existing. Those are passed as parameters to the `LV_DRAW_SW_ROTATE180_RGB565` macro.

But never mind that at the moment. I am not sure as to why there is all of this crazy math that is being used in order to rotate the display 180°. This can be simplified using the code below

### Proposal

```c
static void rotate180_rgb565(const uint16_t * src, uint16_t * dst, int32_t width, int32_t height, int32_t src_stride,
                             int32_t dest_stride)
{
    if(LV_RESULT_OK == LV_DRAW_SW_ROTATE180_RGB565(src, dst, srcWidth, srcHeight, srcStride, dstStride)) {
        return ;
    }
    // not sure if these need to be here or not. I believe it might matter if the macro above is not defined in userland
    LV_UNUSED(srcStride);
    LV_UNUSED(dstStride);

    uint32_t num_pixels = height * width;

    for(int32_t i = 0; i < num_pixels; i++) {
        dst[~i + num_pixels] = src[i];
    }
}
```

The code above yields a 25% increase in speed at a minimum. The results of the speed test are with the code converted to python running the test 5000 times over each function. I believe the `~` operator does the same in C code that it does in Python. In Python `~0` is `-1`.

so basically we are taking a  function that has 2 division and 2 addition, 4 subtraction, 2 multiplication done in for loops and reducing it to a bit operation and an addition done in a for loop. There is no need to muck about with the strides because when rotating 180 degrees the strides are going to be identical. all that is being done is the last pixel is being moved to the first pixel location and vice versa through the entire buffer. 

at a 25% speed increase at a minimum this is something that I feel should be done. This can also be done for the ARGB8888 and RGB888 180° functions as well. The RGB888 function will have some more addition in it.

kisvegabor:
fixes #6827

### Notes

### DIFF
--- a/src/drivers/sdl/lv_sdl_window.c
+++ b/src/drivers/sdl/lv_sdl_window.c
@@ -208,71 +208,44 @@ static inline int sdl_render_mode(void)
 static void flush_cb(lv_display_t * disp, const lv_area_t * area, uint8_t * px_map)
 {
 #if LV_USE_DRAW_SDL == 0
-    lv_area_t rotated_area;
     lv_sdl_window_t * dsc = lv_display_get_driver_data(disp);
     lv_color_format_t cf = lv_display_get_color_format(disp);
 
     if(sdl_render_mode() == LV_DISPLAY_RENDER_MODE_PARTIAL) {
-        lv_display_rotation_t rotation = lv_display_get_rotation(disp);
+        lv_area_t rotated_area = *area;
+        lv_display_rotate_area(disp, &rotated_area);
+
+        int32_t px_map_w = lv_area_get_width(area);
+        int32_t px_map_h = lv_area_get_height(area);
+        uint32_t px_map_stride = lv_draw_buf_width_to_stride(lv_area_get_width(area), cf);
         uint32_t px_size = lv_color_format_get_size(cf);
 
-        if(rotation != LV_DISPLAY_ROTATION_0) {
-            int32_t w = lv_area_get_width(area);
-            int32_t h = lv_area_get_height(area);
-            uint32_t w_stride = lv_draw_buf_width_to_stride(w, cf);
-            uint32_t h_stride = lv_draw_buf_width_to_stride(h, cf);
-            size_t buf_size = w * h * px_size;
+        int32_t fb_stride = disp->hor_res * px_size;
+        uint8_t * fb_start = dsc->fb_act;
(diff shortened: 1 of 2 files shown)


src/drivers/sdl/lv_sdl_window.c
```diff
@@ -208,71 +208,44 @@ static inline int sdl_render_mode(void)
...
         }
-    }
 
-    /* TYPICALLY YOU DO NOT NEED THIS
-     * If it was the last part to refresh update the texture of the window.*/
-    if(lv_display_flush_is_last(disp)) {
-        if(sdl_render_mode() != LV_DISPLAY_RENDER_MODE_PARTIAL) {
-            dsc->fb_act = px_map;
+        if(lv_display_flush_is_last(disp)) {
+            if(sdl_render_mode() != LV_DISPLAY_RENDER_MODE_PARTIAL) {
```
liamHowatt: This will never be true. This condition is in a block where `sdl_render_mode() == LV_DISPLAY_RENDER_MODE_PARTIAL` is true.

kisvegabor: Ah, thanks! It was off by one level.

liamHowatt (approved):

FASTSHIFT (approved):

PR #6838, opened 2024-09-12
feat(animimg): add getter function for underlying animation

lorenzo-arena:
This PR adds a getter function for the animimage class to retrieve the underlying animation reference. This is needed when someone want to tweak the animation process (for example setting it reverse or with a custom range of values) without having to create a new image array.


kisvegabor (dismissed):
Thank you!

liamHowatt (approved):

kisvegabor (approved):

Issue #6771, opened 2024-08-30
indev callback not workming correctly


kdschlosser:
### LVGL version

master branch

### What happened?

When an indev callback is made there is a structure that gets passed and that structure is supposed to get filled by the callback function is there has been user input. There is a field in that structure names `continue_reading`. This flag causes the indev code to continually loop reading the input over and over again. That part works as it should but what is missing from it is updating the display. It loops without ever doing a display refresh. 

I would have imagined the purpose of that field is to offer better performance when scrolling or moving a slider. Issue is if the display doesn't get updated it simply jumps once the touch is released. I am not sure if the `continue_reading` flag had some other purpose and if it does we can tweak the API slightly so it's purpose doesn't change, We add a `bool` return value form the callback function. Returning `true` would instruct LVGL to refresh the display immediately processing the data from the callback. It doesn't matter if there is a change to the user input or not. The continue reading would function as it should. This would only apply if there is some use case for continue reading where the indev driver should be running in a loop without ever updating the display. 

There needs to some way to continually loop the indev otherwise it is only going to get called once every 33 milliseconds and that is not going to provide the best in terms of smooth scrolling. we don't want to return until the user has released the display. having a way to control this aspect of LVGL give the user the ability to made a decision as to what is important to them. Personally I hate it when a UI doesn't respond quickly when scrolling or moving a slider.

I personally like the idea of of being able to control the loop and also the refresh separately and using a return value is a simple way to achieve that.

kisvegabor:
The purpose of `continue_reading` is to make LVGL read out all the buffered data (if any). E.g. now you have 3 data buffered, you can set `continue_reading` twice and leave it `false` on the last call. 

For the behaviour you have described you can call `lv_timer_ready(lv_indev_get_read_timer());` in the `read_cb` but note that, the coordinate and speed calculations in are not considering the read period (which is a problem and should be fixed at some point).

For example if previously `y` way `100` and now it's `110`. LVGL thinks the your finger moved 10px. However within what time frame? 10ms? 20ms? 5 sec? They all result in different speed.

So for now, be careful with pushing new coordinates in random time frames.

kdschlosser:
[...]

PR #6840, opened 2024-09-12
updates indev to allow for placing priority on user input.

kdschlosser:
This is for proof of concept and will need more work to be able to be merged if the idea is accepted.

### Notes

### DIFF
--- a/docs/porting/indev.rst
+++ b/docs/porting/indev.rst
@@ -269,6 +269,20 @@ data instead of directly reading the input device. Setting the
 ``data->continue_reading`` flag will tell LVGL there is more data to
 read and it should call ``read_cb`` again.
 
+
+User input priority
+-------------------
+
+If you want to place priority on user input and updating the display this can be done by returning `true`
+from the read callback when there is user input. This will instruct LVGL to refresh the display immediatly
+after the input has been processed. You can further extend the priority by setting the
+``data->continue_reading`` field to `true` and returning `true`. This will loop the reading of the input
+device and refreshing the display after the input as been processed. Once there is no more user input then
+you would set ``data->continue_reading`` to `false` and return `true`. You only return `true` one time
+after the user inmput ends. Calls made to the read callback after that and there is no user input would
+return `false` from the read callback.
+
+
 Switching the input device to event-driven mode
 -----------------------------------------------
 
--- a/src/drivers/libinput/lv_libinput.c
+++ b/src/drivers/libinput/lv_libinput.c
@@ -58,7 +58,7 @@ lv_libinput_event_t * _get_event(lv_libinput_t * state);
 bool _event_pending(lv_libinput_t * state);
 lv_libinput_event_t * _create_event(lv_libinput_t * state);
 
-static void _read(lv_indev_t * indev, lv_indev_data_t * data);
+static bool _read(lv_indev_t * indev, lv_indev_data_t * data);
 static void _read_pointer(lv_libinput_t * state, struct libinput_event * event);
 static void _read_keypad(lv_libinput_t * state, struct libinput_event * event);
 
@@ -380,7 +380,7 @@ lv_libinput_event_t * _create_event(lv_libinput_t * dsc)
     return evt;
 }
 
-static void _read(lv_indev_t * indev, lv_indev_data_t * data)
+static bool _read(lv_indev_t * indev, lv_indev_data_t * data)
 {
     lv_libinput_t * dsc = lv_indev_get_driver_data(indev);
     LV_ASSERT_NULL(dsc);
@@ -397,6 +397,18 @@ static void _read(lv_indev_t * indev, lv_indev_data_t * data)
     data->key = evt->key_val;
     data->continue_reading = _event_pending(dsc);
 
+    bool res = false;
+
+    if (data->state != dsc->last_event->pressed) res = true;
+
+    if (
+        data->state == LV_INDEV_STATE_PRESSED || data->key != dsc->last_event->key_val ||
+        data->point.x != dsc->last_event->point.x || data->point.y != dsc->last_event->point.y ) {
+
+        res = true;
+        data->continue_reading = true;
+    }
+
     dsc->last_event = *evt; /* Remember the last event for the next call */
 
     pthread_mutex_unlock(&dsc->event_lock);
@@ -404,6 +416,7 @@ static void _read(lv_indev_t * indev, lv_indev_data_t * data)
     if(evt)
         LV_LOG_TRACE("libinput_read: (%04d, %04d): %d continue_reading? %d", data->point.x, data->point.y, data->state,
                      data->continue_reading);
+    return res;
 }
 
 static void _read_pointer(lv_libinput_t * dsc, struct libinput_event * event)
@@ -426,7 +439,7 @@ static void _read_pointer(lv_libinput_t * dsc, struct libinput_event * event)
             pointer_event = libinput_event_get_pointer_event(event);
             break;
         default:
-            return; /* We don't care about this events */
+            return false; /* We don't care about this events */
     }
 
     /* We need to read unrotated display dimensions directly from the driver because libinput won't account
@@ -435,7 +448,7 @@ static void _read_pointer(lv_libinput_t * dsc, struct libinput_event * event)
 
     /* ignore more than 2 fingers as it will only confuse LVGL */
     if(touch_event && (slot = libinput_event_touch_get_slot(touch_event)) > 1)
-        return;
+        return false;
 
     evt = _create_event(dsc);
 
--- a/src/drivers/sdl/lv_sdl_keyboard.c
+++ b/src/drivers/sdl/lv_sdl_keyboard.c
@@ -28,7 +28,7 @@ typedef struct {
 /**********************
  *  STATIC PROTOTYPES
  **********************/
-static void sdl_keyboard_read(lv_indev_t * indev, lv_indev_data_t * data);
+static bool sdl_keyboard_read(lv_indev_t * indev, lv_indev_data_t * data);
 static uint32_t keycode_to_ctrl_key(SDL_Keycode sdl_key);
 static void release_indev_cb(lv_event_t * e);
 
@@ -66,7 +66,7 @@ lv_indev_t * lv_sdl_keyboard_create(void)
  *   STATIC FUNCTIONS
  **********************/
 
-static void sdl_keyboard_read(lv_indev_t * indev, lv_indev_data_t * data)
+static bool sdl_keyboard_read(lv_indev_t * indev, lv_indev_data_t * data)
 {
     lv_sdl_keyboard_t * dev = lv_indev_get_driver_data(indev);
 
@@ -84,6 +84,7 @@ static void sdl_keyboard_read(lv_indev_t * indev, lv_indev_data_t * data)
         data->key = dev->buf[0];
         lv_memmove(dev->buf, dev->buf + 1, len);
     }
+    return false;
 }
 
 static void release_indev_cb(lv_event_t * e)
--- a/src/drivers/sdl/lv_sdl_mouse.c
+++ b/src/drivers/sdl/lv_sdl_mouse.c
@@ -24,7 +24,7 @@
 /**********************
  *  STATIC PROTOTYPES
  **********************/
-static void sdl_mouse_read(lv_indev_t * indev, lv_indev_data_t * data);
+static bool sdl_mouse_read(lv_indev_t * indev, lv_indev_data_t * data);
 static void release_indev_cb(lv_event_t * e);
 
 /**********************
@@ -71,7 +71,7 @@ lv_indev_t * lv_sdl_mouse_create(void)
  *   STATIC FUNCTIONS
  **********************/
 
-static void sdl_mouse_read(lv_indev_t * indev, lv_indev_data_t * data)
+static bool sdl_mouse_read(lv_indev_t * indev, lv_indev_data_t * data)
 {
     lv_sdl_mouse_t * dsc = lv_indev_get_driver_data(indev);
 
@@ -83,6 +83,7 @@ static void sdl_mouse_read(lv_indev_t * indev, lv_indev_data_t * data)
     data->enc_diff = dsc->diff;
     dsc->diff = 0;
 #endif
+    return false;
 }
 
 static void release_indev_cb(lv_event_t * e)
--- a/src/drivers/sdl/lv_sdl_mousewheel.c
+++ b/src/drivers/sdl/lv_sdl_mousewheel.c
@@ -20,7 +20,7 @@
 /**********************
  *  STATIC PROTOTYPES
  **********************/
-static void sdl_mousewheel_read(lv_indev_t * indev, lv_indev_data_t * data);
+static bool sdl_mousewheel_read(lv_indev_t * indev, lv_indev_data_t * data);
 static void release_indev_cb(lv_event_t * e);
 
 /**********************
@@ -62,13 +62,14 @@ lv_indev_t * lv_sdl_mousewheel_create(void)
  *   STATIC FUNCTIONS
  **********************/
 
-static void sdl_mousewheel_read(lv_indev_t * indev, lv_indev_data_t * data)
+static bool sdl_mousewheel_read(lv_indev_t * indev, lv_indev_data_t * data)
 {
     lv_sdl_mousewheel_t * dsc = lv_indev_get_driver_data(indev);
 
     data->state = dsc->state;
     data->enc_diff = dsc->diff;
     dsc->diff = 0;
+    return false;
 }
 
 static void release_indev_cb(lv_event_t * e)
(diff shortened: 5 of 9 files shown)


kisvegabor:
I'm still lost with this feature. It will ensure that changes related to a touch point are displayed immediately. However
- If refreshing is slow (e.g. 100ms) there is no gain as the speed is limited by rendering speed and you will have 10 FPS
- If rendering is fast (e.g. 3 ms) the FPS will be limited by the touch reading speed
  - If reading happens periodically (polling) it will limit the FPS
  - If touch reading is event driver the event fire rate will limit the FPS (can be 100 FPS, but I can't see the value in it)
  
What is the use case the that you are about to optimize?

kisvegabor:
After reading [this comment](https://github.com/lvgl/lvgl/issues/6771#issuecomment-2346239543) I see that your concern is the long running user code in between the 2 calls of `lv_timer_handler`. 

Is it something that really we need to address or
- The user should structure the code to not run so long
- RTOS can be used the manage priorities

kdschlosser:
if the user code is very minimal and lv_task_handler is called often there is still the 33 millisecond timeout that exists for reading touch input that is going to cause a noticeable delay in the scrolling...

kdschlosser:
I did a hackish work around that produces a very similar result. There is a user that tested it and this was their response to it.

https://github.com/lvgl-micropython/lvgl_micropython/issues/128#issuecomment-2367425730

There are no ill effects from using it.

There is a 33 millisecond lag between user input and when LVGL updates the display. How do I know this.. well The display is usually created first and then the indev gets created. These 2 things share the same refresh time of 33 milliseconds. That means then when task_handler gets called the first thing to get done is the display is going to be refreshed, then the indev is going to get called to see if there is any input. task handler will exit and now 33 milliseconds is going to have to pass before the display gets refreshed again... 33 milliseconds is an eternity when dealing with input and extend that out to a total of 33 milliseconds at a minimum from when the display gets touched until the display gets updated is a pretty long time.

What I am addressing is the amount of time it takes for input to get realized on the display from when the actual input occurs. and this is even more important with a single input event that extends over a period of time. Like dragging/scrolling.. 

The latency should not be any more than 10 milliseconds from when input occurs and that input gets realized on the display. Npow I do understand that this may not be 100% possible when the code is running on an MCU.  However, the initial input might have a larger latency but continued user input can have it's latency reduced by a lot.

There is no other way for you to understand what I am talking about unless you actually try it on an MCU when scrolling. Try it without the modification and then try it with and see if you notice a difference.

kisvegabor:
In this case why don't you set the the refresh period to 10 ms?

kdschlosser:
Because the refresh timer for the display uses the same timer. There is also no need to check the indev at 10 milliseconds. it only needs the rapid check if there is touch input so it is able to respond fast when scrolling or releasing a button.

kdschlosser:
it cannot be handled in the read CB because the touch has not been processed by LVGL yet. refreshing the display from inside of the read CB isn't going to update the display based on the touch input that has just occurred.


how I have worked around it is I put the driver into event mode and created a timer to call a function that calls the read function for the indev which in turn calls the indev callback. in the indev callback I have it store the touch information so that once the read code exists and it goes back to the timer callback then I have it check the stored information to see if touch occurred or not and if it has then update the display.

it's a hackish workaround that consumes more resources.

There is no easy way of adding this type of feature at the user level due to all of the indev processing functions not being public.

kisvegabor:
I see the problem, but I feel like the proposed solution is a little bit hacky. I mean to solve that the "app handler" might run too long and there are not enough resources for the UI is a higher level problem (task scheduling). As we are not writing a schedule, I think it shouldn't be handled in LVGL. 

For example LVGL gives you the opportunity to not call the "app handler" if the touchpad is pressed.

kdschlosser:
My point being is there is no way to do it from user code without having to reinvent the wheel. That's because all of the functions that process input are not public....

so when the indev callback gets called it gets called from a non public function that collects the touch data, only after the callback returned does the data get handed off to varying functions to be processed and those processing functions are once again not public. When I say not public I mean they are literally not in any header files. they are only declared/defined in C source files.

If the processing functions were public then it could be done from inside of the indev callback. Inside the callback I could collect the input and call the appropriate processing function and then tell LVGL to refresh the display. I would populate the data structure as if no input has occurred. So the indev mechanics would not process the input once the callback returned because it has already been processed.. 

See what I am saying?

There are also issues with the timer value for the indev being the same things as the refresh time for the display. These should be able to be set separately from each other which they currently are not able to be.

Here is a use case example where this really hits home, The MicroPython Binding.  It only runs on a single core and it doesn't expose FreeRTOS at all if it is even being used on a specific port. No real way to handle scheduling tasks. Everything runs in a single loop. You would have to add a large amount of additional code which is going to take up space and memory on a device that doesn't have a lot of either of those things to do something that would literally be the addition of 2-3 lines of code in LVGL which would take up next to no additional flash and no additional heap memory in order for it to work.

kdschlosser:
There is a need for this kind of ability.

See [HERE](https://forum.lvgl.io/t/gestures-are-slow-perceiving-only-detecting-one-of-5-10-tries/18515/4)

kisvegabor:
To put it simple the proposed functionality is to not exit lv_timet_handler while the touch is pressed. If it's true could something like this be suitable too?

```c
lv_timet_handler();
if(!pressed()) user_code();
```

kdschlosser:
No it couldn't. The reason why is because once the call to `lv_timer_handler` exists then it is going to be 33 milliseconds that has to pass before the timer for the indev will expire to read the touch input. You end up with 33.. That add a lot of lag in the user input. and you also end up with the MCU doing nothing for 33 milliseconds. It's spinning it's wheels... Ever seen a [bullshit grinder](https://www.youtube.com/watch?v=oCaOzV7LCLs)??   When sliding a finger across the display ideally you want to be reading the touch and updating the display as fast as possible. This is what gives that quality feel when using a device that has user input and a GUI.  The touch she be reading as fast as the touch IC is able to read user input. most touch IC's have a refresh in the single digit milliseconds.. 

Ever bought an electronic device that had some kind of a display with some form of user input like a touch screen and you press a button and then wait for the GUI to act on that button press. That is ALWAYS the first thing a user is going to notice. It is what tells the user if the design of the device is quality or not. User input almost always comes before anything else with GUI applications.

LVGL will work as normal when there is no input at all. It only opens the gates when there is input. You don't want to make the default timeout smaller because then it is going to be adding more work when it really doesn't need to be done. You only want to open the gates when there is input. This is something that should be able to be done at runtime and not compile time so it can be turned on and off programmatically maybe based on CPU load. Or maybe the user input has trigger a large tack that needs to be performed so the read loop needs to be interrupted regardless of user input.

By adding a return type to the callback that would instruct the calling code to refresh the display. By setting the continue_reading marker in the data structure would tell the call code to call the callback again after processing the last input. You return True and setting continue_reading to True and you end up with a UI that is responsive to user input and that additional work needed to do that only occurs when there is touch input that has been received. as soon as there is no input the user sets continue_reading to false and returns True only the one time after the touch is no longer detected. all other times touch is not detected false would be returned from the callback so the display will not get refreshed. That process gets reset once touch has been detected. That portion the user can manage as it is just a matter of storing the last touch state to check it. Easy thing to do. 

It requires a lot more work to implement this kind of a feature externally to LVGL because the input processing functions are not declared in any header file, not even the private ones. So there is zero access to them to be able to read the input and then process the input and refresh the display. 

You are talking literally 4 lines of code added and 4 lines that are changed. That is all that is being done to make this work. The rest of the code changes were made in some of the built in drivers. That is not a requirement to do that in order for it to be available to the user. I only changed those to show you what would need to be done to use the feature.

kisvegabor:
I'm pretty sure that the PC doesn't attempt to render 5080 times as you move the mouse. Instead it records all the "states" and let's say in every 16 ms (60Hz) is processes them and show the result. (Still having 300 states for each frame)

kdschlosser:
actually Windows as a default reads it once a millisecond and that value can be adjusted. Gaming mice for example are read at a much faster rate than the default.

I mean think about it. a mouse reads it's position in "dpi" and the number of DPI is able to be set a lot of the time. I have my mouse set to 3000 DPI. how fast would I be able to move my mouse 1 inch? pretty damned quick. if the speed that the data is read from the mouse is not fast enough the mouse would be moving really funky because windows would be trying to set how fast the mouse has moved based on the number of pixels it has moved as reported by the mouse. so in order to have the mouse move properly the rate that the mouse has it's data read needs to be pretty damned high.

kdschlosser:
search for M65 RGB ULTRA polling rate. That mouse is able to report at 8000hz or 8000 times a second.

kdschlosser:
There are a bunch of mice that also poll at 10000Hz.

kisvegabor:
I see, but rendering still happens at 60Hz or 120 Hz. The data is buffered and processed 60 or 120 times in a second.

kdschlosser:
where does the 60 and 120 come from? I am assuming what you mean by that is the display frequency. That is not to be confused with rendering frequency as the display update frequency is simply the number of times the display adapter writes the data in GRAM to the display. How fast the data in gram is updated is only limited by the processing power of the cpu and gpu.

kissa96:
Yes, this is true. RAM can be updated much faster than flushing it to the display. You can actually observe it in some laggy cases, where video memory is properly updated, but the display itself is slow. Kind of a reverse corner case, but it is possible. Gfx cards have inbuilt display driver circuitry, which can be configured to the desired FPS and screen size. It automatically flushes the VRAM contents paralel to whatever the gpu is doing.

kisvegabor:
Okay, so these are seemingly independent:
1. How fast you render? An empty screen can be rendered with 1000 FPS too
2. How fast the rendered content is synced with the display? Typically it's 30, 60 or 120 Hz.

But why would you render frames which won't be displayed? Therefore the rendering time is usually synced with display refresh period.

kdschlosser:
in the world of LVGL when you are touching the display something is almost always changing on the display. This is the reason why you want to increase the frequency in which the display is being updated. You also have to remember that when running on an MCU you are not going to be able to update the display as fast as a PC. so what you are thinking as reading the touch min a continuous loop would be polling the device so fast that changes would not take place you are forgetting the amount of time it is going to take LVGL to render information and get it written to the display. so the loop is not going to be running at full tilt and a lot of times there is going to be a 3-5 millisecond period of time that passes between the actual reads and in that time if there is continued touch there is going to be a change that is taking place. whether it be releasing the display or wanting to scroll or holding the button down there is going to be something that is collected that LVGL would be able to be updated in some manner to that input.

With PC's having so many cores they have a grip of cores that interrupts can take place on so dedicating a core to handle interrupts from input is not going to noticeably decrease system performance at all. the input is able to get processed much faster. whether or not the OS actually does anything with that input is a different story but on an MCU the input is not going to be getting read at the speeds that is seen on the OS and a change is likely going to exist with continued input. If a user has some code that MUST run an interrupt can be used as a way to stop an input polling loop if one is running for some reason. A simple change to a flag to trigger the loop to exit is all that would need to be done.

kisvegabor:
I still don't get it. To put is simple: What's the point of having more often redraws than the screen's update period  (i.e. VSYNC).

kdschlosser:
It's a gaming thing I guess. IDK...

All I can tell you is that being able to tell LVGL to refresh the display and reading the indev in a contentious loop if the state is set to pressed does improve it's responsiveness to user input and it also makes scrolling run smoother as well.

kisvegabor:
How does the result feel if you set `LV_DEF_REFR_PERIOD` to `3`?

kdschlosser:
You don't want to change the default refresh to 3 because that is going to cause a lot of extra work to be done when it may not need to be done. Right now the timer gets checked and if it's not ready the code moves on. If you set it to 3 then every 3 milliseconds the code is going to do in and see if there is anything that needs updating. To do that I believe that each and every single object gets looked at.

IF the mechanics of LVGL set a flag at the screen level or the display level if something has changed in any object and in the task handler that flag could be quickly looked at and move on if not set and if it was set then it could do the update... That would be better because you could eliminate the timer for the display refresh all together as it would update the display when something has changed. To make this work the best you would want to have a pointer to the display stored in the object. for fast access to set the flag instead of having to iterate down the parent tree to locate the display to set the flag. The flag it self would be a bit field in the display structure to minimize the size it takes up. You would however have the 4 byte expenditure per object for the pointer to the display structure. It might be worth the tradeoff  considering how much memory a timer might take up. 

You can tell me if I am wrong on how the mechanics works. But when something happen to an object, the size changes, or a color changes maybe the value changes a flag get set in that object that it is invalid. when task handler is called and the timout has expired for the refresh then all of the objects gets iterated over for that display checking to see if anything is marked as invalid. if something is invalid then rendering takes place and the display gets updated.

It would be better performance wise to iterate to locate the display at object creation and set a field in the object's structure to point to that display. There is still a flag in the object to mark that it needs to be updated and at the same time the display gets marked as needing an update. This way when task handler is called on a display the check to see if there is any updating that needs to take place is a simple check taking no time to do. hence the need to have a refresh timer. This check would get done every single time task handler was called. Because the test is simple the amount of work that needs to be performed is no longer an issue which is the reason why a timer was used in the first place.

kdschlosser:
But we are talking about input here and that does need to have a timer to control how often the polling takes place. and due to the amount of work the polling has the better mechanics is to speed up the polling when there is input and slow it down when there is no input. it's simply a dynamic control for the polling interval for input and refreshing the display when there is input. When there is input there is almost always going to be something that needs to be updated on the display and having to wait that 33 milliseconds for the input to actually be realized on the display is a pretty bad response time.

LVGL is used on a lot of small display type devices, like watches and things like that. It takes no time to send the data to those kinds of displays even when running SPI. A lot of times DMA memory is able to be used coupled with double buffering to further increase the performance. The thing that hurts the performance at that point is going to be that 33 milliseconds but you don't want to do work when there is nothing that needs to be updated. iterating over all of the objects to check to see if something is invalid ends up being wasted resources when there is nothing to update and that time could be spent elsewhere, maybe checking for user input, or reading GPIOs for se3nsor data.

kdschlosser:
I cannot imagine it would be too hard to add the code in to LVGL to set a flag at the display level to see if there are updates needed and to have an object set that flag when something has changed. Then remove the timer for the display refresh and have the task handler look for that marker in the display instead. I am willing to bet you see a pretty good jump in the UI's responsiveness. Then if you couple that with the change to the indev polling the UI would become exceedingly snappy to use. Animations would run smoother and user input would be realized almost instantly.

kisvegabor:
This feature is really simple to add. My concern is that we are solving an issue with it that (I think) we shouldn't solve in LVGL. And it will be a bad precedence for setting the what LVGL is for. 

Still brainstorming: `lv_timer_set_period(lv_display_get_refr_timer(NULL), 3);` while the touch is pressed and `lv_timer_set_period(lv_display_get_refr_timer(NULL), 33);` when it's released?

kdschlosser:
There is no way to call lvgl_refr_now if there is touch input and after the touch has been processed by LVGL. All of the processing functions for input are private. changing the timer still means the task_handler still has to exit allowing user code to run which could be a large ? as to how long it is going to take to run.

Currently in order for the input to be realized by LVGL and actually displayed on the screen it consists of...

```
lv_task_handler -> indev callback - > process input -> exit task handler -> run user code -> lv_task handler -> render -> send to display
```

That is an obnoxious amount of things that have to be done just for input to be realized on the display.

When it could be

```
lv_task_handler -> indev callback -> process input -> render -> send to display -> indev callback -> process input -> render -> .....
```

It is also something that the use is optional. It doesn't increase the resulting binary buy maybe a couple of bytes  and there is an extremely low about of memory use and it's all on the stack and not on the heap, It's not something that really needs to be maintained so to speak and most of the mechanics are already in place to use it. It's a matter of adding a return value from the callback function and if that return value is true then call lvgl_refr_now. It's that simple.

kisvegabor:
Sorry, but I'm still not convinced. 

We can use `lv_display_get_inactive_time()` like this:

```c
if (lv_display_get_inactive_time(NULL) < 100) { /*Use action happened in 100 ms*/
   faster_disp_refr();
} else {
   slower_disp_refr();
   user_code();
}
```

kisvegabor:
Giving up? :smile:

kdschlosser:
I cannot come up with any other way to explain it to you so you can understand the impact it makes in how responsive to input LVGL becomes. Everything runs smoother, from animations to scrolling. anything that is triggered as a result of user input. and once there is no user input it goes back to the normal loop with the normal timers and timeouts. No setting need to be changed to alter the behavior and no timers need to be fiddled with to make it all work. It's super simple to add to LVGL but not so simple to do in user code because of the functions needed being private.

Have you even tried making the code changes to the indev timer callback and implemented them on an actual MCU with an attached display? If you did that you would see what I am talking about. That is the ONLY way you are going to understand. I have exhausted my vocabulary on how to describe it.

So I guess yes I am giving up. There have been a lot of keys pressed since I originally submitted this PR and all my efforts to explain it have not given you a good enough mental picture of the hows and whys. The only thing I can tell you is to add the code to LVGL and write a simple scrolling object and run it on an MCU.

kisvegabor:
I really haven't tried it, but I have conceptual issues with this feature. Anyway, thank you for the effort. Hopefully one day we will get back to this and find a common ground.

PR #6854, opened 2024-09-14
fix(libpng): fix memory leak when draw buf alloc failed

lion2tomato:
fix(libpng): fix memory leak when draw buf alloc failed
fix(imagecache): add cache evict when cache memory is not enough


src/drivers/nuttx/lv_nuttx_image_cache.c
```diff
@@ -60,7 +64,7 @@ static void free_cb(void * draw_buf);
  *   GLOBAL FUNCTIONS
  **********************/
 
-void lv_nuttx_image_cache_init(void)
+void lv_nuttx_image_cache_init(bool use_independent_image_heap)
```
FASTSHIFT, line 67: Please send a separate PR for nuttx image cache modification.

FASTSHIFT (changes requested):

src/drivers/nuttx/lv_nuttx_image_cache.h
```diff
@@ -15,6 +15,7 @@ extern "C" {
  *********************/
 
 #include "../../lv_conf_internal.h"
+#include <stdbool.h>
```
kisvegabor, line 18:
```suggestion
#include LV_STDBOOL_INCLUDE
```

Issue #6687, opened 2024-08-15
lv_obj_add_flag and remove_flag should check if the flag is already set/unset


lorddevereux:
### Introduce the problem

I recently ran into a case where a rogue function was continually calling **lv_obj_remove_flag(widget, LV_OBJ_FLAG_HIDDEN)** on a few widgets.
Every call causes the object to be invalidated and therefore redrawn even though it hasn't actually changed. This caused a massive performance impact (3 FPS -> 30 FPS!). Whilst it should be up to the user to make sure they don't call the function unnecessarily, I think patching to prevent this will save someone debugging in future.

Posting as a discussion before submitting PR in case someone think of a good reason not to do it!

### Proposal

Add to lv_obj_add_flag ~line 233
`if (obj->flags & f)`
`        return;`

Add to lv_obj_remove_flag ~line 270
`if (!(obj->flags & f))`
`        return;`

kisvegabor:
Great catch!  Could you send a Pull request?

Just a minor change for `add` as it should work with ORed flags too (e.g. add `HIDDEN | IGNORE_LAYOUT`):
```c
if ((obj->flags & f) == f) /*Check if all flags are set*/
        return;
```

kisvegabor:
No stale

jurstu:
I believe I made the proposed changes. PR coming.

PR #6859, opened 2024-09-15
perf(obj): skip repeated flag setting operations

jurstu:
issue link:
https://github.com/lvgl/lvgl/issues/6687

### DIFF
--- a/src/core/lv_obj.c
+++ b/src/core/lv_obj.c
@@ -231,7 +231,7 @@ lv_obj_t * lv_obj_create(lv_obj_t * parent)
 void lv_obj_add_flag(lv_obj_t * obj, lv_obj_flag_t f)
 {
     LV_ASSERT_OBJ(obj, MY_CLASS);
-    if((obj->flags & f) == f) /*Check if all flags are set*/
+    if(lv_obj_has_flag(obj, f)) /*Check if all flags are set*/
         return;
 
     bool was_on_layout = lv_obj_is_layout_positioned(obj);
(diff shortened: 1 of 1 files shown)


jurstu:
issue link:
https://github.com/lvgl/lvgl/issues/6687

src/core/lv_obj.c
```diff
@@ -231,6 +231,8 @@ lv_obj_t * lv_obj_create(lv_obj_t * parent)
 void lv_obj_add_flag(lv_obj_t * obj, lv_obj_flag_t f)
 {
     LV_ASSERT_OBJ(obj, MY_CLASS);
+    if((obj->flags & f) == f) /*Check if all flags are set*/
```
FASTSHIFT: Use `lv_obj_has_flag`

src/core/lv_obj.c
```diff
@@ -268,6 +270,8 @@ void lv_obj_add_flag(lv_obj_t * obj, lv_obj_flag_t f)
 void lv_obj_remove_flag(lv_obj_t * obj, lv_obj_flag_t f)
 {
     LV_ASSERT_OBJ(obj, MY_CLASS);
+    if(!(obj->flags & f))
```
FASTSHIFT: `!lv_obj_has_flag(obj, f)`

FASTSHIFT (changes requested):

src/core/lv_obj.c
jurstu: This is a great idea! lv_obj_has_flag function does exactly that!

src/core/lv_obj.c
jurstu: looking into it..

jurstu: This will for sure work for single, un-OR-ed flags.

image

For OR-ed flags that would suggest clearing 2 state bits:
lv_obj_has_flag will return TRUE (causing modification and not return from function) only if both state bits are set. 

This means using this function is logically wrong in this case, since we would like to modify the state if ANY of the flags are set. 

Using lv_obj_has_flag_any:

if(!(lv_obj_has_flag_any(obj, f)))

FASTSHIFT (approved):

kisvegabor (approved):
Nice, thank you!

PR #6861, opened 2024-09-15
feat(opengl): texture caching

### LINKED ISSUE #6017: Implement an OpenGL ES based renderer
### Problem to solve

Similar to the [SDL based render](https://github.com/lvgl/lvgl/issues/6016) we can have  OpenGL ES based renderer too. The concept is the same as in the case of SDL.

### Success criteria

Have a fully functional OpenGL ES based renderer.

### Solution outline

To be determined.

### Rabbit holes

To be determined.

### Testing

It would be great to add unit tests to verify the output, but it's not clear if we can do it in GitHub CI, and if we can get back the result from the OoenGL Textures.

### Teaching

Docs needs to be added.

### Considerations

_No response_

liamHowatt:
Fixes #6017

Similar to LVGL's SDL texture caching. Reuse draw buffers by creating OpenGL textures for them and caching them. In theory, it should be faster because they do not have to be SW rendered again and they should blend faster than they would by software.

### Notes


liamHowatt:
To make reviewing simpler, I recommend viewing lv_draw_opengles.c as a diff of lv_draw_sdl.c

Diff with git diff
```
sed -e "s/sdl/opengles/g" -e "s/SDL/OPENGLES/g" src/draw/sdl/lv_draw_sdl.c | git diff --no-index - src/draw/opengles/lv_draw_opengles.c
```

Diff in VS Code
```
sed -e "s/sdl/opengles/g" -e "s/SDL/OPENGLES/g" src/draw/sdl/lv_draw_sdl.c | code --diff -- - src/draw/opengles/lv_draw_opengles.c
```

Todo:
- Edges of radii appear clipped
- Use OpenGL to draw fills. Do not cache fills as textures.

liamHowatt:
## My benchmark average CPU usage results

Before this PR: 48%
Fills from cached textures: 36%
Fills with shader: 31%
Reuse framebuffer: 28%

Despite better performance overall, some scenes perform worse. Those scenes are:
- moving wallpaper: 2% slower
- multiple labels: 43% slower
- multiple arcs: 10% slower
- widgets demo: 2% slower

It does not improve by increasing the max cached texture count from 128 to 1024.

## Known issues

### TODO, unimplemented

- Rotated layers

### Graphical, unknown cause

Borders
borders_large

Radii. The last row of pixels on each edge seems to be missing.
radii_large
radii2_large

Layer opa - the first image is how it looks in this PR, the second image is how it's supposed to look.
layers image

kisvegabor:
The gray pixel artifact seems like black is blended with the opacity of the pixel (instead of the actual color)

kisvegabor:
It's working, and I've created some benchmarks. It's faster in most of the cases but when a lot of recaching is required it's slower (e.g. in case of arc and multi texts). I don't know why the widget's demo is so much slower. Increasing the cache size 1000 (from 128) changed the render time to 11ms from 16ms. (SW is 8 ms)

image
image

src/draw/opengles/lv_draw_opengles.c
```diff
@@ -0,0 +1,666 @@
...
+#elif LV_COLOR_DEPTH == 24
+    GL_CALL(glTexImage2D(GL_TEXTURE_2D, 0, GL_RGB, texture_w, texture_h, 0, GL_BGR, GL_UNSIGNED_BYTE, opengles_render_buf));
+#elif LV_COLOR_DEPTH == 32
+    GL_CALL(glTexImage2D(GL_TEXTURE_2D, 0, GL_RGBA, texture_w, texture_h, 0, GL_BGRA, GL_UNSIGNED_BYTE,
+                         opengles_render_buf));
+#else
+#error("Unsupported color format")
+#endif
+
+    glGenerateMipmap(GL_TEXTURE_2D);
```
JWBverheesen: I usually follow this call with:
glTexParameteri(GL_TEXTURE_2D, GL_TEXTURE_MAX_LEVEL _mipMapLevel)
glTexParameteri(GL_TEXTURE_2D, GL_TEXTURE_MIN_FILTER, GL_LINEAR_MIPMAP_LINEAR)
glTexParameteri(GL_TEXTURE_2D, GL_TEXTURE_MAG_FILTER, GL_LINEAR)
Or nearest when chosing not to use interpolation.

The first call to lower the default index value of 1000 mipmaplevels to a value more in the range of 1-12. 
The latter might apply more to 3d space where GL_LINEAR_MIPMAP_LINEAR gives the smoothest transition between mipmaps but I use it for 2D purposes as well.

JWBverheesen:
I don't have much time on my hand to look into this at the moment. I can take a deeper look somewhere later this week. Also on the performance difference mentioned above. That does strike me as quite strange. When all objects (and its vertices / indices) are set, rendering should be very fast. In my own project this flow goes like:
- load texture, similar to what is done in this codebase
- Update vertices 
- Create GL attributes, buffers and vertex arrays
Then when no changes to the position, scaling or orientation of the object happens the same process of:
when applicable, Bind 
- Use shader ( I load multiple shader programs, map them to an ID and call it with glUseProgram(_id))
- DrawElements
Also make sure that any change to the texture/ vertex/ object happens in a different thread than the main rendering loop. 
I usually just keep everything pre-rendered and render things directly to the screen versus using framebuffers to render to a texture.

src/drivers/glfw/lv_glfw_window.c
```diff
@@ -300,7 +300,10 @@ static void window_update_handler(lv_timer_t * t)
                 lv_refr_now(texture_disp);
             }
 
-            lv_opengles_render_texture(texture->texture_id, &texture->area, texture->opa, window->hor_res, window->ver_res);
+            lv_area_t clip_area = texture->area;
+            lv_area_move(&clip_area, -texture->area.x1, -texture->area.y1);
+            lv_opengles_render_texture(texture->texture_id, &texture->area, texture->opa, window->hor_res, window->ver_res,
+                                       &clip_area, false);
```
JWBverheesen: When setting LV_USE_DRAW_OPENGLES to 0, so not using the draw_opengles src, this will now flip the window. Setting the flip field to true resolves this. This either needs an #ifdef LV_USE_DRAW_OPENGLES or some other solution in the opengles.c file.

JWBverheesen: I also get some weird result when running the widget demo:
Screenshot at 2024-09-17 15-16-04

kisvegabor:
> When all objects (and its vertices / indices) are set, rendering should be very fast. In my own project this flow goes like:

Note that OpenGL was slower only in the cases when it couldn't reuse the cached textures e.g. because the  color of the labels in the "Multiple labels" text are changing randomly. 

Other than that the rendering times as usually 0ms.

src/drivers/glfw/lv_glfw_window.c
kisvegabor: Did you set the color depth to 32?

JWBverheesen: Thanks for pointing that one out! fixed now

src/draw/opengles/lv_draw_opengles.c
```diff
@@ -0,0 +1,666 @@
...
+    lv_draw_buf_t draw_buf;
+    dest_layer.draw_buf = &draw_buf;
+    lv_draw_buf_init(dest_layer.draw_buf, texture_w, texture_h,
+                     LV_COLOR_FORMAT_ARGB8888, LV_STRIDE_AUTO, opengles_render_buf, sizeof(opengles_render_buf));
+    dest_layer.color_format = LV_COLOR_FORMAT_ARGB8888;
+
+    dest_layer.buf_area = task->_real_area;
+    dest_layer._clip_area = task->_real_area;
+    dest_layer.phy_clip_area = task->_real_area;
+    lv_memzero(opengles_render_buf, lv_area_get_size(&dest_layer.buf_area) * 4 + 100);
```
JWBverheesen: magic number +100 bytes allocated for the render buffer?

kisvegabor: I think it's coming from the SDL renderer. Probably not required, just a leftover.

src/draw/opengles/lv_draw_opengles.c
liamHowatt: Thanks! 1000 is a lot of mipmap levels, especially considering the mipmap is unused since the textures are currently never being scaled relative to the pixels. I had been wondering if we could avoid generating the mipmap. I recall trying to remove it and something broke.

We could probably accelerate LVGL object/image transformations with OpenGL. That can be a future PR. I imagine the mipmap will be important to have for that. In the meantime I'll apply your suggestion to be well-prepared for forward compatibility.

src/drivers/glfw/lv_glfw_window.c
liamHowatt: > When setting LV_USE_DRAW_OPENGLES to 0, so not using the draw_opengles src, this will now flip the window. Setting the flip field to true resolves this. This either needs an #ifdef LV_USE_DRAW_OPENGLES or some other solution in the opengles.c file.

Good catch! I will fix it.

src/draw/opengles/lv_draw_opengles.c
liamHowatt: Correct, it came from the SDL renderer. I was trying to leave working things untouched until the end.

src/draw/opengles/lv_draw_opengles.c
JWBverheesen: I tried linear but that gave some bleeding between labels and background due to interpolation. You want this in 3D space to resolve some artifacts in depth but it doesn't look as good in 2D. Setting To Nearest is better in this case.
glTexParameteri(GL_TEXTURE_2D, GL_TEXTURE_MAX_LEVEL, 20);
glTexParameteri(GL_TEXTURE_2D, GL_TEXTURE_MIN_FILTER, GL_NEAREST_MIPMAP_NEAREST);
glTexParameteri(GL_TEXTURE_2D, GL_TEXTURE_MAG_FILTER, GL_NEAREST);

src/drivers/glfw/lv_opengles_driver.c
```diff
@@ -100,10 +109,17 @@ static const char * fragment_shader =
     "uniform sampler2D u_Texture;\n"
     "uniform int u_ColorDepth;\n"
     "uniform float u_Opa;\n"
+    "uniform bool u_IsFill;\n"
+    "uniform vec3 u_FillColor;\n"
     "\n"
     "void main()\n"
```
JWBverheesen, line 113: I went over the shader as well and see that precision is set to mediump. That is fine for texture coordinates, but the use only seems to be limited to colors. you can use lowp. In large scenes using framents shaders with lowp, midiump or higp can have some influence on performance (but probably not really needed in this case)

liamHowatt:
@JWBverheesen all of your feedback has been addressed. Thank you.

- Added docs
- I fixed the "Graphical, unknown cause" issue for 99% of cases. See the new docs.
- It works for 24 and 16 now. 8 having some issues.
- Unfortunately I couldn't figure out exactly why the texture sometimes needs to be flipped and sometimes doesn't.

I was about to take it out of draft but I see there is now clipping happening with the [advanced usage](https://docs.lvgl.io/master/integration/driver/opengles.html#advanced-usage). I will fix that, rebase, and then take it out of draft.

JWBverheesen:
Looks good from what I can see right now! as to the textures sometimes being flipped:

Most image formats store there image data starting at top-left corner of the image. OpenGL starts bottom-left corner as origin. glTexCoord2f(0,0) = bottom-left and glTexCoord2f(1,1) = top-right corner. Kinda anoying but someting to keep in mind.

JWBverheesen (approved):

liamHowatt:
A known issue I'm going to leave as-is for now is that there is flickering when [advanced usage](https://docs.lvgl.io/master/integration/driver/opengles.html#advanced-usage) is used while `LV_USE_DRAW_OPENGLES` is enabled.

kisvegabor (dismissed):
LGTM! Thank you Liam!

@XuNeo @FASTSHIFT could you also take look at it?

FASTSHIFT (approved):

kisvegabor (approved):

PR #6863, opened 2024-09-16
fix: added bounds check in gif decoder

Alpeev:
Added bound check in the gif header  parsing functions to avoid integer overflow with subsequent allocation of wrongly-sized-buffer;
Also, added checks to avoid buffer overruns in the image parsing functions.

### DIFF
--- a/src/libs/gif/gifdec.c
+++ b/src/libs/gif/gifdec.c
@@ -113,9 +113,17 @@ static gd_GIF * gif_open(gd_GIF * gif_base)
     /* Aspect Ratio */
     f_gif_read(gif_base, &aspect, 1);
     /* Create gd_GIF Structure. */
-#if LV_GIF_CACHE_DECODE_DATA
+    #if LV_GIF_CACHE_DECODE_DATA
+    if(0 == (INT_MAX - sizeof(gd_GIF) - LZW_CACHE_SIZE) / width / height / 5){
+        LV_LOG_WARN("Image dimensions are too large");
+        goto fail;
+    } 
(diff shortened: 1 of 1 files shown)


src/libs/gif/gifdec.c
```diff
@@ -113,9 +113,21 @@ static gd_GIF * gif_open(gd_GIF * gif_base)
     /* Aspect Ratio */
     f_gif_read(gif_base, &aspect, 1);
     /* Create gd_GIF Structure. */
-#if LV_GIF_CACHE_DECODE_DATA
+    if(0 == width || 0 == height){
+        LV_LOG_WARN("Zero size image");
+        goto fail;
+    }
+    #if LV_GIF_CACHE_DECODE_DATA
```
kisvegabor:
```suggestion
#if LV_GIF_CACHE_DECODE_DATA
```

kisvegabor (commented):
Wao, thank you! To be honest, we haven't focused on these kind of issues so far. It's great that now you made the first step! :rocket:

Alpeev:
Thanks!

Alpeev:
Are there any target archs with 16 bit ints?

Alpeev:
Modified so as to avoid problems in arches with 16 bit ints. 
In the long run though, it is a good idea to change all counter varibles either to size_t or fixed-size integers.

kisvegabor (approved):

kisvegabor:
> In the long run though, it is a good idea to change all counter varibles either to size_t or fixed-size integers.

In LVGL we are doing that, however we integrate 3rd party libraries with as minimal modifications as possible.

liamHowatt (approved):

PR #6864, opened 2024-09-17
fix(image): lv_image_set_inner_align() behaviour with LV_IMAGE_ALIGN_STRETCH

rdegliesposti:
If you try to set the inner alignment of an image with `lv_image_set_inner_align` to `LV_IMAGE_ALIGN_STRETCH` when an image source for the widget is not yet set, lvgl will attempt a division by zero which results in a crash.


rdegliesposti:
`update_align` should also reset the scale of the image once the alignment has been changed to something else

kisvegabor (approved):
Both make sense, thank you!

FASTSHIFT (approved):

liamHowatt (approved):

Issue #6865, opened 2024-09-17
lv_obj_add_style () may crash appl


glory-man:
### LVGL version

v8.3, v8.4

### What happened?

There is no checking of limits for
[`obj->style_cnt++;](https://github.com/lvgl/lvgl/blob/03e01bdd3e552da632a9b5e853d2efa3540c02d9/src/core/lv_obj_style.c#L91)
It may overflow 63 limit and become zero value

Also if you add the same style multiple times it will be added many times. Unlike version 7 where the same style could only be added once because the function `lv_obj_add_style()` in v7 uses a `_lv_style_list_remove_style()`  call to prevent multiple insertion of the same style

glory-man:
also I interested what will return `lv_mem_realloc()` couldn't have enough memory to allocate increased styles buffer, or how will work `lv_obj_add_style()` if `lv_mem_realloc()` returned NULL

kisvegabor:
In v9 we already have `LV_ASSERT(obj->style_cnt < 63);` and `    lv_obj_remove_style(obj, style, selector);` to address these. I think the same can be backported to v8. 


Regarding `lv_mem_realloc`, it returns `NULL` on out of memory. It's not ideal for sure, however the approach small memory allocations like this is that if you are out of memory here it's hard to recover and the system will crash anyway.

glory-man:
> Regarding `lv_mem_realloc`, it returns `NULL` on out of memory. It's not ideal for sure, however the approach small memory allocations like this is that if you are out of memory here it's hard to recover and the system will crash anyway.

maybe it makes sense to at least add `LV_ASSERT_MALLOC(obj->styles);` as this realized for labels

kisvegabor:
https://github.com/lvgl/lvgl/pull/6873 is merged, closing. Thank you!

PR #6873, opened 2024-09-18
feat(style): add style_cnt and styles checks

glory-man:
added ASSERT to control style_cnt limits & try to remove style if it already added during adding styles.
Fixes https://github.com/lvgl/lvgl/issues/6865


kisvegabor (approved):
Thank you!

liamHowatt (approved):
Thanks!

I know we're just syncing v8.4 with `master` so it all looks good for now, but why is the limit 63 and not 64? Also, could the `ASSERT` come after the `lv_obj_remove_style` since the remove might open up a new spot?

PR #6888, opened 2024-09-19
fix(fragment): fix NULL pointer access

stachurr:
_Ready_

If called with a lv_fragment_manager_t which has no attached lv_fragment_t's, the LV_LL_READ_BACK() macro will set the pointer to null. This was immediately followed by a dereference.

Now a warning is logged followed by a return of false.

### DIFF
--- a/src/others/fragment/lv_fragment_manager.c
+++ b/src/others/fragment/lv_fragment_manager.c
@@ -188,6 +188,10 @@ bool lv_fragment_manager_send_event(lv_fragment_manager_t * manager, int code, v
     LV_ASSERT_NULL(manager);
     lv_fragment_managed_states_t * p = NULL;
     LV_LL_READ_BACK(&manager->attached, p) {
+        if(!p) {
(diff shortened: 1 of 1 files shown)


kisvegabor (approved):

liamHowatt (commented):
Hi, I don't think the loop will ever be entered if `p` is `NULL`.

`LV_LL_READ_BACK` is a macro defined as this:
```c
#define LV_LL_READ_BACK(list, i) for(i = lv_ll_get_tail(list); i != NULL; i = lv_ll_get_prev(list, i))
```

Is this a problem that you encountered while testing or is it just speculation?

stachurr:
Hi, this was a problem I actively encountered.

kisvegabor:
Can you test with a debugger how is can be `NULL`?

stachurr:
I can't seem to replicate the issue. I must have been doing something wildly wrong to cause that before. I have changed a lot since then. Sorry for wasting your time.

PR #6906, opened 2024-09-22
 adds script to create lvgl_private.h

kdschlosser:
I also added in a test to check the validity of the includes in the event a private header is removed or added.
 I also updated lvgl_private.h because it was missing 5 private header files.


.github/workflows/verify_private_header.yml
```diff
@@ -0,0 +1,22 @@
...
+      - name: Checkout
+        uses: actions/checkout@v4
+        with:
+          persist-credentials: false
+          fetch-depth: 0
+      - name: Setup Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: 3.7
+      - name: Verify lvgl_private.h
```
kisvegabor, line 20: How will it make the CI fail if `lvgl_private.h` was updated? See the code formatter as a [reference](https://github.com/lvgl/lvgl/blob/master/.github/workflows/check_style.yml#L31)

kdschlosser: if a private header is added and the lvgl_private.h file doesn't have the new header file added or of a private header file is removed but it is not removed from lvgl_private.h

as an example. the main branch has 5 private headers that are not in the lvgl_private.h file. it is being overlooked to add/remove from that file.

kisvegabor: I see, but I can't see anything in this CI action which check for a return value and tells job to fail. It runs, modifies some files and returns.

kdschlosser: the action runs a python script and if you look here you can see it raises an error if the existing private header file doesn't have the same contents as what the new header file would have...

https://github.com/lvgl/lvgl/blob/d27f044a183cc8eb8127ba5d08cd156bb9a5ab44/scripts/generate_private_header.py#L84

kdschlosser: It doesn't have to specifically check for a failure. If anything that runs returns a non zero it automatically will trigger a build failure.

src/lvgl_private.h
```diff
@@ -2,6 +2,12 @@
  * @file lvgl_private.h
  *
  */
+ 
+/*
+ *                            DO NOT EDIT
```
kisvegabor, line 7:
```suggestion
 *  DO NOT EDIT
```

.github/workflows/verify_private_header.yml
kisvegabor: Got it, thank you!

src/lvgl_private.h
kdschlosser: I didn't want to fix that if you are not going to use the PR. if you are then I will spend the time to fix it. 😛

kisvegabor: I'm ok with adding this PR :slightly_smiling_face:

kdschlosser: OK then I will fix the formatting error.

liamHowatt (commented):
Careful, this PR will undo #6418

Maybe there can be a list of exceptions to filter out in the python script?

kisvegabor:
> Maybe there can be a list of exceptions to filter out in the python script?

It's a little bit hacky, but probably good enough solution for now.

kdschlosser:
the only place private header files should be included are in the source file. They should not be included in ANY LVGL header file. The only purpose lvgl_private.h serves is to be used in bindings where the size of structures and the fields in structures needs to be known in order to generate the binding code. lvgl_private.h should not be included anywhere in LVGL.

kisvegabor:
@liamHowatt @FASTSHIFT can we reiterate on whether `lv_freetype_private.h` can be added to `lvgl_private.h`. Reading https://github.com/lvgl/lvgl/pull/6418 again I'm not entirely sure what the original problem was.

liamHowatt:
If I recall correctly, `lv_freetype_private.h` included `ft2build.h` which caused some error, but it's under `LV_USE_FREETYPE` so I think it's fine if `LV_USE_FREETYPE` is not enabled. The same applies for `lv_windows_input_private.h`. I can't see anything wrong either :+1:

FASTSHIFT:
> @liamHowatt @FASTSHIFT can we reiterate on whether `lv_freetype_private.h` can be added to `lvgl_private.h`. Reading #6418 again I'm not entirely sure what the original problem was.

How about we rename lv_freetype_private.h to lv_freetype_internal.h so that the script does not match the name.
As for why it is not recommended to expose lv_freetype_private.h to users, it is because the compilation system only exposes freetype targets to LVGL, and does not expose freetype to upper layer applications that use LVGL. Avoid exposing freetype macros and functions to user space to avoid contamination.

cc @liamHowatt

src/lvgl_private.h
```diff
@@ -15,93 +21,98 @@ extern "C" {
...
-#include "core/lv_obj_draw_private.h"
-#include "core/lv_obj_class_private.h"
-#include "core/lv_group_private.h"
-#include "core/lv_obj_event_private.h"
-#include "misc/lv_timer_private.h"
+#include "indev/lv_indev_private.h"
+#include "layouts/lv_layout_private.h"
+#include "libs/barcode/lv_barcode_private.h"
+#include "libs/ffmpeg/lv_ffmpeg_private.h"
+#include "libs/freetype/lv_freetype_private.h"
```
FASTSHIFT, line 57: Rename file `lv_freetype_private.h` -> `lv_freetype_internal.h`, bypassing named matching.

FASTSHIFT (changes requested):

kisvegabor:
@FASTSHIFT So it would look like:
- `lv_*.h`: public headers
- `lv_*_private.h`: for LVGL developers
- `lv_*_internal.h`: for LVGL only, never exposed

I feel like it's not so intuitive. What about `lv_feetype_build.h`? I found some none build related structs and functions there as well. Maybe they could go to a different header?

FASTSHIFT:
> @FASTSHIFT So it would look like:
> 
> * `lv_*.h`: public headers
> * `lv_*_private.h`: for LVGL developers
> * `lv_*_internal.h`: for LVGL only, never exposed
> 
> I feel like it's not so intuitive. What about `lv_feetype_build.h`? I found some none build related structs and functions there as well. Maybe they could go to a different header?

At present, `lv_freetype_private.h` is mainly used to refer to freetype headers and to store internal data structures (often also referring to freetype defined data structures) and macro definitions. There is no real point for users to use these apis, and I can't think of a reason to open these things up to users.

I still recommend using the `lv_freetype_internal.h` scheme to distinguish usage permissions.

kisvegabor:
I feel like `internal` and `private` are too similar. Anyway I have no better idea ether, so lets go for `internal`. It's not a practice, but something that we can judge case by case.

Issue #6477, opened 2024-07-05
Large stride setting for some displays


cristian-stoica:
### Introduce the problem

We have a display that requires a stride alignment value of 64 bytes. For the system to work we need to set LV_DRAW_BUF_STRIDE_ALIGN to 64. Because this setting is used by all buffer allocation functions (through width_to_stride) even small buffers (for letter drawing for example) become very large.

What could be a solution to avoid making large stride requirements, and thus avoid using unnecessary large buffers for temporary objects, when only the display  needs it?

Going further, how should we separate the stride requirements of object buffers (when they are used by accelerators like vglite and are dependent on hw internals) from stride requirements of display frame buffers?

kisvegabor:
What is the use case in which only the display's buffer has special stride requirements? Does it mean that all destination buffers shall have 64 byte stride, but the source buffers (e.g. letters) have different requirements?

cristian-stoica:
Yes, for example the buffers for letters don't require 64bytes stride alignment. Other buffers may require only 8,16 or no stride alignment.
 If we set the stride alignment for framebuffer to 64 bytes we can't set it differently for other buffers. All buffers will get the same stride alignment whether they need it or not.

kisvegabor:
It's not clear to me how can we decide the stride requirements. I know that it depends on the color format but is it different for source and destination buffers too?

cristian-stoica:
The stride requirements can be left to the lv_conf.h as before but I think it will need another macro to differentiate the target buffers.
I have a working prototype that I'll submit soon for review and comments.

cristian-stoica:
@kisvegabor The patch depends on [6530](https://github.com/lvgl/lvgl/pull/6530) being right. @nicusorcitu has some reservations on it so we need your review there.

Regardless, after looking at the patches for the last days I'm not particular fond of the solution and not going to send it for now. I'll just say that it duplicates width_to_stride (maybe also buf_align) into framebuffer-only versions that are used only during display initialization.

kisvegabor:
I don't understand what is the difference between the frame buffer and a layer which created by LVGL internally to snapshot a widget before rotation/scale. From the GPUs point of view the framebuffer and the internal layers are identical render targets. That's why I'm trying to understand if the different constraints apply to other buffer too (besides the frame buffer).

nicusorcitu:
@kisvegabor 
The short story about this ticket is that some display controllers requires a very large alignment (128/192 bytes) for both stride and start address.
```
[...]

PR #6947, opened 2024-09-26
Feat/partial and direct

kisvegabor:
Introduce a new rendering mode which combines the benefits of partial and direct rendering.

The problem was that
- direct mode is great because it provides tearing free result while updating only the changed areas
- but it's slow as usually its located in a slow external RAM
- partial mode is fast and support rotation too, but it's tearing

The advantages of the new  `LV_DISPLAY_RENDER_MODE_DIRECT_AND_PARTIAL`:
- render into a smaller buffer located in fast internal RAM, support rotation too
- the partial buffer should be flushed to a frame buffer
- with 2 partial buffers and DMA refreshing of the display and rendering can happen in parallel
- if there are 2 frame buffers, LVGL will sync them (same for direct mode)

The API is not final yet, and benchmarking is still required.

cc @XuNeo @FASTSHIFT @liamHowatt @nicusorcitu @espzav @tore-espressif 

### Notes

### DIFF
--- a/demos/benchmark/lv_demo_benchmark.c
+++ b/demos/benchmark/lv_demo_benchmark.c
@@ -464,7 +464,7 @@ static scene_dsc_t scenes[] = {
 
     {.name = "Containers",                 .scene_time = 3000, .create_cb = containers_cb},
     {.name = "Containers with overlay",    .scene_time = 3000, .create_cb = containers_with_overlay_cb},
-    {.name = "Containers with opa",        .scene_time = 1000, .create_cb = containers_with_opa_cb},
+    {.name = "Containers with opa",        .scene_time = 3000, .create_cb = containers_with_opa_cb},
     {.name = "Containers with opa_layer",  .scene_time = 3000, .create_cb = containers_with_opa_layer_cb},
     {.name = "Containers with scrolling",  .scene_time = 5000, .create_cb = containers_with_scrolling_cb},
 
--- a/demos/benchmark/lv_demo_benchmark.c
+++ b/demos/benchmark/lv_demo_benchmark.c
@@ -464,7 +464,7 @@ static scene_dsc_t scenes[] = {
 
     {.name = "Containers",                 .scene_time = 3000, .create_cb = containers_cb},
     {.name = "Containers with overlay",    .scene_time = 3000, .create_cb = containers_with_overlay_cb},
-    {.name = "Containers with opa",        .scene_time = 3000, .create_cb = containers_with_opa_cb},
+    {.name = "Containers with opa",        .scene_time = 1000, .create_cb = containers_with_opa_cb},
     {.name = "Containers with opa_layer",  .scene_time = 3000, .create_cb = containers_with_opa_layer_cb},
     {.name = "Containers with scrolling",  .scene_time = 5000, .create_cb = containers_with_scrolling_cb},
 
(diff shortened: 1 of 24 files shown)


kisvegabor:
@nicusorcitu 
This PR adds `lv_display_set_render_(draw)_buffers` and `lv_display_set_frame_(draw)_buffers` separately. I wouldn't like to bloat the API with `lv_dispay_set_frame_buffers_with_stride` but added a stride parameter to `lv_dispay_set_frame_buffers`. 

If agree with that we should remove `lv_display_set_buffers_with _stride` from v9.2.1 otherwise it will be missing from here.

XuNeo:
Correct me if I'm wrong. 
The new render mode can be regarded as partial mode, but we also have Full FB available(one or two), thus driver can do full update tearing free.

Should we do this in driver, instead of LVGL, because it's same as partial mode, and only differ in how to handle the partial update.Either directly to display, or to intermediate FB.

For `lvgl/src/drivers/nuttx/lv_nuttx_fbdev.c`, malloc a buffer and configure LVGL to use it as partial buffer, and we can decide how to update FB based on device capabilities.

nicusorcitu:
@cristian-stoica

cristian-stoica:
@kisvegabor > added a stride parameter to `lv_dispay_set_frame_buffers`.

https://github.com/lvgl/lvgl/pull/7073

kisvegabor:
>  in order to avoid changing the prototype of lv_dispay_set_frame_buffers in a maintenance release.

`lv_dispay_set_frame_buffers` is a new function, so we can't break the API with it. :sunglasses: We can simply make it look like `lv_dispay_set_frame_buffers(disp, buf1, buf2, stride, mode);`

kisvegabor:
> The new render mode can be regarded as partial mode, but we also have Full FB available(one or two), thus driver can do full update tearing free.

Exactly. Rendering into fast partial buffers, and copy them once in to full frame buffer(s) for tearing free result. 

> Should we do this in driver, instead of LVGL, because it's same as partial mode, and only differ in how to handle the partial update.Either directly to display, or to intermediate FB.

I was thinking about it too. However syncing the areas in case of double buffered direct mode is not trivial:
1. the old areas needs to be saved
2. before rendering the areas needs to be synced

Now we could do these only with some non-trivial custom display events.

In the end you can do everything as it were PARTIAL render, but on the last area you can get the current off screen buffer, and display it. See [here](https://github.com/lvgl/lvgl/pull/6947/files#diff-bea1688640231c37c9637d3a073340616cf6a7a677ed130670fd5d66ae1ac34aR245).

kisvegabor:
I'll get back to this soon. I need to find some time for benchmarking.

kisvegabor:
We are still on it.

kisvegabor:
Note stale. We should consider adding triple frame buffer handling here too.

liamHowatt:
So LVGL doesn't have to wait for VSYNC?

kisvegabor:
Or at least wait less. :slightly_smiling_face:

nicusorcitu:
> lv_dispay_set_frame_buffers is a new function, so we can't break the API with it. 😎 We can simply make it look like lv_dispay_set_frame_buffers(disp, buf1, buf2, stride, mode);
>
This was also my approach on this issue. https://github.com/lvgl/lvgl/issues/6477#issuecomment-2295810511

uLipe:
Not stale, it needs rebase.

Issue #4052, opened 2023-03-12
Styles are acting funny


kdschlosser:
### Perform all steps below and tick them with [x]
- [x] Read the [FAQ](https://docs.lvgl.io/master/intro/index.html#faq)
- [x] Check the related part of the [Documentation](https://docs.lvgl.io/)
- [x] Update lvgl to the latest version
- [x] Reproduce the issue in a [Simulator](https://docs.lvgl.io/latest/en/html/get-started/pc-simulator.html)

### Describe the bug
2 sliders are created identically except for the position. Setting the knob opacity to 0 the border and outline widths to 0 for both MAIN and INDICATOR. When the widgets get rendered to the screen the first slider has a border and the second does not. If you click on either of the sliders to move the bar then the border around the first slider disappears

### To Reproduce
https://sim.lvgl.io/v9.0/micropython/ports/javascript/index.html?script_startup=https://raw.githubusercontent.com/lvgl/lvgl/126fbee5ae0f568ed35d03736de00ef093b9bd0a/examples/header.py&script=https://raw.githubusercontent.com/lvgl/lvgl/126fbee5ae0f568ed35d03736de00ef093b9bd0a/examples/widgets/bar/lv_example_bar_5.py&script_direct=ad8dea7d648357165e93fa31e9dcad3a8e194a4f

### Expected behavior
for the styles to not have odd behavior.

kisvegabor:
It's a focus indication. The default theme adds an outline in `FOCUS_KEY` state. If there is a keypad or encoder indev LVGL focuses a widget by default


You can remove it either with
1. `slider1.set_style_outline_width(0, lv.STATE.FOCUS_KEY)`
2. Not adding a default group (`lv_group_set_default`) so the widgets won't be focused automatically.

kdschlosser:
if it was a focus issue then once the widget lost it's focus the outline would go away but then when it regained focus again the outline should come back. That is not the case. Once the focus goes away the outline goes away and if i focus the widget again the outline never comes back. Also the second slider never gets an outline when focus is given to it.

kisvegabor:
It's really a tricky case. This is what happens:
1. `lv_group_set_default` is called in MicroPython somewhere so controllable widget will be added to the default group automatically. This makes possible to use key navigation (TAB, arrows, etc) without any extra action.
[...]

PR #6955, opened 2024-09-27
fix(group): fix get_indev issue

mvladic:
Currently, `get_indev` returns a first POINTER indev **regardless of the given group**. Here is the description of this function:

"Find an indev preferably with POINTER type (because it's the most generic) that **uses the given group**."

The consequence of this is that the function `lv_group_focus_obj' does not work; and the first widget in the group is not focused (LV_STATE_FOCUS_KEY) by default if ENCODER or KEYBOARD is used. Both are working correctly in version 8.4.


mvladic:
Because of this bug, the functions `lv_group_focus_next` and `lv_group_focus_prev` also doesn't work. Maybe someone should check this and confirm.

mvladic:
One test fails. But, if this test expects that FOCUS_KEY state **is not** enabled then it is obvious while it fails because now it will be enabled which is expected in my opinion.

kisvegabor:
It makes perfect sense to use consider only the indevs if the given group. Could you update the screenshot in the test too?

mvladic:
> Could you update the screenshot in the test too?

Done.

mvladic:
Some remark related to this which I think are important. 

- This fix solves multiple issues. The most important is that `lv_group_focus_obj` didn't work.

- But, I believe not everybody will be happy with this change. For example, if you have both pointer and encoder inputs, maybe you don't want focus by default to the first object of the encoder group, because you consider pointer as your primary interface. 

So, only when user starts to use encoder you want to see focused object visual indicator. But, right now, without this change,  when you move encoder, for example to the RIGHT, it will not focus to the first object, but second.

Maybe, auto focus to the first object shouldn't be default. Or, it should be a way for the group to enable/disable this feature.

For the reference, here is the code that focus the first object added to the group and this can't be easily turned off:

image

mvladic:
After further investigation I found out that in issue #4052 the [decision](https://github.com/lvgl/lvgl/issues/4052#issuecomment-1502897210) was made to not focus by default. And here is the [commit](https://github.com/lvgl/lvgl/commit/d6cc2ff76f69bb7bb3f145da70a822c7fab3670e) which made the change.

So, I think this PR should be put on hold for now.

kisvegabor:
True. Let's close it for now.

PR #6960, opened 2024-09-28
chore(lv_conf_internal): avoid LV_CONF_SKIP redefinition warning

glebm:
When `LV_CONF_SKIP` is set from CMake, it is already defined at this point, triggering a warning.


kisvegabor (approved):
Thank you!

liamHowatt (approved):

PR #6973, opened 2024-10-01
chore(cmake): Add `LV_USE_SDL` CMake option

### LINKED ISSUE #6957: lvgl via CMake FetchContent + SDL does not allow for linking SDL to lvgl
### LVGL version

0daebca18b2e7d4c51d5b53e7f2a60df634850f2

### What happened?

I'm trying to set up a project using lvgl via FetchContent and SDL.

The problem is that lvgl's CMake doesn't seem to provide a way to link SDL to `lvgl::lvgl` target when using it via `FetchContent`.

If I don't link SDL, I get a build error:

```
FAILED: _deps/lvgl-build/CMakeFiles/lvgl.dir/src/drivers/sdl/lv_sdl_keyboard.c.o 
/usr/bin/cc -DLV_CONF_INCLUDE_SIMPLE -DLV_CONF_PATH=platforms/sdl3/lv_conf.h -DLV_LVGL_H_INCLUDE_SIMPLE -isystem build/_deps/lvgl-src -isystem platforms/sdl3 -isystem build/_deps/lvgl-build -g -MD -MT _deps/lvgl-build/CMakeFiles/lvgl.dir/src/drivers/sdl/lv_sdl_keyboard.c.o -MF _deps/lvgl-build/CMakeFiles/lvgl.dir/src/drivers/sdl/lv_sdl_keyboard.c.o.d -o _deps/lvgl-build/CMakeFiles/lvgl.dir/src/drivers/sdl/lv_sdl_keyboard.c.o -c build/_deps/lvgl-src/src/drivers/sdl/lv_sdl_keyboard.c
In file included from build/_deps/lvgl-src/src/drivers/sdl/../../display/../misc/../lv_conf_internal.h:54,
                 from build/_deps/lvgl-src/src/drivers/sdl/../../display/../misc/lv_types.h:16,
                 from build/_deps/lvgl-src/src/drivers/sdl/../../display/lv_display.h:16,
                 from build/_deps/lvgl-src/src/drivers/sdl/lv_sdl_window.h:17,
                 from build/_deps/lvgl-src/src/drivers/sdl/lv_sdl_keyboard.h:16,
                 from build/_deps/lvgl-src/src/drivers/sdl/lv_sdl_keyboard.c:9:
platforms/sdl3/lv_conf.h:1041:37: fatal error: SDL3/SDL.h: No such file or directory
 1041 |     #define LV_SDL_INCLUDE_PATH     <SDL3/SDL.h>
```

If I try to link SDL to lvgl like this:

```cmake
target_link_libraries(lvgl::lvgl PRIVATE SDL3::SDL3)
```

Then I get the following error:

```cmake
CMake Error at platforms/sdl3/CMakeLists.txt:5 (target_link_libraries):
  Cannot specify link libraries for target "lvgl::lvgl" which is not built by
  this project.
```

This error is expected, one can't modify targets imported via FetchContent after all.
lvgl currently provides no ways to customize the libraries linked to it.

Adding a variable for this to https://github.com/lvgl/lvgl/blob/master/env_support/cmake/custom.cmake is probably the way to fix this (e.g. `LV_LINK_LIBRARIES` that the user can then set up before FetchContent).
Is this the right place to report this issue?

### How to reproduce?

Create a CMake project following this guide: https://docs.lvgl.io/8.2/get-started/cmake.html

Then, in `lv_conf.h`, `#define LV_USE_SDL 1`

glebm:
Fixes #6957

### DIFF
--- a/Kconfig
+++ b/Kconfig
@@ -1561,6 +1561,9 @@ menu "LVGL configuration"
 		config LV_USE_SDL
 			bool "Use SDL to open window on PC and handle mouse and keyboard"
 			default n
+			help
+				If using lvgl via CMake, you should also set the `LV_USE_SDL` CMake option.
+
 		config LV_SDL_INCLUDE_PATH
 			string "SDL include path"
 			depends on LV_USE_SDL
--- a/env_support/cmake/custom.cmake
+++ b/env_support/cmake/custom.cmake
@@ -8,6 +8,9 @@ option(LV_LVGL_H_INCLUDE_SIMPLE
 option(LV_CONF_INCLUDE_SIMPLE
        "Use #include \"lv_conf.h\" instead of #include \"../../lv_conf.h\"" ON)
 
+option(LV_USE_SDL
+       "Link SDL2::SDL2 target to lvgl. Enable this if you use LV_USE_SDL in your lv_conf.h." OFF)
+
 # Option LV_CONF_PATH, which should be the path for lv_conf.h
 # If set parent path LV_CONF_DIR is added to includes
 if( LV_CONF_PATH )
@@ -39,6 +42,17 @@ endif()
 # Add definition of LV_CONF_SKIP only if needed
 if(LV_CONF_SKIP)
   target_compile_definitions(lvgl PUBLIC LV_CONF_SKIP=1)
+endif()
+
+if(LV_USE_SDL)
+  find_package(SDL2 REQUIRED)
+  if(TARGET SDL2::SDL2)
+    target_link_libraries(lvgl PRIVATE SDL2::SDL2)
+  elseif(TARGET SDL2::SDL2-static)
+    target_link_libraries(lvgl PRIVATE SDL2::SDL2-static)
+  else()
+    message(FATAL_ERROR "SDL2 package does not define a valid target (SDL2::SDL2 or SDL2::SDL2-static)")
+  endif()
 endif()
 
 # Include root and optional parent path of LV_CONF_PATH
--- a/lv_conf_template.h
+++ b/lv_conf_template.h
@@ -1035,7 +1035,9 @@
  * DEVICES
  *==================*/
 
-/** Use SDL to open window on PC and handle mouse and keyboard. */
+/** Use SDL to open window on PC and handle mouse and keyboard.
+ *  If using lvgl via CMake, you should also set the `LV_USE_SDL` CMake option.
+ */
 #define LV_USE_SDL              0
 #if LV_USE_SDL
     #define LV_SDL_INCLUDE_PATH     <SDL2/SDL.h>
--- a/src/lv_conf_internal.h
+++ b/src/lv_conf_internal.h
@@ -3284,7 +3284,9 @@
  * DEVICES
  *==================*/
 
-/** Use SDL to open window on PC and handle mouse and keyboard. */
+/** Use SDL to open window on PC and handle mouse and keyboard.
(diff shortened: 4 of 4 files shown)


kisvegabor:
Thanks!

I have 2 questions: 
-  Should we do the same with e.g. FreeType, X11, OpenGL, etc and all the supported 3rd party libraries?
- Now this part is really messy. We have `LV_CONF_BUILD_DISABLE_EXAMPLES/DEMOS/THORVG`. I wonder if we can figure out a nice unified way of handling these.

glebm:
Yes, all dependencies need to be linked, otherwise it won't build.

> We have LV_CONF_BUILD_DISABLE_EXAMPLES/DEMOS/THORVG. I wonder if we can figure out a nice unified way of handling these.

Yeah, all options that affect building should be controlled via CMake rather than the config file (i.e. we should add a define from CMake when the option is set). I think it can be done in a unified way.

glebm:
Is merging this blocked on anything?

kisvegabor:
Let's discuss the whole thing in https://github.com/lvgl/lvgl/pull/6974.

I've just added a comment.

glebm:
Replied there but I don't think this PR is in any way tied to defconfig. This PR is needed if you use LVGL as a CMake subproject regardless of whether you use defconfig or not.

kisvegabor:
I'm just trying to put the peaces together and find connection between multiple issues.

So main questions is, how to handle external libraries.
- How do we know which libraries needs to be linked? (if `LV_USE_SDL` is enabled?) 
- How tell CMake to link these libs?

The proposed solution could work for SDL, but I'm looking for a more generic solution. Just thinking aloud: Allow passing a path to `lv_conf.h`, e.g. `cmake .. LV_CONF_H=../lv_conf.h` in which CMake finds which libs should be linked. This is where I see the connection with #6974. 

The user should be able to manually pass options, e.g. `cmake .. LV_USE_SDL=1` in case `lv_conf.h` is not used. 

cc @etag4048

glebm:
I thin you're letting the proverbial perfect be the enemy of the good.
This PR is a bugfix and merging it doesn't prevent us from exploring future improvements.

A simple solution to this in the future would be to default the `LV_USE_SDL` CMake option to a value parsed from lv_conf.h or KConfig.

kisvegabor:
> I thin you're letting the proverbial perfect be the enemy of the good.

I'm sorry if it feels so. There were so many changes and attempts in the CMake files that it feels like a huge mess already and finally I'd like to be sure we are not just patching it, but trying to understand the real problems and providing a solution for them.  

Anything we add now is a responsibility as we shouldn't change it later (to avoid API breaks) and we should also maintain it. 

I don't want to make it perfect and I also believe in incremental changes. But let's be sure that what we doing now really brings us one step closer to an imagined good solution. 


---

In light of that using the `LV_USE_SDL` name in CMake sounds good, as we can extend it later with an `lv_conf.h` parser.

glebm:
> In light of that using the LV_USE_SDL name in CMake sounds good, as we can extend it later with an lv_conf.h parser.

So is anything blocking this now?

kisvegabor:
@etag4048 will work on https://github.com/lvgl/lvgl/issues/7223 to have a unified solution for "transferring" LVGL configs to CMake. He is a few weeks holiday, but he will work on it when he is back. 

Until that I wouldn't like to merge a temporal solution go for a  more complete one. 

I'm sorry this issue is so hard and slow to resolve. I hope it will be worth the time and effort.

PR #6974, opened 2024-10-01
chore(scripts): Support defconfig-based configs

glebm:
Adds a new CMake variable, `LV_CONF_DEFCONFIG_PATH`.
If set to a path to a defconfig or a .config file, that file will be used to generate lvgl configuration.

A defconfig file can be generated from a `.config` file using the kconfig-conf command:

```
kconfig-conf --savedefconfig .config Kconfig
```

I'll verify that this works as expected and update my cmake example repo (https://github.com/glebm/lvgl-cmake-example) once my SDL2 PR is merged. Leaving it as draft until then.

This should probably be the recommended way to use lvgl with cmake because it is much simpler and more robust than the manual `lv_conf.h` alternative. Docs can be updated once the dust has settled.

### DIFF
--- a/env_support/cmake/custom.cmake
+++ b/env_support/cmake/custom.cmake
@@ -8,11 +8,34 @@ option(LV_LVGL_H_INCLUDE_SIMPLE
 option(LV_CONF_INCLUDE_SIMPLE
        "Use #include \"lv_conf.h\" instead of #include \"../../lv_conf.h\"" ON)
 
-# Option LV_CONF_PATH, which should be the path for lv_conf.h
-# If set parent path LV_CONF_DIR is added to includes
-if( LV_CONF_PATH )
-    get_filename_component(LV_CONF_DIR ${LV_CONF_PATH} DIRECTORY)
-endif( LV_CONF_PATH )
+set(LV_CONF_PATH "" CACHE STRING
+    "Path to lv_conf.h.")
+set(LV_CONF_DEFCONFIG_PATH "" CACHE STRING
+    "Generate lv_conf.h from the given defconfig. If set, LV_CONF_PATH is unset and LV_CONF_SKIP is set to ON.")
+
+if(LV_CONF_DEFCONFIG_PATH)
+  set(LV_CONF_SKIP ON)
+  unset(LV_CONF_PATH)
+  set(KCONFIG_LV_CONF_DIR "${CMAKE_BINARY_DIR}/lv_kconfig_generated")
+  set(KCONFIG_LV_CONF_PATH "${KCONFIG_LV_CONF_DIR}/lv_kconfig_generated.h")
+  make_directory("${KCONFIG_LV_CONF_DIR}")
+  add_custom_command(
+    OUTPUT "${KCONFIG_LV_CONF_PATH}"
+    COMMAND
+    env KCONFIG_CONFIG='${LV_CONF_DEFCONFIG_PATH}'
+    '${LVGL_ROOT_DIR}/scripts/genconfig.py'
+    --header-path='${KCONFIG_LV_CONF_PATH}'
+    '${LVGL_ROOT_DIR}/Kconfig'
+    DEPENDS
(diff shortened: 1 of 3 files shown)


kisvegabor:
I don't fully get it. If Kconfig isnot used, we need to use  `lv_conf.h`, but if Kconfig is used we can use this new method the generate a list of `#define`s from Kconfig. What is the exact problem it solves?

glebm:
This generates the `#define`s from Kconfig at build time. How else would you do it?

kisvegabor:
Right no we have [`lv_conf_internal.h`](https://github.com/lvgl/lvgl/blob/master/src/lv_conf_internal.h#L106-L112) which maps Kconfig to LVGL defines.

glebm:
No, `lv_conf_internal.h` maps a header generated from Kconfig, not Kconfig itself.

kisvegabor:
Ah, got you! So the use case is that LVGL is used on its own (not in Zephyr, NuttX of ESP-IDF where Kconfig is already handled), and the user uses Kconfig to configure LVGL. After that we need to create `#define`s from `CONFIG_ASD=...` entries in a file. 

Is this correct?

glebm:
That's right!

glebm:
This is for using lvgl from a CMake project (without platform.io etc), not necessarily as a shared library.

I have an example of doing that https://github.com/glebm/lvgl-cmake-example (can be greatly simplified once #6973 and this PR are merged)

kisvegabor:
Okay, see now. Thank you! 

So the first thing we need to do is having a simple and documented way of running `make menuconfig` with a standalone LVGL lib.

After that optionally convert `sdkconfig.default` to `lv_conf.h`

The rest is tightly related to https://github.com/lvgl/lvgl/pull/6973.   With @liamHowatt we were talking about how to release Debian packages for LVGL (to allow installing them with `sudo apt-get install`). Our current best bet is having:
- `lvgl` core LVGL lib, everything enabled which are platform independent
- `lvgl-sdl_window`: depends on `lvgl` and `SDL2` and it contains only [these files ](https://github.com/lvgl/lvgl/tree/master/src/drivers/sdl)
- `lvgl-sdl_draw`: similarly it contains only [these](https://github.com/lvgl/lvgl/tree/master/src/draw/sdl)
- Same for `lvgl-opengl`, `lvgl-freetype`, etc.

IMO in order to do this we should build these extra libs as separate `.so` files. I can't oversee the whole CMake process, but it feels like it could help with linking the external libs too. What do you think?

glebm:
The SDL lib PR (https://github.com/lvgl/lvgl/pull/6973) is not coupled to this in any way.
That PR is needed regardless of whether you use LVGL via defconfig or not.

Likewise, Debian packages are orthogonal to defconfig.
Shipping an `so` file in the current state is not possible because you can't reconfigure it.

I think for that you'd need to split lvgl into parts that depend on `lv_conf.h` and parts that don't, and then ship the config-dependent parts as a source-only package. But again, this is not at all tied to this PR!

kisvegabor:
I've just opened #7223 to discuss our CMake strategy in general. 
Closing it until then.

glebm:
The issue you've linked says nothing about generating lv_conf.h from defconfig in CMake. That issue only discusses CMake options.

kisvegabor:
You are right, this PR is not related to CMake. It seems I'm trying to keep too many things in my head (and don'r read the comment history :smile: )

However to really make it independent from CMake, as first step let's focus on the `scripts` only. Please some docs about the usage to the beginning of `scripts/genconfig.py`

glebm:
> Please some docs about the usage to the beginning of `scripts/genconfig.py`

That script already starts with several paragraphs of documentation. :confused: 
As it comes from Kconfiglib (https://github.com/ulfalizer/Kconfiglib/blob/master/genconfig.py), I don't think we should be modifying it unnecessarily.

glebm:
OK. Anyway, this will remain a draft PR until SDL is supported correctly (e.g. https://github.com/lvgl/lvgl/pull/6973).

PR #6985, opened 2024-10-02
Introducing Declarative Widget and UI Descriptions

kisvegabor:
Hey,

As you might know, we are working on LVGL's UI editor. It will be a developer-centric tool to help you re-implement UI designs quickly and let you add code for the logic. This tool is not for designers (designers should use Figma or similar tools) and not a no-code tool for super beginners. We are targeting professional developers who want to implement UIs quickly in a flexible and maintainable way.

The tool has 3 parts:
1. **The editor itself**: Describe your widgets in XML and see how they look like in an instant preview. This is similar to Google Chrome's inspector, but you can do much more (define the API, constants, font and image manager, add tests, export code, etc.). It also has an outstanding auto-complete engine to let you write UIs very fast.
2. **Figma import**: Lets you transfer parts of a design to the editor, not the whole Figma project, just the components that you select. It gives you and the designer flexibility and freedom, as you don't have to follow strict conventions for compatibility.
3. **Online preview and testing**: The widgets that you develop can be tested in multiple setups (various API parameters, sizes, etc.). All the widgets and tests are combined into a web page to preview them online and share them with the rest of your team.

This is the wireframe layout of the tool:
image

This PR is about 1) and shows how I imagine describing the widgets in XML. You will find the description of the built-in widgets and a [detailed README](https://github.com/kisvegabor/lvgl_upstream/blob/xml-widgets/xmls/README.md) to describe the design.

This tool will be for you, so please let me know what you think! Any feedback is very welcome!

cc @richardgazdik @coevin @liamHowatt @XuNeo @FASTSHIFT @jeremy-baker 

### Notes

### DIFF
--- a/demos/benchmark/lv_demo_benchmark.c
+++ b/demos/benchmark/lv_demo_benchmark.c
@@ -399,18 +399,18 @@ static void widgets_demo_cb(void)
  **********************/
 
 static scene_dsc_t scenes[] = {
-    //    {.name = "Empty screen",               .scene_time = 3000, .create_cb = empty_screen_cb},
-    //    {.name = "Moving wallpaper",           .scene_time = 3000, .create_cb = moving_wallpaper_cb},
-    //    {.name = "Single rectangle",           .scene_time = 3000, .create_cb = single_rectangle_cb},
-    //    {.name = "Multiple rectangles",        .scene_time = 3000, .create_cb = multiple_rectangles_cb},
-    //    {.name = "Multiple RGB images",        .scene_time = 3000, .create_cb = multiple_rgb_images_cb},
-    //    {.name = "Multiple ARGB images",       .scene_time = 3000, .create_cb = multiple_argb_images_cb},
-    //    {.name = "Rotated ARGB images",        .scene_time = 3000, .create_cb = rotated_argb_image_cb},
-    //    {.name = "Multiple labels",            .scene_time = 3000, .create_cb = multiple_labels_cb},
-    //    {.name = "Screen sized text",          .scene_time = 5000, .create_cb = screen_sized_text_cb},
-    //    {.name = "Multiple arcs",              .scene_time = 3000, .create_cb = multiple_arcs_cb},
-    //
-    //    {.name = "Containers",                 .scene_time = 3000, .create_cb = containers_cb},
+    {.name = "Empty screen",               .scene_time = 3000, .create_cb = empty_screen_cb},
+    {.name = "Moving wallpaper",           .scene_time = 3000, .create_cb = moving_wallpaper_cb},
+    {.name = "Single rectangle",           .scene_time = 3000, .create_cb = single_rectangle_cb},
+    {.name = "Multiple rectangles",        .scene_time = 3000, .create_cb = multiple_rectangles_cb},
+    {.name = "Multiple RGB images",        .scene_time = 3000, .create_cb = multiple_rgb_images_cb},
+    {.name = "Multiple ARGB images",       .scene_time = 3000, .create_cb = multiple_argb_images_cb},
+    {.name = "Rotated ARGB images",        .scene_time = 3000, .create_cb = rotated_argb_image_cb},
+    {.name = "Multiple labels",            .scene_time = 3000, .create_cb = multiple_labels_cb},
+    {.name = "Screen sized text",          .scene_time = 5000, .create_cb = screen_sized_text_cb},
+    {.name = "Multiple arcs",              .scene_time = 3000, .create_cb = multiple_arcs_cb},
+
+    {.name = "Containers",                 .scene_time = 3000, .create_cb = containers_cb},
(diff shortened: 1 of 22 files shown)


jusodi:
Since you mention auto-complete in your announcement, my first comment on this would be to make sure that elements & attributes are sanity checked. I personally feel this is what caused WPF to die a slow death (it's probably not dead but it's dead to me because of this). It only choked at run time, which made no sense to me.

TL;DR Do not allow auto-complete (or manual entry) of attributes for elements which do not handle them. Do not allow successful compilation for the same.

kdschlosser:
You need to create schemas for the XML. They need to be made properly and so they make sense. I can help out with doing this if it is needed.

kisvegabor:
Thanks for joining the discussion! :blush: Yeah, we ere also thinking about Schema.  Based on `<api>` and other tags we can generate the schema and do the validation automatically.  My dream si to make it impossible to enter an incorrect value. E.g. when a parameter's type is enum you can't type value just select from a dropdown.

JAremko:
Spec + informative validation errors can help a lot to generating UI with GPT/Claude by cycling the code trough them until all red is gone.

kdschlosser:
I am about 60% done with a script that will generate the schema programmatically. if you are interested...

kdschlosser:
> My dream si to make it impossible to enter an incorrect value. E.g. when a parameter's type is enum you can't type value just select from a dropdown.

Your dream is going to be a very fast reality. This is exactly what xsd files are intended to do. You can use the xsd to populate all of your GUI controls for the various functions and widgets. Right down to the min and max values for the different integer types. All the enumerations will be available as well.....

It is going to generate an xsd file for each of the public header files.

kisvegabor:
According to my vision XSD should be built from the `<api>` tag which should be way easier to parse than H files.

kdschlosser:
yeah but you have to manually update the api xml files. The point to the XSD files is you will not need to even have the api xml files. The entire C API will literally be spelled out in the XSD files.

Instead of breaking from what people know you should be using the C API names in the XML files. Making a completely different API for the XML is going to end up becoming a nightmare to maintain. The LVGL API is pretty large and any changes made in the LVGL API would propagate to the XSD files by running a script. 

The software you are writing would read those XSD files as the way to set up it's controls and it will use to to generate the XML output that ends up being parsed to run the GUI. whether that parsing is done in C code, Python code or whatever other binding code that exists for LVGL it will not matter.

I am sure you are going to be providing some kind of a way to read the XML files to create the GUI right? Well how that is done is using the XSD files to validate the XML that is being parsed. If the names in the XSD files aligns with the C API those same XSD files could be used to write the code that reads the XML files to be able to run whatever it needs to run.

JAremko:
Can API be generated from headers?

kdschlosser:
the parsing of the header files is a snap to do. There are libraries that already do this and they do it quite well. The one I used was castxml. It is so easy to use you should consider using it as a replacement to pycparser..

This is what it outputs....

[lvgl_temp_header.xml.txt](https://github.com/user-attachments/files/17255116/lvgl_temp_header.xml.txt)

kisvegabor:
I was thinking about generating API/Schema from the header, and I'm not sure if it's good idea. For example 
```c
void lv_buttonmatrix_set_map(lv_obj_t * obj, const char * const map[]);
```

doesn't tell that the array should be `NULL` terminated.

Or you don't know if 
```c
void lv_obj_add_flag(lv_obj_t * obj, lv_obj_flag_t f);
```
accepts ORed values.

kdschlosser:
those are things that can be handled by looking at the documentation. you can add comments into the docstrings that don't end up in the final documentation but would be accessible via the XML files. if there is something special about a parameter for a function like you mentioned above add a comment with the parameter name and then a colon and some information that you would be able to use to know what needs to be done. It's a lot easier to go through and add the comments for the functions that have special needs than it is to write an entire XML API for every single bit and piece of LVGL.

JAremko:
> I was thinking about generating API/Schema from the header, and I'm not sure if it's good idea. For example

Thing's like this usually best when they are 95% automated.  Then it saves a lot of time without adding a ton of complexity to get it up to this sweet 98% percent   :smile:

So adjusting generated API by hand is totally ok.  Especially if there's a special mode with runtime check that can generate missing parts or emit warnings.

JAremko:
Also you can create a map of overrides that map function definition to what should be emitted.  This way you won't have to fix API by hand every time. That's basically like creating a patch file and re-applying it.

kisvegabor:
> So adjusting generated API by hand is totally ok. Especially if there's a special mode with runtime check that can generate missing parts or emit warnings.

I agree. We aim for handwriting as a first step and will see how to speed it up.

kisvegabor:
Update: we are working loading XML components to LVGL in run time. We will send a PR shortly.

kisvegabor:
Closing in favor of https://github.com/lvgl/lvgl/pull/7252

PR #6986, opened 2024-10-02
chore(fs): treat '/' as valid driver letter 

XuNeo:
When set `LV_FS_POSIX_LETTER = '/'`, do not generate warnings.

### Notes

### DIFF
--- a/src/libs/fsdrv/lv_fs_win32.c
+++ b/src/libs/fsdrv/lv_fs_win32.c
@@ -17,17 +17,9 @@
 /*********************
  *      DEFINES
  *********************/
-#if LV_FS_WIN32_LETTER == '\0'
-    #error "LV_FS_WIN32_LETTER must be set to a valid value"
-#else
-    #if (LV_FS_WIN32_LETTER < 'A') || (LV_FS_WIN32_LETTER > 'Z')
-        #if LV_FS_DEFAULT_DRIVE_LETTER != '\0' /*When using default drive letter, strict format (X:) is mandatory*/
-            #error "LV_FS_WIN32_LETTER must be an upper case ASCII letter"
-        #else /*Lean rules for backward compatibility*/
-            #warning LV_FS_WIN32_LETTER should be an upper case ASCII letter. \
-            Using a slash symbol as drive letter should be replaced with LV_FS_DEFAULT_DRIVE_LETTER mechanism
-        #endif
-    #endif
+
+#if !LV_FS_IS_VALID_LETTER(LV_FS_WIN32_LETTER)
+    #error "Invalid drive letter"
 #endif
 
 #define MAX_PATH_LEN 256
--- a/scripts/lv_conf_internal_gen.py
+++ b/scripts/lv_conf_internal_gen.py
@@ -163,7 +163,7 @@ for line in fin.read().splitlines():
     fout.write(f'{line}\n')
 
 fout.write(
-'''
+r'''
 
 /*----------------------------------
  * End of parsing lv_conf_template.h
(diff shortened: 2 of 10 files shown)


kisvegabor:
Is using `LV_FS_DEFAULT_DRIVE_LETTER` an option for you?

XuNeo:
Hi @kisvegabor ,

I'm not fully understand how to use `LV_FS_DEFAULT_DRIVE_LETTER`. The basic need is to set image in abs path like "/system/resource/image.bin"

kisvegabor:
You can enable `LV_USE_FS_POSIX` and set
-  `LV_FS_POSIX_LETTER 'A'`
- `LV_FS_DEFAULT_DRIVE_LETTER 'A'`

This way when the driver letter is missing in the path, it will be redirected to posix (as the default letter shows posix).

XuNeo:
That sounds not intuitive.
Set `LV_FS_POSIX_LETTER = '/' ` is more straight forward.

kisvegabor:
However it won't work a relative path.

XuNeo:
Hi @kisvegabor ,

Can we merge this? The definition LV_FS_POSIX_LETTER = '/' has worked well since LVGL v8 and provides a simple, effective solution. I don't see a reason to maintain the warning against its usage."

kisvegabor:
What about this?
```c
#if LV_FS_POSIX_LETTER == '\0'
    #error "LV_FS_POSIX_LETTER must be set to an upper case letter"
#else
    /*Allow only upper case letters and '/'  ('/' is a special case for backward compatibility)*/
    #if (LV_FS_POSIX_LETTER != '/' && (LV_FS_POSIX_LETTER < 'A') || (LV_FS_POSIX_LETTER > 'Z'))
        #error "LV_FS_POSIX_LETTER must be an upper case ASCII letter"
    #endif
#endif
```

If you agree please update the other FS drivers too.

liamHowatt:
When the default drive letter was added, we agreed to keep `'/'` as a valid drive letter. I think this warning was an accident.

XuNeo:
Updated. Please help to review. Thanks!

FASTSHIFT (dismissed):

kissa96 (approved):

src/libs/fsdrv/lv_fs_win32.c
```diff
@@ -17,17 +17,9 @@
...
-    #if (LV_FS_WIN32_LETTER < 'A') || (LV_FS_WIN32_LETTER > 'Z')
-        #if LV_FS_DEFAULT_DRIVE_LETTER != '\0' /*When using default drive letter, strict format (X:) is mandatory*/
-            #error "LV_FS_WIN32_LETTER must be an upper case ASCII letter"
-        #else /*Lean rules for backward compatibility*/
-            #warning LV_FS_WIN32_LETTER should be an upper case ASCII letter. \
-            Using a slash symbol as drive letter should be replaced with LV_FS_DEFAULT_DRIVE_LETTER mechanism
-        #endif
-    #endif
+
+#if !LV_FS_IS_VALID_LETTER(LV_FS_STDIO_LETTER)
```
liamHowatt:
```suggestion
#if !LV_FS_IS_VALID_LETTER(LV_FS_WIN32_LETTER)
```

liamHowatt (changes requested):
Looks good otherwise

src/libs/fsdrv/lv_fs_win32.c
XuNeo: Nice catch. I double checked and still missed this one. :(

liamHowatt (approved):

FASTSHIFT (approved):

andrewleech:
As a follow up to this, I've just found the default `LV_FS_POSIX_LETTER` in  [lv_conf_template.h](https://github.com/lvgl/lvgl/blob/abdcc04f09ef9c958d7f6cfc6579c16cbfe353e9/lv_conf_template.h#L796) is still `'\0'` so causes a compile fail if the template is used as-is.

Could this file get updated to match the new requirements here, I'm not too sure what the preferred character should be?

PR #6993, opened 2024-10-03
fix(obj): fix crash during multiple screen loader 

forqzy:
fix(lv_obj_draw): Update lv_obj_draw.c to avoid crash during multiple screen loader


FASTSHIFT:
Are there any examples of recurrence? Why pass a null pointer to `lv_obj_get_ext_draw_size`?

kisvegabor:
Please send a code snippet to reproduce the original issue.

PR #6999, opened 2024-10-03
docs: fix 'my_disp_flush' in quick-overview

Narukara:
Fix 'my_disp_flush' in quick-overview. In parallel rendering architecture, `lv_color_t*` in `lv_display_flush_cb_t` has been changed to `uint8_t*`.

Related code:

https://github.com/lvgl/lvgl/blob/a853a1289ae40898122d83be1842d0b61c56b15d/src/display/lv_display.h#L81

https://github.com/lvgl/lvgl/blob/a853a1289ae40898122d83be1842d0b61c56b15d/examples/porting/lv_port_disp_template.c#L124-L138

### DIFF
--- a/docs/get-started/quick-overview.rst
+++ b/docs/get-started/quick-overview.rst
@@ -65,15 +65,15 @@ If you would rather try LVGL on your own project follow these steps:
 
      lv_display_set_flush_cb(display, my_disp_flush);
 
-     void my_disp_flush(lv_display_t * disp, const lv_area_t * area, lv_color_t * color_p)
+     void my_disp_flush(lv_display_t * disp, const lv_area_t * area, uint8_t * px_map)
      {
          int32_t x, y;
          /*It's a very slow but simple implementation.
-          *`set_pixel` needs to be written by you to a set pixel on the screen*/
+          *`put_px` needs to be written by you to put a pixel on the screen*/
          for(y = area->y1; y <= area->y2; y++) {
              for(x = area->x1; x <= area->x2; x++) {
-                 set_pixel(x, y, *color_p);
-                 color_p++;
+                 put_px(x, y, *px_map);
+                 px_map++;
              }
          }
 


kissa96 (approved):
LGTM!

docs/get-started/quick-overview.rst
```diff
@@ -65,15 +65,15 @@ If you would rather try LVGL on your own project follow these steps:
...
          int32_t x, y;
          /*It's a very slow but simple implementation.
-          *`set_pixel` needs to be written by you to a set pixel on the screen*/
+          *`put_px` needs to be written by you to put a pixel on the screen*/
          for(y = area->y1; y <= area->y2; y++) {
              for(x = area->x1; x <= area->x2; x++) {
-                 set_pixel(x, y, *color_p);
-                 color_p++;
+                 put_px(x, y, *px_map);
+                 px_map++;
```
kisvegabor, line 76: By this we assume that the display has 8 bit color format. 

```c
  uint16_t * px_map_16 = (uint16_t*) px_map;
                 put_px(x, y, *px_map_16);
                 px_map_16++;   /*Go to the next pixel*/
```

kisvegabor (dismissed):

Narukara:
Thank you for pointing this out 👍

> ```c
>   uint16_t * px_map_16 = (uint16_t*) px_map;
>   for(y = area->y1; y <= area->y2; y++) {
>              for(x = area->x1; x <= area->x2; x++) {
>                  put_px(x, y, *px_map_16);
>                  px_map_16++;   /*Go to the next pixel*/
> ```

Does this code mean the display has 16 bit color format? If my understanding is correct, in `display_flush_cb`, the pointer `px_map` needs to be converted to the appropriate color format?

If that's the case, I have two suggestions:

1. Add more explanation and example code about this to quick-overview.
2. Why not change the type of `px_map` in `lv_display_flush_cb_t` to `void*`? `px_map` is not actually a pointer to `uint8_t`. It points to some unknown type, which is what `void*` is for.

Narukara:
I noticed that the documentation has been reorganized #7136. That's pretty cool. I think I can close this PR.

> 2. Why not change the type of `px_map` in `lv_display_flush_cb_t` to `void*`? `px_map` is not actually a pointer to `uint8_t`. It points to some unknown type, which is what `void*` is for.

Actually I still have doubts about this. Maybe I will submit an issue later?

kisvegabor:
> Why not change the type of px_map in lv_display_flush_cb_t to void*? px_map is not actually a pointer to uint8_t. It points to some unknown type, which is what void* is for.

Actually I don't exactly remember why I used `uint8_t *`. Probably I wanted to emphasize that the buffer is in bytes and not pixels.
Anyway, changing it now would be an API break. :slightly_frowning_face:

PR #7001, opened 2024-10-04
perf(label): simplify handling of bytes overwritten with dots

niklasf:
Dots and null terminator will overwrite at most LV_LABEL_DOT_NUM + 1 bytes, and there will always be a null terminator after the dots. So we do not need to worry about how many characters those bytes represent, or even if we overwrite a multi-byte character partially. Therefore the dynamic memory allocation path for replaced characters can be simplified away.

Use the new fast revert/set to ensure that self size calculations and placing new dots always happens based on text without dots.

### DIFF
--- a/src/widgets/label/lv_label.c
+++ b/src/widgets/label/lv_label.c
@@ -32,7 +32,7 @@
 
 #define LV_LABEL_DEF_SCROLL_SPEED   lv_anim_speed_clamped(40, 300, 10000)
 #define LV_LABEL_SCROLL_DELAY       300
-#define LV_LABEL_DOT_END_INV 0xFFFFFFFF
+#define LV_LABEL_DOT_BEGIN_INV 0xFFFFFFFF
 #define LV_LABEL_HINT_HEIGHT_LIMIT 1024 /*Enable "hint" to buffer info about labels larger than this. (Speed up drawing)*/
 
 /**********************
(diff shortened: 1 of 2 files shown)


kisvegabor:
Hey,

Amazing idea! Could you add a few tests too?

niklasf:
There's already test for rendering with `LV_LABEL_LONG_DOT`, but I guess this could be made unit-testable. Moved `lv_label_set_dots()` and `lv_label_set_dots()` to `lv_label_private.h`, so that they can be exercised directly in a test case.

kisvegabor:
Oh, sorry I forgot that we already have tests for DOTS. In light of that no knew test is required and the last commit can be reverted.

niklasf:
No worries. Reverted to keep the (private) API surface smaller.

kisvegabor (approved):
Thank you, I've just rebased the PR on master.

liamHowatt (approved):
Very nice!

Just flagging this behavior as a possible issue. Not sure:
```c
lv_obj_t * lab = lv_label_create(lv_screen_active());
lv_obj_set_size(lab, 3, 10);
lv_label_set_text(lab, "abc");
lv_label_set_long_mode(lab, LV_LABEL_LONG_DOT);
```

image

It only dots with longer text, i.e. `"abcd"`. No memory issues though.

niklasf:
Thanks for reviewing!

Fwiw, handling very short labels like that appears to be a concious decision, i.e., the previous code also had:
```c
else if(lv_text_get_encoded_length(label->text) <= LV_LABEL_DOT_NUM) {   /*Don't turn to dots all the characters*/
```

kisvegabor:
> Fwiw, handling very short labels like that appears to be a concious decision, i.e., the previous code also had:

Confirmed, it was my decision many years ago. 🙂

PR #7024, opened 2024-10-10
docs(yocto): add guide to integrate lvgl recipe in Yocto

EDGEMTech-GabrielC:
Added Yocto documentation: 
- Basic components of Yocto explained
- How to use the existing lvgl recipe and create a recipe from scratch
- Variable Glossary and link to Yocto terms documentation

The objective is to guide developers that are not familiar with Yocto to understand the basics of Yocto philosophy and the minimal requirements to be able to integrate an application using LVGL in an already existing Yocto project. Usually, sellers like TI, NXP, Toradex etc.. will provide a Yocto project to create an image for a board. It can be tedious to understand what is going on and this guide's purpose is to ease this process.

The guide can followed only using the commands and executing the steps, but any step is explained in detail, with what is created/generated to be able to debug if something went wrong.

### DIFF
--- a/docs/integration/os/yocto/lvgl_recipe.rst
+++ b/docs/integration/os/yocto/lvgl_recipe.rst
@@ -1,0 +1,913 @@
+LVGL in Yocto
+#############
+
+This chapter serves as a guide to help you create a recipe for an application 
+using LVGL. While the process will be demonstrated for the Raspberry Pi 3 
+(64-bit), it can be applied to any board in a similar way.
+
+Build Host Packages 
+*******************
+You must install essential host packages on your build host. The following 
+command installs the host packages based on an Ubuntu distribution
+
+.. code-block::
+
+    sudo apt install gawk wget git diffstat unzip texinfo gcc build-essential \
+    chrpath socat cpio python3 python3-pip python3-pexpect xz-utils \
+    debianutils iputils-ping python3-git python3-jinja2 python3-subunit zstd \ 
+    liblz4-tool file locales libacl1
+
+.. note::
+
+    For host package requirements on all supported Linux distributions, see the 
+    `Required Packages for the Build Host <https://docs.yoctoproject.org/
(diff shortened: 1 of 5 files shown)


docs/integration/os/yocto/lvgl_recipe.rst
```diff
@@ -0,0 +1,913 @@
...
+   CMAKE_PROJECT_VERSION = "2.0"
+
+   do_install() {
+       install -d ${D}${bindir}
+       install -m 0755 ${S}/bin/lvglbenchmark ${D}${bindir}
+   }
+
+The sources come from ``lv_port_linux`` repository. We apply 2 patches to modify the ``CMakeLists.txt`` and ``lv_conf.h``.
+
+Patch 1
```
kisvegabor, line 694: Can we fix LVGL so that these patches are not required?

kisvegabor (approved):
Fantastic work! Thank you!

docs/integration/os/yocto/lvgl_recipe.rst
EDGEMTech-GabrielC: Yes, but I prefer including these patches because they demonstrate how users can apply one or more patches. Since the guide is also intended to provide some tips for Yocto, I think this is a reasonable way to illustrate the process, don’t you agree?

FASTSHIFT (approved):

docs/integration/os/yocto/lvgl_recipe.rst
XuNeo: > demonstrate how users can apply one or more patches.

How about fix those issues and include a demo purpose patch, say change one line of comment instead? So the code can work out of box without patching?

EDGEMTech-GabrielC: I agree with that. I guess I'll wait the current PR on lv_port_linux to fix Wayland and I'll update with your suggestion.

PR #7025, opened 2024-10-10
docs(fbdev): add information about resolution issue

EDGEMTech-GabrielC:
Added information about an issue with fbdev that some users can encounter.


kisvegabor (dismissed):

docs/integration/driver/display/fbdev.rst
```diff
@@ -39,3 +39,25 @@ node on the display (usually this is ``/dev/fb0``).
 If your screen stays black or only draws partially, you can try enabling direct rendering via ``LV_DISPLAY_RENDER_MODE_DIRECT``. Additionally,
 you can activate a force refresh mode with ``lv_linux_fbdev_set_force_refresh(true)``. This usually has a performance impact though and shouldn't
 be enabled unless really needed.
+
+
+Common mistakes
```
XuNeo: Please check if this version is better for understanding:
```

When the Linux kernel initializes, it sets up subsystems like the framebuffer (fbdev) to manage display output. If an HDMI display is connected, the kernel detects it and allocates a portion of RAM as the framebuffer, which holds the pixel data for rendering images.

However, a common issue arises when the display is not powered on during the boot process. If the board is powered on before the screen, the kernel may incorrectly configure the framebuffer resolution. As a result, both the visible and virtual resolutions can be incorrect, leading to display problems.

This issue often occurs with HDMI connections where the display is powered up after the system has already booted.
```

kisvegabor: @EDGEMTech-GabrielC is on a holiday now, and will get back to this soon.

uLipe:
@EDGEMTech-GabrielC  do you have any update on this?

kisvegabor (dismissed):

uLipe (commented):
Nice, and instructive text!

I just would like something that users can do to mitigate the issue, for example if this is an issue of power-up order, instruct users to make sure to respect the correct powerup order to have a proper experience when using the fbdev.

kisvegabor:
Once we are touching the fbdev docs, could you also add a section about [hiding the cursor of the terminal](https://blog.lvgl.io/2018-01-03/linux_fb#hide-the-cursor)?

EDGEMTech-GabrielC:
I explained the correct way to connect everything so the Fbdev issue won't occur. I also added the section about the cursor. I'm not sure what was the issue so I could not provide more information about this. I hope it is what you were expecting.

kisvegabor (approved):
Thank you!

uLipe (approved):

liamHowatt (approved):

PR #7061, opened 2024-10-14
Sysmon

cristian-stoica:
Fixes for performance benchmark code:
- calculate performance averages only for the rendered frames
- completely disable performance banners on display when results are redirected to tty
- add configuration option to disable performance updates during rendering

### Notes
- [x] If you added new options to `lv_conf_template.h` run [lv_conf_internal_gen.py](https://github.com/lvgl/lvgl/blob/master/scripts/lv_conf_internal_gen.py) and update [Kconfig](https://github.com/lvgl/lvgl/blob/master/Kconfig).

### DIFF
--- a/Kconfig
+++ b/Kconfig
@@ -1409,6 +1409,11 @@ menu "LVGL configuration"
 			depends on LV_USE_PERF_MONITOR
 			default n
 
+		config LV_USE_PERF_MONITOR_OBSERVER
+			bool "Prints or displays performance data during rendering"
+			depends on LV_USE_PERF_MONITOR
+			default y
+
 		config LV_USE_MEM_MONITOR
 			bool "Show the used memory and the memory fragmentation"
 			default n
--- a/demos/benchmark/lv_demo_benchmark.c
+++ b/demos/benchmark/lv_demo_benchmark.c
@@ -510,7 +510,7 @@ void lv_demo_benchmark(void)
 #if LV_USE_PERF_MONITOR
     lv_display_t * disp = lv_display_get_default();
     lv_subject_add_observer_obj(&disp->perf_sysmon_backend.subject, sysmon_perf_observer_cb, title, NULL);
-#if LV_USE_PERF_MONITOR_LOG_MODE
+#if LV_USE_PERF_MONITOR_LOG_MODE || !LV_USE_PERF_MONITOR_OBSERVER
     lv_obj_add_flag(title, LV_OBJ_FLAG_HIDDEN);
 #endif
 #else
@@ -575,7 +575,7 @@ static void sysmon_perf_observer_cb(lv_observer_t * observer, lv_subject_t * sub
         scene_name[0] = '\0';
     }
 
-#if !LV_USE_PERF_MONITOR_LOG_MODE
+#if !LV_USE_PERF_MONITOR_LOG_MODE && LV_USE_PERF_MONITOR_OBSERVER
     lv_obj_t * label = lv_observer_get_target(observer);
     lv_label_set_text_fmt(label,
                           "%s"
--- a/demos/benchmark/lv_demo_benchmark.c
+++ b/demos/benchmark/lv_demo_benchmark.c
@@ -510,6 +510,9 @@ void lv_demo_benchmark(void)
 #if LV_USE_PERF_MONITOR
     lv_display_t * disp = lv_display_get_default();
     lv_subject_add_observer_obj(&disp->perf_sysmon_backend.subject, sysmon_perf_observer_cb, title, NULL);
+#if LV_USE_PERF_MONITOR_LOG_MODE
+    lv_obj_add_flag(title, LV_OBJ_FLAG_HIDDEN);
+#endif
 #else
     lv_label_set_text(title, "LV_USE_PERF_MONITOR is not enabled");
 #endif
@@ -563,9 +566,8 @@ static void next_scene_timer_cb(lv_timer_t * timer)
 static void sysmon_perf_observer_cb(lv_observer_t * observer, lv_subject_t * subject)
 {
     const lv_sysmon_perf_info_t * info = lv_subject_get_pointer(subject);
-    lv_obj_t * label = lv_observer_get_target(observer);
-
     char scene_name[64];
+
     if(scenes[scene_act].name[0] != '\0') {
         lv_snprintf(scene_name, sizeof(scene_name), "%s: ", scenes[scene_act].name);
     }
@@ -573,6 +575,8 @@ static void sysmon_perf_observer_cb(lv_observer_t * observer, lv_subject_t * sub
         scene_name[0] = '\0';
     }
 
+#if !LV_USE_PERF_MONITOR_LOG_MODE
+    lv_obj_t * label = lv_observer_get_target(observer);
     lv_label_set_text_fmt(label,
                           "%s"
                           "%" LV_PRIu32" FPS, %" LV_PRIu32 "%% CPU\n"
@@ -581,6 +585,9 @@ static void sysmon_perf_observer_cb(lv_observer_t * observer, lv_subject_t * sub
                           info->calculated.fps, info->calculated.cpu,
                           info->calculated.render_avg_time + info->calculated.flush_avg_time,
                           info->calculated.render_avg_time, info->calculated.flush_avg_time);
+#else
+    LV_UNUSED(observer);
+#endif
 
     /*Ignore the first call as it contains data from the previous scene*/
     if(scenes[scene_act].measurement_cnt != 0) {
--- a/env_support/cmsis-pack/lv_conf_cmsis.h
+++ b/env_support/cmsis-pack/lv_conf_cmsis.h
@@ -823,6 +823,9 @@
 
         /*0: Displays performance data on the screen, 1: Prints performance data using log.*/
         #define LV_USE_PERF_MONITOR_LOG_MODE 0
+
+        /* 0: Display performance data only at the end, 1: Performance data is printed periodically */
+        #define LV_USE_PERF_MONITOR_OBSERVER 1
     #endif
 
     /*1: Show the used memory and the memory fragmentation
--- a/lv_conf_template.h
+++ b/lv_conf_template.h
@@ -937,6 +937,9 @@
 
         /** 0: Displays performance data on the screen; 1: Prints performance data using log. */
         #define LV_USE_PERF_MONITOR_LOG_MODE 0
+
+        /* 0: Display performance data only at the end, 1: Performance data is printed periodically */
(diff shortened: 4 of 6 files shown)


nicusorcitu:
@kisvegabor can we have this included in 9.2.1 as well?

kisvegabor:
I've commented in https://github.com/lvgl/lvgl/pull/7071

cristian-stoica:
not stale

kissa96 (approved):

kisvegabor:
From [here](https://github.com/lvgl/lvgl/pull/7071#issuecomment-2416107545):
> With the patch:
> 
> sysmon: 57 FPS (refr_cnt: 17 | redraw_cnt: 1), refr 6ms (render 102ms | flush 2ms), CPU 38%
> So there were 17 refreshes from which only one was rendered (in 102 ms).
> 
> If there were 2 renders, each lasting 100 ms, the average render time still should be 100ms (which is the case with the patch too)
> 
> However, the 57 FPS is quite misleading as the max possible FPS is 10 with 100 ms rendering time.

Maybe I missed something, but did we came to a conclusion with it?

cristian-stoica:
The remaining question is whether the proposed solution is right as is or we need to make amends.

The problem is two-fold resulting in better numbers than subjective perception would suggest:
- During the LV_SYSMON_REFR_PERIOD_DEF period, averages for CPU and render time includes data also from non-rendered frames
- The rendered count increases also on non-rendered frames.

What is on my mind - and I couldn't tell - is whether the restarting of calculations when a new test start is still in effect (if it was before) or whether there are any leaks of perf data from one test to the next

kisvegabor:
The benchmark is a very sensitive topic, so let take is slow to be sure that we fully understand what's happening. 🙂

Let's start with [this part](https://github.com/lvgl/lvgl/pull/7061/files?diff=unified&w=1#diff-931dc821c9118f8f5432800178c30f1c9bc562ad1c72ff51ebdd8c0aed5232d9L252-L260): `avg_render_time` and `avg_flush_time`.   Just to confirm they don't change at all right? 

> What is on my mind - and I couldn't tell - is whether the restarting of calculations - when a new test start - is still in effect (if it was before) or whether there are any leaks of perf data from one test to the next

[Here](https://github.com/lvgl/lvgl/blob/master/demos/benchmark/lv_demo_benchmark.c#L592) the first measurement data is dropped as it might containt data from the previous scene.

cristian-stoica:
> Let's start with [this part](https://github.com/lvgl/lvgl/pull/7061/files?diff=unified&w=1#diff-931dc821c9118f8f5432800178c30f1c9bc562ad1c72ff51ebdd8c0aed5232d9L252-L260): `avg_render_time` and `avg_flush_time`. Just to confirm they don't change at all right?

Their values and formulas don't change after the patch. Only fps and CPU averages change because they get moved under 'if' together with 'run_cnt`.  When there is no rendering, they all stay zero and added as such in https://github.com/lvgl/lvgl/blob/master/demos/benchmark/lv_demo_benchmark.c#L592

Going in a different direction, an alternative to changing the measurements is to use a config switch to always force rendering. I'm don't know how this can be done.

kisvegabor:
> Their values and formulas don't change after the patch. Only fps and CPU averages change because they get moved under 'if' together with 'run_cnt`. 

I'm testing this PR and I saw a log line like this when I click a button:
```
sysmon: 10 FPS (refr_cnt: 3 | redraw_cnt: 3), refr 9ms (render 1ms | flush 8ms), CPU 9%
```
Here the 10FPS is quite misleading. I don't understand what is the advantage of reporting 10 FPS when the refresh happened in 10ms (it could make 100 FPS). The users might think that it's so slow that even a button click drops the FPS to 10. 

The same in the upstream version:
```
sysmon: 30 FPS (refr_cnt: 9 | redraw_cnt: 5), refr 5ms (render 0ms | flush 9ms), CPU 19%
```
which says that there is no performance drop when a button is clicked.

cristian-stoica:
> ```
> sysmon: 10 FPS (refr_cnt: 3 | redraw_cnt: 3), refr 9ms (render 1ms | flush 8ms), CPU 9%
> ```
> 
> 
>     
>   
> 
> Here the 10FPS is quite misleading.

I agree. Reporting 10FPS for what would be 100fps if all frames were rendered back-to-back is misleading.
I can understand why this is so:  3 rendered frames in the (I assume the default) 300 ms measuring interval result is 10 frames in a second.

Probably If we consider calculating FPS against refresh time instead of refresh count, then the results should be more meaningful.

What is the capability of the device you tested? It might be that 30FPS is not telling the whole truth either if it can sustain 9ms refresh (with render) times.

kisvegabor:
> Probably If we consider calculating FPS against refresh time instead of refresh count (normalizing), then the results will be more meaningful.

I agree, however, what should happen when nothing is rendered? The proposed changes will report 0 FPS, the original will report 30 FPS. 

I still believe 30 FPS is more meaningful to the users as probably they want to know when the performance drops. 
On the screen we can't display too much details, but in the log we already report the refresh count too. 

>  What is the capability of the device you tested? It might be that 30FPS is not telling the whole truth either if it can sustain 9ms refresh (with render) times.

AMD Ryzen 7 5800H  :smile: I was testing it on my notebook.

cristian-stoica:
Personally, I want to measure the performance of the device not of the display.

If you consider like me that the issues mentioned here https://github.com/lvgl/lvgl/pull/7061#issuecomment-2449763803 need to be addressed, then we can find solutions for the corner cases.

I find the report misleading when it says 30FPS with 9ms render time, or 30 fps with no rendering. In fact, if there is no rendering, I think zero is the right number - just don't average it in with the rest of interval measurements. I also find it weird to see 60fps when I have three rendered frames at 200ms+.

In the end, the question is what we really want to show as perf numbers. Otherwise, as I mentioned elsewhere, I would find it easier to drop this patches and switch on a flag that says, 'render all frames'

kisvegabor:
> During the LV_SYSMON_REFR_PERIOD_DEF period, averages for CPU and render time includes data also from non-rendered frames

Let's say only 1 frame was rendered in 30 ms during the 300 ms rendering period. Isn't is correct to say that the CPU usage 10%?


> The refresh count increases also on non-rendered frames.

I don't have a strong opinion about it as I don't use the refresh count on its own for performance benchmarking, however it's probably related to whether we should display 0 FPS or 30FPS when idle. 

I see why you find 30FPS misleading when LVGL did nothing, however from the user's point of view it has the massage of "my display has no speed issues now". In more detail, this how I see it
- **idle**:  0ms rendering time: 30 FPS, 0% CPU usage.
- **update only a small are**:  1ms render time, still 30FPS, 3% CPU. So a little larger load still results in the same FPS number.
- **update a larger are**: 30ms render time, 30FPS, 90% CPU usage. Still can be handles but high CPU usage.
- **very large update**:  60ms render time, 16FPS, 100% CPU usage.

So the load increases the FPS decrease and the CPU usage increases, which makes sense to me.

cristian-stoica:
> Let's say only 1 frame was rendered in 30 ms during the 300 ms rendering period. Isn't is correct to say that the CPU usage 10%?

If CPU is 100% for that frame and idle in the remaining 270 ms, the average will be 10% for the whole period. Continuing my argument, what would the CPU load be if all frames were rendered?


> > The refresh count increases also on non-rendered frames.
> 
> I don't have a strong opinion about it as I don't use the refresh count on its own for performance benchmarking, however it's probably related to whether we should display 0 FPS or 30FPS when idle.

I think refresh count is essential in calculating FPS. Are we free to ignore its value when it is directly linked to FPS?
`info->calculated.fps = 1000 * info->measured.refr_cnt / `time_since_last_report`

XuNeo:
This seems to have been discussed before. I suggest displaying the average valid FPS (i.e., 1/average_refresh_time), which provides meaningful insight.

For example, if a label refreshes every 300ms and each refresh takes 10ms, the refresh times are 10ms, N/A, N/A, ..., N/A, 10ms over 300ms, giving a frame rate of 100 FPS.

Note that we can average the refresh time but should not average the frame rate.

XuNeo:
If no valid refresh happened in the averaging window, then display `--`.

kisvegabor:
@XuNeo 
>  I suggest displaying the average valid FPS (i.e., 1/average_refresh_time), which provides meaningful insight.

Let's say the FPS is limited to 30 and we have 30 1ms renders. It would mean 1000 FPS, which is not that realistic as we didn't have that many Frames Per Second. 

@uLipe 
> One thing that FPS metrics points to is the amount of complete I/O operations, which can be a successful render (even if partially) operation which may take less time than the required to draw the entire frame.

Is the "LVGL checks if there is anything to render, but there isn't therefore it does nothing" also considered as I/O operation? If only real I/O operations matter we should display 3 FPS for a 3 Hz label update. If skipped I/O operations also matter we should display 30 FPS.


@vwheeler63 
> From my viewpoint as system Architect on a firmware system, I want to see the number of times lv_refr() was called in 1 second.

There is no `lv_refr()` function. If you mean `lv_display_refr_timer()` then it's 30FPS for 3Hz label update.

XuNeo:
> @XuNeo 
> 
> >  I suggest displaying the average valid FPS (i.e., 1/average_refresh_time), which provides meaningful insight.
> 
> 
> 
> Let's say the FPS is limited to 30 and we have 30 1ms renders. It would mean 1000 FPS, which is not that realistic as we didn't have that many Frames Per Second. 
> 

This is the difference between render frame rate and display frame rate. We can also show the latter one.

vwheeler63:
> If you mean `lv_display_refr_timer()`

Yes, that is what I meant.  But even if I set the refresh rate to be a different rate, I still want to see how many times it got called per second, i.e. if something is slowing it down (e.g. some long rendering), it will show up in a reduced frame rate.

kisvegabor:
> This is the difference between render frame rate and display frame rate. We can also show the latter one.

I find misleading to report more **F**rames **P**er **S**econds which is actually happens. In the worst case having only 1 render lasing 1 ms, would result in 1000 FPS.

vwheeler63:
> I find misleading to report more **F**rames **P**er **S**econds which is actually happens. In the worst case having only 1 render lasing 1 ms, would result in 1000 FPS.

Yet knowing that LVGL optimizes the drawing so only changed things need to get moved to the display panel seems to make it intuitive for me.  Seeing 1000 FPS, it is clear what is happening.  And when it drops, it makes it clear that some "drawing load" is occurring.  From an Architectural viewpoint (data flows), my eye is focused on ensuring it doesn't drop below 15 or 20 so it doesn't start to look jittery.

kisvegabor:
We are displaying `refresh time (render time + flush time)` too. 
image

Here you can see the average time spent by rendering and flushing. And as it is displayed as time it has a realistic meaning.

cristian-stoica:
not stale

PR #7070, opened 2024-10-15
bugfix for fatfs freetype fps

ThunderDai:
1. when use freetype and fatfs for esp32, then will compilation failed
2. When calculating fps, direct shaping and discarding may result in a significant difference from the actual frame rate. Therefore, rounding should be used more reasonably

### DIFF
--- a/env_support/cmake/esp.cmake
+++ b/env_support/cmake/esp.cmake
@@ -61,7 +61,7 @@ else()
   idf_component_register(SRCS ${SOURCES} ${EXAMPLE_SOURCES} ${DEMO_SOURCES}
       INCLUDE_DIRS ${LVGL_ROOT_DIR} ${LVGL_ROOT_DIR}/src ${LVGL_ROOT_DIR}/../
                    ${LVGL_ROOT_DIR}/examples ${LVGL_ROOT_DIR}/demos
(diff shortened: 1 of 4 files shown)


FASTSHIFT:
Can you submit a separate PR?

ThunderDai:
> Can you submit a separate PR?

hi, do you means one PR just one commit?

FASTSHIFT:
> > Can you submit a separate PR?
> 
> hi, do you means one PR just one commit?

Yes, it is recommended to split the PR for modifications to different modules, so that it is convenient to review and modify them separately.

ThunderDai:
> > > Can you submit a separate PR?
> > 
> > 
> > hi, do you means one PR just one commit?
> 
> Yes, it is recommended to split the PR for modifications to different modules, so that it is convenient to review and modify them separately.

oh, thanks

ThunderDai:
> > > Can you submit a separate PR?
> > 
> > 
> > hi, do you means one PR just one commit?
> 
> Yes, it is recommended to split the PR for modifications to different modules, so that it is convenient to review and modify them separately.

In fact, perhaps I should directly submit an issue? Because the problem I currently need to solve is relatively simple, and there shouldn't be many modifications

PR #7071, opened 2024-10-15
feat(sysmon): optimize frame rate calculation method

cristian-stoica:
(related to https://github.com/lvgl/lvgl/pull/7061)
 Fixes for performance benchmark code:

calculate performance averages only for the rendered frames
completely disable performance banners on display when results are redirected to tty
add configuration option to disable performance updates during rendering

### Notes
- [x] If you added new options to `lv_conf_template.h` run [lv_conf_internal_gen.py](https://github.com/lvgl/lvgl/blob/master/scripts/lv_conf_internal_gen.py) and update [Kconfig](https://github.com/lvgl/lvgl/blob/master/Kconfig).

### DIFF
--- a/src/others/sysmon/lv_sysmon.c
+++ b/src/others/sysmon/lv_sysmon.c
@@ -277,11 +277,10 @@ static void perf_update_timer_cb(lv_timer_t * t)
 
 static void perf_observer_cb(lv_observer_t * observer, lv_subject_t * subject)
 {
-    lv_obj_t * label = lv_observer_get_target(observer);
     const lv_sysmon_perf_info_t * perf = lv_subject_get_pointer(subject);
 
 #if LV_USE_PERF_MONITOR_LOG_MODE
-    LV_UNUSED(label);
+    LV_UNUSED(observer);
     LV_LOG("sysmon: "
            "%" LV_PRIu32 " FPS (refr_cnt: %" LV_PRIu32 " | redraw_cnt: %" LV_PRIu32"), "
            "refr %" LV_PRIu32 "ms (render %" LV_PRIu32 "ms | flush %" LV_PRIu32 "ms), "
@@ -290,6 +289,7 @@ static void perf_observer_cb(lv_observer_t * observer, lv_subject_t * subject)
            perf->calculated.refr_avg_time, perf->calculated.render_avg_time, perf->calculated.flush_avg_time,
            perf->calculated.cpu);
 #else
+    lv_obj_t * label = lv_observer_get_target(observer);
     lv_label_set_text_fmt(
         label,
         "%" LV_PRIu32" FPS, %" LV_PRIu32 "%% CPU\n"
--- a/src/others/sysmon/lv_sysmon.c
+++ b/src/others/sysmon/lv_sysmon.c
@@ -236,34 +236,35 @@ static void perf_update_timer_cb(lv_timer_t * t)
     uint32_t LV_SYSMON_GET_IDLE(void);
 
     lv_sysmon_perf_info_t * info = &disp->perf_sysmon_info;
-    info->calculated.run_cnt++;
+    if(info->measured.render_cnt != 0) {
+        info->calculated.run_cnt++;
 
-    uint32_t time_since_last_report = lv_tick_elaps(info->measured.last_report_timestamp);
-    lv_timer_t * disp_refr_timer = lv_display_get_refr_timer(NULL);
-    uint32_t disp_refr_period = disp_refr_timer->period;
+        uint32_t time_since_last_report = lv_tick_elaps(info->measured.last_report_timestamp);
+        lv_timer_t * disp_refr_timer = lv_display_get_refr_timer(NULL);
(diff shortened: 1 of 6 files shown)


src/others/sysmon/lv_sysmon.c
```diff
@@ -236,34 +242,35 @@ static void perf_update_timer_cb(lv_timer_t * t)
...
-                                                                     info->measured.flush_in_render_elaps_sum) /
-                                                                    info->measured.render_cnt) : 0;
-
-    info->calculated.cpu_avg_total = ((info->calculated.cpu_avg_total * (info->calculated.run_cnt - 1)) +
-                                      info->calculated.cpu) / info->calculated.run_cnt;
-    info->calculated.fps_avg_total = ((info->calculated.fps_avg_total * (info->calculated.run_cnt - 1)) +
-                                      info->calculated.fps) / info->calculated.run_cnt;
-
-    lv_subject_set_pointer(&disp->perf_sysmon_backend.subject, info);
+    if(info->measured.render_cnt != 0) {
```
kisvegabor, line 245: I'm not sure about it. I think the performance needs to be reported even if there were no rendering, just sent some special values, e.g:
```c
info->calculated.fps = 1000 / disp_refr_period;
 info->calculated.cpu = 100 - LV_SYSMON_GET_IDLE();
 info->calculated.refr_avg_time = 0;
 info->calculated.render_avg_time = 0;
...
```

kisvegabor (commented):
Did it solve the performance issue of v9.2.1?

src/others/sysmon/lv_sysmon.c
cristian-stoica: Is there a way to force rendering even when scene doesn't change?

cristian-stoica:
> Did it solve the performance issue of v9.2.1?

These patches do not target the performance issue that we've observed on this branch. @cosmindanielradu19 is looking into that issue.

We've observed for some time issues with the reported data and the averages do not look right when rendering does not occur. It doesn't make sense to average no-render frames with rendered ones - hence these patches.

nicusorcitu:
> Did it solve the performance issue of v9.2.1?
>
No, the commit that introduced the performance drop will be figured out by @cosmindanielradu19

nicusorcitu:
@cristian-stoica 
Please copy-paste the sysmon logs with and without your patch. It will be much cleaner what it fixes and how it influence the FPS value.

cristian-stoica:
[perf_widgets_v9.2-g0237e9fce7.txt](https://github.com/user-attachments/files/17391283/perf_widgets_v9.2-g0237e9fce7.txt)
[perf_widgets_sysmon-v9.2.txt](https://github.com/user-attachments/files/17391285/perf_widgets_sysmon-v9.2.txt)

These are the console logs of widgets test with performance data redirected to console.

cristian-stoica:
Indeed, I'm not exactly clear why for the interval where the frame took 100ms there was only one redraw. By count this is 1FPS but by time it would be 10FPS.
I've analyzed the render/refresh events last week and they seemed OK. One explanation I have is that the animation is really not doing anything more than one frame there.

NB. Please have in mind that patch https://github.com/lvgl/lvgl/pull/7071/commits/869625fd71dd070df8963769d34dc52186edcc95
while removing all the banners, removes also the necessity of drawing the letters in them: the amount of rendered frames drops significantly because of this.

kisvegabor:
If the screen is large, it can really happen that e.g. the first page of the widget's demo is not doing anything (just rendered once).

In this case it can be really misleading if only the CPU usage indicator is updated because it counts as one more render_cnt. However it's updated quite rarely. (300 ms by default) 

For other tests (e.g. the empty screen test) it shouldn't be the case. Those animations should be fast enough to change the scene for all frames. 

---

The FPS values should be normalized to 1 sec even if we have measured the rendering for 300 ms. So if LVGL spent e.g. 200 ms with refreshing. and could render 4 times during this 200 ms, we know that
- the avg. refresh time is 200 / 4 = 50ms
- The FPS is 1000ms / 50ms = 20.

cristian-stoica:
> first page of the widget's demo is not doing anything
That makes perfect sense and clarifies the results. 

The introduction of config option might not be appropriate for the 9.2.1 however the no-banner fix can be cherry-picked independently if it makes sense at this late stage in the release. On the other hand, it has the obvious side-effect that it changes results of the tests when redirection to tty happens so it's a tough call to make.

What are your intentions?

cristian-stoica:
> Could you open a dedicated PR?

Pull request https://github.com/lvgl/lvgl/pull/7061 is for master and this one is for v9.2.1. We have had all the discussions here.

kisvegabor:
I meant keeping only 869625f in the PR but for the sake of simplicity I cherry-picked it [here](https://github.com/lvgl/lvgl/commit/51e1dfcb40c52eb194138c7a4afadc828eb8c4bb)

nicusorcitu:
Thanks Gabor!

kisvegabor:
Closing this PR.

Issue #6014, opened 2024-04-03
Support mouse wheel scroll and other gestures


kisvegabor:
### Problem to solve

The pointer input devices can do more then just press/release. The mouse has a scroll wheel and multiple buttons, touchpads can detect multiple gestures, like rotate, pitch, or two finger scroll. 
 

### Success criteria

The main gestures/actions can be detected and applied:
- scroll with wheel
- handle multiple mouse buttons
- forward complex touch gestures

### Solution outline

In `lv_indev_data_t` we can have a `gesture` field which can contain the detected gesture:
```c
- LV_INDEV_GESTURE_SCROLL, //mouse wheel or 2 fingers
- LV_INDEV_GESTURE_ROTATE, //will set `style_transform_rotate`
- LV_INDEV_GESTURE_SCALE, //will set `style_transform_scale`
- LV_INDEV_GESTURE_BUTTON_1, //indicates clicking with button 1
- LV_INDEV_GESTURE_BUTTON_2, //indicates clicking with button 2
- ...
```
This way the input device can report only a single point (x, y) but it can tell what to do there: rotate, scroll, etc.

As some touch controllers can detect these gestures but other can't we can have a function like:
```c
gesture = lv_indev_detect_gesture_from_points(points, point_cnt);
```
It can be called from the `read_cb` on demand, so LVGL itself needs to process only `LV_INDEV_GESTURE`s.

### Rabbit holes

- We need more info about the capabilities of he touch controllers.
- Detecting gestures can be very complicated. We can consider using a 3rd party library.

### Testing

For processing `LV_INDEV_GESTURE`s unit test work.

`lv_indev_detect_gesture_from_points` can be also tested with a lot of sample data.

### Teaching

Update the docs, example and the existing driver.

kisvegabor:
cc @XuNeo @bjsylvia

kisvegabor:
Not stale

XuNeo:
> This way the input device can report only a single point (x, y) but it can tell what to do there: rotate, scroll, etc.

I like this idea of simplicity. The upper application can use the event directly and ignore multiple touch input points.

It may worth to check how android deals with it, to see if more special actions needed.

> We need more info about the capabilities of he touch controllers.

We can use a configure option. It's less likely we need to change the algorithm between software and hardware in run time.

kisvegabor:
> We can use a configure option. It's less likely we need to change the algorithm between software and hardware in run time.

[...]

PR #7078, opened 2024-10-15
feat(indev): add multi touch gestures

etag4048:

@kisvegabor 
This a draft of the new multi-touch recognition feature

### DIFF
--- a/demos/gestures/lv_demo_gestures.c
+++ b/demos/gestures/lv_demo_gestures.c
@@ -1,0 +1,208 @@
+/*
+ * @file lv_demo_gestures.c
+ * @author EDGEMTech Ltd. erik.tagirov@edgemtech.ch
+ *
+ * This is a simple example program that demonstrates how to use
+ * the gesture recognition API, please refer to lv_indev_gesture.h or the documentation 
+ * for more details
+ *
+ * The demo starts with a single rectangle that is scaled when a pinch gesture
+ * is detected. A single finger moves the rectangle around,
+ * a swipe gesture creates another rectangle to interact with - up to 5 are allowed.
+ * A rotation gesture, obviously rotates the rectangle.
+ */
+
+/*********************
+ *      INCLUDES
+ *********************/
+#include "lv_demo_gestures.h"
+
+#if LV_USE_DEMO_GESTURES
+
+/*********************
+ *      DEFINES
+ *********************/
+
+#define RECT_INIT_WIDTH 300.0
+#define RECT_INIT_HEIGHT 300.0
+#define RECT_COLOR 0xC1BCFF
+
+/**********************
+ *      TYPEDEFS
+ **********************/
+
+/**********************
+ *  STATIC PROTOTYPES
+ **********************/
+
+static void label_scale(lv_event_t *gesture_event);
+static void label_move(lv_event_t *event);
+
+/**********************
+ *  STATIC VARIABLES
+ **********************/
+
+static lv_obj_t *label;
+static lv_style_t label_style;
+static float label_width;
+static float label_height;
+static float label_x;
+static float label_y;
+
+/**********************
+ *      MACROS
+ **********************/
+
+/**********************
+ *   GLOBAL FUNCTIONS
+ **********************/
+
+/**
+ * Entry point it creates the screen, and the label
+ * Set event callbacks on the label
+ */
+void lv_demo_gestures(void)
+{
+    lv_obj_t *rectangle;
+    lv_obj_t *root_view;
+
+    label_width = RECT_INIT_WIDTH;
+    label_height = RECT_INIT_HEIGHT;
+    label_y = label_x = 300;
+
+    root_view = lv_screen_active();
+
+    lv_obj_set_style_bg_color(root_view, lv_color_hex(0xFFFFFF), LV_PART_MAIN);
+    label = lv_label_create(root_view);
+    lv_obj_remove_flag(root_view, LV_OBJ_FLAG_SCROLLABLE);
+
+    lv_label_set_text(label, "Zoom or move");
+    lv_obj_add_flag(label, LV_OBJ_FLAG_CLICKABLE);
+
+    lv_style_init(&label_style);
+    lv_style_set_bg_color(&label_style, lv_color_hex(RECT_COLOR));
+    lv_style_set_bg_opa(&label_style, LV_OPA_COVER);
+
+    lv_style_set_width(&label_style, (int)label_width);
+    lv_style_set_height(&label_style, (int)label_height);
+
+    lv_style_set_x(&label_style, (int)label_x);
+    lv_style_set_y(&label_style, (int)label_y);
+
+    lv_obj_add_style(label, &label_style, LV_STATE_DEFAULT);
+
+    lv_obj_add_event_cb(label, label_scale, LV_EVENT_GESTURE_PINCH, label);
+    lv_obj_add_event_cb(label, label_move, LV_EVENT_PRESSING, label);
+
+}
+
+/**********************
+ *   STATIC FUNCTIONS
+ **********************/
+
+/**
+ * Called when a pinch event occurs - scales the label
+ * @param gesture_event point to a LV_EVENT_PINCH event
+ */
+static void label_scale(lv_event_t *gesture_event)
+{
+
+    static int initial_w = -1;
+    static int initial_h = -1;
+    lv_indev_gesture_recognizer_t *recognizer;
+    lv_indev_gesture_state_t state;
+    lv_point_t center_pnt;
+    float scale;
+
+    recognizer = lv_indev_gesture_get_recognizer(gesture_event);
+    scale = lv_indev_gesture_get_scale(gesture_event);
+    state = lv_indev_gesture_get_state(gesture_event);
+
+    lv_indev_gesture_get_center_point(gesture_event, &center_pnt);
+
+    if (state == LV_INDEV_GESTURE_STATE_ENDED) {
+        /* Pinch gesture has ended - reset the width/height for the next pinch gesture*/
+        initial_w = -1;
+        initial_h = -1;
+
+        LV_LOG_TRACE("label end scale: %g %d\n", scale, state);
+        return;
+    }
+
+    if (initial_h == -1 || initial_w == -1) {
+
+        LV_ASSERT(state == LV_INDEV_GESTURE_STATE_RECOGNIZED);
+
+        /* Pinch gesture has been recognized - this is the first event in a series of recognized events */
+        /* The scaling is applied relative to the original width/height of the rectangle */
+        initial_w = label_width;
+        initial_h = label_height;
+
+        LV_LOG_TRACE("label start scale: %g\n", scale);
+    }
+
+    /* The gesture has started or is on-going */
+
+    /* Avoids a situation where the rectangle becomes too small */
+    if (scale < 0.4) {
+        return;
+    }
+
+    /* Subtract the detection threshold - Avoids choppy scaling */
+    if (scale > 1.0) {
+        scale -= 0.5;
+    }
+
+    label_width = initial_w * scale;
+    label_height = initial_h * scale;
+    label_x = center_pnt.x - label_width / 2;
+    label_y = center_pnt.y - label_height / 2;
+
+    LV_LOG_TRACE("label scale: %g label x: %g label y: %g w: %g h: %g\n",
+            scale, label_x, label_y, label_width, label_height);
+
+    /* Update position and size */
+    lv_style_set_width(&label_style, (int)label_width);
+    lv_style_set_height(&label_style, (int)label_height);
+    lv_style_set_x(&label_style, (int)label_x);
+    lv_style_set_y(&label_style, (int)label_y);
+
+    lv_obj_add_style(label, &label_style, LV_STATE_DEFAULT);
+
+}
+
+/**
+ * Called when a LV_EVENT_PRESSING occurs on the rectangle - moves the label
+ * @param event pointer to the event
+ */
+static void label_move(lv_event_t *event)
+{
+    lv_point_t pnt;
+    lv_indev_t *indev;
+    lv_indev_gesture_recognizer_t *recognizer;
+    lv_indev_gesture_state_t state;
+
+    indev = (lv_indev_t *)lv_event_get_param(event);
+    recognizer = lv_indev_gesture_indev_get_recognizer(indev);
+    state = lv_indev_gesture_recognizer_get_state(recognizer);
+
+    /* Do not move and when a pinch gesture is ongoing */
+    if (recognizer != NULL && state == LV_INDEV_GESTURE_STATE_RECOGNIZED) {
+        return;
+    }
+
+    lv_indev_get_point(indev, &pnt);
+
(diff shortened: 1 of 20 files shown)


etag4048:
Related issue: https://github.com/lvgl/lvgl/issues/6014

src/indev/lv_indev_gesture.h
```diff
@@ -0,0 +1,115 @@
...
+	LV_INDEV_GESTURE_STATE_CANCELED,	/* Canceled - usually a finger is lifted */
+} lv_indev_gesture_state_t;
+
+/* Data structures for touch events - used to repsensent a libinput event */
+/* Emitted by devices capable of tracking identifiable contacts (type B) */
+typedef struct {
+	lv_point_t point;		/* Coordinates of the touch */
+	lv_indev_state_t state;		/* The state i.e PRESSED or RELEASED */
+	uint8_t id;			/* Identification/slot of the contact point */
+	uint32_t ts;			/* Timestamp in milliseconds */
```
kisvegabor: Avoid unclear abbreviations
```suggestion
	uint32_t timestamp;			/* Timestamp in milliseconds */
```

src/indev/lv_indev_gesture.h
```diff
@@ -0,0 +1,115 @@
...
+	lv_indev_state_t state;		/* The state i.e PRESSED or RELEASED */
+} lv_indev_gesture_motion_t;
+
+/* General descriptor for a gesture, used by recognizer state machines to track 
+ * the scale, rotation, and translation */
+typedef struct {
+	lv_indev_gesture_type_t type;    
+
+	/* Motion descriptor, stores the coordinates and velocity of a contact point */
+	lv_indev_gesture_motion_t motions[LV_GESTURE_MAX_TOUCHES];
```
kisvegabor:
```suggestion
	lv_indev_gesture_motion_t motions[LV_GESTURE_MAX_POINTS];
```

src/indev/lv_indev_gesture.h
```diff
@@ -0,0 +1,115 @@
...
+	float p_rotation;
+	uint8_t finger_cnt;		 		/* Current number of contact points */
+
+} lv_indev_gesture_t;
+
+/* Gesture recognizers - pinch gesture */
+typedef struct {
+	lv_indev_gesture_state_t state;
+	lv_indev_gesture_t *info;
+	float scale;
```
kisvegabor: Avoid floats by defaults
```suggestion
	uint32_t scale; /*256: no scale, 128: half size, 512 double size*/
```

src/indev/lv_indev_gesture.h
```diff
@@ -0,0 +1,115 @@
...
+	float rotation;
```
kisvegabor:
```suggestion
	int32_t rotation; /*0.1 degree resolution. E.g. 456 means 45.6 degree*/
```

src/indev/lv_indev_gesture.h
```diff
@@ -0,0 +1,115 @@
...
+ * GLOBAL PROTOTYPES
+ **********************/
+
+/**
+ * Detect a pinch gesture
+ * @param gesture_state 	pointer to a gesture recognizer
+ * @param touches 		pointer to the first element of the collected touch events
+ * @param touch_cnt 		length of passed touch event array.
+ */
+void lv_indev_gesture_detect_pinch(lv_indev_gesture_pinch_t *gesture, lv_indev_touch_data_t *touches, uint16_t touch_cnt);
```
kisvegabor: If only `lv_indev_gesture_pinch_t` and `lv_indev_touch_data_t` are needed on the API level let's create an `lv_indev_gesture_private.h` for the other types and include it only in the C file. 

Or add the non API types in the C file.

kisvegabor:
I couldn't test it yet, but I checked the API in detail. I have a few smaller refactoring ideas to better follow LVGL naming conventions, but nothing major.

However, there is a larger thing: I wonder if we really need the concept of gesture recognizer in `lv_indev_t`. Note that our internal gesture recognizer algorithms are optional. It can happen that the driver already provides the gesture information, so we should provide only the high level information to the indev and keep the internal states hidden. So my notes:
- In `lv_indev_data_t` we can set an `lv_indev_gesture_type_t gesture_type` field and a `void * gesture_data`. `gesture_data` is interpreted according to the type and can be get in the event.  
- The gesture recognizer is really needed for `lv_indev_gesture_detect_pinch` to keep track of its state but it's used only in the driver.
- `lv_indev_gesture_state_t` can be tracked by LVGL as it knows when the gesture was set for the first time. 
- In `lv_indev_data_t` set `point.x/y` to the central point in the driver (if any) or to the primary point. It'd be more explicit.
- We should also compensate the scale thresholds. The user shouldn't see these internals. Instead when gesture becomes active (regardless of the threshold) we should start count the scale from there. So maybe internally we start recognizing the pinch only from 125% scale, but from the user's point of it's the 100% and 126% is the 101%.   
- I'm thinking about if we really need events for each gesture type, or a single `LV_EVENT_GESTURE` (where we can get the gesture type) could be enough. This way we can easily add `GESTURE_START/END` events as well, plus it could be backward compatible with the current gesture event.  

Ideally the scale event of the demo could be this simple:
```c
static void label_scale(lv_event_t *e)
{
   float scale = lv_event_get_pintch_scale(e); //If it's not a PINCH event return 1.0
   if(scale < 0.3) scale = 0.3;

    lv_point_t p;
    lv_indev_get_point(lv_indev_active(), &p);

    label_width = RECT_INIT_WIDTH * scale;
    label_height = RECT_INIT_WIDTH * scale;
    label_x = p.x - label_width / 2;
    label_y = p.y - label_height / 2;

    LV_LOG_TRACE("label scale: %g label x: %g label y: %g w: %g h: %g\n",
            scale, label_x, label_y, label_width, label_height);

    /* Update position and size */
    lv_style_set_width(&label_style, (int)label_width);
    lv_style_set_height(&label_style, (int)label_height);
    lv_style_set_x(&label_style, (int)label_x);
    lv_style_set_y(&label_style, (int)label_y);
}
```

What do you think?

etag4048:
@kisvegabor Thanks for the review, I agree about the separating the indev and the gesture recognizer and will refactor accordingly.

>     * I'm thinking about if we really need events for each gesture type, or a single `LV_EVENT_GESTURE` (where we can get the gesture type) could be enough. This way we can easily add `GESTURE_START/END` events as well, plus it could be backward compatible with the current gesture event.

I think it is useful to have an event per gesture, only for fact that the user can set an event gesture event callback of a specific type, on a specific object. Having a general callback for all gesture types would require the user to always check the type of the gesture.

Thanks for the help with sample program, fetching the gesture info and style updating is much cleaner in your proposal, however the original demo supported subsequent zooms. By using a static variable within the scope of the function, that was used to save the width/height of the rectangle just when the gesture was recognized, the calculation is performed on this initial value during the pinch gesture. At the end in the _ENDED branch it was set back to the sentinel values, ready for the next run. Using a constant means if the rectangle is zoomed once when it's zoomed again it won't take into account that it was already scaled previously (and fetching the width/height of lv_obj_t, results in it being scaled way too much).

kisvegabor:
> I think it is useful to have an event per gesture, only for fact that the user can set an event gesture event callback of a specific type, on a specific object.

I see and it makes sense. So we need to choose between 
1. compatibility with the current simple gesture processing (up/down/left/right). (`LV_EVENT_GESTURE`)
2. saving a some code for the user. (`LV_EVENT_GESTURE_PINCH/ROTATE/etc`)

I still believe we should choose 1) because backward compatibility and consistency is higher priority than saving 2-3 lines of code. Having `GESTURE_BEGIN/END` would be great too as it's similar to `SCROLL_BEGIN/END`.

So let's go for 1) now. 

> however the original demo supported subsequent zooms.

I see, however it's an example to show the major concept. We shouldn't overwhelm the users by shown too much. :slightly_smiling_face:

etag4048:
@kisvegabor No problems, let's go with a single gesture event for all. My plan included doing a nice demo program to demonstrate capabilities of this feature and also test every kind of gestures to see if the API works, and keeping the short snippet for the docs ? Is it ok for you ?

kisvegabor:
Sounds good! Typically the demos are something that can be shown on a fair. So just try to make the demo interesting and ascetic. :slightly_smiling_face:

etag4048:
@kisvegabor I've refactored the code to use `gesture_type` and `gesture_data` instead of the previous approach. However, I don't see how one would dispatch GESTURE_END/START if we don't have access to the state of the gesture inside of `lv_indev.c` I you have an idea please let me know. I we can obtain the state of the gesture via a function call to `lv_event_get_gesture_state`

src/indev/lv_indev_gesture.h
```diff
@@ -0,0 +1,167 @@
...
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+/*********************
+ *      INCLUDES
+ *********************/
+#include "../core/lv_obj.h"
+
```
kisvegabor, line 22: Please add 
```
#if LV_USE_GESTURE_RECOGNITION
```
here and in the other related files too

And on one place:
```
#if LV_USE_FLOAT == 0
#error "LV_USE_FLOAT is required for gesture detection."
#endif
```

src/indev/lv_indev_gesture.c
```diff
@@ -0,0 +1,602 @@
...
+    } else {
+        data->state = LV_INDEV_STATE_PRESSED;
+    }
+
+    switch (recognizer->state) {
+    case LV_INDEV_GESTURE_STATE_RECOGNIZED:
+        lv_indev_get_gesture_center_point(recognizer, &cur_pnt);
+        data->point.x = cur_pnt.x;
+        data->point.y = cur_pnt.y;
+        /* Fall through */
```
kisvegabor: `Fall through` caused warnings in some compilers.

src/indev/lv_indev_gesture.h
```diff
@@ -0,0 +1,167 @@
...
+/* General descriptor for a gesture, used by recognizer state machines to track 
+ * the scale, rotation, and translation NOTE: (this will likely become private) */
+typedef struct {
+
+    /* Motion descriptor, stores the coordinates and velocity of a contact point */
+    lv_indev_gesture_motion_t motions[LV_GESTURE_MAX_POINTS];
+
+    lv_point_t center;                  /* Center point */
+    float scale;                        /* Scale factor & previous scale factor */
```
kisvegabor: Separate the gesture detector logic/data from the generic gesture API. 

The generic gesture API can be part of LVGL unconditionally, we just need to enable/disable the internal gesture detector logic/data. (`LV_USE_GESTURE_RECOGNITION` and `LV_USE_FLOAT`).

src/indev/lv_indev_gesture.h
```diff
@@ -0,0 +1,167 @@
...
+ * GLOBAL PROTOTYPES
+ **********************/
+
+/**
+ * Detects a pinch gesture
+ * @param recognizer        pointer to a gesture recognizer
+ * @param touches           pointer to the first element of the collected touch events
+ * @param touch_cnt         length of passed touch event array.
+ */
+void lv_indev_gesture_detect_pinch(lv_indev_gesture_recognizer_t *recognizer, lv_indev_touch_data_t *touches, uint16_t touch_cnt);
```
kisvegabor: E.g. this part of the the internal gesture recognizer (which is optional), but e.g. `lv_indev_set_gesture_data` is a generic function (always available).

demos/gestures/lv_demo_gestures.c
```diff
@@ -0,0 +1,203 @@
...
+ * Copyright (c) 2024 EDGEMTech Ltd
+ *
+ * Author: EDGEMTech Ltd, Erik Tagirov (erik.tagirov@edgemtech.ch)
+ *
+ ******************************************************************/
+
+/*********************
+ *      INCLUDES
+ *********************/
+#include "lv_demo_gestures.h"
```
kisvegabor: Please make it an example (not a demo).

kisvegabor (commented):
It's shaping nicely!

> However, I don't see how one would dispatch GESTURE_END/START if we don't have access to the state of the gesture inside of lv_indev.c

I have few ideas, but leave it as it is for now. 

---

Before merging we need 2 more things:
- Make the CI pass
- Add some docs

etag4048:
@kisvegabor Ok will refactor accordingly, will be ready on monday.

etag4048:
@kisvegabor CI is fixed, I've also moved the demo to examples instead, I've started to create a private header. however I believe it's better to separate detection logic from the generic API once we removed the reliance on float types. Because the main structure uses floats.

lhdjply:
https://github.com/lvgl/lvgl/blob/a5dce5c48e48a7b3a680d82766e5385dd570cb24/examples/porting/lv_port_indev_template.c#L179-L215

Can you modify the `lv_port_indev_template.c` to support  multi touch gestures

kisvegabor (approved):
Thank you Erik! 

As we discussed during our call still there are few improvement opportunities:
- add more gesture detectors besides pinch: rotate, swipe
- API fine tuning
- evaluate integer only logic (low priority)

Anyway, I think it's a great start and complete enough to merge it. 

@liamHowatt please review it.

kisvegabor:
> Can you modify the lv_port_indev_template.c to support multi touch gestures

Good idea! @etag4048 is on a holiday now, and when he is back a follow up PR will be added to improve a few things. Let's handle it there as well.

src/indev/lv_indev_private.h
```diff
@@ -112,6 +112,9 @@ struct _lv_indev_t {
                                       here by the buttons*/
     lv_event_list_t event_list;
     lv_anim_t * scroll_throw_anim;
+
+    lv_indev_gesture_type_t gesture_type;
+    void * gesture_data;
```
liamHowatt, line 117: Probably this could be `lv_indev_gesture_recognizer_t *` since that's the only thing assigned to it but it's ok.

src/indev/lv_indev_gesture_private.h
```diff
@@ -0,0 +1,93 @@
...
+#endif
+
+/*********************
+ *      INCLUDES
+ *********************/
+#include "../core/lv_obj.h"
+
+#if LV_USE_GESTURE_RECOGNITION
+
+/*********************
```
liamHowatt, line 28: Should include lv_indev_gesture.h

src/drivers/wayland/lv_wayland.c
```diff
@@ -117,11 +116,12 @@ struct input {
...
-        uint32_t x;
-        uint32_t y;
-        lv_indev_state_t state;
-    } touch;
+#if LV_USE_GESTURE_RECOGNITION
+    lv_indev_touch_data_t touches[10];
+    uint8_t touch_event_cnt;
+    uint8_t primary_id;
+    lv_indev_gesture_recognizer_t recognizer;
+#endif
```
liamHowatt, line 124: Does Wayland normal touch still work with `LV_USE_GESTURE_RECOGNITION` disabled? Or does Wayland depend on it being enabled? If it's required, then all the guards could go away and a single `#if !LV_USE_GESTURE_RECOGNITION ... #error ... #endif` could go at the top.

liamHowatt (approved):
High level it looks good. We can address the things I mentioned later.

PR #7100, opened 2024-10-17
feat(font): support 8 bpp font bitmaps

liamHowatt:
1, 2, and 4 bpp are already there. Simply add another `else if` branch for 8 bpp. I did not use memcpy because the length will usually be so short the overhead of the function call(s) may be not be worthwhile. I'm open to a suggestion to use memcpy though.

### Notes

### DIFF
--- a/src/font/lv_font_fmt_txt.c
+++ b/src/font/lv_font_fmt_txt.c
@@ -146,6 +146,15 @@ const void * lv_font_get_bitmap_fmt_txt(lv_font_glyph_dsc_t * g_dsc, lv_draw_buf
                 bitmap_out_tmp += stride;
             }
         }
+        else if(fdsc->bpp == 8) {
+            for(y = 0; y < gdsc->box_h; y ++) {
+                for(x = 0; x < gdsc->box_w; x++, i++) {
+                    bitmap_out_tmp[x] = *bitmap_in;
+                    bitmap_in++;
+                }
+                bitmap_out_tmp += stride;
+            }
+        }
         return draw_buf;
     }
     /*Handle compressed bitmap*/


src/font/lv_font_fmt_txt.c
```diff
@@ -146,6 +146,15 @@ const void * lv_font_get_bitmap_fmt_txt(lv_font_glyph_dsc_t * g_dsc, lv_draw_buf
                 bitmap_out_tmp += stride;
             }
         }
+        else if(fdsc->bpp == 8) {
+            for(y = 0; y < gdsc->box_h; y ++) {
+                for(x = 0; x < gdsc->box_w; x++, i++) {
+                    bitmap_out_tmp[x] = *bitmap_in;
+                    bitmap_in++;
+                }
+                bitmap_out_tmp += stride;
```
FASTSHIFT, line 155:
```c
lv_memcpy(bitmap_out_tmp, bitmap_in, gdsc->box_w);
bitmap_in += gdsc->box_w;
bitmap_out_tmp += stride;
```

kisvegabor: It might not be worth to call memcpy as the fonts are usually small.

FASTSHIFT: > It might not be worth to call memcpy as the fonts are usually small.

If the font's original bitmap can meet the alignment requirements (such as software rendering), can we skip the copy and pass the pointer directly?

kisvegabor (approved):

src/font/lv_font_fmt_txt.c
kisvegabor: We needed to replace the pointer in the `draw_buf` parameter, but that buffer is allocated and we can't just overwrite it. I'd marge as a quick fix and improve it later.  We need to touch fonts anyway to support A1,A2,A4 rendering as well (i.e. don't convert everything to A8 by default)

XuNeo (approved):

src/font/lv_font_fmt_txt.c
liamHowatt: > If the font's original bitmap can meet the alignment requirements (such as software rendering), can we skip the copy and pass the pointer directly?

kisvegabor did it in #7234

PR #7108, opened 2024-10-18
feat(label): add lv_label_set_max_lines()

niklasf:
Adds `lv_label_set_max_lines()`, similar to `lv_spangroup_set_max_lines()`.

This is especially useful in the scenario shown in the test case: The label should be centered vertically, so height needs to be as small as possible (`LV_SIZE_CONTENT`). Therefore setting a fixed height is not enough, and text metrics need to be taken into account.

Implementation is fairly straight-forward after #7001.


liamHowatt (approved):
Thank you!

kisvegabor:
Hi,

Thank you for opening this PR!

I have a slightly different idea in mind. What about introducing `LV_EM(num)` similar to `LV_PCT()`? `em` is a CSS unit and 1em means 1 line height. So it depend on the font size. With this we could do:
```c
lv_obj_set_height(label, LV_EM(3)); 
lv_obj_set_style_max_height(label, LV_EM(3), 0); //3 line height
```

What do you think?

niklasf:
Generally I agree that having `LV_EM(...)` would be awesome.

For this use case it might be awkward to use. I think line height is not nescessarily 1 em, and there's also line space that has to be added, and it might be in pixels.

kisvegabor:
I think it's fine to consider line_space too. E.g.  if
- font line_height is 10
- style line space is 5

`LV_EM(3)` means  `3 * (10 + 5) - 5 = 40` (remove the line space after the last line)

What do you think?

kisvegabor:
@niklasf Are you still interested in this issue/PR?

niklasf:
Yes, still very much interested.

It's tricky, though ...

I don't think `LV_EM(...)` should automatically include line spacing. That would unexpected, because:
* It differs from the usual meaning of *em* in typography.
* It differs from *em* in CSS.
* `LV_EM(...)` would be non-linear, e.g., `LV_EM(2)` would not be twice as large as `LV_EM(1)`.

In CSS there's a good solution: There's a different unit `lh` ("line height"). It's linear because there's no `line-spacing` between lines, only `line-height`, and text is centered within each line.

But removing `text_line_space` and adding `text_line_height` instead would be a breaking change in LVGL.

kisvegabor:
Great! :slightly_smiling_face: 

It seems using `line_space` instead of `line_height` is LVGL was a mistake. :slightly_frowning_face: 

Maybe we can use a custom name to not conflict with CSS and typography standards. E.g. `lv_lineunit`. What  do you think?

niklasf:
How about these steps?

1. Let label respect max height.
2. Introduce new units LV_EM (and maybe LV_LH).
3. Introduce line height style property that behaves like line-height in CSS. Line space remains available for now.

And then to limit maximum lines, users can just avoid using line space and work only with line height.

These features should work nicely together, but they can be considered/implemented individually. Will try to make progress on this over the coming weeks.

Abandoning this PR.

niklasf:
X-Post: Following up in #9646

PR #7119, opened 2024-10-19
feat(ffmpeg): add playback complete event trigger

ericmaryea:
I had a need to trigger an even when video playback completed. The comment for the `LV_EVENT_READY` event indicated that it fit this purpose, but I'm open to suggestions this warrants a different/new event code.

### DIFF
--- a/docs/details/libs/ffmpeg.rst
+++ b/docs/details/libs/ffmpeg.rst
@@ -40,6 +40,13 @@ See the examples below.
 
 .. _ffmpeg_example:
 
+Events
+------
+
+- :cpp:enumerator:`LV_EVENT_READY` Sent when playback is complete and auto-restart is not enabled.
+
+Learn more about :ref:`events`.
+
 Example
 -------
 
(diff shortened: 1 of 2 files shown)


FASTSHIFT (dismissed):

kissa96 (commented):
Please update the corresponding documentation as well. We could also implement an LV_EVENT_DONE for these kind of events

ericmaryea:
@kissa96 I went back and did some looking through where `LV_EVENT_READY` is used elsewhere and it's only a handful of spots so far, but I think most crucially it's used in the deprecated rlottie player for the same purpose I'm using it for here: https://github.com/lvgl/lvgl/blob/84b28ff688b9b8b33682191c37554194d5ae1950/src/libs/rlottie/lv_rlottie.c#L204-L229

So maybe a new event might be wanted at some point to separate from other usages (like dropdowns and file lists being ready), but given that the current comment for `LV_EVENT_READY` is `A process has finished` I don't want to make design decisions for the project/team about what would be used when. That's not to say that that work couldn't be done in this PR, but to me feels like it could lead to things that I would consider out of scope for this work.

kissa96:
All right. I have not seen this rlottie implementation so far, but it seems to me LV_EVENT_READY means multiple things currently. Thanks for the doc update!

kisvegabor (approved):
Looks good, thank you! Using `LV_EVENT_READY` was a great idea! :+1:

kissa96 (approved):

liamHowatt (approved):

PR #7127, opened 2024-10-21
fix(cmake): generate lvgl.pc in CMAKE_CURRENT_BINARY_DIR

niklasf:
Fixes file configured at `${CMAKE_BINARY_DIR}/lvgl.pc` but later installed from

```cmake
install(
  FILES "${CMAKE_CURRENT_BINARY_DIR}/lvgl.pc" # <--
  DESTINATION "${LIB_INSTALL_DIR}/pkgconfig/")
```

leading to

```
CMake Error at ThirdParty/_lvgl/cmake_install.cmake:58 (file):
  file INSTALL cannot find
  "/_build/<subdir>/lvgl/lvgl.pc": No such file or
  directory.
  cmake_install.cmake:47 (include)
```

in multi-directory projects.

Originally introduced via 559618d9d2c66069f61ba673b7bce6bf3e7eb036. Works for me, now, but did I choose the right way to fix this?

### DIFF
--- a/env_support/cmake/custom.cmake
+++ b/env_support/cmake/custom.cmake
@@ -133,8 +133,8 @@ if(NOT LV_CONF_BUILD_DISABLE_DEMOS)
 endif()
 
 
-configure_file("${LVGL_ROOT_DIR}/lvgl.pc.in" ${CMAKE_BINARY_DIR}/lvgl.pc @ONLY)
-configure_file("${LVGL_ROOT_DIR}/lv_version.h.in" ${CMAKE_BINARY_DIR}/lv_version.h @ONLY)
+configure_file("${LVGL_ROOT_DIR}/lvgl.pc.in" ${CMAKE_CURRENT_BINARY_DIR}/lvgl.pc @ONLY)
+configure_file("${LVGL_ROOT_DIR}/lv_version.h.in" ${CMAKE_CURRENT_BINARY_DIR}/lv_version.h @ONLY)
 
 install(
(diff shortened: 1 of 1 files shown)


kissa96 (approved):
LGTM

liamHowatt (commented):
Looks fine. Any reason not to change the other `CMAKE_BINARY_DIR`s to `CMAKE_CURRENT_BINARY_DIR`s?

kisvegabor:
cc @etag4048

etag4048:
In #7161 I've used CMAKE_BINARY_DIR for install directive that installs the pkgconfig, I've tested using CMAKE_CURRENT_BINARY_DIR with this patch and it works.

niklasf:
> Any reason not to change the other `CMAKE_BINARY_DIR`s to `CMAKE_CURRENT_BINARY_DIR`s?

Not sure. Maybe the intent there is to add files to a gloabal `lib`/`bin`/... directory?

etag4048:
Currently the library .a archive files are generated in the ``lib`` directory at the root of the cmake build folder. When installed by default, they end in ``/usr/local/lib`` . I believe we can also generate them in ``<buildir>/lvgl/lib`` and install them to ``/usr/local/lib/`` or ``/usr/local/lib/lvgl`` but for the latter ``pc.in`` file has to be adjusted. 

I believe generating in ``<builddir>/lvgl/lib`` is cleaner, I can add a commit to PR #7161 once this is merged

liamHowatt (approved):

kisvegabor (approved):

Issue #4328, opened 2023-06-30
Press an object obj1 and slide to another object obj1, obj1 received LV_EVENT_CLICKED


MiltonWuSZ:
### About the question
My input device is touchscreen，lvgl version 8.3.7
I create two object `obj1` and `obj2`. I pressed  `obj1` and slide to `obj2`, `obj1` received  event `LV_EVENT_CLICKED`.
My expectation is that no click event occurs, in this scenario, the user cancels the click.
###  Examples and cases
```c
static void event_cb(lv_event_t * e)
{
    if (e->code == LV_EVENT_CLICKED) {
        printf("obj1 clicked\n");
    }
}
static void event2_cb(lv_event_t * e)
{

    if (e->code == LV_EVENT_CLICKED) {
        printf("obj2 clicked\n");
    }
}
void obj_event_test(void)
    lv_obj_t * content = lv_obj_create(lv_scr_act());
    lv_obj_remove_style_all(content);
    lv_obj_set_flex_flow(content, LV_FLEX_FLOW_COLUMN);
    lv_obj_set_size(content, 300,600);
    lv_obj_set_align(content, LV_ALIGN_CENTER);
    lv_obj_set_flex_align(content, LV_FLEX_ALIGN_CENTER, LV_FLEX_ALIGN_CENTER, LV_FLEX_ALIGN_CENTER);
    
    lv_obj_t * obj1 = lv_obj_create(content);
    lv_obj_add_flag(obj1, LV_OBJ_FLAG_CLICKABLE);  
    lv_obj_set_style_bg_color(obj1,lv_color_hex(0xff0000), 0);
    lv_obj_set_style_bg_opa(obj1,255, 0);
    lv_obj_set_size(obj1, 100,100);
    lv_obj_t * txt = lv_label_create(obj1);
    lv_label_set_text(txt, "obj1");  
    lv_obj_add_event_cb(obj1, event_cb, LV_EVENT_CLICKED, NULL);
    
    lv_obj_t * obj2 = lv_obj_create(content);
    lv_obj_add_flag(obj2, LV_OBJ_FLAG_CLICKABLE);  
    lv_obj_set_style_bg_opa(obj2,255, 0);
    lv_obj_set_size(obj2, 100,100);
    lv_obj_t * txt2 = lv_label_create(obj2);
    lv_label_set_text(txt2, "obj2");  
    lv_obj_add_event_cb(obj2, event2_cb, LV_EVENT_CLICKED, NULL);
}
[...]

PR #7138, opened 2024-10-22
fix(indev): keep pointer movement even if previous object has been lost duo to missing press lock

Viatorus:
I think this change here https://github.com/lvgl/lvgl/commit/ecf1ea4d37833cd1b7b3f5d9fb6fabc5acb9850b (from https://github.com/lvgl/lvgl/issues/4328) went in a good direction to not fire LV_EVENT_CLICKED.

But IMHO, one thing is missing. Since there was already an active object, the new position should still be recognized as a movement of the pointer rather than the start of a new movement.

We have a case where a scroll event isn’t recognized because the press event slips from one small object to another, and so on.

### DIFF
--- a/src/indev/lv_indev.c
+++ b/src/indev/lv_indev.c
@@ -1197,8 +1197,11 @@ static void indev_proc_press(lv_indev_t * indev)
 
     /*If a new object was found reset some variables and send a pressed event handler*/
     if(indev_obj_act != indev->pointer.act_obj) {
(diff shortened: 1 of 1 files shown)


Viatorus:
@kisvegabor Maybe you could take a look on this PR also. ;)

kisvegabor:
Ah, I'm sorry. I don't how I missed this PR for 3 weeks. :disappointed: 

I'm sure you are right, but I can't see how this change helps. Could you send a code snippet to reproduce the issue on my end?

PR #7143, opened 2024-10-23
feat(roller): set roller option with a string

Orange-Murker:
This adds the ability to manually set an option in a roller using a string that corresponds to that option.

### Notes

### DIFF
--- a/src/stdlib/builtin/lv_string_builtin.c
+++ b/src/stdlib/builtin/lv_string_builtin.c
@@ -232,6 +232,16 @@ int lv_strcmp(const char * s1, const char * s2)
     return *(const unsigned char *)s1 - *(const unsigned char *)s2;
 }
 
+int lv_strncmp(const char * s1, const char * s2, size_t len)
+{
+    while(len > 0 && *s1 && (*s1 == *s2)) {
+        s1++;
+        s2++;
+        len--;
+    }
+    return *(const unsigned char *)s1 - *(const unsigned char *)s2;
+}
+
 char * lv_strdup(const char * src)
 {
     size_t len = lv_strlen(src) + 1;
--- a/src/stdlib/builtin/lv_string_builtin.c
+++ b/src/stdlib/builtin/lv_string_builtin.c
@@ -235,9 +235,11 @@ int lv_strcmp(const char * s1, const char * s2)
 int lv_strncmp(const char * s1, const char * s2, size_t len)
 {
     while(len > 0 && *s1 && (*s1 == *s2)) {
+        if(--len == 0) {
+            return 0;
+        }
         s1++;
         s2++;
-        len--;
     }
     return *(const unsigned char *)s1 - *(const unsigned char *)s2;
 }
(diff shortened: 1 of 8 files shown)


src/widgets/roller/lv_roller.c
```diff
@@ -275,6 +275,34 @@ void lv_roller_get_selected_str(const lv_obj_t * obj, char * buf, uint32_t buf_s
...
+    size_t options_len = lv_strlen(options);
+
+    bool option_found = false;
+
+    uint32_t current_option = 0;
+    size_t line_start = 0;
+
+    for(size_t i = 0; i < options_len; i++) {
+        if(options[i] == '\n') {
+            // See if this is the correct option
```
kisvegabor:
```suggestion
            /* See if this is the correct option */
```

kisvegabor (commented):
Great feature! Could you also add a [test](https://github.com/lvgl/lvgl/tree/master/tests) for it?

Orange-Murker:
> Great feature! Could you also add a [test](https://github.com/lvgl/lvgl/tree/master/tests) for it?

Sure thing. I will take a look at it

src/stdlib/builtin/lv_string_builtin.c
```diff
@@ -232,6 +232,18 @@ int lv_strcmp(const char * s1, const char * s2)
     return *(const unsigned char *)s1 - *(const unsigned char *)s2;
 }
 
+int lv_strncmp(const char * s1, const char * s2, size_t len)
+{
+    while(len > 0 && *s1 && (*s1 == *s2)) {
```
kissa96, line 241: This function assumes that we want to compare s1 to s2 and not the other way. Please indicate this or handle corner cases.

src/stdlib/builtin/lv_string_builtin.c
```diff
@@ -232,6 +232,18 @@ int lv_strcmp(const char * s1, const char * s2)
...
+        if(--len == 0) {
+            return 0;
+        }
+        s1++;
+        s2++;
+    }
+    return *(const unsigned char *)s1 - *(const unsigned char *)s2;
```
liamHowatt, line 248: Handle the case where `len` is `0`. Neither pointer is safe to dereference in that case.

liamHowatt (commented):
Nice!

src/stdlib/builtin/lv_string_builtin.c
Orange-Murker: There are 3 different cases:
1. The length of s1 is equal to the length of s2
2. The length of s1 is less than the length of s2
3. The length of s1 is greater than the length of s2

In the first case the while loop will break at then end of both strings because `*s1` will be false when we reach `\0`
In the second case the comparison will stop again when we reach the null terminator.
In the third case the comparison will stop because the last character of s2 (`\0`) is different from the nonzero character in s1

The comparison will stop anyways if `len` is shorter than the length of the shortest string.

Am I misunderstanding something?

src/stdlib/builtin/lv_string_builtin.c
Orange-Murker: Good catch. Will fix that now.

kisvegabor (approved):

liamHowatt (approved):

liamHowatt:
Thank you!

Issue #6478, opened 2024-07-05
custom.cmake cannot install lvgl.pc


paulh002:
### LVGL version

v8.4

### What happened?

When I issue sudo make install on raspberry pi I get the error 
`CMake Error at _deps/lvgl-build/cmake_install.cmake:70 (file):
  file INSTALL cannot find "/home/pi/sdrberry/build/lvgl.pc": No such file or
  directory.
Call Stack (most recent call first):
  cmake_install.cmake:47 (include)
`
I don't use this file and also don't know why it is necessary, can it be removed? Or issue fixed. Because cmake will stop and don't install the application.

### How to reproduce?
In CMakeList.txt I fetch lvgl with 
`FetchContent_Declare(lvgl GIT_REPOSITORY  https://github.com/lvgl/lvgl.git GIT_TAG v8.4.0)`
Install with:
`sudo make install`

kisvegabor:
It works for me:
```
cd lvgl 
mkdir build
cd build
cmake ..
make -j
sudo make install
```

Could you test it?

paulh002:
This is not the problem. The make install fails when it is part of an application cmake file and lvgl is included using FetchContent_Declare. The path for lvgl.pc is not correct in that case

kisvegabor:
Please a minimal CMake file file which uses `FetchContent_Declare` to reproduce the issue.

paulh002:
I made a small project which demonstrates the issue.  
You can clone it from https://github.com/paulh002/cmaketest  

to build it:  
git clone https://github.com/paulh002/cmaketest  
cd cmaketest
mkdir build
cd build
cmake ..
make
sudo make install


Result will be:
pi@pi5:~/cmaketest/build $ sudo make install
[ 45%] Built target lvgl
[ 45%] Built target cmaketest
[ 85%] Built target lvgl_examples
[100%] Built target lvgl_demos
Install the project...
-- Install configuration: ""
-- Installing: /usr/local/include/lvgl/../lv_conf.h
CMake Error at _deps/lvgl-build/cmake_install.cmake:70 (file):
  file INSTALL cannot find "/home/pi/cmaketest/build/lvgl.pc": No such file
  or directory.
Call Stack (most recent call first):
  cmake_install.cmake:47 (include)


make: *** [Makefile:100: install] Error 1

paulh002:
[...]

PR #7161, opened 2024-10-27
fix(cmake): fix installation with cmake --install

etag4048:
fixes issue #6787 #6478

The ``DESTINATION`` arg of the ``install`` directives located in ``env_support/cmake/custom.cmake`` being hardcoded to the default value of ``CMAKE_INSTALL_PREFIX`` resulted in a broken ``cmake_install.cmake`` file

If the user wants to use cmake to install lvgl headers and libraries to a custom location it can be done
at installation time with the following command ``cmake --install build_dir --prefix /usr/custom/``

Please note that currently ``lvgl.pc`` is not generated properly when ``--prefix`` is used.
There is an entire discussion on this subject in the cmake forums, the workaround is to set CMAKE_INSTALL_PREFIX at configuration time using the ``-D`` argument

https://discourse.cmake.org/t/how-to-generate-pc-pkg-config-file-supporting-prefix-of-the-cmake-install/4109


kisvegabor:
#7127 is merged. Is it ready to merge too?

etag4048:
@kisvegabor I will attempt to add the missing commit before the Thursday evening, now busy with other things.

kisvegabor:
Let us know when it's ready to review.

etag4048:
@kisvegabor Ready.

kisvegabor (approved):
It's hard to tell for me if it's really better this way, but I trust your expertise :slightly_smiling_face:

etag4048:
@kisvegabor It works with or without specifying a prefix, I've also removed the paths that are destined for executables. Since lvgl is a library, not an executable.

liamHowatt (approved):

PR #7181, opened 2024-10-29
fix(file_explorer): fix navigation when using a keypad indev

Orange-Murker:
It used to be the case that the file explorer entry would open on the `LV_EVENT_VALUE_CHANGED` event which works fine on a touchscreen, but is unusable on a keypad because that event is sent every time a new items is selected.
That has now been replaced by opening the entry on `LV_EVENT_CLICKED` which works on a keypad as well as on a touchscreen.

Another issue that using a keypad introduces was the fact that the table would scroll to the second column where the file kind was stored. This has been replaced by storing any relevant metadata in the cell itself using user data instead of adding unnecessary cells to the table.

In addition to that, this PR fixes the concerns raised in #7124 about ending up with a broken path if the directory fails to open.

### Notes


src/others/file_explorer/lv_file_explorer.c
```diff
@@ -690,12 +729,13 @@ static void sort_by_file_kind(lv_obj_t * tb, int16_t lo, int16_t hi)
...
+    // const char * v = lv_table_get_cell_value(tb, lo, 1);
+    lv_file_explorer_file_table_entry_data_t * v = lv_table_get_cell_user_data(tb, lo, 0);
     while(i <= gt) {
-        int strcmp_result = lv_strcmp(lv_table_get_cell_value(tb, i, 1), v);
-        if(strcmp_result < 0)
+        lv_file_explorer_file_table_entry_data_t * x = lv_table_get_cell_user_data(tb, i, 0);
+        if(x->file_kind < v->file_kind)
             exch_table_item(tb, lt++, i++);
-        else if(strcmp_result > 0)
+        else if(x->file_kind > v->file_kind)
```
liamHowatt: Just a question: are we not sorting by name anymore? Is it feasible to sort by both type and name?

src/others/file_explorer/lv_file_explorer.c
```diff
@@ -525,10 +534,17 @@ static void browser_file_event_handler(lv_event_t * e)
             else { /* Nothing to do*/ }
         }
 
-        lv_fs_dir_t dir;
-        if(lv_fs_dir_open(&dir, file_name) == LV_FS_RES_OK) {
-            lv_fs_dir_close(&dir);
-            show_dir(obj, (char *)file_name);
+        // Navigate to the file path if this is a directory
```
liamHowatt:
```suggestion
        /* Navigate to the file path if this is a directory */
```

liamHowatt (commented):
Wow, it really was unusable with a keypad.

I tested it and I found an issue with pointer indev usage.

```c
lv_table_get_selected_cell(explorer->file_table, &row, &col);
```

This is now running for the event `LV_EVENT_CLICKED` in lv_file_explorer.c but lv_table.c sets the selected cell col and row on `LV_EVENT_PRESSED` and sets col and row it back to `LV_TABLE_CELL_NONE` on `LV_EVENT_RELEASED`. Next, the `LV_EVENT_CLICKED` is sent and the file explorer gets `LV_TABLE_CELL_NONE` as the clicked row and col.

src/others/file_explorer/lv_file_explorer.c
Orange-Murker: My code sorts by the file type which is the same behavior as the original code. I could add sorting by name in a separate PR

liamHowatt:
> wouldn't that only happen if the indev is a pointer or a button

Yes, and we should keep supporting pointers and buttons :slightly_smiling_face:

You can see if you add a log here and click a cell with a pointer, it will always print `65535 65535` now.

```diff
diff --git a/src/others/file_explorer/lv_file_explorer.c b/src/others/file_explorer/lv_file_explorer.c
index 0592b5510..91c777cb2 100644
--- a/src/others/file_explorer/lv_file_explorer.c
+++ b/src/others/file_explorer/lv_file_explorer.c
@@ -496,6 +496,7 @@ static void browser_file_event_handler(lv_event_t * e)
 
         lv_memzero(file_name, sizeof(file_name));
         lv_table_get_selected_cell(explorer->file_table, &row, &col);
+        LV_LOG_USER("%" PRIu32 " %" PRIu32, row, col);
         selected_text = lv_table_get_cell_value(explorer->file_table, row, col);
         file_entry_user_data = lv_table_get_cell_user_data(explorer->file_table, row, col);

```

src/others/file_explorer/lv_file_explorer.c
liamHowatt: Oh I see, before it was sorting by file kind and name because the icon is part of the string.

It used to sort by name and now it doesn't, though. Is that right? The `lv_strcmp` compared the full entry name, not just the type.

Orange-Murker: The files were sorted on the second column of the table(index 1) in the old code:
```c
int strcmp_result = lv_strcmp(lv_table_get_cell_value(tb, i, 1), v);
```
File names were stored in index 0 and file kinds in index 1 so the files were indeed sorted by file kind.

Orange-Murker:
> > wouldn't that only happen if the indev is a pointer or a button
> 
> Yes, and we should keep supporting pointers and buttons 🙂
> 
> You can see if you add a log here and click a cell with a pointer, it will always print `65535 65535` now.
> 
> ```diff
> diff --git a/src/others/file_explorer/lv_file_explorer.c b/src/others/file_explorer/lv_file_explorer.c
> index 0592b5510..91c777cb2 100644
> --- a/src/others/file_explorer/lv_file_explorer.c
> +++ b/src/others/file_explorer/lv_file_explorer.c
> @@ -496,6 +496,7 @@ static void browser_file_event_handler(lv_event_t * e)
>  
>          lv_memzero(file_name, sizeof(file_name));
>          lv_table_get_selected_cell(explorer->file_table, &row, &col);
> +        LV_LOG_USER("%" PRIu32 " %" PRIu32, row, col);
>          selected_text = lv_table_get_cell_value(explorer->file_table, row, col);
>          file_entry_user_data = lv_table_get_cell_user_data(explorer->file_table, row, col);
> ```

Very valid point. I will work on resolving this issue.

src/others/file_explorer/lv_file_explorer_private.h
```diff
@@ -51,6 +51,18 @@ struct _lv_file_explorer_t {
     lv_file_explorer_sort_t sort;
 };
 
+typedef enum {
+    FILE_KIND_DIR,
+    FILE_KIND_IMAGE,
+    FILE_KIND_AUDIO,
+    FILE_KIND_VIDEO,
+    FILE_KIND_FILE
```
liamHowatt: These should be namespaced a bit more.

```suggestion
    LV_FILE_EXPLORER_FILE_KIND_DIR,
    LV_FILE_EXPLORER_FILE_KIND_IMAGE,
    LV_FILE_EXPLORER_FILE_KIND_AUDIO,
    LV_FILE_EXPLORER_FILE_KIND_VIDEO,
    LV_FILE_EXPLORER_FILE_KIND_FILE
```

src/others/file_explorer/lv_file_explorer.c
```diff
@@ -686,12 +724,13 @@ static void sort_by_file_kind(lv_obj_t * tb, int16_t lo, int16_t hi)
     int16_t lt = lo;
     int16_t i = lo + 1;
     int16_t gt = hi;
-    const char * v = lv_table_get_cell_value(tb, lo, 1);
+    // const char * v = lv_table_get_cell_value(tb, lo, 1);
```
liamHowatt:
```suggestion
```

src/others/file_explorer/lv_file_explorer.c
liamHowatt: > File names were stored in index 0 and file kinds in index 1 so the files were indeed sorted by file kind.

Ah I see, got it.

src/others/file_explorer/lv_file_explorer.c
```diff
@@ -644,17 +669,30 @@ static void strip_ext(char * dir)
...
 {
     if(i == j) return;
+    const char * i_value = lv_table_get_cell_value(tb, i, 0);
+
+    char * tmp_i_value = lv_malloc(lv_strlen(i_value) + 1);
+    LV_ASSERT_MALLOC(tmp_i_value);
+    lv_strcpy(tmp_i_value, i_value);
+
+    //*Will be freed when the table is freed. The previous user data is freed when the new one is set*/
+    lv_file_explorer_file_table_entry_data_t * tmp_i_user_data = lv_malloc(sizeof(
```
liamHowatt, line 679: I couldn't find where it's freed. Is it?

liamHowatt (commented):
Sorry for the delay.

The fix makes sense, thanks.

An edge case use-after-free issue with my test directory I'm seeing is also present in `master` also so I won't worry about it here.

src/others/file_explorer/lv_file_explorer.c
Orange-Murker: Inside `lv_table.c`:

The old data is freed when setting new data by using `lv_table_set_cell_user_data`:
```c
if(table->cell_data[cell]->user_data) {
    lv_free(table->cell_data[cell]->user_data);
}
```

And the cell user data is freed by the table destructor

liamHowatt: Oh I see. I missed it because I wasn't expecting the table widget to have the responsibility of calling `lv_free` on a user data field :sweat_smile:

src/others/file_explorer/lv_file_explorer.c
```diff
@@ -522,10 +535,17 @@ static void browser_file_event_handler(lv_event_t * e)
...
-        if(lv_fs_dir_open(&dir, file_name) == LV_FS_RES_OK) {
-            lv_fs_dir_close(&dir);
-            show_dir(obj, (char *)file_name);
+        /* Navigate to the file path if this is a directory */
+        if(file_entry_user_data->file_kind == LV_FILE_EXPLORER_FILE_KIND_DIR) {
+            lv_fs_dir_t dir;
+            lv_fs_res_t res = lv_fs_dir_open(&dir, file_name);
+            if(res == LV_FS_RES_OK) {
+                lv_fs_dir_close(&dir);
+                show_dir(obj, (char *)file_name);
```
liamHowatt, line 544: I was testing and found that I was relying on some unreliable behavior. The filesystem drivers currently cannot tell if a symbolic link is a directory, so the icon becomes `FILE_KIND_FILE` but it was ok because the highlighted logic would always try to open the entry as a dir first, and if it fails, it's a file, so I could still open symlink directories. But after this change, it can no longer. I think I may open a PR for `lv_fs_posix` to recognize symlink directories as directories.

liamHowatt (approved):
Thanks! This is an overall nice refactor besides fixing keypad navigation.

src/others/file_explorer/lv_file_explorer.c
Orange-Murker: I think that it is a good idea to fix this on the fs driver level

liamHowatt: I agree

liamHowatt:
Not stale

liamHowatt:
@uLipe @kisvegabor

kisvegabor (approved):
I've checked it from the high level point of view, as @liamHowatt already checked the details.

Issue #7157, opened 2024-10-25
bug - possible dereference of NULL pointer from lv_subject_add_observer()


pfvogel:
### LVGL version

9.3.0

### What happened?

I ran static analysis tool (CodeChecker) that identified a possible NULL pointer dereference.

In src/others/observer/lv_observer.c:

lv_subject_add_observer() will pass NULL as the obj value to lv_subject_add_observer_obj(). 
 In that routine, obj is assigned to target, and then later observer->cb() will attempt to dereference target.

```
    observer->target = obj;
    observer->for_obj = 1;
    /* subscribe to delete event of the object */
    if(obj != NULL) {
        lv_obj_add_event_cb(obj, unsubscribe_on_delete_cb, LV_EVENT_DELETE, observer);
    }

    /* update object immediately */
    if(observer->cb) observer->cb(observer, subject);
```

`A possible fix would be to move the call inside the test to see if the obj is NULL.  The "for_obj" should probably also be set to 0 to indicate that target is not valid.

```
    observer->target = obj;
    observer->for_obj = 1;
    /* subscribe to delete event of the object */
    if(obj != NULL) {
        observer->for_obj = 0;

        lv_obj_add_event_cb(obj, unsubscribe_on_delete_cb, LV_EVENT_DELETE, observer);

       /* update object immediately */
       if(observer->cb) observer->cb(observer, subject);
    }
```

### How to reproduce?

N/A -- issue found through inspection

liamHowatt:
Hi, good find. Do you have a callback as an example which dereferences `target`? My first impression is that users of `lv_subject_add_observer` should know there isn't a target involved and would not try to access `target`.

pfvogel:
Yes, I think it would be reasonable to expect a user to know that the target is null and not dereference it, but I think that might be easy to overlook.  There are instances of this being done within the same source file (lv_observer.c). The lv_obj_bind_flag_if_eq() routine calls bind_to_bitfield() which creates a observer, passing the obj_flag_observer_cb() as the callback.  The callback calls the either the lv_obj_add_flag() or lv_obj_remove_flag() routine.  The add/remove flag routine() will assert when the obj is checked and found to be NULL.

It looks like there are cases where the target of the observer is expected to a pointer to an object (the "for_obj" flag is true).  Maybe part of the change would be to only allow invoking the callback with a NULL target would be when the for_obj flag is false?

pfvogel:
[...]

PR #7183, opened 2024-10-29
feat(observer): add null pointer check

pfvogel:
Checks were added to the callbacks used in the lv_observer.c file to make sure that the subject and object pointers were not NULL prior to being dereferenced.


liamHowatt:
Fixes #7157

> I would say that this is exactly what `LV_USE_ASSERT_NULL` was made for. You could do
> 
> ```c
> LV_USE_ASSERT_NULL(subject);
> LV_USE_ASSERT_NULL(obj);
> ```
> 
> What do you think?

src/others/observer/lv_observer.c
```diff
@@ -452,6 +452,11 @@ lv_observer_t * lv_obj_bind_checked(lv_obj_t * obj, lv_subject_t * subject)
 #if LV_USE_LABEL
 lv_observer_t * lv_label_bind_text(lv_obj_t * obj, lv_subject_t * subject, const char * fmt)
 {
+    if((subject == NULL) || (obj == NULL)) {
+        LV_LOG_WARN("NULL pointer");
+        return NULL;
```
FASTSHIFT: It is recommended to use `LV_ASSERT_NULL`

FASTSHIFT (changes requested):

src/others/observer/lv_observer.c
pfvogel: I didn't initially use the LV_ASSERT_NULL macro because it looks like the routines are allowed to return NULL and the macro will block execution instead of returning NULL. I can see where it might be a matter of preference on whether the routine blocked when it received invalid input or just returned NULL. Since there were already other checks in the routine that returned NULL, I implemented it to also return NULL. But I think using the macro would be reasonable as well.

pfvogel:
I replaced the tests for NULL with the LV_ASSERT_NULL macro

liamHowatt (commented):
Thanks. I notice you're only checking `obj` now. Is that ok? Did your code analysis tool in #7157 only identify `obj` as being at risk of being `NULL`?

src/others/observer/lv_observer.c
```diff
@@ -452,6 +452,8 @@ lv_observer_t * lv_obj_bind_checked(lv_obj_t * obj, lv_subject_t * subject)
 #if LV_USE_LABEL
 lv_observer_t * lv_label_bind_text(lv_obj_t * obj, lv_subject_t * subject, const char * fmt)
 {
+    LV_ASSERT_NULL(obj);
```
kisvegabor, line 456: I suggest
```suggestion
    LV_ASSERT_NULL(obj);
    LV_ASSERT_NULL(subject);
```
here and for the others too.

pfvogel:
The places where these routines are called from within the code base has already done ASSERT checks on the "subject" value prior to these routines being called, so it is safe within the library.  But since these routines can be called directly by a user as well, it does make sense to add checks for the subject value.

pfvogel:
I have added the assertion checks for the subject as well

kisvegabor (commented):
Thanks! :+1:

src/others/observer/lv_observer.c
```diff
@@ -489,6 +495,8 @@ lv_observer_t * lv_arc_bind_value(lv_obj_t * obj, lv_subject_t * subject)
 #if LV_USE_SLIDER
 lv_observer_t * lv_slider_bind_value(lv_obj_t * obj, lv_subject_t * subject)
 {
+    LV_ASSERT_NULL(obj);
+
```
kisvegabor, line 500: One missed here

pfvogel:
Yep,, missed one somehow.  Thanks!  I added the ASSERT and pushed the change.

liamHowatt (approved):
Thanks!

kisvegabor (approved):

kisvegabor:
@FASTSHIFT please take look

FASTSHIFT (approved):

PR #7186, opened 2024-10-30
fix(script): add ending for raw loader of ThorVG

LinkTBF:
The Lottie file exported from AE does not display correctly after running filetohex.py. It only displays properly if 0x00 is added at the end.

### DIFF
--- a/docs/details/libs/rlottie.rst
+++ b/docs/details/libs/rlottie.rst
@@ -81,13 +81,9 @@ array. E.g.:
 
 .. code-block:: shell
 
-   ./filetohex.py path/to/lottie.json > out.txt
-
-``lvgl/scripts/filetohex.py`` only support one-character string. And it is recommended to end with a null terminate to avoid rendering issues. E.g.:
-
-.. code-block:: shell
-
    ./filetohex.py path/to/lottie.json --filter-character --null-terminate > out.txt
+
+``--filter-character`` filters out non-ASCII characters and ``--null-terminate`` makes sure that a trailing zero is appended to properly close the string.
 
 To create an animation from raw data:
 
--- a/docs/details/libs/rlottie.rst
+++ b/docs/details/libs/rlottie.rst
@@ -83,6 +83,12 @@ array. E.g.:
 
    ./filetohex.py path/to/lottie.json > out.txt
 
+``lvgl/scripts/filetohex.py`` only support one-character string. And it is recommended to end with a null terminate to avoid rendering issues. E.g.:
+
+.. code-block:: shell
+
+   ./filetohex.py path/to/lottie.json --filter-character --null-terminate > out.txt
+
 To create an animation from raw data:
 
 .. code-block:: c
--- a/docs/details/widgets/lottie.rst
+++ b/docs/details/widgets/lottie.rst
@@ -68,13 +68,9 @@ array. E.g.:
 
 .. code-block:: shell
 
-   ./filetohex.py path/to/lottie.json > out.txt
-
-``lvgl/scripts/filetohex.py`` only support one-character string. And it is recommended to end with a null terminate to avoid rendering issues. E.g.:
-
-.. code-block:: shell
-
    ./filetohex.py path/to/lottie.json --filter-character --null-terminate > out.txt
+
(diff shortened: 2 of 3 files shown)


kisvegabor:
This file2hex converts the file to hex as it is. Adding `0x00` to the end would be different than what in the file is. 
I suggest either:
- documenting that 0x00 needs to added manually for lottie files
- add an option to the script, e.g. "--null-terminate" (it also needs to be documented)

LinkTBF:
> This file2hex converts the file to hex as it is. Adding `0x00` to the end would be different than what in the file is. I suggest either:
> 
> * documenting that 0x00 needs to added manually for lottie files
> * add an option to the script, e.g. "--null-terminate" (it also needs to be documented)

OK. And I’d like to modify the code below, as Lottie JSON will throw an out-of-range error if it contains CJK characters.

```python
b = bytearray()

if '--filter-character' in sys.argv:
    s = re.sub(r'[^\x00-\xff]', '', s)
if '--null-terminate' in sys.argv:
    s += '\x00'

b.extend(map(ord, s))
```

Or should we leave filetohex.py unchanged and just include everything in the [lv_lottie documentation](https://docs.lvgl.io/master/details/widgets/lottie.html#overview)?

LinkTBF:
> How can it contains CJK characters?

If the Lottie designer's After Effects interface is in Chinese or Japanese, the layer names will contain CJK characters by default, and these layer names will be directly exported to JSON.
```json
{
"layers": [
    {
      "ddd": 0,
      "ind": 1,
      "ty": 0,
      "nm": "圆形", // CJK Layer Name, means "Circle"
      "refId": "comp_0",
      ...
    }
  ]
}
```
---

> Once it's touched, could also add a line break after every 16 item (by item I mean e.g. `0x00`) in the output?

 _add a line break every 16 item_. Is it like this?

```
0x7b, 0x22, 0x76, 0x22, 0x3a, 0x22, 0x34, 0x2e, 0x38, 0x2e, 0x30, 0x22, 0x2c, 0x22, 0x6d, 0x65,
0x34, 0x39, 0x30, 0x31, 0x39, 0x36, 0x2c, 0x31, 0x5d, 0x2c, 0x22, 0x69, 0x78, 0x22, 0x3a, 0x33,
...
0x00
```

liamHowatt:
For reference https://github.com/lvgl/lvgl/blob/master/examples/widgets/lottie/lv_example_lottie_approve.c#L27

I think maybe we should consider removing the requirement to null-terminate the JSON passed to `lv_lottie_set_src_data`. The length is also passed anyways. I'm not sure what it would take to remove the requirement. I'm not clear on whether it's a ThorVG thing.

kisvegabor:
> I think maybe we should consider removing the requirement to null-terminate the JSON passed to lv_lottie_set_src_data. The length is also passed anyways. I'm not sure what it would take to remove the requirement. I'm not clear on whether it's a ThorVG thing.

I've just tried, but even if the size is reduced by one it's faling. 

> add a line break every 16 item. Is it like this?

Yep, exactly!

LinkTBF:
> Yep, exactly!

OK. I have completed the modifications and added the documentation

kisvegabor (commented):
I think it's fine to keep only the working example in to docs as noone should use non-null terminated string. So I recommend:
```rst
``lvgl/scripts/filetohex.py`` can be used to convert a Lottie file a hex
array. E.g.:
.. code-block:: shell
   ./filetohex.py path/to/lottie.json --filter-character --null-terminate > out.txt

``--filter-character`` filters out non-ASCII characters and ``--null-terminate`` makes sure that a trailing zero is appended to properly close the string. 
```

liamHowatt (approved):
All seems good to me!

Just so you know, this converter -- before and after this PR -- seems to be an ASCII only converter. It loads the input file as UTF-8 and fails if there are non-ASCII characters, unless `--filter-character` is given, where it succeeds and filters them out.

file.txt

```
hello 😀️ world
```

```
python3 scripts/filetohex.py file.txt > out1.txt
  File "scripts/filetohex.py", line 16, in <module>
    b.extend(map(ord, s))
ValueError: byte must be in range(0, 256)
```

kisvegabor (approved):
Great! Thanks!

PR #7187, opened 2024-10-30
feat(obj): add transform matrix attribute

FASTSHIFT:
Add matrix style to achieve high performance (no need to get/set style multiple times) + high precision (using float) + NO layer buffer transform effect drawing.

Enable config:
```bash
LV_USE_VECTOR_GRAPHIC
LV_USE_THORVG_INTERNAL
LV_USE_FLOAT
LV_USE_MATRIX
LV_DRAW_TRANSFORM_USE_MATRIX
LV_USE_VG_LITE_THORVG
LV_USE_DRAW_VG_LITE
LV_VG_LITE_USE_GPU_INIT
```

Test code:
```c
static void timer_cb(lv_timer_t* timer)
{
    lv_obj_t* obj = (lv_obj_t*)lv_timer_get_user_data(timer);

    static float value = 0.1f;
    lv_matrix_t matrix;
    lv_matrix_identity(&matrix);
    lv_matrix_scale(&matrix, value, 1);
    lv_matrix_rotate(&matrix, value * 360);
    lv_obj_set_transform_matrix(obj, &matrix);

    value += 0.001f;

    if(value > 2)
        value = 0.1f;
}

void style_matrix_test(void)
{
    lv_timer_create(timer_cb, 1, obj);
    lv_obj_center(obj);
}
```

https://github.com/user-attachments/assets/32a639d4-b646-495b-9487-f7f6a89d7125

3D effect:

https://github.com/user-attachments/assets/a2fa4990-b906-4dd4-a825-a4ebb4837ed6

Known issues:
1. There may be problems with the traditional lvgl transform.
2. Before setting the transform matrix, you need to manually call lv_obj_invalidate to erase the dirty area covered by the old matrix.
3. May not be cropped accurately.
4. The priority of transform_matrix is ​​higher than other transforms (scale, rotation, pivot). When all are set, only matrix will prevail.

### Notes


scripts/style_api_gen.py
```diff
@@ -94,6 +94,10 @@
   'style_type': 'num',   'var_type': 'int32_t',  'default':0, 'inherited': 0, 'layout': 1, 'ext_draw': 1,
   'dsc': "Skew Widget vertically. The value is interpreted in 0.1 degree units. E.g. 450 means 45 deg."},
 
+{'name': 'TRANSFORM_MATRIX',
+ 'style_type': 'ptr',   'var_type': 'const lv_matrix_t *' ,  'default':'`NULL`', 'inherited': 0, 'layout': 0, 'ext_draw': 0,
+ 'dsc': "An initialized `lv_matrix_t` to describe a matrix."},
```
kisvegabor: Mentioned that it requires `LV_USE_MATRIX` and GPU which supports matrix transformations.

kisvegabor (commented):
It's great! 

> Before setting the transform matrix, you need to manually call lv_obj_invalidate to erase the dirty area covered by the old matrix.

[These](https://github.com/lvgl/lvgl/blob/master/src/core/lv_obj_style.c#L106-L108) should handle it.

---

Can you add a test too?

scripts/style_api_gen.py
FASTSHIFT: Done

FASTSHIFT:
> It's great!
> 
> > Before setting the transform matrix, you need to manually call lv_obj_invalidate to erase the dirty area covered by the old matrix.
> 
> [These](https://github.com/lvgl/lvgl/blob/master/src/core/lv_obj_style.c#L106-L108) should handle it.


I feel that this cannot be handled here either. It is necessary to ensure that the dirty area is updated before the user modifies the matrix. If it is handled in `lv_obj_set_style_transform_matrix`, the matrix obtained by lvgl will be modified.

docs/details/base-widget/styles/style-properties.rst
```diff
@@ -320,7 +320,7 @@ Skew Widget vertically. The value is interpreted in 0.1 degree units. E.g. 450 m
 transform_matrix
 ~~~~~~~~~~~~~~~~
 
-An initialized `lv_matrix_t` to describe a matrix.
+An initialized `lv_matrix_t` to describe a matrix. Requires rendering backend to support matrix transformation
```
kisvegabor:
```suggestion
An initialized `lv_matrix_t` to describe a matrix. Requires a rendering backend which support matrix transformations.
```

kisvegabor (dismissed):

FASTSHIFT:
> > It's great!
> > > Before setting the transform matrix, you need to manually call lv_obj_invalidate to erase the dirty area covered by the old matrix.
> > 
> > 
> > [These](https://github.com/lvgl/lvgl/blob/master/src/core/lv_obj_style.c#L106-L108) should handle it.
> 
> I feel that this cannot be handled here either. It is necessary to ensure that the dirty area is updated before the user modifies the matrix. If it is handled in `lv_obj_set_style_transform_matrix`, the matrix obtained by lvgl will be modified.

I think I have solved this problem by storing matrix in `obj->spec_attr`. It is not necessary to store it in style.

kisvegabor:
> I think I have solved this problem by storing matrix in obj->spec_attr. It is not necessary to store it in style.

I see the problem. The styles are so flexible the I wonder if we can keep using them somehow. E.g. we can use different styles in each state, or even animate the properties. 

What if we add 6 styles properties for each matrix element and write some wrappers to handle them in simper way?

W-Mai:
I think we can use

```cpp
lv_obj_transform(obj, matrix);
lv_obj_set_transform(obj, matrix);
lv_obj_reset_transform(obj, matrix);
```

separately  
1. mul a transform matrix
2. set
3. reset to indent

like js canvas do

FASTSHIFT:
Yes, members of matrix cannot simply interpolate independently for animation.

kisvegabor:
Understood. Please document it and add 1-2 examples and we are good to go. A longer description [here](https://github.com/lvgl/lvgl/blob/master/scripts/style_api_gen.py) would make it for now.

This week I'll work on a detailed draw related docs where we can add more about all these.

FASTSHIFT:
> Understood. Please document it and add 1-2 examples and we are good to go. A longer description [here](https://github.com/lvgl/lvgl/blob/master/scripts/style_api_gen.py) would make it for now.
> 
> This week I'll work on a detailed draw related docs where we can add more about all these.

Is there any advantage to putting the matrix in style? I feel that putting it in obj will make access faster and have less impact on the current code.

kisvegabor:
Sorry, it's not a style property anymore. In this case let's wait a little while I add this new docs.

kisvegabor:
We can merge it without the docs for now, but please add some examples.

FASTSHIFT:
> We can merge it without the docs for now, but please add some examples.

Done~

src/core/lv_obj_pos.c
```diff
@@ -983,6 +984,82 @@ void lv_obj_center(lv_obj_t * obj)
...
+
+    /* Free the matrix */
+    lv_free(obj->spec_attr->matrix);
+    obj->spec_attr->matrix = NULL;
+
+    /* Matrix is cleared. Update the layer type */
+    lv_obj_update_layer_type(obj);
+
+    /* Invalidate the new area */
+    lv_obj_invalidate(obj);
```
liamHowatt, line 1044: Is the double-invalidate correct?

liamHowatt (commented):
It looks good.

src/core/lv_obj_pos.c
FASTSHIFT: The area before and after the transformation is different, so redraw the area before the transformation. Otherwise it will be like this:

https://github.com/user-attachments/assets/1a6f1675-2be0-4854-9169-6315862cc9b7

liamHowatt (dismissed):

XuNeo (dismissed):

kisvegabor (approved):
Look good to me!

XuNeo (approved):

PR #7190, opened 2024-10-30
feat(text): handle Combining Diacritical Marks range when typesetting

### LINKED ISSUE #7090: Combining Diacritical Marks do not render properly
### LVGL version

v8.4.0

### What happened?

I would like to render the test phrase "B̶̹̌e̷̘͝ ̵͙̀N̴̹̂o̷̻͛t̴̰̎ ̵̱̓Ą̶͒f̶̳̕r̵̿ͅa̴̧͝ì̷̬d̵̩́", but instead, LVGL's label widget displays:
image

Each combining mark is rendered as a missing character. If a combining mark is missing from the font, could we instead skip drawing it?

### How to reproduce?

Render "B̶̹̌e̷̘͝ ̵͙̀N̴̹̂o̷̻͛t̴̰̎ ̵̱̓Ą̶͒f̶̳̕r̵̿ͅa̴̧͝ì̷̬d̵̩́"

kisvegabor:
fixes #7090

cc @bryghtlabs-richard 

### Notes


bryghtlabs-richard (approved):
This looks good to me.

kisvegabor:
I was thinking about this issue and as it's doesn't work for both lower and upper case letters and doesn't work for [Devanagari](https://en.wikipedia.org/wiki/Devanagari) I wouldn't merge it.  

For CDM you there is a single unicode character for each combination and it can be used instead.  What do you think?

bryghtlabs-richard:
I just got back from out of town last week.

>For CDM there is a single unicode character for each combination and it can be used instead. What do you think?

Sorry, I'm not sure I understand. Do you mean using pre-composed NFC form characters instead of NFD characters with the base character + CDM? Sadly, we need to handle user-specified input here. Though if you prefer to require applications to normalize their strings to NFC form before displaying, that would be fine if we added it to the LVGL documentation.

Separately, normalizing to NFD and handling CDMs correctly can make the fonts smaller overall, as the total number of glyphs in the font is smaller.

kisvegabor:
No worries! :slightly_smiling_face: 

Yes, I mean normalizing the text to NFC before sending it to LVGL. My problem is that we can't display both upper and lower cased latter correctly, so I wouldn't add a half working "solution".

bryghtlabs-richard:
That works for me, but we should update the documentation to reflect that.

Maybe "LVGL supports most UTF-8 encoded Unicode characters. Combining diacritical marks must be pre-composed in NFC form"

Eventual support for NFD is still welcome as it will cut down the font sizes for us.

PR #7216, opened 2024-11-04
fix(indev): don't reset all indevs if an object is set disabled

### LINKED ISSUE #7152: indev: object does not receive LV_EVENT_RELEASED if an unrelated object state changes to LV_STATE_DISABLED
### LVGL version

9.2.0

### What happened?

Take a look at the example below. Just after the `button` is pressed, the `unrelated_obj` changes it state to disabled. The indev-pointer (and all other indev's) looses the active object because it is reset.

This is annoying and I hope this shouldn't be the case. I think only indev's which do have something related to the `unrelated_obj` should be handled.

---

Why is this happen? Let's take a look at the code.

1. Assuming the button has received `LV_EVENT_PRESSED` after reaching:

https://github.com/lvgl/lvgl/blob/a8d2cc05f116d08a1a0d58f9544e3d4ee9dab9d1/src/indev/lv_indev.c#L728

2. Later, LV_STATE_DISABLED is added to the `unrelated_obj`...

https://github.com/lvgl/lvgl/blob/a8d2cc05f116d08a1a0d58f9544e3d4ee9dab9d1/src/core/lv_obj.c#L306-L314

3. ... which performs an `lv_indev_reset` over all indev's via `indev_reset_core`

https://github.com/lvgl/lvgl/blob/a8d2cc05f116d08a1a0d58f9544e3d4ee9dab9d1/src/indev/lv_indev.c#L412-L417

4. Which just resets any active indev object and adds it to the `reset_query`:

https://github.com/lvgl/lvgl/blob/a8d2cc05f116d08a1a0d58f9544e3d4ee9dab9d1/src/indev/lv_indev.c#L1685-L1691

5. If now the user releases the mouse and then `lv_indev_read` is called, `indev_proc_reset_query_handler` is called...

https://github.com/lvgl/lvgl/blob/a8d2cc05f116d08a1a0d58f9544e3d4ee9dab9d1/src/indev/lv_indev.c#L206-L218

6.  ...which just resets everything remaining

https://github.com/lvgl/lvgl/blob/a8d2cc05f116d08a1a0d58f9544e3d4ee9dab9d1/src/indev/lv_indev.c#L1510-L1516

7. Any active unrelated object? Forgotten. If the user wouldn't release the mouse, the button would receive the `LV_EVENT_PRESSED` again instead.

**History**
This functionality was introduced in https://github.com/lvgl/lvgl/pull/3483 and later improved in https://github.com/lvgl/lvgl/pull/4539. But still with issues as described above.

### How to reproduce?

```c
static void pressed_cb(lv_event_t* e)
{
    LV_LOG_USER("%d", lv_event_get_code(e));
}

static void released_cb(lv_event_t* e)
{
    LV_LOG_USER("%d", lv_event_get_code(e));
}

bool disable = true;
static void timer_cb(lv_timer_t* timer)
{
    lv_obj_t* obj = lv_timer_get_user_data(timer);
    if (disable) {
       lv_obj_add_state(obj, LV_STATE_DISABLED);
    } else {
       lv_obj_remove_state(obj, LV_STATE_DISABLED);
    }
    disable = !disable;
}

void obj_disable_test()
{
    lv_obj_t* button= lv_obj_create(lv_screen_active());
    lv_obj_add_event_cb(button, pressed_cb, LV_EVENT_PRESSED, NULL);
    lv_obj_add_event_cb(button, released_cb, LV_EVENT_RELEASED, NULL);

    lv_obj_t* unrelated_obj = lv_obj_create(lv_screen_active());
    lv_obj_set_pos(unrelated_obj , 100, 100);

    lv_timer_t* timer = lv_timer_create(timer_cb, 1, unrelated_obj );
    lv_timer_set_repeat_count(timer, -1);
}

```

Viatorus:
Fixes #7152

### DIFF
--- a/src/core/lv_obj.c
+++ b/src/core/lv_obj.c
@@ -309,11 +309,6 @@ void lv_obj_add_state(lv_obj_t * obj, lv_state_t state)
 
     lv_state_t new_state = obj->state | state;
     if(obj->state != new_state) {
-
-        if(new_state & ~obj->state & LV_STATE_DISABLED) {
-            lv_indev_reset(NULL, obj);
-        }
-
         update_obj_state(obj, new_state);
     }
 }


kissa96 (commented):
Looks good and seems reasonable, but Im not sure about hidden potential issues, as the lib might expect indevs to be reset in this case. @kisvegabor can confirm

kisvegabor (approved):
Look good, thank you!

liamHowatt:
@niklasf do you have any thoughts about this change? It's the same code as this PR: #6855

niklasf:
I think it's probably generally fine. It's just that users may not expect to receive further input events for a newly disabled widget.

Viatorus:
@niklasf this should still be the case. It is know handled directly in lv_indev.c: https://github.com/lvgl/lvgl/blob/master/src/indev/lv_indev.c#L1276

niklasf:
Oh, nice :+1:

liamHowatt (approved):

PR #7221, opened 2024-11-05
fix(freertos):Compilation warning in non-IDF environments

PeakRacing:
fix: Compilation warning in non-IDF environments


liamHowatt (approved):
Thanks!

kisvegabor (approved):

PR #7227, opened 2024-11-05
fix(event) remove redundant last check

cristian-stoica:
There is no other processing after that. The same result will be return with or without the checks

### Notes


liamHowatt (approved):
It's clearly a good change but I wonder if it breaks the pattern maintained by the previous checks in the function. i.e. if another check were added to this function sometime in the future, before the final return, this condition would need to be added back again.

kisvegabor (approved):
IT's really hard to decide. This function doesn't change frequently.

PR #7233, opened 2024-11-05
feat(x11):  add more XImage formats support

matyay:
The X11 driver queries X server for screen color bit depth via `XDisplayPlanes` but later it assumes that it is always 24bpp RGB888. This PR changes that so the driver can render on 16bpp RGB565 and 8bpp grayscale screens too.

### DIFF
--- a/src/drivers/x11/lv_x11_display.c
+++ b/src/drivers/x11/lv_x11_display.c
@@ -104,6 +104,27 @@ static inline lv_color32_t get_px(color_t p)
 #endif
 
 /**
+ * Converts X11 display planes (bits per pixel) to bytes per pixel.
+ * @param[in] dplanes X11 display planes value
+ */
+static inline int lv_dplanes_to_bytes(int dplanes)
+{
+    switch(dplanes) {
+        case 8:
+            return 1;
+        case 16:
+            return 2;
+        case 24:
+            return 4;
+        case 32:
+            return 4;
+    }
+
+    // Should not happen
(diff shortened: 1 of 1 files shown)


kisvegabor:
Hey, 

What about setting LVGL's color format (`lv_display_set_color_format()`) in `lv_x11_window_create` according the the color format of the hardware.

This way LVGL will render in that color format without any conversations.

matyay:
Hi,

That is another approach but to make it work still some changes in the code are needed. The problem is that the X11 driver assumes that the X11 color format is RGB888, see:
 - https://github.com/lvgl/lvgl/blob/master/src/drivers/x11/lv_x11_display.c#L138 
 - https://github.com/lvgl/lvgl/blob/master/src/drivers/x11/lv_x11_display.c#L341

Also I did a quick check in the code and I can't see `disp->color_format` set by `lv_display_set_color_format()`  used anywhere in X11 driver. The driver relies on `LV_COLOR_DEPTH` and always converts to RGB888 depending on that.

If we want to avoid color conversion in the driver we may check if HW format matches `LV_COLOR_DEPTH` and throw an error if it does not. Still requires changes in the image copy loop code.

kisvegabor:
The default color format of the display is set based on `LV_COLOR_DEPTH`. Regardless of it we can call `lv_display_set_color_format()` based of the hardware's color format.

Although it really requires some changes in the code, I thin it's the way to go.

PR #7234, opened 2024-11-06
feat(font): allow using A1,2,4 bitmaps + handle byte aligned fonts

### LINKED ISSUE #7026: 2bpp/4bpp font glyph format support
### Problem to solve

Both 1/2/4bpp font glyph formats are considered as A8 format during drawing in LVGL 9, we hope the renderer can honor the glyph format of glyph_dsc during label drawing. 

### Success criteria

When the returned format (also the real format) of glyph_dsc (lv_draw_glyph_dsc_t) is A1/2/4, the sw draw unit can draw it correctly.

### Solution outline

the sw draw unit support A1/2/4 glyph formats

### Rabbit holes

no

### Testing

Select 1 builtin font, like lv_font_montserrat_38, just return the original 4bpp bitmap in get_glyph_bitmap callback, then check the letter drawing is correct or not.

### Teaching

N/A

### Considerations

_No response_

kisvegabor:
Fixes #7026 (Let the draw unit use the raw A1,2,4 bitmaps)

You can get the raw (non A8 converted) bitmaps like:
```c
      glyph_draw_dsc->g->req_raw_bitap = 1; //g is lv_font_glyph_dsc_t
      glyph_draw_dsc->glyph_data = lv_font_get_glyph_bitmap(glyph_draw_dsc->g, glyph_draw_dsc->_draw_buf);
```

To let the draw unit decide how to get the bitmap, `lv_font_get_glyph_bitmap` is moved to the draw units. 

Also requires https://github.com/lvgl/lv_font_conv/pull/127 to convert bytes aligned A1,2,4 fonts. 

cc @FASTSHIFT @junminacts @MarkIoanAmat

### Notes

### DIFF
--- a/src/draw/nema_gfx/lv_draw_nema_gfx_label.c
+++ b/src/draw/nema_gfx/lv_draw_nema_gfx_label.c
@@ -137,7 +137,7 @@ static void _draw_nema_gfx_letter(lv_draw_unit_t * draw_unit, lv_draw_glyph_dsc_
 #endif
         }
         else if(glyph_draw_dsc->format >= LV_FONT_GLYPH_FORMAT_A1 &&
-                glyph_draw_dsc->format <= LV_FONT_GLYPH_FORMAT_A8) {
+                glyph_draw_dsc->format <= LV_FONT_GLYPH_FORMAT_A8_ALIGNED) {
             /*Do not draw transparent things*/
             if(glyph_draw_dsc->opa <= LV_OPA_MIN)
                 return;
@@ -153,6 +153,7 @@ static void _draw_nema_gfx_letter(lv_draw_unit_t * draw_unit, lv_draw_glyph_dsc_
             const void * mask_buf = draw_buf->data;
 
 
+            glyph_draw_dsc->glyph_data = lv_font_get_glyph_bitmap(glyph_draw_dsc->g, glyph_draw_dsc->_draw_buf);
             int32_t x = glyph_draw_dsc->letter_coords->x1 - layer->buf_area.x1;
             int32_t y = glyph_draw_dsc->letter_coords->y1 - layer->buf_area.y1;
             int32_t w = glyph_draw_dsc->g->box_w;
--- a/src/font/lv_font.h
+++ b/src/font/lv_font.h
@@ -37,17 +37,23 @@ extern "C" {
 typedef enum {
     LV_FONT_GLYPH_FORMAT_NONE   = 0, /**< Maybe not visible*/
 
-    /**< Legacy simple formats*/
+    /**< Legacy simple formats with no byte padding at end of the lines*/
     LV_FONT_GLYPH_FORMAT_A1     = 0x01, /**< 1 bit per pixel*/
     LV_FONT_GLYPH_FORMAT_A2     = 0x02, /**< 2 bit per pixel*/
     LV_FONT_GLYPH_FORMAT_A4     = 0x04, /**< 4 bit per pixel*/
     LV_FONT_GLYPH_FORMAT_A8     = 0x08, /**< 8 bit per pixel*/
 
-    LV_FONT_GLYPH_FORMAT_IMAGE  = 0x09, /**< Image format*/
+    /**< Legacy simple formats with byte padding at end of the lines*/
+    LV_FONT_GLYPH_FORMAT_A1_ALIGNED = 0x011, /**< 1 bit per pixel*/
+    LV_FONT_GLYPH_FORMAT_A2_ALIGNED = 0x012, /**< 2 bit per pixel*/
+    LV_FONT_GLYPH_FORMAT_A4_ALIGNED = 0x014, /**< 4 bit per pixel*/
+    LV_FONT_GLYPH_FORMAT_A8_ALIGNED = 0x018, /**< 8 bit per pixel*/
+
+    LV_FONT_GLYPH_FORMAT_IMAGE  = 0x19, /**< Image format*/
 
     /**< Advanced formats*/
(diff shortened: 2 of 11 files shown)


junminacts (approved):

liamHowatt:
Will it require fonts to be generated specifically with the `*_ALIGNED` color format? i.e., the font converter will need to be updated?

cc @MarkIoanAmat

kisvegabor:
> Will it require fonts to be generated specifically with the *_ALIGNED color format? i.e., the font converter will need to be updated?


Yes, see:
> Also requires https://github.com/lvgl/lv_font_conv/pull/127 to convert bytes aligned A1,2,4 fonts.

src/draw/nema_gfx/lv_draw_nema_gfx_label.c
MarkIoanAmat, line 1: @kisvegabor There are some more changes needed in this file. You can merge it as it is and I will proceed with the necessary changes.
Thank you.

MarkIoanAmat (approved):

src/draw/nema_gfx/lv_draw_nema_gfx_label.c
kisvegabor: Okay. Once it's merged we can update the GPU driver in a new PRs.

src/font/lv_font.h
```diff
@@ -63,6 +69,10 @@ typedef struct {
     lv_font_glyph_format_t format;  /**< Font format of the glyph see lv_font_glyph_format_t */
     uint8_t is_placeholder: 1;      /**< Glyph is missing. But placeholder will still be displayed*/
 
+    /** 0: Get bitmap should return an A8 or ARGB8888 image.
+     * 1: return the bitmap as it is (Maybe A1/2/4 or any proprietary formats). */
+    uint8_t req_raw_bitmap: 1;
```
liamHowatt, line 74: This is not being set to `1` anywhere yet, is it?

src/font/lv_font_fmt_txt.c
```diff
@@ -86,16 +86,19 @@ const void * lv_font_get_bitmap_fmt_txt(lv_font_glyph_dsc_t * g_dsc, lv_draw_buf
 
     const lv_font_fmt_txt_glyph_dsc_t * gdsc = &fdsc->glyph_dsc[gid];
 
+    if(g_dsc->req_raw_bitmap) return &fdsc->glyph_bitmap[gdsc->bitmap_index];
```
liamHowatt, line 89: I didn't check too deeply but is this ok to be returning a `uint8_t *` while normally from this function a `lv_draw_buf_t *` is returned?

liamHowatt (commented):
Glad to see normal A8 bitmap data can be used directly now.

src/font/lv_font.h
kisvegabor: True, it can be used by some GPUs later.

src/font/lv_font_fmt_txt.c
kisvegabor: That's why it's "raw". The user needs to know/check how to handle it.  In case of `A1/2/4/8 +(_ALIGNED)` is will be a raw byte array.

FASTSHIFT:
We can add a cache manager similar to image cache to the unaligned bitmap fonts, which can ensure that the GPU gets images that meet the requirements and does not need to copy font images frequently. In this way, the current `LV_FONT_GLYPH_FORMAT` format can continue to be used without adding a new format.

FASTSHIFT:
The font drawing buffer of the current mainline code is shared, which does not take into account the situation of GPU asynchronous rendering. Usually, this buffer cannot be modified before the GPU rendering is completed.

I added workarond processing [here](https://github.com/lvgl/lvgl/blob/master/src/draw/vg_lite/lv_draw_vg_lite_label.c#L238) to force the GPU to finish the drawing when drawing each letter. This can ensure the integrity of the drawing buffer, but it will seriously affect the drawing efficiency.

So I think adding a font bitmap cache manager should be the best solution to ensure alignment and drawing efficiency. In the rendering backend without alignment requirements (such as software renderer), the original font data can be directly accessed without caching, minimizing copying.

liamHowatt (approved):

kisvegabor:
> We can add a cache manager similar to image cache to the unaligned bitmap fonts

IMO using the raw data has the advantage of
- can be retrieved immediately 
- no need for caching

> The font drawing buffer of the current mainline code is shared, which does not take into account the situation of GPU asynchronous rendering. Usually, this buffer cannot be modified before the GPU rendering is completed.

It's really a big issue. Maybe it should be the GPU's responsibility to save the draw buffer and free it when the rendering is ready. 

Anyway, using the raw pointer is simple and powerful so I think it's a good first step.

FASTSHIFT (approved):

kisvegabor:
We can consider adding an `lv_font_has_static_bitmap()` function. If it's true, we can know that the command doesn't need to be executed immediately, as the bitmap will stay in place.

FASTSHIFT:
> > We can add a cache manager similar to image cache to the unaligned bitmap fonts
> 
> IMO using the raw data has the advantage of
> 
> * can be retrieved immediately
> * no need for caching
> 
> > The font drawing buffer of the current mainline code is shared, which does not take into account the situation of GPU asynchronous rendering. Usually, this buffer cannot be modified before the GPU rendering is completed.
> 
> It's really a big issue. Maybe it should be the GPU's responsibility to save the draw buffer and free it when the rendering is ready.
> 
> Anyway, using the raw pointer is simple and powerful so I think it's a good first step.

@kisvegabor I have completed the cache of the built-in fonts, you can take a look: https://github.com/lvgl/lvgl/pull/7407

PR #7240, opened 2024-11-06
refact(draw): use task instead of draw_unit as argument to draw functions

cristian-stoica:
This patch relieves draw unit structure from records related to the current task (clip_area and target layer) and puts them back with the task where they belong

This simplifies the signatures of draw functions and permits implementations of different dispatch mechanisms (for example with more than one task given in bulk to be executed by a draw_unit)

### Notes

### DIFF
--- a/src/draw/lv_draw_label.c
+++ b/src/draw/lv_draw_label.c
@@ -201,7 +201,7 @@ void LV_ATTRIBUTE_FAST_MEM lv_draw_letter(lv_layer_t * layer, lv_draw_letter_dsc
     LV_PROFILER_DRAW_END;
 }
 
-void lv_draw_label_iterate_characters(lv_draw_unit_t * draw_unit, const lv_draw_label_dsc_t * dsc,
+void lv_draw_label_iterate_characters(lv_draw_task_t * t, const lv_draw_label_dsc_t * dsc,
                                       const lv_area_t * coords,
                                       lv_draw_glyph_cb_t cb)
 {
@@ -209,7 +209,7 @@ void lv_draw_label_iterate_characters(lv_draw_unit_t * draw_unit, const lv_draw_
     int32_t w;
 
     lv_area_t clipped_area;
-    bool clip_ok = lv_area_intersect(&clipped_area, coords, draw_unit->clip_area);
+    bool clip_ok = lv_area_intersect(&clipped_area, coords, &t->clip_area);
     if(!clip_ok) return;
 
     lv_text_align_t align = dsc->align;
@@ -267,7 +267,7 @@ void lv_draw_label_iterate_characters(lv_draw_unit_t * draw_unit, const lv_draw_
                                                            w, NULL, dsc->flag);
 
     /*Go the first visible line*/
-    while(pos.y + line_height_font < draw_unit->clip_area->y1) {
+    while(pos.y + line_height_font < t->clip_area.y1) {
         /*Go to next line*/
         line_start = line_end;
         line_end += lv_text_get_next_line(&dsc->text[line_start], remaining_len, font, dsc->letter_space, w, NULL, dsc->flag);
@@ -349,8 +349,8 @@ void lv_draw_label_iterate_characters(lv_draw_unit_t * draw_unit, const lv_draw_
             if(sel_start != LV_DRAW_LABEL_NO_TXT_SEL && sel_end != LV_DRAW_LABEL_NO_TXT_SEL) {
 #if LV_USE_BIDI
                 logical_char_pos = lv_text_encoded_get_char_id(dsc->text, line_start);
-                uint32_t t = lv_text_encoded_get_char_id(bidi_txt, next_char_offset);
-                logical_char_pos += lv_bidi_get_logical_pos(bidi_txt, NULL, line_end - line_start, base_dir, t, NULL);
+                uint32_t c_idx = lv_text_encoded_get_char_id(bidi_txt, next_char_offset);
+                logical_char_pos += lv_bidi_get_logical_pos(bidi_txt, NULL, line_end - line_start, base_dir, c_idx, NULL);
 #else
                 logical_char_pos = lv_text_encoded_get_char_id(dsc->text, line_start + next_char_offset);
 #endif
@@ -451,7 +451,7 @@ void lv_draw_label_iterate_characters(lv_draw_unit_t * draw_unit, const lv_draw_
                     fill_area.y2 = fill_area.y1 + underline_width - 1;
 
                     fill_dsc.color = dsc->color;
-                    cb(draw_unit, NULL, &fill_dsc, &fill_area);
+                    cb(t, NULL, &fill_dsc, &fill_area);
                 }
                 if(dsc->decor & LV_TEXT_DECOR_STRIKETHROUGH) {
                     lv_area_t fill_area;
@@ -461,7 +461,7 @@ void lv_draw_label_iterate_characters(lv_draw_unit_t * draw_unit, const lv_draw_
                     fill_area.y2 = fill_area.y1 + underline_width - 1;
 
                     fill_dsc.color = dsc->color;
-                    cb(draw_unit, NULL, &fill_dsc, &fill_area);
+                    cb(t, NULL, &fill_dsc, &fill_area);
                 }
             }
 
@@ -470,7 +470,7 @@ void lv_draw_label_iterate_characters(lv_draw_unit_t * draw_unit, const lv_draw_
                && logical_char_pos >= sel_start && logical_char_pos < sel_end) {
                 draw_letter_dsc.color = dsc->sel_color;
                 fill_dsc.color = dsc->sel_bg_color;
-                cb(draw_unit, NULL, &fill_dsc, &bg_coords);
+                cb(t, NULL, &fill_dsc, &bg_coords);
             }
             else if(recolor_cmd_state == RECOLOR_CMD_STATE_TEXT_INPUT) {
                 draw_letter_dsc.color = recolor;
@@ -479,7 +479,7 @@ void lv_draw_label_iterate_characters(lv_draw_unit_t * draw_unit, const lv_draw_
                 draw_letter_dsc.color = dsc->color;
             }
 
-            lv_draw_unit_draw_letter(draw_unit, &draw_letter_dsc, &pos, font, letter, cb);
+            lv_draw_unit_draw_letter(t, &draw_letter_dsc, &pos, font, letter, cb);
 
             if(letter_w > 0) {
                 pos.x += letter_w + dsc->letter_space;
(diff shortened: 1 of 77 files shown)


src/draw/sw/lv_draw_sw.c
```diff
@@ -486,10 +486,10 @@ static int32_t dispatch(lv_draw_unit_t * draw_unit, lv_layer_t * layer)
     }
 
     t->state = LV_DRAW_TASK_STATE_IN_PROGRESS;
-    draw_sw_unit->base_unit.target_layer = layer;
-    draw_sw_unit->base_unit.clip_area = &t->clip_area;
     draw_sw_unit->task_act = t;
 
+    t->target_layer = layer;
```
kisvegabor: Setting it in `lv_draw_add_task` seems more elegant.

kisvegabor (commented):
This change makes a lot of sense. I don't why I didn't do it like this. :raised_eyebrow: 

It will take some time to test all platforms/GPUs, but that's okay, we are not rushing with v9.3.

cristian-stoica:
Thanks. It'll take a few iterations to resolve the rest of the platforms and I'm taking advantage of the project's automatic testing until Cosmin puts ours in place.

kisvegabor:
> project's automatic testing until Cosmin puts ours in place.

So will you add an NXP build test CI? It would be huge!

src/draw/sw/lv_draw_sw.c
```diff
@@ -537,38 +535,38 @@ static void execute_drawing(lv_draw_sw_unit_t * u)
     lv_draw_task_t * t = u->task_act;
     switch(t->type) {
         case LV_DRAW_TASK_TYPE_FILL:
-            lv_draw_sw_fill((lv_draw_unit_t *)u, t->draw_dsc, &t->area);
+            lv_draw_sw_fill(t, t->draw_dsc, &t->area);
```
nicusorcitu, line 256: why not removing all the other arguments and simply use just `t`?

cristian-stoica: > why not removing all the other arguments and simply use just `t`?

The reach of this change is much bigger and although doable I prefer to not mess with it in this patch.

src/draw/nxp/pxp/lv_draw_pxp_layer.c
```diff
@@ -91,7 +91,7 @@ void lv_draw_pxp_layer(lv_draw_unit_t * draw_unit, const lv_draw_image_dsc_t * d
         area_rot.y2 += coords->y1;
     }
     lv_area_t draw_area;
-    if(!lv_area_intersect(&draw_area, &area_rot, draw_unit->clip_area)) return;
+    if(!lv_area_intersect(&draw_area, &area_rot, &t->`clip_area)) return;
```
nicusorcitu: typo, remove the `

src/draw/nxp/pxp/lv_draw_pxp.c
```diff
@@ -412,12 +412,8 @@ static void _pxp_execute_drawing(lv_draw_pxp_unit_t * u)
         if(!lv_area_intersect(&draw_area, &t->area, u->base_unit.clip_area))
             return;
 
-        int32_t idx = 0;
-        lv_draw_unit_t * draw_unit_tmp = _draw_info.unit_head;
-        while(draw_unit_tmp != (lv_draw_unit_t *)u) {
-            draw_unit_tmp = draw_unit_tmp->next;
-            idx++;
-        }
+        int32_t idx = u->base_unit.idx;
```
nicusorcitu, line 417: I like the patch with unit id idx!

kisvegabor:
In general I think this PR goes to the right direction. :slightly_smiling_face: I like these updates! 

> I've had to make a [minor modification](https://github.com/lvgl/lvgl/compare/6e033fffc902f4e9a9e93b1eb49e65eb50ead3f2..52a5cc6bca77d2cb43b7acca9f8e869c69c65a90) to the code to convince the test suite that there is no memory leak.

Could you tell more about it? Maybe we can find an other solution to avoid adding this field.

cristian-stoica:
> > I've had to make a [minor modification](https://github.com/lvgl/lvgl/compare/6e033fffc902f4e9a9e93b1eb49e65eb50ead3f2..52a5cc6bca77d2cb43b7acca9f8e869c69c65a90) to the code to convince the test suite that there is no memory leak.
> 
> Could you tell more about it? Maybe we can find an other solution to avoid adding this field.

That field should be removed. It's not part of the patch but a means to debug the failed tests.

My expectation was that the added field would not make any difference in the memory leaks or show 8 bytes increase. In practice it reduced the leaks and suggests there are some alignment issues that manifest during memory allocation - not a problem of the patch.

I don't have a solution right now and I could not find a justification in git history for why 32 is the acceptable number of leaked bytes for those tests (as opposed to zero or any other number):

        test_lodepng_1:FAIL: Expected 56 to be less than or equal to 40
        test_libpng_1:FAIL: Expected 40 to be less than or equal to 32

kisvegabor:
It's hard to see why these memory leaks are happening without understanding the very details of the allocator. Anyway, it seems like it's not a LVGL issue as we run that loop 20 times in `test_lodepng.c`

cristian-stoica:
not stale

kisvegabor:
Is there anything esle to do/add here?

cristian-stoica:
it seems there are some conflicts that need to be resolved after recent merges

src/draw/lv_draw_label.c
```diff
@@ -544,7 +545,7 @@ static void draw_letter(lv_draw_unit_t * draw_unit, lv_draw_glyph_dsc_t * dsc,
 
     dsc->letter_coords = &letter_coords;
     dsc->g = &g;
-    cb(draw_unit, dsc, NULL, NULL);
+    cb(t, dsc, NULL, NULL);
```
uLipe, line 603: I know this was already there, but could we check this cb before execute it?

uLipe (commented):
Just a single comment, it is my first pass.

uLipe:
I have a general comment not for this PR for sure, but I noticed this PR will start to pass the draw task for the lower ends of the draw engine, along with type and descriptor, so why not convert those functions to receive the single parameter, the task pointer and the function deals with the inner parameters? It would be scalable and better to maintain as long draw task changes.

cristian-stoica:
> convert those functions to receive the single parameter, the task pointer

This is the natural next step and this issue was raised before here https://github.com/lvgl/lvgl/pull/7240#discussion_r1836558627 . I think it is more robust to do this in a few smaller steps instead of a single big one. It is also easier on reviewers especially with large area changes like this one.

src/draw/lv_draw_label.c
cristian-stoica: do you mean like if (cb!= NULL) ?

uLipe:
@cristian-stoica thanks for the comment, I agree on you of do this reducing the risk.

My point was more to understand if this direction was considérés, which seems to be the case.

src/draw/lv_draw_label.c
uLipe: Yes, something like that, if there as guarantee that cb is always valid, that is okay.

src/draw/vg_lite/lv_draw_vg_lite.c
```diff
@@ -51,6 +51,7 @@ static int32_t draw_delete(lv_draw_unit_t * draw_unit);
 /**********************
  *   GLOBAL FUNCTIONS
  **********************/
+lv_draw_vg_lite_unit_t * _vg_lite_unit;
```
FASTSHIFT: Can we avoid using global variables?

FASTSHIFT (changes requested):

src/draw/vg_lite/lv_draw_vg_lite.c
FASTSHIFT: We can add a pointer to `draw_unit` in the `draw_task` structure, so that we can remove the global variable and facilitate debugging. We can know which `draw_unit` the current drawing task is assigned to.

kisvegabor:
Hi,
Is the PR still active?

cristian-stoica:
Yes it is - It needs some changes after review

src/draw/vg_lite/lv_draw_vg_lite_arc.c
```diff
@@ -51,14 +51,15 @@
 /**********************
  *   GLOBAL FUNCTIONS
  **********************/
+extern lv_draw_vg_lite_unit_t * _vg_lite_unit;
```
FASTSHIFT: Is this global variable still needed?

src/draw/vg_lite/lv_draw_vg_lite_arc.c
```diff
@@ -51,14 +51,15 @@
...
 
-void lv_draw_vg_lite_arc(lv_draw_unit_t * draw_unit, const lv_draw_arc_dsc_t * dsc,
+void lv_draw_vg_lite_arc(lv_draw_task_t * t, const lv_draw_arc_dsc_t * dsc,
                          const lv_area_t * coords)
 {
-    lv_draw_vg_lite_unit_t * u = (lv_draw_vg_lite_unit_t *)draw_unit;
+    lv_draw_vg_lite_unit_t * u = _vg_lite_unit;
```
FASTSHIFT: lv_draw_vg_lite_unit_t * u = t->draw_unit;

src/draw/lv_draw_private.h
```diff
@@ -54,12 +54,18 @@ struct _lv_draw_task_t {
...
      */
     lv_area_t clip_area;
+    lv_layer_t * target_layer;
 
 #if LV_DRAW_TRANSFORM_USE_MATRIX
     /** Transform matrix to be applied when rendering the layer */
     lv_matrix_t matrix;
 #endif
 
+#if LV_USE_PARALLEL_DRAW_DEBUG
```
FASTSHIFT: This draw_unit member can be resident, and there is no need to judge `LV_USE_PARALLEL_DRAW_DEBUG`. Use `t->draw_unit` to access the context.

FASTSHIFT (changes requested):

uLipe (dismissed):

FASTSHIFT (dismissed):

cristian-stoica:
@kisvegabor, the base patch is again complete and tested after resolving yesterday's merge conflicts. @FASTSHIFT, @uLipe I will appreciate a re-review after this rebase.

However, since I've added changes for the remaining draw units that required updates dma2d, nemagfx and dave2d, they will need to be reviewed and tested. I've tried to make the changes complete, but I can't build the code and can't run it. Because of this, I will need help from the maintainers of these draw units to build and test the changes for their ports.

For convenience I've left the changes separate to make the review easier.

kisvegabor (approved):
To avoid conflicts let's merge is now, and we can add follow up PRs for the other draw units if needed.

uLipe (approved):
NIce work!

PR #7248, opened 2024-11-07
fix(lv_draw_sw_line): fix lv_draw_line function causes a crash.

### LINKED ISSUE #7247: lv_draw_line function causes a crash.
### LVGL version

V9.2.2

### What happened?

When #define LV_USE_FLOAT 1 is set, the lv_draw_line function causes a crash.

### How to reproduce?

lv_conf.h
#define LV_USE_FLOAT            1

mycode.c

line_dsc.p1.x = 100;
line_dsc.p1.y = 283.2;
line_dsc.p2.x = 200;
line_dsc.p2.y = 283.0;
lv_draw_line(&layer, &line_dsc);

The program crashes after executing the above code.

I have found the cause, and it is as follows:

in  src/draw/sw/lv_draw_sw_line.c

in function:
void lv_draw_sw_line(lv_draw_unit_t * draw_unit, const lv_draw_line_dsc_t * dsc)

The issue occurs because line_dsc.p1.x = 100; line_dsc.p1.y = 283.2; line_dsc.p2.x = 200; line_dsc.p2.y = 283.0;, which causes the draw_line_skew function to be executed.

In the draw_line_skew function, the lv_point_from_precise function converts a float to an int32_t, which results in a loss of precision. After the conversion, line_dsc.p1.y is set to 283 and line_dsc.p2.y is also set to 283. When executing the operation (LV_ABS(xdiff) << 5) / LV_ABS(ydiff), the value of LV_ABS(ydiff) becomes 0, leading to an abnormal program crash.

anheiqq:
Fixes https://github.com/lvgl/lvgl/issues/7247

### DIFF
--- a/src/draw/sw/lv_draw_sw_line.c
+++ b/src/draw/sw/lv_draw_sw_line.c
@@ -65,8 +65,8 @@ void lv_draw_sw_line(lv_draw_unit_t * draw_unit, const lv_draw_line_dsc_t * dsc)
     if(!is_common) return;
 
     LV_PROFILER_DRAW_BEGIN;
-    if(dsc->p1.y == dsc->p2.y) draw_line_hor(draw_unit, dsc);
-    else if(dsc->p1.x == dsc->p2.x) draw_line_ver(draw_unit, dsc);
+    if((int32_t)dsc->p1.y == (int32_t)dsc->p2.y) draw_line_hor(draw_unit, dsc);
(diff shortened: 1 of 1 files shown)


liamHowatt (approved):
Everything you explained in the issue makes sense to me. It's clear from the `/ LV_ABS(xdiff)` that `draw_line_skew` is not meant to be called with horizontal and vertical point pairs.

kisvegabor (approved):

PR #7251, opened 2024-11-07
fix(sw_blend): add null pointer check for mask_area

pfvogel:
In lv_draw_sw_blend(), the blend_dsc->mask_area can be NULL and it is being dereferenced before being checked.  Add a LV_ASSERT_NULL check to the blend_dsc->mask_area variable.


FASTSHIFT (approved):

liamHowatt (approved):
:+1:

Issue #6915, opened 2024-09-23
Potential API breaks?


kisvegabor:
### Introduce the problem

We are developing LVGL's UI editor and in order to generate code in an elegant way it's important to have a consistent API in LVGL. 

### Proposal

For most of the things I can just create new functions and retire the old ones later, but there are two cases where slightly changing the functions' signature would work better. 

API break in a minor version would be very odd so I'd like to discuss these in the hope that you will have good idea/workaround. So 
- `lv_slider_set_value` now has an `anim_en` parameter too. Having a `set_value` and an `set_value_anim` function would be cleaner. An alternative could be to add `set_value_now` and `set_value_anim` as wrappers around set_value
- `lv_roller_set_options` has an infinite/ non infinite parameter too. There should be a separate `set_mode` function instead. Here the workaround can be to have a `set_option_normal` and `set_options_infinite`. 

I'm still exploring how to add custom code to process non standard cases by a hand written parser but in the meantime it'd be great to hear your thoughts on these. 

Cc @XuNeo

XuNeo:
Can we separate those properties? 
For slider, `value` and `anim`: `enable` or `disable`.
For roller:, `mode`: `normal` or `infinite` and `options`.

kisvegabor:
I've just found a way to describe it in XML. I'll share my concept about XML next week, but in short it should support multiple parameters for a property. E.g.
```xml
<roller options="'a\nb\nc' infinite" />
```
The same could work for slider:
```xml
<slider value="30 anim_on"/>
```

This way we can leave LVGL's API as it is. I realized that LVGL's API is
- mature and battle-tested
- shouldn't be broken

So we should align the XML the C API and not the opposite.

XuNeo:
XML is not for human to write, but for automation and parser.
I have to say I cannot agree with this approach. The whole idea of property API is to make it easy for this scenario.

`<roller options="'a\nb\nc' infinite" />` This is absolutely two properties: `mode` and `options`.

> This way we can leave LVGL's API as it is. I realized that LVGL's API is
> 
> mature and battle-tested
> shouldn't be broken

It shouldn't conflict with each other. We should keep two APIs both available.
Instead of current `roller_set_options`. We should also add `set_mode`.

W-Mai:
```html
<roller options="'a'\n'b'\n" mode="infinite" />
<slider value="20" min="0" max="100" with-animation="on" />
```

I think this way will be more clearer.

And it will be translated to like

```cpp
[...]

PR #7252, opened 2024-11-07
Don't squash - Add declarative XML support

coevin:
This PR proposes changes to LVGL to allow developers the ability to declare components and UIs using XML.

### Notes


kisvegabor:
Thank you @coevin. This is what I had in mind. Next week I'll test it locally, and make it comply with LVGL's coding style/standards. I'll also add at least initial docs, examples and tests.

kisvegabor:
It's cleaned up and the basics are working.
- [Example](https://github.com/lvgl/lvgl/pull/7252/files#diff-7fba216d2a8a38a4d28fa536af0ec9bf2abe38d3fb5806fd855285f22dc0c707)
- [Image](https://github.com/lvgl/lvgl/pull/7252/files#diff-bef60607c4fd26b46384e6e8377d044e4f6328a6c302feda3da817b2014e5752)


What is missing:
- Only a few widgets are added
- The added widgets are not full features
- Later it will handle fonts, images, subjects, and events too
- Need more examples
- Need more test
- Need docs

Anyway, I consider it a good start. I'd merge it now and add smaller incremental changes/upates later.

src/others/xml/parsers/lv_xml_obj_parser.c
```diff
@@ -0,0 +1,81 @@
...
+        if(lv_streq("x", name)) lv_obj_set_x(obj, size_process(value));
+        if(lv_streq("y", name)) lv_obj_set_y(obj, size_process(value));
+        if(lv_streq("width", name)) lv_obj_set_width(obj, size_process(value));
+        if(lv_streq("height", name)) lv_obj_set_height(obj, size_process(value));
+        if(lv_streq("align", name)) lv_obj_set_align(obj, lv_xml_align_string_to_enum_value(value));
+        if(lv_streq("style_radius", name)) lv_obj_set_style_radius(obj, lv_xml_atoi(value), 0);
+        if(lv_streq("style_bg_color", name)) lv_obj_set_style_bg_color(obj, lv_color_hex(lv_xml_strtol(value, NULL, 16)), 0);
+        if(lv_streq("style_text_color", name)) lv_obj_set_style_text_color(obj, lv_color_hex(lv_xml_strtol(value, NULL, 16)),
+                                                                               0);
+        if(lv_streq("styles", name)) lv_xml_style_add_to_obj(state, obj, value);
```
XuNeo: The code is repeated in property, subject and here. I don't think it's a good idea to write a separate lv_xml_xxx_parser for every widget.

kisvegabor: First of all, thank you for looking into it. 

By property do you mean lv_obj_property? If so I see what you mean and it'd be great to have a single place for these kind of codes. However I found obj_property limited in some cases (see [here](https://github.com/lvgl/lvgl/issues/6915#issuecomment-2383074845))

So maybe we should approach it the other way around. Build obj_property on top of the parsers somehow.

XuNeo: >  it'd be great to have a single place for these kind of codes

Yes. We can make it really `unified` API.

I'm aware of the limitation of single value per property.
I'm thinking of the solution to support recursive(or combined) value like property of properties.

```c

typedef struct {
    lv_prop_id_t id; /* Reserve 3 bits to indicate it's a combined property, and how many property values are inside. */
    union {
     lv_property_t * sub_properties[]; /* Pointers to an array of lv_property_t */
    ....
    };
} lv_property_t;
```

I'm unfortunately don't have the time to try it out. It would be really great if it can be used in both subject, xml and language bindings.

By doing so, user implemented widgets only have to provide the property table, all subject, xml, and bindings are able to set/get property use the general property API.

kisvegabor: It seems like a great idea,  but I've just realized something. By using the property API we could replace this line
```c
if(lv_streq("x", name)) lv_obj_set_x(obj, size_process(value));
```
with
```c
if(lv_streq("x", name))  lv_obj_set_property(obj, (lv_property_t){.id=LV_PROPERTY_OBJ_X, .num=size_process(value)});
```

It can be structured differently but in the end `"x"` needs to mapped to `LV_PROPERTY_OBJ_X` and the string value needs to be processed as well.

Once we need to map `"x"` to something we can map it directly to `lv_obj_set_x`. In this sense using the property API seems like an extra abstraction layer. That is to convert the attribute and value to a property descriptor, call the property API which calls the LVGL API, instead of calling the LVGL API directly. Now I can not see the advantage of the property API here. Let me know if I miss something. 

> I'm unfortunately don't have the time to try it out. 

Could you or @FASTSHIFT just try out the example, to make sure that it works on others' computer too. :slightly_smiling_face: You just need to enable an FS driver with letter `'A'`.

XuNeo: `lv_prop_id_t lv_obj_property_get_id(const lv_obj_t * obj, const char * name);` is used to get the property ID, thus no need to string compare it manually.

The name to ID map is automatically generated by script `lvgl/scripts/properties.py`.
See example of `lvgl/src/widgets/property/lv_label_properties.c`
**User only need to write once definition of property ID in widget implementation(E.g `lv_image.c`).** Then all the remaining is automatic. No need to write porting layer for xml manually.

For xml and other bindings, an API to get the value from description will be needed. E.g

`lv_xml_convert_property_value(lv_prop_id_t id, const char* value_string)`

I do in this way in lua:
```c
LUALIB_API lv_property_t luavgl_toproperty(lua_State *L, int idx,
                                           lv_prop_id_t id)
{
  lv_property_t prop = {0};
  prop.id = id;
  prop.num = 0;

  int type = LV_PROPERTY_ID_TYPE(id);
  switch (type) {
  case LV_PROPERTY_TYPE_INT:
    prop.num = luavgl_tointeger(L, idx);
    return prop;
  case LV_PROPERTY_TYPE_COLOR:
    prop.color = luavgl_tocolor(L, idx);
    return prop;
...
  default:
    break;
  }

  prop.id = LV_PROPERTY_ID_INVALID;
  return prop;
}
```

I imagine xml will need this:
```xml
<my_h3 text="$btn_text" align="center" color="#orange"/>
```
key: color
value: "#orange"

```c
id = lv_obj_property_get_id(btn, "color")
lv_property_t value = lv_xml_convert_property_value(id, "#orange")
lv_obj_set_property(btn, &value);
```

kisvegabor: Oh, I see. Sounds promising.

Do you have any idea for converting enums, e.g. `"center"` to `LV_ALIGN_CENTER`.

XuNeo: `align` parameter has int type of property value. Using `"center"` string for convention will need to convert to int by xml processor itself. Similarly for color property with convention of `#orange` and `#F12` etc.

E.g.
```c
LUALIB_API lv_color_t luavgl_tocolor(lua_State *L, int idx)
{
  lv_color_t color = {0};
  if (lua_type(L, idx) == LUA_TSTRING) {
    /* support #RGB and #RRGGBB */
    const char *s = lua_tostring(L, idx);
    if (s == NULL) {
      luaL_error(L, "unknown color");
      return color;
    }

    int len = lv_strlen(s);
    if (len == 4 && s[0] == '#') {
      /* #RGB */
      int r = to_int(s[1]);
      r |= r << 4;
      int g = to_int(s[2]);
      g |= g << 4;
      int b = to_int(s[3]);
      b |= b << 4;
      color = lv_color_make(r, g, b);
    } else if (len == 7 && s[0] == '#') {
      /* #RRGGBB */
      int r = (to_int(s[1]) << 4) | to_int(s[2]);
      int g = (to_int(s[3]) << 4) | to_int(s[4]);
      int b = (to_int(s[5]) << 4) | to_int(s[6]);
      color = lv_color_make(r, g, b);
    } else {
      luaL_error(L, "unknown color format");
      return color;
    }
  } else {
    color = lv_color_hex(luavgl_tointeger(L, idx)); /* make to lv_color_t */
  }

  return color;
}
```

kisvegabor: I see. Probably we can generate it somehow too, as it's a quite straight forward mapping. Anyway, at this point I'd like to see the editor working as a whole first, and after that we can get beck to this. We can change this XML parser part later too without modifying any public APIs.

docs/details/other-components/xml.rst
```diff
@@ -0,0 +1,289 @@
+.. _xml:
+
+====================
+XML - Declarative UI
+====================
+
+Introduction
+------------
+VGL is capable of loading UI elements written in XML.
```
liamHowatt:
```suggestion
LVGL is capable of loading UI elements written in XML.
```

docs/details/other-components/xml.rst
liamHowatt, line 1: Add this file to the `index.rst` in this directory.

src/widgets/slider/lv_slider.h
liamHowatt, line 1: Slider inherits Bar so `lv_bar_set_orientation` is sufficient to set slider orientation. What's the reasoning behind adding this API?

If you proceed with this, please update the slider docs to mention these APIs instead of the Bar ones.

docs/details/other-components/xml.rst
kisvegabor: Fixed

docs/details/other-components/xml.rst
kisvegabor: Fixed

src/widgets/slider/lv_slider.h
kisvegabor: On slider, only `obj` and `slider` API should be used. "Intermediate" types should be wrapped to function with the names of the "final" type.

src/widgets/slider/lv_slider.h
```diff
@@ -125,17 +138,31 @@ bool lv_slider_is_dragged(const lv_obj_t * obj);
...
  * @return          true: in symmetrical mode false : not in
 */
 bool lv_slider_is_symmetrical(lv_obj_t * obj);
 
+/**
+ * Set the orientation of slider.
+ * @param obj           pointer to a slider object
+ * @param orientation   slider  orientation from `lv_slider_orientation_t`
+ */
+void lv_slider_set_orientation(lv_obj_t * obj, lv_slider_orientation_t orientation);
```
liamHowatt: I've just noticed you added this prototype twice in this header file.

src/widgets/slider/lv_slider.h
liamHowatt: Okay

src/widgets/slider/lv_slider.h
kisvegabor: Fixed, thanks!

Kconfig
```diff
@@ -1591,7 +1591,9 @@ menu "LVGL configuration"
 			int "Font manager name max length"
 			depends on LV_USE_FONT_MANAGER
 			default 32
-
+		config LV_USE_XML
+			bool "Enable loading XML UIs runtime"
+			default n
```
uLipe: Minor point.

Default n is not necessary, Kconfig by standard set the option to n when nothing is specified.

docs/details/other-components/xml.rst
```diff
@@ -0,0 +1,290 @@
...
+- XML files can be loaded at runtime (e.g., from an SD card) to change the application build.
+- XML is simpler to write than C, enabling people with different skill sets to create LVGL UIs.
+- XML is textual data, making it easy to parse and manipulate with scripts.
+- XML can be used to generate LVGL code in any language.
+- XML helps to separate the view from the logic.
+
+Limitations:
+
+- Only basic widgets are supported with limited functionality.
+- Only a few style properties are supported.
```
uLipe, line 40: Should we mention what is supported here? 

Is there a list of what supported somewhere? I suggest to point to it, or mention the supported. features, users that read this section may infer by themselves what are the limitations of a new feature, which can impact the overall experience.

lv_conf_template.h
```diff
@@ -1076,11 +1076,14 @@
 #define LV_USE_FONT_MANAGER                     0
 #if LV_USE_FONT_MANAGER
 
-/*Font manager name max length*/
+/**Font manager name max length*/
 #define LV_FONT_MANAGER_NAME_MAX_LEN            32
```
uLipe, line 1080: It is okay to make this symbol hard limited instead to expose it into lv_conf and if not defined there then default to the value above?

src/others/xml/lv_xml_component.c
```diff
@@ -0,0 +1,300 @@
...
+
+    /* Housekeeping */
+    lv_free(filename);
+    lv_free(xml_buf);
+    lv_fs_close(&f);
+
+    return res;
+}
+
+void lv_xml_component_unregister(const char * name)
```
uLipe: Why not return lv_result_t since NULL means invalid parameter? It would be nice to track failures like this.

uLipe (commented):
It was my first pass, just few comments.

lv_conf_template.h
kisvegabor: For the sake of simplicity [lv_conf_internal.h](https://github.com/lvgl/lvgl/blob/master/src/lv_conf_internal.h) defines everything to a default value. So in the end there is no such a thing as a on-defined config

src/others/xml/lv_xml_component.c
kisvegabor: Good idea, added.

docs/details/other-components/xml.rst
kisvegabor: Added.

Kconfig
kisvegabor: Added.

uLipe: Great feel free to resolve here.

docs/details/other-components/xml.rst
uLipe: Feel free to resolve it .

lv_conf_template.h
uLipe: Got it, please feel free to resolve.

demos/widgets/lv_demo_widgets.c
```diff
@@ -262,7 +262,7 @@ static void profile_create(lv_obj_t * parent)
     lv_obj_t * dsc = lv_label_create(panel1);
     lv_obj_add_style(dsc, &style_text_muted, 0);
     lv_label_set_text_static(dsc, "This is a short description of me. Take a look at my profile!");
-    lv_label_set_long_mode(dsc, LV_LABEL_LONG_WRAP);
+    lv_label_set_long_mode(dsc, LV_LABEL_LONG_MODE_WRAP);
```
XuNeo, line 265: Could we move this to another PR too?

docs/details/other-components/xml.rst
```diff
@@ -0,0 +1,301 @@
...
+
+        <view extends="lv_button" width="#size" styles="blue red:pressed">
+            <my_h3 text="$btn_text" align="center" color="#orange" style_text_color:checked="0x00ff00"/>
+        </view>
+    </component>
+
+- ``<component>``: The root element.
+- ``<consts>``: Constants with ``px`` or ``color`` types.
+Constants can later be referenced as ``#name``.
+- ``<params>``: Parameters with ``int``, ``px``, ``string``, ``color``, or ``style`` types.
```
XuNeo: Can constants have types same as params?

docs/details/other-components/xml.rst
```diff
@@ -0,0 +1,301 @@
...
+Naming conventions:
+
+- A standard XML syntax is used.
+- Lowercase letters with ``_`` separation are used for attribute names.
+- The usual variable name rules apply for attribute and tag names: only letters, numbers, `'_'` and can't start with a number.
+- LVGL API is followed as much as possible, e.g., ``align="center"``, ``bg_color="0xff0000"``.
+- ``params`` can be referenced with ``$``
+- ``consts`` can be referenced with ``#``
+- ``styles`` can be attached to states and/or parts like ``styles="red blue:pressed green:focused:scrollbar"``
+- Local styles can be used like  ``<lv_label style_text_color="0xff0000" style_text_color:checked="0x00ff00" ``
```
XuNeo, line 118: Is `style_` in `style_text_color` a must? Use `text_color` in both C and `<styles/>` would look more unified.

`style_text_color:checked` The `:checked` seems not that common for me, maybe we should only apply local styles to default state and use the full parameter of `<styles/>` if different state is needed?

docs/details/other-components/xml.rst
```diff
@@ -0,0 +1,301 @@
...
+Usage
+~~~~~
+
+Once a component is created (e.g., ``my_button``), it can be registered by calling either:
+
+- ``lv_xml_component_register_from_file("A:lvgl/examples/others/xml/my_button.xml");``
+- ``lv_xml_component_register_from_data("my_button", xml_data_of_my_button);``
```
XuNeo, line 125: I think the idea of lv_image_set_src(void*) can apply here too, so `lv_xml_component_register` is enough.

Name of `lv_xml_register_component` looks more similar to `lv_image_set_src`.

src/others/xml/lv_xml.c
```diff
@@ -0,0 +1,389 @@
...
+    return NULL;
+}
+
+static void resolve_params(lv_xml_component_ctx_t * item_ctx, lv_xml_component_ctx_t * parent_ctx,
+                           const char ** item_attrs, const char ** parent_attrs)
+{
+    uint32_t i;
+    for(i = 0; item_attrs[i]; i += 2) {
+        const char * name = item_attrs[i];
+        const char * value = item_attrs[i + 1];
```
XuNeo, line 235: We can avoid the double pointer by define 

```
struct _lv_xml_attr {
    const char * name;
    const char * value;
}
```

docs/details/other-components/xml.rst
```diff
@@ -0,0 +1,301 @@
...
+``lv_obj_t * obj = lv_xml_create(lv_screen_active(), "my_button", NULL);``
+
+The created widget is a normal LVGL widget that can be used like any other manually created widget.
+
+The last parameter can be ``NULL`` or an attribute list, like this:
+
+.. code-block:: c
+
+    /* Can be local */
+    char * my_button_attrs[] = {
```
XuNeo, line 141:
```
typedef struct _lv_xml_attr_t {
    const char * name;
    const char * valud:
} lv_xml_attr_t;

const lv_xml_attr_t my_button_attrs[] = {
    {"x", "10"},
    {"y", "-10"},
};

File: docs/details/other-components/xml.rst [hunk 14]
```diff
@@ -0,0 +1,301 @@
...
+        "x", "10",
+        "y", "-10",
+        "align", "bottom_left",
+        "btn_text", "New button",
+        NULL, NULL,
+    };
+
+    lv_obj_t * btn1 = lv_xml_create(lv_screen_active(), "my_button", my_button_attrs);
```
XuNeo, line 149: Feels like the lv_xml_create is bounded to xml, but the functionality can be in a layer that can be reused by other bindings, just like xml.

src/libs/expat/add_lvgl_if.sh
XuNeo, line 1: Please move expat to another PR and merge it firstly.

Please consider adding option to use external expat lib.

src/widgets/slider/lv_slider.h
XuNeo: Looks like still has two of them :)

Could you move this commit to another PR too? It doesn't seem relevant to xml.

demos/widgets/lv_demo_widgets.c
kisvegabor: It's in a separate commit: https://github.com/lvgl/lvgl/pull/7252/commits/096e2bf0c249666aa980f4c790c9271ad4a5b853

src/widgets/slider/lv_slider.h
kisvegabor: Fixed again :crossed_fingers:

docs/details/other-components/xml.rst
kisvegabor: We can consider removing `style_*` from both places but it would be a big API change. Let's treat it seprately.

As LVGL supports local styles for any part and knob it makes sense to me to allow it in XML too. Note that `:checked` can be omitted  to make it affect the main part and default state.

docs/details/other-components/xml.rst
kisvegabor: In case of images `lv_image_dsc_t` is a special struct where the first byte is a magic number to distinguish it from a string. It's not true here as `xml_data_of_my_button` is just a simple array.

src/others/xml/lv_xml.c
kisvegabor: Good idea. I will add in a new PR because I don't want to introduce such a large refactoring. (We will release v0.1 this week).

docs/details/other-components/xml.rst
kisvegabor: I see what you mean. Probably it can be really unified by obj_prop on this level. Let's reconsider it in a new round.

src/libs/expat/add_lvgl_if.sh
kisvegabor: Expat is tightly related to this XML support and it also contains some modification to make it work better with Emscripten, . Therefore I'd like to keep in this PR. Also, it's not guaranteed that an external expat will work. :slightly_frowning_face: 

Besides I can't really see a reason why someone wanted to use it as an external lib.

docs/details/other-components/xml.rst
kisvegabor: True, fixed

kisvegabor:
@XuNeo Thank you for the detailed review. We will release v0.1 of the Editor this week, so I'd like to address only the major issues in this PR and send smaller ones after it. 

Note that there are 3 commits in the PR. After rebase-merge it will nice and clean in history.

src/libs/expat/add_lvgl_if.sh
XuNeo: I mean that LVGL can use external expat lib, just like thorvg, lz4 etc. If the expat is modified and LVGl won't work with external expat lib, then that fine.

docs/details/other-components/xml.rst
XuNeo: It's still possible to judge if the first argument(string) is ended with ".xml".

kisvegabor: I can't check the string length because it might be a pointer to a some data too.

kisvegabor:
Guys, to release v0.1 of our editor we need this feature merged ideally today. If there are is no principal objections  that we can't solve in a follow up PR, please approve and merge it. 

It's not complete for sure and will receive many updates in the upcoming months.

docs/details/other-components/xml.rst
XuNeo: I mean the first argument, in your example, they are both string but one of them is ended with .xml. :)

examples/others/xml/index.rst
```diff
@@ -0,0 +1,5 @@
+Load components at runtime
+--------------------------
+
+.. lv_example:: others/xml/lv_example_snapshot_1
```
liamHowatt:
```suggestion
.. lv_example:: others/xml/lv_example_xml_1
```

src/others/xml/lv_xml_base_types.c
```diff
@@ -0,0 +1,111 @@
...
+}
+
+lv_flex_align_t lv_xml_flex_align_string_to_enum_value(const char * txt)
+{
+    if(lv_streq("center", txt)) return LV_FLEX_ALIGN_CENTER;
+    if(lv_streq("end", txt)) return LV_FLEX_ALIGN_END;
+    if(lv_streq("start", txt)) return LV_FLEX_ALIGN_START;
+    if(lv_streq("space_around", txt)) return LV_FLEX_ALIGN_SPACE_AROUND;
+    if(lv_streq("space_between", txt)) return LV_FLEX_ALIGN_SPACE_BETWEEN;
+    if(lv_streq("space_evently", txt)) return LV_FLEX_ALIGN_SPACE_EVENLY;
```
liamHowatt:
```suggestion
    if(lv_streq("space_evenly", txt)) return LV_FLEX_ALIGN_SPACE_EVENLY;
```

xmls/globals.xml
```diff
@@ -0,0 +1,45 @@
...
+            <enum name="column_reverse"         help="In a column with reverse"/>
+            <enum name="column_wrap_reverse"    help="In columns with wrap and reverse"/>
+    </enumdef>
+    <enumdef name="lv_flex_align">
+            <enum name="center"         help=""/>
+            <enum name="start"    		help=""/>
+            <enum name="end"       		help=""/>
+            <enum name="space_around"  	help=""/>
+            <enum name="space_between"  help=""/>
+            <enum name="space_evently"  help=""/>
```
liamHowatt:
```suggestion
            <enum name="space_evenly"  help=""/>
```

src/others/xml/lv_xml_style.c
```diff
@@ -0,0 +1,202 @@
...
+ *      INCLUDES
+ *********************/
+#include "../../lvgl.h"
+#if LV_USE_XML
+
+#include "lv_xml_base_types.h"
+#include "lv_xml_parser.h"
+#include "lv_xml_style.h"
+#include "lv_xml_utils.h"
+#include <string.h>
```
liamHowatt: Probably just leftover
```suggestion
```

uLipe (dismissed):

uLipe (dismissed):

src/others/xml/lv_xml_style.c
uLipe: Not really the xml style implementations use some functions from the string module, removing it causes a warning of implicit function definition which may be treated as error by the CI or other plaform that enforces warning as error.

For now I suggest to keep on its place, and later we can revisit a better place to put this header.

uLipe (approved):

src/others/xml/lv_xml_style.c
liamHowatt: Sounds good

liamHowatt (approved):

kisvegabor:
Thank you!

PR #7256, opened 2024-11-07
fix(sw_blend): only assign to mask_stride when it is initialized

pfvogel:
In the various lv_draw_sw_blend_to_*.c files, an assignment is made to the mask_stride variable but for some calls to these functions, the variable has not been initialized.  Change the code to defer the assignment to the paths where the mask_stride has been initialized.

This issue was detected by a static code analysis tool, but it is sort of a "false positive".  The paths that do not initialize mask_stride do not use it either, so the "garbage" value of mask_stride did not affect the operation of the code. 

An alternative simpler fix would be to always initialize "mask_stride" in the lv_draw_sw_blend() routine, so that it has a non-garbage value for all paths for the code.

### DIFF
--- a/src/draw/sw/blend/lv_draw_sw_blend_to_al88.c
+++ b/src/draw/sw/blend/lv_draw_sw_blend_to_al88.c
@@ -200,7 +200,6 @@ void LV_ATTRIBUTE_FAST_MEM lv_draw_sw_blend_color_to_al88(lv_draw_sw_blend_fill_
     int32_t h = dsc->dest_h;
     lv_opa_t opa = dsc->opa;
     const lv_opa_t * mask = dsc->mask_buf;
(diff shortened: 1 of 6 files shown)


kisvegabor:
>  An alternative simpler fix would be to always initialize "mask_stride" in the lv_draw_sw_blend() routine, so that it has a non-garbage value for all paths for the code.

I'd rather go for it instead of inserting a lot of these extra lines.

> This issue was detected by a static code analysis tool, but it is sort of a "false positive". The paths that do not initialize mask_stride do not use it either, so the "garbage" value of mask_stride did not affect the operation of the code.

As we don't actively use a static code analyzer now, I'm sure that there are several similar issues. So are there "true positive" issues reported as well?  :slightly_smiling_face:

PR #7263, opened 2024-11-08
fix(roller): fix typo in scaling; prevent division by zero

pfvogel:
In lv_roller.c, the calculation of the scale_x value in the loop in the transform_vect_recursive() routine was using scale_y instead of scale_x.
After the scale values are calculated, check if either scale value is 0 and change to 1 instead to prevent division by zero.


liamHowatt (approved):
Makes sense to me. Thanks.

FASTSHIFT (approved):

PR #7266, opened 2024-11-09
fix(init): remove double call to lv_draw_sw_deinit

antonio-pasini:
Calling lv_deinit() with LV_USE_DRAW_SW= 1, LV_DRAW_SW_COMPLEX = 1 and LV_USE_OS = LV_OS_FREERTOS will end up calling lw_draw_sw_deinit() twice (cfr. lv_init.c, lines367 and lines 407); this makes the mutex deinit fail, because the lv_mutex_delete() is not checking for pxMutex->xIsInitialized being already pdFALSE.

### DIFF
--- a/src/lv_init.c
+++ b/src/lv_init.c
@@ -384,10 +384,6 @@ void lv_deinit(void)
     lv_span_stack_deinit();
 #endif
 
-#if LV_USE_DRAW_SW
-    lv_draw_sw_deinit();
-#endif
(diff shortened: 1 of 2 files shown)


liamHowatt (approved):
Good catch. It really is called twice. It was a good decision to remove the earlier one, IMO

FASTSHIFT (approved):

PR #7269, opened 2024-11-09
fix(draw): fix incorrect clip area computation when clearing transparent framebuffers

### LINKED ISSUE #7268: Partial Framebuffers with Transparency do not get cleared before rendering
### LVGL version

v9.2.0

### What happened?

When rendering with screen transparency, the framebuffer needs to be cleared before rendering to it. Otherwise data from the previous render could get kept in transparent areas.

This happens in `refr_configured_layer`:
```c
static void refr_configured_layer(lv_layer_t *layer)
{
    // ...

    /*If the screen is transparent initialize it when the flushing is ready*/
    if (lv_color_format_has_alpha(disp_refr->color_format))
    {
        lv_draw_buf_clear(layer->draw_buf, &layer->_clip_area);
    }

    // ...
}
```

which again calls the function `lv_draw_buf_clear`:
```c
void lv_draw_buf_clear(lv_draw_buf_t * draw_buf, const lv_area_t * a)
{
    LV_ASSERT_NULL(draw_buf);

    const lv_image_header_t * header = &draw_buf->header;
    uint32_t stride = header->stride;

    if(a == NULL) {
        uint8_t * buf = lv_draw_buf_goto_xy(draw_buf, 0, 0);
        lv_memzero(buf, header->h * stride);
        return;
    }

    lv_area_t a_draw_buf;
    a_draw_buf.x1 = 0;
    a_draw_buf.y1 = 0;
    a_draw_buf.x2 = draw_buf->header.w - 1;
    a_draw_buf.y2 = draw_buf->header.h - 1;

    lv_area_t a_clipped;
    if(!lv_area_intersect(&a_clipped, a, &a_draw_buf)) return;
    if(lv_area_get_width(&a_clipped) <= 0) return;
    if(lv_area_get_height(&a_clipped) <= 0) return;

    uint8_t * buf = lv_draw_buf_goto_xy(draw_buf, a_clipped.x1, a_clipped.y1);
    uint8_t bpp = lv_color_format_get_bpp(header->cf);
    uint32_t line_length = (lv_area_get_width(&a_clipped) * bpp + 7) >> 3;
    int32_t y;
    for(y = a_clipped.y1; y <= a_clipped.y2; y++) {
        lv_memzero(buf, line_length);
        buf += stride;
    }
}
```

However, the `lv_area_intersect` check fails. That's because the `draw_buf` is **not** positioned at `0,0`, like the following lines assume:
```c
    lv_area_t a_draw_buf;
    a_draw_buf.x1 = 0;
    a_draw_buf.y1 = 0;
    a_draw_buf.x2 = draw_buf->header.w - 1;
    a_draw_buf.y2 = draw_buf->header.h - 1;
```

But instead, it is positioned at `layer->buf_area` of the `refr_configured_layer` function.

This information does not get forwarded to the `lv_draw_buf_clear` function, so I'm unsure how to fix that.

Adding an additional parameter to `lv_draw_buf_clear` probably has unforseen consequences at other points in the code, so I need some guidance of how to fix that.

The result of that error looks like this:
384604453-59c5913e-86c7-4b08-bba1-865e11487207

(only a single "Hello world" should be visible in the middle)

### How to reproduce?

Render with a partial framebuffer with an ARGB8888 screen pixel type.

Use transparency. Observe state bleed between partial framebuffer renders.

Finomnis:
Fixes #7268.

Fixes incorrect clip area computation when clearing the framebuffer of transparent screens.

This is especially an issue with partial renders, as the framebuffer duplicates content into the next writes. This can be seen here:

Screenshot without the fix:
384604453-59c5913e-86c7-4b08-bba1-865e11487207

Screenshot with the fix:
image

### DIFF
--- a/src/core/lv_refr.c
+++ b/src/core/lv_refr.c
@@ -770,9 +770,7 @@ static void refr_configured_layer(lv_layer_t * layer)
     }
     /*If the screen is transparent initialize it when the flushing is ready*/
     if(lv_color_format_has_alpha(disp_refr->color_format)) {
-        lv_area_t clear_area;
-        lv_area_set(&clear_area, layer->_clip_area.x1, layer->_clip_area.y1,
-                    layer->_clip_area.x2, layer->_clip_area.y2);
+        lv_area_t clear_area = layer->_clip_area;
         lv_area_move(&clear_area, -layer->buf_area.x1, -layer->buf_area.y1);
         lv_draw_buf_clear(layer->draw_buf, &clear_area);
     }
--- a/src/core/lv_refr.c
+++ b/src/core/lv_refr.c
@@ -770,7 +770,11 @@ static void refr_configured_layer(lv_layer_t * layer)
     }
     /*If the screen is transparent initialize it when the flushing is ready*/
     if(lv_color_format_has_alpha(disp_refr->color_format)) {
-        lv_draw_buf_clear(layer->draw_buf, &layer->_clip_area);
+        lv_area_t clear_area;
+        lv_area_set(&clear_area, layer->_clip_area.x1, layer->_clip_area.y1,
+                    layer->_clip_area.x2, layer->_clip_area.y2);
+        lv_area_move(&clear_area, -layer->buf_area.x1, -layer->buf_area.y1);
+        lv_draw_buf_clear(layer->draw_buf, &clear_area);
     }
 
     lv_obj_t * top_act_scr = NULL;


liamHowatt (dismissed):
Amazing, thank you!

To reproduce the issue on SDL, use partial mode and call:

```c
lv_display_set_color_format(NULL, LV_COLOR_FORMAT_ARGB8888);
lv_obj_set_style_bg_opa(lv_screen_active(), 8, 0);
lv_example_label_5();
```

Finomnis:
> Amazing, thank you!
> 
> To reproduce the issue on SDL, use partial mode and call:
> 
> ```c
> lv_display_set_color_format(NULL, LV_COLOR_FORMAT_ARGB8888);
> lv_obj_set_style_bg_opa(lv_screen_active(), 8, 0);
> lv_example_label_5();
> ```

@liamHowatt

Awesome, thanks :)
In case you want to add a test for it, I'm unfamiliar with the lvgl test framework.

src/core/lv_refr.c
```diff
@@ -770,7 +770,11 @@ static void refr_configured_layer(lv_layer_t * layer)
     }
     /*If the screen is transparent initialize it when the flushing is ready*/
     if(lv_color_format_has_alpha(disp_refr->color_format)) {
-        lv_draw_buf_clear(layer->draw_buf, &layer->_clip_area);
+        lv_area_t clear_area;
+        lv_area_set(&clear_area, layer->_clip_area.x1, layer->_clip_area.y1,
```
FASTSHIFT: lv_area_t clear_area = layer->_clip_area;

FASTSHIFT (changes requested):

kisvegabor (approved):
A customer just reported this issue, but I couldn't reproduce it. Now I know why: I was using DIRECT rendering (the buffer is always at 0;0) and they were using PARTIAL.

So thank you very much! ❤️

kisvegabor:
@liamHowatt @FASTSHIFT please review.

FASTSHIFT (approved):

Issue #6926, opened 2024-09-24
Dropdown scrollbar never set as LV_STATE_SCROLLED when using encoders


zalexzperez:
### LVGL version

v9.2.0

### What happened?

When a dropdown list is scrolled using an encoder, the scrollbar always has the same color: 
![dropdown3](https://github.com/user-attachments/assets/1e4a68a1-81f0-4992-9ec5-175ce8f3fc65)
(Encoder simulated with mouse wheel)

I specifically set the color for the LV_PART_SCROLLBAR | LV_STATE_SCROLLED with no effect. Here's the code for the list customization:

```
lv_obj_t *obj = lv_dropdown_get_list(parent_obj);
lv_obj_set_style_bg_color(obj, lv_color_hex(0xffd0d0d0), LV_PART_MAIN | LV_STATE_DEFAULT);
lv_obj_set_style_text_font(obj, &lv_font_montserrat_12, LV_PART_MAIN | LV_STATE_DEFAULT);
lv_obj_set_style_bg_color(obj, lv_color_hex(0xff0000ff), LV_PART_SCROLLBAR | LV_STATE_DEFAULT);
lv_obj_set_style_bg_color(obj, lv_color_hex(0xffff0000), LV_PART_SCROLLBAR | LV_STATE_SCROLLED);
```
Then, I looked into the source code, and found this statement in `lv_dropdown.c` :
`lv_obj_scroll_to_y(dropdown->list, line_y1, LV_ANIM_OFF);`

If `LV_ANIM_OFF` is changed to `LV_ANIM_ON`, then it works as expected:
![dropdown4](https://github.com/user-attachments/assets/ac2ab141-a963-4517-9d33-911d44e4fd58)

Fix found with the help of mvladic

liamHowatt:
Nice, I think this is a good fix. I think the scrollbar should have all the same styling capabilities in the dropdown as anywhere else. I wonder if there's any reason to not animate encoder scrolling in the dropdown.

cc @kisvegabor

kisvegabor:
I'm ok with making it animated. However `position_to_selected()` (calls `lv_obj_scroll_to_y`) should have an `anim_en` parameter because no animation should happen when e.g. the drop-down opens.

kissa96:
@zalexzperez Do you want to make a PR out of this, or should I handle it?

zalexzperez:
> @zalexzperez Do you want to make a PR out of this, or should I handle it?

[...]

PR #7271, opened 2024-11-09
feat(dropdown): add animations on rotary event

AndreCostaaa:
Tackles #6926 by enabling animation when selecting the dropdown position following a `LV_EVENT_ROTARY` event

### DIFF
--- a/src/widgets/dropdown/lv_dropdown.c
+++ b/src/widgets/dropdown/lv_dropdown.c
@@ -56,7 +56,7 @@ static lv_result_t btn_release_handler(lv_obj_t * obj);
 static lv_result_t list_release_handler(lv_obj_t * list_obj);
 static void list_press_handler(lv_obj_t * page);
 static uint32_t get_id_on_point(lv_obj_t * dropdown_obj, int32_t y);
-static void position_to_selected(lv_obj_t * obj);
+static void position_to_selected(lv_obj_t * dropdown_obj, lv_anim_enable_t anim_en);
 static lv_obj_t * get_label(const lv_obj_t * obj);
 
 /**********************
@@ -347,7 +347,7 @@ void lv_dropdown_set_selected(lv_obj_t * obj, uint32_t sel_opt)
     dropdown->sel_opt_id_orig = dropdown->sel_opt_id;
 
     if(dropdown->list) {
-        position_to_selected(obj);
+        position_to_selected(obj, LV_ANIM_OFF);
     }
 
     lv_obj_invalidate(obj);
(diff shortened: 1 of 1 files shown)


src/widgets/dropdown/lv_dropdown.c
```diff
@@ -773,7 +773,7 @@ static void lv_dropdown_event(const lv_obj_class_t * class_p, lv_event_t * e)
             }
             else if(dropdown->sel_opt_id + 1 < dropdown->option_cnt) {
                 dropdown->sel_opt_id++;
-                position_to_selected(obj);
+                position_to_selected(obj, LV_ANIM_OFF);
```
liamHowatt: It's my personal opinion that that one should be `LV_ANIM_ON` because key navigation is not too dissimilar to encoder navigation.

Same for the other key event.

src/widgets/dropdown/lv_dropdown.c
```diff
@@ -347,7 +347,7 @@ void lv_dropdown_set_selected(lv_obj_t * obj, uint32_t sel_opt)
     dropdown->sel_opt_id_orig = dropdown->sel_opt_id;
 
     if(dropdown->list) {
-        position_to_selected(obj);
+        position_to_selected(obj, LV_ANIM_OFF);
```
liamHowatt, line 350: For this one, I feel like the public function `lv_dropdown_set_selected` that this is in could get a `lv_anim_enable_t` parameter. It would make it more consistent with other similar APIs such as

- `lv_bar_set_value`
- `lv_roller_set_selected`
- `lv_slider_set_value`
- `lv_gridnav_set_focused`
- `lv_tabview_set_active`
- `lv_tileview_set_tile`

It would be a "breaking change" but a good one, in my opinion.

liamHowatt (commented):
Thank you for tackling the issue!

src/widgets/dropdown/lv_dropdown.c
AndreCostaaa: Thought so aswell but didn't want to introduce any breaking changes to the public api
Should I create a new PR with the breaking change or do it in this one ?

src/widgets/dropdown/lv_dropdown.c
AndreCostaaa: Agreed, will change

src/widgets/dropdown/lv_dropdown.c
liamHowatt: Ok, let's start with no breaking change and then there can be a follow-up PR. Thanks.

liamHowatt (approved):

kisvegabor (approved):
Thank you!

PR #7290, opened 2024-11-14
perf(span): optimize span render performance up by 50%

W-Mai:
It is easy to notice that if you process a snippet each time, you can skip the first step, directly add a complete string to the drawing queue, and draw a piece of text in groups.

image

But there is a one-to-more relationship between span and snippet, which causes drawing a snippet, which may only be part of drawing span. However, c-string does not have the concept of slicing, and only half of the string cannot be intercepted for drawing without copying. So the following transformations were made:
image
image

All APIs mainly related to text drawing and calculation support the calculation and drawing of a given length of text. In this way, you can accurately control how much should be drawn at a time. In addition, the text style drawing, which previously needed to be processed separately within the spangroup, has also been uniformly replaced by the style drawing ability in draw label, which not only ensures the consistency of the code, but also reduces the call of the code.

image

### Notes


lizhaoming634:
Good job!

W-Mai:
There are some errors about the pictures generated by vglite, which need to be generated from a new one.

W-Mai:
CI Problem has been fixed

kisvegabor (commented):
It makes sens and it looks and sounds good! 

There is only one thing that I'm thinking about: in `lv_text_get_next_line` len=0 could also mean to process 0 bytes. Therefore I vote for using `UINT32_MAX` if there is no limit. A define like `#define LV_TEXT_LEN_MAX UINT32_MAX` can be added too.

kisvegabor (approved):
Looks good, thank you!

FASTSHIFT (approved):

PR #7308, opened 2024-11-15
feat(demos): High Resolution Demo

liamHowatt:
Add a new demo. It's the "high resolution demo" because it's intended for larger displays and well-suited to MPUs. Assets are optimized for display sizes of exactly 800x480, 1280x720, or 1920x1080.

### Notes


liamHowatt:
Opening for review now. Optimization can be iterated on in later PRs as discussed.

uLipe:
It is more a general comment, I see these demis are becoming larger and larger, and they will become sometimes distant from the lvgl core library, because they are application focused.

Said this, have you considered to move the demos to a special repository that consumes the core lvgl one?

kisvegabor:
We have discussed with @liamHowatt that we will make it smaller by using SVG images (instead of SVGs in 3 sizes) and also load the fonts from TTF directly. Anyway, as demos are added LVGL really get bigger and bigger.

2-3 year ago we had an `lv_demos` repo but keeping it in sync with LVGL was a pain. Now it's so simple that if something changes in LVGL it can be immediately changed in the demos too.   CI and versioning is also so much simpler. I'd rather remove some demos instead of having multi-repos again.

uLipe:
Oh I see, maybe my understanding of the démos might be wrong on their design.

I was thinking demos as a self contained applications, containing not only the ui but a link to the lvgl core lib which could be a separate repo. 

Thanks for the clarification.

kisvegabor:
> I was thinking demos as a self contained applications, containing not only the ui but a link to the lvgl core lib which could be a separate repo.

This will result in the versioning and compatibility issues that  I've described. E.g. if someone changes a function API in LVGL, it needed to be updated in the demos repo too. Which is cumbersome. :slightly_frowning_face:

kisvegabor (approved):

uLipe:
Hmmm but in that cases we would be suffering on a design/maintenance issues, user application API on LVGL should not break between minor versions.

Between major versions API introduction, breaking or even demo deprecation are expected and the development process should address that.

In any case if there is room for further discussion on this topic I can open an issue dedicated for it.

Sorry for pollution on the review.

uLipe (approved):

kisvegabor:
When we have shifted towards a monorepo we have unified 
- lvgl
- tests
- examples
- demos

However the demos are really getting large with thousands of files. So it might make sense to make separate only the demos. 

Anyway, let's discuss it separately.

PR #7320, opened 2024-11-18
fix(dave2d): fix implicit function definitions

jeremy-baker:
Compilation with the default FSP e2studio project settings results in a compilation failure with implicit functions definitions error

### Notes


src/draw/renesas/dave2d/lv_draw_dave2d.h
```diff
@@ -27,6 +27,9 @@ extern "C" {
 #include "../../lv_draw_triangle.h"
 #include "../../lv_draw_buf.h"
 
+#include "../../lv_draw_image_private.h"
+#include "../../../misc/lv_area_private.h"
```
uLipe: Question, if those include files are intended to be private, they should not be exposed to the public interface file, right? What about adding these include only on file that call these private functions, I know it would involve more effort, but it will keep the design of the public interface consistent.

jeremy-baker: Updated, only added the private includes to those files that actually use the private functions

uLipe (approved):

kisvegabor (approved):

liamHowatt (approved):
Thanks!

PR #7329, opened 2024-11-19
fix(obj): use LV_ASSERT_NULL if LV_ASSERT_OBJ not enabled

pfvogel:
If LV_USE_ASSERT_OBJ is set to 0, no validation of the object is done.  Add call to LV_ASSERT_NULL on the object in this case to at least validate the object pointer is not NULL.


kisvegabor (approved):
Good idea!

liamHowatt:
Moved here #7339

PR #7332, opened 2024-11-20
feat(display): add draw buffer size getter

faxe1008:

- Relax constraint in the vtile method && add a parameter for the buffer size to the flush callback.
I am sure I missed to adjust some documentation somewhere.

### Notes

### DIFF
--- a/src/display/lv_display.c
+++ b/src/display/lv_display.c
@@ -1031,6 +1031,37 @@ void lv_display_rotate_area(lv_display_t * disp, lv_area_t * area)
     }
 }
 
+uint32_t lv_display_get_draw_buf_size(lv_display_t * disp)
+{
+    if(!disp) disp = lv_display_get_default();
+    if(!disp) return 0;
+
+    if(disp->buf_1) {
+        return disp->buf_1->data_size;
+    }
+    return 0;
+}
+
+uint32_t lv_display_get_invalidated_draw_buf_size(lv_display_t * disp, uint32_t width, uint32_t height)
+{
+    if(!disp) disp = lv_display_get_default();
+    if(!disp) return 0;
+
+    if(disp->render_mode == LV_DISPLAY_RENDER_MODE_FULL) {
+        width = lv_display_get_horizontal_resolution(disp);
+        height = lv_display_get_vertical_resolution(disp);
+    }
+
+    lv_color_format_t cf = lv_display_get_color_format(disp);
+    uint32_t stride = lv_draw_buf_width_to_stride(width, cf);
+    uint32_t buf_size = stride * height;
+
+    LV_ASSERT(disp->buf_1 && disp->buf_1->data_size >= buf_size);
+    if(disp->buf_2) LV_ASSERT(disp->buf_2->data_size >= buf_size);
+
+    return buf_size;
+}
+
 lv_obj_t * lv_screen_active(void)
 {
     return lv_display_get_screen_active(lv_display_get_default());
--- a/src/display/lv_display.h
+++ b/src/display/lv_display.h
@@ -569,6 +569,24 @@ lv_draw_buf_t * lv_display_get_buf_active(lv_display_t * disp);
  */
 void lv_display_rotate_area(lv_display_t * disp, lv_area_t * area);
 
+/**
+ * Get the size of the draw buffers
+ * @param disp      pointer to a display
+ * @return          the size of the draw buffer in bytes for valid display, 0 otherwise
+ */
+uint32_t lv_display_get_draw_buf_size(lv_display_t * disp);
+
+/**
+ * Get the size of the invalidated draw buffer. Can be used in the flush callback
+ * to get the number of bytes used in the current render buffer.
+ * @param disp      pointer to a display
+ * @param width     the width of the invalidated area
+ * @param height    the height of the invalidated area
+ * @return          the size of the invalidated draw buffer in bytes, not accounting for
+ *                  any preceeding palette information for a valid display, 0 otherwise
+ */
+uint32_t lv_display_get_invalidated_draw_buf_size(lv_display_t * disp, uint32_t width, uint32_t height);
+
 /**********************
  *      MACROS
  **********************/
--- a/tests/src/test_cases/test_display.c
+++ b/tests/src/test_cases/test_display.c
@@ -1,0 +1,130 @@
+#if LV_BUILD_TEST
+#include "../lvgl.h"
+#include "unity/unity.h"
+
+void setUp(void)
+{
+    /* Function run before every test */
+}
+
+void tearDown(void)
+{
+    /* Function run after every test */
+}
+
+struct display_area_test_set {
+    /* Parameters for setting up the display */
+    uint32_t width;
+    uint32_t height;
+    lv_color_format_t color_format;
+    lv_display_render_mode_t render_mode;
+    /* Parameters for testing */
+    uint32_t invalidated_width;
+    uint32_t invalidated_height;
+    uint32_t expected_buf0_size;
+};
+
+void test_get_drawbuf_size_double_buffered(void)
+{
+    static  LV_ATTRIBUTE_MEM_ALIGN uint8_t buf0[200 + LV_DRAW_BUF_ALIGN];
+    static  LV_ATTRIBUTE_MEM_ALIGN uint8_t buf1[200 + LV_DRAW_BUF_ALIGN];
+
+    lv_display_t * disp = lv_display_create(10, 20);
+    lv_display_set_color_format(disp, LV_COLOR_FORMAT_RGB888);
+
+    lv_display_set_buffers(disp, lv_draw_buf_align(buf0, LV_COLOR_FORMAT_RGB888), lv_draw_buf_align(buf1,
+                                                                                                    LV_COLOR_FORMAT_RGB888), 200, LV_DISPLAY_RENDER_MODE_PARTIAL);
+
+    TEST_ASSERT_EQUAL(200, lv_display_get_draw_buf_size(disp));
+}
+
+void test_get_drawbuf_size_single_buffered(void)
+{
+    static  LV_ATTRIBUTE_MEM_ALIGN uint8_t buf0[200 + LV_DRAW_BUF_ALIGN];
+
+    lv_display_t * disp = lv_display_create(10, 20);
+    lv_display_set_color_format(disp, LV_COLOR_FORMAT_RGB888);
+
+    lv_display_set_buffers(disp, lv_draw_buf_align(buf0, LV_COLOR_FORMAT_RGB888), NULL, 200,
+                           LV_DISPLAY_RENDER_MODE_PARTIAL);
+
+    TEST_ASSERT_EQUAL(200,  lv_display_get_draw_buf_size(disp));
+}
+
+static void exec_invalidated_drawbuf_size_test(const struct display_area_test_set * test_set)
+{
+    uint32_t buffer_size = lv_draw_buf_width_to_stride(test_set->width, test_set->color_format) * test_set->height;
+
+    uint8_t * buf0 = lv_malloc(buffer_size + LV_DRAW_BUF_ALIGN);
+
+    lv_display_t * disp = lv_display_create(test_set->width, test_set->height);
+    lv_display_set_color_format(disp, test_set->color_format);
+
+    lv_display_set_buffers(disp, lv_draw_buf_align(buf0, test_set->color_format), NULL, buffer_size,
+                           test_set->render_mode);
+    uint32_t invalidated_size = lv_display_get_invalidated_draw_buf_size(disp, test_set->invalidated_width,
+                                                                         test_set->invalidated_height);
+
+    TEST_ASSERT_EQUAL(test_set->expected_buf0_size, invalidated_size);
+    lv_free(buf0);
+    lv_display_delete(disp);
+}
+
+void test_get_invalidated_drawbuf_size_rgb888_partial()
+{
+    struct display_area_test_set test_set = {
+        .width = 10,
+        .height = 20,
+        .color_format = LV_COLOR_FORMAT_RGB888,
+        .render_mode = LV_DISPLAY_RENDER_MODE_PARTIAL,
+        .invalidated_width = 5,
+        .invalidated_height = 5,
+        .expected_buf0_size = lv_draw_buf_width_to_stride(5, LV_COLOR_FORMAT_RGB888) * 5,
+    };
+    exec_invalidated_drawbuf_size_test(&test_set);
+}
+
+void test_get_invalidated_drawbuf_size_rgb888_full()
+{
+    struct display_area_test_set test_set = {
+        .width = 10,
+        .height = 20,
+        .color_format = LV_COLOR_FORMAT_RGB888,
+        .render_mode = LV_DISPLAY_RENDER_MODE_FULL,
+        .invalidated_width = 10,
+        .invalidated_height = 20,
+        .expected_buf0_size = lv_draw_buf_width_to_stride(10, LV_COLOR_FORMAT_RGB888) * 20,
+    };
+    exec_invalidated_drawbuf_size_test(&test_set);
+}
+
+void test_get_invalidated_drawbuf_size_i1_full()
+{
+    struct display_area_test_set test_set =  {
+        .width = 180,
+        .height = 90,
+        .color_format = LV_COLOR_FORMAT_I1,
+        .render_mode = LV_DISPLAY_RENDER_MODE_FULL,
+        .invalidated_width = 180,
+        .invalidated_height = 90,
+        .expected_buf0_size = lv_draw_buf_width_to_stride(180, LV_COLOR_FORMAT_I1) * 90,
+    };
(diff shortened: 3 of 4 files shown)


faxe1008:
Don't know how to handle the micropython bindigs - where are they generated/kept?

examples/arduino/LVGL_Arduino/LVGL_Arduino.ino
```diff
@@ -34,7 +34,7 @@ void my_print( lv_log_level_t level, const char * buf )
 #endif
 
 /* LVGL calls it when a rendered image needs to copied to the display*/
-void my_disp_flush( lv_display_t *disp, const lv_area_t *area, uint8_t * px_map)
+void my_disp_flush( lv_display_t *disp, const lv_area_t *area, uint8_t * px_map, uint32_t px_map_size)
```
uLipe: It is not a blocker comment, but I have seeing some PR that break the API and parameter change, would not be good to consider (maybe in a longer term) to pack parameters passed from callbacks or for the API inside of structs? 

It would be easier to maintain and would not break the API itself.

uLipe (dismissed):
LGTM so far, but someone must review as well.

examples/arduino/LVGL_Arduino/LVGL_Arduino.ino
faxe1008: Yeah if not overdone would be nicer thats true. Not sure if passing a const version of the layer is the right approach?

kisvegabor:
Uh, we can't change the API of such an important functions. :slightly_frowning_face: 

What is the actual motivation behind this change?

faxe1008:
Hmm, I know its painful to change this API, let me explain:

The flush callback is in the documentation advertised as a place where besides writing to the display the user can also perform some buffer manipulation:

https://docs.lvgl.io/master/details/main-components/display.html#flush-callback

Typically the setup for these operations is something like this:
```
uint16_t * buf16 = (uint16_t *)px_map; /* Let's say it's a 16 bit (RGB565) display */
    int32_t x, y;
            // manipulate or write
            buf16++;
        }
    }
```
The issue here is that there is an implicit assumption about the buffer size. As an API consumer you have no way of verifying if you are performing an out of bands write/read here. You can in theory get the displays color format and then recalulate the assumed buffer size. But this means as a user you need internal knowledge of how each color format impact the drawbuffer layout requirements in regards to sizing. Take the I1 color format for example. Naively writing the data to a display in the style of the above sample requires you to know that you have to skip the palette bytes and width is rounded to a multiple of 8.

kisvegabor:
When you call `lv_display_set_buffers(disp, buf1, buf2, buf_size_bytes, render_mode)` you need to set the buffer size and it will be saved in the `display` object. There is no API function to get it, but we can easily add one. 

But note that in partial mode if you set a 100kB buffer, but only a 2x2 area is updated, only the first few bytes of the buffer will be filled. Probably it's not an issue for an out-of-bounds check, I just wanted to mention it.

faxe1008:
Thx for the idea, will add a helper for it. That would indeed solve my issue for the time being :^)

kisvegabor:
Great! Feel free to clean this PR or open a new one to add this function.

src/display/lv_display.c
```diff
@@ -1031,6 +1031,36 @@ void lv_display_rotate_area(lv_display_t * disp, lv_area_t * area)
     }
 }
 
+void lv_display_get_draw_buf_size(lv_display_t * disp, uint32_t * buf0_size, uint32_t * buf1_size)
+{
+    if(!disp) disp = lv_display_get_default();
+    if(!disp) return;
+
+    if(disp->buf_1) *buf0_size = disp->buf_1->data_size;
```
kisvegabor: Size is always the same of the two buffers. As the second buffer is optional we can just return the size of the first buffer.

src/display/lv_display.h
```diff
@@ -569,6 +569,23 @@ lv_draw_buf_t * lv_display_get_buf_active(lv_display_t * disp);
...
+/**
+ * Get the size of the draw buffers
+ * @param disp      pointer to a display
+ * @param buf0_size pointer to a variable to store the size of the first buffer
+ * @param buf1_size pointer to a variable to store the size of the second buffer
+ */
+void lv_display_get_draw_buf_size(lv_display_t * disp, uint32_t * buf0_size, uint32_t * buf1_size);
+
+/**
+ * Get the size of the invalidated draw buffer
```
kisvegabor:
```suggestion
 * Get the size of the invalidated draw buffer. Can be used in the flush callback the get the number of bytes used in the current render buffer. 
```

src/display/lv_display.c
```diff
@@ -1031,6 +1031,37 @@ void lv_display_rotate_area(lv_display_t * disp, lv_area_t * area)
...
+    if(!disp) disp = lv_display_get_default();
+    if(!disp) return 0;
+
+    if(disp->render_mode == LV_DISPLAY_RENDER_MODE_FULL) {
+        return disp->buf_act->data_size;
+    }
+
+    lv_color_format_t cf = lv_display_get_color_format(disp);
+    uint32_t stride = width * lv_color_format_get_bpp(cf);
+    stride = (stride + 7) >> 3; /* Round up, convert to bytes */
```
liamHowatt: Just a suggestion.

```suggestion
    uint32_t stride = width * lv_color_format_get_size(cf);
```

faxe1008: Very much appreciated, thanks :D

src/display/lv_display.c
```diff
@@ -1031,6 +1031,35 @@ void lv_display_rotate_area(lv_display_t * disp, lv_area_t * area)
...
+    return 0;
+}
+
+uint32_t lv_display_get_invalidated_draw_buf_size(lv_display_t * disp, uint32_t width, uint32_t height)
+{
+    if(!disp) disp = lv_display_get_default();
+    if(!disp) return 0;
+
+    if(disp->render_mode == LV_DISPLAY_RENDER_MODE_FULL) {
+        return disp->buf_act->data_size;
```
faxe1008: Thinking about it - the user might allocate a larger buffer then actually needed. So instead something like this would make sense, right?

```
        width = lv_display_get_horizontal_resolution(disp);
        height = lv_display_get_vertical_resolution(disp);
```

kisvegabor: Good idea! Yes, in case of FULL mode it makes perfect sense. 

However we need should use `stride` instead of `width` here too.

src/display/lv_display.c
faxe1008: @liamHowatt can not use this one unfortunately since it returns bytes per pixel and in the  case of I1 its rounded.

src/display/lv_display.c
```diff
@@ -1031,6 +1031,38 @@ void lv_display_rotate_area(lv_display_t * disp, lv_area_t * area)
...
+    if(!disp) disp = lv_display_get_default();
+    if(!disp) return 0;
+
+    if(disp->render_mode == LV_DISPLAY_RENDER_MODE_FULL) {
+        width = lv_display_get_horizontal_resolution(disp);
+        height = lv_display_get_vertical_resolution(disp);
+    }
+
+    lv_color_format_t cf = lv_display_get_color_format(disp);
+    uint32_t stride = width * lv_color_format_get_bpp(cf);
```
kisvegabor: This functions is used internally, to handle any GPU specific stride settings: 
```suggestion
    uint32_t stride = lv_draw_buf_width_to_stride(width, cf);
```
Please use it here as well.

kisvegabor (commented):
Note the CI failure here: 
https://github.com/lvgl/lvgl/actions/runs/12135420247/job/33834439204?pr=7332#step:7:4851

src/display/lv_display.c
```diff
@@ -1031,6 +1031,37 @@ void lv_display_rotate_area(lv_display_t * disp, lv_area_t * area)
     }
 }
 
+uint32_t lv_display_get_draw_buf_size(lv_display_t * disp)
+{
+    if(!disp) disp = lv_display_get_default();
+    if(!disp) return 0;
```
uLipe, line 1037: Question: I noticed for invalid cases, for example invalid display pointer we return 0 sized buffer, would not be better to return an invalid size token instead, or if by doing this the API would break an LOG_ERROR on such cases are very welcome to track why invalid display was passed to this function.

src/display/lv_display.c
```diff
@@ -1031,6 +1031,37 @@ void lv_display_rotate_area(lv_display_t * disp, lv_area_t * area)
...
+    if(disp->buf_1) {
+        return disp->buf_1->data_size;
+    }
+    return 0;
+}
+
+uint32_t lv_display_get_invalidated_draw_buf_size(lv_display_t * disp, uint32_t width, uint32_t height)
+{
+    if(!disp) disp = lv_display_get_default();
+    if(!disp) return 0;
```
uLipe, line 1048: Same as above.

src/display/lv_display.h
```diff
@@ -569,6 +569,23 @@ lv_draw_buf_t * lv_display_get_buf_active(lv_display_t * disp);
  */
 void lv_display_rotate_area(lv_display_t * disp, lv_area_t * area);
 
+/**
+ * Get the size of the draw buffers
+ * @param disp      pointer to a display
+ * @return          the size of the draw buffer in bytes
```
uLipe: I suggest: 

```suggestion
 * @return          the size of the draw buffer in bytes for valid display, 0 otherwise
```

src/display/lv_display.h
```diff
@@ -569,6 +569,23 @@ lv_draw_buf_t * lv_display_get_buf_active(lv_display_t * disp);
...
+ */
+uint32_t lv_display_get_draw_buf_size(lv_display_t * disp);
+
+/**
+ * Get the size of the invalidated draw buffer. Can be used in the flush callback
+ * to get the number of bytes used in the current render buffer.
+ * @param disp      pointer to a display
+ * @param width     the width of the invalidated area
+ * @param height    the height of the invalidated area
+ * @return          the size of the invalidated draw buffer in bytes, not accounting for any preceeding palette information
```
uLipe:
```suggestion
 * @return          the size of the invalidated draw buffer in bytes, not accounting for any preceeding palette information for a valid display, 0 otherwise
```

src/display/lv_display.c
faxe1008: Its really a question of general API design. There are two ways of doing it:
`int foo(lv_display_t*,  uint32_t*)`
and `uint32_t foo(lv_display_t*)` and returning a special value.

I have not seen much use of the first one inside the codebase so I went with the second approach :^)

uLipe: Make sense so far, I think error handling would be a separate topic to dicuss, so I will resolve that :)

kisvegabor:
Thanks for the update, but the test still fails :slightly_frowning_face:

faxe1008:
I seperated each test case to easier see which ones are failing:
```
/home/runner/work/lvgl/lvgl/tests/src/test_cases/test_display.c:95:test_get_drawbuf_size_double_buffered:INFO: sysmon: 0 FPS (refr_cnt: 0 | redraw_cnt: 0), refr 0ms (render 0ms | flush 0ms), CPU 0%\n
/home/runner/work/lvgl/lvgl/tests/src/test_cases/test_display.c:27:test_get_drawbuf_size_double_buffered:PASS
/home/runner/work/lvgl/lvgl/tests/src/test_cases/test_display.c:95:test_get_drawbuf_size_single_buffered:INFO: sysmon: 0 FPS (refr_cnt: 0 | redraw_cnt: 0), refr 0ms (render 0ms | flush 0ms), CPU 0%\n
/home/runner/work/lvgl/lvgl/tests/src/test_cases/test_display.c:41:test_get_drawbuf_size_single_buffered:PASS
/home/runner/work/lvgl/lvgl/tests/src/test_cases/test_display.c:95:test_get_invalidated_drawbuf_size_rgb888_partial:INFO: sysmon: 0 FPS (refr_cnt: 0 | redraw_cnt: 0), refr 0ms (render 0ms | flush 0ms), CPU 0%\n
/home/runner/work/lvgl/lvgl/tests/src/test_cases/test_display.c:68:test_get_invalidated_drawbuf_size_rgb888_partial:FAIL: Expected 320 Was 80
/home/runner/work/lvgl/lvgl/tests/src/test_cases/test_display.c:95:test_get_invalidated_drawbuf_size_rgb888_full:INFO: sysmon: 0 FPS (refr_cnt: 0 | redraw_cnt: 0), refr 0ms (render 0ms | flush 0ms), CPU 0%\n
/home/runner/work/lvgl/lvgl/tests/src/test_cases/test_display.c:68:test_get_invalidated_drawbuf_size_rgb888_full:FAIL: Expected 1280 Was 640
/home/runner/work/lvgl/lvgl/tests/src/test_cases/test_display.c:95:test_get_invalidated_drawbuf_size_i1_full:INFO: sysmon: 0 FPS (refr_cnt: 0 | redraw_cnt: 0), refr 0ms (render 0ms | flush 0ms), CPU 0%\n
/home/runner/work/lvgl/lvgl/tests/src/test_cases/test_display.c:68:test_get_invalidated_drawbuf_size_i1_full:FAIL: Expected 5760 Was 2160
/home/runner/work/lvgl/lvgl/tests/src/test_cases/test_display.c:95:test_get_invalidated_drawbuf_size_i1_partial:INFO: sysmon: 0 FPS (refr_cnt: 0 | redraw_cnt: 0), refr 0ms (render 0ms | flush 0ms), CPU 0%\n
/home/runner/work/lvgl/lvgl/tests/src/test_cases/test_display.c:68:test_get_invalidated_drawbuf_size_i1_partial:FAIL: Expected 640 Was 240
```

Seems to me like in `display.c` the `LV_DRAW_BUF_ALIGN` is 1 for 64-bit test but in the `test_display.c` its `64` for some weird reason. Not sure what is going on there @kisvegabor

tests/src/test_cases/test_display.c
```diff
@@ -0,0 +1,130 @@
...
+void test_get_invalidated_drawbuf_size_rgb888_partial()
+{
+    struct display_area_test_set test_set = {
+        .width = 10,
+        .height = 20,
+        .color_format = LV_COLOR_FORMAT_RGB888,
+        .render_mode = LV_DISPLAY_RENDER_MODE_PARTIAL,
+        .invalidated_width = 5,
+        .invalidated_height = 5,
+        .expected_buf0_size = LV_ROUND_UP(5 * 3, LV_DRAW_BUF_STRIDE_ALIGN) * 5,
```
liamHowatt: > Seems to me like in `display.c` the `LV_DRAW_BUF_ALIGN` is 1 for 64-bit test but in the `test_display.c` its `64` for some weird reason. Not sure what is going on there kisvegabor

Can you make the expected result be based on the configured `LV_DRAW_BUF_ALIGN`?

faxe1008: Hmm not sure, I thought the `LV_DRAW_BUF_ALIGN` is only important for allocating the draw buffer correctly. It should not matter in regards to the invalidation size - right?

src/display/lv_display.c
```diff
@@ -1031,6 +1031,37 @@ void lv_display_rotate_area(lv_display_t * disp, lv_area_t * area)
...
+        width = lv_display_get_horizontal_resolution(disp);
+        height = lv_display_get_vertical_resolution(disp);
+    }
+
+    lv_color_format_t cf = lv_display_get_color_format(disp);
+    uint32_t stride = lv_draw_buf_width_to_stride(width, cf);
+    uint32_t buf_size = stride * height;
+
+    LV_ASSERT(!disp->buf_1 || disp->buf_1->data_size >= buf_size);
+    LV_ASSERT(!disp->buf_2 || disp->buf_2->data_size >= buf_size);
```
kisvegabor: `buf_2` is optional:

```suggestion
    LV_ASSERT(disp->buf_1 && disp->buf_1->data_size >= buf_size);
    if(disp->buf_2) LV_ASSERT(disp->buf_2->data_size >= buf_size);
```

tests/src/test_cases/test_display.c
kisvegabor: I found the problem. The test fails with the VGLite test where `width_to_stride` is [overwritten](https://github.com/lvgl/lvgl/blob/master/src/draw/vg_lite/lv_vg_lite_utils.c#L654-L661) and doesn't use the `LV_DRAW_BUF_STRIDE_ALIGN`.

I suggest 

```suggestion
        .expected_buf0_size = lv_draw_buf_width_to_stride(5, test_set.color_format),
```

faxe1008: When originially implementing this I looked at the impl of `lv_draw_buf_width_to_stride` and wondered why this was architectured in a way the the stride calculation functions are overwritable. Guess I now have my answer :D

src/display/lv_display.c
faxe1008: True, fixed that.

tests/src/test_cases/test_display.c
kisvegabor: Yeah, GPUs can have specific stride requirements.

kisvegabor (approved):
Looks good, thank you!

FASTSHIFT (approved):

PR #7339, opened 2024-11-20
fix(obj): use LV_ASSERT_NULL if LV_ASSERT_OBJ not enabled

pfvogel:
(Previous pull request for this got deleted)

When the LV_USE_ASSERT_OBJ configuration is not enabled, no checking is done on the object.  Change the macro to call LV_ASSERT_NULL on on the object instead.


liamHowatt (approved):
Thanks!

kisvegabor (approved):

PR #7341, opened 2024-11-20
fix(sw_blend): make sure mask_stride is initialized

pfvogel:
(Previous pull request for this was deleted).

In some paths through the code, the value of mask_stride did not get initialized (but it was not used in those paths, so no bad behavior resulted from the uninitialized value).  Add initialization to the mask_stride value so that it is initialized for all paths.


uLipe (approved):

kisvegabor (approved):
Nice, thanks!

FASTSHIFT (approved):

PR #7343, opened 2024-11-20
fix(obj_style): transform_scale_(xy)_safe return LV_SCALE_NONE instea…

pfvogel:
If  lv_obj_get_style_transform_scale_x() returns an invalid scale value, the _safe version returns 1,  Where the _safe routine is used, the return value is used to multiply the current scale value, then the resulting scale value is shifted down by 8 bits to produce the final scale value.  If a value of 1 is returned by the _safe function, the resulting scale factor can become 0 after the shift.  The scale value is then used as the denominator of a division, possibly resulting in a divide by zero.
Change the return value for an invalid scale from 1 to LV_SCALE_NONE (defined to be 256).

### DIFF
--- a/src/core/lv_obj_style.h
+++ b/src/core/lv_obj_style.h
@@ -290,13 +290,13 @@ lv_text_align_t lv_obj_calculate_style_text_align(const lv_obj_t * obj, lv_part_
 static inline int32_t lv_obj_get_style_transform_scale_x_safe(const lv_obj_t * obj, lv_part_t part)
 {
     int32_t scale = lv_obj_get_style_transform_scale_x(obj, part);
-    return scale > 0 ? scale : 1;
+    return scale > 0 ? scale : LV_SCALE_NONE;
 }
 
 static inline int32_t lv_obj_get_style_transform_scale_y_safe(const lv_obj_t * obj, lv_part_t part)
 {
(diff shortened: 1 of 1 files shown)


uLipe (commented):
Not a blocker, but since the scale is signed, why not return -1 instead making sure we will never have by hardware a division by zero, since -1 would mean all bits in a word are set?

pfvogel:
The code snippet above from lv_indev.c will result in a divide by zero if the _safe() routine returns 1 and the previous scale value was less than 256.  The result of the multiply will be a value that is less than 256, and then the shift down by 8 will result in a scale value of zero, which becomes the denominator in a division.

These routines are used in 4 places currently, and two of the locations are already checking to replace a scale of 0 with 1.  The other two locations (lv_indev,c and lv_indev_scroll.c) do not have this check.

In retrospect, not making this change and instead fixing lv_indev.c and lv_indev_scroll.c is a better fix, as suggested by @kisvegabor.  I'll suggest that this pull request is closed without merging and instead make the fixes in the other two locations.

uLipe:
I like also the approach, since the problem sounds like a specific thing, so indev is responsible to check the denominator instead of relying on the behavior of an external function.

pfvogel:
The fix was implemented in #7354 instead of this change

PR #7344, opened 2024-11-20
fix(draw_sw_box_shadow): add ASSERT_MALLOC check

pfvogel:
The return value from lv_malloc is not checked to see that it is not NULL in lv_draw_sw_box_shadow().  Add LV_ASSERT_MALLOC check.


uLipe (approved):

kisvegabor (approved):
Thanks!

FASTSHIFT (approved):

PR #7346, opened 2024-11-21
feat(NemaGFX): add freetype vector font support

MarkIoanAmat:
### Added
- FreeType Support, Vector Fonts
- Dry run of Freetype, to avoid costly reallocations

### Fixed
- Stride issue in lv_draw_nema_gfx_img.c

@kisvegabor @liamHowatt 
Waiting for your review and comments, thank you.


src/draw/nema_gfx/lv_draw_nema_gfx_label.c
```diff
@@ -105,6 +131,120 @@ void lv_draw_nema_gfx_label(lv_draw_unit_t * draw_unit, const lv_draw_label_dsc_
...
+static bool new_alloc = false;
+
+static void freetype_outline_event_cb(lv_event_t * e)
+{
+    LV_PROFILER_BEGIN;
+    lv_event_code_t code = lv_event_get_code(e);
+    lv_freetype_outline_event_param_t * param = lv_event_get_param(e);
+    if(new_alloc) {
+        /*Now we can do the nessecary memory allocations*/
+        lv_nema_gfx_path_alloc(param->outline);
```
uLipe: We have allocation here, it will always succeed, what happens if it fails?

kisvegabor:
It's great that you could implement this feature despite of the related docs is missing. :cry: 

With @liamHowatt we will try it out tomorrow. Can we test it like [this](https://github.com/lvgl/lvgl/blob/master/tests/src/test_cases/libs/test_freetype.c#L485-L488)?

cc @FASTSHIFT

src/draw/nema_gfx/lv_draw_nema_gfx_label.c
```diff
@@ -95,6 +112,15 @@ void lv_draw_nema_gfx_label(lv_draw_unit_t * draw_unit, const lv_draw_label_dsc_
 
     nema_set_clip(clip_area.x1, clip_area.y1, lv_area_get_width(&clip_area), lv_area_get_height(&clip_area));
 
+#if LV_USE_FREETYPE && LV_USE_NEMA_VG
+    /*We need to initiallize in every new label task*/
+    static bool is_init = false;
```
kisvegabor:
```suggestion
    static bool is_inited = false;
```

kisvegabor:
I've fixed the CI error in master. Please rebase the PR on `master`.

src/draw/nema_gfx/lv_draw_nema_gfx_label.c
MarkIoanAmat: Changed the name

src/draw/nema_gfx/lv_nema_gfx_path.c
```diff
@@ -0,0 +1,151 @@
...
+    LV_PROFILER_END;
+    return nema_gfx_path;
+}
+
+void lv_nema_gfx_path_alloc(lv_nema_gfx_path_t * nema_gfx_path)
+{
+    LV_PROFILER_BEGIN;
+    nema_gfx_path->path = nema_vg_path_create();
+    nema_gfx_path->paint = nema_vg_paint_create();
+    nema_gfx_path->data = (float *) lv_malloc(nema_gfx_path->data_size * sizeof(float));
```
XuNeo, line 69: The `lv_array` can be used here to eliminate the global variable `data_point ` and `seg_point `.

src/libs/freetype/lv_freetype_outline.c
```diff
@@ -354,6 +360,40 @@ static lv_freetype_outline_t outline_create(
...
+                vectors += 2;
+            }
+        }
+    }
+    /*Also for every contour we may have a line for close*/
+    segments += face->glyph->outline.n_contours;
+    vectors += face->glyph->outline.n_contours;
+
+    lv_nema_gfx_path_t * nema_gfx_path = (lv_nema_gfx_path_t *) param.outline;
+    nema_gfx_path->data_size = vectors * 2;
```
XuNeo: We should avoid doing it in common code. How about still relying on lv_array to dynamically extend the data_size?

src/draw/nema_gfx/lv_draw_nema_gfx_label.c
```diff
@@ -606,4 +755,4 @@ static uint32_t get_glyph_dsc_id(const lv_font_t * font, uint32_t letter)
 
 }
 
-#endif /*LV_USE_NEMA_GFX*/
+#endif /*LV_USE_NEMA_GFX*/
```
XuNeo, line 822: Please add the missing new line at file end.

XuNeo (commented):
Nice work!

src/libs/freetype/lv_freetype_outline.c
MarkIoanAmat: Dynamic allocation in those cases could cause major performance issues. It is quite common to have glyphs described with ~ 150 segments. That could cause 150 reallocations for segments points + 150 reallocations for data points and thats for one letter alone, in each frame (especially if the glyph is not cached that could cause a bigger issue). This block inside lv_freetype_outline.c is only enabled when LV_USE_NEMA_GFX is used, so if this is not a hard request from your side we would like to keep this solution.

src/draw/nema_gfx/lv_draw_nema_gfx_label.c
uLipe: I noticed there is an assert for the internal malloc :)

src/libs/freetype/lv_freetype_outline.c
XuNeo: We can pre-alloc the array to hold 150 elements and doubles it whenever runs out of space.

If it's really critical, then dynamic allocation should be avoided.

> if this is not a hard request from your side we would like to keep this solution.

If it's really needed, we can add EVENT etc for it. Register the event in `src/draw/nema_gfx/lv_draw_nema_gfx_label.c` to handle it.

kisvegabor: I think @XuNeo is referring to [lv_array_resize](https://github.com/lvgl/lvgl/blob/master/src/misc/lv_array.h#L73C6-L73C21) which pre-allocates space for N items.

src/libs/freetype/lv_freetype_outline.c
```diff
@@ -354,6 +360,40 @@ static lv_freetype_outline_t outline_create(
         .delta = 0
     };
 
+#if LV_USE_NEMA_GFX
+#if LV_USE_NEMA_VG
+    /*Calculate Total Segmenets Before decompose */
+    int tag_size = face->glyph->outline.n_points;
```
FASTSHIFT: The outline path length information can be stored in the `lv_freetype_outline_event_param_t` structure, so that the length information can be read in `LV_EVENT_CREATE`, avoiding coupling of NEMA's private data structure in the freetype code.

FASTSHIFT (changes requested):

src/draw/nema_gfx/lv_draw_nema_gfx_label.c
```diff
@@ -95,6 +112,15 @@ void lv_draw_nema_gfx_label(lv_draw_unit_t * draw_unit, const lv_draw_label_dsc_
 
     nema_set_clip(clip_area.x1, clip_area.y1, lv_area_get_width(&clip_area), lv_area_get_height(&clip_area));
 
+#if LV_USE_FREETYPE && LV_USE_NEMA_VG
+    /*We need to initiallize in every new label task*/
+    static bool is_inited = false;
```
FASTSHIFT: You can refer to this PR: https://github.com/lvgl/lvgl/pull/7363 to make modifications without using the `is_inited` variable to implement delayed initialization.

src/libs/freetype/lv_freetype_outline.c
MarkIoanAmat: @XuNeo I will send a new event with the lengths of data points and segments, in that way we will avoid any reference to Nema specific structures, also anyone who wants can handle the event and get the data. I will update the code soon.

MarkIoanAmat: **Note:** As @FASTSHIFT suggested I will add the data needed in lv_freetype_outline_event_param_t structure, so that the length information can be read in LV_EVENT_CREATE

src/draw/nema_gfx/lv_draw_nema_gfx_label.c
uLipe: Please feel free to resolve it.

MarkIoanAmat: @uLipe  Thank you.

src/font/lv_font_fmt_txt.c
```diff
@@ -86,7 +86,11 @@ const void * lv_font_get_bitmap_fmt_txt(lv_font_glyph_dsc_t * g_dsc, lv_draw_buf
 
     const lv_font_fmt_txt_glyph_dsc_t * gdsc = &fdsc->glyph_dsc[gid];
 
-    if(g_dsc->req_raw_bitmap) return &fdsc->glyph_bitmap[gdsc->bitmap_index];
+    if(g_dsc->req_raw_bitmap) {
+        const uint8_t * raw_bitmap_out = &fdsc->glyph_bitmap[gdsc->bitmap_index];
+        draw_buf->data = (uint8_t *) raw_bitmap_out;
+        return draw_buf;
+    }
```
MarkIoanAmat: @kisvegabor Small update here. Just to return always the data in the draw_buf form.

uLipe (dismissed):

src/font/lv_font_fmt_txt.c
kisvegabor: It will lead to memory leak/crash. [Here](https://github.com/lvgl/lvgl/blob/master/src/draw/lv_draw_label.c#L538) we malloc `draw_buf->data`, if you overwrite it here, we won't be able to free it. :slightly_frowning_face:

FASTSHIFT (dismissed):

src/draw/nema_gfx/lv_draw_nema_gfx_label.c
```diff
@@ -267,14 +446,16 @@ static void _draw_label_iterate_characters(lv_draw_unit_t * draw_unit, const lv_
         pos.y += dsc->hint->y;
     }
 
-    uint32_t line_end = line_start + lv_text_get_next_line(&dsc->text[line_start], font, dsc->letter_space, w, NULL,
-                                                           dsc->flag);
+    uint32_t remaining_len = dsc->text_length;
```
FASTSHIFT, line 454: Why do we need to implement `_draw_label_iterate_characters` here? Why not use the general `lv_draw_label_iterate_characters` function?

MarkIoanAmat: @FASTSHIFT we had to do some changes inside those functions mainly for performance improvement reasons.

FASTSHIFT: What exactly has been modified? Can the optimization be put into the public part? 
Because the modification of the public part may not be synchronized to the private modification in time, it will take a lot of time to deal with code consistency issues: https://github.com/lvgl/lvgl/pull/7395

MarkIoanAmat: Yes you are right, we are aware of that. We are looking for other ways to achieve those optimizations, avoiding the modification of the public parts.

src/draw/nema_gfx/lv_draw_nema_gfx_img.c
```diff
@@ -170,13 +172,25 @@ static void _draw_nema_gfx_img(lv_draw_unit_t * draw_unit, const lv_draw_image_d
...
-        nema_bind_tex(NEMA_TEX3, (uintptr_t)NEMA_VIRT2PHYS(mask_buf), mask->header.w, mask->header.h,
-                      lv_nemagfx_mask_cf_to_nema(mask->header.cf),
-                      mask->header.stride, NEMA_FILTER_BL);
+    if(!has_transform && masked && !recolor) {
+        if(dsc->bitmap_mask_src->header.cf == LV_COLOR_FORMAT_A8 || LV_COLOR_FORMAT_L8) {
+            blending_mode |= NEMA_BLOP_STENCIL_TXTY;
+            const lv_image_dsc_t * mask = dsc->bitmap_mask_src;
+            const void * mask_buf = mask->data;
+
+            lv_area_t * image_area, mask_area;
```
liamHowatt: `image_area` should be const since `coords`, a const pointer, is being assigned to it.

Prefer one line per declaration.

```suggestion
            const lv_area_t * image_area,
            lv_area_t mask_area;
```

src/draw/nema_gfx/lv_draw_nema_gfx_img.c
```diff
@@ -230,7 +244,277 @@ static uint32_t lv_nemagfx_mask_cf_to_nema(lv_color_format_t cf)
         case LV_COLOR_FORMAT_A1:
             return NEMA_A1;
         case LV_COLOR_FORMAT_A2:
-            return NEMA_A2;
+            return NEMA_A2;/**
+ * MIT License
+ *
+ * -----------------------------------------------------------------------------
```
liamHowatt: It looks like the contents of lv_draw_nema_gfx_fill.c was accidentally pasted into the middle of this file.

liamHowatt (commented):
I think I chose the wrong time to review. It's clearly being worked on right now.

src/draw/nema_gfx/lv_draw_nema_gfx_img.c
```diff
@@ -230,7 +244,277 @@ static uint32_t lv_nemagfx_mask_cf_to_nema(lv_color_format_t cf)
...
+                }
+
+                if(!has_transform && masked && !recolor) {
+                    if(dsc->bitmap_mask_src->header.cf == LV_COLOR_FORMAT_A8 || LV_COLOR_FORMAT_L8) {
+                        blending_mode |= NEMA_BLOP_STENCIL_TXTY;
+                        const lv_image_dsc_t * mask = dsc->bitmap_mask_src;
+                        const void * mask_buf = mask->data;
+
+                        const lv_area_t * image_area;
+                        lv_area_t mask_area;
```
MarkIoanAmat: @liamHowatt here is your suggestion, no clue how I did that :) Working on it.

MarkIoanAmat:
@liamHowatt @kisvegabor Please for your review/comments.

src/draw/nema_gfx/lv_draw_nema_gfx_label.c
```diff
@@ -320,49 +503,132 @@ static void _draw_label_iterate_characters(lv_draw_unit_t * draw_unit, const lv_
...
     uint32_t nema_dsc_color = nema_rgba(dsc_col32.red, dsc_col32.green, dsc_col32.blue, dsc_col32.alpha);
     lv_color32_t dsc_sel_col32 = lv_color_to_32(dsc->sel_color, dsc->opa);
     uint32_t nema_dsc_sel_color = nema_rgba(dsc_sel_col32.red, dsc_sel_col32.green, dsc_sel_col32.blue,
                                             dsc_sel_col32.alpha);
-
+    uint32_t blend_color;
+    uint8_t blend_alpha = 255;
     _set_color_blend(nema_dsc_color, dsc_col32.alpha);
-
     uint32_t is_dsc_color_enabled = 1U;
```
liamHowatt: Also unused

src/draw/nema_gfx/lv_draw_nema_gfx_img.c
```diff
@@ -106,6 +106,8 @@ static void _draw_nema_gfx_img(lv_draw_unit_t * draw_unit, const lv_draw_image_d
     lv_layer_t * layer = draw_unit->target_layer;
     const lv_image_dsc_t * img_dsc = dsc->src;
 
+    bool masked = dsc->bitmap_mask_src != NULL ? true : false ;
```
liamHowatt:
```suggestion
    bool masked = dsc->bitmap_mask_src != NULL;
```

src/draw/nema_gfx/lv_draw_nema_gfx_img.c
```diff
@@ -170,13 +172,26 @@ static void _draw_nema_gfx_img(lv_draw_unit_t * draw_unit, const lv_draw_image_d
...
 
-    if(dsc->bitmap_mask_src != NULL) {
-        blending_mode |= NEMA_BLOP_STENCIL_TXTY;
-        const lv_image_dsc_t * mask = dsc->bitmap_mask_src;
-        const void * mask_buf = mask->data;
-        nema_bind_tex(NEMA_TEX3, (uintptr_t)NEMA_VIRT2PHYS(mask_buf), mask->header.w, mask->header.h,
-                      lv_nemagfx_mask_cf_to_nema(mask->header.cf),
-                      mask->header.stride, NEMA_FILTER_BL);
+    if(!has_transform && masked && !recolor) {
+        if(dsc->bitmap_mask_src->header.cf == LV_COLOR_FORMAT_A8 || LV_COLOR_FORMAT_L8) {
```
liamHowatt:
```suggestion
        if(dsc->bitmap_mask_src->header.cf == LV_COLOR_FORMAT_A8 || dsc->bitmap_mask_src->header.cf == LV_COLOR_FORMAT_L8) {
```

src/draw/nema_gfx/lv_nema_gfx_path.c
```diff
@@ -0,0 +1,151 @@
...
+    path->data[data_point++] = y;
+}
+
+void lv_nema_gfx_path_end(lv_nema_gfx_path_t * path)
+{
+    /* Do Path end jobs....whatever*/
+    seg_point = 0;
+    data_point = 0;
+
+}
```
liamHowatt, line 156: Is this finished?

src/draw/nema_gfx/lv_nema_gfx_path.c
```diff
@@ -0,0 +1,151 @@
...
+#if LV_USE_NEMA_VG
+
+#include "lv_nema_gfx_path.h"
+
+/**********************
+ *   GLOBAL FUNCTIONS
+ **********************/
+
+static int data_point = 0;
+static int seg_point  = 0;
```
liamHowatt, line 47: Can these move into `lv_nema_gfx_path_t`?

src/draw/nema_gfx/lv_nema_gfx_path.h
```diff
@@ -0,0 +1,97 @@
...
+    int seg_size;
+} lv_nema_gfx_path_t;
+
+/**********************
+ * GLOBAL PROTOTYPES
+ **********************/
+
+lv_nema_gfx_path_t * lv_nema_gfx_path_create(void);
+
+void lv_nema_gfx_path_alloc(lv_nema_gfx_path_t * nema_gfx_path);
```
liamHowatt, line 67: Can these be combined?

src/misc/lv_types.h
```diff
@@ -269,6 +269,8 @@ typedef struct _lv_qrcode_t lv_qrcode_t;
 
 typedef struct _lv_freetype_outline_vector_t lv_freetype_outline_vector_t;
 
+typedef struct _lv_freetype_outline_sizes_t lv_freetype_outline_sizes_t;
```
liamHowatt: Typically, fully private types do not go in lv_types.h.

src/libs/freetype/lv_freetype_private.h
```diff
@@ -65,12 +65,18 @@ struct _lv_freetype_outline_vector_t {
     int32_t y;
 };
 
+struct _lv_freetype_outline_sizes_t {
+    int32_t segments_size;
+    int32_t data_size;
+};
```
liamHowatt:
```suggestion
typedef struct {
    int32_t segments_size;
    int32_t data_size;
} lv_freetype_outline_sizes_t;
```

src/libs/freetype/lv_freetype_outline.c
```diff
@@ -331,10 +331,48 @@ static lv_freetype_outline_t outline_create(
...
+                segments++;
+                vectors += 2;
+            }
+        }
+    }
+    /*Also for every contour we may have a line for close*/
+    segments += face->glyph->outline.n_contours;
+    vectors += face->glyph->outline.n_contours;
+
+    param.sizes.data_size = vectors * 2;
```
liamHowatt, line 373: Sorry, just for my understanding, why is it `* 2`?

liamHowatt (commented):
There was a concern in #7395 about using the private text API. I see that a few more functions are still used. It's also related to [FASTSHIFT's comment](https://github.com/lvgl/lvgl/pull/7346#discussion_r1867234165). They are:
```
lv_text_get_next_line
lv_text_encoded_get_char_id
lv_text_encoded_letter_next_2
lv_text_is_marker
```

<hr>

Static function `get_glyph_dsc_id` in lv_draw_nema_gfx_label.c is unused now that it's no longer being called.

src/libs/freetype/lv_freetype_outline.c
MarkIoanAmat: We count the number of floats, so it is vectors * 2

src/draw/nema_gfx/lv_nema_gfx_path.h
MarkIoanAmat: There will be cases afterwards on accelerating Vector Graphics where we may need both

src/draw/nema_gfx/lv_nema_gfx_path.c
MarkIoanAmat: By now is is

src/draw/nema_gfx/lv_nema_gfx_path.c
MarkIoanAmat: No we need them as static as they are incremental during the population of arrays. Inside `lv_nema_gfx_path_t `we have the full length.

src/draw/nema_gfx/lv_draw_nema_gfx_label.c
MarkIoanAmat: @liamHowatt I will remove the unused functions. Also,  will create new PR regarding the decoupling of our implementation from the private text API. Lets proceed with this one for the time being.

src/font/lv_font_fmt_txt.c
MarkIoanAmat: If user specifies `req_raw_bitmap = 1` how could we know if the glyph data is uint8_t * or a draw buffer ? For example using the default font e.g `LV_FONT_MONTSERRAT_14`  in `lv_draw_glyph_dsc_t` format is `LV_FONT_GLYPH_FORMAT_A4` for both cases, either  I choose `req_raw_bitmap = 1` or not.

kisvegabor: With `req_raw_bitmap = 1` the returned pointer of `lv_font_get_glyph_bitmap` is pointer to the raw bitmap and `draw_buf` is not used.

MarkIoanAmat: Currently `req_raw_bitmap = 1` is only valid for `lv_font_fmt_txt` type of font descriptors. What if we use several font descriptors ? For example if we use ttf abd we have requested `req_raw_bitmap = 1` We will wait for a uint8_t * in glyph data again but there a we will have a draw_buf instead.

MarkIoanAmat: ** I meant Tiny TTF in my previous post **
So either we have to also check for the font descriptor type e.g `lv_font_get_bitmap_fmt_txt` or to my understanding the user should only care about the font format and read it always in the dame data structure, if possible, whe we talk about LV_FONT_GLYPH_FORMAT_A1 to LV_FONT_GLYPH_FORMAT_A8_ALIGNED formats

MarkIoanAmat: update: Font information is not available inside draw_letter_cb, so we can not check about the origin of the font descriptor. So finally we are missing a way finding if the glyph data retrieved from `lv_font_get_glyph_bitmap` is uint8_t * or other struct when we have `req_raw_bitmap` enabled.

uLipe (dismissed):

liamHowatt (changes requested):
I think MarkIoanAmat is still addressing the feedback, uLipe

MarkIoanAmat:
@liamHowatt Just finished with the proposed changes. Waiting for your review.
@kisvegabor

uLipe (dismissed):

src/draw/nema_gfx/lv_draw_nema_gfx_label.c
```diff
@@ -64,18 +84,36 @@ static inline uint8_t _bpp_nema_gfx_format(lv_draw_glyph_dsc_t * glyph_draw_dsc)
...
                          const lv_font_t * font, uint32_t letter);
 
-static uint32_t get_glyph_dsc_id(const lv_font_t * font, uint32_t letter);
+static uint8_t hex_char_to_num(char hex);
 
-static int unicode_list_compare(const void * ref, const void * element)
-{
-    return ((int32_t)(*(uint16_t *)ref)) - ((int32_t)(*(uint16_t *)element));
-}
+static bool is_raw_bitmap;
```
liamHowatt, line 89: Please move it into the draw unit struct if possible.

kisvegabor:
cc @FASTSHIFT @XuNeo pleasae take a look at this PR.

src/draw/nema_gfx/lv_draw_nema_gfx_label.c
MarkIoanAmat: @liamHowatt lets proceed with this PR as it refers to freetype support. I will reform  lv_draw_nema_gfx_label.c in next PR including the #7395. Thank you.

src/draw/nema_gfx/lv_nema_gfx_path.c
```diff
@@ -0,0 +1,151 @@
...
+
+lv_nema_gfx_path_t * lv_nema_gfx_path_create(void)
+{
+    LV_PROFILER_BEGIN;
```
FASTSHIFT: Please ues LV_PROFILER_DRAW_BEGIN/END

src/draw/nema_gfx/lv_nema_gfx_path.c
```diff
@@ -0,0 +1,151 @@
...
+    lv_free(nema_gfx_path);
+    LV_PROFILER_END;
+}
+
+void lv_nema_gfx_path_move_to(lv_nema_gfx_path_t * path, float x, float y)
+{
+    LV_ASSERT_NULL(path);
+    path->seg[seg_point++] = NEMA_VG_PRIM_MOVE;
+    path->data[data_point++] = x;
+    path->data[data_point++] = y;
```
FASTSHIFT, line 110: Can we add an array out of bounds check? It is necessary from the point of view of project experience.

src/draw/nema_gfx/lv_nema_gfx_path.h
```diff
@@ -0,0 +1,97 @@
...
+/**********************
+ *      TYPEDEFS
+ **********************/
+
+typedef struct {
+    NEMA_VG_PATH_HANDLE path;
+    NEMA_VG_PAINT_HANDLE paint;
+    float * data;
+    uint8_t * seg;
+    int data_size;
```
FASTSHIFT: It is recommended to use `uint32_t` to ensure that the data length is consistent across platforms.

src/libs/freetype/lv_freetype_outline.c
```diff
@@ -331,10 +331,48 @@ static lv_freetype_outline_t outline_create(
...
+
+
     lv_result_t res;
     lv_freetype_outline_event_param_t param;
-
     lv_memzero(&param, sizeof(param));
+
+    /*Calculate Total Segmenets Before decompose */
+    int tag_size = face->glyph->outline.n_points;
+    int segments = 0;
```
FASTSHIFT: int32_t

tests/src/test_cases/libs/test_freetype.c
```diff
@@ -29,359 +29,359 @@ static const char * UNIVERSAL_DECLARATION_OF_HUMAN_RIGHTS_JP =
...
-    {NULL, 2, {4428, 1950}, {0, 0}, {0, 0}},
-    {NULL, 1, {5370, 3023}, {0, 0}, {0, 0}},
-    {NULL, 2, {5739, 3105}, {0, 0}, {0, 0}},
-    {NULL, 4, {5825, 2855}, {5790, 2998}, {0, 0}},
-    {NULL, 4, {5872, 2613}, {5860, 2712}, {0, 0}},
-    {NULL, 2, {5479, 2499}, {0, 0}, {0, 0}},
-    {NULL, 4, {5434, 2752}, {5469, 2605}, {0, 0}},
-    {NULL, 4, {5370, 3023}, {5400, 2900}, {0, 0}},
-    {NULL, 0, {0, 0}, {0, 0}, {0, 0}},
+    {NULL, 1, {5888, 6889}, {0, 0}, {0, 0}, {0, 0}},
```
FASTSHIFT, line 32: You can try to update this value. You can use `TEST_PRINTF` [here](https://github.com/lvgl/lvgl/blob/master/tests/src/test_cases/libs/test_freetype.c#L525) to print out `segments_size` and `data_size` and update them to the `outline_data_U9F98` array.

FASTSHIFT: You can use ChatGPT or python script to regenerate code based on the `TEST_PRINTF` log content.

MarkIoanAmat: I populate segments_size and data_size only once in the `LV_EVENT_CREATE `. In each push point created, segments_size  and data_size are always zero as param is always initialized inside [here](https://github.com/lvgl/lvgl/blob/master/src/libs/freetype/lv_freetype_outline.c#L247)

MarkIoanAmat: Inside push point we don't have the information of `segments_size `and `data_size` in order to send them again in each push point to do the comparison with `outline_data_U9F98`

kisvegabor (approved):
As it seems there is an agreement on the generic FreeType part I suggest merging it now. It will receive further updates regarding Nema in the near future.

FASTSHIFT (approved):

kisvegabor:
We have discussed it Liam that it can be merged.

PR #7354, opened 2024-11-22
fix(indev): prevent division by zero

pfvogel:
Add checks to the scale values to insure that they are not zero when the scale is used as the denominator in calculations.


kisvegabor (approved):
Thank you!

FASTSHIFT (approved):

PR #7368, opened 2024-11-28
feat(windows): add more personalized flags (frameless, top_level) to the windows emulator

cheungxiongwei:
Previously, the `WS_EX_CLIENTEDGE` flag was used, which had its own border, affecting the visual appearance of the frameless window. Now a flat style is used to facilitate users to customize the frameless window.

[normal Figure 1, befor Figure 2, after Figure 3]

Snipaste_2024-11-28_14-12-42

### Notes


FASTSHIFT (dismissed):

FASTSHIFT:
cc @MouriNaruto please take a look

MouriNaruto:
Here are my opinions:

- Exposing the "lv_windows_center_window" API to the public instead of private may be a good idea for some users. However the API should be more generalized. For example, I suggest the prototype of this API should be the following definition:
  > void lv_windows_center_window(lv_display_t * display, HWND parent_window);
- We should not break the LVGL Windows backend APIs. (at least before LVGL v10, a.k.a. keep the ABI compatibility at least for every major version.) It seems your addition to the "lv_windows_create_display" API can implemented with Windows SetWindowLongPtr API which can called after the window creation. So, I think we should expose APIs which accept the bool value when calling with the names like lv_windows_set_top_level and lv_windows_set_frameless.

Kenji Mouri

kisvegabor:
> So, I think we should expose APIs which accept the bool value when calling with the names like "lv_windows_set_top_level" and "lv_windows_set_frameless".

I agree with it.

cheungxiongwei:
It might be better not to expose the `lv_windows_center_window` function directly to users. If it is exposed, you may need to create a new `lv_windows_show_window()` function. Additionally, users would then need to call the `lv_windows_center_window` function between `lv_windows_create_display` and `lv_windows_show_window` 。

This increases the possibility of novice users making mistakes, so I recommend using the centering strategy by default. Let users focus more on the lvgl content.

cheungxiongwei:
```c
 void lv_windows_set_top_level(lv_display_t * display,bool top_level) {
    HWND window_handle = lv_windows_get_display_window_handle(display);

    LONG_PTR ex_style = GetWindowLongPtr(window_handle, GWL_EXSTYLE);
    if (top_level) {
        ex_style |= WS_EX_TOPMOST;
        ex_style |= WS_EX_APPWINDOW;
    }else {
        ex_style &= ~WS_EX_TOPMOST;
    }

    SetWindowLongPtr(window_handle, GWL_EXSTYLE, ex_style);
    SetWindowPos(window_handle, top_level ? HWND_TOPMOST : HWND_NOTOPMOST, 0, 0, 0, 0,SWP_NOMOVE | SWP_NOSIZE | SWP_NOACTIVATE| SWP_FRAMECHANGED);
}

void lv_windows_set_frameless(lv_display_t * display,bool frameless) {
    HWND window_handle = lv_windows_get_display_window_handle(display);
    
    LONG_PTR style = GetWindowLongPtr(window_handle, GWL_STYLE);
    if (frameless) {
        style &= ~WS_OVERLAPPEDWINDOW;
        style |= WS_POPUP;
    }else {
        style &= ~WS_POPUP;
        style |= WS_OVERLAPPEDWINDOW;
    }

    SetWindowLongPtr(window_handle, GWL_STYLE, style);
    SetWindowPos(window_handle, NULL, 0, 0, 0, 0,SWP_NOMOVE | SWP_NOSIZE | SWP_NOZORDER | SWP_FRAMECHANGED);
}
```
After testing, lv_windows_set_top_level can be called normally, while lv_windows_set_frameless will cause content rendering errors when called after the window is created.

It is worth noting that these two functions cannot be called in a locked scope. They need to be unlocked before calling, otherwise they will deadlock.

example 1
```c
lv_lock();
// ... other code
    lv_windows_set_top_level(display, top_level); // error
lv_unlock();
```

example 2
```c
lv_windows_set_top_level(display, top_level); // ok

lv_lock();
// ... other code
lv_unlock();
```

example 3
```c
lv_lock();
// ... other code
    lv_unlock();
    lv_windows_set_top_level(display, top_level); // ok
    lv_lock();
lv_unlock();
```

cheungxiongwei:
This `main.c` is an MVP.
```c
#include <lvgl.h>
#include <lvgl/src/osal/lv_os.h>
#include <stdio.h>

static void switch_frameless_cb(lv_event_t *e) {
    bool checked = lv_obj_has_state((lv_obj_t*)lv_event_get_target(e),LV_STATE_CHECKED);
    printf("Switch FrameLess:%d\n",checked);

    lv_unlock();
    lv_windows_set_frameless(lv_display_get_default(),checked);
    lv_lock();
}

static void switch_toplevel_cb(lv_event_t *e) {
    bool checked = lv_obj_has_state((lv_obj_t*)lv_event_get_target(e),LV_STATE_CHECKED);
    printf("Switch TopLevel: %d\n",checked);
    lv_unlock();
    lv_windows_set_top_level(lv_display_get_default(),checked);
    lv_lock();
}

int main()
{
    lv_init();

    int32_t zoom_level = 100;
    bool allow_dpi_override = false;
    bool simulator_mode = false;
    bool top_level = false;
    bool frameless = false;
    lv_display_t* display = lv_windows_create_display(
            L"LVGL Display Window",
            600,
            600,
            zoom_level,
            allow_dpi_override,
            simulator_mode
            );
    if (!display)
    {
        return -1;
    }

    lv_windows_set_frameless(display, frameless);
    lv_windows_set_top_level(display, top_level);
    
    lv_indev_t* pointer_device = lv_windows_acquire_pointer_indev(display);
    if (!pointer_device)
    {
        return -1;
    }

    lv_indev_t* keypad_device = lv_windows_acquire_keypad_indev(display);
    if (!keypad_device)
    {
        return -1;
    }

    lv_indev_t* encoder_device = lv_windows_acquire_encoder_indev(display);
    if (!encoder_device)
    {
        return -1;
    }

    lv_obj_t* screen = lv_obj_create(lv_screen_active());
    lv_obj_remove_style_all(screen);
    lv_obj_set_size(screen,LV_PCT(100),LV_PCT(100));
    lv_obj_set_scroll_dir(screen,LV_DIR_NONE);
    lv_obj_set_style_bg_color(screen, lv_color_hex(0x2b2d30), LV_PART_MAIN);
    lv_obj_set_style_bg_opa(screen, LV_OPA_COVER, LV_PART_MAIN);

    lv_obj_set_flex_flow(screen,LV_FLEX_FLOW_ROW);
    lv_obj_set_flex_align(screen,LV_FLEX_ALIGN_CENTER,LV_FLEX_ALIGN_CENTER,LV_FLEX_ALIGN_SPACE_EVENLY);
    lv_obj_set_flex_grow(screen,12);

    lv_obj_set_style_text_color(screen,lv_color_white(),LV_PART_MAIN);
    lv_obj_set_style_pad_gap(screen,12,0);

    // [1]
    lv_obj_t* label_frameless = lv_label_create(screen);
    lv_label_set_text(label_frameless, "Switch FrameLess:");
    lv_obj_add_event_cb(lv_switch_create(screen),switch_frameless_cb,LV_EVENT_VALUE_CHANGED,0);

    // [2]
    lv_obj_t *label_toplevel = lv_label_create(screen);
    lv_label_set_text(label_toplevel, "Switch TopLevel:");
    lv_obj_add_event_cb(lv_switch_create(screen),switch_toplevel_cb,LV_EVENT_VALUE_CHANGED,0);

    lv_unlock();

    while (1)
    {
        uint32_t time_till_next = lv_timer_handler();
        lv_delay_ms(time_till_next);
    }

    return 0;
}
```

src/drivers/windows/lv_windows_display.c
```diff
@@ -115,6 +119,26 @@ int32_t lv_windows_dpi_to_physical(int32_t logical, int32_t dpi)
  *   STATIC FUNCTIONS
  **********************/
 
+static void __stdcall lv_windows_center_window(HWND hwnd)
```
uLipe: I noticed the Win API calls have its return not checked, as well this function is void, if changing it breaks the API would not be possible to at least put LOG_WARNIG/LOG_ERR to track possible erros that may happen? 

I'm afraid of silent failure here may become hard to debug.

uLipe (commented):
LGTM so far, I have just a minor question / concern

src/drivers/windows/lv_windows_display.c
MouriNaruto: I agree with you. Also, I think we should not use stdcall to define LVGL APIs.

Kenji Mouri

uLipe: if stdcall is required in windows platform you can define a noop version when not in Windows version

MouriNaruto: I don't think stdcall is necessary when not calling the Windows APIs.

cheungxiongwei: You are right, I have made the changes accordingly.

kisvegabor:
```c
 bool top_level = false;
    bool frameless = false;
    lv_display_t* display = lv_windows_create_display(
            L"LVGL Display Window",
            600,
            600,
            zoom_level,
            allow_dpi_override,
            simulator_mode
            );
    if (!display)
    {
        return -1;
    }

    lv_windows_set_frameless(display, frameless);
    lv_windows_set_top_level(display, top_level);
```

This direction look good to me. Please update the code accordingly.

src/drivers/windows/lv_windows_display.c
```diff
@@ -126,12 +190,14 @@ static unsigned int __stdcall lv_windows_display_thread_entrypoint(
         window_style &= ~(WS_SIZEBOX | WS_MAXIMIZEBOX | WS_THICKFRAME);
     }
 
+    DWORD window_ex_style =  WS_EX_APPWINDOW;
+
     HWND window_handle = CreateWindowExW(
-                             WS_EX_CLIENTEDGE,
+                             window_ex_style,
```
MouriNaruto: It seems we can use WS_EX_APPWINDOW instead of declaring window_ex_style variable.

src/drivers/windows/lv_windows_display.c
```diff
@@ -143,6 +209,8 @@ static unsigned int __stdcall lv_windows_display_thread_entrypoint(
         return 0;
     }
 
+    LV_ASSERT(lv_windows_center_window(window_handle));
```
MouriNaruto, line 230: You need to mention this behavior change in the comment of lv_windows_create_display API.

src/drivers/windows/lv_windows_display.c
```diff
@@ -111,10 +111,74 @@ int32_t lv_windows_dpi_to_physical(int32_t logical, int32_t dpi)
...
+
+    SetWindowLongPtr(window_handle, GWL_STYLE, style);
+    SetWindowPos(window_handle, NULL, 0, 0, 0, 0, SWP_NOMOVE | SWP_NOSIZE | SWP_NOZORDER | SWP_FRAMECHANGED);
+}
+
 /**********************
  *   STATIC FUNCTIONS
  **********************/
 
+static int lv_windows_center_window(HWND hwnd)
```
MouriNaruto: Maybe we should use window_handle instead of hwnd to improve readability and keep the coding style.

Also, some variables like windowWidth should use snake case like window_width to keep the coding style.

src/drivers/windows/lv_windows_display.c
```diff
@@ -111,10 +111,74 @@ int32_t lv_windows_dpi_to_physical(int32_t logical, int32_t dpi)
...
+    if(top_level) {
+        ex_style |= WS_EX_TOPMOST;
+        ex_style |= WS_EX_APPWINDOW;
+    }
+    else {
+        ex_style &= ~WS_EX_TOPMOST;
+    }
+
+    SetWindowLongPtr(window_handle, GWL_EXSTYLE, ex_style);
+    SetWindowPos(window_handle, top_level ? HWND_TOPMOST : HWND_NOTOPMOST, 0, 0, 0, 0,
```
MouriNaruto: We should try best to keep one parameter for each line when we calling the function to improve the code readability. You also should keep maximum 80 half-wide characters per line.

For example, learn the coding style when I calling CreateWindowEx in lv_windows_display_thread_entrypoint function.
```
    HWND window_handle = CreateWindowExW(
                             WS_EX_CLIENTEDGE,
                             L"LVGL.Window",
                             data->title,
                             window_style,
                             CW_USEDEFAULT,
                             0,
                             data->hor_res,
                             data->ver_res,
                             NULL,
                             NULL,
                             NULL,
                             data);
```

There are some similar code snippets you can improve.

docs/details/integration/driver/windows.rst
```diff
@@ -82,6 +84,9 @@ Usage
             return -1;
         }
 
+        lv_windows_set_frameless(display, frameless);
+        lv_windows_set_top_level(display, top_level);
+
         lv_lock();
```
liamHowatt, line 90: These should be moved after the lock. Even though these APIs are thread safe now, they may not always be.

liamHowatt:
@MouriNaruto does it all look ok now? If the API I mentioned is thread safe then [my comment](https://github.com/lvgl/lvgl/pull/7368#discussion_r1918906415) can be ignored and we can approve and merge.

docs/details/integration/driver/windows.rst
MouriNaruto: I agree with your concern.

PR #7389, opened 2024-12-01
Do not squash. support second value in property

XuNeo:
Support two simple value(4byte) in one property.

Search though all the widget set APIs, two value seems enough. With exception for lv_table `lv_table_set_cell_value` etc. that has 3 parameters: col, row, value.

This method is more simple compare to PR https://github.com/lvgl/lvgl/pull/7382 .
And most importantly, this method is still able to be automated for bindings: we can get all the value types from ID and construct the property value correspondingly, without need to handle it case by case.

Tested with roller `lv_roller_set_options` as shown in the test case.

### Notes


FASTSHIFT (dismissed):

src/widgets/slider/lv_slider.c
```diff
@@ -49,6 +49,52 @@ static void update_knob_pos(lv_obj_t * obj, bool check_drag);
...
+    {
+        .id = LV_PROPERTY_SLIDER_MIN_VALUE,
+        .setter = NULL,
+        .getter = lv_slider_get_min_value,
+    },
+    {
+        .id = LV_PROPERTY_SLIDER_MAX_VALUE,
+        .setter = NULL,
+        .getter = lv_slider_get_max_value,
+    },
```
XuNeo, line 79: Please do note that `RANGE` property supports two values for min and max, but the getter does not support. This matches with the C code that has to use MIN_VALUE, and MAX_VALUE.

kisvegabor:
It's really nice! A few notes:
- [Calendar APIs](https://github.com/lvgl/lvgl/blob/master/src/widgets/calendar/lv_calendar.h#L65) also have 3 arguments
- How can we handle e.g. `void lv_span_set_text(lv_span_t * span, const char * text)` where the `span` is not a widget just some arbitrary data  
- Just to confirm: widget creation is out of scope here, right? (We assume that widget is already created). Some example where it might be interesting. Add a tab to the tabview or add a series to the chart. 
- Can we pass `const char * map[] = {"a", "b", "\n"}` to a button matrix?

W-Mai (approved):

XuNeo:
> [Calendar APIs](https://github.com/lvgl/lvgl/blob/master/src/widgets/calendar/lv_calendar.h#L65) also have 3 arguments

The first one is obj itself, thus two values for one property, which is covered.

> lv_span_set_text(lv_span_t * span, const char * text)

span is not an widget, thus cannot use the widget property API.

> widget creation is out of scope here, right? 

Yes. property only takes care of property. The widget creation has other tricks. For example, dynamically registering all the  widgets class to lvgl.
```c
lv_obj_t * lv_obj_create_from_name(const char* class_name, parent)
{
  iterate_lv_obj_classes(lvgl, clz) {
    if(clz->name == class_name){
        return lv_obj_class_create_obj(parent, clz);
    }
  }
}

```

> Can we pass const char * map[] = {"a", "b", "\n"} to a button matrix?

Yes. It can be passed as a raw `void*` pointer.

kisvegabor:
> The first one is obj itself, thus two values for one property, which is covered.

Sorry, I meant obj + 3 others. See

```c
void lv_calendar_set_today_date(lv_obj_t * obj, uint32_t year, uint32_t month, uint32_t day);
```

> span is not an widget, thus cannot use the widget property API.

Are there any ways to support them too? Maybe we can change something in LVGL itself?

liamHowatt (commented):
I notice you're using `LV_ANIM_ON/OFF` and `true/false` interchangeably in the tests for `enable2` when it's an anim enable param. Also, `enable` is a bool but `lv_anim_enable_t` is an enum which is `int`-sized. Maybe there needs to be a union member `int enum;`.

Edit: I understand that on modern architectures, including 64 bit, `bool` parameters use a full register or 4 bytes on the stack so it's not a serious issue.

src/core/lv_obj_property.h
```diff
@@ -83,12 +88,28 @@ struct _lv_property_name_t {
...
     lv_prop_id_t id;
     union {
-        int32_t num;                /**< Number integer number (opacity, enums, booleans or "normal" numbers)*/
-        bool enable;                /**< booleans*/
-        const void * ptr;           /**< Constant pointers  (font, cone text, etc)*/
-        lv_color_t color;           /**< Colors*/
-        lv_value_precise_t precise; /**< float or int for precise value*/
-        lv_point_t point;           /**< Point*/
+        struct {
+            union {
```
uLipe: num + num_2 don't feels that means two properties with two values, not a blocker but for clarification why not suggest something like C++ does for their map structure by prepending first_ and second_ prefixes for the properties?

for example we would have first_num, and second_num , which also would enforce the correct order of filling the properties.

liamHowatt:
> Thank you Liam for noticing it. I mistakenly mixing them. How about change the enum to define LV_ANIM_ON as true and OFF as false?

At first thought is sounds like a bad idea but the more I think about it, I think it could be ok. I approve that approach.

src/core/lv_obj_property.h
kisvegabor: cc @XuNeo

kisvegabor (dismissed):
As this feature is an ongoing development and it's not the final state, I can approve it.

FASTSHIFT (dismissed):

src/core/lv_obj_property.h
XuNeo: I makes sense. Instead of using first and second, I simply put arg1 and arg2 so when two args are needed, it's specified explicitly.

E.g.
```
    prop.id = LV_PROPERTY_ROLLER_OPTIONS;
    prop.arg1.ptr = "One\nTwo\nThree";
    prop.arg2.num = LV_ROLLER_MODE_NORMAL;
    lv_obj_set_property(obj, &prop);
```

uLipe (approved):

kisvegabor (approved):

kisvegabor:
Thank you!

FYI, I was able to not squash it again :smile:

PR #7398, opened 2024-12-03
feat(vg_lite): reduce unnecessary path quality settings

FASTSHIFT:
Set the default path quality to HIGH to avoid aliasing. The current rendering speed bottleneck is no longer GPU.

### Notes

### DIFF
--- a/src/draw/vg_lite/lv_draw_vg_lite_arc.c
+++ b/src/draw/vg_lite/lv_draw_vg_lite_arc.c
@@ -83,7 +83,6 @@ void lv_draw_vg_lite_arc(lv_draw_unit_t * draw_unit, const lv_draw_arc_dsc_t * d
     LV_PROFILER_DRAW_BEGIN;
 
     lv_vg_lite_path_t * path = lv_vg_lite_path_get(u, VG_LITE_FP32);
-    lv_vg_lite_path_set_quality(path, VG_LITE_HIGH);
     lv_vg_lite_path_set_bounding_box_area(path, &clip_area);
 
     float radius_out = dsc->radius;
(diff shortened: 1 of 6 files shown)


kisvegabor:
>  The current rendering speed bottleneck is no longer GPU.

Is it memory, CPU, or LVGL overhead now?

kisvegabor:
cc @uLipe

FASTSHIFT:
> > The current rendering speed bottleneck is no longer GPU.
> 
> Is it memory, CPU, or LVGL overhead now?

Currently, in our application scenario, the CPU is the performance bottleneck. Using the [GPU idle flush](Currently, in our application scenario, the CPU is the performance bottleneck, especially when the component complexity is high.
Using the GPU idle flush mechanism can make the GPU and CPU run in parallel.) mechanism can make the GPU and CPU run in parallel.

XuNeo (approved):
Can we keep it as a configuration option in case we need it later.

uLipe (commented):
I like the idea, but as @XuNeo mentioned, this option could be a choice that can be selected in the lv_conf.h at build time, please notice using high quality paths tends to increase the current consumption which is not desired for some battery powered applications, so I think users might wanted this option could be selectable at least in build time to avoid them to go into the LVGL source tree and modify it by themselves,.

FASTSHIFT:
> Can we keep it as a configuration option in case we need it later.

I would like to consider introducing configuration items only when there is a clear demand. Currently, image quality has a higher priority.

FASTSHIFT:
The current modification does not affect the current rendering efficiency, because the previous drawing layer (such as draw arc, draw letter) code will actively set the quality to high. This modification only sets the default rendering quality to high, so that the drawing layer code can be less.

liamHowatt (approved):
Are you sure `path->base.quality` isn't uninitialized in some cases?

FASTSHIFT:
> Are you sure `path->base.quality` isn't uninitialized in some cases?

Yes, path creation and destruction are unified interfaces.

Issue #7372, opened 2024-11-29
Font Converter Error


Allen-azh:
### LVGL version

v8.2.0

### Platform

Coter-M4F  with freertos.

### What happened?

1. Use Font Converter to generate a library containing eight Chinese characters in bin format. (计订认讥讨让讪讫)
2. After loading the font library, use lv_label_set_text(label, "讧"); Set the lable.
Ideally, lvgl reports an error indicating that the font library cannot be found.
Current situation: lvgl has drawn "讧" using the glyph of "计".
![Snipaste_2024-11-29_17-05-46](https://github.com/user-attachments/assets/a6db9699-8fab-4bef-888b-ddbe4e8b5a57)
[font.zip](https://github.com/user-attachments/files/17956874/font.zip)

Allen-azh:
@kisvegabor

Allen-azh:
https://github.com/lvgl/lvgl/pull/7404#issuecomment-2552226223

have figured out

PR #7404, opened 2024-12-03
fix(font): fix get glyph id bug.

Allen-azh:
For more details, see [Font Converter Error #7372](https://github.com/lvgl/lvgl/issues/7372)

### DIFF
--- a/src/font/lv_font_fmt_txt.c
+++ b/src/font/lv_font_fmt_txt.c
@@ -270,7 +270,7 @@ static uint32_t get_glyph_dsc_id(const lv_font_t * font, uint32_t letter)
         }
         else if(fdsc->cmaps[i].type == LV_FONT_FMT_TXT_CMAP_FORMAT0_FULL) {
             const uint8_t * gid_ofs_8 = fdsc->cmaps[i].glyph_id_ofs_list;
-            if(0 == gid_ofs_8[rcp] && letter != fdsc->cmaps[i].range_start) continue;
+            if(gid_ofs_8[rcp] == 0 && letter != fdsc->cmaps[i].range_start) continue;
             glyph_id = fdsc->cmaps[i].glyph_id_start + gid_ofs_8[rcp];
         }
         else if(fdsc->cmaps[i].type == LV_FONT_FMT_TXT_CMAP_SPARSE_TINY) {
--- a/src/font/lv_font_fmt_txt.c
+++ b/src/font/lv_font_fmt_txt.c
@@ -270,6 +270,9 @@ static uint32_t get_glyph_dsc_id(const lv_font_t * font, uint32_t letter)
         }
         else if(fdsc->cmaps[i].type == LV_FONT_FMT_TXT_CMAP_FORMAT0_FULL) {
             const uint8_t * gid_ofs_8 = fdsc->cmaps[i].glyph_id_ofs_list;
+            /* The first character is always valid and should have offset = 0
+             * However if a character is missing it also has offset=0.
+             * So if there is a 0 not on the first position then it's a missing character */
             if(gid_ofs_8[rcp] == 0 && letter != fdsc->cmaps[i].range_start) continue;
             glyph_id = fdsc->cmaps[i].glyph_id_start + gid_ofs_8[rcp];
         }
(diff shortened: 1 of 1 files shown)


src/font/lv_font_fmt_txt.c
```diff
@@ -270,6 +270,7 @@ static uint32_t get_glyph_dsc_id(const lv_font_t * font, uint32_t letter)
         }
         else if(fdsc->cmaps[i].type == LV_FONT_FMT_TXT_CMAP_FORMAT0_FULL) {
             const uint8_t * gid_ofs_8 = fdsc->cmaps[i].glyph_id_ofs_list;
+            if(0 == gid_ofs_8[rcp] && letter != fdsc->cmaps[i].range_start) continue;
```
FASTSHIFT: Please use `gid_ofs_8[rcp] == 0` make it readable. Modern compilers can also warn about misspelled equals signs.

FASTSHIFT:
cc @W-Mai

src/font/lv_font_fmt_txt.c
kisvegabor:
```suggestion
            if(gid_ofs_8[rcp] == 0 && letter != fdsc->cmaps[i].range_start) continue;
```

kisvegabor: I've just fixed it.

kisvegabor:
The proposed solution seems more like a workaround but a not a real solution to the problem. Does it work with a font converted to C? 

I've taken a deeper look and it seems `glyph_id_ofs_list` is generated incorrectly. These `0`s in the array look bad. It should be like 0, 1, 2, 3, 4, 5 ...
image

Allen-azh:
> it work with a font converted to C

Yes, it work with a font converted to C.

Allen-azh:
> The proposed solution seems more like a workaround but a not a real solution to the problem. Does it work with a font converted to C?提出的解决方案似乎更像是一种解决方法，但并不是问题的真正解决方案。它适用于转换为 C 的字体吗？
> 
> I've taken a deeper look and it seems `glyph_id_ofs_list` is generated incorrectly. These `0`s in the array look bad. It should be like 0, 1, 2, 3, 4, 5 ...我仔细研究了一下，它似乎 `glyph_id_ofs_list` 生成不正确。数组中的 `0` s 看起来很糟糕。它应该是 0、1、2、3、4、5 ... image

Yes, it work with a font converted to C.

src/font/lv_font_fmt_txt.c
```diff
@@ -270,6 +270,7 @@ static uint32_t get_glyph_dsc_id(const lv_font_t * font, uint32_t letter)
         }
         else if(fdsc->cmaps[i].type == LV_FONT_FMT_TXT_CMAP_FORMAT0_FULL) {
             const uint8_t * gid_ofs_8 = fdsc->cmaps[i].glyph_id_ofs_list;
+            if(gid_ofs_8[rcp] == 0 && letter != fdsc->cmaps[i].range_start) continue;
```
kisvegabor, line 276:
```suggestion
            /* The first character is always valid and should have offset = 0
             * However if a character is missing it also has offset=0.
             * So if there is a 0 not on the first position then it's a missing character */
            if(gid_ofs_8[rcp] == 0 && letter != fdsc->cmaps[i].range_start) continue;
```

kisvegabor:
I've started to Debug it and finally understood the issue. The first glyph is mapped to index=0, but the missing characters are also mapped to 0. In light of that, I think your suggestion is correct because:
1. The first character is always valid and should have index=0
2. index=0 later always means a missing character

I've added this as a comment in the code.

kisvegabor (approved):

Allen-azh:
Hi,

Thank you for your feedback and for understanding the issue! I'm glad my suggestion helped resolve the problem. I also appreciate you adding the comment in the code; it makes things much clearer. I hope this change effectively resolves the issue. Feel free to reach out if there are more opportunities for improvements.

Thanks again for your effort and support!

Allen-azh:
@FASTSHIFT The issue has been resolved. Could you kindly review it at your earliest convenience?

FASTSHIFT (approved):

kisvegabor:
Thank you @Allen-azh% :blush:

PR #7407, opened 2024-12-04
perf(vg_lite): add font cache reference counting support

FASTSHIFT:
* Add built-in font cache to completely solve the performance problems caused by repeated conversion and decompression of built-in fonts, and the output image can ensure that it meets the GPU alignment requirements.

* Based on the reference counting capability of cache, it can perfectly adapt to GPU asynchronous rendering without waiting for the font image buffer to be released.

* Using config control, the old rendering mode can continue to be used on devices with limited memory.
 
### Notes


src/draw/vg_lite/lv_vg_lite_utils.c
```diff
@@ -1273,6 +1292,11 @@ void lv_vg_lite_flush(struct _lv_draw_vg_lite_unit_t * u)
         LV_PROFILER_DRAW_END;
         return;
     }
+
+    /* When the GPU is in idle state, it means that the GPU is not accessing
+     * any resources and the cache reference state can be released.
+     */
+    lv_vg_lite_clear_all_pending(u);
```
uLipe: Question: The fact of the GPU being idle would not mean it will not access the same cache entries again, so why not keep then and just invalidate upon an explicit "cache miss"?

src/font/lv_font_fmt_txt.c
```diff
@@ -22,6 +22,11 @@
     #define font_rle LV_GLOBAL_DEFAULT()->font_fmt_rle
 #endif /*LV_USE_FONT_COMPRESSED*/
 
+#if LV_FONT_FMT_TXT_CACHE_GLYPH_CNT > 0
+    #define font_bitmap_cache LV_GLOBAL_DEFAULT()->font_bitmap_cache
```
uLipe: Are those meant to be read_only? if so why not:

```suggestion
    #define font_bitmap_cache() LV_GLOBAL_DEFAULT()->font_bitmap_cache
```
To imply this is a macro ?

FASTSHIFT: This is not read-only, we need to create a cache manager object and store it on this pointer.

src/draw/vg_lite/lv_vg_lite_utils.c
FASTSHIFT: The cache entry needs to be locked while the GPU is reading the image resource, because the GPU only records the pointer of the image instead of copying the image, so it is necessary to prevent the cache manager from mistakenly kicking out the image resource that the GPU is referencing.
The meaning of pending is to record the locked cache entry in the list, and when the GPU completes the work, the cache is unlocked uniformly, and the cache is not cleared.

src/draw/vg_lite/lv_draw_vg_lite_label.c
```diff
@@ -239,14 +254,26 @@ static void draw_letter_bitmap(lv_draw_vg_lite_unit_t * u, const lv_draw_glyph_d
         lv_vg_lite_path_drop(u, path);
     }
 
-    /* TODO: The temporary buffer of the built-in font is reused.
-     * You need to wait for the GPU to finish using the buffer before releasing it.
-     * Later, use the font cache for management to improve efficiency.
-     */
+#if LV_FONT_FMT_TXT_CACHE_GLYPH_CNT > 0
```
kisvegabor: If we are rendering a FreeType or TinyTTF font it will be cached because of `LV_FONT_FMT_TXT_CACHE_GLYPH_CNT`. So it seems seems more like a generic bitmap font cache, not related only to the built-in format.

kisvegabor:
I have a questions: What if `LV_FONT_FMT_TXT_CACHE_GLYPH_CNT` is 8 and the GPU's command buffer size is 16. It means not all glyphs can be kept alive.

src/draw/vg_lite/lv_draw_vg_lite_label.c
FASTSHIFT: At present, both freetype and tinyttf support cache reference counting. Later, we can consider centralized management of font cache, and the font back-end only needs to focus on adapting the creation method of font bitmap.
At present, we can implement a built-in font independent cache, and then gradually modify and refactor:)

FASTSHIFT:
> I have a questions: What if `LV_FONT_FMT_TXT_CACHE_GLYPH_CNT` is 8 and the GPU's command buffer size is 16? It means not all glyphs can be kept alive.

Yes, we must ensure that the total number of font caches exceeds the maximum number of concurrent GPU holdings.
This is usually not a big problem because the font cache size is usually configured to be large to keep the frame rate stable for long text drawing. In addition, LVGL checks whether the GPU has completed drawing before each commit to the GPU and releases the cache reference in time.

src/font/lv_font_fmt_txt.c
uLipe: Nice, fair enough.

src/draw/vg_lite/lv_vg_lite_utils.c
uLipe: Makes sense now, thank you for clarification!

src/font/lv_font_fmt_txt.c
```diff
@@ -591,3 +543,244 @@ static int unicode_list_compare(const void * ref, const void * element)
...
+    font_bitmap_cache_data_t search_key = {
+        .fdsc = fdsc,
+        .glyph_index = glyph_index,
+    };
+
+    lv_cache_entry_t * entry = lv_cache_acquire_or_create(font_bitmap_cache, &search_key, NULL);
+
+    if(!entry) {
+        LV_LOG_TRACE("Glyph %d not found in cache", glyph_index);
+        return NULL;
```
liamHowatt:
```suggestion
        LV_PROFILER_FONT_END;
        return NULL;
```

liamHowatt (commented):
Looks good. I can't comment on all the "bigger picture" effects of this. Looks like kisvegabor is thinking about it, at least.

src/font/lv_font_fmt_txt.c
FASTSHIFT: Nice catch!

liamHowatt (dismissed):

src/draw/vg_lite/lv_draw_vg_lite_label.c
kisvegabor: I see, but it feels a little messy to me. What about this?
- Keep everything as it is in this PR, just change the name from `LV_FONT_FMT_TXT_CACHE_GLYPH_...` to `LV_FONT_CACHE_GLYPH_...` to make it sound generic. 

In a follow up PR we can remove the TinyTTF and FreeType bitmap caches.

FASTSHIFT: Done.

uLipe (dismissed):
LGTM!

src/draw/vg_lite/lv_draw_vg_lite_label.c
kisvegabor: Thanks!

src/font/lv_font_fmt_txt.c
```diff
@@ -591,3 +543,249 @@ static int unicode_list_compare(const void * ref, const void * element)
...
+#if LV_FONT_CACHE_GLYPH_CNT > 0
+
+/**
+ * Get the glyph's bitmap from the cache.
+ * @param g_dsc the glyph descriptor
+ * @param fdsc the font descriptor
+ * @param glyph_index the index of the glyph in the font
+ * @return pointer to the draw buffer
+ */
+static void * get_bitmap_cached(lv_font_glyph_dsc_t * g_dsc, const lv_font_fmt_txt_dsc_t * fdsc, uint32_t glyph_index)
```
kisvegabor: It also should be moved to a generic place. E.g. `lv_font.c`.

kisvegabor:
I'm thinking about how to keep things cleaner without API break. So now we might want 2 things:
1. Get the raw bitmap: no caching is needed if its format is suitable and we are reading it from flash. 
2. Get a cached and (maybe) formatted bitmap. We can add a new get bitmap_get_cached function for that. If it returns an error (because all glyphs are being used and nothing can be dropped) the GPU can know that it should flush the command list.

Having a not cached but formatted doesn't make much sense.

So ideally we should have
```c
//Return NULL if no raw bitmap if available
//Uses lv_font_get_glyph_bitmap internally
//Can be used in GPUs in case of aligned formats
const uint8_t * lv_font_get_glyph_bitmap_raw(lv_font_glyph_dsc_t * g_dsc);

//Return NULL if there is no more space in the cache now. 
//Basically your new `get_bitmap_cached` function
lv_draw_buf_t *  lv_font_get_glyph_bitmap_cached(lv_font_glyph_dsc_t * g_dsc, lv_color_format_t cf);
```
What do you think?

---

Potential issue: What if 2 draw units (e.g. VGLite and SW) are drawing letters at the same time? Let's say the cache size 4 glyphs. VGLite want's to draw exactly 4 bitmaps, so it quickly queries them ans starts adding more draw tasks. In the meantime SW can't proceed with letter drawing as there is no more space in the cache. In this how will SW wait for a new space in the cache? Just loop in a while and keep trying?


Performance: Please benchmark the performance impact of caching on a the screen sized text scene of the benchmark demo.

FASTSHIFT:
> Potential issue: What if 2 draw units (e.g. VGLite and SW) are drawing letters at the same time? Let's say the cache size is 4 glyphs. VGLite wants to draw exactly 4 bitmaps, so it quickly queries them and starts adding other draw tasks. In the meantime SW can't proceed with letter drawing as there is no more space in the cache. In this case how will SW wait for a new space in the cache? Just loop in a while and keep trying?

In this case, it just skips drawing and logging, and tells the user to increase the cache space. 
The font cache is usually configured to be large, usually up to hundreds of characters, and it is unlikely that allocation will fail. It is not recommended to introduce too complex a design for this low-probability situation, as it will involve many complex synchronization issues.

kisvegabor:
What about my API suggestions above?

FASTSHIFT:
> What about my API suggestions above?

I feel that it is possible to use the draw_buf interface to determine whether the alignment requirements are met without distinguishing whether the resource passes through the cache.
When the resource meets the alignment requirements, draw_buf directly executes the original address of the resource without the need for a second copy through the cache.
When the alignment requirements cannot be met, the font image is copied to the draw_buf that meets the alignment requirements through the cache.
In my PR, req_raw_bitmap can still work normally without passing through the cache. From the perspective of GPU backend developers, it is still more desirable that the font framework shields the details of resource processing as much as possible and uses a unified format for data exchange, which is easier to expand and maintain without writing too many judgment branches.

FASTSHIFT:
@kisvegabor @MarkIoanAmat 
OK, let's align the target behavior. Now there are several known situations:
* GPU rendering + aligned font image resources -> directly use the font image in flash, without copying and reference counting management, theoretically the highest access efficiency.
* GPU rendering + unaligned font image resources(non-A8 format/compressed format) -> use the font cache framework for conversion (this PR), to ensure that the GPU receives an image that meets the alignment requirements, and use the reference counting mechanism to lock the cache to prevent the cache from being released when the GPU reads the font resource. The GPU backend dereferences the cache after rendering.
* CPU software rendering + aligned/unaligned font image resources (A8 format) -> directly use the font image in flash
* CPU software rendering + aligned/unaligned font image resources (non-A8 format/compressed format) -> use the font cache framework for conversion to reduce the frequency of image conversion.

FASTSHIFT:
I noticed a problem. If the font file only marks that it is aligned, but does not indicate what the specific alignment parameters are (such as the first address, stride information), this design is prone to bugs from project experience.
Because it is possible that users may try to use a set of resources in different projects. If different GPUs are used, the alignment requirements are also different, which will eventually cause strange software failures. It will take a lot of effort to locate the problem, and finally find that the resource parameter settings are wrong.

The recommended draw_buf interface is used to pass font image data, so that the data obtained by the GPU can be checked. Don't worry about extra copies. If the GPU alignment requirements are met, draw_buf->data can directly point to the original font image address to achieve zero-copy drawing.

@MarkIoanAmat I would also like to ask a question. Is there a big difference between FLASH access speed and RAM on the platform you use? Because the access speed of RAM is usually much faster than FLASH. If frequently accessed font images can be cached in RAM, GPU drawing efficiency can be improved.

kisvegabor:
Thanks for the summary but I'm still not full convinced. We should keep the complexity as low as possible. In the complex case ("converted" bitmaps) now we have multi level cache. E.g. in case of FreeType:
- FreeType might cache the bitmaps
- We also cache the bitmaps in this PR in a global bitmap cache
- The GPU also saves a reference to the cache entry so that it can release them.

In the process we need to hope that there i enough cache for the command list.

I think something simpler would be better.

I'd summarize the cases like this:
- Synchronous rendring (there is no command list, rendering happens immediately, e.g. SW render)
   - no cache cache life cycle management is needed
   - aligned static bitmaps can be rendered as they are
   - unaligned static bitmaps needs to be converted to the correct format and can be rendered after that without issues
   - dynamically created bitmaps (e.g. FreeType) can be rendered as they are. FreeType/TinyTTF cache might or might not be used, it doesn't matter. 
- Asynchronous rendering (GPU with command list or a single blending with interrupt)
  - aligned static bitmaps can be rendered as they are, a list of draw_bufs can be saved in the draw unit to free them later
  - dynamically created bitmaps needs to have a local copy in the draw_bufs as they can be removed from FreeType cache at any time
  - draw_bufs with a pointer to a static bitmap or malloced bitmap can be stored and manged identically and freed when the GPU is ready
 
So what are the real life scenarios:
- SW is simple: 
   - if it's a static A8 bitmap, draw it
   - if it's a static A1/2/4 bitmap, convert it and draw it (we might add A1, A2, A4 blending support to SW instead of caching A8)
   - if it's e.g. a FreeType bitmap, it's A8 anyway, just draw it   
- In case of GPUs 
  - We should avoid converting static bitmaps. Instead the static fonts should be provided in an aligned format. We don't need to prepare for any theoretical GPUs. **Instead we should consider only VG-Lite, Nema, and Dave2D. Can you confirm if 1 byte alignment is enough for these for A1/2/4/8 formats?**
  - If it's a dynamic format, just copy the data to a GPU managed draw buf
  - So in real life, in case of GPUs the "Convert A2 to A8 and cache the result to save the conversation time" shouldn't happen. If we really want to cache the conversation result, than this cache should be local to the built in font engine (sorry for asking @FASTSHIFT to make it global. :cry: ). However, I'd really rather add A1/2/4 blending support instead. 
  
  
So in summary:
- I wouldn't add this cache
- I would add A1/2/4 blending support to SW
- I would make the GPUs save the draw_buf with malloc-ed data or just a pointer static bitmaps

What do you think?

FASTSHIFT:
> Instead we should consider only VG-Lite, Nema, and Dave2D. Can you confirm if 1 byte alignment is enough for these for A1/2/4/8 formats?

One-byte alignment does not apply to VG-Lite, which requires 64-byte alignment at the first memory address and 16px alignment at the width. In addition, this GPU can only support A4 and A8 images, A1 and A2 are not supported formats.

kisvegabor:
> Does that mean that every text drawn produces a copy and dynamic memory allocation? This affects the drawing performance of long text. And every GPU has to write the same logic again. Or if you are on a GPU platform and want to achieve better performance, you must regenerate new aligned font files and give up using LVGL's built-in fonts.

I wanted to say this: "It's bad, but I think we cannot avoid it. Imagine that FreeType has no cache or just a small cache so it can remove its cached data before it was rendered". And I realized that it's the whole point of having an LVGL controlled cache... 🤦

So, when we can and should use direct static data access?
1. SW render with A8 bitmaps
2. SW render with A1/2/4 if we add the required blending
3. GPUs is they support 1 byte alignment, 1 byte stride with A1/2/4/8. 

When do we need a global LVGL cache?
1. If FreeType or TinyTTF is used to replace their bitmap cache
2. If the built-in fonts needs to be converted (not supported color format, compressed, unaligned, wrong stride, etc)

From [this comment](https://github.com/lvgl/lvgl/pull/7407#issuecomment-2544902824) of @FASTSHIFT 

> The font backend only needs to check whether the given font image is static or converted. If it is static, the reference count processing is ignored. In this way, the GPU rendering backend does not care whether the font is static or converted, or whether it is a freetype/tinyttf font.

Sounds good, make sense. A few notes:
- `lv_font_get_glyph_bitmap_static()` is needed to decide if just a pointer can be stored
- [Here](https://github.com/lvgl/lvgl/pull/7407/files#diff-9d50f3563437f6e2a4ab70489e05c3c633d9a88267036929eddf37f968ef1b4cR706) the `draw_buf` is created by the cache, so reference counting is needed for static bitmaps too, otherwise the `draw_buf` itself can be removed. Maybe we can provide a the `draw_buf` externally as a parameter so the user manages the draw_buf's lifecycle and cache only sets the width, height, data, etc.  
- IMO we should make as much as we can in this PR to see the whole thing together. Including:
   - Add the global cache support (mostly done, needs to be moved to `lv_font` from `lv_font_fmt_txt`)
   - Update APIs (mostly done)
   - Disable FreeType and TyinyTTF bitmap cache
   - Figure out something for smaller cache size than command list size 
   - Benchmark the performance

W-Mai:
without builtin font cache may cause this mem leak I think.

https://github.com/lvgl/lvgl/actions/runs/12427226274/job/34696709168?pr=7490#step:7:1575

kisvegabor:
@FASTSHIFT I think your :+1: on my previous comment means you agree with it. Do you plan to update this PR in light of that?

FASTSHIFT:
@kisvegabor 

> Disable FreeType and TyinyTTF bitmap cache

Is this a big change? I plan to merge the global font cache in the next PR, implement the built-in font cache in this PR, and make small changes.

> Figure out something for smaller cache size than command list size

The solution is to print an error log to notify the user to increase the font cache (the lv_cache framework will automatically print it). If the cache is too small, cache misses will occur frequently, and the performance will be worse than turning off the cache.

> Benchmark the performance

I made some time to do it this week.

FASTSHIFT:
> Benchmark the performance

It's here:

Test code:

```c
static void label_redraw_timer_cb(lv_timer_t* timer)
{
    lv_obj_t* label = (lv_obj_t*)lv_timer_get_user_data(timer);
    lv_obj_invalidate(label);
}

static void label_stress(void)
{
    lv_obj_t* label = lv_label_create(lv_scr_act());
    lv_obj_set_width(label, lv_pct(100));
    lv_label_set_long_mode(label, LV_LABEL_LONG_WRAP);
    lv_obj_set_style_text_font(label, &lv_font_montserrat_48, 0);
    lv_label_set_text(label, "ABCDEFGHIJKLMNOPQRSTUVWXYZ"
                             "abcdefghijklmnopqrstuvwxyz"
                             "0123456789");
    lv_timer_create(label_redraw_timer_cb, 10, label);
}
```

Software render, 12.6ms:

image

[trace_sw_letter_48.txt](https://github.com/user-attachments/files/18228049/trace_sw_letter_48.txt)

VG-Lite render + Disable font cache:

image

[trace_gpu_letter_48_no_cache.txt](https://github.com/user-attachments/files/18228031/trace_gpu_letter_48_no_cache.txt)

VG-Lite render + Enable font cache:

image

[trace_gpu_letter_48_cache.txt](https://github.com/user-attachments/files/18228045/trace_gpu_letter_48_cache.txt)

kisvegabor:
> Is this a big change? I plan to merge the global font cache in the next PR, implement the built-in font cache in this PR, this will impact the current architecture.

 Are there any issues adding the global cache in a single PR? It would allow us to test, evaluate, and benchmark the whole thing and merge it with great confidence. 

> Increase the cache capacity + limit the number of characters written to the GPU command buffer (on VG-Lite I set it to flush every 8 letters).

Having a constant max letter count definitely would make it more predictable. However it can happen that certain label draw tasks are dispatched to a multi core SW renderer and other to VG-Lite. The problematic scenario is 
- The cache size is 8
- VG-Lite already queued 8 letters 
- SW is rendering in the meantime, but the is no place for in the cache. 

We can say that the cache size can be large enough, but when it's not large enough, it will lead to non deterministic rendering issues. The best would be to say that SW render needs to wait until there is place in the cache. Maybe we can add a mutex or a thread sync for cache full? We can still add warning when there is no place in the cache but we should handle it explicitly. 



>> Benchmark the performance

> It's here:


Fantastic! Did the cache make and difference in case of SW render too?

uLipe:
Just adding an extra point regarding the predictable drawing, caches are source of indeterminism, regardless the system you don't have an 100% bounded latency because you don't have a model of how much time a cache entry will become invalid ( cache miss) or when it becomes valid ( cache hit ), and a potential invalid leads to cache flush which add a refill uncertainty. 

So this VGLite case here the same issue will be present when doing multicore rendering and mixing sw + hardware draw engines, the cache présence here will be aimed to increase throughput, but it will not enforce any bounded latency ( of course the jitter here might be negligible on the majority of real time systems).

FASTSHIFT:
> Fantastic! Did the cache make and difference in case of SW render too?

Software rendering performance is improved by 30%, and rendering time is reduced from 12.6ms to 8.8ms.

image

[trace_sw_letter_48_cache.txt](https://github.com/user-attachments/files/18238331/trace_sw_letter_48_cache.txt)

FASTSHIFT:
> Are there any issues adding the global cache in a single PR? It would allow us to test, evaluate, and benchmark the whole thing and merge it with great confidence.

How to coexist with freetype's outline mode needs further investigation, because it stores vector data instead of draw_buf. I want to make the first minimal change (the current modification of this PR) and take the first step in performance optimization :)

W-Mai:
I think the case where the cache is too small is a very extreme boundary situation. Secondly, a safer way is to draw placeholder and print warning.

If the cache is full, the system heap allocates font information, because some embedded system main heaps cannot allocate memory will panic (NuttX), so this scheme is not appropriate.

Our products uses cache count in [256, 1024).

kisvegabor:
> Software rendering performance is improved by 30%, and rendering time is reduced from 12.6ms to 8.8ms.

Awesome! 

>  I want to make the first minimal change (the current modification of this PR) and take the first step in performance optimization :)

Okay, let's target the minimal changes to merge this PR. :slightly_smiling_face: Please
- Move the cache code to an `lv_font_cache.c/h` file (or similar) to indicate that it's a global feature and to make it easier to change later
- If the alignment, stride, and color format is correct, don't allocate the buffer in the `draw_buf`
- Add these to using e.g. A4 bitmaps directly:
   -  in `lv_font_t` `uint8_t static_bitmap : 1;`
   - `lv_font_has_static_bitmap(font)`
   - `void lv_font_get_raw_gitmap(font, &draw_buf);`

Are you okay with this?

FASTSHIFT:
@kisvegabor 
> Move the cache code to an lv_font_cache.c/h file (or similar) to indicate that it's a global feature and to make it easier to change later

Done.

> If the alignment, stride, and color format is correct, don't allocate the buffer in the draw_buf
Add these to using e.g. A4 bitmaps directly:
in lv_font_t uint8_t static_bitmap : 1;
lv_font_has_static_bitmap(font)
void lv_font_get_raw_gitmap(font, &draw_buf);

I don't quite understand how to implement this part? Can you give more details?

kisvegabor:
> If the alignment, stride, and color format is correct, don't allocate the buffer in the draw_buf

I suggest using `lv_draw_buf_init(draw_buf, w, h, cf, static_bitmap_pointer, size)` of `lv_draw_buf_create`. 

> Add these to using e.g. A4 bitmaps directly:
> in lv_font_t uint8_t static_bitmap : 1;

I believe it's clear.

> lv_font_has_static_bitmap(font);

Just getter for the `static_bitmap` flag.

> void lv_font_get_raw_gitmap(font, &draw_buf);

A function to skip the cache and get the bitmap as it is. Can return `LV_RES_INVALID` is called on a non static bitmap font.

src/font/lv_font.c
```diff
@@ -46,16 +47,46 @@ const void * lv_font_get_glyph_bitmap(lv_font_glyph_dsc_t * g_dsc, lv_draw_buf_t
...
     LV_ASSERT_NULL(font_p);
+#if LV_FONT_CACHE_GLYPH_CNT > 0
+    /* If the font cache is enabled, try to use it to get the glyph bitmap */
+    return lv_font_cache_get_glyph_bitmap(g_dsc, draw_buf);
+#else
     return font_p->get_glyph_bitmap(g_dsc, draw_buf);
+#endif
+}
+
+lv_result_t lv_font_get_glyph_raw_bitmap(lv_font_glyph_dsc_t * g_dsc, lv_draw_buf_t * draw_buf)
```
kisvegabor: Sorry just noticed the inconsitency (although you implemented it as I suggested):

```suggestion
lv_result_t lv_font_get_glyph_static_bitmap(lv_font_glyph_dsc_t * g_dsc, lv_draw_buf_t * draw_buf)
```
Let's use the `static_bitmap` expression everywhere.

src/font/lv_font_cache.c
```diff
@@ -0,0 +1,212 @@
...
+
+    lv_cache_release(font_bitmap_cache, g_dsc->entry, NULL);
+    g_dsc->entry = NULL;
+}
+
+/**********************
+ *   STATIC FUNCTIONS
+ **********************/
+
+static inline bool font_is_built_in(const lv_font_t * font)
```
kisvegabor: I suggest not treating the built-in fonts in a special way. It seems here the gist is if the font is static or not.

kisvegabor (commented):
Thank you! I'll test and review it in more detail. Just added a few early comments. 

It seems this is still missing, right?
> I suggest using lv_draw_buf_init(draw_buf, w, h, cf, static_bitmap_pointer, size) of lv_draw_buf_create.

FASTSHIFT:
Not stale.

kisvegabor:
@FASTSHIFT approximately when will you have time for this?

FASTSHIFT:
> @FASTSHIFT approximately when will you have time for this?

Let's continue~

src/font/lv_font_cache.c
FASTSHIFT: At this stage, the built-in static font has not added this static_bitmap tag. I first use this function to judge. Although it is a bit strange, it can correctly distinguish and is forward compatible.

src/font/lv_font.c
FASTSHIFT: Done.

FASTSHIFT:
@kisvegabor 
> I suggest using lv_draw_buf_init(draw_buf, w, h, cf, static_bitmap_pointer, size) of lv_draw_buf_create.

Done. See [here](https://github.com/lvgl/lvgl/pull/7407/files#diff-fe5a07e82bdd5f437fac7a7c1db046822161078311c44bd29c7bd51bc11df37fR107).

FASTSHIFT:
It should be noted that the current font bitmap design does not take alignment into consideration, and most fonts cannot meet the first address alignment requirement.
image

FASTSHIFT:
If `LV_DRAW_BUF_ALIGN` is set to 1, zero-copy rendering of the font can be achieved:

`LV_DRAW_BUF_ALIGN=4`:

image

`LV_DRAW_BUF_ALIGN=1`:

image

src/font/lv_font_cache.c
```diff
@@ -0,0 +1,230 @@
...
+    /* If the alignment, stride, and color format is correct, bypass the cache */
+    if(g_dsc->format == LV_FONT_GLYPH_FORMAT_A8
+       && raw_glyph_bitmap == lv_draw_buf_align(raw_glyph_bitmap, LV_COLOR_FORMAT_A8)
+       && g_dsc->box_w == lv_draw_buf_width_to_stride(g_dsc->box_w, LV_COLOR_FORMAT_A8)) {
+
+        const uint32_t a8_stride = gdsc->box_w * sizeof(uint8_t);
+
+        /* Use the static bitmap draw buffer to avoid memory allocation */
+        lv_draw_buf_init(
+            &font_static_bitmap_draw_buf,
```
kisvegabor: It can happen that 2 draw units are using the `font_static_bitmap_draw_buf` at the same time. 

IMO the user can always supply a `draw_buf *` to this function and we can initialize it here, or use it in `get_bitmap_cached` below. This way the life cycle of `draw_buf` managed on the user's side which is more flexible and transparent.

src/font/lv_font_cache.c
kisvegabor: I see. Then let's name this function `has_static_bitmap`.

FASTSHIFT: Done

src/font/lv_font_cache.c
FASTSHIFT: Done, I added a check on `draw_buf` to bypass the font cache when the `draw_buf` pointer is passed in externally.
`font_static_bitmap_draw_buf` is only used in `LV_OS_NONE` mode, there will be no multithreading issues.

kisvegabor:
> Done, I added a check on draw_buf to bypass the font cache when the draw_buf pointer is passed in externally.
font_static_bitmap_draw_buf is only used in LV_OS_NONE mode, there will be no multithreading issues.

It's not necessarily true. On my notebook I can run 4 SW render units with pthread, all using a static bitmaps.

src/font/lv_font_cache.c
```diff
@@ -0,0 +1,240 @@
...
+    /* If the alignment, stride, and color format is correct, bypass the cache */
+    if((g_dsc->format == LV_FONT_GLYPH_FORMAT_A8 || g_dsc->format == LV_FONT_GLYPH_FORMAT_A8_ALIGNED)
+       && raw_glyph_bitmap == lv_draw_buf_align((void *)raw_glyph_bitmap, LV_COLOR_FORMAT_A8)
+       && g_dsc->box_w == lv_draw_buf_width_to_stride(g_dsc->box_w, LV_COLOR_FORMAT_A8)) {
+
+        const uint32_t a8_stride = gdsc->box_w * sizeof(uint8_t);
+
+        /* Use the static bitmap draw buffer to avoid memory allocation */
+        lv_draw_buf_init(
+            &font_static_bitmap_draw_buf,
```
FASTSHIFT: @kisvegabor Is there any multi-threaded competition risk for this `font_static_bitmap_draw_buf` global variable?

FASTSHIFT: I will remove it temporarily and add it again when necessary.

FASTSHIFT:
> I'm still puzzled with a few things. So we cache `draw_buf`s where the data is either malloced or points to a static bitmap.
> 
> You mentioned that the cache size should be larger than the command list size of GPU, but imagine this:
> 
> 1. The font cache size is 32
> 2. VGLite works on a large and complex gradient (2-3 commands, and 30 characters after that)
> 3. In the meantime SW wants to draw 3 characters.
> 
> In this case there won't be place for the last character in SW.
> 
> Note that it can happen only if
> 
> * there is an OS.
> * For some reason the letters are drawn by 2 draw units at the same time (usually one draw unit can do it, but it's not necessarily the case)
> 
> So if the the requested character cannot be cached we can either:
> 
> * Use a mutex somehow to wait until there will be space in the cache (VG-Lite finishes)
> * Go uncached as a fallback: but in this case the draw_buf returned by `lv_font_get_glyph_bitmap` either cached or not cached. So the user needs to free the data or not, free the draw_buf or not, so it's complicated.
> 
> That's why I suggested to **always** use the `draw_buf` provided in `lv_font_get_glyph_bitmap`, so that the user can handle it (save in a pending list, and free it when needed, etc.) in my view the cache should save only the A8 bitmaps and assign them to the provided draw_buf data's.
> 
> When VG-Lite finishes the pending list can be cleared like this:
> 
> 1. if `draw_buf->header.flags & LV_IMAGE_FLAGS_ALLOCATED` free the data: maybe it was uncached, or anything
> 2. if `draw_buf->header.flags & LV_IMAGE_FLAGS_CACHED` (new flag): tell the cache to release this data.
> 
> @W-Mai
> 
> What's happening now with images if the cache size 10kB, it's full with one image that is being rendered by VGLite and an other image is requested by an other draw unit in the meantime?

I understand your concern. In this PR, you can use the `draw_buf` provided by `lv_font_get_glyph_bitmap` to be compatible with it. Currently vg-lite uses `draw_glyph_dsc->g->entry` to determine whether it is in cache, rather than draw_buf flags, which currently works well and can work with cache font and none cache font.
Is there anything else that needs to be modified?

FASTSHIFT:
Tips: `lv_cache_release` is used to reduce the reference count by one, but does not release memory. When the cache is full and a new cache object needs to be inserted, the cache manager will first try to remove the cache object with a reference count of 0.
`lv_cache_drop` is used to free cache object memory.

kisvegabor:
Sorry for nit picking that much, but it's a very important feature and I'd like to be sure we do it as good as possible.

After seeing your implementation, I think we should have this API:
```c
/*Same as it is in master. Don't use cache*/
const void * lv_font_get_glyph_bitmap(lv_font_glyph_dsc_t * g_dsc, lv_draw_buf_t * draw_buf);

/*If possible load the cached bitmap in the draw_buf.
 *If there is not place in the cache go uncached automatically (call lv_font_get_glyph_bitmap)
 *The draw_buf is stored and managed by the user, we take care of the bitmap only*/
void lv_font_get_glyph_bitmap_cached(lv_font_glyph_dsc_t * g_dsc, lv_draw_buf_t * draw_buf);

/*Release the cache if it was cached, or free the memory if it wan't cached. Use a flag or anything else to figure out was was cached.*/
void lv_font_release_bitmap_cached(lv_draw_buf_t * draw_buf);
```

IMO it would be cleanest and most explicit way to handle it. What do you think?

W-Mai:
As an aside, I think we could have a logic like this if we want to display even after the cache is full:

1. `!cache.full` && `cache.evictable`: insert into cache
2. `cache.full` && `cache.evictable`: evict entries and insert into cache
3. `cache.full` && `!cache.evictable`: skip cache and just decode directly like no cache cases.

However, there may be times when you want to take a fail to decode solution when the cache is full and can't be evicted, to make sure that the memory usage is under control. So we can use a configuration to control this behavior.

FASTSHIFT:
> Sorry for nit picking that much, but it's a very important feature and I'd like to be sure we do it as good as possible.
> 
> After seeing your implementation, I think we should have this API:
> 
> ```c
> /*Same as it is in master. Don't use cache*/
> const void * lv_font_get_glyph_bitmap(lv_font_glyph_dsc_t * g_dsc, lv_draw_buf_t * draw_buf);
> 
> /*If possible load the cached bitmap in the draw_buf.
>  *If there is not place in the cache go uncached automatically (call lv_font_get_glyph_bitmap)
>  *The draw_buf is stored and managed by the user, we take care of the bitmap only*/
> void lv_font_get_glyph_bitmap_cached(lv_font_glyph_dsc_t * g_dsc, lv_draw_buf_t * draw_buf);
> 
> /*Release the cache if it was cached, or free the memory if it wan't cached. Use a flag or anything else to figure out was was cached.*/
> void lv_font_release_bitmap_cached(lv_draw_buf_t * draw_buf);
> ```
> 
> IMO it would be cleanest and most explicit way to handle it. What do you think?

I have a question. If we let users manage draw_buf, how should we guide users to manage draw_buf correctly and efficiently without causing errors (for example: memory leaks, use after free)?

kisvegabor:
@W-Mai 
> However, there may be times when you want to take a fail to decode solution when the cache is full and can't be evicted, to make sure that the memory usage is under control. So we can use a configuration to control this behavior.

I agree, also with having a config for it. As a summary, what kid of caches do we have now?
- image decoder
- image header cache
- this font cache
- ?

---

@FASTSHIFT 
I have a question. If we let users manage draw_buf, how should we guide users to manage draw_buf correctly and efficiently without causing errors (for example: memory leaks, use after free)?

I think there are real life scenarios here:
1. Get only one bitmap in "sync" rendering, e.g. SW: Simple, `draw_buf` is a local variable
2. Get multiple draw_buf's in "async" rendering, e.g. VG-Lite: Complex anyways. 

I think what I proposed is not the different from what we  have in this PR now. The user needs to get the bitmap, use it, and release it. In my suggestion the main difference is that user also needs to allocate (or use a memory pool) from `draw_buf`s.

W-Mai:
> I agree, also with having a config for it. As a summary, what kid of caches do we have now?
> 
> * image decoder
> * image header cache
> * this font cache
> * ?

- Image Decoder Cache
- Image Decoder Header Cache
- Font Cache
  - Freetype Face Cache
  - Freetype Glyph Cache
  - Freetype Bitmap Cache
  - Freetype Outline Cache
  - TinyTTF Glyph Cache
  - TinyTTF Bitmap Cache
  - `This Cache`
- SDL Texture Cache
- VG_Lite Cache
  - Gradient Cache
  - Outline Cache

kisvegabor:
@W-Mai 
Wow, okay. Let's treat it separately and open an issue once if we can make some unification/improvements.

kisvegabor:
@FASTSHIFT Do you see when you can allocate time for this PR?

cosmindanielradu19:
Hi, I've tried this branch with our VGLite implementation, and I've observed a performance drop on screen sized text benchmark test when the cache is enabled (tested on RT1170 without SW rendering, running from FLASH).

no cache -Screen sized text, 90%, 8, 120, 110, 10
10 glyphs cache - Screen sized text, 96%, 3, 290, 281, 9
25 glyphs cache - Screen sized text, 92%, 6, 152, 142, 10
50 glyphs cache - Screen sized text, 93%, 6, 138, 131, 7
150 glyphs cache - Screen sized text, 92%, 7, 121, 114, 7

Is LV_FONT_CACHE_GLYPH_CNT the only config needed by this patch in order to benefit for all the improvement that was specified above? Is it anything else needed?

FASTSHIFT:
> Hi, I've tried this branch with our VGLite implementation, and I've observed a performance drop on screen sized text benchmark test when the cache is enabled (tested on RT1170 without SW rendering, running from FLASH).
> 
> no cache -Screen sized text, 90%, 8, 120, 110, 10 10 glyphs cache - Screen sized text, 96%, 3, 290, 281, 9 25 glyphs cache - Screen sized text, 92%, 6, 152, 142, 10 50 glyphs cache - Screen sized text, 93%, 6, 138, 131, 7 150 glyphs cache - Screen sized text, 92%, 7, 121, 114, 7
> 
> Is LV_FONT_CACHE_GLYPH_CNT the only config needed by this patch in order to benefit for all the improvement that was specified above? Is it anything else needed?

Emmm, after I tested and enabled the font cache(`LV_FONT_CACHE_GLYPH_CNT=256`), the average rendering time of screen-sized text decreased significantly: 42ms -> 37ms

image

From the perspective of the single letter drawing process, it significantly reduces the bitmap copy time and GPU synchronization waiting time:

image

FASTSHIFT:
> @FASTSHIFT Do you see when you can allocate time for this PR?

OK. I found that we may also need to pass in the alignment parameter requirements when requesting the bitmap from the font module, because different rendering backends have different alignment requirements. For example, software rendering has no alignment requirements and can directly use the original bitmap. However, VG-Lite's alignment requirements are very strict and must be passed through the font cache to reduce conversion.
Do you have any good suggestions on how to pass this parameter?

nicusorcitu:
Why not making glyph cache buffers in A4 format and all aligned to 8 bytes so that every render backend stride can be satisfied and use original bitmaps (zero copy).
SW A4 format can be added but the GPU stride is limited by hardware.

nicusorcitu:
The main reason I mentioned A4 format is to save some bitmap space but at least keeping the stride requirement.
A4 requires 8 bytes stride alignment while the current A8 require 16 bytes stride alignment.

I believe the purpose should be that any draw unit to utilize the original bitmaps (cached bitmaps) as they are, without any copy. Any if they are created with stride_align = 1 then it will be no benefit for GPUs.

kisvegabor:
>  Do you have any good suggestions on how to pass this parameter?

As we don't store draw unit specific alignment requirements anywhere. We can check if `lv_draw_buf_align` returns the same pointer. If so, it's aligned , if not it need to be aligned. 

> The main reason I mentioned A4 format is to save some bitmap space but at least keeping the stride requirement.
A4 requires 8 bytes stride alignment while the current A8 require 16 bytes stride alignment.

We can either 
1. develop a solution for returning cached bitmaps with a requested format (that potentially caching the same bitmap as A4 and A8 too) and save some RAM 
2. or accept the extra RAM usage (at  least for now), merge the basics of this feature and consider updating it in a new iteration. 

I vote for 2) otherwise I feel like we need to wait lot before doing the first steps. 

Of course we need to make sure that @nicusorcitu can also measure a performance boost.

Will also test it on our end:
- @liamHowatt can you run some test with SW render on the U5 ST board?
- @uLipe can you run some test with SW render on an Renesas RA8D1-EK

---

I still believe this more explicit API would be cleaner:  https://github.com/lvgl/lvgl/pull/7407#issuecomment-2662296912

nicusorcitu:
@cosmindanielradu19 please run one more time and check any boost performance with this PR.

cosmindanielradu19:
Hi, I've managed to do some benchmarks using RT1170, tested with VGLite only, compiled with -Os, ran from FLASH.

image

It seems that the performance is lower with this patch on our side. There are also performance differences between the current master branch and the font_cache branch even when the FONT_CACHE_GLYPH_CNT is set to 0.

FASTSHIFT:
> It seems that the performance is lower with this patch on our side. There are also performance differences between the current master branch and the font_cache branch even when the FONT_CACHE_GLYPH_CNT is set to 0.

@cosmindanielradu19 

Can you use this LOG to print whether there are frequent cache misses?

image

This is my log. When all bitmaps are loaded, the log is no longer printed:
```bash
[User]  (0.572, +572)    font_bitmap_create_cb: Created font bitmap cache entry for glyph_index 45, size 14x18 lv_font_cache.c:208
[User]  (0.572, +0)      font_bitmap_create_cb: Created font bitmap cache entry for glyph_index 80, size 15x14 lv_font_cache.c:208
[User]  (0.573, +1)      font_bitmap_create_cb: Created font bitmap cache entry for glyph_index 83, size 8x14 lv_font_cache.c:208
[User]  (0.573, +0)      font_bitmap_create_cb: Created font bitmap cache entry for glyph_index 70, size 14x14 lv_font_cache.c:208
[User]  (0.573, +0)      font_bitmap_create_cb: Created font bitmap cache entry for glyph_index 78, size 24x14 lv_font_cache.c:208
[User]  (0.573, +0)      font_bitmap_create_cb: Created font bitmap cache entry for glyph_index 74, size 5x19 lv_font_cache.c:208
[User]  (0.573, +0)      font_bitmap_create_cb: Created font bitmap cache entry for glyph_index 81, size 15x19 lv_font_cache.c:208
[User]  (0.573, +0)      font_bitmap_create_cb: Created font bitmap cache entry for glyph_index 84, size 13x14 lv_font_cache.c:208
[User]  (0.573, +0)      font_bitmap_create_cb: Created font bitmap cache entry for glyph_index 86, size 14x14 lv_font_cache.c:208
[User]  (0.573, +0)      font_bitmap_create_cb: Created font bitmap cache entry for glyph_index 69, size 15x19 lv_font_cache.c:208
[User]  (0.573, +0)      font_bitmap_create_cb: Created font bitmap cache entry for glyph_index 77, size 3x19 lv_font_cache.c:208
[User]  (0.573, +0)      font_bitmap_create_cb: Created font bitmap cache entry for glyph_index 85, size 11x17 lv_font_cache.c:208
[User]  (0.573, +0)      font_bitmap_create_cb: Created font bitmap cache entry for glyph_index 66, size 13x14 lv_font_cache.c:208
[User]  (0.574, +1)      font_bitmap_create_cb: Created font bitmap cache entry for glyph_index 13, size 4x8 lv_font_cache.c:208
[User]  (0.574, +0)      font_bitmap_create_cb: Created font bitmap cache entry for glyph_index 68, size 13x14 lv_font_cache.c:208
[User]  (0.574, +0)      font_bitmap_create_cb: Created font bitmap cache entry for glyph_index 79, size 14x14 lv_font_cache.c:208
[User]  (0.574, +0)      font_bitmap_create_cb: Created font bitmap cache entry for glyph_index 72, size 15x19 lv_font_cache.c:208
[User]  (0.575, +1)      font_bitmap_create_cb: Created font bitmap cache entry for glyph_index 15, size 4x4 lv_font_cache.c:208
[User]  (0.575, +0)      font_bitmap_create_cb: Created font bitmap cache entry for glyph_index 49, size 16x18 lv_font_cache.c:208
[User]  (0.575, +0)      font_bitmap_create_cb: Created font bitmap cache entry for glyph_index 82, size 15x19 lv_font_cache.c:208
[User]  (0.575, +0)      font_bitmap_create_cb: Created font bitmap cache entry for glyph_index 71, size 11x19 lv_font_cache.c:208
[User]  (0.576, +1)      font_bitmap_create_cb: Created font bitmap cache entry for glyph_index 67, size 15x19 lv_font_cache.c:208
[User]  (0.576, +0)      font_bitmap_create_cb: Created font bitmap cache entry for glyph_index 75, size 9x24 lv_font_cache.c:208
[User]  (0.577, +1)      font_bitmap_create_cb: Created font bitmap cache entry for glyph_index 87, size 16x14 lv_font_cache.c:208
[User]  (0.578, +1)      font_bitmap_create_cb: Created font bitmap cache entry for glyph_index 39, size 14x18 lv_font_cache.c:208
[User]  (0.580, +2)      font_bitmap_create_cb: Created font bitmap cache entry for glyph_index 52, size 15x18 lv_font_cache.c:208
[User]  (0.583, +3)      font_bitmap_create_cb: Created font bitmap cache entry for glyph_index 42, size 4x18 lv_font_cache.c:208
[User]  (0.584, +1)      font_bitmap_create_cb: Created font bitmap cache entry for glyph_index 38, size 14x18 lv_font_cache.c:208
[User]  (0.585, +1)      font_bitmap_create_cb: Created font bitmap cache entry for glyph_index 37, size 19x18 lv_font_cache.c:208
[User]  (0.587, +2)      font_bitmap_create_cb: Created font bitmap cache entry for glyph_index 47, size 17x18 lv_font_cache.c:208
[User]  (0.591, +4)      font_bitmap_create_cb: Created font bitmap cache entry for glyph_index 36, size 17x18 lv_font_cache.c:208
[User]  (0.593, +2)      font_bitmap_create_cb: Created font bitmap cache entry for glyph_index 55, size 20x18 lv_font_cache.c:208
[User]  (0.594, +1)      font_bitmap_create_cb: Created font bitmap cache entry for glyph_index 73, size 14x19 lv_font_cache.c:208
[User]  (0.595, +1)      font_bitmap_create_cb: Created font bitmap cache entry for glyph_index 46, size 21x18 lv_font_cache.c:208
[User]  (0.602, +7)      font_bitmap_create_cb: Created font bitmap cache entry for glyph_index 52, size 9x10 lv_font_cache.c:208
[User]  (0.602, +0)      font_bitmap_create_cb: Created font bitmap cache entry for glyph_index 68, size 8x8 lv_font_cache.c:208
[User]  (0.602, +0)      font_bitmap_create_cb: Created font bitmap cache entry for glyph_index 83, size 5x8 lv_font_cache.c:208
[User]  (0.602, +0)      font_bitmap_create_cb: Created font bitmap cache entry for glyph_index 70, size 8x8 lv_font_cache.c:208
[User]  (0.602, +0)      font_bitmap_create_cb: Created font bitmap cache entry for glyph_index 79, size 8x8 lv_font_cache.c:208
[User]  (0.602, +0)      font_bitmap_create_cb: Created font bitmap cache entry for glyph_index 84, size 7x8 lv_font_cache.c:208
[User]  (0.602, +0)      font_bitmap_create_cb: Created font bitmap cache entry for glyph_index 74, size 2x11 lv_font_cache.c:208
[User]  (0.602, +0)      font_bitmap_create_cb: Created font bitmap cache entry for glyph_index 91, size 7x8 lv_font_cache.c:208
[User]  (0.602, +0)      font_bitmap_create_cb: Created font bitmap cache entry for glyph_index 69, size 9x11 lv_font_cache.c:208
[User]  (0.602, +0)      font_bitmap_create_cb: Created font bitmap cache entry for glyph_index 85, size 6x10 lv_font_cache.c:208
[User]  (0.602, +0)      font_bitmap_create_cb: Created font bitmap cache entry for glyph_index 89, size 8x8 lv_font_cache.c:208
[User]  (0.602, +0)      font_bitmap_create_cb: Created font bitmap cache entry for glyph_index 27, size 3x8 lv_font_cache.c:208
[User]  (0.602, +0)      font_bitmap_create_cb: Created font bitmap cache entry for glyph_index 17, size 9x10 lv_font_cache.c:208
[User]  (0.602, +0)      font_bitmap_create_cb: Created font bitmap cache entry for glyph_index 39, size 8x10 lv_font_cache.c:208
[User]  (0.602, +0)      font_bitmap_create_cb: Created font bitmap cache entry for glyph_index 49, size 9x10 lv_font_cache.c:208
[User]  (0.602, +0)      font_bitmap_create_cb: Created font bitmap cache entry for glyph_index 13, size 3x5 lv_font_cache.c:208
[User]  (0.602, +0)      font_bitmap_create_cb: Created font bitmap cache entry for glyph_index 6, size 12x10 lv_font_cache.c:208
[User]  (0.603, +1)      font_bitmap_create_cb: Created font bitmap cache entry for glyph_index 36, size 10x10 lv_font_cache.c:208
[User]  (0.603, +0)      font_bitmap_create_cb: Created font bitmap cache entry for glyph_index 54, size 9x10 lv_font_cache.c:208
[User]  (0.603, +0)      font_bitmap_create_cb: Created font bitmap cache entry for glyph_index 71, size 6x11 lv_font_cache.c:208
[User]  (0.603, +0)      font_bitmap_create_cb: Created font bitmap cache entry for glyph_index 15, size 3x3 lv_font_cache.c:208
[User]  (0.603, +0)      font_bitmap_create_cb: Created font bitmap cache entry for glyph_index 78, size 13x8 lv_font_cache.c:208
[User]  (0.603, +0)      font_bitmap_create_cb: Created font bitmap cache entry for glyph_index 30, size 8x6 lv_font_cache.c:208
[User]  (0.603, +0)      font_bitmap_create_cb: Created font bitmap cache entry for glyph_index 12, size 8x7 lv_font_cache.c:208
[User]  (0.603, +0)      font_bitmap_create_cb: Created font bitmap cache entry for glyph_index 77, size 2x11 lv_font_cache.c:208
[User]  (0.603, +0)      font_bitmap_create_cb: Created font bitmap cache entry for glyph_index 86, size 8x8 lv_font_cache.c:208
[User]  (0.603, +0)      font_bitmap_create_cb: Created font bitmap cache entry for glyph_index 73, size 8x11 lv_font_cache.c:208
[User]  (0.604, +1)      font_bitmap_create_cb: Created font bitmap cache entry for glyph_index 9, size 4x14 lv_font_cache.c:208
[User]  (0.604, +0)      font_bitmap_create_cb: Created font bitmap cache entry for glyph_index 93, size 2x14 lv_font_cache.c:208
[User]  (0.604, +0)      font_bitmap_create_cb: Created font bitmap cache entry for glyph_index 10, size 4x14 lv_font_cache.c:208
[User]  (0.845, +241)    font_bitmap_create_cb: Created font bitmap cache entry for glyph_index 24, size 8x10 lv_font_cache.c:208
[User]  (0.845, +0)      font_bitmap_create_cb: Created font bitmap cache entry for glyph_index 23, size 9x10 lv_font_cache.c:208
[User]  (0.845, +0)      font_bitmap_create_cb: Created font bitmap cache entry for glyph_index 20, size 8x10 lv_font_cache.c:208
[User]  (0.845, +0)      font_bitmap_create_cb: Created font bitmap cache entry for glyph_index 26, size 8x10 lv_font_cache.c:208
[User]  (1.175, +330)    font_bitmap_create_cb: Created font bitmap cache entry for glyph_index 19, size 8x10 lv_font_cache.c:208
[User]  (1.175, +0)      font_bitmap_create_cb: Created font bitmap cache entry for glyph_index 21, size 10x10 lv_font_cache.c:208
[User]  (1.818, +643)    font_bitmap_create_cb: Created font bitmap cache entry for glyph_index 22, size 8x10 lv_font_cache.c:208
[User]  (1.818, +0)      font_bitmap_create_cb: Created font bitmap cache entry for glyph_index 18, size 4x10 lv_font_cache.c:208
[User]  (1.818, +0)      font_bitmap_create_cb: Created font bitmap cache entry for glyph_index 25, size 9x10 lv_font_cache.c:208
[User]  (2.045, +227)    font_bitmap_create_cb: Created font bitmap cache entry for glyph_index 34, size 21x18 lv_font_cache.c:208
[User]  (2.045, +0)      font_bitmap_create_cb: Created font bitmap cache entry for glyph_index 89, size 14x14 lv_font_cache.c:208
Benchmark Summary (9.3.0 dev)
Name, Avg. CPU, Avg. FPS, Avg. time, render time, flush time
Screen sized text, 93%, 24, 37, 31, 6
All scenes avg.,93%, 24, 37, 31, 6
```

You can modify the benchmark program to only test the "Screen sized text" test case:

image

Can you provide your `lv_conf.h` file and screen color depth and resolution information? :>

cosmindanielradu19:
> Can you provide your `lv_conf.h` file and screen color depth and resolution information? :>

Sure, here it is [mcux_config.txt](https://github.com/user-attachments/files/19086802/mcux_config.txt) (it's uploaded as .txt because git doesn't let me in other format).
The color depth is 16 bits, and the resolution is 720x1280, so it's a HD panel.

> 
> You can modify the benchmark program to only test the "Screen sized text" test case:
> 

As for the performance, I've seen that the results of the screen sized text are not the same when running all the benchmark vs running only that specific test.

Although, let's summarize the results only for that specific test.

image

I've got the results by running from FLASH and RAM. It seems like the results from RAM are identical between the master branch and this PR, but when LV_FONT_CACHE_GLYPH_CNT is set to 256, the performance is decreasing.


> This is my log. When all bitmaps are loaded, the log is no longer printed. In theory, the number of your logs should be the same as mine.

It looks like the number of misses is the same as yours.

```
[User] font_bitmap_create_cb: Created font bitmap cache entry for glyph_index 45, size 14x18
[User] font_bitmap_create_cb: Created font bitmap cache entry for glyph_index 80, size 15x14
[User] font_bitmap_create_cb: Created font bitmap cache entry for glyph_index 83, size 8x14
[User] font_bitmap_create_cb: Created font bitmap cache entry for glyph_index 70, size 14x14
[User] font_bitmap_create_cb: Created font bitmap cache entry for glyph_index 78, size 24x14
[User] font_bitmap_create_cb: Created font bitmap cache entry for glyph_index 74, size 5x19
[User] font_bitmap_create_cb: Created font bitmap cache entry for glyph_index 81, size 15x19
[User] font_bitmap_create_cb: Created font bitmap cache entry for glyph_index 84, size 13x14
[User] font_bitmap_create_cb: Created font bitmap cache entry for glyph_index 86, size 14x14
[User] font_bitmap_create_cb: Created font bitmap cache entry for glyph_index 69, size 15x19
[User] font_bitmap_create_cb: Created font bitmap cache entry for glyph_index 77, size 3x19
[User] font_bitmap_create_cb: Created font bitmap cache entry for glyph_index 85, size 11x17
[User] font_bitmap_create_cb: Created font bitmap cache entry for glyph_index 66, size 13x14
[User] font_bitmap_create_cb: Created font bitmap cache entry for glyph_index 13, size 4x8
[User] font_bitmap_create_cb: Created font bitmap cache entry for glyph_index 68, size 13x14
[User] font_bitmap_create_cb: Created font bitmap cache entry for glyph_index 79, size 14x14
[User] font_bitmap_create_cb: Created font bitmap cache entry for glyph_index 72, size 15x19
[User] font_bitmap_create_cb: Created font bitmap cache entry for glyph_index 15, size 4x4
[User] font_bitmap_create_cb: Created font bitmap cache entry for glyph_index 49, size 16x18
[User] font_bitmap_create_cb: Created font bitmap cache entry for glyph_index 82, size 15x19
[User] font_bitmap_create_cb: Created font bitmap cache entry for glyph_index 71, size 11x19
[User] font_bitmap_create_cb: Created font bitmap cache entry for glyph_index 67, size 15x19
[User] font_bitmap_create_cb: Created font bitmap cache entry for glyph_index 75, size 9x24
[User] font_bitmap_create_cb: Created font bitmap cache entry for glyph_index 87, size 16x14
[User] font_bitmap_create_cb: Created font bitmap cache entry for glyph_index 39, size 14x18
[User] font_bitmap_create_cb: Created font bitmap cache entry for glyph_index 52, size 15x18
[User] font_bitmap_create_cb: Created font bitmap cache entry for glyph_index 42, size 4x18
[User] font_bitmap_create_cb: Created font bitmap cache entry for glyph_index 38, size 14x18
[User] font_bitmap_create_cb: Created font bitmap cache entry for glyph_index 37, size 19x18
[User] font_bitmap_create_cb: Created font bitmap cache entry for glyph_index 47, size 17x18
[User] font_bitmap_create_cb: Created font bitmap cache entry for glyph_index 36, size 17x18
[User] font_bitmap_create_cb: Created font bitmap cache entry for glyph_index 55, size 20x18
[User] font_bitmap_create_cb: Created font bitmap cache entry for glyph_index 73, size 14x19
[User] font_bitmap_create_cb: Created font bitmap cache entry for glyph_index 46, size 21x18
[User] font_bitmap_create_cb: Created font bitmap cache entry for glyph_index 34, size 21x18
[User] font_bitmap_create_cb: Created font bitmap cache entry for glyph_index 89, size 14x14
[User] font_bitmap_create_cb: Created font bitmap cache entry for glyph_index 52, size 9x10
[User] font_bitmap_create_cb: Created font bitmap cache entry for glyph_index 68, size 8x8
[User] font_bitmap_create_cb: Created font bitmap cache entry for glyph_index 83, size 5x8
[User] font_bitmap_create_cb: Created font bitmap cache entry for glyph_index 70, size 8x8
[User] font_bitmap_create_cb: Created font bitmap cache entry for glyph_index 79, size 8x8
[User] font_bitmap_create_cb: Created font bitmap cache entry for glyph_index 84, size 7x8
[User] font_bitmap_create_cb: Created font bitmap cache entry for glyph_index 74, size 2x11
[User] font_bitmap_create_cb: Created font bitmap cache entry for glyph_index 91, size 7x8
[User] font_bitmap_create_cb: Created font bitmap cache entry for glyph_index 69, size 9x11
[User] font_bitmap_create_cb: Created font bitmap cache entry for glyph_index 85, size 6x10
[User] font_bitmap_create_cb: Created font bitmap cache entry for glyph_index 89, size 8x8
[User] font_bitmap_create_cb: Created font bitmap cache entry for glyph_index 27, size 3x8
[User] font_bitmap_create_cb: Created font bitmap cache entry for glyph_index 17, size 9x10
[User] font_bitmap_create_cb: Created font bitmap cache entry for glyph_index 39, size 8x10
[User] font_bitmap_create_cb: Created font bitmap cache entry for glyph_index 49, size 9x10
[User] font_bitmap_create_cb: Created font bitmap cache entry for glyph_index 13, size 3x5
[User] font_bitmap_create_cb: Created font bitmap cache entry for glyph_index 6, size 12x10
[User] font_bitmap_create_cb: Created font bitmap cache entry for glyph_index 36, size 10x10
[User] font_bitmap_create_cb: Created font bitmap cache entry for glyph_index 54, size 9x10
[User] font_bitmap_create_cb: Created font bitmap cache entry for glyph_index 71, size 6x11
[User] font_bitmap_create_cb: Created font bitmap cache entry for glyph_index 15, size 3x3
[User] font_bitmap_create_cb: Created font bitmap cache entry for glyph_index 78, size 13x8
[User] font_bitmap_create_cb: Created font bitmap cache entry for glyph_index 30, size 8x6
[User] font_bitmap_create_cb: Created font bitmap cache entry for glyph_index 12, size 8x7
[User] font_bitmap_create_cb: Created font bitmap cache entry for glyph_index 77, size 2x11
[User] font_bitmap_create_cb: Created font bitmap cache entry for glyph_index 86, size 8x8
[User] font_bitmap_create_cb: Created font bitmap cache entry for glyph_index 73, size 8x11
[User] font_bitmap_create_cb: Created font bitmap cache entry for glyph_index 9, size 4x14
[User] font_bitmap_create_cb: Created font bitmap cache entry for glyph_index 93, size 2x14
[User] font_bitmap_create_cb: Created font bitmap cache entry for glyph_index 10, size 4x14
[User] font_bitmap_create_cb: Created font bitmap cache entry for glyph_index 18, size 4x10
[User] font_bitmap_create_cb: Created font bitmap cache entry for glyph_index 22, size 8x10
[User] font_bitmap_create_cb: Created font bitmap cache entry for glyph_index 21, size 10x10
[User] font_bitmap_create_cb: Created font bitmap cache entry for glyph_index 23, size 9x10
[User] font_bitmap_create_cb: Created font bitmap cache entry for glyph_index 20, size 8x10
[User] font_bitmap_create_cb: Created font bitmap cache entry for glyph_index 19, size 8x10
[User] font_bitmap_create_cb: Created font bitmap cache entry for glyph_index 25, size 9x10
[User] font_bitmap_create_cb: Created font bitmap cache entry for glyph_index 24, size 8x10
[User] font_bitmap_create_cb: Created font bitmap cache entry for glyph_index 26, size 8x10
```

cosmindanielradu19:
> @cosmindanielradu19 
> 
> 
> 
> I noticed several configuration optimization items. The lvgl configuration items are a crucial factor affecting performance.
> 
> 
> 
> ```bash
> 
> # The upper limit of animation frame rate is adjusted to 60FPS
> 
> CONFIG_LV_DEF_REFR_PERIOD=16
> 
> 
> 
> # Disable all ASSERT check (There are a lot of checks in the lv_cache module, which will affect the running speed.)
> 
> CONFIG_LV_USE_VGLITE_ASSERT=0
> 
> CONFIG_LV_USE_ASSERT_NULL=0
> 
> CONFIG_LV_USE_ASSERT_MALLOC=0
> 
> ```
> 
> 
> 
> And adjust the compilation optimization level to O3 to ensure that the code runs at the highest efficiency.

The runs from RAM were using -O3 optimization flag.

I'll come back with another set of results based on the asserts update suggestions

FASTSHIFT:
> > @cosmindanielradu19
> > I noticed several configuration optimization items. The lvgl configuration items are a crucial factor affecting performance.
> > ```shell
> > # The upper limit of animation frame rate is adjusted to 60FPS
> > 
> > CONFIG_LV_DEF_REFR_PERIOD=16
> > 
> > 
> > 
> > # Disable all ASSERT check (There are a lot of checks in the lv_cache module, which will affect the running speed.)
> > 
> > CONFIG_LV_USE_VGLITE_ASSERT=0
> > 
> > CONFIG_LV_USE_ASSERT_NULL=0
> > 
> > CONFIG_LV_USE_ASSERT_MALLOC=0
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > And adjust the compilation optimization level to O3 to ensure that the code runs at the highest efficiency.
> 
> The runs from RAM were using -O3 optimization flag.
> 
> I'll come back with another set of results based on the asserts update suggestions

Can you help compare the performance data of LV_OS_NONE (without OS)?

cosmindanielradu19:
> > > @cosmindanielradu19
> 
> > > I noticed several configuration optimization items. The lvgl configuration items are a crucial factor affecting performance.
> 
> > > ```shell
> 
> > > # The upper limit of animation frame rate is adjusted to 60FPS
> 
> > > 
> 
> > > CONFIG_LV_DEF_REFR_PERIOD=16
> 
> > > 
> 
> > > 
> 
> > > 
> 
> > > # Disable all ASSERT check (There are a lot of checks in the lv_cache module, which will affect the running speed.)
> 
> > > 
> 
> > > CONFIG_LV_USE_VGLITE_ASSERT=0
> 
> > > 
> 
> > > CONFIG_LV_USE_ASSERT_NULL=0
> 
> > > 
> 
> > > CONFIG_LV_USE_ASSERT_MALLOC=0
> 
> > > ```
> 
> > > 
> 
> > > 
> 
> > >     
> 
> > >       
> 
> > >     
> 
> > > 
> 
> > >       
> 
> > >     
> 
> > > 
> 
> > >     
> 
> > >   
> 
> > > And adjust the compilation optimization level to O3 to ensure that the code runs at the highest efficiency.
> 
> > 
> 
> > The runs from RAM were using -O3 optimization flag.
> 
> > 
> 
> > I'll come back with another set of results based on the asserts update suggestions
> 
> 
> 
> Can you help compare the performance data of LV_OS_NONE (without OS)?

Sure, I'll come back with results for both FreeRTOS and OS_NONE

FASTSHIFT:
@cosmindanielradu19 

Is the delay after your `lv_timer_handler` using a fixed time or a dynamic time as described in the [document](https://docs.lvgl.io/master/details/integration/adding-lvgl-to-your-project/timer_handler.html)? 

image

If a fixed time is used, LVGL will not be able to start rendering in time, which will also affect rendering performance.

cosmindanielradu19:
> @cosmindanielradu19
> 
> Is the delay after your `lv_timer_handler` using a fixed time or a dynamic time as described in the [document](https://docs.lvgl.io/master/details/integration/adding-lvgl-to-your-project/timer_handler.html)?
> 
> image
> 
> If a fixed time is used, LVGL will not be able to start rendering in time, which will also affect rendering performance.

Yes, a dynamic time is used

```c
    for (;;)
    {
        vTaskDelay(lv_timer_handler());
    }
```

FASTSHIFT:
> What is the purpose of this caching in the first place? To use the correctly aligned bitmaps with good stride, right?
> 
> What if we say this:
> 
> * Update the font converter with bitmap alignment and bitmap stride options
> * It's the user's responsibility to use a bitmap which works well with the GPU.
> * FreeType and TinyTTF will create `draw_buf`s anyway which the required buffers settings.
> 
> So no caching is required, but C array fonts can be used as static bitmaps. It's solve a lot of issues at once. We do the same with images too.
> 
> This way to current behavior in `lv_font_get_glyph_bitmap` to convert bitmaps to `A8` can be an optional helper to provide full control to the user.
> 
> What do you think?

I agree with this solution. In this PR, I deleted the cache of built-in fonts and only kept the freetype bitmap reference count adaptation of vg_lite.

kisvegabor:
Just to make sure that we are on the same page. In order to make rendering faster we should add these to the font converter:
1. `--stride` to set the stride for each bitmaps
2. Store the `stride` in `lv_font_glyph_dsc_t`
2. `--align` to set the alignment of each bitmaps, meaning all bitmaps will be padded to make sure that the alignment of the next bitmap is fulfilled. 
3. We should add `LV_ATTRIBUTE_MEM_ALIGN` (or `LV_ATTRIBUTE_FONT_BITMAP_ALIGN`) before the bitmap arrays to make sure that the large array is aligned. 

After that we VGLite, Nema, and DAVE/2D can be improved to use the bitmaps directly if
- `glyph_draw_dsc->format == LV_FONT_GLYPH_FORMAT_A1/2/4/8_ALIGNED`
- `glyph_draw_dsc->stride` is correct
- the raw bitmap is aligned.

Do you agree?

FASTSHIFT:
> Just to make sure that we are on the same page. In order to make rendering faster we should add these to the font converter:
> 
> 1. `--stride` to set the stride for each bitmaps
> 2. Store the `stride` in `lv_font_glyph_dsc_t`
> 3. `--align` to set the alignment of each bitmaps, meaning all bitmaps will be padded to make sure that the alignment of the next bitmap is fulfilled.
> 4. We should add `LV_ATTRIBUTE_MEM_ALIGN` (or `LV_ATTRIBUTE_FONT_BITMAP_ALIGN`) before the bitmap arrays to make sure that the large array is aligned.
> 
> After that we VGLite, Nema, and DAVE/2D can be improved to use the bitmaps directly if
> 
> * `glyph_draw_dsc->format == LV_FONT_GLYPH_FORMAT_A1/2/4/8_ALIGNED`
> * `glyph_draw_dsc->stride` is correct
> * the raw bitmap is aligned.
> 
> Do you agree?

I totally agree.

FASTSHIFT:
@kisvegabor Can we merge this PR for VG-Lite to draw freetype fonts first? :)

uLipe (approved):

kisvegabor (approved):

PR #7420, opened 2024-12-06
perf(os): optimize OS API calls without OS mode

FASTSHIFT:
Since compilation does not necessarily optimize cross-file empty functions well (-O3 optimization alone is not enough unless LTO optimization is enabled), In the absence of an operating system, use inline functions to help compile optimizations and avoid the call overhead of the OS API to ensure no performance penalty.

Especially in scenarios where large amounts of text are used, cache requests will be very frequent, resulting in many meaningless `lv_mutex` calls.

The following figure compares the results of the [decompilation analysis tool](https://github.com/NationalSecurityAgency/ghidra) and the trace data.

Fix before:
image

image

Fix after:
image

image

It is measured that the `lv_cache_acquire` function can reduce the time consumption by more than 100us per frame on real devices. 

### Notes


FASTSHIFT:
Assembly code and address of `lv_mutex_lock`:
image

After calculation, the address difference between the `lv_cache_acquire` function and the `lv_mutex_lock` function is more than 25K bytes, which can easily trigger an instruction cache miss and waste cache line space.

FASTSHIFT:
image
The same effect can be achieved by using static inline, which can maintain the characteristics of the function.

uLipe (dismissed):
Nice work!

uLipe (approved):

kisvegabor (approved):
Great!

PR #7429, opened 2024-12-09
feat(demo): add smartwatch demo

fbiego:
Smart watch UI demo

https://github.com/user-attachments/assets/7fc62cd0-7d83-494f-958f-128f3676604d

### Notes


Kconfig
```diff
@@ -2013,6 +2013,9 @@ menu "LVGL configuration"
 			bool "vector graphic demo"
 			default n
 			depends on LV_USE_VECTOR_GRAPHIC
+		config LV_USE_DEMO_SMARTWATCH
+			bool "Smartwatch demo"
+			default n
```
uLipe: Nit: Kconfig symbols that are bool, are default disabled, you can just rremove this line

demos/smartwatch/lv_demo_smartwatch.c
```diff
@@ -0,0 +1,272 @@
...
+                lv_sqrt_res_t res;
+                lv_sqrt(x_sqr, &res, 0x8000); /*Use lvgl's built in sqrt root function*/
+                x = r - res.i;
+            }
+
+            /*Translate the item by the calculated X coordinate*/
+            lv_obj_set_style_translate_x(child, circular_scroll ? x : 0, 0);
+
+            /* Use some opacity with larger translations */
+            // lv_opa_t opa = lv_map(x, 0, r, LV_OPA_TRANSP, LV_OPA_COVER);
```
uLipe: If they are not used, could you remove these lines?

demos/smartwatch/lv_demo_smartwatch.c
```diff
@@ -0,0 +1,272 @@
...
+ *   STATIC FUNCTIONS
+ **********************/
+
+static void dialog_close_event_cb(lv_event_t * e)
+{
+    lv_event_code_t event_code = lv_event_get_code(e);
+    lv_obj_t * target = lv_event_get_target(e);
+    if(event_code == LV_EVENT_CLICKED) {
+        lv_obj_add_flag(dialog_window, LV_OBJ_FLAG_HIDDEN);
+    }
```
uLipe, line 372: an else with a LOG_WARN /INFO  would be good here to indicate unhandled events that may happen, it would be a nice debug source.

demos/smartwatch/lv_demo_smartwatch.h
```diff
@@ -0,0 +1,462 @@
...
+#error "LV_FONT_MONTSERRAT_20 text support is required. Enable it in lv_conf.h (LV_FONT_MONTSERRAT_20 1)"
+#endif
+#if LV_FONT_MONTSERRAT_46 == 0
+#error "LV_FONT_MONTSERRAT_46 text support is required. Enable it in lv_conf.h (LV_FONT_MONTSERRAT_46 1)"
+#endif
+#if LV_FONT_MONTSERRAT_48 == 0
+#error "LV_FONT_MONTSERRAT_48 text support is required. Enable it in lv_conf.h (LV_FONT_MONTSERRAT_48 1)"
+#endif
+#if LV_USE_STDLIB_MALLOC == LV_STDLIB_BUILTIN && LV_MEM_SIZE < 144 * 1024
+#warning "It's recommended to have at least 144kB RAM for the smartwatch demo"
```
uLipe: I think memory requirements needs to be strict to maximize the user experience, so building the demo on a platform that has less RAM than the required should not be allowed.

Showing an error instead of warning, would enforce people to optimize the RAM consumption to get this demo runnable on more platforms.

uLipe (commented):
Great work!

It is just my first pass.

Kconfig
fbiego: the other symbols for demos have been kept this way. I'm following the same to maintain consistency

demos/smartwatch/lv_demo_smartwatch_easter_egg.c
```diff
@@ -0,0 +1,340 @@
...
+
+/*********************
+ *      INCLUDES
+ *********************/
+#include "lv_demo_smartwatch.h"
+#if LV_USE_DEMO_SMARTWATCH
+
+#include "lv_demo_smartwatch_easter_egg.h"
+
+#include "misc/lv_timer_private.h"
```
kisvegabor:
```suggestion
#include "../../lvgl_private.h"
```

demos/smartwatch/lv_demo_smartwatch_notifications.c
```diff
@@ -0,0 +1,344 @@
+/**
```
kisvegabor, line 1: Could you add more realistic sample messages?

demos/smartwatch/lv_demo_smartwatch_home.h
```diff
@@ -0,0 +1,116 @@
...
+ * Create the home page. Called only once.
+ */
+void lv_demo_smartwatch_home_create(void);
+
+/**
+ * load the home screen
+ * @param anim_type    screen load animation to use
+ * @param time    animation time
+ * @param delay     delay time before loading the screen
+ */
```
kisvegabor: Please format these like this:
```suggestion
 * Load the home screen
 * @param anim_type    screen load animation to use
 * @param time         animation time
 * @param delay        delay time before loading the screen
 */
```

kisvegabor (commented):
First of all great work! 

As I was testing it and looking into the code I got some ideas:
1. What about using a tileview instead of gestures for navigation? This way a unified scroll-based navigation could be realised across the demo. 
2. It'd be amazing if we could support 240x240 ... 400x400 screens with this demo. It seems to me by using the same assets just a little but larger paddings and  % units it would be possible. What do you think?

demos/smartwatch/lv_demo_smartwatch.h
liamHowatt, line 1: Nice API. Would it be possible to use the same prefix for all the functions the user is meant to call? Either `lv_demo_smartwatch_` or `lv_smartwatch_`. If some aren't meant to be called by users, mayebe create an `lv_demo_smartwtch_private.h` as is done in other places in LVGL. The `img_` image declarations could go there too, if you decide to do it.

liamHowatt (commented):
Nice!

demos/smartwatch/lv_demo_smartwatch.c
uLipe: EDIT: Nevermind, it seems  I was looking on a old code revision

demos/smartwatch/lv_demo_smartwatch.c
uLipe: EDIT: NVM i was looking on an old code revision

Kconfig
uLipe: I understand this could be a nit, but on newly added PR the Kconfig bool symbols are being addressed to reflect the Kconfig standards, it would be good that newly added ones follow its good practices and we would enforce to do the same on the other places where this occur.

So please reconsider it.

uLipe (changes requested):
Please have a check on the Kconfig comment, and please let the reviewer to resolve the discussion except on trivial cases.

Otherwise, this demo is awesome!

demos/smartwatch/lv_demo_smartwatch_notifications.c
fbiego: I have updated it. As for the tileview I was considering using it but since the original code did not have this, it would take longer to adapt it to the existing but we can still update it later. Definitely, making it support up to 400x400 is in the plan

Kconfig
fbiego: updated

uLipe (dismissed):
LGTM, although someone more experienced on the demos might have to approve :)

demos/smartwatch/lv_demo_smartwatch_private.h
```diff
@@ -0,0 +1,245 @@
...
+
+
+#if LV_USE_DEMO_SMARTWATCH
+
+
+/*********************
+ *      DEFINES
+ *********************/
+
+LV_IMG_DECLARE(img_weather_day_0);          // assets/dy-0.png
```
liamHowatt:
```suggestion
LV_IMG_DECLARE(img_weather_day_0);          /* assets/dy-0.png */
```

And others in the file

liamHowatt (commented):
Otherwise, looks great to me!

demos/smartwatch/lv_demo_smartwatch_private.h
fbiego: updated

liamHowatt (commented):
Sorry for spreading out my review. I noticed some warnings and a few more things while trying it out.

- C standard `(void)` for no parameters: https://github.com/liamHowatt/lvgl/commit/c994109ff6c34935011b02a5b447af800bf524ab
- Automatic pre-commit typo hook results: https://github.com/liamHowatt/lvgl/commit/4dc22266c6b81995182bef7492d9765c65cc24f6
- unuseds and mismatched signedness comparisons. This one fixes warnings about things that aren't very important, but someone would probably open a PR for them soon after merging anyways: https://github.com/liamHowatt/lvgl/commit/1ceeb74a7131477d4a1a87006a971d8156127190

I thought CI caught these but I guess not for demos.

Also, what's the recommended screen size?

uLipe:
@liamHowatt does the CI build use the option `-Werror` ? Or the warnings are just not appearing as they shoud be?

fbiego:
> Also, what's the recommended screen size?
240x240, it will also be adapted for up to 400x400

liamHowatt (dismissed):
> 240x240, it will also be adapted for up to 400x400

Got it, thanks

liamHowatt:
> @liamHowatt does the CI build use the option `-Werror` ? Or the warnings are just not appearing as they shoud be?

@uLipe I just checked and it looks like the standard warnings used in the CI are used only for `lvgl` and `lvgl_examples`, not `lvgl_demos`.

uLipe:
@liamHowatt should we address this on this PR? Otherwise I can check it, and address on a separate one to not mix the subjects, what do you think?

uLipe (dismissed):
Again, great work!

lhdjply:
A very good example. Just one suggestion: it would be best to include the original image files (such as PNG files) as well.

lhdjply:
https://github.com/lvgl/lvgl/blob/cc43f12257604f885806e97930b62ff4f577cdd6/tests/src/lv_test_conf_full.h#L107-L118

Please adding #define LV_USE_DEMO_SMARTWATCH 0 to this file ,it can test the demo in the CI environment.

demos/smartwatch/lv_demo_smartwatch_private.h
```diff
@@ -0,0 +1,245 @@
...
+
+
+#if LV_USE_DEMO_SMARTWATCH
+
+
+/*********************
+ *      DEFINES
+ *********************/
+
+LV_IMG_DECLARE(img_weather_day_0);          /* assets/dy-0.png */
```
lhdjply: Please use `LV_IMAGE_DECLARE` to replace `LV_IMG_DECLARE`. Because if `LV_IMG_DECLARE` is used, the trailing semicolon needs to be removed.

demos/smartwatch/lv_demo_smartwatch.c
```diff
@@ -178,7 +178,11 @@ bool lv_demo_smartwatch_get_load_app_list(void)
 
 static void dialog_close_event_cb(lv_event_t * e)
 {
-    lv_obj_add_flag(dialog_window, LV_OBJ_FLAG_HIDDEN);
+    lv_event_code_t event_code = lv_event_get_code(e);
+    if(event_code == LV_EVENT_CLICKED) {
+        lv_obj_add_flag(dialog_window, LV_OBJ_FLAG_HIDDEN);
+    }
+
 }
```
liamHowatt: FYI, if you added this just to silence the unused parameter warning, you can also do `LV_UNUSED(e);`. It's ok to have unused parameters.

liamHowatt (commented):
> Please adding #define `LV_USE_DEMO_SMARTWATCH 1` to this file ,it can test the demo in the CI environment.

@uLipe I missed this. You explicitly opt specific demos into the CI apparently. My bad. Thanks @lhdjply

demos/smartwatch/lv_demo_smartwatch.c
fbiego: got it, thanks. I wasn't sure what to do with the unused parameters

kisvegabor:
I works well! :tada: 

Just a few things that I have found:
- The circular lists as straight by default and they needed to be scrolled a little to make them circlar
- Can you use tileview when loading e.g. the WhatsApp and other messages?
- Could you add an indication (e.g. arrow, or little bullets) about the the possible scroll directions and position to improve the user experience? I feel like it's a little bit easy to get lost in the menu :slightly_smiling_face:

fbiego:
@kisvegabor please review,
made changes
- tileview on notifications screen
- fixed circular scrolled lists
- added scroll direction hints

kisvegabor:
Thank you! It's very useful! 

As new user of the smartwatch I found a few further optimization options:


image
On this screen loading the QR code from the top makes me assume that I can also scroll there. I suggest loading it without scroll or allow scrolling to it.

---

image
There are no hints here.

---

image
What about placing this next to the above 2 weather screens to keep the weather app less scattered?
 
---

image
As this is the main screen can we add permanent hints here?
E.g. icons like a ⚙️ on the right, a ☀️ on the bottom, 🎵 or so on the top and ✉️ on the right.

---

In general can you show the hints for ~1000 ms when a tile is loaded a hide it with a fade animation? This way I don't have to make a mistake first to see what I can do. 

---

What do you think?

fbiego:
Got it, working on the changes.

> image
> What about placing this next to the above 2 weather screens to keep the weather app less scattered?

However, i'm not sure what you mean by this. Could you clarify?
The current Weather app is structured as below

Screenshot 2025-01-13 at 10 52 40

kisvegabor:
Oh, I see the problem. Having a horizontally scrollable tileplaced next other tiles would be hard to use, as you need scroll through it's content to change tile.  Please ignore this comment.

kisvegabor (approved):
I like it! As we have discussed we can improve the design a little in a new iteration. 

It's also great that you have enabled the demo in `lv_test_conf_full.h`.

liamHowatt (approved):

uLipe (approved):

PR #7430, opened 2024-12-09
feat(refr): improve performance measurement

FASTSHIFT:
Clean up performance blind spots to facilitate time-consuming analysis of the entire rendering process.

image

### Notes


FASTSHIFT:
I noticed a few points that could be optimized:

1. LVGL v9 introduces more draw task dispatch mechanism than LVGL v8. It needs to copy draw dsc to the task linked list frequently and release memory immediately after drawing. It has been measured that the creation and destruction of the draw task takes about 400us in a trace rendering of 16ms a frame. Can we implement in no-OS mode draw dsc directly using stack parameters without additional memory copy? Or introduce a draw dsc pool, draw dsc memory uniformly allocated from the pool, minimize malloc/free and memory copy.



2. `lv_obj_get_style_opa_recursive` also consumed a lot of time, the total rendering time of 16ms was close to 700us, the reason why it was slow was that every time it needed to recurse upward to the deepest parent for transparency blending calculation, can we use the stack to cache the opa value? It only needs to be transparently blended with the top of the stack each time, there is no need for a recursive parent, and the cache value can be stored in the layer structure.

uLipe (approved):

kisvegabor:
Very insightful! Can you share the whole table from this?
image


 > lv_obj_get_style_opa_recursive also consumed a lot of time,
 
Good idea! The layer can save the current `opa`, and when a child comes,
1. it saves the layer opa
2. it gets its own `opa`
3. updates the layer `opa`
4. restore the layer open when all its children are rendered

We can do the opa scaling in `lv_draw_label` (and similar functions) when the draw task is created.  

In the same way we might want to save some inherited properties too. E.g. the color filter is repeatedly called multiply times for for all the color like properties.   An optimization similar to opa could work like:
1. The screen gets its `color_filter_dsc` in a non inherited way and saves it in the layer
2. The child checks if `color_filter_dsc` is set in its styles and if not set just use the layer's `color_filter_dsc`
3. Whatever the final `color_filter_dsc` is, it should be saved in the layer, and restored (from a local variable) when all the children are rendered.   

Text related properties can be cached too, but the gain is less obvious as its not sure that the text properties will be required later. 

> LVGL v9 introduces more draw task dispatch mechanism than LVGL v8. It needs to copy draw dsc to the task linked list frequently and release memory immediately after drawing.

It's definitely more complicated. I'd approach it differently than the mempool. My idea is that what if we create the `draw_task` first, and set the `draw_dsc` directly in it. E.g.
Instead of:
```c
    lv_draw_label_dsc_t txt_dsc;
    lv_draw_label_dsc_init(&txt_dsc);
    lv_obj_init_draw_label_dsc(obj, LV_PART_MAIN, &txt_dsc);
    txt_dsc.text = "Some text";
    lv_area_t txt_area = {...};
    lv_draw_label(layer, &txt_dsc, &txt_area);
```
Do this:
```c
lv_draw_task_t * t = lv_draw_task_create(LV_DRAW_TASK_TYPE_LABEL); //Just allocation
lv_draw_label_dsc_init(t->draw_dsc);
lv_obj_init_draw_label_dsc(obj, LV_PART_MAIN, t->draw_dsc);
t->draw_dsc->text = "Some text";
lv_area_t txt_area = {...};
lv_draw_task_set_area(t, txt_area);  //New function
lv_draw_task_add_to_layer(layer, t); //New function ()
```

We could gradually migrate to this drawing in the widgets starting with most intensive draw task creators.  

I suspect that the `memzero` in `lv_draw_label_dsc_init` (and others) takes a lot of time too. Maybe it can be also saves somehow. 

---

Anyway, this PR looks good 😊

FASTSHIFT:
> Very insightful! Can you share the whole table from this?

It's here: [trace.txt](https://github.com/user-attachments/files/18071727/trace_no_opa_return.txt)

FASTSHIFT:
> It's definitely more complicated. I'd approach it differently than the mempool. 

In fact, malloc/free has a greater impact. Our system uses the libc standard memory allocation interface, which makes it easy to analyze memory leaks and memory trampling problems. However, since it is a public system API, each operation involves locking and unlocking operations, which introduces additional overhead.
The benefit of introducing a memory pool is that it can reduce malloc/free calls. In single-threaded mode, there is no need to lock, and there is no overhead for memory defragmentation. We can consider reintroducing the `lv_mem_buf_get/release` interface of LVGL v8 to minimize dynamic memory allocation in limited rendering time.

image

[trace.txt](https://github.com/user-attachments/files/18072604/trace.txt)

FASTSHIFT:
The performance of the two heap modes is actually similar:

system heap:
image

lvgl buildin heap(TLSF):
image

FASTSHIFT:
> I suspect that the memzero in lv_draw_label_dsc_init (and others) takes a lot of time too. Maybe it can be also saves somehow.

There is no memory clearing operation inside `lv_obj_init_draw_label_dsc` , mainly due to the time consumption of style reading:
image

image

`LV_OBJ_STYLE_CACHE` is turned on.

FASTSHIFT:
In any case, we can merge this performance measurement PR first, and more detailed performance optimization issues can be discussed in a separate PR :)

XuNeo (approved):

kisvegabor:
Thank you the log!

> As shown in the figure below, in the total rendering time of 13ms, dynamic memory allocation takes a total of about 250us:

It's only 2% now for all mallocs. I seems like a reasonable overhead to me. With some complex logic maybe we could save half of it. IMO it doesn't worth the effort. 

`lv_obj_get_style_opa_recursive` is 1ms, which is 7%. By optimizing it we could save ~6%. I think the `color_filter`s is even more. So I'd rather focus on these where we hope 10+ % gain.

PR #7446, opened 2024-12-11
fix(draw_rect): refactor to insure header is initialized

pfvogel:
Static analysis tool is reporting that the "header" variable may be uninitialized prior to use.  This appears to be a false positive, as the path through the code does not use "header" for the case where it is not initialized.  To eliminate the analysis tool warnings, refactor the code to always initialize "header"

### DIFF
--- a/src/draw/lv_draw_rect.c
+++ b/src/draw/lv_draw_rect.c
@@ -201,11 +201,12 @@ void lv_draw_rect(lv_layer_t * layer, const lv_draw_rect_dsc_t * dsc, const lv_a
         if(src_type == LV_IMAGE_SRC_VARIABLE || src_type == LV_IMAGE_SRC_FILE) {
             res  = lv_image_decoder_get_info(dsc->bg_image_src, &header);
         }
-        else if(src_type == LV_IMAGE_SRC_UNKNOWN) {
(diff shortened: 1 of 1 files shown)


uLipe (commented):
Would you share you testbench / static analysis tool in order to enable us to reproduce something similar? 

Thank you!

pfvogel:
I am using [codechecker](https://codechecker.readthedocs.io/en/latest) and it was pretty simple to run on the code.

The script I use to run it in a build directory looks like:

```
make clean

CodeChecker log -b make -o compilation.json
CodeChecker analyze ./compilation.json --output ./reports --enable sensitive --ctu
CodeChecker parse --export html --output html/reports ./reports
```
clang and cppcheck should be installed as well to get all of the checks to run.

uLipe:
Thank you @pfvogel this could open the path to creat a github action that reports this static analysis automatically.

uLipe (approved):
:LGTM so far, just need someone else to approve

pfvogel:
We use that tool on a number of our projects and has been very helpful pointing out potential issues.  There are some false positives, but overall it works very well.  It can take a while to run, though.

liamHowatt (approved):

PR #7447, opened 2024-12-11
fix(draw_sw_img): avoid divide by zero

pfvogel:
Add check to set buf_stride to 1 if it was 0 to avoid a division by zero.

### DIFF
--- a/src/draw/sw/lv_draw_sw_img.c
+++ b/src/draw/sw/lv_draw_sw_img.c
@@ -388,6 +388,9 @@ static void recolor_only(lv_draw_unit_t * draw_unit, const lv_draw_image_dsc_t *
     uint8_t * tmp_buf;
     int32_t buf_h;
     uint32_t buf_stride = blend_w * px_size;
+    if(buf_stride == 0) {
+        buf_stride = 1;
+    }
     buf_h = MAX_BUF_SIZE / buf_stride;
     if(buf_h > blend_h) buf_h = blend_h;
     tmp_buf = lv_malloc(buf_stride * buf_h);


src/draw/sw/lv_draw_sw_img.c
```diff
@@ -388,6 +388,9 @@ static void recolor_only(lv_draw_unit_t * draw_unit, const lv_draw_image_dsc_t *
     uint8_t * tmp_buf;
     int32_t buf_h;
     uint32_t buf_stride = blend_w * px_size;
+    if(buf_stride == 0) {
+        buf_stride = 1;
+    }
```
FASTSHIFT, line 393: Wouldn't it be better to use return?

pfvogel: The buf_stride is used to generate the size of a temporary buffer for recoloring.  The line after the buf_stride value is used does a range check to make sure the allocated area is in range:

```
    buf_h = MAX_BUF_SIZE / buf_stride;
    if(buf_h > blend_h) buf_h = blend_h;
    tmp_buf = lv_malloc(buf_stride * buf_h);
```
So the allocated buffer size is constrained, so this part should be safe.

However, if buf_stride is zero somehow (generated from this equation)

`    uint32_t buf_stride = blend_w * px_size;`

then this assignment later in the code:

`  blend_dsc.src_stride = blend_w * px_size;`

will also be zero.  This will probably cause problems when lv_draw_sw_blend() is called later with blend_dsc being passed as a parameter.

Maybe it would be better to add an assertion that buf_stride is not zero after it is calculated?

uLipe (approved):

FASTSHIFT (approved):

PR #7449, opened 2024-12-11
fix(draw_sw_triangle): add null check

pfvogel:
In lv_draw_sw_triangle(), grad is allowed to be NULL but is dereferenced without checking to see if it is NULL.  Add NULL check before dereferencing it.


uLipe (commented):
Sounds a trivial change but before approve, could you point where are some examples of grad gets de-referenced?

uLipe:
Oh you're right, sorry I overlooked it.

uLipe (approved):

liamHowatt (approved):

PR #7457, opened 2024-12-13
feat(osal): import sdl2 for thread

M1NGS:
Through SDL2, LVGL can easily achieve multi-threaded rendering on PC (OS2/BSD/MacOS/Windows/Linux/Android/iOS) or game consoles (n3ds/ngage/ps2/psp/switch) without additional work.
In addition, a benefit is that for cross-platform applications/games, we don't have to modify lv_config.h before build library.

### DIFF
--- a/src/osal/lv_sdl2.c
+++ b/src/osal/lv_sdl2.c
@@ -45,6 +45,10 @@ lv_result_t lv_thread_init(lv_thread_t * thread, lv_thread_prio_t prio, void (*c
     thread->callback = callback;
     thread->user_data = user_data;
     thread->thread = SDL_CreateThreadWithStackSize(generic_callback, "LVGL", stack_size, thread);
+    if(thread->thread == NULL) {
+        LV_LOG_ERROR("Error: %s", SDL_GetError());
+        return LV_RESULT_INVALID;
+    }
     return LV_RESULT_OK;
 }
 
@@ -53,7 +57,7 @@ lv_result_t lv_thread_delete(lv_thread_t * thread)
     int ret;
     SDL_WaitThread(thread->thread, &ret);
     if(ret != 0) {
-        LV_LOG_WARN("Error: %d", ret);
+        LV_LOG_ERROR("Error: %d", ret);
         return LV_RESULT_INVALID;
     }
 
@@ -65,7 +69,7 @@ lv_result_t lv_mutex_init(lv_mutex_t * mutex)
     *mutex = SDL_CreateMutex();
 
     if(*mutex == NULL) {
-        LV_LOG_WARN("Error: %s", SDL_GetError());
+        LV_LOG_ERROR("Error: %s", SDL_GetError());
         return LV_RESULT_INVALID;
     }
     else {
@@ -77,7 +81,7 @@ lv_result_t lv_mutex_lock(lv_mutex_t * mutex)
 {
     int ret = SDL_LockMutex(*mutex);
     if(ret) {
-        LV_LOG_WARN("Error: %d", ret);
+        LV_LOG_ERROR("Error: %d", ret);
         return LV_RESULT_INVALID;
     }
     else {
@@ -89,7 +93,7 @@ lv_result_t lv_mutex_lock_isr(lv_mutex_t * mutex)
 {
     int ret = SDL_LockMutex(*mutex);
     if(ret) {
-        LV_LOG_WARN("Error: %d", ret);
+        LV_LOG_ERROR("Error: %d", ret);
         return LV_RESULT_INVALID;
     }
     else {
@@ -101,7 +105,7 @@ lv_result_t lv_mutex_unlock(lv_mutex_t * mutex)
 {
     int ret = SDL_UnlockMutex(*mutex);
     if(ret) {
-        LV_LOG_WARN("Error: %d", ret);
+        LV_LOG_ERROR("Error: %d", ret);
         return LV_RESULT_INVALID;
     }
     else {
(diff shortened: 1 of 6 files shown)


src/osal/lv_sdl2.c
```diff
@@ -0,0 +1,171 @@
...
+}
+
+/**********************
+ *   STATIC FUNCTIONS
+ **********************/
+
+static int generic_callback(void * user_data)
+{
+    lv_thread_t * thread = user_data;
+    thread->callback(thread->user_data);
```
uLipe, line 179: Could you check if the callback is valid before calling it, please?

src/osal/lv_sdl2.c
```diff
@@ -0,0 +1,171 @@
...
+    thread->thread = SDL_CreateThreadWithStackSize(generic_callback, "LVGL", stack_size, thread);
+    return LV_RESULT_OK;
+}
+
+lv_result_t lv_thread_delete(lv_thread_t * thread)
+{
+    int ret;
+    SDL_WaitThread(thread->thread, &ret);
+    if(ret != 0) {
+        LV_LOG_WARN("Error: %d", ret);
```
uLipe: I think it is better ot use LOG_ERR instead.

src/osal/lv_sdl2.c
```diff
@@ -0,0 +1,171 @@
...
+
+    return LV_RESULT_OK;
+}
+
+lv_result_t lv_mutex_init(lv_mutex_t * mutex)
+{
+    *mutex = SDL_CreateMutex();
+
+    if(*mutex == NULL) {
+        LV_LOG_WARN("Error: %s", SDL_GetError());
```
uLipe: Also here

src/osal/lv_sdl2.c
```diff
@@ -0,0 +1,171 @@
...
+    else {
+        return LV_RESULT_OK;
+    }
+}
+
+lv_result_t lv_mutex_lock(lv_mutex_t * mutex)
+{
+    int ret = SDL_LockMutex(*mutex);
+    if(ret) {
+        LV_LOG_WARN("Error: %d", ret);
```
uLipe: Also here

src/osal/lv_sdl2.c
```diff
@@ -0,0 +1,171 @@
...
+    else {
+        return LV_RESULT_OK;
+    }
+}
+
+lv_result_t lv_mutex_lock_isr(lv_mutex_t * mutex)
+{
+    int ret = SDL_LockMutex(*mutex);
+    if(ret) {
+        LV_LOG_WARN("Error: %d", ret);
```
uLipe: Also here/

src/osal/lv_sdl2.c
```diff
@@ -0,0 +1,171 @@
...
+    else {
+        return LV_RESULT_OK;
+    }
+}
+
+lv_result_t lv_mutex_unlock(lv_mutex_t * mutex)
+{
+    int ret = SDL_UnlockMutex(*mutex);
+    if(ret) {
+        LV_LOG_WARN("Error: %d", ret);
```
uLipe: And also here.

src/osal/lv_sdl2.c
```diff
@@ -0,0 +1,171 @@
...
+lv_result_t lv_mutex_delete(lv_mutex_t * mutex)
+{
+    SDL_DestroyMutex(*mutex);
+    return LV_RESULT_OK;
+}
+
+lv_result_t lv_thread_sync_init(lv_thread_sync_t * sync)
+{
+    sync->mutex = SDL_CreateMutex();
+    sync->cond = SDL_CreateCond();
```
uLipe, line 129: Hmm different from other objects you are not testing they are valid after allocating the, they will be always valid, otherwise, can you validate their creation before moving forward?

uLipe (changes requested):
I just loved this patch, I have only some nits regarding pointer validation and logs.

src/osal/lv_sdl2.c
M1NGS: No problem, I created it using lv_pthread.c as a template and didn't add any extra logic，i also think that check return is necessary, but I'm not sure if there is another reason why lv_pthread.c & lv_windows.c doesn't do that.

I checked some callers like
```c
#if LV_USE_OS
static void render_thread_cb(void * ptr)
{
    lv_draw_sw_unit_t * u = ptr;

    lv_thread_sync_init(&u->sync);
``` 
```c
void lv_draw_init(void)
{
#if LV_USE_OS
    lv_thread_sync_init(&_draw_info.sync);
#endif
}
```
They didn't check the return.

However, I'll implement error handling for against a rainy day

src/osal/lv_sdl2.c
M1NGS: noticed

src/osal/lv_sdl2.c
M1NGS: I think the validity check of the callback should be do in lv_thread_init(), such as
```c
lv_result_t lv_thread_init(lv_thread_t * thread, lv_thread_prio_t prio, void (*callback)(void *), size_t stack_size,
                           void * user_data)
{
    LV_UNUSED(prio);
    if (callback == NULL)
    {
        //Handling error
    }
    thread->callback = callback;
    thread->user_data = user_data;
``` 
how did you like it?

src/osal/lv_sdl2.c
M1NGS: noticed

src/osal/lv_sdl2.c
M1NGS: noticed

src/osal/lv_sdl2.c
M1NGS: noticed

src/osal/lv_sdl2.c
M1NGS: noticed

src/osal/lv_sdl2.c
M1NGS: I think the callback check is not necessary. Users are unlikely to explicitly enter NULL, and other incorrect type inputs will be warned by the compiler.

uLipe (approved):
LGTM!

kisvegabor (approved):
Nice! I love it too!

kisvegabor:
Thank you!

PR #7459, opened 2024-12-15
fix(examples): fix infinite scroll crash

liamHowatt:
Infinite (or near enough) recursion can occur in some cases. The `lv_obj_scroll_by` calls cause the event callback to recur. Make the function return if it is reentered with the same object.

Reproducable on Emscripten (i.e. the docs) by scrolling to the bottom (-30) https://docs.lvgl.io/master/examples.html#infinite-scrolling
I tested the fix with Emscripten locally.

### Notes

### DIFF
--- a/examples/scroll/lv_example_scroll_7.c
+++ b/examples/scroll/lv_example_scroll_7.c
@@ -17,6 +17,10 @@ static lv_obj_t * load_item(lv_obj_t * parent, int32_t num)
 
 static void update_scroll(lv_obj_t * obj)
 {
+    bool is_already_in = (bool)(uintptr_t)lv_obj_get_user_data(obj);
+    if(is_already_in) return;
+    lv_obj_set_user_data(obj, (void *)(uintptr_t)true);
+
     int32_t top_num_original = top_num;
     int32_t bottom_num_original = bottom_num;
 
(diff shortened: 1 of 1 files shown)


examples/scroll/lv_example_scroll_7.c
```diff
@@ -63,6 +67,8 @@ static void update_scroll(lv_obj_t * obj)
     if(bottom_num != bottom_num_original) {
         lv_label_set_text_fmt(low_label, "current smallest\nloaded value:\n%"PRId32, bottom_num);
     }
+
+    lv_obj_set_user_data(obj, (void *)(uintptr_t)false);
```
faxe1008: Idea: why not just store a NULL since this is what the expression expands to anyways and above store `obj` again in the user data? Reduces the amount of casting.

liamHowatt: I agree about the casting being distracting. `NULL` is a nice `false`, but I can't get on board with `obj` as the `true` without confusing readers. Check what I just pushed. This example is already using global variables so another one can be appropriate.

This example shows the working principle but I was wondering if users would benefit from something easier to copy-paste into their projects. I.e. callbacks for loading and dropping items. See the question in #7388 after it was merged.

kisvegabor (approved):
Not the most elegant thing, but let's do it this way as a hotfix :sweat_smile:

uLipe (approved):

PR #7464, opened 2024-12-16
feat(drm): add support for GBM buffer object to increase performance

etag4048:
Add the option to use MESA GBM buffers to be able to use dma-buf buffer sharing. This increases performance on platforms
where the graphics driver supports exporting dmabuf  buffers.

### DIFF
--- a/src/drivers/display/drm/lv_linux_drm.c
+++ b/src/drivers/display/drm/lv_linux_drm.c
@@ -23,6 +23,7 @@
 #include <drm_fourcc.h>
 
 #include "../../../stdlib/lv_sprintf.h"
+#include "../../../draw/lv_draw_buf.h"
 
 #if LV_LINUX_DRM_GBM_BUFFERS
 
@@ -168,7 +169,6 @@ static void drm_dmabuf_set_active_buf(lv_event_t * event)
     drm_dev_t * drm_dev;
     lv_display_t * disp;
     lv_draw_buf_t * act_buf;
-    int res;
     int i;
 
     disp = (lv_display_t *) lv_event_get_current_target(event);
@@ -190,6 +190,7 @@ static void drm_dmabuf_set_active_buf(lv_event_t * event)
 
         struct dma_buf_sync sync_req;
         sync_req.flags = DMA_BUF_SYNC_START | DMA_BUF_SYNC_RW;
+        int res;
 
         if((res = ioctl(drm_dev->act_buf->handle, DMA_BUF_IOCTL_SYNC, &sync_req)) != 0) {
             LV_LOG_ERROR("Failed to start DMA-BUF R/W SYNC res: %d", res);
--- a/src/drivers/display/drm/lv_linux_drm.c
+++ b/src/drivers/display/drm/lv_linux_drm.c
@@ -24,6 +24,14 @@
 
 #include "../../../stdlib/lv_sprintf.h"
 
+#if LV_LINUX_DRM_GBM_BUFFERS
+
+    #include <gbm.h>
+    #include <linux/dma-buf.h>
+    #include <sys/ioctl.h>
+
+#endif
+
 /*********************
  *      DEFINES
  *********************/
@@ -34,6 +42,8 @@
 #else
     #error LV_COLOR_DEPTH not supported
 #endif
+
+#define BUFFER_CNT 2
 
 /**********************
  *      TYPEDEFS
@@ -67,7 +77,8 @@ typedef struct {
     drmModePropertyPtr plane_props[128];
     drmModePropertyPtr crtc_props[128];
     drmModePropertyPtr conn_props[128];
-    drm_buffer_t drm_bufs[2]; /*DUMB buffers*/
+    drm_buffer_t drm_bufs[2];
+    drm_buffer_t * act_buf;
 } drm_dev_t;
 
 /**********************
@@ -94,12 +105,25 @@ static int drm_allocate_dumb(drm_dev_t * drm_dev, drm_buffer_t * buf);
 static int drm_setup_buffers(drm_dev_t * drm_dev);
 static void drm_flush_wait(lv_display_t * drm_dev);
 static void drm_flush(lv_display_t * disp, const lv_area_t * area, uint8_t * px_map);
+static void drm_dmabuf_set_active_buf(lv_event_t * event);
 
 static uint32_t tick_get_cb(void);
+
+#if LV_LINUX_DRM_GBM_BUFFERS
+
+    static int create_gbm_buffer(drm_dev_t * drm_dev, drm_buffer_t * buf);
+
+#endif
 
 /**********************
  *  STATIC VARIABLES
  **********************/
+
+#if LV_LINUX_DRM_GBM_BUFFERS
+
+    static struct gbm_device * gbm_device;
+
+#endif
 
 /**********************
  *      MACROS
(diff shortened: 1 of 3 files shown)


src/drivers/display/drm/lv_linux_drm.c
```diff
@@ -67,7 +77,8 @@ typedef struct {
     drmModePropertyPtr plane_props[128];
     drmModePropertyPtr crtc_props[128];
     drmModePropertyPtr conn_props[128];
-    drm_buffer_t drm_bufs[2]; /*DUMB buffers*/
+    drm_buffer_t drm_bufs[2];
```
liamHowatt:
```suggestion
    drm_buffer_t drm_bufs[BUFFER_CNT];
```

src/drivers/display/drm/lv_linux_drm.c
```diff
@@ -130,9 +154,55 @@ lv_display_t * lv_linux_drm_create(void)
...
+    drm_dev = (drm_dev_t *) lv_display_get_driver_data(disp);
+    act_buf = lv_display_get_buf_active(disp);
+
+    if(drm_dev->act_buf == NULL) {
+
+        for(i = 0; i < BUFFER_CNT; i++) {
+            if(act_buf->unaligned_data == drm_dev->drm_bufs[i].map) {
+                drm_dev->act_buf = &drm_dev->drm_bufs[i];
+                LV_LOG_TRACE("Set active buffer idx: %d", i);
+            }
```
liamHowatt:
```suggestion
                LV_LOG_TRACE("Set active buffer idx: %d", i);
                break;
            }
```

src/drivers/display/drm/lv_linux_drm.c
```diff
@@ -167,6 +237,11 @@ void lv_linux_drm_set_file(lv_display_t * disp, const char * file, int64_t conne
     lv_display_set_buffers(disp, drm_dev->drm_bufs[1].map, drm_dev->drm_bufs[0].map, buf_size,
                            LV_DISPLAY_RENDER_MODE_DIRECT);
 
+
+    /* Set the handler that is called before a redraw occurs to set the active buffer/plane
+     * when GBM buffers are used the DMA_BUF_SYNC_START is issued there */
+    lv_display_add_event_cb(disp, drm_dmabuf_set_active_buf, LV_EVENT_REFR_REQUEST, drm_dev);
```
liamHowatt: Are you sure it shouldn't be `LV_EVENT_REFR_START`?

src/drivers/display/drm/lv_linux_drm.c
```diff
@@ -130,9 +154,55 @@ lv_display_t * lv_linux_drm_create(void)
...
     lv_display_set_flush_cb(disp, drm_flush);
 
+
     return disp;
 }
 
+/* Called by LVGL when there is something that needs redrawing
+ * it sets the active buffer. if GBM buffers are used, it issues a DMA_BUF_SYNC
+ * ioctl call to lock the buffer for CPU access, the buffer is unlocked just
+ * before the atomic commit */
```
liamHowatt, line 165: > the buffer is unlocked just before the atomic commit

Is that a small race condition?

src/drivers/display/drm/lv_linux_drm.c
```diff
@@ -783,11 +884,117 @@ static int drm_allocate_dumb(drm_dev_t * drm_dev, drm_buffer_t * buf)
...
 }
 
+#if LV_LINUX_DRM_GBM_BUFFERS
+
+static int create_gbm_buffer(drm_dev_t * drm_dev, drm_buffer_t * buf)
+{
+
+    struct gbm_bo * gbm_bo;
+    int prime_fd;
+    uint32_t handles[4] = {0}, pitches[4] = {0}, offsets[4] = {0};
```
liamHowatt, line 896: Are all 4 elements of each of these arrays used?

src/drivers/display/drm/lv_linux_drm.c
```diff
@@ -783,11 +884,117 @@ static int drm_allocate_dumb(drm_dev_t * drm_dev, drm_buffer_t * buf)
...
+
+    prime_fd = gbm_bo_get_fd_for_plane(gbm_bo, 0);
+
+    if(prime_fd < 0) {
+        LV_LOG_ERROR("Failed to get prime fd for plane 0");
+        return -1;
+
+    }
+
+    buf->map = mmap(NULL, buf->size, PROT_READ | PROT_WRITE, MAP_SHARED, prime_fd, 0);
```
liamHowatt, line 947: Should it be `munmap`ped anywhere or does it live for the duration of the program?

liamHowatt (commented):
Nice! Mainly just questions.

src/drivers/display/drm/lv_linux_drm.c
kisvegabor: `LV_EVENT_REFR_REQUEST` is sent when something happened on the screen and a refresh should happen later. 

`LV_EVENT_REFR_START` is sent when the refresh is really started.

kisvegabor:
Not stale

src/drivers/display/drm/lv_linux_drm.c
ghost: The DRM function below, ``drmModeAddFB2``, expects an array. Some adapters feature multiple planes, it allows to perform hardware composition. However this driver only uses 1 plane.

src/drivers/display/drm/lv_linux_drm.c
ghost: The current implementation, doesn't close the drm device nor does it un-map the buffer before quitting - So I didn't focus on this part.

src/drivers/display/drm/lv_linux_drm.c
ghost: It depends on the implementation of the IOCTL request, I haven't checked if it can cause a race condition. I know that without it, it glitches heavily.

src/drivers/display/drm/lv_linux_drm.c
ghost: I tried it works with start too, thanks for the suggestion

etag4048:
@kisvegabor @liamHowatt Here are the fixes based on your feedback, sorry for the delay.

etag4048:
@kisvegabor gentle ping

liamHowatt (approved):
Thanks!

src/drivers/display/drm/lv_linux_drm.c
```diff
@@ -130,9 +155,56 @@ lv_display_t * lv_linux_drm_create(void)
...
+        }
+
+
+#if LV_LINUX_DRM_GBM_BUFFERS
+
+        struct dma_buf_sync sync_req;
+        sync_req.flags = DMA_BUF_SYNC_START | DMA_BUF_SYNC_RW;
+        int res;
+
+        if((res = ioctl(drm_dev->act_buf->handle, DMA_BUF_IOCTL_SYNC, &sync_req)) != 0) {
```
kisvegabor, line 195: Note that LVGL also [swaps the buffers](https://github.com/lvgl/lvgl/blob/master/src/core/lv_refr.c#L1281-L1289). In light of that is it still required?

ghost: The current implementation uses 2 separate buffers acting as frame-buffers that are set on the graphics cards. I am not sure if libdrm support only 1 buffer. Never tried such as a use case.

kisvegabor: Isn't it a standard double buffered DIRECT mode rendering? ([here](https://github.com/lvgl/lvgl/pull/7464/files#diff-f987953b6b5639df99192994498283ec69629b506874914605a14fd7afd12728R239-L168))

What I meant is LVGL takes care of swapping the buffers if two buffers are set, so this part shouldn't be required. But maybe I'm missing something.

ghost: @kisvegabor Correct, It is using direct mode and LVGL still takes care of the swapping. The block of code you commented on, retrieves the active/current buffer before it starts to get drawn to. Then in the flush_cb this buffer is committed (passed to the DRM subsystem, so that it get displayed). Then after the atomic commit, the DRM driver waits for the page flip event to arrive on the fd of the DRM device in the flush_wait_cb, once it does the next flush can happen. And the active is again retrieved from LVGL and commited..

kisvegabor: Oh, I understand now. LVGL swaps its own buffers but DRM buffers in the driver are different `struct`s and you need to keep the twp in sync. All good, approving.

kisvegabor (approved):

PR #7481, opened 2024-12-19
feat(drivers): add evdev discovery

liamHowatt:
Add support for finding evdev devices and creating indevs for them automatically. Hot-plugging is effectively made possible. Read the updated docs.

### Notes


kisvegabor:
Nice! Please update the docs as well.

liamHowatt:
cc @niklasf

src/drivers/evdev/lv_evdev.c
```diff
@@ -159,6 +207,65 @@ static void _evdev_read(lv_indev_t * indev, lv_indev_data_t * data)
...
+    lv_evdev_t * dsc = lv_indev_get_driver_data(indev);
+    LV_ASSERT_NULL(dsc);
+    close(dsc->fd);
+    lv_free(dsc);
+}
+
+#ifndef BSD
+static void _evdev_discovery_indev_try_create(const char * file_name)
+{
+    if(lv_strlen(file_name) <= 5
```
niklasf: Could use `lv_strncmp`

src/drivers/evdev/lv_evdev.c
```diff
@@ -171,10 +278,77 @@ lv_indev_t * lv_evdev_create(lv_indev_type_t indev_type, const char * dev_path)
...
+        goto err_after_open;
+    }
+    dsc->st_dev = sb.st_dev;
+    dsc->st_ino = sb.st_ino;
+
+    lv_indev_t * ex_indev = NULL;
+    while((ex_indev = lv_indev_get_next(ex_indev))) {
+        if(lv_indev_get_read_cb(ex_indev) != _evdev_read) continue;
+        lv_evdev_t * ex_dsc = lv_indev_get_driver_data(ex_indev);
+        if(dsc->st_dev == ex_dsc->st_dev && dsc->st_ino == ex_dsc->st_ino) {
```
niklasf: Not sure if it's a realistic use case: Would it be conceivable that someone might be using multiple instances of the same device, maybe with multiple displays?

niklasf:
I like the feature. Implementation looks good, too.

src/drivers/evdev/lv_evdev.c
kisvegabor: Interesting idea. I can hardly imagine it in real life, but if possible IMO we can leave this opportunity open.

kisvegabor:
not stale
cc @liamHowatt

src/drivers/evdev/lv_evdev.c
liamHowatt: I added this check to circumvent a race condition.

The way discovery starts up:

1. An inotify watcher is created for `/dev/input/` new devices in the future.
2. `/dev/input/` is iterated for already connected devices.
3. inotify events arrive to attach new devices.

The reason 1 and 2 aren't swapped is because if a device is added between those two steps, it will never be noticed by LVGL. So, the watcher is created (1) and then the directory is scanned (2).

This presents another problem, though. If a device is plugged in between 1 and 2, **it will be added twice**, so I added this inode check to make sure we don't add the same device twice.

> Would it be conceivable that someone might be using multiple instances of the same device, maybe with multiple displays?

That's a good point. I didn't think of that. Someone might want to explicitly call `lv_evdev_create` twice with the same path on purpose.

I pushed a commit just now that moves the logic into `_evdev_discovery_indev_try_create` so the check only happens for discovered devices. Let me know if it's ok.

liamHowatt:
@niklasf fyi: https://github.com/lvgl/lvgl/pull/7481#discussion_r1905813661

src/drivers/evdev/lv_evdev.c
```diff
@@ -229,9 +230,21 @@ static void _evdev_discovery_indev_try_create(const char * file_name)
     lv_indev_t * indev = lv_evdev_create(LV_INDEV_TYPE_NONE, dev_path);
     if(indev == NULL) return;
 
+    lv_evdev_t * dsc = lv_indev_get_driver_data(indev);
+
+    lv_indev_t * ex_indev = NULL;
```
kisvegabor, line 244: Please add a comment here about what this sections does, or wrap it into a function.

kisvegabor: Thanks!

kisvegabor (dismissed):
Looks, good, thanks!

uLipe (dismissed):
LGTM!

uLipe:
I think we need to wait for @niklasf review since he was participating on the review cycle before merging it.

liamHowatt:
In any case, a partner is facing an issue so I will put it back into draft until it's root-caused.

niklasf (approved):
Oh, yeah, sorry. Looks good to me. (Based on code review only. I did not actually try it.)

liamHowatt:
Ready for review again.

The problem was that the `/dev/input/` directory is not present when there are no input devices. Rather than simply being empty when the last device is removed, it is removed from the filesystem. I had not accounted for that, but I've pushed a fix and they confirmed it works.

> Oh, yeah, sorry. Looks good to me. (Based on code review only. I did not actually try it.)

Thanks, no worries. It's your feature so I just wanted you to be aware/involved.

uLipe (approved):

kisvegabor (approved):
Great! Thank you!

PR #7487, opened 2024-12-19
feat(disp): support subscription and unsubscription of vsync event

bjsylvia:

### Notes

### DIFF
--- a/src/display/lv_display.c
+++ b/src/display/lv_display.c
@@ -963,6 +963,49 @@ void lv_display_delete_refr_timer(lv_display_t * disp)
     disp->refr_timer = NULL;
 }
 
+lv_result_t lv_display_send_vsync_event(lv_display_t * disp, void * param)
+{
+    if(!disp) disp = lv_display_get_default();
+    if(!disp) return LV_RESULT_INVALID;
+
+    if(disp->vsync_count > 0)
+        return lv_display_send_event(disp, LV_EVENT_VSYNC, param);
+
+    return LV_RESULT_INVALID;
+}
+
+bool lv_display_register_vsync_event(lv_display_t * disp, lv_event_cb_t event_cb, void * user_data)
+{
+    if(!disp) disp = lv_display_get_default();
+    if(!disp) return false;
+
+    lv_display_add_event_cb(disp, event_cb, LV_EVENT_VSYNC, user_data);
+
+    /*only send once*/
+    if(disp->vsync_count == 0)
+        lv_display_send_event(disp, LV_EVENT_VSYNC_REQUEST, disp);
+
+    disp->vsync_count++;
+    return true;
+}
+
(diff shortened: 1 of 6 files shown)


kisvegabor:
As VSYNC handling is low level driver thing, I wonder what could be a practical use case for using it on a higher level outside of the driver.

If there is no such a practical use case, I think I'd make more sense to keep these local in the driver, instead of introducing a generic solution in LVGL events.

bjsylvia:
> As VSYNC handling is low level driver thing, I wonder what could be a practical use case for using it on a higher level outside of the driver.
> 
> If there is no such a practical use case, I think I'd make more sense to keep these local in the driver, instead of introducing a generic solution in LVGL events.

The original intention of this patch is to provide the ability to subscribe to vsync event related to the display in multi-instance mode. In other words, if the display is associated with a real display device, it will provide hardware vsync feedback; if the display is associated with a virtual device, it will provide virtual vsync feedback. The events between displays are not related to each other.

kisvegabor:
What is a "virtual device"? My understanding was that even in multi instance mode there is a singe real display which can handle its VSYNC and other driver related things internally.

uLipe:
@bjsylvia when you say virtual device are you mean to say something like standard virtualizaion such as virtio, and this device is something like virtio-display?

bjsylvia:
Yes, the virtual device can be virtio or other forms. A unified API can achieve application developers' transparency to this. They only need to register/unregister to achieve the effect of monitoring a specific display vsync event. The display adaptation layer will be responsible for periodically sending vsync events.

kisvegabor:
I still don't understand why would any module need to know the VSYNC event (other than the driver internally). Please mention some use cases.

uLipe:
@bjsylvia any comments?

uLipe:
Resuming this, @kisvegabor I think I can think of a potential use case.

LVGL today is mostly timed driven, which poll for a timer and consumes all events queued into its core and generate a set of draw tasks do behave according it.

But what if we could pilot the LVGL core using a single system event that does that? In that cases the event core would be less loaded and the LVGL motor would be running in sync with the display framerate allowing us make all the temporizations even more precise since they will became in sync to the display hardware.

It can be an Advanced option of usage from LVGL.

Maybe I can draw some diagrams to elucidate better what came into my mind.

uLipe:
BTW: Not stale

kisvegabor:
@uLipe 
If we want to merge it docs need to be added [here](https://github.com/lvgl/lvgl/pull/7487/files#diff-72f29d8b0e420df7d686bdd729cb761b9dfa391efc9865f0e323cfbc5234e424L111-R112) and [here](https://docs.lvgl.io/master/details/main-components/display/refreshing.html#decoupling-the-display-refresh-timer).

uLipe:
Not stale

kisvegabor (approved):

uLipe (approved):

PR #7488, opened 2024-12-19
feat(obj): support lv_obj_create_ex and lv_image_create_ex with creation information

bjsylvia:
For certain application scenarios, there is a demand for fully customized component styles, without relying on the default theme, in order to avoid performance issues introduced by secondary configuration.

### Notes

### DIFF
--- a/src/core/lv_obj.c
+++ b/src/core/lv_obj.c
@@ -210,13 +210,13 @@ const lv_obj_class_t lv_obj_class = {
  *   GLOBAL FUNCTIONS
  **********************/
 
-lv_obj_t * lv_obj_create(lv_obj_t * parent)
+lv_obj_t * lv_obj_create_ex(lv_obj_t * parent, lv_obj_create_info_t * create_info)
 {
     LV_LOG_INFO("begin");
     lv_obj_t * obj = lv_obj_class_create_obj(MY_CLASS, parent);
     LV_ASSERT_NULL(obj);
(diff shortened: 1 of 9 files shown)


kisvegabor:
The same can be achieved by calling `lv_obj_remove_style_all(obj)` after `lv_obj_create()`;

I was also thinking about introducing `lv_cont` which is a transparent widget and could be used as wrapper element.

bjsylvia:
> The same can be achieved by calling `lv_obj_remove_style_all(obj)` after `lv_obj_create()`;
> 
> I was also thinking about introducing `lv_cont` which is a transparent widget and could be used as wrapper element.

There are some use cases where native components such as buttons, text, etc. are needed, but not all default theme configurations are required. The problem with operating after `lv_obj_create` is that `lv_theme_apply` introduces a significant delay, affecting the wake-up delay in the scenario.

So the background of the patch is to avoid unnecessary configurations and optimize processing time.

kisvegabor:
I see the problem. Is this the case with many widget or only with `lv_obj`? (`lv_image` is also changed in the PR but usually it doesn't have any theme styles).

Anyway, instead of adding a new API to all widgets, what about adding a function to temporarily disable applying the theme. 
E.g. `lv_theme_disable/enable(lv_theme_t * th)`. So you can guard your calls with it.

kisvegabor:
Great! If it also works can you modify this PR?

PR #7494, opened 2024-12-20
perf(bin_decoder): improve the decoding performance of a8

W-Mai:
1. Variable `A8` should not enter cache. Entering cache in the past will cause unexpected errors when the memory address is the same.

2. Reduce decoder_data allocation and release

3. Reduce `decoded draw_buf` allocation and release

image

and it will reduce the `get_decoder_data` allocation and this draw buf allocation below

image

cc @XuNeo 

### Notes

### DIFF
--- a/src/libs/bin_decoder/lv_bin_decoder.c
+++ b/src/libs/bin_decoder/lv_bin_decoder.c
@@ -275,13 +275,20 @@ lv_result_t lv_bin_decoder_open(lv_image_decoder_t * decoder, lv_image_decoder_d
             }
         }
         else if(LV_COLOR_FORMAT_IS_ALPHA_ONLY(cf)) {
-            /*Alpha only image will need decoder data to store pointer to decoded image, to free it when decoder closes*/
-            decoder_data_t * decoder_data = get_decoder_data(dsc);
-            if(decoder_data == NULL) {
-                return LV_RESULT_INVALID;
+            if(cf == LV_COLOR_FORMAT_A8) {
+                res = LV_RESULT_OK;
+                use_directly = true;
+                dsc->decoded = (lv_draw_buf_t *)image;
             }
+            else {
+                /*Alpha only image will need decoder data to store pointer to decoded image, to free it when decoder closes*/
+                decoder_data_t * decoder_data = get_decoder_data(dsc);
+                if(decoder_data == NULL) {
+                    return LV_RESULT_INVALID;
+                }
 
-            res = decode_alpha_only(decoder, dsc);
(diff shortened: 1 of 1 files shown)


kisvegabor:
In general, why A8 is special? Why only A8 can be used directly and A4 not?

W-Mai:
> In general, why A8 is special? Why only A8 can be used directly and A4 not?

Because A1/2/4 needs bin decoder to decode them to A8 format and cached, but A8 can be draw directly. 

In fact, A8 hasn't done anything before I modified it. it just assigned it a new cache buf, then copied the old A8 image into it and returned the new A8 buf.

XuNeo:
> Entering cache in the past will cause unexpected errors when the memory address is the same.

What is the issue?

W-Mai:
> > Entering cache in the past will cause unexpected errors when the memory address is the same.
> 
> What is the issue?

I encountered a problem. When I didn't open asan and used tlsf heap to allocate mem, because there was no other (or alloc but free quickly balanced) heap allocation behavior between the two drawing characters, the assigned draw buf was still the same address when drawing the next character. This causes the cache to hit, resulting in the use of draw buf used in the previous drawing character when drawing image. As a result, it leads to drawing problems.

kisvegabor:
> Because A1/2/4 needs bin decoder to decode them to A8 format and cached, but A8 can be draw directly.

I can't fully oversee the problem, but what bothers me is that for GPUs A1/2/4/8 are the same. Maybe it's irrelevant. 


> This causes the cache to hit, resulting in the use of draw buf used in the previous drawing character when drawing image. As a result, it leads to drawing problems.

It seems like a more generic issue than only A8. (This is my last work day this year, so I cannot immerse in it now.)

W-Mai:
> > Because A1/2/4 needs bin decoder to decode them to A8 format and cached, but A8 can be draw directly.
> 
> I can't fully oversee the problem, but what bothers me is that for GPUs A1/2/4/8 are the same. Maybe it's irrelevant.

But LVGL will decoded A1/2/4 to A8 here and push the decoded A8 image into cache here:

https://github.com/lvgl/lvgl/blob/cdaaeb81dc3e1f9023749d1a79c5c7970cad10c1/src/libs/bin_decoder/lv_bin_decoder.c#L866-L894

But in fact, A8 doesn't to be decoded again, it doesn't need to run this process.

The current code's behavior with A8 is just to create a new buff and copy the original A8 data into it, which is an unnecessary process

kisvegabor (approved):
Finally understood, sorry for being so slow. I got tired by the end if the year :smile:

uLipe (approved):

PR #7504, opened 2024-12-23
perf(vg_lite): improve path append data performance

FASTSHIFT:
Use reserved space in advance + write memory directly for optimization, reduce boundary judgment and avoid using `memcpy` to copy small memory.

Can improve performance by more than 40%.

image

### Notes

### DIFF
--- a/src/draw/vg_lite/lv_draw_vg_lite_vector.c
+++ b/src/draw/vg_lite/lv_draw_vg_lite_vector.c
@@ -346,8 +346,6 @@ static void lv_path_to_vg(lv_vg_lite_path_t * dest, const lv_vector_path_t * src
         if((point)->y > max_y) max_y = (point)->y;  \
     } while(0)
 
-#define VLC_SET_OP_CODE(ptr, op) (*((uint32_t*)ptr) = (op))
-
 #define COPY_POINT_NEXT()        \
     do {                         \
         CMP_BOUNDS(point);       \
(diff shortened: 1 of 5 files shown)


src/draw/vg_lite/lv_vg_lite_utils.c
```diff
@@ -1076,7 +1076,7 @@ bool lv_vg_lite_path_check(const vg_lite_path_t * path)
         case VG_LITE_DRAW_FILL_PATH:
         case VG_LITE_DRAW_FILL_STROKE_PATH: {
                 /* Check end op code */
-                uint8_t end_op_code = VLC_GET_OP_CODE(end - fmt_len);
+                uint8_t end_op_code = LV_VG_LITE_PATH_GET_OP_CODE(end - fmt_len);
```
uLipe, line 1079: I noticed there is a subtraction to an unsigned destination value, there is no issue if fmt_len is higher than end (which leads to wrapping values)?  Or this situation is guaranteed that will never happen?

src/draw/vg_lite/lv_draw_vg_lite_vector.c
```diff
@@ -371,30 +369,30 @@ static void lv_path_to_vg(lv_vg_lite_path_t * dest, const lv_vector_path_t * src
     for(uint32_t i = 0; i < op_size; i++) {
         switch(ops[i]) {
             case LV_VECTOR_PATH_OP_MOVE_TO: {
-                    VLC_SET_OP_CODE(path_data++, VLC_OP_MOVE);
+                    LV_VG_LITE_PATH_SET_OP_CODE(path_data++, uint32_t, VLC_OP_MOVE);
```
uLipe, line 372: This is something that I get confused sometimes, the path_data next point is getting incremented after each operation added to the path buffer, but there are no bound check after every operation, is there a guarantee that the next free point in the path buffer will be always pointing of a valid bound of the path buffer?

uLipe (commented):
Nice work! I just left some questions for my understanding.

src/draw/vg_lite/lv_vg_lite_utils.c
FASTSHIFT: This won't be a problem. end is a pointer to the address after the last byte at the end of path. `fmt_len` indicates the unit length of path, which is within 4, so no errors can occur.
This code indicates that the last data of the path is fetched. The last data of the closed path should be `VLC_OP_END`.

src/draw/vg_lite/lv_draw_vg_lite_vector.c
FASTSHIFT: No boundary check required. Because the previous `lv_vg_lite_path_reserve_space` ensures that enough space is allocated ahead of time for safe writing, reducing repeated boundary checks can significantly improve efficiency, especially if the path data is long.
In addition, I also call `LV_ASSERT_MSG` after the loop to check whether the converted path matches the calculated path exactly, and the program will report an error if it does not match.

uLipe: Perfect, thank you for the clarification @FASTSHIFT

uLipe (approved):

XuNeo (approved):

PR #7509, opened 2024-12-24
feat(draw): Enable TSC Compression scheme for frame buffer in Nema acceleration

MarkIoanAmat:
## TSC Compression for FB
While using NemaGFX to accelerate LVGL, we have the option to use ThinkSilicon's proprietary TSC compression algorithms to encode/decode the frame buffer in favor of memory consumption.
### Steps to Enable
 - Enable TSC Frame buffer type. e.g `lv_display_set_color_format(display, LV_COLOR_FORMAT_NEMA_TSC12A);`
 - Bypass `lv_draw_sw_blend `function with the use of the define below inside in `lv_conf.h` 
         `#define LV_DRAW_BLEND_CUSTOM LV_DRAW_BLEND_CUSTOM_NEMA_GFX`
### Features not supported while using TSC compression for frame buffer
 - LV_COLOR_FORMAT_ARGB8565 
 - LV_COLOR_FORMAT_RGB565A8 
 - Blending mode LV_BLEND_MODE_SUBTRACTIVE


MarkIoanAmat:
@kisvegabor @liamHowatt

src/draw/nema_gfx/lv_draw_nema_gfx_blend.c
```diff
@@ -0,0 +1,260 @@
...
+
+void lv_draw_sw_blend(lv_draw_unit_t * draw_unit, const lv_draw_sw_blend_dsc_t * blend_dsc)
+{
+    /*Do not draw transparent things*/
+    if(blend_dsc->opa <= LV_OPA_MIN) return;
+    if(blend_dsc->mask_buf && blend_dsc->mask_res == LV_DRAW_SW_MASK_RES_TRANSP) return;
+
+    lv_area_t blend_area;
+    if(!lv_area_intersect(&blend_area, blend_dsc->blend_area, draw_unit->clip_area)) return;
+
```
uLipe, line 61: I like this initial checking, but I'd like to see an extra information to aid people to track when the things go wrong here, simply putting log-err or log-warning would be helpful here.

uLipe (commented):
First pass, I have a small comment.

kisvegabor:
Let's consider https://github.com/lvgl/lvgl/pull/7531 too, as it aims to solve the same issue.

MarkIoanAmat:
@kisvegabor 
I had a quick look at #7531 to my understanding these 2 solutions are complementary as they solve different problems. PR #7509 aims to by pass the whole lv_draw_sw_blend function and use a custom one. On the other one PR #7531 allow registering different blending handlers inside lv_draw_sw_blend. Also PR #7509 is adding a special feature which requires the presence of NemaGPU Plus the TSC Compression scheme for FB and it is working under specific constrictions. I can not think of a way merging those two solutions without polluting the common code, which is not preferred of course. 
Waiting for your feedback. Thank you.

kisvegabor:
It seems to me in https://github.com/lvgl/lvgl/pull/7531 we can register a blend callbacks for TSC color formats without any issues. However, we should really add the `LV_COLOR_FORMAT_NEMA_TSCxxx` color formats.

MarkIoanAmat:
@kisvegabor for sure we have to add the new color formats. Regarding the custom handlers proposed in #7531, they aim to replace/add to the specific blending functions like `lv_draw_sw_blend_image_to_rgb565` etc. On the other hand my PR aims to change the whole structure of lv_draw_sw_blend, doing changes long before the point where the custom handlers will be called. For example, we cannot use code like ```
fill_dsc.dest_buf = lv_draw_layer_go_to_xy(layer, blend_area.x1 - layer->buf_area.x1,
                                                   blend_area.y1 - layer->buf_area.y1);
```
on a TSC Frame buffer. The other way could be creating a custom handler regarding blending to a TSC FB, but we would have to pollute the public code of lv_draw_sw_blend with code specific for TSC which of course is not acceptable and sustainable for future releases.

**Comment by MarkIoanAmat** [comment]
@kisvegabor for sure we have to add the new color formats. Regarding the custom handlers proposed in #7531 they aim to replace/add to the specific blending functions like `lv_draw_sw_blend_image_to_rgb565 `etc. 
On the other hand my PR aims to change the whole structure of _lv_draw_sw_blend_, doing changes long before the point where the custom handlers will be called. For example, we cannot use code like the bellow on a TSC Frame buffer. 

`fill_dsc.dest_buf = lv_draw_layer_go_to_xy(layer, blend_area.x1 - layer->buf_area.x1,blend_area.y1 - layer->buf_area.y1);`

(We cannot move the pointer inside a TSC encoded memory area)

The other way could be creating a custom handler regarding blending to a TSC FB, but we would have to pollute the public code of _lv_draw_sw_blend_ with code specific for TSC which of course is not acceptable and sustainable for future releases.

**Comment by MarkIoanAmat** [comment]
@kisvegabor My proposal would be to keep both solutions as they are referring to different things

File: lv_conf_template.h [hunk 2]
```diff
@@ -137,6 +137,12 @@
  * - Rendering engine needs to support 3x3 matrix transformations. */
 #define LV_DRAW_TRANSFORM_USE_MATRIX            0
 
+/** Selectively disable lv_draw_sw_blend function and use a custom one. Possible options:
+ * - LV_DRAW_BLEND_CUSTOM_NONE
+ * - LV_DRAW_BLEND_CUSTOM_NEMA_GFX */
+
+#define LV_DRAW_BLEND_CUSTOM LV_DRAW_BLEND_CUSTOM_NONE
```
kisvegabor, line 144: As it's realted to SW render please move it to the other LV_DRAW_SW_ config option and call it `LV_DRAW_SW_BLEND_CUSTOM`

scripts/lv_conf_internal_gen.py
```diff
@@ -41,6 +41,9 @@
 #define LV_OS_SDL2          7
 #define LV_OS_CUSTOM        255
 
+#define LV_DRAW_BLEND_CUSTOM_NONE 0
+#define LV_DRAW_BLEND_CUSTOM_NEMA_GFX 1
```
kisvegabor, line 61:
```suggestion
#define LV_DRAW_SW_BLEND_CUSTOM_NONE 0
#define LV_DRAW_SW_BLEND_CUSTOM_NEMA_GFX 1
```

src/draw/nema_gfx/lv_draw_nema_gfx_blend.c
kisvegabor: It's not an error, just an early return if trying to draw outside of the current buffer area. It can happen if it wasn't filtered earlier.

kisvegabor (commented):
I understand now! 

> We cannot move the pointer inside a TSC encoded memory area

It really happens earlier and we should be able to overwrite it too. 


Ideally it'd be better make it also dynamic to allow having e.g. a Nema TSC frame buffer and an external display controller (ILI9341 with normal RGB565 color format) What do you think about it?

kisvegabor:
@MarkIoanAmat could you rebase it on #7531?

MarkIoanAmat:
@kisvegabor  Will do. Thank you.

PR #7531, opened 2024-12-31
feat(draw/sw): allow custom handlers

faxe1008:
This allows LVGL users to register custom software draw handlers. It is beneficial to be able to add blending support for more unusual color formats that are unlikely to be included in the upstream repository (e.g., e-Ink displays with 7-indexed color formats).
It also allows users to register custom handlers if the behavior of the existing ones regarding performance or rules (e.g., for I1, when is a pixel considered to be on?) does not meet their needs.

### Notes

### DIFF
--- a/lv_conf_template.h
+++ b/lv_conf_template.h
@@ -215,6 +215,7 @@
 
     /** Enable drawing complex gradients in software: linear at an angle, radial or conical */
     #define LV_USE_DRAW_SW_COMPLEX_GRADIENTS    0
+
 #endif
 
 /*Use TSi's aka (Think Silicon) NemaGFX */
--- a/src/draw/sw/lv_draw_sw.c
+++ b/src/draw/sw/lv_draw_sw.c
@@ -92,6 +92,8 @@ void lv_draw_sw_init(void)
 #if LV_USE_VECTOR_GRAPHIC && LV_USE_THORVG
     tvg_engine_init(TVG_ENGINE_SW, 0);
 #endif
+
+    lv_ll_init(&LV_GLOBAL_DEFAULT()->draw_sw_blend_handler_ll, sizeof(lv_draw_sw_custom_blend_handler_t));
 }
 
 void lv_draw_sw_deinit(void)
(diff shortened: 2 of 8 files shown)


uLipe (changes requested):
I like the idea, the only point that I concerns me is about the configuration name, so instead of : `LV_DRAW_SW_CUSTOM_HANDLERS` , why not something like: `LV_DRAW_SW_CUSTOM_BLEND_HANDLERS` ?

I understand the name is large , but the first option imply you are enabling an option that allows to register any kind of custom handlers, since in the reality you are just enabling a feature that allows to override the blend handler. 

Other than that, nice work!

faxe1008:
True, the name you suggested is better. Adjusted accordingly :^)

kisvegabor:
What a coincidence! https://github.com/lvgl/lvgl/pull/7509 aims to solve the same issue. This PR is more flexible as it allows registering multiple handlers at runtime, but the solution in https://github.com/lvgl/lvgl/pull/7509 is much simpler. 

cc @MarkIoanAmat

kisvegabor:
Not stale, let's keep it going. Sorry for my absence.

kisvegabor:
I was talking with @MarkIoanAmat about this topic and we concluded that your version should be used with a little modification. 
Instead of adding new `blend_color` and `blend_image` functions the whole `lv_draw_sw_blend`. It is because [this part](https://github.com/lvgl/lvgl/blob/master/src/draw/sw/blend/lv_draw_sw_blend.c#L83-L85) can be used only with standard color formats. E.g. it won't know how to go to x=10, y=22 on an indexed 3 bit colored frame buffer. However, if call the user's function at the very beginning the user can de any magic with exotic color formats. 

Do you agree with this direction? If you are you interested in updating the PR accoridngly?

kisvegabor:
I've checked our schedule , however we very busy with the preparation to Embedded World @liamHowatt can work on in only at the beginning of March. If someone is interested in picking it up, it'd be more than welcome

faxe1008:
Quite busy as well but I will see if i can manage something in next two weeks :^)

faxe1008:
@kisvegabor pushed a change with your suggested changes :^)

src/draw/sw/lv_draw_sw.c
```diff
@@ -340,4 +344,59 @@ static void execute_drawing(lv_draw_sw_unit_t * u)
...
+            lv_ll_remove(&LV_GLOBAL_DEFAULT()->draw_sw_blend_handler_ll, handler);
+            lv_free(handler);
+            return true;
+        }
+    }
+
+    return false;
+}
+
+lv_draw_sw_blend_handler_t lv_draw_sw_get_blend_handler(lv_color_format_t cf)
```
kisvegabor:
```suggestion
lv_draw_sw_blend_handler_t lv_draw_sw_get_blend_handler(lv_color_format_t dest_cf)
```

kisvegabor (commented):
Thank you Fabian! I'm really thankful! 

It looks good to me, just one minor suggestion. 

cc @MarkIoanAmat

src/draw/sw/lv_draw_sw.c
faxe1008: Done :^)

lv_conf_template.h
```diff
@@ -215,6 +215,9 @@
 
     /** Enable drawing complex gradients in software: linear at an angle, radial or conical */
     #define LV_USE_DRAW_SW_COMPLEX_GRADIENTS    0
+
+    /* Enable registering of custom SW draw handlers */
+    #define LV_DRAW_SW_CUSTOM_BLEND_HANDLERS  0
```
kisvegabor: IMO this config is not required as enabling this feature by default doesn't make a big difference.

kisvegabor (commented):
Sorry, just a last one.

lv_conf_template.h
faxe1008: I added it as a config because imo. especially when working on resource constraint devices you should only pay for what you use (flash/ram). Maybe its kind of zephyr-ish in it's approach but I feel like generally the lower the footprint, the better. On the flip side this makes testing more difficult of course, if every feature has its own toggle :^)

kisvegabor: Yeah, it's really a trade off. On the flip side (besides the more complex testing) I also see the issue of more config options. In this case the extra memory usage is only a few bytes which is less of an issue compared to the testing and config complexity increase. 

So unless there is a principal objection, please remove this config.

kisvegabor (approved):
Looks good, thank you!

uLipe (approved):

PR #7542, opened 2025-01-04
fix(label): limit self size by max_height style

niklasf:
Small part of #7108 that is relevant without the full feature: Let labels respect the `max_height` property.


liamHowatt (approved):
Makes sense to me. I assume if the max height style is changed, the `LV_EVENT_STYLE_CHANGED` is handled and `invalid_size_cache` is set to `true` in `lv_label_refr_text` so the cached size is recalculated.

uLipe (approved):

PR #7548, opened 2025-01-06
fix(widgets): fix an incorrect text replacement

DDcrow:
There is an incorrect text replacing that replaces "lv_obj_event_base" with "LV_EVENT_base" in widget template, this commit will reverse this replacement.


uLipe (approved):

kisvegabor (approved):
Thanks!

Issue #7552, opened 2025-01-06
LVGL9.2.2 lv_file_explorer.c has Minor bugs that affect functionality.


lifeng5478:
### LVGL version

V9.2.2

### Platform

![1736161076424](https://github.com/user-attachments/assets/eebbff2d-d3d2-4027-9cb4-ffc72159b88d)
this! After the folder is entered, it cannot be re-entered. I fixed Like This! For your information!

### What happened?

When the file manager enters the secondary directory, it reverts again and cannot be entered again.

### How to reproduce?

static void browser_file_event_handler(lv_event_t * e)
{
    lv_event_code_t code = lv_event_get_code(e);
    lv_obj_t * obj = lv_event_get_user_data(e);
    lv_file_explorer_t * explorer = (lv_file_explorer_t *)obj;

    if(code == LV_EVENT_VALUE_CHANGED) {
        char file_name[LV_FILE_EXPLORER_PATH_MAX_LEN];
        const char * str_fn = NULL;
        uint32_t row;
        uint32_t col;

        lv_memzero(file_name, sizeof(file_name));
        lv_table_get_selected_cell(explorer->file_table, &row, &col);
        str_fn = lv_table_get_cell_value(explorer->file_table, row, col);

        str_fn = str_fn + 5;
        if((lv_strcmp(str_fn, ".") == 0))  return;

        if((lv_strcmp(str_fn, "..") == 0) && (lv_strlen(explorer->current_path) > 3)) {
            strip_ext(explorer->current_path);
            /*Remove the last '/' character*/
            strip_ext(explorer->current_path);
            lv_snprintf((char *)file_name, sizeof(file_name), "%s", explorer->current_path);

        }
        else {
            **_if(lv_strcmp(str_fn, "..") != 0) {
                if(explorer->current_path[lv_strlen(explorer->current_path) - 1] != '/') {
                    strip_ext(explorer->current_path);
                    strip_ext(explorer->current_path);

                    lv_snprintf((char *)file_name, sizeof(file_name), "%s/%s", explorer->current_path, str_fn);
                }
                else {
                lv_snprintf((char *)file_name, sizeof(file_name), "%s%s", explorer->current_path, str_fn);
                }
            }_**
        }

        lv_fs_dir_t dir;
        if(lv_fs_dir_open(&dir, file_name) == LV_FS_RES_OK) {
            lv_fs_dir_close(&dir);
            show_dir(obj, (char *)file_name);
        }
        else {
[...]

PR #7558, opened 2025-01-07
fix(file_explorer): leave and re-enter directory

lifeng5478:
lv_file_explorer.c has Minor bugs that affect functionality.

### DIFF
--- a/src/others/file_explorer/lv_file_explorer.c
+++ b/src/others/file_explorer/lv_file_explorer.c
@@ -478,7 +478,6 @@ static void browser_file_event_handler(lv_event_t * e)
 {
     lv_event_code_t code = lv_event_get_code(e);
     lv_obj_t * obj = lv_event_get_user_data(e);
-
     lv_file_explorer_t * explorer = (lv_file_explorer_t *)obj;
 
     if(code == LV_EVENT_VALUE_CHANGED) {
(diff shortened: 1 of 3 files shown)


src/others/file_explorer/lv_file_explorer.c
```diff
@@ -603,10 +607,13 @@ static void show_dir(lv_obj_t * obj, const char * path)
     lv_strlcpy(explorer->current_path, path, sizeof(explorer->current_path));
     lv_label_set_text_fmt(explorer->path_label, LV_SYMBOL_EYE_OPEN" %s", path);
 
+    /*
     size_t current_path_len = lv_strlen(explorer->current_path);
     if((*((explorer->current_path) + current_path_len) != '/') && (current_path_len < LV_FILE_EXPLORER_PATH_MAX_LEN)) {
         *((explorer->current_path) + current_path_len) = '/';
     }
+    */
```
liamHowatt, line 615: This can be removed.

liamHowatt (commented):
Fixes #7552

Use the astyle from here: https://github.com/lvgl/astyle and run `python3 scripts/code-format.py`

This was already fixed in `master`. @C47D would it be enough to cherry-pick #7552, do you think? It's tricky because so many fixes have been made for file_explorer in master, including with lv_table, and a lv_file_explorer test case. cc @lhdjply

I'm inclined to just work with this specific fix if it works, but I wouldn't want to make `release/v9.2` less stable with new code that has not been broadly validated.

lhdjply:
@kisvegabor @liamHowatt Will V9.2.3 be released? If not, this fix doesn't make sense without a release.

PR #7568, opened 2025-01-08
feat(obj): add dragable flag

### LINKED ISSUE #7505: lv_obj_set_drag()
### Introduce the problem

There is no similar interface available

### Proposal

_No response_

lhdjply:
Fixes #7505 

### Notes


liamHowatt (approved):
Looks good, if we all agree that it's a feature that should be added.

cc @XuNeo to double check the obj property updates.

I noticed these things but maybe they're expected, and up to the user to deal with.
- draggable objects with children will scroll the children when dragged (fine?)
- labels are not draggable (fine?)

lhdjply:
> * draggable objects with children will scroll the children when dragged (fine?)
> * labels are not draggable (fine?)

1. If the parent object is dragged, the child object should definitely move together with the parent object. So I think this is fine.
2. For labels, you need to add `lv_obj_add_flag(label, LV_OBJ_FLAG_CLICKABLE);` to make them draggable.

liamHowatt:
> If the parent object is dragged, the child object should definitely move together with the parent object. So I think this is fine.

In general I agree but did you know this is what happens? See this video:

[Screencast from 01-09-2025 09:22:59 PM.webm](https://github.com/user-attachments/assets/5302af6a-eceb-4995-942d-31df205567f4)

IAMMX:
> In general I agree but did you know this is what happens? See this video:

This mainly involves both dragging and sliding actions taking effect at the same time, which indeed makes it hard to judge.

lhdjply:
This mainly involves both dragging and sliding actions taking effect at the same time, which indeed makes it hard to judge.

liamHowatt:
Okay. I've already approved.

AndreCostaaa:
> > If the parent object is dragged, the child object should definitely move together with the parent object. So I think this is fine.
> 
> In general I agree but did you know this is what happens? See this video:
> Screencast.from.01-09-2025.09.22.59.PM.webm

We could check that the parent is being dragged and block the children from getting scrolled. Not sure if it could break something else or if it's hard/impossible to do

kisvegabor:
I've removed this feature in v8 because I haven't seen much real life use cases for it. For the few use cases adding the simple event from the example is better than having it in LVGL as  built-in feature. What do you think?

lhdjply:
> I've removed this feature in v8 because I haven't seen much real life use cases for it. For the few use cases adding the simple event from the example is better than having it in LVGL as  built-in feature. What do you think? 

So, this PR is not recommended？

PR #7572, opened 2025-01-08
fix(wayland): delete the allocated display and buffers

ThomasDevoogdt:
The `lv_wayland_window_create` calls `lv_{display,draw_buf}_create`, so `lv_wayland_window_close` should call at some point `lv_{draw_buf_destroy,display_delete}`.


liamHowatt (approved):
Thanks. cc @etag4048

uLipe (approved):

PR #7573, opened 2025-01-08
fix(memcpy): ensure volatile qualifier for destination pointer in lv_…

### LINKED ISSUE #7406: Optimization Issue in lv_memcpy Due to Macro Usage
### LVGL version

v9.2,2

### Platform

STM32H755 Platform
GCC (-O3)
-ffunction-sections
-fdata-sections

### What happened?

The current implementation of lv_memcpy uses macros like _COPY and _REPEAT8, which can lead to undefined or unintended behavior under high compiler optimization levels  (-O3). This results in hard faults in some cases. 

Here’s the refactored version of the function using inline functions:

```
#include <stdint.h>
#include <stddef.h>

#define _ALIGN_MASK (sizeof(uint32_t) - 1)

/* Inline function to copy a single byte */
static inline void copy_byte(volatile uint8_t **dst, const uint8_t **src) {
    **dst = **src;
    (*dst)++;
    (*src)++;
}

/* Inline function to copy a single 32-bit word */
static inline void copy_word(volatile uint32_t **dst, const uint32_t **src) {
    **dst = **src;
    (*dst)++;
    (*src)++;
}

/* Inline function for repeated byte copying */
static inline void repeat_byte_copy(volatile uint8_t **dst, const uint8_t **src, size_t times) {
    for (size_t i = 0; i < times; i++) {
        copy_byte(dst, src);
    }
}

/* Inline function for repeated word copying */
static inline void repeat_word_copy(volatile uint32_t **dst, const uint32_t **src, size_t times) {
    for (size_t i = 0; i < times; i++) {
        copy_word(dst, src);
    }
}

void lv_memcpy(void *dst, const void *src, size_t len) {
    volatile uint8_t *d8 = (volatile uint8_t *)dst;
    const uint8_t *s8 = (const uint8_t *)src;

    /* Simplify for small memories */
    if (len < 16) {
        while (len) {
            copy_byte(&d8, &s8);
            len--;
        }
        return;
    }

    uintptr_t d_align = (uintptr_t)d8 & _ALIGN_MASK;
    uintptr_t s_align = (uintptr_t)s8 & _ALIGN_MASK;

    /* Byte copy for unaligned memories */
    if (s_align != d_align) {
        while (len > 32) {
            repeat_byte_copy(&d8, &s8, 32);
            len -= 32;
        }
        while (len) {
            copy_byte(&d8, &s8);
            len--;
        }
        return;
    }

    /* Align the memories */
    if (d_align) {
        d_align = _ALIGN_MASK + 1 - d_align;
        while (d_align && len) {
            copy_byte(&d8, &s8);
            d_align--;
            len--;
        }
    }

    volatile uint32_t *d32 = (volatile uint32_t *)d8;
    const uint32_t *s32 = (const uint32_t *)s8;

    while (len > 32) {
        repeat_word_copy(&d32, &s32, 8);
        len -= 32;
    }

    d8 = (volatile uint8_t *)d32;
    s8 = (const uint8_t *)s32;
    while (len) {
        copy_byte(&d8, &s8);
        len--;
    }
}
```

### How to reproduce?

_No response_

1saeed:
…memcpy

Using the volatile qualifier prevents the compiler from optimizing out memory operations, which is critical when dealing with unaligned addresses or memory-mapped hardware regions like external SDRAM.
Fixes: [lvgl#7406](https://github.com/lvgl/lvgl/issues/7406)

### Notes


kisvegabor:
Thank you! Would you mind adding a test case [here](https://github.com/lvgl/lvgl/blob/master/tests/src/test_cases/test_mem.c) for unaligened memcpy.

1saeed:
> Thank you! Would you mind adding a test case [here](https://github.com/lvgl/lvgl/blob/master/tests/src/test_cases/test_mem.c) for unaligened `memcpy?

@kisvegabor 
I would but I tried many times and realized that it's not possible to create a fail test for unaligned memory access that works reliably on all platforms. Different hardware and architectures handle unaligned access in various ways. For example, ARM Cortex-M processors may handle unaligned access in internal SRAM by splitting it into smaller aligned accesses. On the other hand, external SDRAM with stricter memory rules might raise errors, return wrong data, or behave unpredictably.

Compilers and optimizations can also hide unaligned access problems, making them harder to detect. Additionally, features like the UNALIGN_TRP bit in ARM processors, which can catch unaligned accesses, are not available on all platforms. These differences make it impossible to create a single test that works everywhere

liamHowatt:
kisvegabor, fyi unaligned accesses are handled gracefully on x86 which you may already know.

liamHowatt:
This may be the answer we're looking for:

https://gcc.gnu.org/onlinedocs/gcc/Instrumentation-Options.html#index-fsanitize_003dalignment

> -fsanitize=alignment
> 
> This option enables checking of alignment of pointers when they are dereferenced, or when a reference is bound to insufficiently aligned target, or when a method or constructor is invoked on insufficiently aligned object.

tests/src/test_cases/test_mem.c
```diff
@@ -24,4 +24,28 @@ void test_mem_buf_realloc(void)
...
+
+    /* Use unaligned addresses by offsetting by 1 */
+    uint8_t *unaligned_source = source + 1;
+    uint8_t *unaligned_destination = destination + 1;
+
+    /* Perform memcpy */
+    lv_memcpy(unaligned_destination, unaligned_source, 18);
+
+    /* Verify that the copied values match the source */
+    for (int i = 0; i < 18; i++) {
```
liamHowatt: Should it be `19`?

tests/src/test_cases/test_mem.c
```diff
@@ -24,4 +24,28 @@ void test_mem_buf_realloc(void)
...
+    uint8_t destination[20];
+
+    /* Initialize source with some values */
+    for (int i = 0; i < 20; i++) {
+        source[i] = (uint8_t)(i + 1);
+    }
+
+    /* Use unaligned addresses by offsetting by 1 */
+    uint8_t *unaligned_source = source + 1;
+    uint8_t *unaligned_destination = destination + 1;
```
liamHowatt: What if `source` is placed at `0x0000ffff` on the stack? Adding 1 would align it, instead of unalign it. I recommend defining `source` and `destination` as:

```c
union {
    uint8_t u8[20];
    uint32_t dummy_u32;
} source;
...
uint8_t *unaligned_source = source.u8 + 1;
```

to ensure 32 bit alignment of the array so the `+ 1` is guaranteed unaligned.

1saeed: @liamHowatt Thanks! Your recommendation is excellent and it's a robust way.

liamHowatt (approved):
To fix formatting next time, install this: https://github.com/lvgl/astyle and run `python3 scripts/code-format.py`. Or copy the diff from the CI failure output.

uLipe (approved):

Issue #5396, opened 2024-01-19
Run the CI tests in more configuratios


kisvegabor:
### Introduce the problem

We should add more config option for the CI, especially for `LV_BIN_DECODER_RAM_LOAD` and parallel rendering. 

### Proposal

@XuNeo

XuNeo:
Need also include NEON acceleration to CI test. See https://github.com/lvgl/lvgl/pull/5526.

kisvegabor:
Also test with various cache settings.

kisvegabor:
Test OBJ_ID too. #5544

XuNeo:
Current CI has enabled them in lv_test_conf_full.h
```
#define LV_USE_OBJ_ID           1
#define LV_USE_OBJ_ID_BUILTIN   1
```

kisvegabor:
Test with `LV_DRAW_SW_COMPLEX 0`

XuNeo:
Not stale.

kisvegabor:
Add NuttX build to test the CI. See https://github.com/lvgl/lvgl/issues/6797

kisvegabor:
This one is also related: https://github.com/lvgl/lvgl/issues/3127

kisvegabor:
Cover this one too: https://github.com/lvgl/lvgl/issues/5152

AndreCostaaa:
After taking a look at the current Build CI i've noticed that not all test options are currently being run

https://github.com/lvgl/lvgl/blob/25232d92e7d2526dc2504b2fc0455cd2d17c6aa8/tests/CMakeLists.txt#L111-L153

https://github.com/lvgl/lvgl/blob/25232d92e7d2526dc2504b2fc0455cd2d17c6aa8/.github/workflows/ccpp.yml#L17-L21

Is there a specific reason behind this ? @kisvegabor

kisvegabor:
To be honest I don't why `OPTIONS_NORMAL_8BIT` is missing. That's why we need to review these sooner than later :slightly_smiling_face:

kisvegabor:
Let's get back to this after v9.4

AndreCostaaa:
With the recent additions of CI tests I would say we can close this issue

Issue #7223, opened 2024-11-05
Rework and unify CMake


kisvegabor:
### Introduce the problem

Now our CMake "policy" is not clear. The way we link 3rd party lib or enable/disable demos and examples is messy. We need a unified way of handling all these.

### Proposal

After having a discussion with @etag4048 and @glebm I believe we should do this:
- CMake should use `lv_conf.h` or `Kconfig` configs
- From these the enabled options should be extracted
- CMake should compile and link accordingly. 

What do you think?

glebm:
That makes sense.

kisvegabor:
@etag4048 will be back next week and can look into it.

uLipe:
What about let cmake to handle only the source tree that is platform agnostic of the lvgl and register it in the package registry from cmake.

Zephyr and NuttX do that and this facilitates a lot the creation of out of tree projects that can just use the cmake's find package that creates lvgl as an archive file add to the target list then the platform specific project just links it against the final binary.

kisvegabor:
I'm open to any ideas/solution. My goal is to create something flexible and stable that can be used for a few years.

However note that it's not only about 3rd party libs, but internal LVGL configs too. E.g. if `LV_BUILD_EXAMPLES == 0` we can skip the building and linking of the examples. Same for each demo.

tdjastrzebski:
I am not a CMake expert but I used to build LVGL 8.x with my own build tool (cppbuild) and it worked just great. No issues at all.
Now LVGL 9.x got more sophisticated. `thorvag/rapidjson` library was added which contains two files (`stdint.h`, `inttypes.h`) causing conflicts. Includes like `#include LV_DRAW_DMA2D_HAL_INCLUDE` are used which my tool cannot resolve.
All in all, it might be easier to go with the mainstream and use CMake. LVGL contains (intends to contain?) a support document dedicated to CMake config (https://docs.lvgl.io/integration/build/cmake) but this page appears to be missing, 404 error is returned. Is it because it is in works?

kisvegabor:
@tdjastrzebski 
> here are only three above issues which, when fixed manually allow my tool to work correctly again

Do you remember what issues did you have? Could you add a build log?

@uLipe 
> One thing that is also could be useful here is to pick the lv_conf.h and generate a lvconf.cmake that can be included into the main cmake file allowing for example to select which directories should be considered during build system generation.

Sounds good to me!

tdjastrzebski:
[...]

PR #7574, opened 2025-01-08
ci: add 8bit build option

AndreCostaaa:
Related issue: #5396

### DIFF
--- a/.github/workflows/ccpp.yml
+++ b/.github/workflows/ccpp.yml
@@ -11,6 +11,7 @@ jobs:
     if: ${{ github.event_name != 'pull_request' || github.repository != github.event.pull_request.head.repo.full_name }}
     runs-on: ubuntu-22.04
     strategy:
+      fail-fast: false
       matrix:
         # A valid option parameter to the cmake file.
         # See BUILD_OPTIONS in tests/CMakeLists.txt.
@@ -31,6 +32,7 @@ jobs:
 
   build-windows:
     strategy:
+      fail-fast: false
       matrix:
         # A valid option parameter to the cmake file.
         # See BUILD_OPTIONS in tests/CMakeLists.txt.
--- a/.github/workflows/ccpp.yml
+++ b/.github/workflows/ccpp.yml
@@ -65,7 +65,7 @@ jobs:
     - name: Install prerequisites
       run: scripts\install-prerequisites.bat
     - name: Build
-      if: matrix.build_option != OPTIONS_NORMAL_8BIT
+      if: matrix.build_option != 'OPTIONS_NORMAL_8BIT'
       run: python tests/main.py --build-option=${{ matrix.build_option }} build
       env:
         CC: ${{ matrix.compiler }}
--- a/.github/workflows/ccpp.yml
+++ b/.github/workflows/ccpp.yml
@@ -14,7 +14,8 @@ jobs:
       matrix:
         # A valid option parameter to the cmake file.
         # See BUILD_OPTIONS in tests/CMakeLists.txt.
-        build_option: ['OPTIONS_16BIT',
+        build_option: ['OPTIONS_NORMAL_8BIT',
+                       'OPTIONS_16BIT',
                        'OPTIONS_24BIT',
                        'OPTIONS_FULL_32BIT',
                        'OPTIONS_VG_LITE',
--- a/.github/workflows/ccpp.yml
+++ b/.github/workflows/ccpp.yml
@@ -36,8 +36,7 @@ jobs:
       matrix:
         # A valid option parameter to the cmake file.
         # See BUILD_OPTIONS in tests/CMakeLists.txt.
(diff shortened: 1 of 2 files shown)


AndreCostaaa:
I guess this answers https://github.com/lvgl/lvgl/issues/5396#issuecomment-2575080065 @kisvegabor 

Would it be okay to add an exception to windows runs? It could be useful to make sure it still builds in others platforms

uLipe:
I'm wondering if we could override to a supported color depth in CI scenarios for windows, @kisvegabor

AndreCostaaa:
> I'm wondering if we could override to a supported color depth in CI scenarios for windows, @kisvegabor

We can just remove that build option. What do you think ?

uLipe:
> We can just remove that build option. What do you think ?

For me it seems okay, @kisvegabor ?

.github/workflows/ccpp.yml
```diff
@@ -7,19 +7,21 @@ on:
     branches: [ master, release/v8.* ]
 
 jobs:
-  build:
+  build-ubuntu:
     if: ${{ github.event_name != 'pull_request' || github.repository != github.event.pull_request.head.repo.full_name }}
     runs-on: ubuntu-22.04
     strategy:
+      fail-fast: false
```
liamHowatt, line 14: Why not? Just curious

.github/workflows/ccpp.yml
```diff
@@ -28,36 +30,45 @@ jobs:
...
     - uses: actions/checkout@v4
     - name: Install prerequisites
       run: scripts\install-prerequisites.bat
-    - uses: ilammy/msvc-dev-cmd@v1
+
+    - if: matrix.compiler == 'gcc'
+      uses: ammaraskar/gcc-problem-matcher@master
+
+    - if: matrix.compiler == 'gcc'
+      uses: ammaraskar/gcc-problem-matcher@master
```
liamHowatt: Accidental duplicate

liamHowatt (commented):
We disable other things for Windows so it's ok to do, in my opinion.

We could probably build Windows in 8 bit if we just disabled `lv_windows` for the 8 bit build. I don't see an easy way though. If we could just have `if(!8bit)` here: https://github.com/lvgl/lvgl/blob/62291fa1744811a0ad6969a65d1f426387719914/tests/CMakeLists.txt#L419-L420 but it's not that easy.

.github/workflows/ccpp.yml
AndreCostaaa: Currently, the developer never gets to know if the new commit broke a specific build option or all of them. Since build options aren't dependent of each other, I believe a failure in the 8 bit option shouldn't cancel testing the 32 bit option for instance

.github/workflows/ccpp.yml
AndreCostaaa: Oh, thank you

AndreCostaaa:
> We could probably build Windows in 8 bit if we just disabled lv_windows for the 8 bit build. I don't see an easy way though.

What would be the difference between building windows in 8 bit while disabling `lv_windows` or not building windows in 8 bit at all?

liamHowatt (approved):
Sorry for the delay.

> What would be the difference between building windows in 8 bit while disabling `lv_windows` or not building windows in 8 bit at all?

Some 8bit-only piece of code might raise a warning/error only with MSCV but not Ubuntu GCC. We could build all of LVGL except `lv_windows`. It's an extreme edge case though :slightly_smiling_face:

AndreCostaaa:
> Some 8bit-only piece of code might raise a warning/error only with MSCV but not Ubuntu GCC. We could build all of LVGL except `lv_windows`. It's an extreme edge case though 🙂

I think doing that could eventually be done when #7223 eventually gets fixed

If we could simply toggle the compilation of a certain file based on the config it would probably be very easy to create a config that would simply not try to build `lv_window_context.c` for that specific use-case

uLipe (approved):

liamHowatt:
@kisvegabor final question before merging. Are we ok to bring up the number of CI jobs from 24 to 30?

AndreCostaaa:
Just to clarify as that increase may be surprising - The increase is due to creating a separate Windows job for each configuration, as was already done for Ubuntu. Previously, two Windows jobs tested all configurations; now, there are five jobs, each testing a single configuration.

This change allows the 8-bit build option to be cleanly disabled on Windows without modifying the `tests/main.py` script. It also makes the workflow file more readable by explicitly showing which configurations are being tested.

FASTSHIFT (approved):

liamHowatt (approved):
Let's do it

kisvegabor:
Sorry for missing it so long. 

I need to find a better strategy to prioritize GitHub issues/PRs. @uLipe, @liamHowatt please write me on Slack if I'm missing something for this long.

PR #7579, opened 2025-01-09
fix(os): add support for thread names

### LINKED ISSUE #7217: Set different names for threads created by different units
### Introduce the problem

The OS abstraction layer defines and uses a single name "lvglDraw" for all LVGL threads when LV_USE_OS is enabled.

Because a single thread name is used this can lead to confusion when identifying running threads. For example, during FreeRTOS profiling I've found that identifying SW and VGlite threads was confusing enough that I've made a patch to address it in my setup. This is resolved by adding a 'thread_name' argument to `lv_thread_init`.

Is this something you also find useful? I know there are a bunch of OSs that don't use thread names and there are also written applications that may need updating if the lv_thread_init signature is changed. If it shows interest, I'll submit a PR for review.

Thanks

### Proposal

```
-lv_result_t lv_thread_init(lv_thread_t * thread, lv_thread_prio_t prio, void (*callback)(void *), size_t stack_size,
+lv_result_t lv_thread_init(lv_thread_t * thread, const char * const pcName,
+		lv_thread_prio_t prio, void (*callback)(void *), size_t stack_size,
                            void * user_data);
```

cristian-stoica:
This patch allows different draw units to name their own threads differently so they can be distinguished during debugging.

Fixes https://github.com/lvgl/lvgl/issues/7217

### Notes

### DIFF
--- a/env_support/cmsis-pack/lv_os_custom_c.txt
+++ b/env_support/cmsis-pack/lv_os_custom_c.txt
@@ -36,10 +36,11 @@
  *   GLOBAL FUNCTIONS
  **********************/
 
-lv_result_t lv_thread_init(lv_thread_t * thread, lv_thread_prio_t prio, void (*callback)(void *), size_t stack_size,
+lv_result_t lv_thread_init(lv_thread_t * thread, const char * const name, lv_thread_prio_t prio, void (*callback)(void *), size_t stack_size,
                            void * user_data)
 {
     LV_UNUSED(thread);
+    LV_UNUSED(name);
     LV_UNUSED(callback);
     LV_UNUSED(prio);
     LV_UNUSED(stack_size);
--- a/examples/porting/osal/lv_example_osal.c
+++ b/examples/porting/osal/lv_example_osal.c
@@ -51,7 +51,8 @@ void lv_example_osal(void)
         LV_LOG_ERROR("Error initializing thread sync");
     }
 
-    if(lv_thread_init(&increment_thread, LV_THREAD_PRIO_MID, increment_thread_entry, 2048, NULL) != LV_RESULT_OK) {
+    if(lv_thread_init(&increment_thread, "inc_th", LV_THREAD_PRIO_MID, increment_thread_entry, 2048,
+                      NULL) != LV_RESULT_OK) {
         LV_LOG_ERROR("Error initializing thread");
     }
 }
--- a/src/draw/nxp/pxp/lv_draw_pxp.c
+++ b/src/draw/nxp/pxp/lv_draw_pxp.c
@@ -95,7 +95,7 @@ void lv_draw_pxp_init(void)
     draw_pxp_unit->base_unit.name = "NXP_PXP";
 
 #if LV_USE_PXP_DRAW_THREAD
-    lv_thread_init(&draw_pxp_unit->thread, LV_THREAD_PRIO_HIGH, _pxp_render_thread_cb, 2 * 1024, draw_pxp_unit);
+    lv_thread_init(&draw_pxp_unit->thread, "pxpdraw", LV_THREAD_PRIO_HIGH, _pxp_render_thread_cb, 2 * 1024, draw_pxp_unit);
 #endif
 #endif /*LV_USE_DRAW_PXP*/
 }
--- a/src/draw/nxp/vglite/lv_draw_vglite.c
+++ b/src/draw/nxp/vglite/lv_draw_vglite.c
@@ -118,7 +118,8 @@ void lv_draw_vglite_init(void)
     draw_vglite_unit->base_unit.name = "NXP_VGLITE";
 
 #if LV_USE_VGLITE_DRAW_THREAD
-    lv_thread_init(&draw_vglite_unit->thread, LV_THREAD_PRIO_HIGH, _vglite_render_thread_cb, 2 * 1024, draw_vglite_unit);
+    lv_thread_init(&draw_vglite_unit->thread, "vglitedraw", LV_THREAD_PRIO_HIGHEST, _vglite_render_thread_cb, 2 * 1024,
+                   draw_vglite_unit);
 #endif
 }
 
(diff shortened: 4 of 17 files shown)


env_support/cmsis-pack/lv_os_custom_c.txt
```diff
@@ -36,10 +36,11 @@
  *   GLOBAL FUNCTIONS
  **********************/
 
-lv_result_t lv_thread_init(lv_thread_t * thread, lv_thread_prio_t prio, void (*callback)(void *), size_t stack_size,
+lv_result_t lv_thread_init(lv_thread_t * thread, const char * const pcName, lv_thread_prio_t prio, void (*callback)(void *), size_t stack_size,
                            void * user_data)
 {
     LV_UNUSED(thread);
+    LV_UNUSED(pcName);
```
uLipe: Nit: pcName does not seem to follow the snake case standard even though this file is a custom one.

examples/porting/osal/lv_example_osal.c
```diff
@@ -51,7 +51,8 @@ void lv_example_osal(void)
         LV_LOG_ERROR("Error initializing thread sync");
     }
 
-    if(lv_thread_init(&increment_thread, LV_THREAD_PRIO_MID, increment_thread_entry, 2048, NULL) != LV_RESULT_OK) {
+    if(lv_thread_init(&increment_thread, "inc_th", LV_THREAD_PRIO_MID, increment_thread_entry, 2048,
```
uLipe, line 54: I would suggest a more expressive name, in a OS environment short names like that may collide with other threads, for example: `lv_inc_thread` or `lv_increment_th` tells more about the thread.

src/draw/nxp/pxp/lv_draw_pxp.c
```diff
@@ -95,7 +95,7 @@ void lv_draw_pxp_init(void)
     draw_pxp_unit->base_unit.name = "NXP_PXP";
 
 #if LV_USE_PXP_DRAW_THREAD
-    lv_thread_init(&draw_pxp_unit->thread, LV_THREAD_PRIO_HIGH, _pxp_render_thread_cb, 2 * 1024, draw_pxp_unit);
+    lv_thread_init(&draw_pxp_unit->thread, "pxp_th", LV_THREAD_PRIO_HIGH, _pxp_render_thread_cb, 2 * 1024, draw_pxp_unit);
```
uLipe: same here

src/draw/nxp/vglite/lv_draw_vglite.c
```diff
@@ -118,7 +118,8 @@ void lv_draw_vglite_init(void)
     draw_vglite_unit->base_unit.name = "NXP_VGLITE";
 
 #if LV_USE_VGLITE_DRAW_THREAD
-    lv_thread_init(&draw_vglite_unit->thread, LV_THREAD_PRIO_HIGH, _vglite_render_thread_cb, 2 * 1024, draw_vglite_unit);
+    lv_thread_init(&draw_vglite_unit->thread, "vglite_th", LV_THREAD_PRIO_HIGHEST, _vglite_render_thread_cb, 2 * 1024,
```
uLipe: although here seems good, on a vglite MPU environment we will probably other threads with vglite naming, maybe put the LV prefix would be valid as well.

src/draw/renesas/dave2d/lv_draw_dave2d.c
```diff
@@ -99,7 +99,8 @@ void lv_draw_dave2d_init(void)
     lv_ll_init(&_ll_Dave2D_Tasks, 4);
 
 #if LV_USE_OS
-    lv_thread_init(&draw_dave2d_unit->thread, LV_THREAD_PRIO_HIGH, _dave2d_render_thread_cb, 8 * 1024, draw_dave2d_unit);
+    lv_thread_init(&draw_dave2d_unit->thread, "dave2d_th", LV_THREAD_PRIO_HIGH, _dave2d_render_thread_cb, 8 * 1024,
```
uLipe: same for above

src/draw/sw/lv_draw_sw.c
```diff
@@ -85,7 +85,8 @@ void lv_draw_sw_init(void)
         draw_sw_unit->base_unit.name = "SW";
 
 #if LV_USE_OS
-        lv_thread_init(&draw_sw_unit->thread, LV_THREAD_PRIO_HIGH, render_thread_cb, LV_DRAW_THREAD_STACK_SIZE, draw_sw_unit);
+        lv_thread_init(&draw_sw_unit->thread, "lvglDraw", LV_THREAD_PRIO_HIGH, render_thread_cb, LV_DRAW_THREAD_STACK_SIZE,
```
uLipe: Nit: name of the all other threads follow snake case convention, I think should too.

uLipe (changes requested):
Nice work! I have some small nits, could you check them please?

cristian-stoica:
> Nice work! I have some small nits, could you check them please?

I'm using FreeRTOS conventions for thread name variable - hence "pcName'. If we change the case to follow LVGL convention we might as well use an LVGL(ish) variable name if you have a suggestion.

Adding "lv_" prefix to names requires a name array length above 10 which is FreeRTOS default. I don't know about other OS' but I would not make this change a requirement.

uLipe:
About the snake case convention, I think we can just use `name` because the method is self explanatory.

> Adding "lv_" prefix to names requires a name array length above 10 which is FreeRTOS default. I don't know about other OS' but I would not make this change a requirement. The alternative is of course to come up with shorter names - any ideas?
> 
> For thread names, I've reused only "lvglDraw" name which did not follow snake convention originally - that can be fixed.
One suggestion here is just avoid prefixes at all, we know that all of the named things are actually threads, so we can just the name of the subject of that thread with some sort of abbreviation and keep the name into 10-sized string since the FreeRTOS are the current constraint we have so possible names:

`swdraw` -> Software Render Thread
`vglitedraw ` -> VGLite Render Thread
`pxpdraw` -> pxp based draw thread

Other OS like Zephyr and QNX use this approach by just setting the subject of that thread without prefixes.

What do you think @cristian-stoica ?

uLipe (approved):
Again, Nice work!

FASTSHIFT (approved):

kisvegabor:
E.g. `nemagfx` and `dma2d` could have been `dma2d_draw` and `nemagfx_draw` for consistency with `pxpdraw` and `vglitedraw`, but that's not a big deal. 

Thank you for the update!

PR #7582, opened 2025-01-09
feat(scripts/LVGLImage.py): adds RGB565 dithering support

kdschlosser:
The title says it all.

### Notes

### DIFF
--- a/scripts/LVGLImage.py
+++ b/scripts/LVGLImage.py
@@ -497,6 +497,7 @@ class LVGLImage:
                  data: bytes = b'') -> None:
         self.stride = 0  # default no valid stride value
         self.premultiplied = False
+        self.rgb565_dither = False
         self.set_data(cf, w, h, data)
 
     def __repr__(self) -> str:
@@ -571,6 +572,9 @@ class LVGLImage:
 
         self.stride = stride
         self.data = bytearray(b''.join(data_out))
+
+    def set_rgb565_dither(self, value):
+        self.rgb565_dither = value
 
     def premultiply(self):
         """
@@ -996,6 +1000,7 @@ class LVGLImage:
                 color |= (g >> 2) << 5
                 color |= (b >> 3) << 0
                 return uint16_t(color)
+
         elif cf == ColorFormat.RGB565A8:
 
             def pack(r, g, b, a):
@@ -1017,20 +1022,80 @@ class LVGLImage:
         w, h, rows, _ = reader.asRGBA8()
         rawdata = bytearray()
         alpha = bytearray()
-        for row in rows:
+        for y, row in enumerate(rows):
             R = row[0::4]
             G = row[1::4]
             B = row[2::4]
             A = row[3::4]
-            for r, g, b, a in zip(R, G, B, A):
+            for x, (r, g, b, a) in enumerate(zip(R, G, B, A)):
                 if cf == ColorFormat.RGB565A8:
                     alpha += uint8_t(a)
+
+                if (
+                    self.rgb565_dither and
+                    cf in (ColorFormat.RGB565, ColorFormat.RGB565A8, ColorFormat.ARGB8565)
+                ):
+                    r, g, b = dither(x, y, r, g, b)
+
                 rawdata += pack(r, g, b, a)
 
         if cf == ColorFormat.RGB565A8:
             rawdata += alpha
 
         self.set_data(cf, w, h, rawdata)
+
+
+red_thresh = [
+  1, 7, 3, 5, 0, 8, 2, 6,
+  7, 1, 5, 3, 8, 0, 6, 2,
+  3, 5, 0, 8, 2, 6, 1, 7,
+  5, 3, 8, 0, 6, 2, 7, 1,
+  0, 8, 2, 6, 1, 7, 3, 5,
+  8, 0, 6, 2, 7, 1, 5, 3,
+  2, 6, 1, 7, 3, 5, 0, 8,
+  6, 2, 7, 1, 5, 3, 8, 0
+]
+
+green_thresh = [
+  1, 3, 2, 2, 3, 1, 2, 2,
+  2, 2, 0, 4, 2, 2, 4, 0,
+  3, 1, 2, 2, 1, 3, 2, 2,
+  2, 2, 4, 0, 2, 2, 0, 4,
+  1, 3, 2, 2, 3, 1, 2, 2,
+  2, 2, 0, 4, 2, 2, 4, 0,
+  3, 1, 2, 2, 1, 3, 2, 2,
+  2, 2, 4, 0, 2, 2, 0, 4
+]
+
+blue_thresh = [
+  5, 3, 8, 0, 6, 2, 7, 1,
+  3, 5, 0, 8, 2, 6, 1, 7,
+  8, 0, 6, 2, 7, 1, 5, 3,
+  0, 8, 2, 6, 1, 7, 3, 5,
+  6, 2, 7, 1, 5, 3, 8, 0,
+  2, 6, 1, 7, 3, 5, 0, 8,
+  7, 1, 5, 3, 8, 0, 6, 2,
+  1, 7, 3, 5, 0, 8, 2, 6
+]
+
+
+# Get 16bit closest color
+def closest_rb(c):
+    return c >> 3 << 3
+
+
+def closest_g(c):
+    return c >> 2 << 2
+
+
+def dither(x, y, r, g, b):
+    tresshold_id = ((y & 7) << 3) + (x & 7)
+
+    return (
+        closest_rb(min(r + red_thresh[tresshold_id], 0xff)) & 0b11111000,
+        closest_g(min(g + green_thresh[tresshold_id], 0xff)) & 0b11111100,
+        closest_rb(min(b + blue_thresh[tresshold_id], 0xff)) & 0b11111000
+    )
 
 
 class RLEHeader:
@@ -1212,7 +1277,8 @@ class PNGConverter:
                  align: int = 1,
                  premultiply: bool = False,
                  compress: CompressMethod = CompressMethod.NONE,
-                 keep_folder=True) -> None:
+                 keep_folder=True,
+                 rgb565_dither=False) -> None:
         self.files = files
         self.cf = cf
         self.ofmt = ofmt
@@ -1223,6 +1289,7 @@ class PNGConverter:
         self.premultiply = premultiply
         self.compress = compress
         self.background = background
+        self.rgb565_dither = rgb565_dither
 
     def _replace_ext(self, input, ext):
         if self.keep_folder:
@@ -1243,6 +1310,8 @@ class PNGConverter:
             else:
                 img = LVGLImage().from_png(f, self.cf, background=self.background)
                 img.adjust_stride(align=self.align)
+                img.set_rgb565_dither(self.rgb565_dither)
+
                 if self.premultiply:
                     img.premultiply()
                 output.append((f, img))
@@ -1274,6 +1343,9 @@ def main():
             "XRGB8888", "RGB565", "RGB565A8", "ARGB8565", "RGB888", "AUTO",
             "RAW", "RAW_ALPHA"
         ])
+
+    parser.add_argument('--rgb565dither', action='store_true',
+                        help="use dithering to correct banding in gradients", default=False)
 
     parser.add_argument('--premultiply', action='store_true',
                         help="pre-multiply color with alpha", default=False)
@@ -1334,7 +1406,8 @@ def main():
                              align=args.align,
                              premultiply=args.premultiply,
                              compress=compress,
-                             keep_folder=False)
+                             keep_folder=False,
+                             rgb565_dither=args.rgb565dither)
     output = converter.convert()
     for f, img in output:
         logging.info(f"len: {img.data_len} for {path.basename(f)} ")
--- a/scripts/LVGLImage.py
+++ b/scripts/LVGLImage.py
@@ -1079,22 +1079,13 @@ blue_thresh = [
 ]
 
 
-# Get 16bit closest color
-def closest_rb(c):
-    return c >> 3 << 3
-
-
-def closest_g(c):
-    return c >> 2 << 2
-
-
 def dither(x, y, r, g, b):
     treshold_id = ((y & 7) << 3) + (x & 7)
 
     return (
-        closest_rb(min(r + red_thresh[treshold_id], 0xff)) & 0b11111000,
-        closest_g(min(g + green_thresh[treshold_id], 0xff)) & 0b11111100,
-        closest_rb(min(b + blue_thresh[treshold_id], 0xff)) & 0b11111000
+        (min(r + red_thresh[treshold_id], 0xFF) >> 3) << 3,
+        (min(g + green_thresh[treshold_id], 0xFF) >> 2) << 2,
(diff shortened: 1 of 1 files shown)


scripts/LVGLImage.py
```diff
@@ -1033,6 +1045,59 @@ def pack(r, g, b, a):
...
+def closest_rb(c):
+    return c >> 3 << 3
+
+
+def closest_g(c):
+    return c >> 2 << 2
+
+
+def dither(x, y, r, g, b):
+    tresshold_id = ((y & 7) << 3) + (x & 7)
```
uLipe: Nit: typo `thesshold_id`

uLipe (commented):
Looks good, I have a small nit.

scripts/LVGLImage.py
kdschlosser: all fixed.

liamHowatt (approved):
Awesome! I imagine if more dithering color formats are added `--rgb565dither` might become an automatic subset of some future `--dither` flag.

Out of curiosity, what kind of dithering is this? Is it Floyd-Steinberg?

Original
image

Dithered
image

Dithered zoomed in
image

scripts/LVGLImage.py
```diff
@@ -1033,6 +1049,41 @@ def pack(r, g, b, a):
...
+  8, 0, 6, 2, 7, 1, 5, 3,
+  0, 8, 2, 6, 1, 7, 3, 5,
+  6, 2, 7, 1, 5, 3, 8, 0,
+  2, 6, 1, 7, 3, 5, 0, 8,
+  7, 1, 5, 3, 8, 0, 6, 2,
+  1, 7, 3, 5, 0, 8, 2, 6
+]
+
+
+
```
XuNeo: Please remove one extra line. It's not documented, but the file is originally formatted using yapf.

XuNeo (commented):
Do we have other better algorithm?
For example, the pngquant output looks even better, using 8-bit.
```
pngquant test.png
```

kdschlosser:
It's actually not working for some reason. I don't know the code in that file or what the best way is to get the variable set so it knows to do the dithering.

This is what it should look like..

Original Image
image


RGB565 Without dithering
image


RGB565 with dithering
image

liamHowatt:
Indeed, the image I sent did not in fact get dithered :laughing:. That code is not reached in the debugger.

liamHowatt (changes requested):
(see previous comment)

kdschlosser:
That dithering algorithm is the fastest one I have ever seen. I don't remember where I came across the algorithm.

kdschlosser:
The thing is I don't know why it is not reached. If I eliminate the if statement it works fine. I am not sure why the rgb565 flag is not getting set. I don't know the mechanics of what is being called and when. It might be easier if the person that wrote the script edited it and moved things where they need to be moved to in order to get it to work,...

scripts/LVGLImage.py
```diff
@@ -1243,6 +1296,8 @@ def convert(self):
             else:
                 img = LVGLImage().from_png(f, self.cf, background=self.background)
                 img.adjust_stride(align=self.align)
+                img.set_rgb565_dither(self.rgb565_dither)
```
XuNeo: The PNG is converted to LVGL image in method `from_png`. So the dither parameter should pass directly to it.

```suggestion
                img = LVGLImage().from_png(f, self.cf, background=self.background, rgb565dither= self.rgb565_dither)
                img.adjust_stride(align=self.align)
```

kdschlosser: OK cool I will make the changes... I will also make the code I added PEP8 compliant.

scripts/LVGLImage.py
```diff
@@ -1037,10 +1036,10 @@ def pack(r, g, b, a):
                 ):
                     treshold_id = ((y & 7) << 3) + (x & 7)
 
-                    r = (min(r + red_thresh[treshold_id], 0xFF) >> 3) << 3
-                    g = (min(g + green_thresh[treshold_id], 0xFF) >> 2) << 2
-                    b = (min(b + blue_thresh[treshold_id], 0xFF) >> 3) << 3
-                   
+                    r = min(r + red_thresh[treshold_id], 0xFF) & 0xF8
```
XuNeo, line 1039: It seems there's no need to `& 0xF8` etc. manually. The pack method will only keep 5,6,5 bits anyway.

kdschlosser: It is better to leave it there for purpose of knowing what is actually taking place and that only the higher 5, 6m, and 5 bits of the RGB is what is used.. I do know that the pack if going to bit shift the bottom 3, 2 and 3 bits away anyhow. For the purposes of clarity in the code I do recommend leaving it there. The bit operation is cheap to perform and doesn't cause any kind king of an appreciable difference  in how long it takes to convert the image..

If anything I actually think that perhaps what needs to be done is removing the rgb565 requirement and making it a general flag that will work across all color formats as I am sure it would help with any color anomalies that occur when lowering the color depth in general and not for rgb565 alone.

XuNeo: That's fine to me. Thank you.

XuNeo (approved):

kdschlosser:
@XuNeo 

Question for ya...

Have you seen the Python Imaging Library (pillow, aka PIL)? It would be a replacement for using pyPNG and PngQuant. PIL has built in handling of indexed images. It also has these codecs that are built into it. 

* BLP (Blizzard Mipmap Format)
* BMP
* BUFR
* CUR (Windows Cursor)
* DCX (Indel DCX)
* DDS (DirectDraw Surface)
* EPS (Encapsulated Postscript)
* FITS
* FLI (Autodesk FLI/FLC Animation)
* FPX (FlashPix)
* FTEX (Texture File Format)
* GBR (GIMP brush file)
* GD (GD uncompressed images)
* GIF (Compuserve)
* GRIB
* HDF5
* ICNS (Mac OS icns resource)
* ICO (Windows Icon)
* IM (IFUNC Image Memory)
* IMT (IM Tools)
* IPTC
* JPEG2000
* JPEG
* MCIDAS (McIdas area file)
* MIC (Microsoft Image Composer)
* MPEG
* MPO
* MSP (Windows Paint)
* Palm
* PCD (Kodak PhotoCD)
* PCX (Paintbrush)
* PDF
* PIXAR (PIXAR raster image)
* PNG
* PNGA
* PPM (Pbmplus image)
* PSD (Adobe Photoshop)
* QOI (Quite OK Image)
* SGI
* SPIDER
* SUN (Sun Raster File)
* TGA (Targa)
* TIFF (Adobe TIFF)
* WAL (Quake2 Texture)
* WEBP
* WMF (Windows Metafile)
* XBM (X11 Bitmap)
* XPM (X11 Pixel Map)
* XVThumb (XV thumbnail image)

PIL is extensible and is easy to add a codec to it.  A codec for LVGL "BIN" format would be easy to add. If a codec was written then you can convert to and from from any of the above image formats. This would also include reading and writing LVGL image C source files as well. Could even add one that reads and writes LVGL MicroPython py files if ya wanted to.

Here are the docs on how to add an image plugin.
https://pillow.readthedocs.io/en/stable/handbook/writing-your-own-image-plugin.html

There are 3rd party codec plugins that have been made for PIL as well. AVIF, HEIF, JPEG-LS, SVG, JPEG-XL, ASTC, PVRTC, etc... Just to name a few...


Thought it would be worth mentioning instead of limiting the the image converter script to only PNG and only being able to deal with indexed images from a machine running a posix OS. PIL is cross platform so it would run on Posix like systems including macOS and also Windows. It doesn't have any external dependencies so no additional packages or shared libraries need to be installed to be able to use it.

XuNeo:
I initially used Pillow but switched to pngquant for its superior quality when processing Indexed8 images. 

Since Indexed8 (1 byte/pixel) is widely favored for its small size, `pngquant` was chosen as the final tool.

All other image formats can be firstly converted to PNG and then to LVGL's bin file.

`icu` image converter does support a few more image formats https://github.com/W-Mai/icu

scripts/LVGLImage.py
kdschlosser: Out of question how come the file isn't to PEP8 standards? That is the de-facto standard used for CPython source files.

The space still needed to be removed because it is not PEP8 compliant either. But the rest of the file has all kinds of formatting issues if it is to be in PEP8 compliance.

If you unzip the attached file and edit the index.html file and replace all instances of `file:///LVGLImage.py` with `file:///{absolute_path_to}/LVGLImage.py` where `{absolute_path_to}` is the absolute path to `LVGLImage.py` and then open up the index.html file in your browser you will see all of the code inspection errors/warnings. There are 167 of them. 

[code_inspection.zip](https://github.com/user-attachments/files/18449546/code_inspection.zip)

you can also see the results in the attached text file

[inspection_results.txt](https://github.com/user-attachments/files/18449877/inspection_results.txt)

XuNeo: Feel free to reformat the script to follow standards. Document which standard is used in beginning of the script so others can follow it too.

XuNeo: Oh, one more thing, please do remember to use a second commit for formatting, so it's not mixed with the commit for dithering support.

kdschlosser: I didn't change any of the original codes formatting. If you are OK with that being done I will be more than happy to and I will submit a separate PR for that.

liamHowatt (approved):
Very nice

img_demo_high_res_smart_home_widget2_bg_lg

kdschlosser:
I did come up with ma nice way of handling  the use of the dithering if it is wanting to be added to the gradients.

```c

uint8_t *red_thresh = NULL;
uint8_t *green_thresh = NULL;
uint8_t *blue_thresh = NULL;


bool rgb565_dither_init(void)
{
    if (red_thresh == NULL) {
        red_thresh = (uint8_t *)malloc(64);
        if (red_thresh != NULL) {
            memcpy(red_thresh,
                (uint8_t []){
                1, 7, 3, 5, 0, 8, 2, 6,
                7, 1, 5, 3, 8, 0, 6, 2,
                3, 5, 0, 8, 2, 6, 1, 7,
                5, 3, 8, 0, 6, 2, 7, 1,
                0, 8, 2, 6, 1, 7, 3, 5,
                8, 0, 6, 2, 7, 1, 5, 3,
                2, 6, 1, 7, 3, 5, 0, 8,
                6, 2, 7, 1, 5, 3, 8, 0
            }, 64);
        }
    }

    if (green_thresh == NULL) {
        green_thresh = (uint8_t *)malloc(64);
        if (green_thresh != NULL) {
            memcpy(green_thresh,
                (uint8_t []){
                1, 3, 2, 2, 3, 1, 2, 2,
                2, 2, 0, 4, 2, 2, 4, 0,
                3, 1, 2, 2, 1, 3, 2, 2,
                2, 2, 4, 0, 2, 2, 0, 4,
                1, 3, 2, 2, 3, 1, 2, 2,
                2, 2, 0, 4, 2, 2, 4, 0,
                3, 1, 2, 2, 1, 3, 2, 2,
                2, 2, 4, 0, 2, 2, 0, 4
            }, 64);
        }
    }

    if (blue_thresh == NULL) {
        blue_thresh = (uint8_t *)malloc(64);
        if (blue_thresh != NULL) {
            memcpy(blue_thresh,
                (uint8_t []){
                5, 3, 8, 0, 6, 2, 7, 1,
                3, 5, 0, 8, 2, 6, 1, 7,
                8, 0, 6, 2, 7, 1, 5, 3,
                0, 8, 2, 6, 1, 7, 3, 5,
                6, 2, 7, 1, 5, 3, 8, 0,
                2, 6, 1, 7, 3, 5, 0, 8,
                7, 1, 5, 3, 8, 0, 6, 2,
                1, 7, 3, 5, 0, 8, 2, 6
            }, 64);
        }
    }

    if (red_thresh == NULL || blue_thresh == NULL || green_thresh == NULL) return false;
    else return true;
}
```

This will allow for the lowest amount of memory use if the feature is not used without making it a compile time option. Once a gradient is rendered the arrays can be freed from memory if resources are tight or it can be left for repeated use for better performance is resources are not an issue.

having the ability to use the dithering at runtime is an important thing to have because some gradients the banding is not as noticeable as others so being able to selectively pick what gradients to use it on is important. being able to turn it "on" and "off" at runtime also makes it possible to save some RAM if a gradient is being rendered to a buffer only one time and that buffer is being used in place of having to render the gradient over and over again. The threshold arrays can be freed since they are no longer being used.

This is what I did for the header file for the dithering..


```c
#ifndef __RGB565_DITHER_H__
    #define __RGB565_DITHER_H__

    #define CALC_THRESHOLD(x, y)  (uint8_t)((y & 7) << 3) + (x & 7)

    extern uint8_t *red_thresh;
    extern uint8_t *green_thresh;
    extern uint8_t *blue_thresh;

    bool rgb565_dither_init(void);

    static inline void rgb565_dither_pixel(uint8_t treshold_id, uint16_t *pixel)
    {
        *pixel = (((((*pixel >> 8) & 0xF8) + red_thresh[treshold_id]) << 8) |
                  ((((*pixel >> 3) & 0xFC) + green_thresh[treshold_id]) << 3) |
                  ((((*pixel & 0x1F) << 3) + blue_thresh[treshold_id]) >> 3));
    }

#endif
```

The dither function is an inline function and the arrays are extern. The threshold calculation is done using a macro. 

The code above is written for the in place modification of a buffer. So the pixel is able to be written and the buffer then passed to this function for modification. Or it can be adjusted so that it will work with an lc_color_t structure and the alteration can be made prior to the color being written to the buffer,

kdschlosser:
@liamHowatt 

can you do me a favor and post the original PNG you used for that gradient you posted. There is something funky going on with the image converterm that is causing the dithering to come out all screwed up and I am wondering if it is pyPNG doing it.

I created a test gradient using the colors at opposite ends of the gradient that you posted and using the RGB565 dithering with the python imaging library instead of pyPNG this is what it comes out as...


image


That is wayyy different looking than what you posted. Then again I am also starting off with RGB565 colors and that could skew the results. This is why I would like to have the original image you used.

kdschlosser:
When zoomed into the image you posted this is what is seen...

image


and when zoomed into the same exact area for the image I posted this is what is seen.

image


The pattern seen in the coloring is completely different. So something is altering it.

liamHowatt:
> can you do me a favor and post the original PNG you used for that gradient you posted.

Here:

img_demo_high_res_smart_home_widget2_bg_lg

kdschlosser:
something is without a doubt altering the pixel data. This is an RGB565 dithered output using PIL and the dithering.

image

It is very different than your output

405017261-53db6f6b-4d16-41b8-9ba7-ac9b2e3d40e2

I don't know what is altering the pixel data in that way..

what are the script command you used when you converted the image?

liamHowatt:
```
python3 scripts/LVGLImage.py --rgb565dither --ofmt PNG --cf ARGB8565 '/home/liam/Downloads/Demos(29)/img_demo_high_res_smart_home_widget2_bg_lg.png'
```

`pypng` in `pip freeze` is `pypng==0.20220715.0`

kdschlosser:
I am not sure what is happening in the conversion but something is making a mess of the colors. I tested just the RGB565 dithering and it's 100% not being caused by that... I suspect it is being caused by pyPNG's encoding. Have you tested to see what the BIN output looks like on a display? That is what the script is really intended for is converting from PNG to LVGL BIN and not PNG to PNG.

PR #7583, opened 2025-01-09
feat(anim): add a pause method

### LINKED ISSUE #7029: lv anim support pause/resume
### Problem to solve

anim requires pause/resume in some cases to keep continuity

### Success criteria

pause the anim, sleep several seconds, the resume it started with the same anim value.

### Solution outline

add new control API, like lv_anim_pause/resume

### Rabbit holes

No

### Testing

pause the anim, sleep several seconds, the resume it started with the same anim value.

### Teaching

Doc the new API

### Considerations

_No response_

AndreCostaaa:
Closes #7029 

This PR adds the ability to pause and resume animations

### DIFF
--- a/src/misc/lv_anim.c
+++ b/src/misc/lv_anim.c
@@ -44,6 +44,7 @@ static void anim_mark_list_change(void);
 static void anim_completed_handler(lv_anim_t * a);
 static int32_t lv_anim_path_cubic_bezier(const lv_anim_t * a, int32_t x1,
                                          int32_t y1, int32_t x2, int32_t y2);
+static void lv_anim_pause_for_internal(lv_anim_t * a, uint32_t ms);
 static void resolve_time(lv_anim_t * a);
 static bool remove_concurrent_anims(lv_anim_t * a_current);
 static void remove_anim(void * a);
@@ -502,24 +503,25 @@ uint32_t lv_anim_resolve_speed(uint32_t speed_or_time, int32_t start, int32_t en
 
 bool lv_anim_is_paused(lv_anim_t * a)
 {
+    LV_ASSERT_NULL(a);
     return a->is_paused;
 }
 
 void lv_anim_pause(lv_anim_t * a)
 {
-    lv_anim_pause_for(a, LV_ANIM_PAUSE_FOREVER);
+    LV_ASSERT_NULL(a);
+    lv_anim_pause_for_internal(a, LV_ANIM_PAUSE_FOREVER);
 }
 
 void lv_anim_pause_for(lv_anim_t * a, uint32_t ms)
 {
-    a->is_paused = true;
-    a->pause_time = lv_tick_get();
-    a->pause_duration = ms;
+    LV_ASSERT_NULL(a);
+    lv_anim_pause_for_internal(a, ms);
 }
 
 void lv_anim_resume(lv_anim_t * a)
 {
-
+    LV_ASSERT_NULL(a);
     a->is_paused = false;
     a->pause_duration = 0;
 }
@@ -692,6 +694,13 @@ static int32_t lv_anim_path_cubic_bezier(const lv_anim_t * a, int32_t x1, int32_
     new_value += a->start_value;
 
     return new_value;
+}
+
+static void lv_anim_pause_for_internal(lv_anim_t * a, uint32_t ms){
+
+    a->is_paused = true;
+    a->pause_time = lv_tick_get();
+    a->pause_duration = ms;
 }
 
 static void resolve_time(lv_anim_t * a)
--- a/src/misc/lv_anim.c
+++ b/src/misc/lv_anim.c
@@ -501,26 +501,30 @@ uint32_t lv_anim_resolve_speed(uint32_t speed_or_time, int32_t start, int32_t en
     return LV_CLAMP(min_time * 10, time, max_time * 10);
 }
 
-bool lv_anim_is_paused(lv_anim_t * a){
+bool lv_anim_is_paused(lv_anim_t * a)
+{
     return a->is_paused;
 }
 
-void lv_anim_pause(lv_anim_t * a){
+void lv_anim_pause(lv_anim_t * a)
+{
     lv_anim_pause_for(a, LV_ANIM_PAUSE_FOREVER);
 }
 
-void lv_anim_pause_for(lv_anim_t * a, uint32_t ms){
+void lv_anim_pause_for(lv_anim_t * a, uint32_t ms)
+{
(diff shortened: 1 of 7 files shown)


src/misc/lv_anim.c
```diff
@@ -499,6 +500,30 @@ uint32_t lv_anim_resolve_speed(uint32_t speed_or_time, int32_t start, int32_t en
     return LV_CLAMP(min_time * 10, time, max_time * 10);
 }
 
+bool lv_anim_is_paused(lv_anim_t * a)
+{
+    return a->is_paused;
+}
+
+void lv_anim_pause(lv_anim_t * a)
```
uLipe, line 510: What happens if `a` is NULL? Is guaranteed this will never occur? Additionally since this a function that does not return value, I would like to see at least a LOG_ERR/WRN when things get wrong, it would facillitate and avoid silent failures.

src/misc/lv_anim.c
```diff
@@ -499,6 +500,30 @@ uint32_t lv_anim_resolve_speed(uint32_t speed_or_time, int32_t start, int32_t en
...
+{
+    lv_anim_pause_for(a, LV_ANIM_PAUSE_FOREVER);
+}
+
+void lv_anim_pause_for(lv_anim_t * a, uint32_t ms)
```
uLipe, line 516: Please check also here.

src/misc/lv_anim.c
```diff
@@ -499,6 +500,30 @@ uint32_t lv_anim_resolve_speed(uint32_t speed_or_time, int32_t start, int32_t en
...
+{
+    a->is_paused = true;
+    a->pause_time = lv_tick_get();
+    a->pause_duration = ms;
+}
+
+void lv_anim_resume(lv_anim_t * a)
```
uLipe, line 522: And here.

uLipe (changes requested):
Nice work! I just have a small concern pointed, could you check please?

src/misc/lv_anim.c
AndreCostaaa: If we really want  to make sure that `a` is valid, we would need to go through the animation list to find `a`.

Having said that I agree with checking if it's NULL as it's always a good practice when dealing with pointers

uLipe: There are some asserting macros for NULL you can use on these cases, my intention here is just to provide some hint for the user  if something goes wrong, making easier to debug when the test coverage does not reach some potential edge case.

src/misc/lv_anim.c
```diff
@@ -694,6 +696,13 @@ static int32_t lv_anim_path_cubic_bezier(const lv_anim_t * a, int32_t x1, int32_
     return new_value;
 }
 
+static void lv_anim_pause_for_internal(lv_anim_t * a, uint32_t ms){
```
uLipe: Nit: brackets should be opened on the next line.

AndreCostaaa: Yep, forgot to run the formatting script before the commit. You can check the latest commit

src/misc/lv_anim.c
AndreCostaaa: Yep, I agree with you

uLipe (dismissed):

src/misc/lv_anim.h
```diff
@@ -146,6 +147,9 @@ struct _lv_anim_t {
 
     /* Animation system use these - user shouldn't set */
     uint32_t last_timer_run;
+    bool is_paused;                           /**<Indicates that the animation is paused */
```
kisvegabor: To save a few bytes please move it above `uint8_t reverse_play_in_progress : 1;`
```suggestion
    uint8_t is_paused :1;                           /**<Indicates that the animation is paused */
```

kisvegabor (commented):
Looks good to me, just on note.

uLipe (dismissed):

src/misc/lv_anim.c
```diff
@@ -519,8 +546,13 @@ static void anim_timer(lv_timer_t * param)
 
     while(a != NULL) {
         uint32_t elaps = lv_tick_elaps(a->last_timer_run);
-        a->act_time += elaps;
 
+        if(a->is_paused) {
+            a->is_paused = (lv_tick_get() - a->pause_time) < a->pause_duration;
```
liamHowatt: Would this be applicable?

```suggestion
            uint32_t time_spent_paused = lv_tick_get() - a->pause_time;
            if(time_spent_paused >= a->pause_duration) {
                a->is_paused = false;
                uint32_t pause_overrun = time_spent_paused - a->pause_duration;
                a->act_time += pause_overrun;
            }
```

liamHowatt: Also, maybe instead of `lv_tick_get() - a->pause_time`, `lv_tick_elaps(a->pause_time)` could be used to handle tick overflow wrap-around as used by `elaps`.

AndreCostaaa: Wouldn't updating the `act_time` mean a small "jump" when the animation is resumed ? It would probably go unnoticed as long as the `pause_overrun` is not too big but it got me wondering

liamHowatt: There would be a jump, ~~but there would be an even bigger jump at the next refresh without it~~. Here's a test case that passes with the change and fails without

```c
void test_anim_pause_for(void)
{
    int32_t var;

    /*Start an animation*/
    lv_anim_t a;
    lv_anim_init(&a);
    lv_anim_set_var(&a, &var);
    lv_anim_set_values(&a, 0, 100);
    lv_anim_set_exec_cb(&a, exec_cb);
    lv_anim_set_duration(&a, 100);
    lv_anim_t * animation = lv_anim_start(&a);

    lv_anim_pause_for(animation, 20);

    lv_test_wait(40);
    TEST_ASSERT_EQUAL(19, var);
}
```

liamHowatt: Also this more important issue. The overrun is lost and not counted. In fairness, it would only be noticeable if the refresh rate was very low.

```c
    ...
    lv_anim_t * animation = lv_anim_start(&a);

    lv_anim_pause_for(animation, 20);

    lv_test_wait(40);
    lv_test_wait(40);
    TEST_ASSERT_EQUAL(59, var); // "Expected 59 Was 39"
}
```

AndreCostaaa: Thank you for providing these test cases and i agree with you, will add the test cases and update the code. Nice catch !

liamHowatt (approved):
Thanks! Nice catch with the `FOREVER`.

kisvegabor (approved):
Thank you!

PR #7587, opened 2025-01-10
fix(label): add assignment to uninitialized variable

pfvogel:
If the alignment value is not LEFT, RIGHT or CENTER, the x is not assigned.  Add assignment to set x to zero if the alignment value is not LEFT, RIGHT or CENTER


uLipe (approved):

kisvegabor (approved):
Thanks!

PR #7602, opened 2025-01-14
fix(text): fix oob read for utf8-next

faxe1008:
Originally raised in zephyr clone https://github.com/zephyrproject-rtos/lvgl/pull/55 

### Notes

### DIFF
--- a/src/misc/lv_text.c
+++ b/src/misc/lv_text.c
@@ -649,6 +649,12 @@ static uint32_t lv_text_utf8_next(const char * txt, uint32_t * i)
     uint32_t i_tmp = 0;
     if(i == NULL) i = &i_tmp;
 
+    /* Ensure the string is not null */
+    if(txt == NULL || *i >= lv_strlen(txt)) {
+        (*i)++;
+        return result;
+    }
+
     /*Normal ASCII*/
     if(LV_IS_ASCII(txt[*i])) {
         result = txt[*i];
--- a/src/misc/lv_text.c
+++ b/src/misc/lv_text.c
@@ -429,7 +429,7 @@ int32_t lv_text_get_width_with_flags(const char * txt, uint32_t length, const lv
     lv_text_cmd_state_t cmd_state = LV_TEXT_CMD_STATE_WAIT;
 
     if(length != 0) {
-        while(i < length) {
+        while(txt[i] != '\0' && i < length) {
             uint32_t letter;
             uint32_t letter_next;
             lv_text_encoded_letter_next_2(txt, &letter, &letter_next, &i);
@@ -650,8 +650,7 @@ static uint32_t lv_text_utf8_next(const char * txt, uint32_t * i)
     if(i == NULL) i = &i_tmp;
 
     /* Ensure the string is not null */
-    if(txt == NULL || *i >= lv_strlen(txt)) {
-        (*i)++;
+    if(txt == NULL || txt[*i] == '\0') {
         return result;
     }
 
--- a/tests/src/test_cases/test_txt.c
+++ b/tests/src/test_cases/test_txt.c
@@ -142,27 +142,27 @@ void test_txt_next_line_should_handle_empty_string(void)
 void test_lv_text_encoded_letter_next_2_should_handle_null_pointer(void)
 {
     const char * txt = NULL;
-    uint32_t letter = 0, letter_next = 0, ofs = 0;
+    uint32_t letter = 100, letter_next = 101, ofs = 0;
 
     lv_text_encoded_letter_next_2(txt, &letter, &letter_next, &ofs);
 
     /* Expect both letter and letter_next to be 0 because the input string is NULL */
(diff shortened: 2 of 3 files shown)


src/misc/lv_text.c
```diff
@@ -649,6 +649,9 @@ static uint32_t lv_text_utf8_next(const char * txt, uint32_t * i)
     uint32_t i_tmp = 0;
     if(i == NULL) i = &i_tmp;
 
+    /* Ensure the string is not null */
+    if(txt == NULL || txt[*i] == '\0') return result;
```
uLipe: Suggest to track potential NULL strings by adding a LOG_WRN or a LOG_ERR

uLipe (commented):
Nice catch, just a small note.

src/misc/lv_text.c
faxe1008: Added a warning logging :^)

kisvegabor:
I'm not sure, but it seems the problem is rather [here](https://github.com/lvgl/lvgl/pull/7598/files#diff-87868f4fa80d2fbd02400750545f9d5db3581d6d096b84487e1421d6464f9e6bR893). Why do we need to overwrite the area to the whole widget area?

kisvegabor (dismissed):
Thanks!

kisvegabor:
Note the CI failure. I've checked it in detail, but I could reproduce it locally too.

faxe1008:
Yeah the issue that this loop will go infinitely now:
```
int32_t lv_text_get_width(const char * txt, uint32_t length, const lv_font_t * font, int32_t letter_space)
{
    if(txt == NULL) return 0;
    if(font == NULL) return 0;
    if(txt[0] == '\0') return 0;

    uint32_t i                   = 0;
    int32_t width             = 0;

    if(length != 0) {
        while(i < length) {
            uint32_t letter;
            uint32_t letter_next;
            lv_text_encoded_letter_next_2(txt, &letter, &letter_next, &i);

            int32_t char_width = lv_font_get_glyph_width(font, letter, letter_next);
            if(char_width > 0) {
                width += char_width;
                width += letter_space;
            }
        }

        if(width > 0) {
            width -= letter_space; /*Trim the last letter space. Important if the text is center
                                      aligned*/
        }
    }

    return width;
}
```

Because i is not incremented when hitting the end of the string

src/misc/lv_text.c
```diff
@@ -649,6 +649,13 @@ static uint32_t lv_text_utf8_next(const char * txt, uint32_t * i)
     uint32_t i_tmp = 0;
     if(i == NULL) i = &i_tmp;
 
+    /* Ensure the string is not null */
+    if(txt == NULL || *i >= lv_strlen(txt)) {
+        LV_LOG_WARN("txt is NULL or empty");
```
kisvegabor: I saw lot of these log messages, because it's valid that the next letter in `lv_text_encoded_letter_next_2` is `\0` when we are the last but one letter. In light of that I think this log should be removed.

kisvegabor (commented):
Thank you for the fix!

src/misc/lv_text.c
faxe1008: Removed the log.

faxe1008:
Rereading the PR just a thought:

It strikes me as kind of odd that for some of the other tests to pass we have to increment i even if it is beyond the string length:
```
if(txt == NULL || *i >= lv_strlen(txt)) {
        (*i)++;
        return result;
}
```
From the semantics of the api docs this seems wrong to me in retrospect:
```
/**
 * Decode an UTF-8 character from a string.
 * @param txt pointer to '\0' terminated string
 * @param i start byte index in 'txt' where to start.
 *          After call it will point to the next UTF-8 char in 'txt'.
 *          NULL to use txt[0] as index
 * @return the decoded Unicode character or 0 on invalid UTF-8 code
 */
```
If we are already at the last UTF-8 codepoint there is no next one. Incrementing anyways feels wrong. I'd advocate in fixing the users of said method and do not increment `i`. What do you think @uLipe  @kisvegabor ?

kisvegabor:
You are perfectly right. I've just looked deeper into it, and could make it work without out of bounds incrementing. What do you think?

uLipe:
I like the idea as well, if you are at the bounds even though it does not causes functional impact, incrementing is conceiptially wrong

kisvegabor:
@uLipe Could you approve it if it looks good?

uLipe (approved):

faxe1008:
Appreciate the approval but the discussed change is still missing :D

kisvegabor:
It's added in [18232f3](https://github.com/lvgl/lvgl/pull/7602/commits/18232f30ad2e2f6928f41fe4c25b98e659e318d8). See above. Or did I miss something?

faxe1008:
@kisvegabor All good then, did not see that you added something on top 😃

liamHowatt (approved):

PR #7609, opened 2025-01-15
fix(chart): avoid divide by zero

pfvogel:
For a BAR chart with a series count of zero, the count is being used in a divisor to adjust the x position.  Change to bypass the x adjustment if the series count is zero,

### DIFF
--- a/src/widgets/chart/lv_chart.c
+++ b/src/widgets/chart/lv_chart.c
@@ -261,10 +261,15 @@ void lv_chart_get_point_pos_by_id(lv_obj_t * obj, lv_chart_series_t * ser, uint3
         }
 
         p_out->x = (int32_t)((int32_t)(w + block_gap) * id) / chart->point_cnt;
-        p_out->x += block_w * ser_idx / ser_cnt;
+        if(ser_cnt > 0) {
+            p_out->x += block_w * ser_idx / ser_cnt;
 
-        int32_t col_w = (block_w - (ser_gap * (ser_cnt - 1))) / ser_cnt;
-        p_out->x += col_w / 2;
+            int32_t col_w = (block_w - (ser_gap * (ser_cnt - 1))) / ser_cnt;
+            p_out->x += col_w / 2;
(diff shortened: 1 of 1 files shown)


uLipe (changes requested):
Please add a LOG or ASSERT when a zero occurs

uLipe:
Otherwise, nice work!

uLipe:
Fair enough, so since it is not clear in that cases what is a fatal error or not, I still think at least a LOG_WRN should be called to provide a hint to debug on cases where this could be an issue, what do you think @pfvogel ?

pfvogel:
Adding a WARN message seems reasonable, since this does look like an unusual case.  I have added it and pushed the change to the pull request

uLipe (dismissed):

FASTSHIFT (dismissed):

src/widgets/chart/lv_chart.c
```diff
@@ -261,10 +261,13 @@ void lv_chart_get_point_pos_by_id(lv_obj_t * obj, lv_chart_series_t * ser, uint3
         }
 
         p_out->x = (int32_t)((int32_t)(w + block_gap) * id) / chart->point_cnt;
-        p_out->x += block_w * ser_idx / ser_cnt;
+        if(ser_cnt > 0) {
+            LV_LOG_WARN("bar chart series count is zero");
```
XuNeo: The `if(ser_cnt > 0) {`  does not match with the warning?

XuNeo (changes requested):

src/widgets/chart/lv_chart.c
pfvogel: You are right, good catch.  I will fix this and push again.  Thanks!

pfvogel:
Updated version has been pushed

uLipe (approved):

XuNeo (approved):

liamHowatt (approved):

PR #7610, opened 2025-01-15
fix(bin_decoder): check divisor is zero

pfvogel:
If the color format is unknown, the bits per pixel variable is set to zero.  This value is then used as a divisor.  Add assertion prior to division to check that the "bpp" value is not zero

### DIFF
--- a/src/libs/bin_decoder/lv_bin_decoder.c
+++ b/src/libs/bin_decoder/lv_bin_decoder.c
@@ -838,6 +838,12 @@ static lv_result_t decode_alpha_only(lv_image_decoder_t * decoder, lv_image_deco
     uint32_t rn;
     decoder_data_t * decoder_data = dsc->user_data;
     uint8_t bpp = lv_color_format_get_bpp(dsc->header.cf);
+
+    if(bpp == 0) {
+        LV_LOG_ERROR("Error color format: %d", dsc->header.cf);
+        return LV_RESULT_INVALID;
+    }
+
     uint32_t w = (dsc->header.stride * 8) / bpp;
     uint32_t buf_stride = (w * 8 + 7) >> 3; /*stride for img_data*/
     uint32_t buf_len = w * dsc->header.h; /*always decode to A8 format*/


uLipe (dismissed):

src/libs/bin_decoder/lv_bin_decoder.c
```diff
@@ -838,6 +838,7 @@ static lv_result_t decode_alpha_only(lv_image_decoder_t * decoder, lv_image_deco
     uint32_t rn;
     decoder_data_t * decoder_data = dsc->user_data;
     uint8_t bpp = lv_color_format_get_bpp(dsc->header.cf);
+    LV_ASSERT(bpp != 0);
```
FASTSHIFT:
```c
if(bpp == 0) {
  LV_LOG_ERROR("Error color format: %d", dsc->header.cf);
  retrun LV_RESULT_INVALID;
}
```

pfvogel:
I've made the suggested change to replace the LV_ASSERT with a check/LOG_ERROR/return INVALID

XuNeo (commented):
But the function is already guarded by `LV_COLOR_FORMAT_IS_ALPHA_ONLY`

See 
```
if(LV_COLOR_FORMAT_IS_ALPHA_ONLY(cf)) {
        res = decode_alpha_only(decoder, dsc);
    }
```

How the assert is triggered?

pfvogel:
In the section of code in lines 250-290:

```
       lv_color_format_t cf = image->header.cf;                      // line 257
...
        else if(LV_COLOR_FORMAT_IS_ALPHA_ONLY(cf)) {                 // line 277
...
                res = decode_alpha_only(decoder, dsc);               // line 290
```

So the guard is validating the 'cf' value from the image header, but the 'bpp' value is calculated from the dsc header, so it may not be the same value.
Maybe instead the dsc header should be updated from the image header to make them consistent?

pfvogel:
The division by zero issue was detected from a static analysis tool review of the code, rather than an actual runtime detection.

The issue is that the macro that checks to see if the 'cf' value is alpha is using the value of

`((lv_image_dsc_t *)dsc->src)->header.cf`

while the 'bpp' value is calculated from

`dsc->header.cf`

which may not be the same value for 'cf'

I have attached the report output from the static analysis that shows the path and assumptions needed to trigger it, if that helps to define the issue.

[lv_bin_decoder.c_clangsa_24fc6b21620914413bc7a206f8317a68.plist.zip](https://github.com/user-attachments/files/18524834/lv_bin_decoder.c_clangsa_24fc6b21620914413bc7a206f8317a68.plist.zip)

uLipe (approved):

FASTSHIFT (approved):

PR #7611, opened 2025-01-15
chore(draw): fix unused function warning

thughes:
Building with clang and `-Wunused-function` gives the following warning:

```
src/draw/lv_draw.c:37:24: error: unused function
'get_layer_size_kb' [-Werror,-Wunused-function]
static inline uint32_t get_layer_size_kb(uint32_t size_byte)
                       ^
```

`get_layer_size_kb` is only used from the `LV_LOG_LEVEL_INFO` macro, so only include it if the log level is set to INFO or lower.


FASTSHIFT (approved):

thughes:
@uLipe @kisvegabor @liamHowatt It looks like I need someone else to review. Would you mind taking a look at this trivial change?

XuNeo (approved):

PR #7617, opened 2025-01-16
nuttx: replace LV_LOG_USER with LV_LOG_INFO

michallenc:
Debug outputs are changed from `LV_LOG_USER` to `LV_LOG_INFO`. These are just informational outputs (device opened, framebuffer heap info, etc.) and their priority should be lower compared to error and warning outputs. It imho just does't make any sense to have these as `LV_LOG_INFO`. If I build LVGL with `LV_LOG_LEVEL_ERROR` I expect only error outputs, not information like device successfully opened.

### DIFF
--- a/src/drivers/nuttx/lv_nuttx_entry.c
+++ b/src/drivers/nuttx/lv_nuttx_entry.c
@@ -295,7 +295,7 @@ static void check_stack_size(void)
 {
     pthread_t tid = pthread_self();
     ssize_t stack_size = pthread_get_stacksize_np(tid);
-    LV_LOG_USER("tid: %d, Stack size : %zd", (int)tid, stack_size);
+    LV_LOG_INFO("tid: %d, Stack size : %zd", (int)tid, stack_size);
 
     if(stack_size < LV_NUTTX_MIN_STACK_SIZE) {
         LV_LOG_ERROR("Stack size is too small. Please increase it to %d bytes or more.",
--- a/src/drivers/nuttx/lv_nuttx_fbdev.c
+++ b/src/drivers/nuttx/lv_nuttx_fbdev.c
@@ -105,7 +105,7 @@ int lv_nuttx_fbdev_set_file(lv_display_t * disp, const char * file)
         LV_LOG_ERROR("Error: cannot open framebuffer device");
         return -errno;
     }
-    LV_LOG_USER("The framebuffer device was opened successfully");
+    LV_LOG_INFO("The framebuffer device was opened successfully");
 
     if(ioctl(dsc->fd, FBIOGET_VIDEOINFO, (unsigned long)((uintptr_t)&dsc->vinfo)) < 0) {
         LV_LOG_ERROR("ioctl(FBIOGET_VIDEOINFO) failed: %d", errno);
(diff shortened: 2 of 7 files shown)


kisvegabor (approved):
Makes sense to me. cc @XuNeo

src/drivers/nuttx/lv_nuttx_entry.c
```diff
@@ -295,7 +295,7 @@ static void check_stack_size(void)
 {
     pthread_t tid = pthread_self();
     ssize_t stack_size = pthread_get_stacksize_np(tid);
-    LV_LOG_USER("tid: %d, Stack size : %zd", (int)tid, stack_size);
```
FASTSHIFT, line 298: Instead, I think it's necessary to use user-level printing because lvgl needs to run on many hardware platforms and we have to make sure lvgl has been properly initialized with fb. In the past, many lvgl display errors were found to be caused by fb driver reporting errors. These logs can effectively reduce problem location time.

FASTSHIFT (changes requested):

src/drivers/nuttx/lv_nuttx_entry.c
FASTSHIFT: This log is only printed once when the lvgl is started and does not have much effect on code execution.

michallenc: But in that case why not just have error/warning outputs when something fails? We could verbose the framebuffer heap etc. in case of error for example. The output also collides with NuttX shell. For example my terminal output after boot is:

```
N[LVGL] [User]	(1.440, +1440)	 check_stack_size: tid: 3, Stack size : 8192 lv_nuttx_entry.c:297
u[LVGL] [User]	(1.440, +0)	 lv_nuttx_lcd_create: lcd /dev/lcd0 opening lv_nuttx_lcd.c:77
t[LVGL] [User]	(1.450, +10)	 lv_nuttx_lcd_create: lcd /dev/lcd0 open success lv_nuttx_lcd.c:84
tShell (NSH) NuttX-12.8.0
```

And I use LCD driver, it seems there is more verbose output with framebuffer. It is not so much about code execution, it just make an unnecessary mess on the command line in my opinion. It may complicate terminal parsing for some automated tools for example.

XuNeo: `printf` and `syslog` will collides anyway because of the two tasks have different priority, but there are two buffers can be used to minimize the effect.

1. `CONFIG_STDIO_BUFFER_SIZE` should be enlarged, so printf can output the content in batch. In the example you provided, it seems this buffer is disabled.
2. `CONFIG_SYSLOG_BUFSIZE` is used for syslog buffer.

michallenc: Yes, the output is good with different buffers, good point. But still I think only errors should be printed imho.

michallenc:
@XuNeo @FASTSHIFT @kisvegabor Hi, any update before this gets closed by the bot? So far it's +1 and -1.

XuNeo (approved):
Looks good to me!

XuNeo:
> I know we are arguing about marginal stuff here, the outputs do not cause performance issues as they are in initialization stage only.

Hi @michallenc, you’re right.

I agree with the change since this should be an info-level log — once the project is set up correctly, it’s unlikely to be relevant.

However, FASTSHIFT prefers to keep it always printed because, in our experience at Xiaomi, we've encountered numerous setup issues across multiple projects. Having this information readily available makes troubleshooting much easier.

This is the reason why LV_LOG_USER is added instead of INFO/ERROR etc.

> If I build LVGL with LV_LOG_LEVEL_ERROR I expect only error outputs, not information like device successfully opened.

But your point also makes sense. If the one-time log doesn't hurt, maybe we should keep it?

FASTSHIFT:
> And what is the current output for too small cache? The user doesn't know the lags are caused by this value anyway. And if LVGL has some automatic check, then it can be LV_LOG_ERROR output, similar to stack size control.

There is no clear right or wrong evaluation standard for the image cache size. It needs to be fine-tuned according to the RAM size provided by the user's hardware and the complexity of the application to achieve a balance between performance and memory usage.

michallenc:
@FASTSHIFT @XuNeo ok, so let's close it if you have a use case for the verbose print out.

PR #7619, opened 2025-01-16
ci: cancel previous workflow run on new commit

AndreCostaaa:
Currently, when a new commit gets pushed for a specific PR, multiple workflows will run at the same time for different commits in the same PR

Since we most likely only care about the latest commit made to the PR, this concurrency rule cancels the previous workflow run when a new commit is made to the same `github.ref` which will, in the case of a push be the branch name

I've also added the same rule to some checks that only happen in master for the same reason, when merging multiple PR's one after the other the same reasoning is probably justified

I didn't create an issue before so I apologize if you feel like this is not a good idea

@liamHowatt

### DIFF
--- a/.github/workflows/arduino.yml
+++ b/.github/workflows/arduino.yml
@@ -4,6 +4,13 @@ on:
     branches: [ master, release/v8.* ]
   pull_request:
     branches: [ master, release/v8.* ]
+
+# https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#concurrency
+# Ensure that only one commit will be running tests at a time on each PR
+concurrency:
+  group: ${{ github.ref }}-${{ github.workflow }} 
+  cancel-in-progress: true
+
 jobs:
   lint:
     if: ${{ github.event_name != 'pull_request' || github.repository != github.event.pull_request.head.repo.full_name }}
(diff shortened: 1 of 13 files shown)


XuNeo (approved):

liamHowatt (approved):
> I didn't create an issue before so I apologize if you feel like this is not a good idea

It's fine to do!

Thanks!

liamHowatt (commented):
I noticed in `.github/workflows/compile_docs.yml` there is already a line later which is `concurrency: docs-build-and-deploy`.

https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#jobsjob_idconcurrency

https://github.com/lvgl/lvgl/blob/2b97e0c022c7d859db7f64d3cb998fdd99bca69b/.github/workflows/compile_docs.yml#L21

I think it ensures only one instance runs across all branches.

AndreCostaaa:
Oh right, didn't see that one

I would be in favor of removing the old one as the new one prevents the whole workflow from running as that one only prevents that specific job. The reasoning here is that if a new job gets added for some reason the same rule would have to be added to that new job

Also the old one prevents the workflow from running if there something is merged to `master` and to `release/X` in a short interval

What do you think ?

PR #7632, opened 2025-01-20
feat(osal): add lv_os_get_idle_percent for linux

### LINKED ISSUE #7592: implement get_idle_percent for linux
### Introduce the problem

I was wondering if it would be possible to create a function similar to lv_os_get_idle_percent() from freertos:
https://github.com/lvgl/lvgl/blob/d9677acd93f6eeeb9f50f68363516dfae2233570/src/osal/lv_freertos.c#L408
but make it in lv_pthread.c for linux?

### Proposal

_No response_

AndreCostaaa:
Fixes #7592 

- Implemented lv_os_get_idle_percent for linux by reading the `proc/uptime` file which returns the active and idle times of the system since boot. From `man proc_uptime`:

```
proc_uptime(5)                                                                                   File Formats Manual                                                                                  proc_uptime(5)

NAME
       /proc/uptime - system uptime

DESCRIPTION
       /proc/uptime
              This file contains two numbers (values in seconds): the uptime of the system (including time spent in suspend) and the amount of time spent in the idle process.
``` 

The values are in seconds with two decimal values

```bash
cat /proc/uptime
7673.63 88861.75
```
This implementation doesn't use floating point values

- Created a default `lv_os_get_idle_percent` that calls `lv_timer_get_idle` for other os

### DIFF
--- a/src/osal/lv_linux.c
+++ b/src/osal/lv_linux.c
@@ -97,7 +97,7 @@ uint32_t lv_os_get_idle_percent(void)
 
 static lv_result_t lv_read_proc_stat(lv_proc_stat_t * result)
 {
-    FILE *fp = fopen(LV_UPTIME_MONITOR_FILE, "r");
+    FILE * fp = fopen(LV_UPTIME_MONITOR_FILE, "r");
 
     if(!fp) {
         LV_LOG_ERROR("Failed to open " LV_UPTIME_MONITOR_FILE);
--- a/src/osal/lv_linux.c
+++ b/src/osal/lv_linux.c
@@ -11,7 +11,9 @@
 
 #if LV_USE_OS != LV_OS_NONE && defined(__linux__)
 
+#include "../core/lv_global.h"
 #include "../misc/lv_log.h"
+#include "lv_linux_private.h"
 #include <stdio.h>
 
 /*********************
@@ -22,23 +24,10 @@
 
 #define LV_PROC_STAT_VAR_FORMAT        " %" PRIu32
 #define LV_PROC_STAT_IGNORE_VAR_FORMAT " %*" PRIu32
-#define LV_PROC_STAT_PARAMS_LEN        7
 
 /**********************
  *      TYPEDEFS
  **********************/
-typedef union {
-    struct {
-        /*
-         *  We ignore the iowait column as it's not reliable
-         *  We ignore the guest and guest_nice columns because they're accounted
-         *   for in user and nice respectively
-         */
-        uint32_t user, nice, system, idle, /*iowait,*/ irq, softirq,
-                 steal /*, guest, guest_nice*/;
-    } fields;
-    uint32_t buffer[LV_PROC_STAT_PARAMS_LEN];
-} lv_proc_stat_t;
 
 /**********************
  *  STATIC PROTOTYPES
@@ -51,11 +40,11 @@ static uint32_t lv_proc_stat_get_total(const lv_proc_stat_t * p);
  *  STATIC VARIABLES
  **********************/
 
-static lv_proc_stat_t last_proc_stat;
-
 /**********************
  *      MACROS
  **********************/
+
+#define last_proc_stat LV_GLOBAL_DEFAULT()->linux_last_proc_stat
 
 /**********************
  *   GLOBAL FUNCTIONS
--- a/src/osal/lv_linux.c
+++ b/src/osal/lv_linux.c
@@ -100,6 +100,7 @@ static int lv_proc_get_uptime(uint32_t * active_s, int * active_ms, uint32_t * i
         LV_LOG_WARN("Failed to parse " LV_UPTIME_MONITOR_FILE);
         return -1;
     }
+    return 0;
 }
 
 #endif
--- a/src/osal/lv_linux.c
+++ b/src/osal/lv_linux.c
@@ -36,7 +36,7 @@ uint32_t lv_os_get_idle_percent(void)
     uint32_t delta_uptime_s, delta_idletime_s;
     int delta_uptime_ms, delta_idletime_ms;
 
-    // Calculate the delta first to avoid overflowing
+    /* Calculate the delta first to avoid overflowing */
     lv_proc_get_delta(uptime_s, uptime_ms, last_uptime_s,
                       last_uptime_ms, &delta_uptime_s, &delta_uptime_ms);
 
@@ -44,7 +44,7 @@ uint32_t lv_os_get_idle_percent(void)
                       last_idletime_ms, &delta_idletime_s,
                       &delta_idletime_ms);
 
-    /* From here onwards, there's no risk as long as we call this function regularly */
+    /* From here onwards, there's no risk of overflowing as long as we call this function regularly */
 
     /* Update for next call */
     last_uptime_s = uptime_s;
--- a/src/osal/lv_linux.c
+++ b/src/osal/lv_linux.c
@@ -18,9 +18,9 @@
  *      DEFINES
  *********************/
 
-#define LV_UPTIME_MONITOR_FILE	       "/proc/stat"
+#define LV_UPTIME_MONITOR_FILE         "/proc/stat"
 
-#define LV_PROC_STAT_VAR_FORMAT	       " %" PRIu32
+#define LV_PROC_STAT_VAR_FORMAT        " %" PRIu32
 #define LV_PROC_STAT_IGNORE_VAR_FORMAT " %*" PRIu32
 
 /**********************
@@ -31,7 +31,7 @@
  *  STATIC PROTOTYPES
  **********************/
 
-static lv_result_t lv_proc_get_uptime(uint32_t *active, uint32_t *idle);
+static lv_result_t lv_proc_get_uptime(uint32_t * active, uint32_t * idle);
 
 /**********************
  *  STATIC VARIABLES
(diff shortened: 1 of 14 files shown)


src/osal/lv_linux.c
```diff
@@ -0,0 +1,106 @@
...
+static int last_uptime_ms, last_idletime_ms;
+
+uint32_t lv_os_get_idle_percent(void)
+{
+
+    uint32_t delta_active_s, delta_idle_s;
+    int delta_active_ms, delta_idle_ms;
+    {
+
+        // UINT32_MAX seconds > 136 years
```
uLipe: Please use multiline comments style.

src/osal/lv_linux.c
```diff
@@ -0,0 +1,106 @@
...
+        uint32_t active_s, idletime_s;
+        // Range is [0: 99[
```
uLipe: Also here

src/osal/lv_linux.c
```diff
@@ -0,0 +1,106 @@
...
+    }
+}
+
+static int lv_proc_get_uptime(uint32_t * active_s, int * active_ms, uint32_t * idle_s, int * idle_ms)
+{
+
+    FILE * fp = fopen(LV_UPTIME_MONITOR_FILE, "r");
+
+    if(!fp) {
+        LV_LOG_WARN("Failed to open " LV_UPTIME_MONITOR_FILE);
```
uLipe: I think this can be a LOG_ERR, even though is not a catastrophic failure, is not intended that the percentage module to work without access to the file.

src/osal/lv_linux.c
```diff
@@ -0,0 +1,106 @@
...
+
+    int err = fscanf(fp,
+                     "%" PRIu32 ".%d"
+                     " %" PRIu32 ".%d",
+                     active_s, active_ms, idle_s, idle_ms);
+
+    fclose(fp);
+
+    if(err != 4) {
+        LV_LOG_WARN("Failed to parse " LV_UPTIME_MONITOR_FILE);
```
uLipe: Same as above.

uLipe (changes requested):
Nice work, I just placed a couple of nits.

src/osal/lv_linux.c
AndreCostaaa: Yep, you're right

AndreCostaaa:
Thanks @uLipe ! :rocket: 

I was wondering about two things actually

1. Was putting the function in a new `lv_linux.c` file the way to do this? Since the same function could be used by `SDL2` and `pthread` or even if none of these things is defined it didn't feel right to put it in one of those files

2. Also not sure if using a `defined(__linux__)` clause is the way or if we should a add a new `LV_OS_LINUX` define ? Maybe we could do something like

```c
#ifndef LV_OS_LINUX && defined(__linux__)
#define LV_OS_LINUX
#endif 
```

Any thoughts ?

src/osal/lv_linux.c
```diff
@@ -0,0 +1,106 @@
...
+
+        /* UINT32_MAX seconds > 136 years */
+        uint32_t active_s, idletime_s;
+        /* Range is [0: 99[ */
+        int active_ms, idletime_ms;
+
+        int err = lv_proc_get_uptime(&active_s, &active_ms, &idletime_s, &idletime_ms);
+
+        if(err < 0) {
+            return UINT_MAX;
```
liamHowatt:
```suggestion
            return UINT32_MAX;
```

src/osal/lv_linux.c
```diff
@@ -0,0 +1,106 @@
...
+    {
+
+        /* UINT32_MAX seconds > 136 years */
+        uint32_t active_s, idletime_s;
+        /* Range is [0: 99[ */
+        int active_ms, idletime_ms;
+
+        int err = lv_proc_get_uptime(&active_s, &active_ms, &idletime_s, &idletime_ms);
+
+        if(err < 0) {
```
liamHowatt: Recommend using `lv_result_t` for success and failure of new LVGL functions.

liamHowatt (commented):
Oooh I think this is generally a good way to approach it! I can't think of any better way to handle it, RE the things you mentioned.

It works well when I test.

Opening, reading, and closing the file must not be much slower than open, read, seek(0, 0), read, seek(0, 0), read, (repeat). If we did that we'd have to add a deinit handler to close the file so I think it's fine as-is.

Please add the section titles from `src/misc/lv_templ.c` to `lv_linux.c` and `lv_os_none.c`.

Can you use `int32_t` instead of `int`? We try to avoid it throughout LVGL. For the return value of `fscanf` it's good though.

liamHowatt:
Fyi, when I run `stress` on 3 cores, I see LVGL reporting 50% CPU usage while `htop` shows what looks like 75%.

Screenshot from 2025-01-23 15-12-31

AndreCostaaa:
Can you try it now @liamHowatt ?

Results should now perfectly match

Screenshot From 2025-01-23 22-30-39
Screenshot From 2025-01-23 22-30-48
Screenshot From 2025-01-23 22-30-56


PS: taking the screenshot makes the usage spike for a bit so that's why the screenshots don't show 50% when using 6 cores and 25% with 3

src/osal/lv_os_none.c
liamHowatt, line 1: Please use the template for this one too

src/osal/lv_rtthread.c
```diff
@@ -11,6 +11,7 @@
 #if LV_USE_OS == LV_OS_RTTHREAD
 
 #include "../misc/lv_log.h"
+#include "../misc/lv_timer.h"
```
liamHowatt, line 14: To be consistent with what you've done in the other files, put this in a `#ifndef __linux__`

liamHowatt (commented):
Nice! They match now, awesome!

src/osal/lv_rtthread.c
AndreCostaaa: Sorry, not familiar with RTThread, is it something that can be used with linux ? I thought it was a different os completely ?

AndreCostaaa: I just checked and yes, they're totally different OS. I made a mistake in line 188

src/osal/lv_os_none.c
AndreCostaaa: Fixed, sorry about that

liamHowatt (commented):
Thanks. You made it look easy!

Just one more thing I want to clarify.

https://man7.org/linux/man-pages/man5/proc_stat.5.html

You're grabbing `user` -- time spent in normal programs. Do you think it would make sense to also add the value of `system`, i.e. `active_user + active_system`?

`system` is time spent in Linux kernel code. If I run `stress --io 4` it spends most of the time in kernel code as shown by the **red** color in `htop`. That's real CPU utilization just like `user`.

Here is `stress --io 4`

image


It's rarer to see `nice` -- low priority tasks -- in `htop` but it's **blue** when it happens. For me I see some `nice` activity after taking a screenshot with PrtScr. Maybe that's just my system. You can use the command `nice -n 20 [command]` to run something with the lowest priority to cause it. The strangest part is that your PR _does_ pick up this measurement in the percentage somehow! (but not `system` for some reason) so maybe this value does not need to be added?

Here is `nice -n 20 stress -c 4`

image

Edit: It makes sense that `nice` is included as part of `user`

> nice   (2) Time spent in user mode with low priority (nice).

because it's part of user.

AndreCostaaa:
Yes you're right

There are 10 columns in `/proc/stat`

```bash
  user   (1) Time spent in user mode.

                     nice   (2) Time spent in user mode with low priority (nice).

                     system (3) Time spent in system mode.

                     idle   (4) Time spent in the idle task.  This value should be USER_HZ times the second entry in the /proc/uptime pseudo-file.

                     iowait (since Linux 2.5.41)
                            (5) Time waiting for I/O to complete.  This value is not reliable, for the following reasons:

                            •  The  CPU  will  not  wait for I/O to complete; iowait is the time that a task is waiting for I/O to complete.  When a CPU goes into idle state for outstanding task I/O, another task
                               will be scheduled on this CPU.

                            •  On a multi-core CPU, the task waiting for I/O to complete is not running on any CPU, so the iowait of each CPU is difficult to calculate.

                            •  The value in this field may decrease in certain conditions.

                     irq (since Linux 2.6.0)
                            (6) Time servicing interrupts.

                     softirq (since Linux 2.6.0)
                            (7) Time servicing softirqs.

                     steal (since Linux 2.6.11)
                            (8) Stolen time, which is the time spent in other operating systems when running in a virtualized environment

                     guest (since Linux 2.6.24)
                            (9) Time spent running a virtual CPU for guest operating systems under the control of the Linux kernel.

                     guest_nice (since Linux 2.6.33)
                            (10) Time spent running a niced guest (virtual CPU for guest operating systems under the control of the Linux kernel).
``` 

We can start by ignoring column 5 since it's not considered reliable.

Column user and nice are increased here: https://elixir.bootlin.com/linux/v6.12.6/source/kernel/sched/cputime.c#L121
They depend on the niceness level. 

These two columns also get added alongside columns `guest` and `guest nice` here: https://elixir.bootlin.com/linux/v6.12.6/source/kernel/sched/cputime.c#L153-L159
Meaning columns `guest` and `guest nice` can be ignored, so columns 9 and 10

Steal time is updated here: https://elixir.bootlin.com/linux/v6.12.6/source/kernel/sched/cputime.c#L211
So it should be used to count alongside the total time

For the system, irq and softirq, they don't overlap so they must get added for the total time spent on the kernel : https://elixir.bootlin.com/linux/v6.12.6/source/kernel/sched/cputime.c#L188

So in the end. Columns 1, 2, 3, 4, 6, 7 and 8 should all be accounted for the total time wile column 4 is the total idle time. 



> Edit: It makes sense that `nice` is included as part of `user`
> 
> > nice   (2) Time spent in user mode with low priority (nice).
> 
> because it's part of user.

This was my first guess but looking into the kernel code I don't see how that can be as they get added separately. So i'm not too sure. You can grep `CPUTIME_USER` in https://elixir.bootlin.com/linux/v6.12.6/source/kernel/sched/cputime.c and see that every time that every code branch updates either user  or nice but not both

AndreCostaaa:
The latest commit now correctly accounts for all columns based on the description in my previous comment

I tested the `nice` time with `nice -n 20 stress -c X`  and the results are correct. Can you double-check this? I still find it intriguing that your previous test showed the correct values @liamHowatt

src/osal/lv_linux.c
```diff
@@ -0,0 +1,133 @@
...
+ *      TYPEDEFS
+ **********************/
+typedef union {
+    struct {
+        /*
+         *  We ignore the iowait column as it's not reliable
+         *  We ignore the guest and guest_nice columns because they're accounted
+         *   for in user and nice respectively
+         */
+        uint32_t user, nice, system, /*iowait,*/ idle, irq, softirq,
```
liamHowatt:
```suggestion
        uint32_t user, nice, system, idle, /*iowait,*/ irq, softirq,
```

liamHowatt (commented):
Thanks Andre, it works great! One small comment change and I'll approve.

The below is just something interesting, not feedback:

I understand why it showed 100% when I `stress`ed `nice` but `nice` wasn't being measured yet. It was because `idle` is divided by the total, and no matter what the total is, idle is a small number so a small number divided by anything is close to zero. 0 idle means 100% usage.

I confirmed `nice` is the only counter that changes meaningfully while stressing `nice`

```
$ head -n 1 /proc/stat && sleep 10 && head -n 1 /proc/stat
cpu  784546 76622 263316 4495620 5366 0 5375 0 0 0
cpu  784614 80516 263357 4495622 5366 0 5375 0 0 0
```

(delta 3894 over 10 seconds, divided by 4 for 4 CPUs, divided by 100 for 100 Hz = 9.735 seconds spent in `nice`)

liamHowatt (dismissed):

AndreCostaaa:
> I understand why it showed 100% when I `stress`ed `nice` but `nice` wasn't being measured yet. It was because `idle` is divided by the total, and no matter what the total is, idle is a small number so a small number divided by anything is close to zero. 0 idle means 100% usage.

I guess we would've caught the error if we tried to aim for a lower percentage :thinking: 

> Thanks Andre, it works great! One small comment change and I'll approve.

Thank you for your help in getting this across the finish line!  :smile:  :rocket:

src/osal/lv_linux.c
```diff
@@ -0,0 +1,133 @@
...
+ **********************/
+
+/**********************
+ *   GLOBAL FUNCTIONS
+ **********************/
+
+uint32_t lv_os_get_idle_percent(void)
+{
+    lv_proc_stat_t proc_stat;
+    {
```
kisvegabor, line 56: Why do we need `{` here?

src/osal/lv_linux.c
```diff
@@ -0,0 +1,133 @@
...
+ **********************/
+
+static lv_result_t lv_read_proc_stat(lv_proc_stat_t * result);
+static uint32_t lv_proc_stat_get_total(const lv_proc_stat_t * p);
+
+/**********************
+ *  STATIC VARIABLES
+ **********************/
+
+static lv_proc_stat_t last_proc_stat;
```
kisvegabor: It should go to `lv_global.h`. In this file you can easily reference it with `#define last_proc_stat LV_GLOBAL_DEFAULT()->linux_last_proc_stat`

kisvegabor (commented):
I almost had some meaningful comments but just seen that all of them are handled, so I added only 2 minor remarks.

 Great work! :rocket:

src/osal/lv_linux.c
AndreCostaaa: Sounds good. Done !

src/osal/lv_linux.c
AndreCostaaa: It's a minor artifact from a previous version where I had defined four variables within a small scope to keep them contained and avoid their use later in the function. We can leave this as is - the `err` variable can be discarded earlier, and it also provides a visual separation between the delta calculation and the actual percentage calculation

If you still think it should be removed despite this explanation, let me know, and I’ll make the change

liamHowatt (approved):

src/osal/lv_linux.c
liamHowatt: I have done it before too: https://github.com/lvgl/lvgl/pull/6042#discussion_r1556344898

I avoid it in LVGL because it's uncommon here and lesser-known. Gabor, it's an anonymous block/scope. Any local variables declared in it cannot be accessed afterward. In long functions there can be lots of local variables, some of which were just temporaries, so they can go in an anonymous scope so they go out of scope after they're done being used for an intermediate calculation.

uLipe (approved):

uLipe:
NIce work! Just wait from @kisvegabor as wel.

FASTSHIFT (approved):

src/osal/lv_linux.c
```diff
@@ -0,0 +1,122 @@
...
+    return (proc_stat.fields.idle * 100) / total;
+}
+
+/**********************
+ *   STATIC FUNCTIONS
+ **********************/
+
+static lv_result_t lv_read_proc_stat(lv_proc_stat_t * result)
+{
+    FILE * fp = fopen(LV_UPTIME_MONITOR_FILE, "r");
```
FASTSHIFT, line 89: Can this `FILE * fp` be stored in the global context? Avoid repeated opening and closing of files to improve operating efficiency.

AndreCostaaa: Already discussed above with @liamHowatt https://github.com/lvgl/lvgl/pull/7632#pullrequestreview-2570854198

It would also mean we need a deinit function so we can close the file at the end which is incompatible with the current api

FASTSHIFT: OK.

PR #7633, opened 2025-01-20
feat(xml): allow to set obj id through xml

gabordemooij:
Allows to set id of an object through xml.


kisvegabor:
Hey,

This it the first contribution the new XML support feature! Thank you very much! 

In the end, I think it will be a little more complicated. See the discussion [here](https://github.com/lvgl/lvgl/issues/7476#issuecomment-2577497768).

There are some Editor related features in pipeline where this feature will be required. The requirements are still forming, so I'd like to postpone it a little.

gabordemooij:
Yes, of course, I understand. No problem. I need this feature right away so I am implementing it in my own forked branch. When your implementation is finished I will replace it. Sounds good.

PR #7638, opened 2025-01-21
fix(indev): add the missing wait_until_release flag to clean up

yimoxiao:
fix(indev): When releasing, satisfying wait until release and reset indev will cause the next clicked to fail

Example Code：
```
static void event_cb(lv_event_t * e)
{
  if(code == LV_EVENT_PRESSED) {
    lv_indev_wait_release(lv_indev_get_act());
  } else if(code == LV_EVENT_PRESS_LOST) {
    LV_LOG_INFO("press lost");
    lv_obj_t* obj = lv_event_get_target(e);
    lv_indev_reset(lv_indev_get_act(), obj);
  }
}

static void demo()
{
  lv_obj_t* btn = lv_button_create(lv_scr_act());
  lv_obj_t* label = lv_label_create(btn);
  lv_label_set_text(label, "Button");
  lv_obj_set_size(btn, 100, 100);
  lv_obj_center(btn);
  lv_obj_remove_flag(btn, LV_OBJ_FLAG_PRESS_LOCK);

  lv_obj_add_event_cb(btn, event_cb, LV_EVENT_ALL, NULL);
}
```

### DIFF
--- a/src/indev/lv_indev.c
+++ b/src/indev/lv_indev.c
@@ -1367,7 +1367,10 @@ static void indev_proc_release(lv_indev_t * indev)
 
     if(indev->wait_until_release) {
         lv_obj_send_event(indev->pointer.act_obj, LV_EVENT_PRESS_LOST, indev_act);
-        if(indev_reset_check(indev)) return;
+        if(indev_reset_check(indev)) {
+            indev->wait_until_release = 0;
+            return;
+        }
 
         indev->pointer.act_obj  = NULL;
         indev->pointer.last_obj = NULL;
--- a/tests/src/test_cases/test_indev_wait_release.c
+++ b/tests/src/test_cases/test_indev_wait_release.c
@@ -1,0 +1,59 @@
+#if LV_BUILD_TEST
+#include "../lvgl.h"
+#include "../lv_test_indev.h"
+#include "unity/unity.h"
+
+void setUp(void)
+{
+    /* Function run before every test */
+}
+
+void tearDown(void)
+{
+    /* Function run after every test */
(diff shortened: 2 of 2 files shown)


liamHowatt (commented):
Thanks. Just wondering, any reason not to have the `indev->wait_until_release = 0;` inside `indev_reset_core`?

yimoxiao:
> Thanks. Just wondering, any reason not to have the `indev->wait_until_release = 0;` inside `indev_reset_core`?

Because indev reset may be used in various parts of lvgl, for example, when processing the pressed state of pointer input devices, a new object was found the previous was lost, which will trigger indev reset. But if `indev->wait_until_release` is reset to 0 at this time, then the new object will be pressed. I think the idea is unreasonable, which can be seen from the design idea of `​​LV_OBJ_FLAG_PRESS_LOCK` and the fact that `indev_proc_press` will first judge wait_until_release.
There will be many places in the project that call indev reset, which will also cause some unreasonable scenes.

FASTSHIFT:
Nice catch! Can you add a testcase to close this loophole?

kisvegabor:
Nice catch indeed! A test would be really welcome. It could serve as a good starting point for the future contributors of the indevs. 

You can use [this API](https://github.com/lvgl/lvgl/blob/master/tests/src/lv_test_indev.h) to emulate clicks.

yimoxiao:
I just submitted a test case. Could you let me know if this implementation seems reasonable?

FASTSHIFT (approved):

uLipe (approved):

PR #7641, opened 2025-01-21
fix(example):  add cast for chart example

DevAnyKR:
void * requires casting

### Notes

### DIFF
--- a/examples/widgets/chart/lv_example_chart_4.c
+++ b/examples/widgets/chart/lv_example_chart_4.c
@@ -5,7 +5,7 @@
 static void draw_event_cb(lv_event_t * e)
 {
     lv_draw_task_t * draw_task = lv_event_get_draw_task(e);
-    lv_draw_dsc_base_t * base_dsc = lv_draw_task_get_draw_dsc(draw_task);
+    lv_draw_dsc_base_t * base_dsc = (lv_draw_dsc_base_t *)lv_draw_task_get_draw_dsc(draw_task);
 
     if(base_dsc->part != LV_PART_ITEMS) {
         return;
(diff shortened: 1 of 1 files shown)


XuNeo:
Is it really needed to cast from `void *`?

DevAnyKR:
> Is it really needed to cast from `void *`?

Cannot initialize entity of type "lv_draw_dsc_base_t *" using value of type "void *". C/C++(144)

I don't know. ask MS.

XuNeo:
I mean that what is the compile warning or error, which compiler is used. I believe there are lots of converting from void * to other type of pointers.

If there's really need, all of them should be fixed.

DevAnyKR:
That error is the output that appears when compiling with msvc.
As you said, all of them should be fixed, but all issues related to this have been closed automatically.

AndreCostaaa:
> Is it really needed to cast from `void *`?

The cast is necessary when using a C++ compiler as an implicit void* cast is not valid C++

FASTSHIFT:
In C, the void* type does not require a cast, see [here](https://stackoverflow.com/questions/605845/should-i-cast-the-result-of-malloc-in-c/605858#605858). My guess is that you copied the example code into a cpp file and compiled it. If so, cast it only in your application, but the example written as C does not need to be modified.

uLipe:
I think even though is not necessary having a cast removes any possibillity of build errors where this warning may be promoted to an error -Werror cases, which may depend on compiler, using this example in a C++ env or even if a new C standard treats this as warning.

In any case there is no bad thing on doing this cast and it may save some hours of silent build erros failure.

What do you think folks?

AndreCostaaa:
I agree with @uLipe. There's no harm in adding a cast and making a `.c` file compilable with a cpp compiler
It also adds more clarity, as to what the function returns and may be easier to spot bugs in the future

uLipe (approved):

PR #7642, opened 2025-01-21
Patch prvCheckMutexInit() in lv_freertos.c

trbljump:
When using lvgl with LV_OS_FREERTOS and SMP on a rp2040, the `configASSERT(portGET_CRITICAL_NESTING_COUNT() == 0 )` assertion was failing. As per https://github.com/FreeRTOS/FreeRTOS-Kernel/issues/1187, replacing `_enter_critical()` / `_exit_critical()` with `vTaskSuspendAll()` / `xTaskResumeAll()` fixes the crash.

Note that `prvCheckMutexInit()` is also called from `lv_mutex_lock_isr()` and thus `lv_lock_isr()`which is a problem but I didn't find any reference to the latter in the current codebase. Can it really happen that a mutex is used from an ISR without having been created beforehand (I know, I know...).


trbljump:
Please abandon this PR. I mistakenly created it from release/9.2 and this was not my intention, sorry for the inconvenience, I'll resubmit it to master later.

PR #7655, opened 2025-01-24
feat(demos): add TTF and span text benchmark

kar98knb:
Add testcases for both TTF text and span text in benchmark, both have bitmap rendering mode and outline rendering mode.

### Notes


liamHowatt (commented):
Thanks. The benchmark is often run on platforms without filesystems or TTF enabled. Adding benchmark scenes that depend of the filesystem or a lib like TinyTTF can be considered, but it should be a configuration option so the benchmark can continue to run on e.g. MCUs without SD cards. cc @kisvegabor

Kconfig
```diff
@@ -956,6 +956,10 @@ menu "LVGL configuration"
 		config LV_USE_FONT_PLACEHOLDER
 			bool "Enable drawing placeholders when glyph dsc is not found"
 			default y
+
+		config LV_TEST_FONT_PATH
```
FASTSHIFT: Since this configuration is only used by the benchmark, please move this configuration below the `LV_USE_DEMO_BENCHMARK` configuration, rename it to `LV_DEMO_BENCHMARK_FREETYPE_FONT_PATH` and add the dependency `LV_USE_FREETYPE`.

FASTSHIFT (changes requested):

demos/benchmark/lv_demo_benchmark.c
```diff
@@ -75,50 +98,104 @@ static void next_scene_timer_cb(lv_timer_t * timer);
...
 
-static void moving_wallpaper_cb(void)
+static void moving_wallpaper_cb(benchmark_context_t * context)
 {
     lv_obj_set_style_pad_all(lv_screen_active(), 0, 0);
-    LV_IMAGE_DECLARE(img_benchmark_lvgl_logo_rgb);
+    LV_IMAGE_DECLARE(img_benchmark_lvgl_logo_argb);
 
-    lv_obj_t * img = lv_image_create(lv_screen_active());
+    lv_obj_t * img = lv_img_create(lv_screen_active());
```
FASTSHIFT: Why changed?

demos/benchmark/lv_demo_benchmark.c
```diff
@@ -75,50 +98,104 @@ static void next_scene_timer_cb(lv_timer_t * timer);
...
     lv_obj_set_size(img, lv_pct(150), lv_pct(150));
-    lv_image_set_src(img, &img_benchmark_lvgl_logo_rgb);
+    lv_image_set_src(img, &img_benchmark_lvgl_logo_argb);
```
FASTSHIFT: Why change image source?

demos/benchmark/lv_demo_benchmark.c
```diff
@@ -75,50 +98,104 @@ static void next_scene_timer_cb(lv_timer_t * timer);
...
 static void summary_create(void);
 
-static void rnd_reset(void);
-static int32_t rnd_next(int32_t min, int32_t max);
-static lv_color_t rnd_color(void);
+static void rnd_reset(benchmark_context_t * context);
+static int32_t rnd_next(int32_t min, int32_t max, benchmark_context_t * context);
 static void shake_anim_y_cb(void * var, int32_t v);
-static void fall_anim(lv_obj_t * obj, int32_t y_max);
+static void fall_anim(lv_obj_t * obj, int32_t y_max, benchmark_context_t * context);
```
FASTSHIFT: It is recommended to put context in the first parameter.

demos/benchmark/lv_demo_benchmark.c
```diff
@@ -490,9 +584,64 @@ static uint32_t rnd_act;
  *   GLOBAL FUNCTIONS
  **********************/
 
+static void benchmark_context_init(benchmark_context_t * context)
```
FASTSHIFT: It is recommended to put the memory allocation and initialization of the context into one function to keep the dynamically allocated memory consistent with the life cycle of the object itself:
```
static benchmark_context_t * benchmark_context_create(void)
```

demos/benchmark/lv_demo_benchmark.c
```diff
@@ -490,9 +584,64 @@ static uint32_t rnd_act;
...
+    lv_obj_set_style_bg_opa(context->label_perf, LV_OPA_COVER, 0);
+    lv_obj_set_style_bg_color(context->label_perf, lv_color_white(), 0);
+    lv_obj_set_style_text_color(context->label_perf, lv_color_black(), 0);
+    lv_obj_set_width(context->label_perf, lv_pct(100));
+#if LV_USE_PERF_MONITOR_LOG_MODE
+    lv_obj_add_flag(context->label_perf, LV_OBJ_FLAG_HIDDEN);
+#endif
+}
+
+static void benchmark_context_deinit(benchmark_context_t * context)
```
FASTSHIFT, line 661: Suggested name:
```
static void benchmark_context_destroy(benchmark_context_t * context)
```
Release the memory pointed to by the context pointer in this function.

demos/benchmark/lv_demo_benchmark.c
```diff
@@ -490,9 +584,64 @@ static uint32_t rnd_act;
...
+{
+#if LV_USE_FREETYPE
```
FASTSHIFT, line 663: No need to add macro judgment.

demos/benchmark/lv_demo_benchmark.c
```diff
@@ -34,6 +34,15 @@
     #warning "It's recommended to have at least 128kB RAM for the benchmark"
 #endif
 
+#if LV_USE_TINY_TTF
```
FASTSHIFT: TinyTTF can also test file mode and non-file mode.

demos/benchmark/lv_demo_benchmark.c
```diff
@@ -47,13 +56,27 @@
 #define FALL_HEIGHT     80
 #define PAD_BASIC       8
 
+
+#define LV_TEST_FONT_STRING_1 "法律之前人人平等,并有权享受法律的平等保护,不受任何歧视。人人有权享受平等保护,以免受违反本宣言的任何歧视行为以及煽动这种歧视的任何行为之害"
```
FASTSHIFT: `LV_TEST_FONT_STRING_CHINESE`
It is recommended to use meaningful names, and we may introduce tests in other languages, such as Arabic, in the future.

demos/benchmark/lv_demo_benchmark.c
```diff
@@ -47,13 +56,27 @@
...
+#define LV_TEST_FONT_STRING_2 "No one shall be subjected to arbitrary arrest, detention or exile. Everyone is entitled in full equality to a fair and public hearing by an independent and impartial tribunal, in the determination of his rights and obligations and of any criminal charge against him. No one shall be subjected to arbitrary interference with his privacy, family, home or correspondence, nor to attacks upon his honour and reputation. Everyone has the right to the protection of the law against such interference or attacks."
```
FASTSHIFT: `LV_TEST_FONT_STRING_ENGLISH`

demos/benchmark/lv_demo_benchmark.c
```diff
@@ -548,34 +688,39 @@ static void load_scene(uint32_t scene)
...
+        lv_timer_set_period(timer, scenes[context->scene_act].scene_time);
     }
 }
 
 #if LV_USE_PERF_MONITOR
 static void sysmon_perf_observer_cb(lv_observer_t * observer, lv_subject_t * subject)
 {
     const lv_sysmon_perf_info_t * info = lv_subject_get_pointer(subject);
-    char scene_name[64];
+    benchmark_context_t * context = (benchmark_context_t *)(observer->user_data);
```
FASTSHIFT: No cast is required for `void*` types, and please use functions to access object data.
```suggestion
    benchmark_context_t * context = lv_observer_get_user_data(observer);
```

demos/benchmark/lv_demo_benchmark.c
```diff
@@ -490,9 +584,64 @@ static uint32_t rnd_act;
...
+        lv_tiny_ttf_destroy(context->font_tinyttf);
+    }
+#endif
+    lv_obj_delete(context->label_perf);
+}
+
 void lv_demo_benchmark(void)
 {
-    scene_act = 0;
+    benchmark_context_t * context = (benchmark_context_t *)lv_malloc(sizeof(benchmark_context_t));
```
FASTSHIFT: Use `lv_malloc_zeroed`.

demos/benchmark/lv_demo_benchmark.c
```diff
@@ -830,10 +970,10 @@ static lv_obj_t * card_create(void)
     lv_obj_set_size(panel, 270, 120);
     lv_obj_set_style_pad_all(panel, 8, 0);
 
-    LV_IMAGE_DECLARE(img_benchmark_avatar);
+    LV_IMG_DECLARE(img_transform_avatar_15);
```
FASTSHIFT: Why changed?

demos/benchmark/lv_demo_benchmark.c
```diff
@@ -854,12 +994,39 @@ static lv_obj_t * card_create(void)
     return panel;
 }
 
-static void rnd_reset(void)
+#if LV_USE_FREETYPE || TEST_TINY_TTF
```
FASTSHIFT: Span widget can also be tested using the default font, without relying on freetype and tinyttf.

kisvegabor:
I agree with @liamHowatt that it doesn't really test LVGL itself, but rather a 3rd party library. I rather suggest adding `lv_example_tinyttf_3()` to perform a benchmark specifically for TinyTTF.

FASTSHIFT:
> I agree with @liamHowatt that it doesn't really test LVGL itself, but rather a 3rd party library. I rather suggest adding `lv_example_tinyttf_3()` to perform a benchmark specifically for TinyTTF.

On the contrary, I think it is very meaningful to add 3rd lib to the benchmark. Because in actual business scenarios, a large number of third-party libraries are used, such as freetype and libpng. If these tests are added to the benchmark, it can be closer to the actual business scenario and make the data more representative.

In addition, we can also consider adding performance tests in CI to monitor the performance changes of each PR on LVGL. As LVGL functions become more and more complex, the problem of performance regression becomes more and more obvious, and it is very necessary to introduce a performance monitoring mechanism.

demos/benchmark/lv_demo_benchmark.c
kar98knb: The deinit functions lv_freetype_font_delete() and lv_tiny_ttf_destory() are not supported if don't enable these macros, and will lead to compile error.

kisvegabor:
I see your point but I'd approach it differently. If we make the 3rd party libs mandatory in the benchmark it might not be possible to run them on all the systems. E.g. there is no file system access, no memory for ThorVG, cannot link FreeType, etc. 
Also there are multiple 3rd party libs for the same task:
- FreeType, TinyTTF
- ThorVG, VGLite
- libpng, lodepng

Which one should be benchmarked? 

So I'd rather say to have some separate targeted benchmarks:
- PNG decoding (or lossless ARGB images in general, so that RLE, LZ4 and C array can be compared too)
- JPG decoding (or lossy RGB in genral)
- Vector fonts
- Video playback
- Gif
- Anything else?

This we can benchmark and compare the different solution.

> In addition, we can also consider adding performance tests in CI to monitor the performance changes of each PR on LVGL. As LVGL functions become more and more complex, the problem of performance regression becomes more and more obvious, and it is very necessary to introduce a performance monitoring mechanism.


There is a long running PR about it. I was talking to the guys behind it  and they will pick it up again soon: https://github.com/lvgl/lvgl/pull/6486

Kconfig
```diff
@@ -2007,6 +2007,12 @@ menu "LVGL configuration"
 			select LV_FONT_MONTSERRAT_24
 			select LV_FONT_MONTSERRAT_26
 			select LV_USE_DEMO_WIDGETS
+
+		config LV_DEMO_BENCHMARK_FREETYPE_FONT_PATH
+			string "Path to font file which is used in benchmark test"
+			default "./lvgl/tests/src/test_files/fonts/noto/NotoSansSC-Regular.ttf"
+			depends on LV_USE_DEMO_BENCHMARK && (LV_USE_FREETYPE || LV_USE_TINY_TTF)
```
FASTSHIFT: tinyttf cannot use the `./` path of freetype. You can add an independent path configuration `LV_DEMO_BENCHMARK_TINY_TTF_FONT_PATH` to tinyttf.

lv_conf_template.h
```diff
@@ -603,6 +603,9 @@
 /** Enable drawing placeholders when glyph dsc is not found. */
 #define LV_USE_FONT_PLACEHOLDER 1
 
+/*Path to font file which is used in benchmark test*/
+#define LV_TEST_FONT_PATH "./lvgl/tests/src/test_files/fonts/noto/NotoSansSC-Regular.ttf"
```
FASTSHIFT: `lv_conf_template` needs to be synchronized with `Kconfig`, `LV_DEMO_BENCHMARK_FREETYPE_FONT_PATH`
`LV_DEMO_BENCHMARK_TINY_TTF_FONT_PATH`

FASTSHIFT (changes requested):

FASTSHIFT:
> I see your point but I'd approach it differently. If we make the 3rd party libs mandatory in the benchmark it might not be possible to run them on all the systems. E.g. there is no file system access, no memory for ThorVG, cannot link FreeType, etc. Also there are multiple 3rd party libs for the same task:
> 
> * FreeType, TinyTTF
> * ThorVG, VGLite
> * libpng, lodepng
> 
> Which one should be benchmarked?

We don’t have to force the use of third-party libraries in the benchmark. My idea is to only do verification when the third-party library is enabled, otherwise just skip the test. Users only need to care about the third-party libraries they use.

demos/benchmark/lv_demo_benchmark.c
kar98knb: Is there a need to test span widget without freetype and tinyttf?

demos/benchmark/lv_demo_benchmark.c
kar98knb: Is this necessary?

FASTSHIFT (dismissed):

uLipe (dismissed):

liamHowatt (commented):
I tried TinyTTF. Can it be centered?

https://github.com/user-attachments/assets/c37a3a1e-6266-4c07-b708-f85b76e45996

kar98knb:
> I tried TinyTTF. Can it be centered?
> 
>  simplescreenrecorder-2025-02-12_14.20.04.mp4

Done.

FASTSHIFT (dismissed):

FASTSHIFT (dismissed):

lv_conf_template.h
```diff
@@ -1264,6 +1264,14 @@
 /** Benchmark your system */
 #define LV_USE_DEMO_BENCHMARK 0
 
+#if LV_USE_DEMO_BENCHMARK
+    /** Path to FreeType font file which is used in benchmark test */
+    #define LV_DEMO_BENCHMARK_FREETYPE_FONT_PATH "./lvgl/demos/benchmark/assets/NotoSansSC-Regular.ttf"
```
kisvegabor, line 1306: We don't allow changing other assets either. I think it should be fixed without a config option too.

kisvegabor (changes requested):
Alright, I'm with this approach, however while I was testing it I have found some strange issues.

TinyTTF test are not working for me. Either crashing or failing in exotic ways. Do they work on your end? 

If I close the SDL window while running FreeType tests, I get these logs:
```
[Warn]	(0.000, +0)	 drop_all_cb: entry (0x562f67a32a90) is still referenced (1) lv_cache_lru_rb.c:401
[Warn]	(0.000, +0)	 drop_all_cb: entry (0x562f67a32800) is still referenced (1) lv_cache_lru_rb.c:401
[Warn]	(0.000, +0)	 drop_all_cb: 2 entries are still referenced lv_cache_lru_rb.c:406
```

I believe it's triggered from `lv_deinit` from `lv_freetype_uninit` because the fonts are not freed. We should add some DELETED events to the benchmark's main container to free the context on deletion.

lv_conf_template.h
kar98knb: TinyTTF is working for me, though it comsumes CPU(my device is a M1 MacBook Air):

https://github.com/user-attachments/assets/7fbcb653-3461-434f-9e50-8b22b340ed8a

The fonts and other resources(if any) are freed in the function `benchmark_context_deinit`, line 676, which will be called after all scenes are finished.

Sorry I didn't get your last point. Though in our scenario, the path should be a configuration since we could use our own font.

kar98knb:
remove abandoned

kisvegabor:
Sorry for the long delay, it seems I missed this PR.  Next time please re-request review, so that I'll be notified about it.

image

---

I still believe that we should have these font/text tests as examples to keep the benchmark unified. If the benchmark is not the same for all devices it will be harder to compare the devices. Now the "All scenes avg" represents the overall performance and it's a good way to quickly compare boards. However, if some of the devices uses FreeType, and some not, the end result might vary a lot. 

I also think that there are too many libraries and cases to compare. (See below). 

So in my opinion we should add simple very targeted functions the benchmark these. As these should be very simple the users can easily modify them if needed. 

```c
/*Render  raw, PNG, JPG, GIF, LZ4, RLE4, I8, etc images from file and data and measure the rendering time*/
lv_example_benchmark_image_decoders_from_file(); 
lv_example_benchmark_image_decoders_from_data();
```
```c
/*Render label and spans from built-in fonts (compressed, bpp 4/8, etc), binfont, tinyttf, freetype, etc*/
lv_example_benchmark_text_from_file(); 
lv_example_benchmark_text_from_data();
```

FASTSHIFT:
OK, let's use the separate benchmark design pattern. @kar98knb

kar98knb:
ok

PR #7659, opened 2025-01-26
fix(chart): fix last point not drawn in scatter chart

niklasf:
With the loop from 0 to chart->point_cnt (exclusive), a line is drawn to the last point, but not the point itself. Fix off-by-one error.


niklasf:
Oops. Fixed typo in branch name. New pr: #7665.

PR #7665, opened 2025-01-27
fix(chart): fix last point not drawn in scatter chart

niklasf:
With the loop from 0 to chart->point_cnt (exclusive), a line is drawn to the last point, but not the point itself. Fix off-by-one error.


uLipe (commented):
General comment, is it possible to put a test case for it?

niklasf:
Test via reference screenshot should be possible. Will do.

liamHowatt (commented):
LGTM

> `/*Don't draw the first point. A second point is also required to draw the line*/`

Seems odd to not draw the first point

niklasf:
Done.

> Seems odd to not draw the first point

Clarified that comment. It's really about not drawing a line *to* the first point.

Also took the opportunity to adjust `id2` of the draw tasks to match those of other chart types.

Referene image before the fix (note there's no point in the top right corner):

chart_scatter

kisvegabor (approved):
:+1:

liamHowatt (approved):

PR #7666, opened 2025-01-27
fix(bar): fix bar indicator length error

### LINKED ISSUE #7664: Reversed Bars have a small indicator when at min value
### LVGL version

v9.2.2

### Platform

Custom Linux distribution on ATSAMA5D29, forked off lv_port_linux. 

### What happened?

Two bars are created one from 0-100 and one from -100-0. In order to make them both empty, I set the first one to 0 and the second to -100. The first bar shows no indicator as it should, but the second one shows a pixel wide indicator at the top of the bar. I believe this is a bug as it should register it as 0 width and not draw it. Looking into it, it seems that the check on lv_bar.c:428 doesn't work properly for reversed bars. Trying to find out why or if that's really the issue. 

### How to reproduce?

```c
neg_power_bar = lv_bar_create(lv_screen_active());
/* add styles, etc */
lv_bar_set_range(neg_power_bar, 0, -100); 
lv_bar_set_value(neg_power_bar, -100, LV_ANIM_OFF); /* shows a small indicator at top of bar*/
```

sbesse-nam:
found that length is always 2 more than it should be for reverse drawn vertical bars. This change doesn't seem to break anything else, and it doesn't accumulate over multiple draws. 

Fixes #7664 

### Notes


uLipe (commented):
One general comment, is there a way to test against the horizontal as well? Also is it possible to put a test case that evidences the fix resolves it? Thank you!

sbesse-nam:
Hi, I see that it failed the bar test. I don't know what this consists of, so maybe you could provide some input as to what the problem was. 

Here's the test case that I was using 

````
static lv_obj_t *neg_power_bar;
neg_power_bar = lv_bar_create(lv_screen_active());
lv_obj_set_size(neg_power_bar, 10, 100);
lv_bar_set_range(neg_power_bar, 100, 0); 
lv_obj_align(neg_power_bar, LV_ALIGN_LEFT_MID, 5, 55); 
lv_bar_set_value(neg_power_bar, 0, LV_ANIM_OFF);
usleep(10000);
lv_bar_set_value(neg_power_bar, 100, LV_ANIM_ON);
`````

before my change, there is a small indicator at the top of the bar when the value is 0. After the change that indicator properly goes away when the value is 0. After the change the bar also properly fills up from top to bottom, which is the case before the change as well.

**Comment by liamHowatt** [comment]
It's a screenshot test that ensures that test case UIs continue to produce an expected resulting image.

https://github.com/lvgl/lvgl/blob/0f67eb67cfb545afde03bb602b002c0b5464e1f0/tests/ref_imgs/widgets/bar_1.png

https://github.com/lvgl/lvgl/blob/0f67eb67cfb545afde03bb602b002c0b5464e1f0/tests/src/test_cases/widgets/test_bar.c#L196-L206

You can run the tests locally and replace the existing screenshot with the new one generated after your new change, if your change produces a more correct bar widget.

You can add a new test case which tests vertical reversed bars set to the `-100` value that causes the issue.

**Comment by sbesse-nam** [comment]
Analyzing the screenshot in GIMP I measured the indicator lengths on both. I found 188px for the top to bottom indicator and 184px for the bottom to top indicator. This is consistent with my findings that the top to bottom indicator is always slightly longer than it should be. No new test is needed since it isn't caused by the -100 value, but is shared by all reversed vertical bars. I only have LVGL setup for cross compilation at the moment, but I will try to get a setup going on windows and generate a new test image, verifying that the indicator lengths match exactly for the vertical bars. As a side note, both horizontal bar indicators measured 185px, so I believe this problem is unique to reversed vertical bars. The discrepancy between horizontal and non-reversed vertical indicator length is probably slightly difference antialiasing.

**Comment by sbesse-nam** [comment]
Sorry have run out of time recently. If someone else has a setup and can generate new test cases that would solve this testing issue. If not, I will revisit when I have a bit more time. Thanks

**Comment by kisvegabor** [comment]
@liamHowatt, @uLipe if you have time, could you take over this PR?

**Comment by liamHowatt** [comment]
I do not understand it at all. I think there must be a better way to fix it, but I'm not sure how, so this is great for now!

Here is what the new test image looks when sbesse-nam's fix is not applied. The red arrows are fixed by this PR. I have to say there also appears to be something wrong with the filled reversed horizontal bar.

Untitled drawing

**Comment by liamHowatt** [comment]
There's a complex vg lite failure. Will revisit it.

**Comment by liamHowatt** [comment]
Actually the CI passed. It must have been something wrong with my local tests. Ready for review.

**Review by uLipe — DISMISSED** [review]

**Review by kisvegabor — DISMISSED** [review]

File: src/widgets/bar/lv_bar.c [hunk 1]
```diff
@@ -459,6 +459,9 @@ static void draw_indic(lv_event_t * e)
     }
 
     /*Do not draw a zero length indicator but at least call the draw task event*/
+    /* small fix for reversed vertical indicator, no idea if this happens with horizontal as well */
```
kisvegabor: Sorry for the long delay. 

Please check it with horizontal as well to have a fix the covers both cases.

sbesse-nam:
Re-synced with master to pass tests. 

Doing testing and measuring I'm concluding that the horizontal bar appearance is a separate issue. The bar indicator length is correct for the filled bars, 300px, but they look off because it appears that the radius isn't being properly applied when the bar is full, even though the radius is correctly applied when the horizontal bar is partially full. I would recommend merging and closing this issue, as it fixes the indicator length problem with reversed vertical bars, and if there's interest, look at the radius issue with horizontal bars separately.

src/widgets/bar/lv_bar.c
kisvegabor: Thank you! Please remove this comment.

liamHowatt:
I believe I've completely fixed the problem for all cases here: https://github.com/liamHowatt/lvgl/commit/d8ebe76b54229aaa3f8f60d57847d69eb57f1e58

If you agree, I'll push it to this branch.

kisvegabor:
@liamHowatt 
Great! It seems cleaner. Please open a new PR with your suggested changes.

liamHowatt:
#8199

kisvegabor:
Closed in in favor of https://github.com/lvgl/lvgl/pull/8200

PR #7673, opened 2025-01-28
perf(draw): optimize lv_draw_label() with text_length for text_local

niklasf:
Introduce `lv_strnlen()` and `lv_strndup()`, mirroring the standard C library.

When using `lv_draw_label()` with `text_length` and `text_local` to draw a slice of a larger string, copy only the required part.

Depending on the size of the larger string, this can be anywhere from insignificant to a major improvement.

I think the additional standard library functions are useful, but it would also be possible to manually malloc and copy diretcly in the draw function.

### DIFF
--- a/src/stdlib/builtin/lv_string_builtin.c
+++ b/src/stdlib/builtin/lv_string_builtin.c
@@ -191,6 +191,14 @@ size_t lv_strlen(const char * str)
     return i;
 }
 
+size_t lv_strnlen(const char * str, size_t max_len)
+{
+    size_t i = 0;
+    while(i < max_len && str[i]) i++;
+
+    return i;
+}
+
 size_t lv_strlcpy(char * dst, const char * src, size_t dst_size)
 {
     size_t i = 0;
@@ -255,6 +263,17 @@ char * lv_strdup(const char * src)
     if(dst == NULL) return NULL;
 
     lv_memcpy(dst, src, len); /*memcpy is faster than strncpy when length is known*/
+    return dst;
+}
+
+char * lv_strndup(const char * src, size_t max_len)
+{
+    size_t len = lv_strnlen(src, max_len);
+    char * dst = lv_malloc(len + 1);
+    if(dst == NULL) return NULL;
+
+    lv_memcpy(dst, src, len);
+    dst[len] = '\0';
     return dst;
 }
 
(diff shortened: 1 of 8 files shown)


src/stdlib/builtin/lv_string_builtin.c
```diff
@@ -191,6 +191,14 @@ size_t lv_strlen(const char * str)
     return i;
 }
 
+size_t lv_strnlen(const char * str, size_t max_len)
+{
```
uLipe, line 195: Could you track for NULL here, it can be an assert or log it will make the debug easy if the things here get wrong

src/stdlib/builtin/lv_string_builtin.c
```diff
@@ -258,6 +266,17 @@ char * lv_strdup(const char * src)
     return dst;
 }
 
+char * lv_strndup(const char * src, size_t max_len)
```
uLipe, line 269: Also here, it returns null silently when malloc fails

src/stdlib/clib/lv_string_clib.c
```diff
@@ -102,6 +107,17 @@ char * lv_strdup(const char * src)
     return dst;
 }
 
+char * lv_strndup(const char * src, size_t max_len)
```
uLipe, line 110: also here

uLipe (changes requested):
Nice work!

I just have a couple of nits.

src/stdlib/builtin/lv_string_builtin.c
niklasf: ~~I am a bit hesitant, as the assert handler and logging machinery may itself use the stdlib and at least the other stdlib functions don't do it.~~ I don't think I can do it in the standard library without creating a circular dependency, because the assert handler/logging machinery may itself use the stdlib. Maybe I should just add assertions to the call in `lv_draw_label.c`?

liamHowatt (dismissed):

src/stdlib/builtin/lv_string_builtin.c
niklasf: Updated my commit with that.

liamHowatt (approved):

liamHowatt:
@uLipe I think we should try to make any functions that mimic C library functions as compliant as possible. i.e., the caller handles failures instead of crashing in str(n)dup, and invalid input is UB.

uLipe (approved):

PR #7522, opened 2024-12-26
fix(scroll): scrolling animation and finger dragging conflict during continuous sliding, causing flickering

C-Ackerman:
Continuous and rapid sliding will conflict with the scrolling animation, causing scrolling to flicker，the scrolling animation needs to be stopped before dragging the finger

image

Modify as follows: Increase animation duration to increase recurrence probability
`/src/core/lv_obj_scroll.c
+#define SCROLL_ANIM_TIME_MIN    1000    /*ms*/
+#define SCROLL_ANIM_TIME_MAX    1500    /*ms*/`

Example Code：
`void lv_example_flex_4(void)
{

    lv_obj_t * cont = lv_obj_create(lv_screen_active());
    lv_obj_set_size(cont, 300, 220);
    lv_obj_center(cont);
    lv_obj_set_flex_flow(cont, LV_FLEX_FLOW_COLUMN_REVERSE);
    // lv_obj_clear_flag(cont, LV_OBJ_FLAG_SCROLL_ELASTIC);
    lv_obj_add_flag(cont, LV_OBJ_FLAG_SNAPPABLE);
    lv_obj_set_scroll_snap_y(cont, LV_SCROLL_SNAP_END);
    lv_obj_set_style_bg_color(cont, lv_color_hex(0x9F79EE), 0);

    uint32_t i;
    for(i = 0; i < 6; i++) {
        lv_obj_t * obj = lv_obj_create(cont);
        lv_obj_set_size(obj, 100, 50);
        lv_obj_remove_flag(obj, LV_OBJ_FLAG_SCROLLABLE);
        lv_obj_set_style_bg_color(obj, lv_color_hex(0x90EE90), 0);

        lv_obj_t * label = lv_label_create(obj);
        lv_label_set_text_fmt(label, "Item: %"LV_PRIu32, i);
        lv_obj_center(label);
    }
}`

https://github.com/user-attachments/assets/edbb6a48-3476-423a-9802-60096ea27894

PR #7674, opened 2025-01-28
chore(obj): make lv_obj_get_scroll_{left,right,top,bottom}() const

niklasf:


liamHowatt (commented):
I think this convention could be avoided since it could lead to code-churn, i.e., a function that changes the const-ness of its parameters will effect the API and also cause any callers of it to maybe change their own parameters from const back to mutable if it later goes from const back to mutable. Also casting away const in the function because "we know it won't change", but it actually could. See https://github.com/lvgl/lvgl/pull/7522/files#diff-990eed2994668918be796257f40d31dbed25a57146ef4c586a5fc8dc2ecd03d5R445-R449

I don't think it's useful to tell the caller that the obj they pass will not be changed, since an `lv_obj_t` should never live in read-only memory or have its fields become frozen at any point.

I appreciate the contribution as always, though.

niklasf:
My motivation here is exactly about that guarantee while writing a custom widget: Now and for future 9.x versions, can I assume that these particular functions do not do anything that needs non-const `lv_obj_t*`, like recursively submiting events when I am using them from an event handler.

I think that's useful information, but, as you're saying, only if it does not change and cause churn. So I understand if we don't want to commit to const-ness for these functions.

liamHowatt (approved):
I just realized it's already being done in many getters and it hasn't caused code-churn, and these functions are good candidates too. I never noticed it before. Sorry for the friction.

kisvegabor (approved):
Thank you! I agree that it's a good practice to use `const` where possible.

PR #7700, opened 2025-02-02
feat(build): add new build system

### LINKED ISSUE #7223: Rework and unify CMake

AndreCostaaa:
Fixes #7223 

1. No more globbing
2. Config is loaded from `.config` file generated using Kconfig
3. Each subfolder generates a static library based on the imported config


CMakeLists.txt
```diff
@@ -20,7 +20,24 @@ elseif(ZEPHYR_BASE)
...
 else()
-  include(${CMAKE_CURRENT_LIST_DIR}/env_support/cmake/custom.cmake)
+  include(cmake/top.cmake)
+  add_subdirectory(libs)
+  add_subdirectory(src)
+  add_subdirectory(demos)
+  add_subdirectory(examples)
+
+
+    file(GLOB_RECURSE EXAMPLE_SOURCES ${LVGL_ROOT_DIR}/examples/*.c)
```
uLipe, line 30: Why do we need to keep the globbing with the examples?

uLipe (commented):
Nice progress!

Just had a first pass.

CMakeLists.txt
AndreCostaaa: We don't, just haven't gotten around to adding the CMakelists to examples
I am trying to compile a demo for now after all the changes

AndreCostaaa:
Status: Currently the conditional compilation is working but the C files that are wrapped around #if XXX #endif don't actually get compiled as I'm missing the lv_conf.h file

I don't want to remove these for now if we want to still be able to compile using the old way + the config passed through still needs to be accessible by the code

AndreCostaaa:
Closing this now for 3 reasons

1. This won't work with arduino as it doesn't do conditional compiling of source files
2. A bit out of scope of #7223 as that issue is only meant to fix linking with external libraries, examples and demos
3. `lv_conf.h` must be kept so moving everything to `kconfig` isn't the way to go

PR #7743, opened 2025-02-10
feat(draw_vector): add partial rendering support

etag4048:
Feature that adds partial rendering to the ThorVG integration - this was needed to render custom shapes
in a memory constrained environment that could not accommodate a large draw buffer using the canvas.

This solution runs the existing example located in the `demos/vector_graphics` directory, but there is an issue with the pattern fill. 
To perform rendering over multiple passes a transform with a y_offset is applied on each pass.

### Notes
   Waiting on #7241 to get merged before adding a section on vector graphics.

- Add [Examples](https://github.com/lvgl/lvgl/tree/master/examples)
  The example is located in the `demos/vector_graphics` directory

### DIFF
--- a/demos/vector_graphic/lv_demo_vector_graphic.c
+++ b/demos/vector_graphic/lv_demo_vector_graphic.c
@@ -23,6 +23,21 @@
 /**********************
  *  STATIC PROTOTYPES
  **********************/
+
+/**********************
+ *  STATIC VARIABLES
+ **********************/
+static int is_partially_buffered = 0;
+static lv_draw_buf_t * draw_buf;
+lv_obj_t * demo_container;
+
+/**********************
+ *      MACROS
+ **********************/
+
+/**********************
+ *   STATIC FUNCTIONS
+ **********************/
 static void draw_pattern(lv_vector_dsc_t * ctx, lv_vector_path_t * path)
 {
     lv_vector_path_clear(path);
@@ -228,12 +243,24 @@ static void draw_vector(lv_layer_t * layer)
 
     draw_shapes(ctx, path);
     draw_lines(ctx, path);
-    draw_pattern(ctx, path);
+    if(is_partially_buffered == 0) {
+        /*TODO: Currently not supported in partial mode*/
+        draw_pattern(ctx, path);
+    }
     draw_radial_gradient(ctx, path);
     draw_gradient(ctx, path);
     draw_blend(ctx, path);
     draw_arc(ctx, path);
-    lv_draw_vector(ctx); // submit draw
+
+    if(is_partially_buffered == 1) {
+        lv_vector_dsc_set_draw_buf(ctx, draw_buf);
+        lv_draw_vector_partial(ctx, demo_container);
+
+    }
+    else {
+        lv_draw_vector(ctx); // submit draw
+    }
+
     lv_vector_path_delete(path);
     lv_vector_dsc_delete(ctx);
 }
@@ -241,7 +268,7 @@ static void draw_vector(lv_layer_t * layer)
 static void delete_event_cb(lv_event_t * e)
 {
     lv_obj_t * obj = lv_event_get_target(e);
-    lv_draw_buf_t * draw_buf = lv_canvas_get_draw_buf(obj);
+    draw_buf = lv_canvas_get_draw_buf(obj);
     lv_draw_buf_destroy(draw_buf);
 }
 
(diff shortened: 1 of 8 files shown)


etag4048:
The CI has failed due to leaks in the load svg stuff. But otherwise seems to be ready.

kisvegabor:
@etag4048 Fantastic! :tada: 
Just one note: In `lv_draw_sw_vector` we should detect if the target layer's color format is not supported by ThorVG (e.g. RGB565) and allocate a `draw_buf` automatically (if not allcated yet). The size can come from config like `LV_DRAW_SW_TEMP_BUF_SIZE` or so. 


@FASTSHIFT Look at [this](https://github.com/lvgl/lvgl/pull/7743/files#diff-d35784339390a3a5830b889e93d2fdf984f52f12d8bdd5282b4d86cec0150795R448)! ThorVG support un-alpha premultipled format. I think we should migrate to it. What do you think?

onecoolx:
I have a suggestion. I think partial mode should be invisible to developers. People should not care whether it is partial mode or not. So I think y_offset and x_offset should be in `lv_layer_t`. When we call lv_draw_vector, we can use these two offsets to translate when creating tvg_canvas in thorvg backend to draw the correct partial buffer. Similarly, draw_buf should also be created in thorvg's drawing code, after all, vglite does not need it.  In this way the code and API will be much simpler and cleaner.

onecoolx:
> @onecoolx I agree of making the API more uniform, but don't agree on partial mode being invisible to developers. I made this PR for a client that is using the NR5340 which has only 512K of RAM, So I would like to retain the possibility to configure the size of the buffer manually depending on the size of shape that needs to be drawn. I also need to be able to redraw only the changed part. Not quite sure what you mean with layer_t, is it an lv_obj_t backed by a SIMPLE layer ?

I mean the `lv_layer_t` data structure, the first parameter of each drawing function, if configured in partial mode, the `clip_area` in `lv_layer_t` will be the size of the partial buffer.

cc @kisvegabor

demos/vector_graphic/lv_demo_vector_graphic.c
```diff
@@ -228,20 +243,32 @@ static void draw_vector(lv_layer_t * layer)
 
     draw_shapes(ctx, path);
     draw_lines(ctx, path);
-    draw_pattern(ctx, path);
+    if(is_partially_buffered == 0) {
+        /*TODO: Currently not supported in partial mode*/
```
uLipe, line 247: Please replace the TODO with a relevant LOG call (a LOG_WARN should be sufficient)

demos/vector_graphic/lv_demo_vector_graphic.c
```diff
@@ -228,20 +243,32 @@ static void draw_vector(lv_layer_t * layer)
...
+        draw_pattern(ctx, path);
+    }
     draw_radial_gradient(ctx, path);
     draw_gradient(ctx, path);
     draw_blend(ctx, path);
     draw_arc(ctx, path);
-    lv_draw_vector(ctx); // submit draw
+
+    if(is_partially_buffered == 1) {
```
uLipe, line 255: nit: if it is bool please use true or false for readabillity.

demos/vector_graphic/lv_demo_vector_graphic.c
```diff
@@ -228,20 +243,32 @@ static void draw_vector(lv_layer_t * layer)
...
+        lv_vector_dsc_set_draw_buf(ctx, draw_buf);
+        lv_draw_vector_partial(ctx, demo_container);
+
+    }
+    else {
+        lv_draw_vector(ctx); // submit draw
```
uLipe, line 261: Comment is self explanatory, it can be removed.

src/draw/lv_draw_vector.c
```diff
@@ -778,27 +839,58 @@ void lv_vector_dsc_skew(lv_vector_dsc_t * dsc, float skew_x, float skew_y)
     lv_matrix_skew(&(dsc->current_dsc.matrix), skew_x, skew_y);
 }
 
-void lv_vector_for_each_destroy_tasks(lv_ll_t * task_list, vector_draw_task_cb cb, void * data)
+void lv_vector_dsc_set_blend_color(lv_vector_dsc_t * dsc, lv_color_t blend_color)
+{
+    if(dsc == NULL) {
```
uLipe, line 844: please add a LOG_ERROR when parameter validation fails on void type functions.

uLipe (changes requested):
Nice work! It is just my first pass, please check some small nits I left.

kisvegabor:
Why did you close it?

PR #7744, opened 2025-02-10
fix(nema): fix indexed image error

### LINKED ISSUE #7275: Visualization of indexed images seems to be corrupted on ST when NEMA GFX is enabled
### LVGL version

v9.3

### What happened?

I have a simple application showing an image on a STM32U5 board.  I've enabled `LTDC` and `NEMA_GFX` in `lv_conf.h` and I'm testing the visualization of an image exported with the LVGLImage.py tool using different image format.

If the image is exported using e.g. RGB565 format it is displayed correctly
IMG_0327

If the image is exported using e.g. I8 format it is not displayed correctly
IMG_0326

If I set `NEMA_GFX` acceleration to `0`, then both images are displayed correctly

### How to reproduce?

_No response_

william-spinelli:
Fixes #7275

The visualization of indexed images (in particular I8 images) is corrupt when NEMA acceleration is enabled.

The root cause of the problem seems to be related to the way in which NEMA indexes the color palette in I8 format.  See the following comments for a detailed explanation
* https://github.com/lvgl/lvgl/issues/7275#issuecomment-2644550771
* https://github.com/lvgl/lvgl/issues/7275#issuecomment-2643781026

Based on the analysis, NEMA_L8 format seems to be assuming that for example the color with index 0x03 is in position 0x30 of the palette array.

An option `--nemagfx` has been added to the conversion script.  When enabled the high and low nibble in the color indexes are swapped for I8 images

Additionally, LUT indexing is enabled in the commands to NEMA accelerator.  Otherwise NEMA will not take into consideration the color palette in the LUT

### Notes

### DIFF
--- a/scripts/LVGLImage.py
+++ b/scripts/LVGLImage.py
@@ -498,6 +498,7 @@ class LVGLImage:
         self.stride = 0  # default no valid stride value
         self.premultiplied = False
         self.rgb565_dither = False
+        self.nema_gfx = False
         self.set_data(cf, w, h, data)
 
     def __repr__(self) -> str:
@@ -840,7 +841,8 @@ class LVGLImage:
                  filename: str,
                  cf: ColorFormat = None,
                  background: int = 0x00_00_00,
-                 rgb565_dither=False):
+                 rgb565_dither=False,
+                 nema_gfx=False):
         """
         Create lvgl image from png file.
         If cf is none, used I1/2/4/8 based on palette size
@@ -848,6 +850,7 @@ class LVGLImage:
 
         self.background = background
         self.rgb565_dither = rgb565_dither
+        self.nema_gfx = nema_gfx
 
         if cf is None:  # guess cf from filename
             # split filename string and match with ColorFormat to check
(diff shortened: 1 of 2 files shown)


liamHowatt (dismissed):
Amazing that you were able to figure this out. Nice catch with the blending mode flag. I can't understand the reasoning behind swapping the endianness of a byte's nibbles. Maybe it's to do with the architecture and how it processes I4. Thanks for explaining everything in your comments.

Everything looks good to me. The `--nemagfx` flag is fine because it can be extended with other Nema quirks that apply to other image types if found later. My only critique is that `lv_nemagfx_cf_to_nema` could do what `lv_palette_cf_to_nema` does, if it gets corrected. Not critical.

cc @MarkIoanAmat

williams-one:
Thanks for the feedback @liamHowatt!  I also assume that the way in which they manipulate indexes is due to the structure of the hardware.  I have tried to find some references in the docs or on the internet, but I couldn't find anything...  For sure it would have been more reassuring 😃 

You are also right `lv_palette_cf_to_nema` is not needed and can be replaced with `lv_nemagfx_cf_to_nema`.
I will collect all the other feedbacks and then fix everything according to your suggestions.

MarkIoanAmat:
@williams-one Great work done here !! Indeed indexing manipulation is due to hardware limitations. Also, as manual mentions, you cannot use bilinear filtering in LUT type textures, as the filtering would be applied on the LUT indices (before the final texture composition) again due to HW structure.

williams-one:
@MarkIoanAmat Yes let's say that there was also a good component of luck in finding the solution😄
Now I will remove `lv_palette_cf_to_nema` and replace it with `lv_nemagfx_cf_to_nema` as suggested by @liamHowatt and redo all the tests

uLipe (approved):

liamHowatt (approved):
Thanks @MarkIoanAmat for looking at it! Thanks again @williams-one!

PR #7745, opened 2025-02-10
Further XML related updates

kisvegabor:
### Notes


kisvegabor:
@liamHowatt @uLipe I've create meaningful commits to make it easier to review.

src/others/xml/parsers/lv_xml_buttonmatrix_parser.c
```diff
@@ -0,0 +1,170 @@
...
+{
+    LV_UNUSED(attrs);
+
+    void * item = lv_buttonmatrix_create(lv_xml_state_get_parent(state));
+    return item;
+}
+
+void lv_xml_buttonmatrix_apply(lv_xml_parser_state_t * state, const char ** attrs)
+{
+    void * item = lv_xml_state_get_item(state);
```
uLipe, line 50: Okay I should not be too strict but, function that does not return error we should provide parameter validation plus a LOG_ERR for tracking, otherwise silent failures would be a pain to debug.

src/others/xml/parsers/lv_xml_buttonmatrix_parser.c
```diff
@@ -0,0 +1,170 @@
...
+    }
+}
+
+/**********************
+ *   STATIC FUNCTIONS
+ **********************/
+
+static lv_buttonmatrix_ctrl_t ctrl_text_to_enum_value(const char * txt)
+{
+    if(lv_streq("none", txt)) return LV_BUTTONMATRIX_CTRL_NONE;
```
uLipe, line 137: Here too, what are the consequences of txt being NULL?

src/widgets/buttonmatrix/lv_buttonmatrix.h
```diff
@@ -40,8 +56,8 @@ typedef enum {
     LV_BUTTONMATRIX_CTRL_CLICK_TRIG   = 0x0200, /**< 1: Enables sending LV_EVENT_VALUE_CHANGE on CLICK, 0: sends LV_EVENT_VALUE_CHANGE on PRESS. */
     LV_BUTTONMATRIX_CTRL_POPOVER      = 0x0400, /**< Show button text in a pop-over while being pressed. */
     LV_BUTTONMATRIX_CTRL_RECOLOR      = 0x0800, /**< Enable text recoloring with `#color` */
-    LV_BUTTONMATRIX_CTRL_RESERVED_2   = 0x1000, /**< Reserved for later use */
-    LV_BUTTONMATRIX_CTRL_RESERVED_3   = 0x2000, /**< Reserved for later use */
+    LV_BUTTONMATRIX_CTRL_RESERVED_1   = 0x1000, /**< Reserved for later use */
```
uLipe, line 59: Is it sure this will the position to be used on the future, if not 100% of chance I would remove them rather use a reserved field that may end up forgotten if the feature is not implemented.

uLipe (commented):
Some small nits, otherwise LGTM

xmls/lv_buttonmatrix.xml
```diff
@@ -31,16 +31,15 @@ Example
 			<enum name="popover"     help="Show a popover when pressing this key"/>
 			<enum name="reserved_1"  help="Reserved for later use"/>
```
liamHowatt:
```suggestion
			<enum name="popover"     help="Show a popover when pressing this key"/>
			<enum name="recolor"     help="Enable text recoloring with `#color`"/>
			<enum name="reserved_1"  help="Reserved for later use"/>
```

tests/src/test_cases/xml/test_xml_buttonmatrix.c
```diff
@@ -0,0 +1,34 @@
...
+{
+    /* Function run after every test */
+    lv_obj_clean(lv_screen_active());
+}
+
+void test_xml_buttonmatrix_with_attrs(void)
+{
+
+    const char * attrs_1[] = {
+        "map", "'1' '2' '3' '\n' '4\\'s value' '\n' '5' '6' '7' '8' '9' '\n' 'Escape \\ 10'",
```
liamHowatt, line 21: Are you sure the screenshot result is what it should be? The rendered text is "4\\'s value". FYI, in a C string like this, `"\\"` is parsed as `"\"`, so the bytes of this string are `4\'s value`. You are displaying the `\`, when, intuitively I'd expect escaping the `'` to display the `'` but not the `\`.

Looking at the parser, it would add complexity to handle what I mentioned, since the string should need to have bytes moved left to overwrite the `\`, so I completely understand leaving it as-is.

src/others/xml/parsers/lv_xml_buttonmatrix_parser.c
```diff
@@ -0,0 +1,170 @@
...
+                                str_array_size += 4;
+                                str_array = lv_realloc(str_array, sizeof(char *) * str_array_size);
+                            }
+                            if(lv_streq("\\n", text_start)) text_start = "\n";
+                            str_array[btn_cnt] = lv_strdup(text_start);
+                            btn_cnt++;
+                            in_text = false;
+                        }
+                    }
+                }
```
liamHowatt, line 93: The sequence `\\'` should not escape the `'`, but it would escape it with this logic here.

If you had a test case `"'\\\\'"` (C parsed as `'\\'`), you would get an unterminated `'` issue, but really it's terminated.

tests/src/test_cases/xml/test_xml_buttonmatrix.c
liamHowatt: There is no test case for `\\`. In C, you'd express that as `\\\\`.

src/others/xml/parsers/lv_xml_buttonmatrix_parser.c
```diff
@@ -0,0 +1,170 @@
...
+    if(lv_streq("width_14", txt)) return LV_BUTTONMATRIX_CTRL_WIDTH_14;
+    if(lv_streq("width_15", txt)) return LV_BUTTONMATRIX_CTRL_WIDTH_15;
+    if(lv_streq("hidden", txt)) return LV_BUTTONMATRIX_CTRL_HIDDEN;
+    if(lv_streq("no_repeat", txt)) return LV_BUTTONMATRIX_CTRL_NO_REPEAT;
+    if(lv_streq("disabled", txt)) return LV_BUTTONMATRIX_CTRL_DISABLED;
+    if(lv_streq("checkable", txt)) return LV_BUTTONMATRIX_CTRL_CHECKABLE;
+    if(lv_streq("checked", txt)) return LV_BUTTONMATRIX_CTRL_CHECKED;
+    if(lv_streq("click_trig", txt)) return LV_BUTTONMATRIX_CTRL_CLICK_TRIG;
+    if(lv_streq("popover", txt)) return LV_BUTTONMATRIX_CTRL_POPOVER;
+    if(lv_streq("reserved_1", txt)) return LV_BUTTONMATRIX_CTRL_RESERVED_1;
```
liamHowatt:
```suggestion
    if(lv_streq("popover", txt)) return LV_BUTTONMATRIX_CTRL_POPOVER;
    if(lv_streq("recolor", txt)) return LV_BUTTONMATRIX_CTRL_RECOLOR;
    if(lv_streq("reserved_1", txt)) return LV_BUTTONMATRIX_CTRL_RESERVED_1;
```

src/others/xml/parsers/lv_xml_buttonmatrix_parser.h
```diff
@@ -0,0 +1,41 @@
+/**
+ * @file lv_xml_buttonmatrix_parser.h
+ *
+ */
+
+#ifndef LV_BUTTONMATRIX_XML_PARSER_H
```
liamHowatt:
```suggestion
#ifndef LV_XML_BUTTONMATRIX_PARSER_H
```
And others

src/font/lv_font_montserrat_14.c
```diff
@@ -2187,5 +2187,5 @@ lv_font_t lv_font_montserrat_14 = {
 #endif
     .dsc = &font_dsc           /*The custom font data. Will be accessed by `get_glyph_bitmap/dsc` */
 };
-
+const lv_font_t * lv_font_montserrat_14_p = &lv_font_montserrat_14;
```
liamHowatt: Is it used somewhere?

src/font/lv_font.c
```diff
@@ -30,6 +30,12 @@
  *  STATIC VARIABLES
  **********************/
 
+/**********************
+ *  GLOBAL VARIABLES
+ **********************/
+
+const lv_font_t * lv_font_default = LV_FONT_DEFAULT;
```
liamHowatt: Could it change? If not:

```suggestion
const lv_font_t const * lv_font_default = LV_FONT_DEFAULT;
```

liamHowatt (commented):
Thanks for splitting it into separate commits. Impressive work.

src/others/xml/parsers/lv_xml_buttonmatrix_parser.c
kisvegabor: Now the assumption is that the XML is always correct as it's coming from the editor which has a validator. 

Regardless of it,  agree that basic asserts would be useful. These are missing from everywhere not only from the button matrix, so I'd add them in a new PR.

src/widgets/buttonmatrix/lv_buttonmatrix.h
kisvegabor: Sorry, I don't get it. Do you recommend removing the RESERVED enums?

tests/src/test_cases/xml/test_xml_buttonmatrix.c
kisvegabor: This part is not final yet. I'm still not sure what kind of input method can we come up with in the editor. Now even a `\n` is interpreted literally in an XML. So I wouldn't touch it until we know more.

src/font/lv_font_montserrat_14.c
kisvegabor: Upps, thanks, removed.

src/others/xml/parsers/lv_xml_buttonmatrix_parser.h
kisvegabor: Fixed

src/font/lv_font.c
kisvegabor: Done

liamHowatt (approved):

src/others/xml/parsers/lv_xml_buttonmatrix_parser.c
uLipe: Fair enough, I will be resolving since it is a larger issue by itself the miss of some asserts

src/widgets/buttonmatrix/lv_buttonmatrix.h
uLipe: Yes that is the point, having a reserved field it is a try to foresee the future on the code, which in general results on a dead enum that is never used.

uLipe (approved):
I left some comments but they actually are not that hard blockers, so I will approve it too :) 

Great work!

src/widgets/buttonmatrix/lv_buttonmatrix.h
kisvegabor: Okay, I'll remove it in the next one.

PR #7772, opened 2025-02-13
feat(thorvg): use LVGL's malloc/realloc/zalloc/free

kisvegabor:
Now there are lot of hard to debug out of memory issues using ThorVG's Lottie. With this updated 
- we have ASSERTs on out of memory
- we can monitor the memory usage
- we can use a single heap for LVGL and Lottie

These new `lv_*` symbols are also easy to refactor to any `tvg_*` if there from ThorVG.

cc @FASTSHIFT @XuNeo @onecoolx @hermet

### Notes

### DIFF
--- a/examples/widgets/lottie/lv_example_lottie_1.c
+++ b/examples/widgets/lottie/lv_example_lottie_1.c
@@ -7,17 +7,17 @@
  */
 void lv_example_lottie_1(void)
 {
-    extern const uint8_t lv_example_lottie_approve[];
-    extern const size_t lv_example_lottie_approve_size;
+    extern const uint8_t rain[];
+    extern const size_t rain_size;
 
     lv_obj_t * lottie = lv_lottie_create(lv_screen_active());
-    lv_lottie_set_src_data(lottie, lv_example_lottie_approve, lv_example_lottie_approve_size);
+    lv_lottie_set_src_data(lottie, rain, rain_size);
 
 #if LV_DRAW_BUF_ALIGN == 4 && LV_DRAW_BUF_STRIDE_ALIGN == 1
     /*If there are no special requirements, just declare a buffer
       x4 because the Lottie is rendered in ARGB8888 format*/
-    static uint8_t buf[64 * 64 * 4];
-    lv_lottie_set_buffer(lottie, 64, 64, buf);
+    static uint8_t buf[364 * 364 * 4];
+    lv_lottie_set_buffer(lottie, 364, 364, buf);
 #else
     /*For GPUs and special alignment/strid setting use a draw_buf instead*/
     LV_DRAW_BUF_DEFINE(draw_buf, 64, 64, LV_COLOR_FORMAT_ARGB8888);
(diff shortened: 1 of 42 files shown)


FASTSHIFT:
@hermet Can we add a wrapper to thorvg's memory allocation interface? It is convenient for users to customize the dynamic memory allocation interface

hermet:
> @hermet Can we add a wrapper to thorvg's memory allocation interface? It is convenient for users to customize the dynamic memory allocation interface, such as `tvg_malloc` `tvg_free`.

@FASTSHIFT Yes, we have a plan it, just delayed... sorry about that. :-)

FASTSHIFT (approved):

kisvegabor:
> Yes, we have a plan it, just delayed... sorry about that. :-)

No worries, happy to hear that it's not forgotten. 😊

I suggest merging this one for now, and we can use the upstream ThorVG solution once it's added.

hermet:
https://github.com/thorvg/thorvg/pull/3244
```
    void* operator new(std::size_t size) {
        return tvg::malloc(size);
    }
    
    void operator delete(void* ptr) noexcept {
        tvg::free(ptr);
    }
namespace tvg {
    template<typename T = void*>
    static inline T calloc(size_t nmem, size_t size)
    {
        return static_cast<T>(std::calloc(nmem, size));
    }

    template<typename T = void*>
    static inline T realloc(void* ptr, size_t size)
    {
        return static_cast<T>(std::realloc(ptr, size));
    }

    static inline void free(void* ptr)
    {
        std::free(ptr);
    }
}
```

kisvegabor:
Thank you @hermet! Two questions:
1. What is the recommended way of overwriting `tvg:malloc/free/etc`?
2. What do you think about adding a `TVG_ASSERT()` after malloc-s? It can be an empty define by default, but it'd be helpful for us to find where the out of memory issues happened by defining the assert to `while(1);`

hermet:
@kisvegabor Hello, I think you can change the code `tvgCommon.h` like this:

```
//tvgCommon.h
namespace tvg {
    template<typename T = void*>
    static inline T malloc(size_t size)
    {
        //add LVGL_MALLOC_ASSERT() here if you need
        return static_cast<T>(lvg_malloc(size));
    }

    template<typename T = void*>
    static inline T calloc(size_t nmem, size_t size)
    {
        return static_cast<T>(lvg_calloc(nmem, size)); 
    }

    template<typename T = void*>
    static inline T realloc(void* ptr, size_t size)
    {
        return static_cast<T>(lvg_realloc(ptr, size));
    }

    static inline void free(void* ptr)
    {
        lvg_free(ptr);
    }
}
```

hermet:
> It works for me, thank you. Could you also comment on this?
> 
> > What do you think about adding a TVG_ASSERT() after malloc-s? It can be an empty define by default, but it'd be helpful for us to find where the out of memory issues happened by defining the assert to while(1);

@kisvegabor Hi, I can't confirm this right now. Sorry about that. So far, ThorVG has no assertion policy, as it provides no significant benefit to the project but is mostly a matter of contributors preference or habit (so far). I guess the assert could be inside of lvgl memory allocators (i.e `lvg_malloc(size)`) if you are considering OOM. Otherwise, You can simply put the LVGL own assert() in the above thorvg custom wrappers like:
```
    template<typename T = void*>
    static inline T malloc(size_t size)
    {
        auto p = static_cast<T>(lvg_malloc(size));
        lvgl_assert(p);
        return p;
    }
```

Please let me know If I am missed something. Thanks. :)

kisvegabor:
Adding the asserts in your template is a good idea! When will you release these changes?

kisvegabor:
I see. In this case I suggest merging this PR for now, and when we update ThorVg in a few month we can use your solution.

uLipe (approved):

PR #7778, opened 2025-02-15
fix(lodepng): fix memory allocation interface mismatch

ccruquan:
By default, the data for PNG images is allocated using the default allocator, specifically through `lv_mem_alloc`. However, if the user defines LODEPNG_NO_COMPILE_ALLOCATORS, the allocation is done via `lodepng_malloc`. When freeing the memory, it is important to use the corresponding free function based on the allocation method used.

### DIFF
--- a/src/extra/libs/png/lv_png.c
+++ b/src/extra/libs/png/lv_png.c
@@ -16,6 +16,11 @@
 /*********************
  *      DEFINES
  *********************/
+#ifdef LODEPNG_COMPILE_ALLOCATORS
+#define lv_png_free(ptr) lv_mem_free((ptr))
+#else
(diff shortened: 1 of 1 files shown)


FASTSHIFT (approved):

uLipe (approved):

liamHowatt (approved):
It would also be acceptable to always call `lodepng_free` in `lv_png.c`.

https://github.com/lvgl/lvgl/blob/d30df363ff69b9ad9706a533764a577b7ed9cc3e/src/extra/libs/png/lodepng.c#L73-L97

PGNetHun (approved):

kisvegabor:
Thank you! The faling CI is unrelated.

PR #7779, opened 2025-02-15
fix(tiny_ttf): fix GPOS lookup list table address

matevzmihalic:
Lookup list offset is specified from start of GPOS table, not from start of file: [specs](https://learn.microsoft.com/en-us/typography/opentype/spec/gpos#:~:text=lookupListOffset-,Offset%20to%20LookupList%20table%2C%20from%20beginning%20of%20GPOS%20table.,-GPOS%20Header%2C%20version)

### DIFF
--- a/src/libs/tiny_ttf/stb_truetype_htcw.h
+++ b/src/libs/tiny_ttf/stb_truetype_htcw.h
@@ -2758,7 +2758,7 @@ static stbtt_int32 stbtt__GetGlyphGPOSInfoAdvance(const stbtt_fontinfo * info, i
     if(ttUSHORT(data, 2 + info->gpos) != 0) return 0;  // Minor version 0
 
     lookupListOffset = ttUSHORT(data, 8 + info->gpos);
-    lookupList = lookupListOffset;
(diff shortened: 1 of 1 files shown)


matevzmihalic:
In [upstream](https://github.com/nothings/stb/blob/master/stb_truetype.h#L2506-L2512) this is fixed by adding the gpos offset to data at the start, but this PR achieves the same result.

Without this fix incorrect data gets loaded as gpos table and with a specific ttf file it caused significantly longer rendering times.

liamHowatt:
Thank you. To streamline the fix, would it make more sense to replace it with the full contents of upstream or not?

kisvegabor (approved):

matevzmihalic:
I was testing label render times with the same font, but different glyph sets and was surprised when a font with just basic latin character set was slower than the full font which was 10x bigger.
Then I've run the test with small font file in profiler and saw that it took a lot of time in `stbtt__GetGlyphGPOSInfoAdvance` so I went through that function with debugger and found out what was wrong.

PR #7782, opened 2025-02-16
fix(thorvg): fix internal types of rasterXYFlip

pharuxtan:
Before the C++ compiler couldn't decide the template of std::min (that happen if `int` type is not an `int32_t`), so explicitly changing all types to int32_t fixes this.


FASTSHIFT (dismissed):

liamHowatt (approved):
Thanks for fixing it upstream too!

PGNetHun (approved):

PR #7789, opened 2025-02-18
fix(example_grad): return value from lv_style_get_prop not checked

pfvogel:
In the frac_1_event_cb routine, a call is made to lv_style_get_prop but the return value is not checked.  The lv_style_value returned is not populated but is used later in the code.
Add a check for the return value and a warning message on failure.


pfvogel:
There are more instances of this same type of this issue in this file.  I'm going to close this pull request and include the fixes for all instances in the new pull request

PR #7794, opened 2025-02-18
fix(indev): decouple scroll momentum decay from indev read loop (#6832)

### LINKED ISSUE #6832: LV_INDEV_MODE_EVENT is not compatible with scroll momentum
### LVGL version

v9.2.0

### What happened?

Try to throw a scrollable list with the input device created by `lv_sdl_mouse_create()`. Scrolling stops immediately, there is no scroll momentum.

This was working in LVGL 8. Bisected to 2b0587197c038b960b69920930c43877fb6b91c8 as the first bad commit.

If I understand correctly, this happens due to the decay in:

https://github.com/lvgl/lvgl/blob/559618d9d2c66069f61ba673b7bce6bf3e7eb036/src/indev/lv_indev.c#L1254-L1255

This assumes that `indev_proc_press()` is called at a roughly fixed interval. But if an input driver with `LV_INDEV_MODE_EVENT` delivers events much more quickly (like SDL does), the scroll vector decays to 0 almost immediately.

### How to reproduce?

Using the same example again, with the SDL input driver.

```c
static void demo()
{
    lv_obj_t * screen = lv_obj_create(NULL);
    lv_obj_set_size(screen, 512, 600);
    lv_obj_set_flex_flow(screen, LV_FLEX_FLOW_COLUMN);

    for (int i = 0; i < 100; i++)
    {
        lv_obj_t * btn = lv_button_create(screen);
        lv_obj_set_size(btn, lv_pct(100), LV_SIZE_CONTENT);
    }

    lv_disp_load_scr(screen);
}

int main()
{
    lv_init();

    lv_sdl_window_create(512, 600);
    lv_sdl_mouse_create();

    demo();

    while(true) {
        uint32_t delay = lv_timer_handler();
        usleep(delay * 1000);
    }

    return 0;
}
```

niklasf:
Fixes #6832.

As suggested in the issue, this computes the scroll throw vector based on a short history.

Initially I computed the speed over the entire history up to a maximum time, but then it depends a lot on individual history entries making the cut or not and can feel a bit inconsistent.

Next, I applied exponential decay to history entries, `x * exp(-t / 33.0)`, which felt much smoother.

To avoid floating point arithmetic, I tried some series expansions, but ultimately simple linear interpolation seems good enough.

Tested with SDL (where the full history is often relevant) and evdev with default refresh rate (where usually ~2 entries make the cut).

### DIFF
--- a/src/indev/lv_indev.c
+++ b/src/indev/lv_indev.c
@@ -192,6 +192,8 @@ void indev_read_core(lv_indev_t * indev, lv_indev_data_t * data)
         data->key = LV_KEY_ENTER;
     }
 
+    data->timestamp = lv_tick_get();
+
     if(indev->read_cb) {
         LV_TRACE_INDEV("calling indev_read_cb");
         indev->read_cb(indev, data);
@@ -245,11 +247,12 @@ void lv_indev_read(lv_indev_t * indev)
         indev->state = data.state;
 
         /*Save the last activity time*/
+        indev->timestamp = data.timestamp;
         if(indev->state == LV_INDEV_STATE_PRESSED) {
-            indev->disp->last_activity_time = lv_tick_get();
+            indev->disp->last_activity_time = data.timestamp;
         }
         else if(indev->type == LV_INDEV_TYPE_ENCODER && data.enc_diff) {
-            indev->disp->last_activity_time = lv_tick_get();
+            indev->disp->last_activity_time = data.timestamp;
         }
 
         if(indev->type == LV_INDEV_TYPE_POINTER) {
@@ -438,9 +441,8 @@ void lv_indev_stop_processing(lv_indev_t * indev)
 
 void lv_indev_reset_long_press(lv_indev_t * indev)
 {
-    indev->long_pr_sent         = 0;
-    indev->longpr_rep_timestamp = lv_tick_get();
-    indev->pr_timestamp         = lv_tick_get();
+    indev->long_pr_sent = 0;
+    indev->longpr_rep_timestamp = indev->pr_timestamp = lv_tick_get();
 }
 
 void lv_indev_set_cursor(lv_indev_t * indev, lv_obj_t * cur_obj)
@@ -792,7 +794,7 @@ static void indev_keypad_proc(lv_indev_t * i, lv_indev_data_t * data)
     /*Key press happened*/
     if(data->state == LV_INDEV_STATE_PRESSED && prev_state == LV_INDEV_STATE_RELEASED) {
         LV_LOG_INFO("%" LV_PRIu32 " key is pressed", data->key);
-        i->pr_timestamp = lv_tick_get();
+        i->pr_timestamp = i->timestamp;
 
         /*Move the focus on NEXT*/
         if(data->key == LV_KEY_NEXT) {
@@ -838,19 +840,19 @@ static void indev_keypad_proc(lv_indev_t * i, lv_indev_data_t * data)
         }
 
         /*Long press time has elapsed?*/
-        if(i->long_pr_sent == 0 && lv_tick_elaps(i->pr_timestamp) > i->long_press_time) {
+        if(i->long_pr_sent == 0 && lv_tick_diff(i->timestamp, i->pr_timestamp) > i->long_press_time) {
             i->long_pr_sent = 1;
             if(data->key == LV_KEY_ENTER) {
-                i->longpr_rep_timestamp = lv_tick_get();
+                i->longpr_rep_timestamp = i->timestamp;
 
                 if(send_event(LV_EVENT_LONG_PRESSED, indev_act) == LV_RESULT_INVALID) return;
             }
         }
         /*Long press repeated time has elapsed?*/
         else if(i->long_pr_sent != 0 &&
-                lv_tick_elaps(i->longpr_rep_timestamp) > i->long_press_repeat_time) {
+                lv_tick_diff(i->timestamp, i->longpr_rep_timestamp) > i->long_press_repeat_time) {
 
-            i->longpr_rep_timestamp = lv_tick_get();
+            i->longpr_rep_timestamp = i->timestamp;
 
             /*Send LONG_PRESS_REP on ENTER*/
             if(data->key == LV_KEY_ENTER) {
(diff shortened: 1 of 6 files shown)


niklasf:
Not stale.

src/indev/lv_indev.c
```diff
@@ -1283,12 +1297,21 @@ static void indev_proc_press(lv_indev_t * indev)
...
-    indev->pointer.scroll_throw_vect.x = (indev->pointer.scroll_throw_vect.x + indev->pointer.vect.x) / 2;
-    indev->pointer.scroll_throw_vect.y = (indev->pointer.scroll_throw_vect.y + indev->pointer.vect.y) / 2;
+    indev->pointer.vect_hist[indev->pointer.vect_hist_index] = indev->pointer.vect;
+    indev->pointer.vect_hist_timestamp[indev->pointer.vect_hist_index] = lv_tick_get();
+    indev->pointer.vect_hist_index = (indev->pointer.vect_hist_index + 1) % LV_INDEV_VECT_HIST_SIZE;
+
+    indev->pointer.scroll_throw_vect.x = 0;
+    indev->pointer.scroll_throw_vect.y = 0;
+    for(int i = 0; i < LV_INDEV_VECT_HIST_SIZE; i++) {
+        int32_t t = lv_tick_elaps(indev->pointer.vect_hist_timestamp[i]);
```
liamHowatt: It'd be nice if there was a `lv_ms_elaps` so you could reuse the gotten tick. Just thinking of NuttX where the tick getter is clock_gettime. I don't know though.

src/indev/lv_indev_private.h
```diff
@@ -82,6 +83,9 @@ struct _lv_indev_t {
         lv_point_t last_point; /**< Last point of input device.*/
         lv_point_t last_raw_point; /**< Last point read from read_cb. */
         lv_point_t vect; /**< Difference between `act_point` and `last_point`.*/
+        lv_point_t vect_hist[LV_INDEV_VECT_HIST_SIZE];
+        uint32_t   vect_hist_timestamp[LV_INDEV_VECT_HIST_SIZE];
+        uint8_t    vect_hist_index;
```
liamHowatt, line 90: I guess 100 bytes is the price of progress

src/indev/lv_indev_private.h
```diff
@@ -20,6 +20,7 @@ extern "C" {
 /*********************
  *      DEFINES
  *********************/
+#define LV_INDEV_VECT_HIST_SIZE 8
```
liamHowatt, line 24: While testing, I found that `1` feels closest to `master`, which is basically what it was. `8` truly _flies_! Could the change be too shocking for existing users?

liamHowatt (commented):
It's a great experience!

Nice implementation.

niklasf:
Maybe `LV_INDEV_VECT_HIST_SIZE` should be compile-time configurable, so that users who know that their indev read loop runs at a given frequency can pick a lower value?

Or maybe it's back to the drawing board, to see if something similar can be achieved without explicitly keeping the history around.

niklasf:
Still interested to get this over the finish line, but need some direction (https://github.com/lvgl/lvgl/pull/7794#issuecomment-2781329552).

kisvegabor:
Hi Niklas,

I've just tested your PR and in general I like it, The scrolling experience is much better. However I'm not sure about one thing.

In theory having 2 samples like
```
0ms: y=0
10ms: y=100
```

and having 6 samples like
```
0ms: y=0
2ms: y=20
4ms: y=40
6ms: y=60
8ms: y=80
10ms: y=100
```
Should result in a similar (same?) momentum.

Right now, not considering the decay, in the first case the total momentum is 100, in the second case it's 300. 

I feel like somehow we should also consider how much time elapsed between the samples to calculate a speed.

E.g. in the first case: 
```
1. sample: 100px/10ms = 10px/ms
```
In the second case
```
5 samples: e.g. (100px - 80px) / (10ms-8ms) = 20px / 2ms = 10px/ms for each.
```

So in the end I think we should
1. allow the `read_cb` to also pass a timestamp for the sample (in case they were buffered and we ready all the samples at once). We can set the timestamp before calling the `read_cb` that can be overwritten in the `read_cb` if needed.
2. calculate the average of a velocities by giving less weight if the timestamp is older.

What do you think?

kisvegabor:
The more I think about the more correct this simple feature seems. Even if the read period changes it will be correct as it doesn't matter that we have  5px/sample change in every 10 ms, or 50 px/sample change over 100 ms. It will result in the same vector (10 x 5 = 50 and 1 x 50 = 50).  

Just one thing came to my mind:
Please add a timestamp filed to `lv_indev_data_t` and set it to `lv_tick_get()` before calling the `read_cb`. This way `read_cb` can overwrite it if needed and we can have the correct timestamp if the indev driver buffered some points and they will be read by using the `continue_reading` flag.

niklasf:
Added that timestamp and a new utility `uint32_t lv_tick_diff(uint32_t tick, uint32_t prev_tick)` to use it. That's essentially the `lv_ms_elaps` that @liamHowatt suggested above.

src/indev/lv_indev.c
niklasf: Implemented

niklasf:
Rebased for CI.

niklasf:
Not stale. Waiting for next round of review.

liamHowatt (commented):
I think it's good. It really more faithful to how touch screen throw behaves on the devices that users are familiar with.

The only thing missing is mentioning in the indev docs that timestamp can be overwritten if a more accurate one is available.

> Maybe `LV_INDEV_VECT_HIST_SIZE` should be compile-time configurable, so that users who know that their indev read loop runs at a given frequency can pick a lower value?

I think it's ok without a config option.

> Or maybe it's back to the drawing board, to see if something similar can be achieved without explicitly keeping the history around.

100 bytes isn't that much, all things considered.

niklasf:
Rebased & resolved conflicts.

> The only thing missing is mentioning in the indev docs that timestamp can be overwritten if a more accurate one is available.

Added to the *Buffered Reading* section.

liamHowatt (approved):
Thanks!

kisvegabor (approved):
Thank you!

PR #7796, opened 2025-02-19
feat(blend): add blend difference mode

fbiego:
LV_BLEND_MODE_DIFFERENCE
This is the absolute difference between background and foreground color

> LV_ABS(foreground - background)

### Notes

### DIFF
--- a/src/draw/sw/blend/lv_draw_sw_blend_to_rgb565.c
+++ b/src/draw/sw/blend/lv_draw_sw_blend_to_rgb565.c
@@ -495,6 +495,10 @@ static void LV_ATTRIBUTE_FAST_MEM i1_image_blend(lv_draw_sw_blend_image_dsc_t * 
                               ((((dest_buf_u16[dest_x] >> 5) & 0x3F) * ((l8_to_rgb565(chan_val) >> 2) & 0x3F) >> 6) << 5) |
                               (((dest_buf_u16[dest_x] & 0x1F) * (l8_to_rgb565(chan_val) & 0x1F)) >> 5);
                         break;
+                    case LV_BLEND_MODE_DIFFERENCE:
+                        // Difference blending mode
+                        res = (LV_ABS(dest_buf_u16[dest_x] - l8_to_rgb565(chan_val)));
+                        break;
                     default:
                         LV_LOG_WARN("Not supported blend mode: %d", dsc->blend_mode);
                         return;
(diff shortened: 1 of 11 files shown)


src/draw/sw/blend/lv_draw_sw_blend_to_rgb565.c
```diff
@@ -495,6 +495,10 @@ static void LV_ATTRIBUTE_FAST_MEM i1_image_blend(lv_draw_sw_blend_image_dsc_t *
                               ((((dest_buf_u16[dest_x] >> 5) & 0x3F) * ((l8_to_rgb565(chan_val) >> 2) & 0x3F) >> 6) << 5) |
                               (((dest_buf_u16[dest_x] & 0x1F) * (l8_to_rgb565(chan_val) & 0x1F)) >> 5);
                         break;
+                    case LV_BLEND_MODE_DIFFERENCE:
+                        // Difference blending mode
```
kisvegabor:
```suggestion
                        /*Difference blending mode*/
```

src/draw/sw/blend/lv_draw_sw_blend_to_rgb888.c
```diff
@@ -922,6 +922,11 @@ static inline void LV_ATTRIBUTE_FAST_MEM blend_non_normal_pixel(uint8_t * dest,
             res[1] = (dest[1] * src.green) >> 8;
             res[2] = (dest[2] * src.red) >> 8;
             break;
+        case LV_BLEND_MODE_DIFFERENCE:
+            res[0] = LV_ABS(dest[0] - src.blue);
```
kisvegabor: I'm not sure if it's needed, but it'd make sense as the the numbers are unsigned by default, so `1 - 2 = 254`
```suggestion
            res[0] = LV_ABS((int16_t)dest[0] - src.blue);
```

xmls/globals.xml
```diff
@@ -77,6 +77,7 @@
             <enum name="additive"    	help=""/>
             <enum name="subtractive"  	help=""/>
             <enum name="multiply"  	  	help=""/>
+			<enum name="difference"  	help=""/>
```
kisvegabor: See the indentation. Use spaces instead of tabs.

kisvegabor (commented):
It is a great first step, however we can also consider adding the blend mode handling to the color fill blend functions. 

Right now there is no way to handle blend modes in color fills, as if you set `style_blend_mode`, a new layer will be created, and that layer will be blended as an image.

So it's good for now, but we should keep in mind that there are some more opportunities here.

src/draw/sw/blend/lv_draw_sw_blend_to_rgb888.c
fbiego: Integer promotions happen automatically in C when performing arithmetic operations, but maybe we can add it?

```
uint8_t x = 1;
uint8_t y = 2;
uint8_t res = LV_ABS(x - y);
LV_LOG_USER("Without cast 1 - 2 = %d", res);
res = LV_ABS((int16_t)x - y);
LV_LOG_USER("With cast 1 - 2 = %d", res);
```

> [User]  (0.496, +496)    main: Without cast 1 - 2 = 1 main.cpp:327
> [User]  (0.496, +0)      main: With cast 1 - 2 = 1 main.cpp:329

uLipe (approved):
NIce one @fbiego !

As I understood correctly the support for different blending functions needs to be ported across the GPUS we have supported no? 

So can I consider i will need to take this to the VGLites and D/AVE GPU, right @kisvegabor ?

kisvegabor:
It's not mandatory to support all the draw features on all the draw units, but now the opportunity is there. However, if you know counterpart of it in VGLite and Dave2D a PR would be welcome for sure!

kisvegabor (approved):

PR #7799, opened 2025-02-19
fix(scripts): initialize reserved fields of lv_image_dsc_t and lv_image_header_t to prevent compiler warnings

sparkles43:
Previously the LVGLImage.py script did not initialize reserved fields of the lv_image_dsc_t and lv_image_header_t. This led to a flood of compiler warnings about uninitialized variables.

This PR sets these variables to zero and NULL respectively.

### DIFF
--- a/scripts/LVGLImage.py
+++ b/scripts/LVGLImage.py
@@ -355,8 +355,10 @@ const lv_image_dsc_t {varname} = {{
   .header.w = {w},
   .header.h = {h},
   .header.stride = {stride},
+  .header.reserved_2 = 0,
   .data_size = sizeof({varname}_map),
   .data = {varname}_map,
+  .reserved = NULL,
 }};
 
 '''


uLipe (approved):
Thanks for the correction, this source of warning could be a pain on projects that promotes warnings to errors.

liamHowatt (approved):
Just had a thought. If we replace these `reserved` fields someday, they will actually have to be anonymous unions so that old code assigning a value to `reserved` will still compile.

```c
typedef struct {
    lv_image_header_t header;   /**< A header describing the basics of the image*/
    uint32_t data_size;         /**< Size of the image in bytes*/
    const uint8_t * data;       /**< Pointer to the data of the image*/
    union {const void * reserved; void * some_new_field};
} lv_image_dsc_t;
```

PGNetHun (approved):

uLipe (approved):

XuNeo (approved):
I think it would be automatically initialized to zero. https://en.cppreference.com/w/c/language/struct_initialization

If the compiler you used generates warning, then we can do the simple fix. 
So looks good to me.

fbiego (approved):

PR #7800, opened 2025-02-19
fix(scale): assert mode value is valid

pfvogel:
If an invalid scale mode is used, the scale_get_tick_points() does not assign values to the tick points and uninitialized values are used to draw lines.  The following sequence of code causes  the issue when rendered:

```
    lv_obj_t * scale = lv_scale_create(lv_screen_active());

    lv_scale_set_post_draw(scale, true);
    lv_scale_set_draw_ticks_on_top(scale, true);
    lv_scale_set_label_show(scale, false);
    // set an invalid value for the mode
    lv_scale_set_mode(scale, LV_SCALE_MODE_HORIZONTAL_TOP |
                             LV_SCALE_MODE_ROUND_OUTER);
```
Add an assertion message to check that the mode is one of the valid values.

### DIFF
--- a/src/widgets/scale/lv_scale.c
+++ b/src/widgets/scale/lv_scale.c
@@ -1065,6 +1065,14 @@ static void scale_get_tick_points(lv_obj_t * obj, const uint32_t tick_idx, bool 
     int32_t major_len = 0;
     int32_t radial_offset = 0;
 
+    LV_ASSERT_MSG((scale->mode == LV_SCALE_MODE_HORIZONTAL_TOP) ||
+                  (scale->mode == LV_SCALE_MODE_HORIZONTAL_BOTTOM) ||
+                  (scale->mode == LV_SCALE_MODE_VERTICAL_LEFT) ||
+                  (scale->mode == LV_SCALE_MODE_VERTICAL_RIGHT) ||
+                  (scale->mode == LV_SCALE_MODE_ROUND_INNER) ||
+                  (scale->mode == LV_SCALE_MODE_ROUND_OUTER),
+                  "unknown value for scale mode");
+
     if(is_major_tick) {
         major_len = lv_obj_get_style_length(obj, LV_PART_INDICATOR);
         radial_offset = lv_obj_get_style_radial_offset(obj, LV_PART_INDICATOR);


uLipe (approved):

liamHowatt (commented):
Thanks. Can this check be put in `lv_scale_set_mode`? That's where the erroneous value of `mode` would originate and it'd only be checked once.

pfvogel:
Moving the assertion there makes sense from a functional perspective, but the check that is complaining is on the "using" the value path as opposed to the "setting" the value path.
If the assertion was only added to the `lv_scale_set_mode()` routine, the checker would still complain -- the path that is failing starts at the `lv_scale_event()` call for the `LV_EVENT_DRAW_POST` event.

liamHowatt:
I don't understand. Where does `scale->mode` become a bad value before `lv_scale_set_mode` is called?

pfvogel:
In practice, scale->mode would only be set through the lv_scale_set_mode() function and having a check there would prevent the issue from arising when the scale is drawn.  The static code analysis tool isn't sophisticated enough to create an object, set a bad scale value and then cause the scale to be rendered to detect the issue.  Instead, it doesn't make any assumptions about the values used by functions, and is only constrained by the requirements needed to exercise the code path and assertions along that path.  So in this case, the analysis starts at the lv_scale_event() and determines values needed to exercise the various the code paths.  In this case, the code relies on the previously set value for scale->mode, so there are effectively no constraints on the mode value imposed by the lv_scale_event() routine.  The code analysis then reports that an invalid value for scale->mode will result in uninitialized values being used to draw lines.
Moving the assertion to the lv_scale_set_mode() would fix the functional issue but the static code analysis would still fail.  Would it make sense to make a new assertion type that could be used for analysis checks, but does not result in code being generated when compiled?

liamHowatt:
I imagine this line of reasoning exploding into 100s of possible checks being added. There's a reason the struct members are defined in `lv_scale_private.h`. It's the user's fault if they include private headers and set the mode to an erroneous value without the setter. If there could only be one check, I'd say put it in the setter, so the API abiding user is given immediate feedback about their mistake.

> Would it make sense to make a new assertion type that could be used for analysis checks, but does not result in code being generated when compiled?

Something like `LV_ASSERT_STATIC_ANALYSYS`? I think it's possible. Did your static analysis tool report many more issues?

pfvogel:
I agree that a user shouldn't include the private header, and that would make it difficult to change the mode value between it being set and its later use.
The analysis tool has about 30 high priority issues remaining, with about half of these in the red-black code and string routines.  These are probably false positives, so having an assertion macro should let me clear those without any impact to the code.

kisvegabor:
Thank you @liamHowatt for bringing this to my attention. If there are no 1000+ issues like this I think we can fix them, as it would be a good point in the eyes of the automotive UI developer if LVGL is compliant with some standard static analysis tools. However, in this case we should add a static analysis tool to our CI. 

@pfvogel  Which tool do you use?

pfvogel:
I've been using [CodeChecker](https://codechecker.readthedocs.io/en/latest/).  Free, built on top of clang and works very well in a cmake/make environment.  It can take a while to run, maybe 20-30 mins in my virtual machine.

At the moment, there are about 27 high priority issues and about 800 medium priority issues. I think the majority of the remaining high priority issues are false positives, and most of the 800 medium issues are related to type conversions without casts.

fbiego (approved):

PR #7808, opened 2025-02-20
feat(sdl): add window icon settings

luozhenhua:
Fixes #xxxx 

### Notes

### DIFF
--- a/src/drivers/sdl/lv_sdl_window.c
+++ b/src/drivers/sdl/lv_sdl_window.c
@@ -3,6 +3,10 @@
  *
  */
 
+/**
+ * Modified by NXP in 2025
+ */
+
 /*********************
  *      INCLUDES
  *********************/
@@ -187,6 +191,15 @@ void lv_sdl_window_set_title(lv_display_t * disp, const char * title)
 {
     lv_sdl_window_t * dsc = lv_display_get_driver_data(disp);
     SDL_SetWindowTitle(dsc->window, title);
+}
+
+void lv_sdl_window_set_icon(lv_display_t * disp, void * icon, int32_t width, int32_t height)
+{
+    lv_sdl_window_t * dsc = lv_display_get_driver_data(disp);
+    SDL_Surface * iconSurface = SDL_CreateRGBSurfaceWithFormatFrom(icon, width, height, 32, width * 4,
+                                                                   SDL_PIXELFORMAT_ARGB8888);
+    SDL_SetWindowIcon(dsc->window, iconSurface);
+    SDL_FreeSurface(iconSurface);
 }
 
 void * lv_sdl_window_get_renderer(lv_display_t * disp)
--- a/src/drivers/sdl/lv_sdl_window.c
+++ b/src/drivers/sdl/lv_sdl_window.c
@@ -20,7 +20,11 @@
 #ifndef __USE_ISOC11
     #define __USE_ISOC11
 #endif
-#include <stdlib.h>
+#ifndef _WIN32
+    #include <stdlib.h>
+#else
+    #include <malloc.h>
+#endif /* _WIN32 */
 
 #define SDL_MAIN_HANDLED /*To fix SDL's "undefined reference to WinMain" issue*/
 #include "lv_sdl_private.h"
--- a/src/drivers/sdl/lv_sdl_window.h
+++ b/src/drivers/sdl/lv_sdl_window.h
@@ -3,6 +3,10 @@
  *
  */
 
+/**
+ * Modified by NXP in 2025
+ */
+
 #ifndef LV_SDL_WINDOW_H
 #define LV_SDL_WINDOW_H
 
@@ -44,6 +48,8 @@ void lv_sdl_window_set_zoom(lv_display_t * disp, float zoom);
 float lv_sdl_window_get_zoom(lv_display_t * disp);
 
 void lv_sdl_window_set_title(lv_display_t * disp, const char * title);
+
+void lv_sdl_window_set_icon(lv_display_t * disp, void * icon, int32_t width, int32_t height);
 
 void * lv_sdl_window_get_renderer(lv_display_t * disp);
 
--- a/src/drivers/sdl/lv_sdl_keyboard.c
+++ b/src/drivers/sdl/lv_sdl_keyboard.c
@@ -143,7 +143,7 @@ void lv_sdl_keyboard_handler(SDL_Event * event)
         case SDL_TEXTINPUT: {                   /*Text input*/
                 const size_t len = lv_strlen(dsc->buf) + lv_strlen(event->text.text);
                 if(len < KEYBOARD_BUFFER_SIZE - 1)
-                    strcat(dsc->buf, event->text.text);
+                    lv_strcat(dsc->buf, event->text.text);
             }
             break;
         default:


src/drivers/sdl/lv_sdl_window.h
```diff
@@ -45,6 +49,8 @@ float lv_sdl_window_get_zoom(lv_display_t * disp);
 
 void lv_sdl_window_set_title(lv_display_t * disp, const char * title);
 
+void lv_sdl_window_set_icon(lv_display_t * disp, uint16_t * icon);
```
FASTSHIFT: Does this function require the image size to be passed in? What if the user passes in a different size?

FASTSHIFT (changes requested):

src/drivers/sdl/lv_sdl_window.c
```diff
@@ -181,6 +189,14 @@ void lv_sdl_window_set_title(lv_display_t * disp, const char * title)
     SDL_SetWindowTitle(dsc->window, title);
 }
 
+void lv_sdl_window_set_icon(lv_display_t * disp, uint16_t * icon)
+{
+    lv_sdl_window_t * dsc = lv_display_get_driver_data(disp);
```
AndreCostaaa, line 198:
```suggestion
    lv_sdl_window_t * dsc = (lv_sdl_window_t *) lv_display_get_driver_data(disp);
```

So this code stays compatible with c++

src/drivers/sdl/lv_sdl_window.h
ZongchunYu: Thanks for your comments.  we will update this function to add image size.

src/drivers/sdl/lv_sdl_window.c
ZongchunYu: Thank you, we will update this line.

liamHowatt: We only do these casts in the examples, AFAIK. There are a lot of places all over LVGL where implicit `void` pointer coercion happens

AndreCostaaa: Currently, yes, i believe you're right but wouldn't you agree that it's a good idea to start adding these ?

liamHowatt: Personally, no! :laughing: is there a strong case for it?

AndreCostaaa: Fair point ahaha

1. Explicitness and readability, it tells the rest of the world you know what you're expecting from the function
2. Portability. The code simply doesn't compile with a c++ compiler 

```c
typedef struct {
	int x;
} bar;
bar a;
void *foo(void)
{
	return &a;
}
int main(void)
{
	bar *b = foo();

	b->x = 10;
}
```

```bash
> gcc -Wall -Wextra -pedantic -o main main.c
> g++ -Wall -Wextra -pedantic -o main main.c
main.c: In function ‘int main()’:
main.c:16:21: error: invalid conversion from ‘void*’ to ‘bar*’ [-fpermissive]
   16 |         bar *b = foo();
      |                  ~~~^~
      |                     |
      |                     void*
```

3. Maintainability : 

This compiles (and shouldn't!) - Spot the mistake:

```c
typedef struct {
	int x;
} bar;

typedef struct {
	float x;
} baz;

bar a;

void *foo(void)
{
	return &a;
}

int main(void)
{
	baz *b;

	b = foo();

	b->x = 10;
}
```

This doesn't compile

```c
typedef struct {
	int x;
} bar;

typedef struct {
	float x;
} baz;

bar a;

void *foo(void)
{
	return &a;
}

int main(void)
{
	baz *b;

	b = (bar *)foo();

	b->x = 10;
}
```
```bash
> gcc -Wall -Wextra -pedantic -o main main.c
main.c: In function ‘main’:
main.c:22:11: error: assignment to ‘baz *’ from incompatible pointer type ‘bar *’ [-Wincompatible-pointer-types]
   22 |         b = (bar *)foo();
      |           ^
```

liamHowatt: For (1.), why would someone compile a C file as C++? The examples are copy-pasted sometimes so it makes sense, but why for anything else?

I think (2.) is a 50/50 chance whether they put the correct cast. What if they do `b = (baz *)foo();`? In any case it's 50% better than nothing so you've convinced me, somewhat.

AndreCostaaa: 1) Well, since there are some C and Cpp files in lvgl the simplest way to compile an app using lvgl is by collecting every .c, .cpp and .S file using eg `find` and then compiling everything in a single command :

```bash
g++ $SRC_FILES -o app
```
Since there are .cpp files inside the source tree, the only way this can possible work is with a c++ compiler but if the C code isn't C++ compatible you remove that option from people

Of course you can always decide that you don't want to support building lvgl that way which is totally acceptable aswell

2) i don't think the problem is using the correct cast, in general the first time the code is written, the variable type (and thus the cast) are correct and you won't have any problems. The cast will save you in future updates, when for some reason you update the type of `b` and don't realize that an implicit cast to the previous type exists

liamHowatt: 1. I see the convenience of being able to do that, but C is more than a subset. It's its own language with divergent semantics.

2. Ah true it will help you if `b` changes, which is worth it enough. But not for the return value of `foo`. I guess the refactorer should check these users anyways.

liamHowatt: This line will not compile in C++ because `new` is a reserved keyword.

https://github.com/lvgl/lvgl/blob/b7061642a0c78e9f92b0cbabfb7215ba797a1c6f/src/stdlib/lv_mem.c#L141

AndreCostaaa: > 1. I see the convenience of being able to do that, but C is more than a subset. It's its own language with divergent semantics.

True but I don't think there's anything in the lvgl source code folder that would make it impossible to be C++ compliant. Most of the changes simply rely on an explicit cast or simple variable rename

> This line will not compile in C++ because `new` is a reserved keyword.

Yep, good one

This command will give you the whole list of things non-C++ compliant. Spoiler alert, there's a lot of them

```bash
g++ -c $(find src -type f \( -name '*.S' -o -name '*.cpp' -o -name '*.c' \))
```

ZongchunYu: Hello AndreCostaaa, liamHowatt,  Thanks for your comments!  May I know the conclusion for this suggest change?

luozhenhua: Since there is no agreement on the c++ compatible update, can we update the PR to address other review comments first?

AndreCostaaa: Yes, feel free to ignore this suggested change in your PR update

Thank you for your work and sorry for hijacking your PR for discussing lvgl c++ compliance :)

luozhenhua: Thanks for your review and comments.

FASTSHIFT (approved):

luozhenhua:
Hi @liamHowatt and @AndreCostaaa, could you please help to review the updated PR and approve if no further update is needed? Thanks.

uLipe (approved):

PR #7810, opened 2025-02-20
refactor(scripts): initialize lv_image_dsc_t by using C++20 compatible designated initializers

sparkles43:
In our project we use a single C++ compiler to compile resources like icons and images. Up to now the `LVGLImage.py` generates C code that uses nested designated initializers which is not compatible with C++.  Starting at C++ >=20 it has become possible to use ("unnested") designated initializers.

Therefore, I would like to propose a change of the code to make it use designated initializers but only in a not nested fashion. From my understanding this is compatible to the C standard.

_Remark: do we have a CI job for testing 'compilability' of LVGLImage.py generated code?_

Merge requirements:

- [ ] <https://github.com/lvgl/lvgl/pull/7799>

### DIFF
--- a/scripts/LVGLImage.py
+++ b/scripts/LVGLImage.py
@@ -349,13 +349,15 @@ uint8_t {varname}_map[] = {{
 }};
 
 const lv_image_dsc_t {varname} = {{
-  .header.magic = LV_IMAGE_HEADER_MAGIC,
-  .header.cf = LV_COLOR_FORMAT_{cf.name},
-  .header.flags = {flags},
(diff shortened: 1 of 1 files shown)


liamHowatt:
Can you compile the C files in your project as C?

The only C files we try to make C++ friendly are the examples because people copy the code into their Arduino sketches.

> do we have a CI job for testing 'compilability' of LVGLImage.py generated code?

It would be good to have and easy to add. But I'm not sure if there's a strong case to make the CI check with anything except a C compiler.

PR #7811, opened 2025-02-20
fix(obj): remove all events from the object

Pavi92:
In function lv_obj_remove_event_cb only the first event was removed
respectively marked for deletion.
Changed function prototype to match the prototype
of lv_obj_remove_event_cb_with_user_data.
Simply call this function with user_data = NULL to remove all events.
Expanded if condition in function lv_obj_remove_event_cb_with_user_data
to simply delete all events matching the event_cb without taking
user_data into consideration if user_data is NULL.

### DIFF
--- a/src/core/lv_obj_event.c
+++ b/src/core/lv_obj_event.c
@@ -136,7 +136,20 @@ bool lv_obj_remove_event_dsc(lv_obj_t * obj, lv_event_dsc_t * dsc)
 
 uint32_t lv_obj_remove_event_cb(lv_obj_t * obj, lv_event_cb_t event_cb)
 {
-    return lv_obj_remove_event_cb_with_user_data(obj, event_cb, NULL);
+    LV_ASSERT_NULL(obj);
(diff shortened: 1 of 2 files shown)


uLipe (dismissed):

Pavi92:
@uLipe what does this error mean "This branch must not contain merge commits"
I already did a rebase onto master

liamHowatt (commented):
Thanks. It makes sense.

Making `lv_obj_remove_event_cb` remove all event callbacks is important, I think.

However I think it should be done without changing the behavior of `lv_obj_remove_event_cb_with_user_data`. There may be applications that are relying on it to remove event callbacks with user_data of `NULL` only.

Could you just fix `lv_obj_remove_event_cb` by making it not return early?

Pavi92:
@liamHowatt done

uLipe (approved):

liamHowatt (approved):
Thanks!

PR #7828, opened 2025-02-23
fix(obj): modify style_cnt to uint16_t

yushuailong:
There are many combinations of `state `and `part` in widget. We should expand `style_cnt `so that each `selector `has a corresponding `style`. 

### Notes

### DIFF
--- a/src/core/lv_obj_private.h
+++ b/src/core/lv_obj_private.h
@@ -65,11 +65,11 @@ struct _lv_obj_t {
     lv_area_t coords;
     lv_obj_flag_t flags;
     lv_state_t state;
+    uint16_t style_cnt;
     uint16_t layout_inv : 1;
     uint16_t readjust_scroll_after_layout : 1;
     uint16_t scr_layout_inv : 1;
     uint16_t skip_trans : 1;
-    uint16_t style_cnt  : 6;
     uint16_t h_layout   : 1;
     uint16_t w_layout   : 1;
     uint16_t is_deleting : 1;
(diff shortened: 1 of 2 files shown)


src/core/lv_obj_private.h
```diff
@@ -65,11 +65,11 @@ struct _lv_obj_t {
     lv_area_t coords;
     lv_obj_flag_t flags;
     lv_state_t state;
+    uint16_t style_cnt;
```
FASTSHIFT: Is UINT8 sufficient?

FASTSHIFT (dismissed):

src/core/lv_obj_private.h
yushuailong: No, user customization is not considered. The widget has a maximum of 256 states and 7 parts, which can be combined into thousands of possible states.

yushuailong:
> We use only 7 bits from that `uint16_t` bit field now. What about making using `style_cnt :9` (resulting in max 512 styles)?

512 is a large number, but it may still overflow. The other bits of uint16_t can be used for subsequent expansion. What do you think?

kisvegabor:
I'd like to introduce extra memory usage for all widgets only if there is a real need. Even 63 styles worked in 99.99% of the cases. 

If you need thousands of styles it will be very slow to read style properties from them anyways. 

So I vote for using 512.

yushuailong:
I agree with this statement, maybe keeping the status quo is a better choice

kisvegabor:
Why did you close it after modifying to 511?

PR #7830, opened 2025-02-24
fix(obj_style): fix possible memory leak

Simen-z:
Fixed a memory leak in function lv_obj_remove_style. When obj->style_cnt is 0, the original logic realloc a zero byte of memory, but some realloc implementations align the bytes upward instead of directly free, causing a memory leak of one aligned byte.


FASTSHIFT (approved):

kisvegabor:
`lv_realloc` handles it before calling any platform specific allocators. See [here](https://github.com/lvgl/lvgl/blob/master/src/stdlib/lv_mem.c#L151-L155).

PR #7832, opened 2025-02-24
fix(nuttx_fbdev): fix memory buffer handling

W-Mai:
mmap fb mem addr

Improved the memory buffer handling in the framebuffer device driver for NuttX.

This change adjusts the calculation of the buffer offset and ensures that the framebuffer memory is correctly mapped.

consecutive: The **offset** of the `fbmem` address **between** two `displays` in the **same** `panel` is 0.

- For consecutive address cases
    - Calculate the `offset` based on `yoffset`.
    - Since the `fblen` is the size of two buffers when addresses are **consecutive**, it needs to be divided by **2**.
- For non-continuous address cases
    - Use the difference between the `fbmem` address **offsets** of the two displays as the `offset` directly.
    - `fblen` is an independent value for each.

Perform `mmap` operations based on the `offset` and `fblen`.

### Notes

### DIFF
--- a/src/drivers/nuttx/lv_nuttx_fbdev.c
+++ b/src/drivers/nuttx/lv_nuttx_fbdev.c
@@ -328,27 +328,26 @@ static int fbdev_get_pinfo(int fd, FAR struct fb_planeinfo_s * pinfo)
 
 static int fbdev_init_mem2(lv_nuttx_fb_t * dsc)
 {
-    uintptr_t buf_offset;
     struct fb_planeinfo_s pinfo;
-    int ret;
-    void * mem2;
+    void * phy_mem1;
+    void * phy_mem2;
 
+    int ret;
+
+    /* Get display[0] planeinfo */
     lv_memzero(&pinfo, sizeof(pinfo));
+    pinfo.display = dsc->pinfo.display;
+    if((ret = fbdev_get_pinfo(dsc->fd, &pinfo)) < 0) return ret;
+    phy_mem1 = pinfo.fbmem;
 
     /* Get display[1] planeinfo */
-
+    lv_memzero(&pinfo, sizeof(pinfo));
     pinfo.display = dsc->pinfo.display + 1;
+    if((ret = fbdev_get_pinfo(dsc->fd, &pinfo)) < 0) return ret;
+    phy_mem2 = pinfo.fbmem;
 
-    if((ret = fbdev_get_pinfo(dsc->fd, &pinfo)) < 0) {
-        return ret;
-    }
-
-    mem2 = mmap(NULL, pinfo.fblen, PROT_READ | PROT_WRITE, MAP_SHARED | MAP_FILE, dsc->fd, 0);
-
-    if(mem2 == MAP_FAILED) {
-        LV_LOG_ERROR("ERROR: mmap failed: %d\n", errno);
-        return -errno;
-    }
+    lv_uintptr_t offset = (lv_uintptr_t)phy_mem2 - (lv_uintptr_t)phy_mem1;
+    bool is_consecutive = offset == 0;
 
     /* Check bpp */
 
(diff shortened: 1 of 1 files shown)


kisvegabor (dismissed):

src/drivers/nuttx/lv_nuttx_fbdev.c
```diff
@@ -342,6 +343,13 @@ static int fbdev_init_mem2(lv_nuttx_fb_t * dsc)
         return ret;
     }
 
+    fbmem = mmap(NULL, pinfo.fblen, PROT_READ | PROT_WRITE, MAP_SHARED | MAP_FILE, dsc->fd, 0);
+
+    if(fbmem == MAP_FAILED) {
+        LV_LOG_ERROR("ERROR: ioctl(FBIOGET_PLANEINFO) failed: %d\n", errno);
+        return EXIT_FAILURE;
```
liamHowatt: The caller of `fbdev_init_mem2` treats return values `< 0` as an error. `EXIT_FAILURE` is usually `1`. Please change this to -1 or change the caller.

Edit: maybe return `-errno`?

src/drivers/nuttx/lv_nuttx_fbdev.c
```diff
@@ -342,6 +343,13 @@ static int fbdev_init_mem2(lv_nuttx_fb_t * dsc)
         return ret;
     }
 
+    fbmem = mmap(NULL, pinfo.fblen, PROT_READ | PROT_WRITE, MAP_SHARED | MAP_FILE, dsc->fd, 0);
+
+    if(fbmem == MAP_FAILED) {
+        LV_LOG_ERROR("ERROR: ioctl(FBIOGET_PLANEINFO) failed: %d\n", errno);
```
liamHowatt:
```suggestion
        LV_LOG_ERROR("ERROR: mmap failed: %d\n", errno);
```

Please also update the log message at the other `mmap` location too.

liamHowatt (commented):
Makes sense, thanks!

uLipe (dismissed):
Except from @liamHowatt comments LGTM

src/drivers/nuttx/lv_nuttx_fbdev.c
W-Mai: Done

src/drivers/nuttx/lv_nuttx_fbdev.c
W-Mai: Done

liamHowatt (dismissed):
The other error log message for the main `dsc->mem` `mmap` should be updated also, but approving regardless. Thanks!

src/drivers/nuttx/lv_nuttx_fbdev.c
```diff
@@ -342,6 +343,13 @@ static int fbdev_init_mem2(lv_nuttx_fb_t * dsc)
         return ret;
     }
 
+    mem2 = mmap(NULL, pinfo.fblen, PROT_READ | PROT_WRITE, MAP_SHARED | MAP_FILE, dsc->fd, 0);
```
FASTSHIFT: There will be problems with framebuffers with non-contiguous addresses, because `mmap` will always return the address of the first fb. The reason for using `pinfo.fbmem` before is that it can be used to get the address of the second fb, and `pinfo.display` is used to tell the driver which `fbmem` should be reported.

FASTSHIFT (changes requested):

src/drivers/nuttx/lv_nuttx_fbdev.c
liamHowatt: Good catch. I just realized the two `mmap` calls are identical so they would return the same address.

W-Mai: I'll try to fix it later

FASTSHIFT: This is my version:

```c
static int fbdev_init_mem2(lv_nuttx_fb_t * dsc)
{
    struct fb_planeinfo_s pinfo;
    int ret;

    lv_memzero(&pinfo, sizeof(pinfo));

    /* Get display[0] planeinfo */

    if((ret = fbdev_get_pinfo(dsc->fd, &pinfo)) < 0) {
        return ret;
    }

    /* Get the physical address of the first plane */
    void * phy_mem_1 = pinfo.fbmem;

    lv_memzero(&pinfo, sizeof(pinfo));
    pinfo.display = 1;

    if((ret = fbdev_get_pinfo(dsc->fd, &pinfo)) < 0) {
        return ret;
    }

    /* Check bpp */
    if(pinfo.bpp != dsc->pinfo.bpp) {
        LV_LOG_WARN("mem2 bpp is incorrect: %d", pinfo.bpp);
        return -EINVAL;
    }

    /* Calculate the offset of the second plane */
    uintptr_t buf_offset = pinfo.fbmem - phy_mem_1;

    if(buf_offset == 0) {
        dsc->mem2_yoffset = dsc->vinfo.yres;
        dsc->mem2 = mmap(NULL, pinfo.fblen, PROT_READ | PROT_WRITE, MAP_SHARED | MAP_FILE, dsc->fd,
                         dsc->vinfo.yres * dsc->pinfo.stride);
        LV_LOG_USER("Use consecutive mem2 = %p, yoffset = %" LV_PRIu32,
                    dsc->mem2, dsc->mem2_yoffset);
    }
    else {
        /* Check the buffer address offset,
         * It needs to be divisible by pinfo.stride
         */
        if((buf_offset % dsc->pinfo.stride) != 0) {
            LV_LOG_WARN("It is detected that buf_offset(%" PRIuPTR ") "
                        "and stride(%d) are not divisible, please ensure "
                        "that the driver handles the address offset by itself.",
                        buf_offset, dsc->pinfo.stride);
        }

        dsc->mem2_yoffset = buf_offset / dsc->pinfo.stride;
        dsc->mem2 = mmap(NULL, pinfo.fblen, PROT_READ | PROT_WRITE, MAP_SHARED | MAP_FILE, dsc->fd, (off_t)buf_offset);
        LV_LOG_USER("Use non-consecutive mem2 = %p, yoffset = %" LV_PRIu32,
                    dsc->mem2, dsc->mem2_yoffset);
    }

    if(dsc->mem2 == MAP_FAILED) {
        LV_LOG_ERROR("mmap mem2 failed: %d", errno);
        return -errno;
    }

    return 0;
}
```

FASTSHIFT:
> It's amazing that we even wrote code that even had the same variable names and ideas. It's exactly the same as the code I hadn't finished on my computer two hours ago. I'm still testing possible boundary cases for `fb` drivers and memory isolation in the nuttx OS.
> 
> image

Great minds think alike! (“英雄所见略同！”)

W-Mai:
Modifications:

consecutive: The **offset** of the `fbmem` address **between** two `displays` in the **same** `panel` is 0.

- For consecutive address cases
    - Calculate the `offset` based on `yoffset`.
    - Since the `fblen` is the size of two buffers when addresses are **consecutive**, it needs to be divided by **2**.
- For non-continuous address cases
    - Use the difference between the `fbmem` address **offsets** of the two displays as the `offset` directly.
    - `fblen` is an independent value for each.

Perform `mmap` operations based on the `offset` and `fblen`.

kisvegabor:
@liamHowatt can review it in a more meaningful way than I could do, so I pass it to him.

liamHowatt (approved):
I didn't know framebuffer drivers return the same `fbmem` in `fbdev_get_pinfo` if it's contiguous. Everything looks correct from what I can judge.

W-Mai:
> I didn't know framebuffer drivers return the same `fbmem` in `fbdev_get_pinfo` if it's contiguous. Everything looks correct from what I can judge.

Some old drivers of Linux and NuttX do like this. I hav't to follow the rule to maintain forward compatibility 😮‍💨

FASTSHIFT (approved):

PR #7842, opened 2025-02-25
fix(txt): text command escaping

OptionalValue:
Text commands were processed even when correctly escaped.
This resulted in several problems were the width of the text was calculated wrong, causing for example incorrect alignments.

Example:
A right aligned label with the text "XX##YY" would render partially out of the widget as only the "XX" is used to calculate the width of the text to align it. Rest of the text is skipped as it's considered a text command.
Note: the text that is rendered is correct, resulting in the expected text "XX##YY"

### DIFF
--- a/src/misc/lv_text.c
+++ b/src/misc/lv_text.c
@@ -146,7 +146,7 @@ bool lv_text_is_cmd(lv_text_cmd_state_t * state, uint32_t c)
             ret = true;
         }
         /*Other start char in parameter is escaped cmd. char*/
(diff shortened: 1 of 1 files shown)


C47D:
Hi, is it possible to add unit tests for this?

uLipe (commented):
Nice! Could you add a test to ensure covering this case? 

Thank you!

OptionalValue:
Sure, I'll take a look at your current tests and see if I can integrate it :)

PR #7855, opened 2025-02-26
feat(refr): add global recolor

### LINKED ISSUE #7837: Color filter cannot be effective on image
### LVGL version

master

### Platform

Eclipse with SDL on Linux

### What happened?

Color filter cannot take effect on image. If `recolor_opa `is not set for `lv_image`, `color_filter `will not take effect.
modify [here](https://github.com/lvgl/lvgl/blob/728546adb7d7bbce71b177132b0c249e4dd47dcc/src/core/lv_obj_draw.c#L218) can solved my problem. But this may not be reasonable

### How to reproduce?

```
 static lv_color_t darken(const lv_color_filter_dsc_t * dsc, lv_color_t color, lv_opa_t opa)
 {
    LV_UNUSED(dsc);
    return lv_color_darken(color, opa);
 }

static void test(void)
{
    static lv_color_filter_dsc_t color_filter;
    lv_color_filter_dsc_init(&color_filter, darken);

    lv_obj_t * avatar = lv_image_create(lv_screen_active());
    LV_IMAGE_DECLARE(img_demo_widgets_avatar);
    lv_image_set_src(avatar, &img_demo_widgets_avatar);
    lv_obj_set_style_color_filter_dsc(avatar, &color_filter, 0);
    lv_obj_set_style_color_filter_opa(avatar, LV_OPA_50, 0);
    lv_obj_center(avatar);
}

```
Image

yushuailong:
Fixes #7837

```c
static void test(void)
{
    lv_obj_set_style_recolor(lv_screen_active(), lv_color_black(), 0);
    lv_obj_set_style_recolor_opa(lv_screen_active(), LV_OPA_80, 0);

    lv_demo_widgets();
}
```

use layer to cache the global recolor, It can be effective for all widgets.

### Notes


src/core/lv_obj_style.c
```diff
@@ -640,6 +642,40 @@ void lv_obj_update_layer_type(lv_obj_t * obj)
     }
 }
 
+
+lv_color32_t lv_obj_get_style_color_filter_resolved(const lv_obj_t * obj, lv_part_t part)
+{
+    const lv_color_filter_dsc_t * f = lv_obj_get_style_color_filter_dsc(obj, part);
+    if(f && f->filter_cb) {
+        lv_opa_t opa = lv_obj_get_style_color_filter_opa(obj, part);
+        lv_color_t color = f->filter_cb(f, lv_color_black(), LV_OPA_COVER);
```
kisvegabor: It took me 1 hours to understand this line, so I suggest adding a comment here :smile: 
```suggestion
        /*Pass LV_OPA_COVER and any colors. It will return  the pure color that the filter uses to mix the real input color with later*/
        lv_color_t color = f->filter_cb(f, lv_color_black(), LV_OPA_COVER);
```

However, it doesn't work in all cases. The callback can be any complex. E.g. do different things based in the input opacity (< 50% darken, >50% lighten), or based on the color (change the hue).

In light of that I'm not sure if we can cache the color filter's color in the layer.  We can save just a bit indicating if any of the parents had color filter. In most of the cases it will be `0` so we can skip a lot of recursive `style_get` queries.

kisvegabor (commented):
Great implementation, but there one thing that we should consider.

src/core/lv_obj_style.c
yushuailong: Thanks for your reply, I think I encountered some problems, for an image widget, we have its `recolor`, `recolor_opa`, and `filter_cb`, `filter_opa` styles. we ultimately need to set that to `lv_draw_image_dsc_t`, and I'm not sure how to calculate to get the correct result.

For example, if an image widget’s `recolor` and `recolor_opa` are never set, but it gets `filter_cb` and `filter_opa` from its parent, we cannot get the correct result from `filter_cb` (because it needs a color as a parameter).

Do you have any suggestions for this? Should we consider adding a new feature to implement this function and keep the existing `color_filter`?

kisvegabor: What if we keep `color_filter` as it is and make `recolor` not only an image but a generic style property, (`style_recolor` and `style_recolor_opa`). It should be not inherited but calculated recursively and cached in the layer (as it's done in the PR).

`color_filter` is still useful as we can do any magic in it but it's more expansive.  We can start using  `recolor` with default theme too and add a config option to enable/disable `color_filter` as it requires a lot of style queries.

In light of that we can keep the `lv_obj_get_style_bg_color_filtered` and similar functions but if `color_filter` is disabled, `lv_obj_style_apply_color_filter` just returns the color as it is. 

We should also add a helper function, like `lv_obj_style_apply_recolor(layer, obj, color)` which is used in `draw_dsc` init functions.

Regarding images, we can just get the global `recolor` and mix it with the `image_recolor`, and start deprecating `image_recolor`.

What do you think?

yushuailong: I think I understand what you mean, this is a great idea, I will update this PR later :grinning:

yushuailong:
@kisvegabor Hi, could you please review it again when you have time?

W-Mai (approved):

src/core/lv_obj_style.h
```diff
@@ -82,7 +82,7 @@ bool lv_obj_replace_style(lv_obj_t * obj, const lv_style_t * old_style, const lv
  *
  * lv_obj_remove_style(obj, NULL, LV_PART_MAIN | LV_STATE_ANY); //Remove all styles from the main part
  *
- * lv_obj_remove_style(obj, NULL, LV_PART_ANY | LV_STATE_ANY); //Remove all styles
+ * lv_obj_remove_style(obj, NULL, LV_3_ANY | LV_STATE_ANY); //Remove all styles
```
FASTSHIFT: Typo?

FASTSHIFT (changes requested):

src/themes/default/lv_theme_default.c
```diff
@@ -308,16 +292,16 @@ static void style_init(my_theme_t * theme)
     lv_style_set_pad_column(&theme->styles.btn, LV_DPX_CALC(theme->disp_dpi, 5));
     lv_style_set_pad_row(&theme->styles.btn, LV_DPX_CALC(theme->disp_dpi, 5));
 
-    lv_color_filter_dsc_init(&theme->dark_filter, dark_color_filter_cb);
-    lv_color_filter_dsc_init(&theme->grey_filter, grey_filter_cb);
-
     style_init_reset(&theme->styles.pressed);
-    lv_style_set_color_filter_dsc(&theme->styles.pressed, &theme->dark_filter);
-    lv_style_set_color_filter_opa(&theme->styles.pressed, 35);
+    lv_style_set_recolor(&theme->styles.pressed, lv_color_black());
```
kisvegabor, line 296: See `grey_filter_cb`. It uses lighten or darken based on opa. Here we should use `white` or `black` based on `theme->base.flags & MODE_DARK`

lv_conf_template.h
```diff
@@ -1136,6 +1136,8 @@
 /** Enable loading XML UIs runtime */
 #define LV_USE_XML    0
 
+/*1: Enable color filter style*/
+#define LV_USE_COLOR_FILTER     1
```
kisvegabor: Let's make it 0 by default, we will describe it in the release note.

kisvegabor (commented):
Looks great! Just two minor comments.

kisvegabor:
It seems the CI wasn't just a temporarily issue. Do you have any ideas why it fails suddenly?

src/themes/default/lv_theme_default.c
yushuailong:
```c
static lv_color_t dark_color_filter_cb(const lv_color_filter_dsc_t * f, lv_color_t c, lv_opa_t opa)
{
    LV_UNUSED(f);
    return lv_color_darken(c, opa);
}

lv_color_filter_dsc_init(&theme->dark_filter, dark_color_filter_cb);

lv_style_set_color_filter_dsc(&theme->styles.pressed, &theme->dark_filter);

```
Before I changed it, the pressed state did not check `theme->base.flags & MODE_DARK`, it was always `black`, should we need change it?

yushuailong:
> It seems the CI wasn't just a temporarily issue. Do you have any ideas why it fails suddenly?

I don't know why CI failed. Now other PR also encounter this problem.

FASTSHIFT (approved):

uLipe (approved):

src/misc/lv_style.c
```diff
@@ -415,6 +417,7 @@ lv_style_value_t lv_style_prop_get_default(lv_style_prop_t prop)
         case LV_STYLE_LINE_COLOR:
         case LV_STYLE_TEXT_COLOR:
         case LV_STYLE_IMAGE_RECOLOR:
+        case LV_STYLE_RECOLOR:
```
liamHowatt, line 420: Should a case for `LV_STYLE_RECOLOR_OPA` be added too?

src/themes/default/lv_theme_default.c
liamHowatt: @kisvegabor do you have any more comments on this thread?

liamHowatt (approved):
Great work!

@XuNeo were the style properties updated correctly?

src/misc/lv_style.c
yushuailong: I think `LV_STYLE_RECOLOR_OPA `should have the same default value of 0 as `LV_STYLE_IMAGE_RECOLOR_OPA`, so there is no case added. What do you think?

liamHowatt: Ah I see, it's zeroed if not handled explicitly. Got it.

src/themes/default/lv_theme_default.c
kisvegabor: It was intentional to always make the pressed state darker so let's keep it as it is.

kisvegabor:
Sorry for the delay. Please fix [this last thing](https://github.com/lvgl/lvgl/pull/7855#discussion_r2009071015), and we are good to go.

yushuailong:
> Sorry for the delay. Please fix [this last thing](https://github.com/lvgl/lvgl/pull/7855#discussion_r2009071015), and we are good to go.

hi, long time no see, we don't need to modify it, I have resolved it

kisvegabor:
Awesome, thank you!

kisvegabor:
I was benchmarking how many style property gets we can save with `LV_USE_COLOR_FILTER 0`, but it's not as much as I hoped because  `LV_OBJ_STYLE_CACHE` skips a lot of style gets earlier. 

Anyway I think it's pretty useful feature. :rocket:

PR #7856, opened 2025-02-26
fix(port_releaser): always use remote when pushing and add token

AndreCostaaa:
This fixes the current CI problem, but won't fix the whole script ~~I believe~~ by just merging it

Since the script is pulling using `https`, github will ask for a username and a token. The `GITHUB_TOKEN` that github provides you inside the workflow won't allow you to push to a different repository even if it's part of the same organization

The way I would solve this is by generating a `fine-grained token`  associated with the `lvgl` organization with `write permissions` on the repositories you wish.

For this: 
1. Click on your profile icon on the top right
2. Go to settings
3. Scroll all the way down to  `Developer Settings`
4. `Personal Tokens` > `Fine-Grained Tokens` > `Generate New Token`
5. Select the `lvgl` organization
6. Select `All Repositories` or `Only Select Repositories` based on preference. I suggest `Only Select Repositories` since it's safer.
    - The `lvgl/lvgl` repository doesn't need to be added here, only the `lvgl/lv_port_XXX` repositories that you wish to update
7. Add the `Contents` read + write permission

Now add this token as a github secret to this repository

1. Go to this repository's settings 
2. `Secretes and Variables` > `Actions` > `New Repository Secret`
3. Add the token as a secret. Name it `PORT_RELEASER_GITHUB_TOKEN`

Now this new token can be used to clone the repositories with the correct permissions which will allow you to write i.e push to them by cloning them in this format `https://{token}@github.com/lvgl/lv_port...`

This PR updates the script according to my explanations above

## Some security concerns

**This secret could be leaked**. Github checks for the scripts outputs matching the secrets and redacts it but a malicious person could still get his hands on it if he tries really hard. See [here](https://docs.github.com/en/actions/security-for-github-actions/security-guides/security-hardening-for-github-actions).
 
Since this workflow is only ran when a merge to master happens, the reviewers just need to pay extra attention to the changes made to this script/ workflow and not merge anything that seems suspicious

~~Additionally, someone could potentially change the workflow and the CI to trigger on a PR, and leak the token this way, for this, the best way would be to not allow workflows to run when the workflow file is changed inside a PR. This can be done by using the `CODEOWNERS` feature [here](https://docs.github.com/en/actions/security-for-github-actions/security-guides/security-hardening-for-github-actions)~~

The whole interaction with the secrets, scripts and workflows make this impossible to test reliably but I believe this PR fixes or at least gets you closer to having a working workflow

Hope this helps. Let me know if you have any questions

@kisvegabor @liamHowatt

### DIFF
--- a/.github/workflows/release_branch_updater.yml
+++ b/.github/workflows/release_branch_updater.yml
@@ -1,8 +1,6 @@
 name: Port repo release update
 
 on:
-  pull_request:
-    branches: [ master ]
   push:
     branches:
       - master
--- a/.github/workflows/release_branch_updater.yml
+++ b/.github/workflows/release_branch_updater.yml
@@ -1,6 +1,8 @@
 name: Port repo release update
 
 on:
+  pull_request:
+    branches: [ master ]
   push:
     branches:
       - master
--- a/.github/workflows/release_branch_updater.yml
+++ b/.github/workflows/release_branch_updater.yml
@@ -31,4 +31,6 @@ jobs:
           git config --global user.email 'lvgl-bot@users.noreply.github.com'
 
       - name: run the script
-        run: python3 lvgl/scripts/release_branch_updater.py --oldest-major 9
+        with:
+          GITHUB_TOKEN: ${{ secrets.PORT_RELEASER_GITHUB_TOKEN }}
+        run: python3 lvgl/scripts/release_branch_updater.py --oldest-major 9 --github_token "$GITHUB_TOKEN"
--- a/.github/workflows/release_branch_updater.yml
+++ b/.github/workflows/release_branch_updater.yml
@@ -35,4 +35,4 @@ jobs:
       - name: run the script
         env:
           GITHUB_TOKEN: ${{ secrets.PORT_RELEASER_GITHUB_TOKEN }}
-        run: python3 lvgl/scripts/release_branch_updater.py --oldest-major 9 --github_token "$GITHUB_TOKEN"
+        run: python3 lvgl/scripts/release_branch_updater.py --oldest-major 9 --github-token "$GITHUB_TOKEN"
--- a/.github/workflows/release_branch_updater.yml
+++ b/.github/workflows/release_branch_updater.yml
@@ -33,6 +33,6 @@ jobs:
           git config --global user.email 'lvgl-bot@users.noreply.github.com'
 
       - name: run the script
-        with:
+        env:
           GITHUB_TOKEN: ${{ secrets.PORT_RELEASER_GITHUB_TOKEN }}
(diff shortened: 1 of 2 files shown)


liamHowatt (commented):
Deferring to @kisvegabor to create the `PORT_RELEASER_GITHUB_TOKEN`. I tried and do not have the ability.

It looks great! And wow, the `-u` flag really isn't required. I did not know. Great catch with the output blocking.

kisvegabor:
I've just created `PORT_RELEASER_GITHUB_TOKEN`.

AndreCostaaa:
Any way to trigger the workflow manually for this PR so we can check?

liamHowatt (commented):
I guess adding `pull_request:` under `on:` wouldn't work, because your fork doesn't have the token. Maybe @kisvegabor can copy your branch under his https://github.com/kisvegabor/lvgl_fork to let it run there?

AndreCostaaa:
> I guess adding pull_request: under on: wouldn't work, because your fork doesn't have the token

I believe it would. In theory a PR open by someone has access to the repository secrets in order for the workflow to run

I can add a temporary commit to trigger the workflow and then revert it but I would prefer triggering it manually since it has the `workflow_dispatch` option


It won't work in @kisvegabor's fork because the secret is associated with this repository unfortunately

AndreCostaaa:
Oh so TIL that github doesn't give access to secrets that are triggered by pull requests from forked repositories, [here](https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#workflows-in-forked-repositories)

I guess what I said before is wrong :
> Additionally, someone could potentially change the workflow and the CI to trigger on a PR, and leak the token this way

Even if someone changes the workflow to trigger on a PR, github won't let the workflow access the secret. You can see that the token doesn't exist in this [run](https://github.com/lvgl/lvgl/actions/runs/13653626641/job/38167620596) for instance

```bash
[release_branch_updater.py] Warning: No github token was provided for this production run. Continuing anyway...
```

I fixed some extra stuff about the script and reverted the run on pull requests
I'm leaving the revert commit so you can easily find the previous commits where the runs failed because of the missing secret since this PR can get merge-squashed anyway

Guess we'll know if everything works when this gets merged

liamHowatt (approved):
I had the same though process, but thought it would be unlikely to let anybody print the token in their PR CI.

Good catches with `env` and `--github-token`!

kisvegabor:
Just to confirm, anyone can send a PR in which they print any secrets and the y will see them in the logs? 

It's not the case, right?

AndreCostaaa:
Exactly, it cant't happen. Workflows triggered by pull requests don't have access to repository secrets

kisvegabor (approved):
All good then! Let's see what happens after merging.

AndreCostaaa:
Success ! 😁 🚀

PR #7859, opened 2025-02-27
fix(jpg):More rational identification

cnjhb:
The header of the jpg is just 0xffd8,the current way of recognizing it is not proper.

Fixes #xxxx 

### Notes

### DIFF
--- a/src/extra/libs/sjpg/lv_sjpg.c
+++ b/src/extra/libs/sjpg/lv_sjpg.c
@@ -885,7 +885,7 @@ static void decoder_close(lv_img_decoder_t * decoder, lv_img_decoder_dsc_t * dsc
 
 static int is_jpg(const uint8_t * raw_data, size_t len)
 {
-    const uint8_t jpg_signature[] = {0xFF, 0xD8, 0xFF,  0xE0,  0x00,  0x10, 0x4A,  0x46, 0x49, 0x46};
+    const uint8_t jpg_signature[] = {0xFF, 0xD8, };
     if(len < sizeof(jpg_signature)) return false;
     return memcmp(jpg_signature, raw_data, sizeof(jpg_signature)) == 0;
 }


src/extra/libs/sjpg/lv_sjpg.c
```diff
@@ -885,7 +885,7 @@ static void decoder_close(lv_img_decoder_t * decoder, lv_img_decoder_dsc_t * dsc
 
 static int is_jpg(const uint8_t * raw_data, size_t len)
 {
-    const uint8_t jpg_signature[] = {0xFF, 0xD8, 0xFF,  0xE0,  0x00,  0x10, 0x4A,  0x46, 0x49, 0x46};
+    const uint8_t jpg_signature[] = {0xFF, 0xD8, };
```
XuNeo, line 888: Should we verify both SOI + Identifier("JFIF\0")?

```
typedef struct _JFIFHeader
{
  BYTE SOI[2];          /* 00h  Start of Image Marker     */
  BYTE APP0[2];         /* 02h  Application Use Marker    */
  BYTE Length[2];       /* 04h  Length of APP0 Field      */
  BYTE Identifier[5];   /* 06h  "JFIF" (zero terminated) Id String */
  BYTE Version[2];      /* 07h  JFIF Format Revision      */
  BYTE Units;           /* 09h  Units used for Resolution */
  BYTE Xdensity[2];     /* 0Ah  Horizontal Resolution     */
  BYTE Ydensity[2];     /* 0Ch  Vertical Resolution       */
  BYTE XThumbnail;      /* 0Eh  Horizontal Pixel Count    */
  BYTE YThumbnail;      /* 0Fh  Vertical Pixel Count      */
} JFIFHEAD;
```

cnjhb: you are right

PGNetHun: It would be better to verify these 2 parts (practically exclude APP0 length field):
1. SOI + APP0 marker (4 bytes: [0..3])
2. "JFIF\0" identifier (5 bytes: [6..10])

If it is implemented in an "early return" way (meaning that comparing the first 4 bytes (32 bits) with a constant), then recognizing a non-JPEG image would be very fast on 32 bit processors (because not comparing bytes one-by-one).
If it passes (SOI + APP0 marker) found, then just check for "JFIF\0".
APP0 length field would be ignored in this case.

Opinion?

PGNetHun (changes requested):

src/extra/libs/sjpg/lv_sjpg.c
cnjhb: > It would be better to verify these 2 parts (practically exclude APP0 length field):
> 
>     1. SOI + APP0 marker (4 bytes: [0..3])
> 
>     2. "JFIF\0" identifier (5 bytes: [6..10])
> 
> 
> If it is implemented in an "early return" way (meaning that comparing the first 4 bytes (32 bits) with a constant), then recognizing a non-JPEG image would be very fast on 32 bit processors (because not comparing bytes one-by-one). If it passes (SOI + APP0 marker) found, then just check for "JFIF\0". APP0 length field would be ignored in this case.
> 
> Opinion?

You're right too, but I don't have much time at the moment.

liamHowatt:
The v8.4 CI has been fixed. Please rebase so the fixed CI can run.

PR #7927, opened 2025-03-12
Dropped touchscreen input bugfix

### LINKED ISSUE #5211: Touch screen input events missed

thomaskbrooksee:
Fixes LVGL issue #5211 on my device, based on LVGL PR #5662. Touchscreen inputs are being dropped/ignored on low-framerate devices when LVGL detects a press and release in the same frame. We were able to use evtest to confirm that inputs for quick taps were being dropped.

LVGL's evdev event loop already uses a boolean flag to handle this issue for keypresses and mouse clicks, but it's not being used for screen taps. This fix improves the accuracy of tapping buttons and swiping on the screen.

5211 and 5662 reported that the previous fix introduced issues with dragging on certain devices. I have not been able to reproduce these issues in our own app.


uLipe:
@thomaskbrooksee is it ready for reviewing? It seems this PR is bringing lots of unrelated commits, could you check this commit history and make sure that only PR relevant commits are being pushed?

Thank you :)

uLipe (changes requested):
Given a temporary -1 until the commit history gets addressed, happy to review it when ready

thomaskbrooksee:
Umm... This is my first time doing a PR, please bear with me...

I'm pretty sure I branched off of the v9.2.2 tag. So I should merge it into master?

liamHowatt:
Please base your branch on `master`.

PR #7934, opened 2025-03-14
arch(cmake): Add native Kconfig support to cmake

### LINKED ISSUE #7223: Rework and unify CMake

Daxidz:
Fixes #7223 (first part of the work)

It allow to build lvgl with Kconfig directly, without the need of an integration on top. It also creates the CONFIG_LV_USE_* as cmake variable, to be used in future work.

Allow to use the `-DLV_USE_KCONFIG` cmake option to use Kconfig instead of `lv_conf.h`.
It now uses a preprocessed `lv_conf_internal.h` to extract the `CONFIG_LV_USE_*` and offer them as cmake variables. 

### Notes
  -  To be discussed what to update/add


Daxidz:
@kisvegabor I updated the documentation with the latest changes. It seems the CI is currently broken on the lv_bindings task.

docs/src/details/integration/building/cmake.rst
```diff
@@ -70,11 +77,27 @@ If you use cmake to install lvgl 3 folders will be created.
 
 Build with Command line
 -----------------------
+You can also build your project using the command line. Run the following commands to use the presets:
+
+.. code-block:: bash
+
+    cmake --preset windows-base
```
kisvegabor: Please describe in 1-2 sentences what is a CMake preset. What other common presets are available?

docs/src/details/integration/building/cmake.rst
```diff
@@ -70,11 +77,27 @@ If you use cmake to install lvgl 3 folders will be created.
...
+    ctest --preset windows-base_dbg
+
 
-You can also build your project using the command line. Run the following commands
+To not use preset, use these commands:
 
-- ``cmake --preset windows-base``
-- ``cmake --build --preset windows-base_dbg``
-- ``ctest --preset windows-base_dbg``
+.. code-block:: bash
```
kisvegabor: I suggest starting with these instead of presets and these commands are more generic.

docs/src/details/integration/building/cmake.rst
```diff
@@ -70,11 +77,27 @@ If you use cmake to install lvgl 3 folders will be created.
...
+    cd <lvgl_repo>
+    mkdir build
+    cd build
+    cmake ..            # Configure phase
+    cmake --build .     # Build phase
 
+    # Method 2
+    cd <lvgl_repo>
+    cmake -B build      # Configure phase
+    cmake --build build # build phase
```
kisvegabor: Please describe how to add lvgl into a parent CMake project. (including the the demos, examples and thorvg libraries)

kisvegabor (commented):
I've tested the Kconfig part and it worked very well! Just a few comments on the docs.

kisvegabor:
Please also add these to `.gitignore`:
```
	.config
	config.h
	defconfig
```

Daxidz:
@kisvegabor documentation was updated according to the remarks, I let you check

docs/src/details/integration/building/cmake.rst
```diff
@@ -19,22 +25,63 @@ You need to install
...
+    cmake --build .     # Build phase
 
-More configurations will be added once available.
+    # Method 2
+    cd <lvgl_repo>
+    cmake -B build      # Configure phase
+    cmake --build build # build phase
+
+
+You can also use the presets:
```
kisvegabor: I think preset related build instructions should be documented at the section of the presets.

kisvegabor (commented):
Thanks, just one minor comment.

docs/src/details/integration/building/cmake.rst
Daxidz: Agreed, it is now better organized

kisvegabor (dismissed):
Looks good, thank you!

scripts/generate_cmake_variables.py
```diff
@@ -0,0 +1,85 @@
...
+    fout = open(path_output, "w", newline='')
+
+    # If we use Kconfig, we must check the CONFIG_LV_USE_* defines
+    if kconfig:
+        CONFIG_PATTERN="#define CONFIG_LV_USE"
+        CONFIG_PREFIX=""
+    # Otherwise check the LV_USE_* defines
+    else:
+        CONFIG_PATTERN="#define LV_USE"
+        CONFIG_PREFIX="CONFIG_"
```
liamHowatt, line 51: What about options that don't start with `LV_USE`? I.e. `LV_FONT_FMT_TXT_LARGE`

scripts/generate_cmake_variables.py
```diff
@@ -0,0 +1,85 @@
...
+
+            conf = line.split(" ")
+
+            name = conf[1]
+
+            value = line.split(name, 1)[1]
+
+            # Only write enabled configs
+            if value.endswith(f'1'):
+                fout.write(f'set({CONFIG_PREFIX}{name} {value})\n')
```
liamHowatt: What about lines like this
```c
#define LV_USE_ASSERT_NULL          1   /**< Check if the parameter is NULL. (Very fast, recommended) */
```

configs/defconfigs/minimal_defconfig
liamHowatt, line 1: I'm seeing an issue that I don't understand.

```
cmake -B build -DPython_EXECUTABLE:FILEPATH=/home/liam/lvgl/lvgl_py_venv/bin/python -DLV_USE_KCONFIG=ON -DLV_DEFCONFIG_PATH=configs/defconfigs/minimal_defconfig
cmake --build build
```

```
In file included from /home/liam/lvgl/lvgl/src/draw/vg_lite/../../../lvgl.h:76,
                 from /home/liam/lvgl/lvgl/src/draw/vg_lite/../../lvgl.h:16,
                 from /home/liam/lvgl/lvgl/src/draw/vg_lite/lv_draw_vg_lite_label.c:10:
/home/liam/lvgl/lvgl/src/draw/vg_lite/../../../src/widgets/slider/lv_slider.h:22:2: error: #error "lv_slider: lv_bar is required. Enable it in lv_conf.h (LV_USE_BAR 1)"
   22 | #error "lv_slider: lv_bar is required. Enable it in lv_conf.h (LV_USE_BAR 1)"
      |  ^~~~~
```

Bar is not enabled according to both `minimal_defconfig` and the generated `lv_conf_expanded.h`

```
# CONFIG_LV_USE_BAR is not set
```

```
#define LV_USE_BAR 0
```

However when I go into `menuconfig` to enable it, it's already checked.

image

liamHowatt (commented):
Wow, nice!

scripts/generate_cmake_variables.py
Daxidz: For now I only convert LV_USE_* configs, as they are the one used by the build system. The other ones are used by the code itself, which make it less useful to also make them available in cmake. But I can change this to convert **all** configs.

configs/defconfigs/minimal_defconfig
Daxidz: The issue is that the current system does not regenerate a `.config` when using a defconfig via `-DLV_DEFCONFIG_PATH`. `menuconfig` will work on `.config`, which is why you see it set here (probably from a previous configuration phase) but actually it does not reflect the defconfig which was used. I'll fix this.

scripts/generate_cmake_variables.py
Daxidz: Nice catch, I'll use the `conf[2]` as the value

configs/defconfigs/minimal_defconfig
Daxidz: Fixed. It now generates `.config` in the lvgl directory directly, there was a missmatch between the input .config and the output .config.
It also makes it possible to now run a first cmake configuration using `-DLV_DEFCONFIG_PATH=...`, then modify the resulting `.config` using menuconfig and then run the cmake configuration without `-DLV_DEFCONFIG_PATH`, which will make it use the `.config`.

scripts/generate_cmake_variables.py
Daxidz: I changed this to convert **all** configuration to cmake variables. It now uses the `conf[2]` as the value which simplify the code and better handles the corner cases like you described

scripts/generate_cmake_variables.py
Daxidz: Actually, the other configs are much harder to parse, as they may be some strings, macros, or contain spaces. We must decide if we need those configs as cmake variables, or if only having LV_USE_* config as cmake variable is enough.

configs/defconfigs/minimal_defconfig
liamHowatt: Nice, thank you. When I run `savedefconfig`, what do you see in your new generated `defconfig`?

Daxidz: It should be the non default values (from the base Kconfig parsing), with the value assigned in the `.config` used while executing `savedefconfig`.

kisvegabor (dismissed):

kisvegabor (approved):

uLipe (approved):

PR #7949, opened 2025-03-18
ci: emulated perf workflow

AndreCostaaa:
This PR adds a new workflow to run performance benchmarks in an **emulated ARM environment**

LVGL runs on top of [so3](https://github.com/smartobjectoriented/so3), a lightweight operating system. The docker image used is provided by `so3` with a special config for theses tests. You can read all the details about the images used [here](https://github.com/smartobjectoriented/so3/blob/main/docker/README.md). The file also explains how you can run this image locally 

This PR adds:
 
- **Script Checks**: Ensures the workflow scripts are working before execution.
- **Benchmark Workflow**: Runs performance tests in an emulated setup.    
- **Stable Timing**: Since it’s emulated, results don’t depend on the runner’s hardware. 
   - I've personally tested running on a 2 Core Celeron N3060 @ 2.48 GHz, and while the app takes ~8 minutes to run, the results are the same as in my 12 Core Ryzen 5 3600 running @ 4.20 GHz.
   - In a standard github runner, the whole workflow takes ~3 minutes to run 
- **PR Comments**: If this runs on a PR, it’ll compare results with previous benchmarks (if they exist) and comment on the findings.  
- **Master**: Stores results in a "pre-release" when merged to `master`.  

### Why This?  
To make sure performance stays in check and we can **track regressions easily**. No more guessing if something slowed down

You can check [this repository](https://github.com/AndreCostaaa/lvgl-copy) to see what it looks like

These two PRs [1](https://github.com/AndreCostaaa/lvgl-copy/pull/1) [2](https://github.com/AndreCostaaa/lvgl-copy/pull/2) showcase the PR part. Including the comparison of results between `master` and PR 2.

This is the ["pre-release"](https://github.com/AndreCostaaa/lvgl-copy/releases/tag/emulated-benchmark-latest) this workflow adds

Not sure if we need to add some documentation about this somewhere, i've tried to be very explicit and added some documentation to the different python scripts

Let me know what you think

@kisvegabor @liamHowatt (cc @daniel-rossier)


AndreCostaaa:
Seem like there’s a permission problem when trying to write the PR comment

AndreCostaaa:
I was able to figure it out

When running a workflow triggered by the `pull_request` event no permission is given to the `GITHUB_TOKEN`

So in order to make it work i've separated the "running the benchmark" and the "checking results" parts 
The results are passed as artifacts between the workflows

Right now the second workflow isn't triggered because the workflow file must exist in the `master` branch which it doesn't yet. I could split this PR into two 

1. Add the workflow that checks the results, creates the PR comment and the Pre-Release
2. Add the workflow that runs the benchmark and exports the results

This way, when PR 2 is triggered, the workflow will already exist in `master` and it will be triggered, causing it to create the PR comment which would also create the "baseline" results

Let me know what you prefer !

AndreCostaaa:
That's right :)

AndreCostaaa:
Hey just pushed some updates: 

- Updated the docker image. The new version adds `python`, `pcpp` and removes a lot of verbosity.
  - #8090 removes `pcpp` but it's not a big dependency so it doesn't hurt to have it in the image
  - Based on the discussion in #8064, i believe I will have to update the image again to point to lvgl io instead of so3 
- Added a way to easily add new dependencies to the docker image via a shell script
- Added a disclaimer to the PR comment that gets generated
- Very rarely, the script fails because of a run condition when creating a loop device, see https://github.com/lvgl/lvgl/actions/runs/14432699079/job/40469393600 so to circumvent failing the CI in case that happens I added some retry logic because it should never fail twice in a row

kisvegabor:
Thank you Andre!
Let's get back to this after the release of v9.3.

liamHowatt (dismissed):
Cool sanity check step. Everything I could see made sense. This is super cool!

One thing I noticed, it looks like

`scripts/perf/tests/serialize_results/input/results-32b-lv_conf_perf32b.json` and `scripts/perf/tests/serialize_results/input/results-32b-lv_conf_perf32b_a.json` are duplicates and

`scripts/perf/tests/serialize_results/input/results-32b-lv_conf_perf32b.mpk` and `scripts/perf/tests/serialize_results/input/results-32b-lv_conf_perf32b_b.mpk` are duplicates.

It's fine, is it to test it against multiple inputs?

AndreCostaaa:
> Cool sanity check step. Everything I could see made sense. This is super cool!

Thanks Liam :smile: 

> One thing I noticed, it looks like
> 
> `scripts/perf/tests/serialize_results/input/results-32b-lv_conf_perf32b.json` and `scripts/perf/tests/serialize_results/input/results-32b-lv_conf_perf32b_a.json` are duplicates and
> 
> `scripts/perf/tests/serialize_results/input/results-32b-lv_conf_perf32b.mpk` and `scripts/perf/tests/serialize_results/input/results-32b-lv_conf_perf32b_b.mpk` are duplicates.
> 
> It's fine, is it to test it against multiple inputs?

Yep the script handle the three possible cases:

1. json and mpk both exist - So there are new results that need to get added to the mpk file
2. json exist but mpk doesn't exist - New config, need to generate a new mpk file
3. json doesn't exist but mpk exists - Old config, we have history but there are no more results

The `_a.json` is used to test case 2 and the `_b.mpk` is used to test case 3. I just copied and renamed them so that's why they are identical

AndreCostaaa:
> Cool sanity check step. Everything I could see made sense. This is super cool!

Thanks Liam :smile: 

> One thing I noticed, it looks like
> 
> `scripts/perf/tests/serialize_results/input/results-32b-lv_conf_perf32b.json` and `scripts/perf/tests/serialize_results/input/results-32b-lv_conf_perf32b_a.json` are duplicates and
> 
> `scripts/perf/tests/serialize_results/input/results-32b-lv_conf_perf32b.mpk` and `scripts/perf/tests/serialize_results/input/results-32b-lv_conf_perf32b_b.mpk` are duplicates.
> 
> It's fine, is it to test it against multiple inputs?

Yep the script handles the three possible cases:

1. json and mpk both exist - So there are new results that need to get added to the mpk file
2. json exist but mpk doesn't exist - New config, need to generate a new mpk file
3. json doesn't exist but mpk exists - Old config, we have history but there are no more results

The `_a.json` is used to test case 2 and the `_b.mpk` is used to test case 3. I just copied and renamed them so that's why they are identical

kisvegabor:
@AndreCostaaa please add some docs (README?) to describe how to run it locally.

AndreCostaaa:
Sure. Inside the tests/README file?

kisvegabor:
Sounds good to me

tests/README.md
```diff
@@ -71,3 +71,33 @@ There are some custom, LVGL specific asserts:
...
+```
+
+2. Run the benchmark demo
+
+```bash
+docker run -t --privileged \
+   --name lv_perf_test_32b \
+   -v /dev:/dev \
+   -v $(pwd)/lv_conf_perf32b.h:/so3/usr/lib/lv_conf.h \
+   -v $(pwd):/so3/usr/lib/lvgl \
```
liamHowatt:
```suggestion
   -v $(pwd):/so3/usr/lvgl \
```

to avoid

```
CMake Error at lib/lvgl/env_support/cmake/os_desktop.cmake:126 (message):
  Configuration file: /so3/usr/lv_conf.h - not found
  lib/lvgl/CMakeLists.txt:23 (include)
```

AndreCostaaa: Actually I need to get this merged for it to be fixed here: https://github.com/smartobjectoriented/so3/pull/166

What is happening is that since `LV_CONF_PATH` is not set, the new cmake update tries to find `lv_conf.h` alongside the root CMakeLists.txt which, in case of so3, is inside `so3/usr`

https://github.com/lvgl/lvgl/blob/a051ee4e7971b8409e4f7d140bf45a165315773b/env_support/cmake/os_desktop.cmake#L115

I could also modify line 98 to mount `lv_conf` inside `/so3/usr` directly and it would work but would like to override the actual `lv_conf.h` from so3 and not hack through a quick fix

AndreCostaaa:
https://github.com/smartobjectoriented/so3/pull/166 has now been merged and this PR has been updated and ready for another round of reviews @liamHowatt @kisvegabor .
I modified the image version, removed the retry logic from the CI as it should no longer be necessary and modified the docs so it made more sense after the other PR got merged. You can see all of that in the latest two commits bf77d58 and 34e096e

AndreCostaaa:
https://github.com/smartobjectoriented/so3/pull/166 has now been merged and this PR has been updated and ready for another round of reviews @liamHowatt @kisvegabor .
I modified the image version, removed the retry logic from the CI as it should no longer be necessary and modified the docs so it made more sense after the other PR got merged. You can see all of that in the latest two commits bf77d58 and 34e096e

scripts/perf/tests/benchmark_results_comment/no_mpk/expected.md
```diff
@@ -0,0 +1,83 @@
+Hi :wave:, thank you for your PR!
```
kisvegabor, line 1: What does `no_mpk` mean?

tests/README.md
```diff
@@ -44,6 +44,34 @@ docker run --rm -it -v $(pwd):/work lvgl_test_env "./tests/main.py"
 
 This ensures you are testing in a consistent environment with the same dependencies as the CI pipeline.
 
+## Running automatically
+
+GitHub's CI automatically runs these tests on pushes and pull requests to `master` and `releasev8.*` branches.
```
kisvegabor:
```suggestion
GitHub's CI automatically runs these tests on pushes and pull requests to `master` and `release/v8.*` branches.
```

tests/README.md
```diff
@@ -78,31 +107,36 @@ For full usage options, run:
...
-   - If the compare fails an `<image_name>_err.png` file will be created with the rendered content next to the reference image.
-- `TEST_ASSERT_EQUAL_COLOR(color1, color2)` Compare two colors.
+```bash
+docker run -t --privileged \
+   --name lv_perf_test_32b \
+   -v /dev:/dev \
+   -v $(pwd)/lv_conf_perf32b.h:/so3/usr/lib/lv_conf.h \
+   -v $(pwd):/so3/usr/lib/lvgl \
+   ghcr.io/smartobjectoriented/so3-lvperf32b:main
+```
```
kisvegabor, line 146: Can you add a python script that hides all the details? E.g.
```
benchmark_emu.py --conf perf32b
benchmark_emu.py run
```

kisvegabor (commented):
I've just checked the high level things. We will see better how it works in the practice.

We have som many tests now, that a bullet point summary at the beginning of teh README would be great too.

scripts/perf/tests/benchmark_results_comment/no_mpk/expected.md
AndreCostaaa: MessagePack is the file format used to store the performance history: https://msgpack.org/index.html
There's one `mpk` file per config so this the expected comment when such file doesn't exist (i.e. we don't have a history file for that specific config)

tests/README.md
AndreCostaaa: Good idea !

AndreCostaaa:
Concerns should be handled in latest 3 commits @kisvegabor

kisvegabor:
It works well, thank you! I just got these warnings:
```
Warning: The guest is now late by 0.0 to 1.0 seconds
Warning: The guest is now late by 1.0 to 2.0 seconds
Warning: The guest is now late by 2.0 to 3.0 seconds
Warning: The guest is now late by 1.0 to 2.0 seconds
Warning: The guest is now late by 2.0 to 3.0 seconds
Warning: The guest is now late by 3.0 to 4.0 seconds
Warning: The guest is now late by 4.0 to 5.0 seconds
Warning: The guest is now late by 5.0 to 6.0 seconds
```

I don't know if it's normal.

kisvegabor (approved):

AndreCostaaa:
It is! Since QEMU is slowed down it's basically telling you the difference in the emulation vs real world

uLipe (approved):

AndreCostaaa:
Thank you for approving. I'll merge this now as is so we can iterate on it to make it perfect 👌

PR #7954, opened 2025-03-19
fix(Widgets): Optimise widget event callbacks to remove unnecessary calls to the event handler.

pete-pjb:
Some Background:
I have an application which has many nested tabviews running on a full HD screen and I noticed there were a massive amount of calls to the tabview event handler going on.  Upon investigation I found the 'tabview', 'tileview' and 'textarea' widgets' constructor functions route **ALL EVENTS** to the event callback.  Edited the  code to just route the events the handler needs to handle for each widget.  Saves on many unnecessary function calls and pointer lookup and assignments.

### DIFF
--- a/.github/workflows/build_micropython.yml
+++ b/.github/workflows/build_micropython.yml
@@ -33,7 +33,7 @@ jobs:
     - name: Clone lv_micropython
       run: |
         git clone https://github.com/lvgl/lv_micropython.git .
-        git checkout master
+        git checkout bf110742007ed2519e0d688aa58ef5c2f0acfaf0
     - name: Initialize lv_bindings submodule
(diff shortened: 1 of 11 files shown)


pete-pjb:
Not sure why the Code Formatting failed I ran the script....

.github/workflows/build_micropython.yml
PGNetHun, line 1: You can discard changes in this file, as MicroPython CI build is fixed and merged to main branch.

pete-pjb: I have reverted the file now... Thanks.

kisvegabor (approved):
Very reasonable update! Thank you!

liamHowatt (approved):

PR #7955, opened 2025-03-19
feat(image): add LV_IMAGE_ALIGN_CONTAIN & LV_IMAGE_ALIGN_COVER to scale images without changing aspect ratio

AndyEveritt:
Adds a new `lv_image_align_t` mode called `LV_IMAGE_ALIGN_AUTO_SCALE` which will behave similar to `LV_IMAGE_ALIGN_STRETCH` but does not modify the aspect ratio of the image. ie the image will be scaled the same in x & y so that the image fits the size of the object.

By default the image will be centered if it's aspect ratio does not match the aspect ratio of the object. This can be adjusted with `lv_image_set_offset_x()` and `lv_image_set_offset_y()`

image

Also adds a few helper methods to `lv_image.h`
- `lv_image_get_width(lv_obj_t* obj)`: Gets the width of the image before any transformations
- `lv_image_get_height(lv_obj_t* obj)`: Gets the height of the image before any transformations
- `lv_image_get_scaled_width(lv_obj_t* obj)`: Gets the width of the image after transformations have been applied
- `lv_image_get_scaled_height(lv_obj_t* obj)`: Gets the height of the image after transformations have been applied

> Note that the above methods refer to the actual image being displayed not the `lv_obj_t*` object

An example of where the helper methods would be useful:
```cpp
lv_obj_t* obj = lv_image_create();
lv_image_set_inner_align(obj, LV_IMAGE_ALIGN_AUTO_SCALE);  // set the image to scale without changing apect ratio
lv_image_set_src(obj, PATH_TO_IMAGE);
lv_image_set_offset_x(obj, (lv_obj_get_width(obj) - lv_image_get_scaled_width(obj)) / 2)  // sets the offset so that the image is positioned center right
```

### Notes


PGNetHun (commented):
Nice work!

Could you please add one test with the 2 cases also to see that scaling on both axis works as expected?
1. image width > height (wide image)
2. image height > width (tall image)

Image widget tests:
https://github.com/lvgl/lvgl/blob/master/tests/src/test_cases/widgets/test_image.c

AndyEveritt:
Ok, silly question but how do I run the tests to confirm they work?

PGNetHun:
> Ok, silly question but how do I run the tests to confirm they work?

It's not a silly question, I had to check it also :-) 
(did not do code change for a long time)

Basically a Python script runs the tests:
https://github.com/lvgl/lvgl/blob/master/tests/README.md

PGNetHun (dismissed):

AndyEveritt:
@PGNetHun I've added the tests

PGNetHun (dismissed):
Perfect!
Thank you!

AndyEveritt:
@PGNetHun is there anything left for me to do for this to be merged? I don't want to cause any unnecessary delays on my end

PGNetHun:
> @PGNetHun is there anything left for me to do for this to be merged? I don't want to cause any unnecessary delays on my end

Nothing else is left for you, PR just needs another approve.

@kisvegabor @liamHowatt

uLipe (changes requested):
Nice work, I only would ask you to check the commit history, it needs to be clean and describes what a particular commit does.

The last one just means that a commit from a particular sha1 has been merged to this branch, happy to approve when it gets fixed.

AndyEveritt:
@uLipe The last commit was just bringing this branch up to date with the mainline master branch. Not sure why it gave the sha1 hash directly. Not sure how to change it now.

@liamHowatt I do not mind whether it is merged with/without squashing. It is up to you imo. I've updated the PR title so hopefully that is ok now. In response to your question about aligning the image, see the example I put in the first message in this PR.

uLipe:
@AndyEveritt probably something when wrong during rebasing against the master? 

What you can do in yout local copy is to use the interative rebase to drop this commit and restart the rebase procedure again:

```
$ git rebase -i HEAD~7
```

When selecting the commits pick the latest strange commit, on the rebase file and set to drop it, the local copy will be restored in state prior the rebase, then try to rebase again:

```
git fetch origin master #Assuming your origin remote points to the lvgl upstream repo
git rebase origin/master
```

And check the history again:

```
git log 
```

Please let me know if this works.

src/widgets/image/lv_image.c
```diff
@@ -540,6 +540,58 @@ int32_t lv_image_get_scale_y(lv_obj_t * obj)
     return img->scale_y;
 }
 
+int32_t lv_image_get_width(lv_obj_t * obj)
```
kisvegabor: We should avoid using names which conflict with `lv_obj_*` API names.as it would confuse bindings and our XML parser as well. 

 ```suggestion
int32_t lv_image_get_src_width(lv_obj_t * obj)
```

File: src/widgets/image/lv_image.h [hunk 2]
```diff
@@ -51,8 +51,9 @@ typedef enum {
     LV_IMAGE_ALIGN_RIGHT_MID,
     LV_IMAGE_ALIGN_CENTER,
     LV_IMAGE_ALIGN_AUTO_TRANSFORM,
-    LV_IMAGE_ALIGN_STRETCH,
-    LV_IMAGE_ALIGN_TILE,
+    LV_IMAGE_ALIGN_STRETCH,    // Set X and Y scale to fill the Widget's area
+    LV_IMAGE_ALIGN_TILE,       // Tile image to fill Widget's area. Offset is applied to shift the tiling.
+    LV_IMAGE_ALIGN_AUTO_SCALE, // Scale the image to fit without changing its aspect ratio
```
kisvegabor: Please use `/**/` comments

I wonder if we can find a CSS counterpart for it and use that name.

src/widgets/image/lv_image.c
```diff
@@ -540,6 +540,58 @@ int32_t lv_image_get_scale_y(lv_obj_t * obj)
...
+{
+    LV_ASSERT_OBJ(obj, MY_CLASS);
+
+    lv_image_t * img = (lv_image_t *)obj;
+    update_align(obj);
+
+    return img->h;
+}
+
+int32_t lv_image_get_scaled_width(lv_obj_t * obj)
```
kisvegabor:
```suggestion
int32_t lv_image_get_transformed_width(lv_obj_t * obj)
```

src/widgets/image/lv_image.c
AndyEveritt: OK do you have a proposed name you are happy with?

src/widgets/image/lv_image.h
AndyEveritt: Possibly something like this
https://developer.mozilla.org/en-US/docs/Web/CSS/object-fit

It might also be worth adding the equivalent to "cover"

AndyEveritt:
@uLipe is this required if doing a squash merge? Looks like that is how most other PRs have been merged anyway

src/widgets/image/lv_image.h
AndyEveritt: @kisvegabor I have change the name to `LV_IMAGE_ALIGN_CONTAIN` and added `LV_IMAGE_ALIGN_COVER` which match the CSS naming I linked.

Personally I think `LV_IMAGE_ALIGN_FIT` and `LV_IMAGE_ALIGN_FILL` are more descriptive, and since `LV_IMAGE_ALIGN_STRETCH` doesn't follow the CSS naming anyway I would be tempted to use those instead. Please let me know what you think. I don't have a strong opinion either way so am happy to go with your decision.

tests/ref_imgs/widgets/image_cover.png
PGNetHun, line 1: Is this correct?
I thought that cover means to fill full space + clip parts outside of box, like this:

image

AndyEveritt: That is how I would have chosen to do it. But I think to make that change it would also affect anyone who manually scales an image to be larger than the obj size. Since that could be a potential ui breaking change for some people, I decided to leave it to be consistent with existing lvgl behaviour

AndyEveritt: @PGNetHun see almost all the other image test png (eg https://github.com/lvgl/lvgl/blob/master/tests/ref_imgs/widgets/image_scale_pivot_center.png). They all exhibit the same behaviour. I would propose that if displaying the image outside the bounds of the object is unintended behaviour then an issue should be raised and it should be dealt with in its own PR instead of combining it with this one.

PGNetHun: OK

PGNetHun (dismissed):

src/widgets/image/lv_image.c
```diff
@@ -445,7 +445,7 @@ void lv_image_set_inner_align(lv_obj_t * obj, lv_image_align_t align)
     if(align == img->align) return;
 
     /*If we're removing STRETCH, reset the scale*/
-    if(img->align == LV_IMAGE_ALIGN_STRETCH) {
+    if(img->align == LV_IMAGE_ALIGN_STRETCH || img->align == LV_IMAGE_ALIGN_CONTAIN) {
         lv_image_set_scale(obj, LV_SCALE_NONE);
     }
```
liamHowatt: How come it doesn't apply for `LV_IMAGE_ALIGN_COVER` too?

tests/ref_imgs_vg_lite/widgets/image_cover.png
liamHowatt, line 1: I debugged why the top middle image doesn't draw outside its area. It's because the external draw area is calculated only if the scale is != 256. In that case, the scale is 256 so it's bounded by min(obj_w, src_w).

Here is where it happens.

https://github.com/lvgl/lvgl/blob/ac54e6bb577168befd81131407c6507a57e82e6b/src/widgets/image/lv_image.c#L701-L711

But it's not enough to add ` || img->align == LV_IMAGE_ALIGN_COVER` because `lv_image_buf_get_transformed_area` still won't know how much to extend the area.

I'm fine to leave it for now.

image

liamHowatt (dismissed):

tests/ref_imgs_vg_lite/widgets/image_cover.png
AndyEveritt: Nice debugging. In my opinion drawing the image outside the bounds on the object is a bug, but it is not caused by this PR so I would propose it being fixed separately if others agree that it is in fact unintended behaviour.

src/widgets/image/lv_image.c
AndyEveritt: Fixed

tests/ref_imgs_vg_lite/widgets/image_cover.png
liamHowatt: Okay sounds good

liamHowatt (approved):
Thanks very much!

uLipe (approved):

kisvegabor:
Awesome! Thank you!

PR #7975, opened 2025-03-23
fix(draw_sw_vector): fix thorvg canvas colorspace

pharuxtan:
Colors from LVGL are normally alpha-premultiplied, so ThorVG don't need to recompute them.

### DIFF
--- a/src/draw/sw/lv_draw_sw_vector.c
+++ b/src/draw/sw/lv_draw_sw_vector.c
@@ -494,7 +494,7 @@ void lv_draw_sw_vector(lv_draw_task_t * t, const lv_draw_vector_task_dsc_t * dsc
     }
 
     Tvg_Canvas * canvas = tvg_swcanvas_create();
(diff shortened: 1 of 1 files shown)


pharuxtan:
Just saw the PR for this case https://github.com/lvgl/lvgl/pull/7771

By setting the ThorVG canvas target to non premultiplied, my FPS drops and make my fonts render bad.

While waiting the support of both colorspace, could it be reverted to its initial state before the merge of https://github.com/lvgl/lvgl/commit/fc5c1563853eb8fa44c01e85bb117a956a5dfe75?
What do you think?

cc @FASTSHIFT @kisvegabor

uLipe (approved):

kisvegabor (approved):

kisvegabor:
Thank you for reporting it. 

FYI, @fbiego is working ARGB8888 premultipled rendering support in LVGL here: https://github.com/lvgl/lvgl/pull/7979

We were planning to support ARGB8888 premultipled only for destination color format, but it seems we need to add it for source as well. 
 
With this we would be able to use the Lottie rendered content as it is (ARGB8888 premultipled).

PR #7979, opened 2025-03-24
feat(sw_blend) : add argb8888-premultiplied support

fbiego:
- add color format LV_COLOR_FORMAT_ARGB8888_PREMULTIPLIED
- add color blend to argb8888-premultiplied
- add image blend to argb8888-premultiplied

### Notes


lv_conf_template.h
```diff
@@ -169,6 +169,7 @@
     #define LV_DRAW_SW_SUPPORT_RGB888       1
     #define LV_DRAW_SW_SUPPORT_XRGB8888     1
     #define LV_DRAW_SW_SUPPORT_ARGB8888     1
+    #define LV_DRAW_SW_SUPPORT_ARGB8888_PREMULTIPLIED 1
```
uLipe, line 172: I think this name can be abbreviated to keep the consistency of maximum size when compared to the other options.

src/draw/sw/blend/lv_draw_sw_blend_to_argb8888_premultiplied.c
```diff
@@ -0,0 +1,326 @@
...
+    color_argb.green = (color_argb.green * color_argb.alpha) >> 8;
+    color_argb.blue  = (color_argb.blue  * color_argb.alpha) >> 8;
+
+
+    /* Simple fill */
+    if(mask == NULL && opa >= LV_OPA_MAX) {
+        uint32_t color32 = lv_color_to_u32(dsc->color);
+        uint32_t * dest_buf = dsc->dest_buf;
+        for(y = 0; y < h; y++) {
+            for(x = 0; x < w - 16; x += 16) {
```
uLipe: You are doing a loop unrolling to reduce the branching instructions, which is a good technique to improve speed, I just would like to see some validation for dest_buffer it should be aligned to a 16 byte boundary as well its size.

src/draw/sw/blend/lv_draw_sw_blend_to_argb8888_premultiplied.c
```diff
@@ -0,0 +1,326 @@
...
+    lv_color_mix_with_alpha_cache_init(&cache);
+
+    /* Iterate over every pixel in the image */
+    for(int32_t y = 0; y < h; y++) {
+        for(int32_t x = 0; x < w; x++) {
+            lv_color32_t src = src_buf_c32[x];
+
+            /* Convert from premultiplied alpha back to standard RGB */
+            if(src.alpha > 0) {
+                src.red   = (src.red   * 255) / src.alpha;
```
uLipe: does the src.alpha changes in runtime? If not it is possible to perform a pre computation and uses the reciprocal multipliciation to avoid the division instruction here.

uLipe (commented):
My first pass.

src/draw/sw/blend/lv_draw_sw_blend_to_argb8888_premultiplied.c
fbiego: I agree with your suggestion and will update the code to use precomputed reciprocal multiplication instead of division for better performance. Thanks for the optimization tip!

src/draw/sw/blend/lv_draw_sw_blend_to_argb8888_premultiplied.c
fbiego: Thanks for the insight, i'm still learning about these optimizations. What issue does it cause if not aligned?

lv_conf_template.h
fbiego: I would need suggestions for this, currently I'm thinking
`LV_DRAW_SW_SUPPORT_ARGB8888_PM`

uLipe: I think PM as a suffix like you suggested is good enough, the idea is mostly to keep the naming size consistent.

src/draw/sw/blend/lv_draw_sw_blend_to_argb8888_premultiplied.c
uLipe: The most problematic issue is buffer overflow that might happen when the buffers and its size are not aligned with the loop unrolling factor, which is 16 in the case above, when the alignment does not happen the increment by 16 at each interaction may lead the index to point outside of the buffer, corrupting some another variable.

kisvegabor:
Thank you! I'll test it today. 

I think we should add support for 
- in `lv_draw_sw_blend_to_argb8888_premultiplied.c`
   - argb8888 and argb8888_premultiplied source images
- in `lv_draw_sw_blend_to_argb8888.c`, `lv_draw_sw_blend_to_xrgb8888.c`, `lv_draw_sw_blend_to_rgb888.c`, `lv_draw_sw_blend_to_rgb565.c`
   - argb8888_premultiplied source images (similar to argb8888)

Please also add support for premultipled images in `LVGLImage.py`

Please also add a test similar to `test_render_to_rgb565.c` and also add a premultiplied image [here](https://github.com/lvgl/lvgl/blob/master/demos/render/lv_demo_render.c#L333-L359).

kisvegabor:
@fbiego I added the reference images and made the premultiplied test work too. Please check the results.

fbiego:
> I added the reference images and made the premultiplied test work too. Please check the results.

Thanks

lv_conf_template.h
fbiego: @kisvegabor suggested keeping it as it is

kisvegabor:
I think I fixed the transformation edge issues. Please check the reference images.

fbiego:
awesome, that fixes the edge issues.

As for the RGB888 & XRGB8888 images drawn when transformed, this might be the cause
https://github.com/lvgl/lvgl/blob/5c5ae5d2bc9b3626bfed246a8533a1b1a3aeba20/src/draw/sw/lv_draw_sw_img.c#L461-L464

kisvegabor:
> As for the RGB888 & XRGB8888 images drawn when transformed, this might be the cause


Good observation, but that's okay.

uLipe (approved):

liamHowatt (approved):

PR #8017, opened 2025-03-28
fix(draw): fix "blend_non_normal_pixel: Not supported blend mode" issue when using lv_demo_smartwatch which compiled by MSVC

MouriNaruto:
image

In the recent days, one of LVGL Visual Studio repository users told me that there are some issues mentioned in the screenshot when he trying to use lv_demo_smartwatch.

This issue seems something familiar for me. So, I solve the issue for that user quickly. I think I should create a PR instead of an issue, and hope the issue can be fixed as soon as possible.

Kenji Mouri

### DIFF
--- a/src/draw/lv_draw_image.h
+++ b/src/draw/lv_draw_image.h
@@ -71,7 +71,7 @@ struct _lv_draw_image_dsc_t {
     /**Describes how to blend the pixels of the image to the background.
      * See `lv_blend_mode_t` for more details.
      */
-    lv_blend_mode_t blend_mode : 3;
+    uint16_t blend_mode : 4;
 
     /**1: perform the transformation with anti-alaising */
     uint16_t antialias          : 1;
--- a/src/draw/lv_draw_image.h
+++ b/src/draw/lv_draw_image.h
@@ -71,7 +71,7 @@ struct _lv_draw_image_dsc_t {
     /**Describes how to blend the pixels of the image to the background.
      * See `lv_blend_mode_t` for more details.
      */
-    uint16_t blend_mode : 4;
+    lv_blend_mode_t blend_mode : 4;
 
     /**1: perform the transformation with anti-alaising */
     uint16_t antialias          : 1;


uLipe:
@MouriNaruto could you describe why the type of the blend mode is the cause of the issue above?

Thank you :)

MouriNaruto:
> @MouriNaruto could you describe why the type of the blend mode is the cause of the issue above?
> 
> Thank you :)

Because MSVC will consider the enum types as int32_t (not the uint32_t you think) by default.

Kenji Mouri

MouriNaruto:
image
> The constant-expression must have int type and can be negative.

Reference: https://learn.microsoft.com/en-us/cpp/c-language/c-enumeration-declarations?view=msvc-170#syntax

uLipe:
@MouriNaruto thanks for clarifying it.

uLipe (dismissed):

MouriNaruto:
@kisvegabor @liamHowatt ping

Kenji Mouri

src/draw/lv_draw_image.h
```diff
@@ -71,7 +71,7 @@ struct _lv_draw_image_dsc_t {
     /**Describes how to blend the pixels of the image to the background.
      * See `lv_blend_mode_t` for more details.
      */
-    lv_blend_mode_t blend_mode : 3;
+    uint16_t blend_mode : 3;
```
kisvegabor: I'd rather go with this to keep the type for easier debugging.
```suggestion
    uint16_t blend_mode : 4;
```

MouriNaruto: I have followed your suggestions.

Kenji Mouri

MouriNaruto:
@kisvegabor @uLipe @liamHowatt Ping

Kenji Mouri

FASTSHIFT (dismissed):

src/draw/lv_draw_image.h
```diff
@@ -71,7 +71,7 @@ struct _lv_draw_image_dsc_t {
     /**Describes how to blend the pixels of the image to the background.
      * See `lv_blend_mode_t` for more details.
      */
-    lv_blend_mode_t blend_mode : 3;
+    uint16_t blend_mode : 4;
```
kisvegabor: Ah, sorry I meant this:
```suggestion
    lv_blend_mode_t blend_mode : 4;
```
to keep the type. 😓

MouriNaruto: But it will still be the issue again when lv_blend_mode_t adds more enum items in the future. So, maybe we should have a way to let other guys to know that.

MouriNaruto: Note: Only C++11/C23 or later can limit the type of enum constants.
https://en.cppreference.com/w/c/language/enum

MouriNaruto: I followed your suggestion. But I hope there is some way to warn someone who wants to add enum items to that enum type.

liamHowatt (approved):

FASTSHIFT (approved):

PR #8025, opened 2025-03-31
Bugfix and improvement for the logging

jefdriesen:
The first commit fixes a crash in the log function when both LV_LOG_PRINTF is enabled and a log callback is registered. In that case the varargs are used twice, which results in a crash.

The second commit changes the logic to prefer the registered log function over LV_LOG_PRINTF to make it possible to replace the logging destination at runtime.

### Notes

### DIFF
--- a/src/misc/lv_log.c
+++ b/src/misc/lv_log.c
@@ -96,13 +96,6 @@ void lv_log_add(lv_log_level_t level, const char * file, int line, const char * 
 #endif
         static const char * lvl_prefix[] = {"Trace", "Info", "Warn", "Error", "User"};
 
-#if LV_LOG_PRINTF
-        printf("[%s]" LOG_TIMESTAMP_FMT " %s: ",
-               lvl_prefix[level], LOG_TIMESTAMP_EXPR func);
-        vprintf(format, args);
-        printf(LOG_FILE_LINE_FMT "\n" LOG_FILE_LINE_EXPR);
-        fflush(stdout);
-#else
         if(custom_print_cb) {
             char buf[512];
             char msg[256];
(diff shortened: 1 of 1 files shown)


src/misc/lv_log.c
```diff
@@ -111,6 +104,15 @@ void lv_log_add(lv_log_level_t level, const char * file, int line, const char *
                         lvl_prefix[level], LOG_TIMESTAMP_EXPR func, msg LOG_FILE_LINE_EXPR);
             custom_print_cb(level, buf);
         }
+        else {
+#if LV_LOG_PRINTF
```
uLipe, line 107: I think you can put the whole else block inside of the ifdef instead of making it noop.

Is there any special reason to make it noop?

src/misc/lv_log.c
```diff
@@ -126,15 +128,16 @@ void lv_log(const char * format, ...)
...
-#if LV_LOG_PRINTF
-    vprintf(format, args);
-#else
     if(custom_print_cb) {
         char buf[512];
         lv_vsnprintf(buf, sizeof(buf), format, args);
         custom_print_cb(LV_LOG_LEVEL_USER, buf);
     }
+    else {
+#if LV_LOG_PRINTF
```
uLipe, line 136: Also here

uLipe (commented):
I just have a comment regarding the coverage of the ifdef

src/misc/lv_log.c
jefdriesen: No particular reason. I have adjusted the code.

uLipe (approved):

jefdriesen:
Fixed a small typo in the commit message.

uLipe (approved):

kisvegabor (approved):
Thank you!

PR #8041, opened 2025-04-02
fix(wayland): Remove the XDG_RUNTIME_DIR check

jefdriesen:
The XDG_RUNTIME_DIR environment variable isn't used anywhere in the driver code. If the WAYLAND_DISPLAY variable is set to an absolute path, the XDG_RUNTIME_DIR variable is also not required.

### Notes

### DIFF
--- a/src/drivers/wayland/lv_wayland.c
+++ b/src/drivers/wayland/lv_wayland.c
@@ -169,8 +169,6 @@ struct application {
     struct xdg_wm_base * xdg_wm;
 #endif
 
-    const char * xdg_runtime_dir;
-
 #ifdef LV_WAYLAND_WINDOW_DECORATIONS
     bool opt_disable_decorations;
 #endif
@@ -2419,9 +2417,6 @@ static void wayland_init(void)
         sme_init_buffer,
         sme_free_buffer
     };
-
-    application.xdg_runtime_dir = getenv("XDG_RUNTIME_DIR");
-    LV_ASSERT_MSG(application.xdg_runtime_dir, "cannot get XDG_RUNTIME_DIR");
 
     // Create XKB context
     application.xkb_context = xkb_context_new(XKB_CONTEXT_NO_FLAGS);


uLipe:
Hello, thanks for the PR!

Is there any negative impacts on keep the XDG_RUNTIME_DIR variable? Or it is just a cleanup?

jefdriesen:
The `XDG_RUNTIME_DIR` env variable isn't required when using the `WAYLAND_DISPLAY` env variable directly. But when  `XDG_RUNTIME_DIR` is not set, the assert immediately kills the application and thus prevents it from running. That's the problem this PR fixes. It's not just a cleanup.

AndreCostaaa (approved):
This is okay, here's the reasoning:

Basically we either need the XDG_RUNTIME_DIR or WAYLAND_DISPLAY env variables so this asserts too early as it could still work if `WAYLAND_DISPLAY` is the absolute path to the display

Docs: https://manpages.debian.org/experimental/libwayland-doc/wl_display_connect.3.en.html

So this would work with this PR but not without it

```bash
$ export WAYLAND_DISPLAY=/run/user/1000/wayland-0
$ unset XDG_RUNTIME_DIR
$ ./lvglsim-wayland 
[Warn]	(0.000, +0)	lv_init: Memory integrity checks are enabled via LV_USE_ASSERT_MEM_INTEGRITY which makes LVGL much slower lv_init.c:309
[Warn]	(0.000, +0)	lv_init: Object sanity checks are enabled via LV_USE_ASSERT_OBJ which makes LVGL much slower lv_init.c:313
[Warn]	(0.000, +0)	lv_init: Style sanity checks are enabled that uses more RAM lv_init.c:317
[User]	(645.375, +645375)	init_pointer_evdev: Using evdev automatic discovery. evdev.c:161
[User]	(645.375, +0)	discovery_cb: new 'ABS' device discovered evdev.c:110
[User]	(645.404, +29)	discovery_cb: new 'REL' device discovered evdev.c:110
```

Removing the assert and not having either of the env variables will result in this being output which is still fine to me

```bash
$ unset WAYLAND_DISPLAY
$ unset XDG_RUNTIME_DIR
$ ./lvglsim-wayland 
[Warn]	(0.000, +0)	lv_init: Memory integrity checks are enabled via LV_USE_ASSERT_MEM_INTEGRITY which makes LVGL much slower lv_init.c:309
[Warn]	(0.000, +0)	lv_init: Object sanity checks are enabled via LV_USE_ASSERT_OBJ which makes LVGL much slower lv_init.c:313
[Warn]	(0.000, +0)	lv_init: Style sanity checks are enabled that uses more RAM lv_init.c:317
error: XDG_RUNTIME_DIR is invalid or not set in the environment.
[Error]	(0.000, +0)	wayland_init: Asserted at expression: application.display (failed to connect to Wayland server) lv_wayland.c:2416
```

uLipe (approved):

PR #8047, opened 2025-04-03
fix: table get cell area error

AndreCostaaa:
Fixes #7381. I don't have the permissions to push into that branch and seems the author is not been available. Hope it's okay

See problem in #7381, this PR completes the original one by simply checking for the specific config that was causing the failing CI. I believe it's okay to do it this way because if one day the `invalidate_area` function is changed, the test needs to be updated anyway

@liamHowatt @kisvegabor

### DIFF
--- a/src/widgets/table/lv_table.c
+++ b/src/widgets/table/lv_table.c
@@ -1091,18 +1091,35 @@ static void get_cell_area(lv_obj_t * obj, uint32_t row, uint32_t col, lv_area_t 
     for(c = 0; c < col; c++) {
         area->x1 += table->col_w[c];
     }
+    /* Traverse the current row from the first until the penultimate column.
+     * Increment the text width if the cell has the LV_TABLE_CELL_CTRL_MERGE_RIGHT control,
+     * exit the traversal when the current cell control is not LV_TABLE_CELL_CTRL_MERGE_RIGHT */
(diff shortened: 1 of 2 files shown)


src/widgets/table/lv_table.c
```diff
@@ -1091,18 +1091,35 @@ static void get_cell_area(lv_obj_t * obj, uint32_t row, uint32_t col, lv_area_t
     for(c = 0; c < col; c++) {
         area->x1 += table->col_w[c];
     }
-
+    /* Traverse the current row from the first until the penultimate column.
+     * Increment the text width if the cell has the LV_TABLE_CELL_CTRL_MERGE_RIGHT control,
```
kisvegabor:
```suggestion
     * Increment the offset if the cell has the LV_TABLE_CELL_CTRL_MERGE_RIGHT control,
```

kisvegabor (commented):
Looks good, thank you! Really creative test case! 

Just one minor comment.

liamHowatt (approved):
Arrgh it was failing in the VG Lite test, I am blind. Thank you for fixing it! It's brilliant!

AndreCostaaa:
No worries! And thanks 😀

uLipe (approved):

PR #8052, opened 2025-04-04
build(custom): fix include path of lv_conf.cmake

### LINKED ISSUE #8046: cmake throws error on Ubuntu build
### LVGL version

v9.3.0

### Platform

Ubuntu 22.04

### What happened?

Initially, the cmake failed to find "pcpp" (which was not required previously).  I installed it using "apt install pcc" and now cmake fails with

```
-- The C compiler identification is GNU 11.4.0
-- The CXX compiler identification is GNU 11.4.0
-- Found assembler: /usr/bin/cc
-- Found Python: /usr/bin/python3.10 (found version "3.10.12") found components: Interpreter 
error: unknown option '--passthru-defines'
Error during preprocessing: Command '['pcpp', '-o', '/home/paul/lvgl/build/tmp.h', '--passthru-defines', '--line-directive=', '/home/paul/lvgl/src/lv_conf_internal.h', '-I/home/paul/lvgl', '-I/home/paul/lvgl/..', '-I/home/paul/lvgl/src']' returned non-zero exit status 8.
CMake Error at env_support/cmake/custom.cmake:72 (message):
  preprocess_lv_conf_internal.py failed with return code: 1
  CMakeLists.txt:23 (include)

```

Reverting this commit
```
commit 371c21bbdaa8216f2a0bef0069db7ae1a727d80d
Author: David Truan <david.truan@edgemtech.ch>
Date:   Thu Apr 3 14:42:18 2025 +0200

    arch(cmake): add native Kconfig support to cmake (#7934)
```

lets the current head build without the error

### How to reproduce?

```
mkdir build ; cd build ; cmake ..
```

AndreCostaaa:
Fixes #8046 

The reason the CI wasn't failing is because the `CMAKE_CURRENT_BINARY_DIR` and `CMAKE_BINARY_DIR` point to the same path when building lvgl alone : 

```bash
#inside lvgl
mkdir build; cmake ..
```

But in other repositories like the `lv_port_pc_vscode`, these paths differ as `CMAKE_CURRENT_BINARY_DIR` will be `lv_port_pc_vscode/build/lvgl` whereas `CMAKE_BINARY_DIR` will be `lv_port_pc_vscode/build` when doing something like

```bash
#inside lv_port_pc_vscode
mkdir build; cmake ..
```

cc @Daxidz

### DIFF
--- a/env_support/cmake/custom.cmake
+++ b/env_support/cmake/custom.cmake
@@ -34,6 +34,8 @@ file(GLOB_RECURSE THORVG_SOURCES ${LVGL_ROOT_DIR}/src/libs/thorvg/*.cpp ${LVGL_R
 add_library(lvgl ${SOURCES})
 add_library(lvgl::lvgl ALIAS lvgl)
 
+set(PCPP_DEFINITIONS_LIST "")
+
 if ( LV_USE_KCONFIG )
   # kconfig.cmake will generate the .config
   # and autoconf.h, which will be used by lv_conf_kconfig.h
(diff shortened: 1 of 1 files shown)


Daxidz:
Nice catch, I guess I was to focused on my testing setup. Will be more careful in the future

AndreCostaaa:
No worries. Happens to the best of us :)

kisvegabor (dismissed):
Thank you!

AndreCostaaa:
There were actually two separate issues involved:

Issue 1: As described in the PR, the build would fail when the LVGL CMakeLists.txt wasn’t the first one invoked.
- Fixed by [b6b36a6](https://github.com/lvgl/lvgl/pull/8052/commits/b6b36a6b136b3c40ea81da2160f8c2a1317b2d71)

Issue 2: As reported by @pfvogel in [issue #8046](https://github.com/lvgl/lvgl/issues/8046), the build would break when building LVGL directly inside the lvgl folder with a lv_conf.h also located there. See [this comment](https://github.com/lvgl/lvgl/issues/8046#issuecomment-2778843737) for details.
- Fixed by [5bc874c](https://github.com/lvgl/lvgl/pull/8052/commits/5bc874ca2d4a801f1d0c2688a27166c5036b1a5d)

The third commit also resolves an issue where multiple definitions couldn’t be passed to `lv_conf_internal_gen.py`, which caused the CI to fail due to incorrect handling of LV_CONF_PATH for the tests

kisvegabor (approved):
Thank you!

It's also related: https://github.com/lvgl/lvgl/pull/8069

PGNetHun (approved):

Pavi92:
there is also a problem with the generate_cmake_variables.py.
if the line in lv_conf_expanded.h is something like ```#define LV_USE_NAMEX (LV_USE_NAME1 | LV_USE_NAME2 | LV_USE_NAME3)```
the corresponding line in lv_conf.cmake is not generated correctly. It will look like ```set(CONFIG_LV_USE_NAMEX (LV_USE_NAME1)``` and cmake will complain about the missing closing bracket.

```diff
diff --git a/scripts/generate_cmake_variables.py b/scripts/generate_cmake_variables.py
index 33a230c86..cf83f1b4b 100755
--- a/scripts/generate_cmake_variables.py
+++ b/scripts/generate_cmake_variables.py
@@ -92,14 +92,14 @@ def generate_cmake_variables(path_input: str, path_output: str, kconfig: bool):
         # For the rest of config, simply add CONFIG_ and write
         # all LV_USE_* configs, as these are the one needed in cmake
         elif line.startswith(f'{CONFIG_PATTERN}'):
-
+            
             parts = line.split()
             if len(parts) < 3:
                 continue
 
+            value = ' '.join(parts[2:])
+            value = value.strip()
             name = parts[1]
-            value = parts[2].strip()
-
             fout.write(f'set({CONFIG_PREFIX}{name} {value})\n')
 
 

```

AndreCostaaa:
I think that your fix won't work because CMake's OR operator is the keyword `OR` and not `|` afaik. We could detect that the line being parsed contains `|` and replace it with `OR` but that will start to get really complicated.

The C language is too rich and I don't think trying to support every possible way of defining a macro needs to be supported in the `generate_cmake_variables.py` script when we can always simplify the `lv_conf.h` defines into `#define LV_USE_XX 1` in one way or another. What do you think?

Pavi92:
I agree it would get really complicated.
Maybe it would be better to to create a file lv_conf.cmake with the default configuration as cache variables and use cmakes configure_file command to create the lv_conf.h from these variables ?

AndreCostaaa:
The intended approach is that `lv_conf.h` and `.config` serve as the primary entry points for configuring LVGL—whether manually or through tools like `menuconfig`. Introducing a new `lv_conf.cmake` file as an entrypoint doesn't feel like simplifying the process; in fact, i would argue it complicates it. I could be wrong though

kisvegabor:
I merged it to fix the CMake issue

liamHowatt:
Thank you :pray:

FYI I opened a PR for something else. I'm open to another solution. I added reproduce info. #8079

PR #8058, opened 2025-04-06
feat(linux/fbdev): support non-mmappable frame buffers

jcmvbkbc:
Linux framebuffer may not support mmap(). Introduce configuration symbol LV_LINUX_FBDEV_MMAP and use pwrite() to copy data to the framebuffer when it's set to 0.

### DIFF
--- a/src/drivers/display/fb/lv_linux_fbdev.c
+++ b/src/drivers/display/fb/lv_linux_fbdev.c
@@ -59,7 +59,9 @@ typedef struct {
     struct fb_var_screeninfo vinfo;
     struct fb_fix_screeninfo finfo;
 #endif /* LV_LINUX_FBDEV_BSD */
+#if LV_LINUX_FBDEV_MMAP
     char * fbp;
+#endif
     uint8_t * rotated_buf;
     size_t rotated_buf_size;
     long int screensize;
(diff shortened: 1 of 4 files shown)


src/drivers/display/fb/lv_linux_fbdev.c
```diff
@@ -249,14 +253,26 @@ void lv_linux_fbdev_set_force_refresh(lv_display_t * disp, bool enabled)
  *   STATIC FUNCTIONS
  **********************/
 
+static void write_to_fb(lv_linux_fb_t * dsc, uint32_t fb_pos, const void * data, size_t sz)
+{
+#if LV_LINUX_FBDEV_MMAP
+    uint8_t * fbp = (uint8_t *)dsc->fbp;
+    lv_memcpy(&fbp[fb_pos], data, sz);
+#else
+    pwrite(dsc->fbfd, data, sz, fb_pos);
```
FASTSHIFT: Add return value checking and error `errno` printing.

jcmvbkbc: Done.

src/drivers/display/fb/lv_linux_fbdev.c
```diff
@@ -249,14 +253,27 @@ void lv_linux_fbdev_set_force_refresh(lv_display_t * disp, bool enabled)
...
  **********************/
 
+static void write_to_fb(lv_linux_fb_t * dsc, uint32_t fb_pos, const void * data, size_t sz)
+{
+#if LV_LINUX_FBDEV_MMAP
+    uint8_t * fbp = (uint8_t *)dsc->fbp;
+    lv_memcpy(&fbp[fb_pos], data, sz);
+#else
+    if(pwrite(dsc->fbfd, data, sz, fb_pos) < 0)
+        perror("Error writing to fbfd");
```
FASTSHIFT: LV_LOG_ERROR("write failed: %d", errno);

jcmvbkbc: Other error reporting in this file used `perror()`, that's why I did it that way.
In places where `LV_LOG_ERROR()` is used it tends to print `strerror(errno)` rather than `errno`.
I've changed it as you suggested.

FASTSHIFT (approved):

src/drivers/display/fb/lv_linux_fbdev.c
FASTSHIFT: It is more appropriate to use `LV_LOG`, which is more convenient for managing log output.

FASTSHIFT: Yes, you can consider printing `strerror(errno)`.

kisvegabor (approved):
Thank you! What is the device/platform where mmap didn't work?

jcmvbkbc:
> What is the device/platform where mmap didn't work?

It's an SPI-attached LCD display on noMMU linux, in my case it's a ilitek,ili9341 on Espressif ESP32-S3-BOX3.

PR #8059, opened 2025-04-06
fix(font): allow non-constant LV_FONT_DEFAULT again (fixes #7788)

### LINKED ISSUE #7788: `LV_FONT_DEFAULT` must now be a compile-time constant
### Introduce the problem

Due to

```c
const lv_font_t * const lv_font_default = LV_FONT_DEFAULT;
```

introduced in ad8a9d95873c05474905eb498b9d6b5118b9e866 , `LV_FONT_DEFAULT` must now be a compile-time constant. Before I was using something like

```c
#define LV_FONT_DEFAULT my_default_font()
```

with:

```c
const lv_font_t* my_default_font()
{
    static const lv_font_t* font = nullptr;
    if (!font) {
        font = lv_freetype_font_create(/* ... */);
    }
    return font;
}
```

Is there another way to use default fonts that are not compile time constants?

### Proposal

_No response_

niklasf:
Fixes #7788.

Allows `#define LV_FONT_DEFAULT MyDefaultFont()` again after ad8a9d95873c05474905eb498b9d6b5118b9e866.

### DIFF
--- a/src/font/lv_font.c
+++ b/src/font/lv_font.c
@@ -34,8 +34,6 @@
  *  GLOBAL VARIABLES
  **********************/
 
-const lv_font_t * const lv_font_default = LV_FONT_DEFAULT;
-
 /**********************
  * GLOBAL PROTOTYPES
  **********************/
@@ -153,7 +151,7 @@ int32_t lv_font_get_line_height(const lv_font_t * font)
 
 const lv_font_t * lv_font_get_default(void)
 {
-    return lv_font_default;
+    return LV_FONT_DEFAULT;
 }
 
 /**********************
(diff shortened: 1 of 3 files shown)


liamHowatt (commented):
Thank you. Please also update `fonts.rst` and `component_library.rst`.

kisvegabor:
It's really more flexible like this. Thank you! 

Please update the docs as Liam suggested and we should be good to go.

niklasf:
You mean the reference to `lv_font_default`?

There`s still

```c
lv_xml_register_font(NULL, "lv_font_default", lv_font_get_default());
```

so I believe from XML perspective the constant still exists.

liamHowatt:
Oh hmm I see what you mean. It exists in XML but not the normal LVGL API. @kisvegabor should the `lv_xml_register_font` be changed?

kisvegabor (approved):
Ah, true. No update is needed indeed.

liamHowatt (approved):

PR #8068, opened 2025-04-07
fix(makefile): Path fixes in component.mk

rbudai98:
This commit fixes paths for makefile-systems that uses component.mk

The paths of components do not reflect the actual position of specific components. This commit fixes these path to be able to use the component.mk to include them. It was tested with building the library and initializing it with: `lv_init()`

### DIFF
--- a/component.mk
+++ b/component.mk
@@ -3,64 +3,89 @@
 COMPONENT_SRCDIRS := . \
                   src \
                   src/core \
-                  src/draw \
                   src/extra \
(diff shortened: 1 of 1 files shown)


PGNetHun (approved):
I've checked only a few directories, they are indeed in a new path.
Could you please provide screenshot evidence that build really finishes successfully?
Thank you!

FASTSHIFT (approved):

PR #8070, opened 2025-04-08
perf(draw): support draw task dsc alloc together to reduce the malloc call times

terry0012:
Improve the performance of draw task create.

Previously, draw task and dsc were allocated separately, 
Now they can be allocated together to reduce the number of `lv_malloc `call times.
And `lv_draw_add_task` still support allocated separately if the type is `LV_DRAW_TASK_TYPE_NONE`

### Notes

### DIFF
--- a/src/draw/lv_draw.c
+++ b/src/draw/lv_draw.c
@@ -13,6 +13,8 @@
 #include "../misc/lv_area_private.h"
 #include "../misc/lv_assert.h"
 #include "lv_draw_private.h"
+#include "lv_draw_mask_private.h"
+#include "lv_draw_vector_private.h"
 #include "sw/lv_draw_sw.h"
 #include "../display/lv_display_private.h"
 #include "../core/lv_global.h"
@@ -33,6 +35,7 @@
  **********************/
 static bool is_independent(lv_layer_t * layer, lv_draw_task_t * t_check);
 static void cleanup_task(lv_draw_task_t * t, lv_display_t * disp);
+static inline size_t get_draw_dsc_size(lv_draw_task_type_t type);
 static lv_draw_task_t * get_first_available_task(lv_layer_t * layer);
 
 #if LV_LOG_LEVEL <= LV_LOG_LEVEL_INFO
@@ -92,10 +95,12 @@ void * lv_draw_create_unit(size_t size)
     return new_unit;
 }
 
-lv_draw_task_t * lv_draw_add_task(lv_layer_t * layer, const lv_area_t * coords)
+lv_draw_task_t * lv_draw_add_task(lv_layer_t * layer, const lv_area_t * coords, lv_draw_task_type_t type)
 {
     LV_PROFILER_DRAW_BEGIN;
-    lv_draw_task_t * new_task = lv_malloc_zeroed(sizeof(lv_draw_task_t));
+    size_t dsc_size = get_draw_dsc_size(type);
+    LV_ASSERT_FORMAT_MSG(dsc_size > 0, "Draw task size is 0 for type %d", type);
+    lv_draw_task_t * new_task = lv_malloc_zeroed(LV_ALIGN_UP(sizeof(lv_draw_task_t), 8) + dsc_size);
     LV_ASSERT_MALLOC(new_task);
     new_task->area = *coords;
     new_task->_real_area = *coords;
@@ -104,6 +109,8 @@ lv_draw_task_t * lv_draw_add_task(lv_layer_t * layer, const lv_area_t * coords)
 #if LV_DRAW_TRANSFORM_USE_MATRIX
     new_task->matrix = layer->matrix;
 #endif
+    new_task->type = type;
+    new_task->draw_dsc = (uint8_t *)new_task + LV_ALIGN_UP(sizeof(lv_draw_task_t), 8);
     new_task->state = LV_DRAW_TASK_STATE_QUEUED;
 
     /*Find the tail*/
@@ -324,7 +331,6 @@ lv_draw_task_t * lv_draw_get_next_available_task(lv_layer_t * layer, lv_draw_tas
 {
     LV_PROFILER_DRAW_BEGIN;
 
-
     /*If the first task is screen sized, there cannot be independent areas*/
     if(layer->draw_task_head) {
         int32_t hor_res = lv_display_get_horizontal_resolution(lv_refr_get_disp_refreshing());
@@ -441,7 +447,6 @@ void lv_draw_layer_init(lv_layer_t * layer, lv_layer_t * parent_layer, lv_color_
     LV_PROFILER_DRAW_END;
 }
 
-
 void * lv_draw_layer_alloc_buf(lv_layer_t * layer)
 {
     LV_PROFILER_DRAW_BEGIN;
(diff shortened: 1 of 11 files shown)


src/draw/lv_draw.c
```diff
@@ -589,7 +638,6 @@ static void cleanup_task(lv_draw_task_t * t, lv_display_t * disp)
         draw_label_dsc->text = NULL;
     }
 
-    lv_free(t->draw_dsc);
```
terry0012, line 592: here should consider separately alloc.

src/draw/lv_draw.c
```diff
@@ -589,7 +638,10 @@ static void cleanup_task(lv_draw_task_t * t, lv_display_t * disp)
...
+    /* draw_dsc is alloc sepreated with draw_task */
+    if(t->draw_dsc != (char *)t + sizeof(lv_draw_task_t)) {
```
FASTSHIFT: (uint8_t *)t

terry0012:
1. updated lv_draw_image to adapt the new draw task alloc method. 
@FASTSHIFT

FASTSHIFT (dismissed):

liamHowatt (commented):
If `lv_draw_task_t` has an alignment less than the draw descriptor, the draw descriptor could be misaligned. I.e., on a 32-bit platform, the alignment of `lv_draw_task_t` is 32 bits, and if `sizeof(lv_draw_task_t) % 8 != 0` and a future draw descriptor has a `double` or `int64_t` type member (8 byte alignment), then unaligned memory access could cause issues if the arch doesn't support unaligned access if 64-bit types are treated differently than pairs of 32-bit values on the arch.

C11 has `alignof` and `max_align_t` but those are C11 and not totally portable as they're from a new standard.

I'd recommend

```c
lv_draw_task_t * new_task = lv_malloc_zeroed(LV_ALIGN_UP(sizeof(lv_draw_task_t), 8) + dsc_size);
```

and

```c
new_task->draw_dsc = (uint8_t *)new_task + LV_ALIGN_UP(sizeof(lv_draw_task_t), 8);
```

and

```c
    if(t->draw_dsc != (uint8_t *)t + LV_ALIGN_UP(sizeof(lv_draw_task_t), 8)) {
        lv_free(t->draw_dsc);
    }
```

Assuming 8 byte (64 bit) alignment is conservative enough. A value of `16` could be better.

One other thought. That `free` condition may be risky if the draw dsc size is zero and the allocator placed the allocation of the dsc after the draw task on its own, in a separate allocation. In that case it should be freed, but wouldn't be, because the pointer will alias and be equal. I recommend adding a member to `lv_draw_task_t`

```c
    bool dsc_is_contiguous;
```

As an absolute way to know whether to free. It would occupy tail-padding and not make the struct larger regardless.

liamHowatt:
Here is a concrete example. I have an open PR that adds a new draw dsc with a type `lv_3dtexture_id_t` that is framework-specific. It could be a 64 bit type.

https://github.com/lvgl/lvgl/pull/8033/files#diff-3a4ca0357860507f59273a5ec876b3a0954845443d29f12e6246ef7df33e447dR28-R32

terry0012:
> If `lv_draw_task_t` has an alignment less than the draw descriptor, the draw descriptor could be misaligned. I.e., on a 32-bit platform, the alignment of `lv_draw_task_t` is 32 bits, and if `sizeof(lv_draw_task_t) % 8 != 0` and a future draw descriptor has a `double` or `int64_t` type member (8 byte alignment), then unaligned memory access could cause issues if the arch doesn't support unaligned access if 64-bit types are treated differently than pairs of 32-bit values on the arch.
> 
> C11 has `alignof` and `max_align_t` but those are C11 and not totally portable as they're from a new standard.
> 
> I'd recommend
> 
> ```c
> lv_draw_task_t * new_task = lv_malloc_zeroed(LV_ALIGN_UP(sizeof(lv_draw_task_t), 8) + dsc_size);
> ```
> 
> and
> 
> ```c
> new_task->draw_dsc = (uint8_t *)new_task + LV_ALIGN_UP(sizeof(lv_draw_task_t), 8);
> ```
> 
> and
> 
> ```c
>     if(t->draw_dsc != (uint8_t *)t + LV_ALIGN_UP(sizeof(lv_draw_task_t), 8)) {
>         lv_free(t->draw_dsc);
>     }
> ```
> 
> Assuming 8 byte (64 bit) alignment is conservative enough. A value of `16` could be better.
> 

Agree.

> One other thought. That `free` condition may be risky if the draw dsc size is zero and the allocator placed the allocation of the dsc after the draw task on its own, in a separate allocation. In that case it should be freed, but wouldn't be, because the pointer will alias and be equal. I recommend adding a member to `lv_draw_task_t`
> 
> ```c
>     bool dsc_is_contiguous;
> ```
> 
> As an absolute way to know whether to free. It would occupy tail-padding and not make the struct larger regardless.
How about only support contiguous draw_buf and draw_dsc allocation ?
Now all the draw task has been changed to contiguous allocation, is there any case need to allocate separately in future?

liamHowatt:
> How about only support contiguous draw_buf and draw_dsc allocation ?

Fine with me.

> Now all the draw task has been changed to contiguous allocation, is there any case need to allocate separately in future?

Hmm good question. I can't think of a case. Even vector has a special `lv_draw_vector_task_dsc_t` which is only used with `lv_draw_task_t`.

terry0012:
> > How about only support contiguous draw_buf and draw_dsc allocation ?
> 
> Fine with me.
> 
> > Now all the draw task has been changed to contiguous allocation, is there any case need to allocate separately in future?
> 
> Hmm good question. I can't think of a case. Even vector has a special `lv_draw_vector_task_dsc_t` which is only used with `lv_draw_task_t`.

OK. I think we just only support contiguous allocation now.
If there is a case need to allocate separately, we can add it back.
How do you think? @FASTSHIFT @kisvegabor

FASTSHIFT (dismissed):

src/draw/lv_draw.c
```diff
@@ -92,10 +95,12 @@ void * lv_draw_create_unit(size_t size)
     return new_unit;
 }
 
-lv_draw_task_t * lv_draw_add_task(lv_layer_t * layer, const lv_area_t * coords)
+lv_draw_task_t * lv_draw_add_task(lv_layer_t * layer, const lv_area_t * coords, lv_draw_task_type_t type)
 {
     LV_PROFILER_DRAW_BEGIN;
-    lv_draw_task_t * new_task = lv_malloc_zeroed(sizeof(lv_draw_task_t));
+    size_t dsc_size = get_draw_dsc_size(type);
+    LV_ASSERT(dsc_size > 0);
```
FASTSHIFT: Use `LV_ASSERT_MSG_FMT` to print the `type` value.

src/draw/lv_draw.c
```diff
@@ -535,6 +540,49 @@ static bool is_independent(lv_layer_t * layer, lv_draw_task_t * t_check)
...
+
+        /* no struct match for LV_DRAW_TASK_TYPE_MASK_BITMAP, set it to zero now */
+        case LV_DRAW_TASK_TYPE_MASK_BITMAP:
+            return 0;
+#if LV_USE_VECTOR_GRAPHIC
+        case LV_DRAW_TASK_TYPE_VECTOR:
+            return sizeof(lv_draw_vector_task_dsc_t);
+#endif
+        default:
+            return 0;
```
FASTSHIFT: Remove the `default` branch. When adding a new type, the compiler will automatically warn the developer of the case to be handled.

FASTSHIFT (approved):

liamHowatt (approved):
Thanks! I think it's a great enhancement.

PR #8080, opened 2025-04-10
fix(conf): include LV_CONF_PATH only if LV_CONF_INCLUDE_SIMPLE is not defined

Pavi92:


liamHowatt:
@Pavi92 see this PR I just opened to see if it's related #8079

PR #8100, opened 2025-04-15
ci(docs): fix commit ref when workflow is triggered by master push

AndreCostaaa:
The docs building CI is broken again because it's grabbing an old version of LVGL to build with the current docs
image

The commit reference being used by the CI is the one currently being pointed by the lvgl submodule in [lv_web_emscripten](https://github.com/lvgl/lv_web_emscripten)

It looks like this issue has been around for a while but went unnoticed since it didn’t cause the CI to fail until now. I guess a bunch of minor issues with the docs CI all decided to surface at the same time.

This PR sets the repo url and commit reference explicitly for the build docs script

Sources for the variables used [here](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/accessing-contextual-information-about-workflow-runs#github-context)

@liamHowatt @kisvegabor


kisvegabor:
Thank you! Let's see! 

I bypass the reviews.

PR #8110, opened 2025-04-15
feat(obj_property): make lv_property_t reusable between widgets.

### LINKED ISSUE #7982: lv_obj_set_property and lv_obj_set_properties stomp over the id of the passed in property/ies
### LVGL version

v9.2.0-493-g56320599c

### Platform

Linux x64 through lv_port_pc_vscode example project.

### What happened?

Reading the docs when I came across the widget property API and wanted to use it to make it easier to create more data driven widget setup templates. When trying it out I found that both of property set methods stomp over the id of the passed in properties under certain circumstances, like when setting flags. Making you only able to use them for 1 widget.

Traced it to the following line https://github.com/lvgl/lvgl/blob/b88fb674d3e54e27d05ef7595a77052c20453698/src/core/lv_obj_property.c#L272

I've fixed the issue locally be removing the line and so far have had no issues. Is this some old remnant from when the system worked slightly differently or am I missing some use case were this line is very neccessary?

### How to reproduce?

Copy the following snippet into a test project, build and run.
```c
const lv_property_t toggleButtonProps[] = {
	{.id = LV_PROPERTY_OBJ_FLAG_CHECKABLE, .enable = true},
};

lv_obj_t* layout = lv_obj_create(lv_screen_active());
lv_obj_set_size(layout, 300, 75);
lv_obj_align(layout, LV_ALIGN_TOP_MID, 0, 2);
lv_obj_set_flex_flow(layout, LV_FLEX_FLOW_ROW);

lv_obj_t* button1 = lv_button_create(layout);
lv_obj_set_properties(button1, toggleButtonProps, 1);
lv_obj_t* button1Label = lv_label_create(button1);
lv_label_set_text(button1Label, "Btn1");
		
lv_obj_t* button2 = lv_button_create(layout);
lv_obj_set_properties(button2, toggleButtonProps, 1);
lv_obj_t* button2Label = lv_label_create(button2);
lv_label_set_text(button2Label, "Btn2");

```
Observe that the toggle property is only applied to Btn1

andrimarjonsson:
Fixes #7982

Removed code in `obj_property` which stomped over the passed in property id when it needed to be forwarded to the widget for handling. This basically makes your conveniently setup widget property sets single use.

After the change, code like the following works as expected:

```c
const lv_property_t toggleButtonProps[] = {
	{.id = LV_PROPERTY_OBJ_FLAG_CHECKABLE, .enable = true},
};

lv_obj_t* layout = lv_obj_create(lv_screen_active());
lv_obj_set_size(layout, 300, 75);
lv_obj_align(layout, LV_ALIGN_TOP_MID, 0, 2);
lv_obj_set_flex_flow(layout, LV_FLEX_FLOW_ROW);

lv_obj_t* button1 = lv_button_create(layout);
lv_obj_set_properties(button1, toggleButtonProps, 1);
lv_obj_t* button1Label = lv_label_create(button1);
lv_label_set_text(button1Label, "Btn1");
		
lv_obj_t* button2 = lv_button_create(layout);
lv_obj_set_properties(button2, toggleButtonProps, 1);
lv_obj_t* button2Label = lv_label_create(button2);
lv_label_set_text(button2Label, "Btn2");
```

As this is my first pull request, please let me know if I missed something and I'll get right on it.


uLipe (dismissed):

liamHowatt (commented):
Good catch. Thanks for opening a PR for it. Please add/update a test case in `tests/src/test_cases/widgets/test_obj_property.c`. The example you gave in your issue would be great.

Also do you think it'd be possible to make it so the `const` `* value` parameter

https://github.com/lvgl/lvgl/blob/8f94109d9826c7c9c30e382056f49c73d7803e33/src/core/lv_obj_property.c#L109

isn't having `const` cast away here?

https://github.com/lvgl/lvgl/blob/8f94109d9826c7c9c30e382056f49c73d7803e33/src/core/lv_obj_property.c#L124

Perhaps all you could do to fix it would be

```c
lv_property_t value_copy = *value;
return obj_property(obj, value->id, &value_copy, true);
```

andrimarjonsson:
> Good catch. Thanks for opening a PR for it. Please add/update a test case in `tests/src/test_cases/widgets/test_obj_property.c`. The example you gave in your issue would be great.

Thanks and will get on it, as soon as I get the latest building. Was testing this via the `lv_port_pc_vscode` but when switching to vanilla lvgl and following the steps laid out in https://docs.lvgl.io/master/details/integration/building/cmake.html#integrating-lvgl-cmake I get the following error:
```
CMake Error at build/lv_conf.cmake:141:
  Parse error.  Function missing ending ")".  End of file reached.
  env_support/cmake/custom.cmake:107 (include)
  CMakeLists.txt:24 (include)


CMake Error at env_support/cmake/custom.cmake:107 (include):
  include could not load requested file:

    /home/andrimar/Projects/lvgl/build/lv_conf.cmake
  CMakeLists.txt:24 (include)


-- Configuring incomplete, errors occurred!
```
 
> Also do you think it'd be possible to make it so the `const` `* value` parameter
> 
> https://github.com/lvgl/lvgl/blob/8f94109d9826c7c9c30e382056f49c73d7803e33/src/core/lv_obj_property.c#L109
> 
> isn't having `const` cast away here?
> 
> https://github.com/lvgl/lvgl/blob/8f94109d9826c7c9c30e382056f49c73d7803e33/src/core/lv_obj_property.c#L124
> 
No sadly because `obj_property` is doing double duty as both a getter and setter. Where the getter branch requires a non const pointer to be passed in.

> Perhaps all you could do to fix it would be
> 
> ```c
> lv_property_t value_copy = *value;
> return obj_property(obj, value->id, &value_copy, true);
> ```
Functionally, yes. However I feel this is a wasteful workaround that doesn't really address either of the two issues I have with this code. The first one being, and this is my personal opinion, the code paths for get and set should have been kept separate. Merging them like done here via `obj_property`, only diverging on a bool at various places in the code, is inviting issues like the one being fixed here to happen. It may be that this way was necessary because of code size, or some other embedded domain requirement that I'll admit I'm not that familiar with, in which case this is a pure logic issue and should be fixed as such and as you correctly pointed out earlier, tests should be provided to cover the weakness of the approach taken.

Now, any help on my CMake issue would be greatly appreciated so I can get on with writing and running tests for this part of the API :)

AndreCostaaa:
Hey, I can help you with your cmake issue. Could you please provide your `lv_conf.h` that you are using to compile lvgl?

Also if you're trying to add a new test, the tests can simply be compiled using the `test/main.py` script

andrimarjonsson:
> Hey, I can help you with your cmake issue. Could you please provide the `lv_conf.h` you are using to compile lvgl?
> 
> Also if you're trying to add a new test, the tests can simply be compiled using the `test/main.py` script

Thank you for the offer and suggestion. I actually just figured it out. Long story short, had done some pulls from main lvgl repo and forgotten about the fact that `lv_conf.h` is not under source control. A quick diff and manual local merge with the updated 
`lv_conf_template.h` and all is well again.

I'll take a look at the test runner you mentioned as well, once I'm done with writing the test.

liamHowatt (commented):
Thank you.

Sorry, I see that it really is just a simple case of the semantics of `const lv_property_t * value` being ignored (const is cast away in the function). Usually this kind of thing isn't tested but we can keep the test. Please fix the formatting issue reported by the CI.

Do you think a better way to fix it would be to change

```c
return obj_property(obj, value->id, (lv_property_t *)value, true);
```

to

```c
lv_property_t value_copy = *value;
return obj_property(obj, value->id, &value_copy, true);
```

because `obj_property` is called somewhere else and the "stomping" is desired in that case.

XuNeo:
> ```c
> lv_property_t value_copy = *value;
> return obj_property(obj, value->id, &value_copy, true);
> ```

I think this is a good idea.
And thanks for the fix!

liamHowatt:
Sorry, I see you already tried explaining before, but I'm afraid I don't understand your argument.

>  the code paths for get and set should have been kept separate. Merging them like done here via `obj_property`, only diverging on a bool at various places in the code, is inviting issues like the one being fixed here to happen.

I understand you don't like the way the logic is structured. What is preventing us from fixing the `const *` bug?

> As for the cases where the "stomping" is desired, I assume you are referring to when the caller is ´lv_obj_get_property´. Sure, in that case the caller expects the parameter passed in to be filled by the function.

Yes, the caller does expect that. It sounds like we agree. Can we fix the `const *` bug?

PR #8118, opened 2025-04-16
fix(makefile): Extend component.mk with missing elements

rbudai98:
This commit adds missing elements to component.mk


kisvegabor (approved):
Thank you!

uLipe (approved):

Issue #6763, opened 2024-08-29
v9.3 planning


kisvegabor:
**Update:**
Feature freeze week until June1, Release is planned to June 3. 

### Introduce the problem

**Goals**
I have 2 goals for 2024 Q3 and Q4:
1. Make LVGL more appealing for MPUs: ready-to-use projects, Linux drivers, more docs, heavier features 
2. Improve the developer experience: docs and CI review, more examples, etc

**Schedule**
Targeting 2025 Q1

### Proposal

In light of that, I would like to focus on these features and updates:
**Features**
- [ ] ~Blurring https://github.com/lvgl/lvgl/issues/6018 (deferred as low priority)~
- [x] Tiled rendering https://github.com/lvgl/lvgl/pull/6761
- [x] OpenGL rendering https://github.com/lvgl/lvgl/issues/6017
- [x] Multi-touch (scroll, zoom, rotate) and mouse wheel scroll https://github.com/lvgl/lvgl/issues/6014
- [ ] ~Widget caching  https://github.com/lvgl/lvgl/issues/4910 (deferred in lack time)~
- [x] SVG https://github.com/lvgl/lvgl/issues/6022
- [x] Improve the span widget https://github.com/lvgl/lvgl/issues/6019
- [x] Double and triple click: https://github.com/lvgl/lvgl/pull/6187
- [x] Font manager https://github.com/lvgl/lvgl/issues/6806

**Docs**
- [ ] Review and complete the docs https://github.com/lvgl/lvgl/issues/6671 (partly done)
- [x] Yocto and Build root docs
- [x] Add the docs page for all chip vendor partners
- [ ] Write detailed tutorials about the important features (cancelled)
- [ ] Add more examples where needed (deferred to the next release, collecting ideas now)
 
**Developer experience**
- [x] Have 5 repos for MPU-based dev boards
- [x] Have repos for all the officially supported boards 
- [x] Control the number of open issues and PRs (Adding workflows internally to have better focus on issues/PRs)
- [x] API unification (for the editor and binding generation)
- [ ] ~Clean up the CI https://github.com/lvgl/lvgl/issues/5396 (deferred n lack of time)~
- [ ] ~Separate lv_conf.h to platform and UI-specific parts https://github.com/lvgl/lvgl/issues/6010 (cancelled)~
- [x] Have a demo for MPUs (Full HD)

---

Leave a comment if you would like to see other features too in v9.3. 

Also, let us know if you are interested in contributing to any of these features. 

[...]

PR #8122, opened 2025-04-16
feat(nxp/g2d): performance and function optimization

anaGrad:
Added DMA double buffer implementation for Wayland driver.

### Notes


kisvegabor:
Is this critical for you the have [in v9.3](https://github.com/lvgl/lvgl/issues/6763#issuecomment-2802811878)?

uLipe:
Just did a pass, the changes are not too trivial, so I suggest to post-pone it, even though it looks good to me.

nicusorcitu:
This PR improves the performance on wayland buffering, it offers support for dma buffers to avoid any extra memcopy.

Should be any reason to postpone? If there is a wayland rework in progress, we can delay this PR. But I believe this improvement is good to have it in 9.3.

kisvegabor:
I see, thanks. @etag4048 please take a look at this PR.

cosmindanielradu19:
Hi,
Are there any concerns regarding this PR being merged in 9.3 release? Is there anything that should be modified before being approved?
Thanks

kisvegabor:
@etag4048 thank you! 
I'd like it to be part of v9.3 so I wonder if you can look into it this week.

src/drivers/wayland/lv_wayland.c
```diff
@@ -2918,7 +2919,7 @@ bool lv_wayland_timer_handler(void)
         }
     }
 
-    return true;
+    return idle_time;
```
ghost: The return type of this function was `bool`. It returned `false` when the window is hidden (behind another window or minimized).. Now it returns the time until the next tick. However, not visible in the diff  it still returns `false` `lv_wayland_timer_handler:2901` and on line `2878` let's change the return type to `int32_t` and return `-1` when the window is hidden or closed. Please see how it is used here:  https://github.com/lvgl/lv_port_linux/blob/dcdb0b66a74225398934c301e4b83bb13330f9cf/src/lib/display_backends/wayland.c#L130

src/drivers/wayland/lv_wayland.c
```diff
@@ -2608,6 +2625,23 @@ lv_display_t * lv_wayland_window_create(uint32_t hor_res, uint32_t ver_res, char
...
+                                    stride);
+    window->lv_disp_draw_buf2 = lv_draw_buf_create(
+                                    hor_res,
+                                    ver_res,
+                                    lv_display_get_color_format(window->lv_disp),
+                                    stride);
+
+
+    lv_display_set_draw_buffers(window->lv_disp, window->lv_disp_draw_buf1, window->lv_disp_draw_buf2);
+    lv_display_set_render_mode(window->lv_disp, LV_DISPLAY_RENDER_MODE_FULL);
```
ghost: Wouldn't `LV_DISPLAY_RENDER_MODE_DIRECT` be faster? to only modify the areas that have changed directly in the draw buffer. `MODE_FULL` redraws the entire window on each call. Please see the early return in `lv_inv_area` in `lv_refr.c`

src/drivers/wayland/lv_wayland.c
```diff
@@ -2599,15 +2869,26 @@ lv_display_t * lv_wayland_window_create(uint32_t hor_res, uint32_t ver_res, char
...
 
-    window->lv_disp_draw_buf = lv_draw_buf_create(
-                                   hor_res,
-                                   ver_res / LVGL_DRAW_BUFFER_DIV,
-                                   lv_display_get_color_format(window->lv_disp),
-                                   stride);
+#if LV_WAYLAND_BUF_COUNT == 1
+    window->body->buffers[0]->lv_draw_buf = lv_draw_buf_create(
+                                                hor_res,
+                                                ver_res,
```
ghost: `ver_res` should be divided by `LVGL_DRAW_BUFFER_DIV` here

src/drivers/wayland/lv_wayland.c
```diff
@@ -2002,14 +2193,25 @@ static bool resize_window(struct window * window, int width, int height)
...
 
-        window->lv_disp_draw_buf = lv_draw_buf_reshape(
-                                       window->lv_disp_draw_buf,
-                                       lv_display_get_color_format(window->lv_disp),
-                                       width, height / LVGL_DRAW_BUFFER_DIV, stride);
+
+        window->body->buffers[0]->lv_draw_buf = lv_draw_buf_resize(
+                                                    window->body->buffers[0]->lv_draw_buf,
+                                                    lv_display_get_color_format(window->lv_disp),
+                                                    width, height / LVGL_DRAW_BUFFER_DIV, stride);
```
ghost: missing  `lv_display_set_draw_buffers(window->lv_disp, window->body->buffers[0]->lv_draw_buf, NULL)` here - omitting it causes crashes on window resizing.

src/draw/lv_draw_buf.c
```diff
@@ -397,6 +412,28 @@ lv_draw_buf_t * lv_draw_buf_reshape(lv_draw_buf_t * draw_buf, lv_color_format_t
     return draw_buf;
 }
 
+lv_draw_buf_t * lv_draw_buf_resize(lv_draw_buf_t * draw_buf, lv_color_format_t cf, uint32_t w, uint32_t h,
```
ghost: I don't see the need to add a resize function that doesn't really resize, but simply destroys and re-creates the buffer. Destroying and re-creating a buffer can be handled on the driver side.

src/drivers/wayland/lv_wayland.c
anaGrad: We tried testing with both render mode full and direct. Although direct does bring some improvement in the render time, there is a big spike in CPU usage for some of the tests in the benchmark. We are planning on finding the origin of this spikes, but until then, we decided to keep render mode full, as it is still a big improvement from render mode partial. Please see the results in the table below. A good example of the spikes I am talking about is the multiple labels test.

LV_DISPLAY_RENDER_MODE_FULL
| Name | Avg. CPU | Avg. FPS | Avg. time | render time | flush time |
| --- | --- | --- | --- | --- | --- |
| Empty screen | 16% | 53 | 1 | 1 | 0 |
| Moving wallpaper | 10% | 60 | 2 | 2 | 0 |
| Single rectangle | 7% | 60 | 1 | 1 | 0 |
| Multiple rectangles | 12% | 60 | 3 | 3 | 0 |
| Multiple RGB images | 13% | 59 | 3 | 3 | 0 |
| Multiple ARGB images | 11% | 60 | 3 | 3 | 0 |
| Rotated ARGB images | 15% | 60 | 3 | 3 | 0 |
| Multiple labels | 14% | 60 | 3 | 3 | 0 |
| Screen sized text | 24% | 60 | 6 | 6 | 0 |
| Multiple arcs | 9% | 60 | 1 | 1 | 0 |
| Containers | 15% | 59 | 4 | 4 | 0 |
| Containers with overlay | 23% | 57 | 11 | 11 | 0 |
| Containers with opa | 19% | 59 | 6 | 6 | 0 |
| Containers with opa_layer | 27% | 59 | 9 | 9 | 0 |
| Containers with scrolling | 17% | 60 | 3 | 3 | 0 |
| Widgets demo | 35% | 55 | 8 | 8 | 0 |
| All scenes avg. |16% | 58 | 4 | 4 | 0 |

LV_DISPLAY_RENDER_MODE_DIRECT
| Name | Avg. CPU | Avg. FPS | Avg. time | render time | flush time |
| --- | --- | --- | --- | --- | --- |
| Empty screen | 17% | 53 | 1 | 1 | 0 |
| Moving wallpaper | 10% | 60 | 2 | 2 | 0 |
| Single rectangle | 6% | 60 | 0 | 0 | 0 |
| Multiple rectangles | 17% | 60 | 3 | 2 | 1 |
| Multiple RGB images | 21% | 60 | 2 | 2 | 0 |
| Multiple ARGB images | 21% | 59 | 2 | 2 | 0 |
| Rotated ARGB images | 15% | 59 | 2 | 2 | 0 |
| Multiple labels | 51% | 60 | 11 | 8 | 3 |
| Screen sized text | 23% | 59 | 6 | 6 | 0 |
| Multiple arcs | 30% | 59 | 1 | 1 | 0 |
| Containers | 12% | 59 | 1 | 1 | 0 |
| Containers with overlay | 24% | 57 | 11 | 11 | 0 |
| Containers with opa | 15% | 60 | 1 | 1 | 0 |
| Containers with opa_layer | 16% | 59 | 3 | 3 | 0 |
| Containers with scrolling | 16% | 60 | 3 | 3 | 0 |
| Widgets demo | 31% | 60 | 6 | 6 | 0 |
| All scenes avg. |20% | 59 | 3 | 3 | 0 |

anaGrad:
> Well the news is not good, it broke the normal non dma buf operation. Investigating why.

Hi @etag4048 ! I handled your comments in the last 4 commits. Let me know if there is anything else you would like to discuss regarding this PR.

kisvegabor:
Hi,

Which  `LV_SYSMON_GET_IDLE` do you use?
Please make sure that you are using ` lv_os_get_idle_percent` and enable `LV_USE_OS LV_OS_PTHREAD`. It might affect the reported CPU usage.

anaGrad:
> Hi,
> 
> Which `LV_SYSMON_GET_IDLE` do you use? Please make sure that you are using ` lv_os_get_idle_percent` and enable `LV_USE_OS LV_OS_PTHREAD`. It might affect the reported CPU usage.

Hi @kisvegabor ! The results from the tables are extracted from running with `lv_os_get_idle_percent` and `LV_OS_PHTREAD` set in lv_conf.

nicusorcitu:
I agree the wayland need a drastic refactoring, but for now this optimization adds a boost on the g2d implementation.
Can be this PR included in LVGL 9.3?
Thanks.

kisvegabor:
If Wayland needs to be reworked anyways and this PR wont't make the current state worse in general, I'm ok.

@etag4048 @andrecostaa please review and test it on some of our dev boards.

AndreCostaaa:
> @etag4048 @andrecostaa please review and test it on some of our dev boards.

I will take a look at it this afternoon
Btw you tagged the wrong person :smiling_face_with_tear:

andrecostaa:
It's true! Lol you tagged the wrong person this comment

AndreCostaaa:
Hello everyone

After an internal discussion there are some concerns about the `wayland` driver changes. I'm worried that the conditional compilation additions will make this driver that much harder to maintain. A major rewrite is planned and this change could make it that much more difficult to maintain and actually rewrite it so I propose taking some time to really get a grasp of this driver and to actually propose a good architecture around `dmabuf`. What i have in mind is separating the driver into multiple parts:

- `lv_wayland_common.c` - Common code
- `lv_wayland_dmabuf.c` - Code specific to `dmabuf`
- `lv_wayland.c` - Keep this the same way as it is in current master branch minus the common code that would have been moved to `lv_wayland_common.c`

This would, in my opinion, create an easier way to maintain this driver and to be able to completely rewrite it later without fear of breaking your changes

If you have any concerns about this approach, please let me know, i'm happy to discuss it further

nicusorcitu:
Hi @AndreCostaaa 
We aimed to include these optimization changes into lvgl 9.3 - so definitely there is no time to rework the code as you proposed.
dma-buf add a good improvement in performance, even the code is not perfect.
How would you like to continue with these changes?
Thank you for reviewing it.
Nicu.

AndreCostaaa:
Hi @nicusorcitu, to be clear, we also want to add this changes to v9.3. I will take on the refactoring responsibility and we'll check the progress mid next week and take a decision based on the status then

Just to be sure, is everything you wanted to change included in this PR already or are you still planning on modifying something ?

anaGrad:
> Hi @nicusorcitu, to be clear, we also want to add this changes to v9.3. I will take on the refactoring responsibility and we'll check the progress mid next week and take a decision based on the status then
> 
> Just to be sure, is everything you wanted to change included in this PR already or are you still planning on modifying something ?

There are no more changes we plan to add to this PR.

src/drivers/wayland/lv_wayland.c
```diff
@@ -1260,6 +1293,25 @@ static const struct xdg_wm_base_listener xdg_wm_base_listener = {
...
+static void
+dmabuf_format(void * data, struct zwp_linux_dmabuf_v1 * zwp_linux_dmabuf, uint32_t format)
+{
+    /* XXX: deprecated */
+}
+
+static const struct zwp_linux_dmabuf_v1_listener dmabuf_listener = {
+    dmabuf_format,
+    dmabuf_modifiers
+};
```
AndreCostaaa: Deprecated but still used for versions < 4, so we should probably support them _for now_ just in case
We also need to use the feedback listener to make sure that the color format set by the user is available for `dmabuf`

See the current `shm_format` function to see what it should be done here, when using `dmabuf` it's my understanding we shouldn't use `shm`

AndreCostaaa (changes requested):
Hi there,

The way i see it, we either use SHM or DMABUF based on the LV_WAYLAND_USE_DMABUF config.
In the future I would like to try to do this in runtime but for now, let's keep things simple

For that, the create window decorations function still uses `smm`. Can you replace this with `dmabuf` ?

https://github.com/lvgl/lvgl/blob/94bdc2daa4f33c2b1fa821aca67d3bc47d97ec66/src/drivers/wayland/lv_wayland.c#L1819-L1834

Once that's complete, there's no longer a need to initialize smm, so we can fully decouple smm from dmabuf rather than having them coexist:

https://github.com/lvgl/lvgl/blob/94bdc2daa4f33c2b1fa821aca67d3bc47d97ec66/src/drivers/wayland/lv_wayland.c#L2443-L2444

Also we need to actually check for the `dmabuf` supported formats to avoid having weird issues if the format specified by the user is not actually supported by the compositor. You already setup the listener for the format callback for <v4, we need to check the color format there the same way it's done in `shm_format` and the same should be done for v4
with the `feedback` listener: https://wayland.app/protocols/linux-dmabuf-v1#zwp_linux_dmabuf_feedback_v1:event:format_table

Let me know what you think

anaGrad:
> For that, the create window decorations function still uses `smm`. Can you replace this with `dmabuf` ?
> 
Hi! I will get working on this right away. After that, I will look into what you said about the color formats.

anaGrad:
Hi @AndreCostaaa! Upon testing I saw that I have a question related to window decorations. While using the master for lv_port_linux and lvgl, it seems to be missing the buttons from the decorations, even though the code is passing through the logic for the buttons. Is this a known issue?
image

AndreCostaaa:
>  Is this a known issue?

Not really an issue in my system. What compositor is your distribution using ?

anaGrad:
> Not really an issue in my system. What compositor is your distribution using ?

It is using Weston.

AndreCostaaa:
Maybe a weston issue, checking internally if it's known. I can attest that it works correctly on gnome tho
image

AndreCostaaa:
Also, if there's a problem with weston, let's treat it in another PR, if you're able to get the window decorations to look the same with the dmabuf implementation it's perfectly fine

I'll also test it on gnome to make sure there are no regressions there

AndreCostaaa:
Hi again @anaGrad. Just checked and it's _supposed_ to be working

Could you make sure the issue still arises after doing a clean build and could you tell me the weston version you're running ?

anaGrad:
> Hi again @anaGrad. Just checked and it's _supposed_ to be working
> 
> Could you make sure the issue still arises after doing a clean build and could you tell me the weston version you're running ?

I tried a clean build using master branches from both lvgl and lv_port_linux and the behavior is the same. I am running weston version 12.0.4.

anaGrad:
> Could you provide me:
> 
> 1. Your `lv_conf.h`
> 2. The application and wayland logs using WAYLAND_DEBUG=1
> 3. Your weston.ini config file

Sure thing, here are the files requested: 

[lv_conf.txt](https://github.com/user-attachments/files/20326765/lv_conf.txt)
[Wayland_debug_logs.txt](https://github.com/user-attachments/files/20326771/Wayland_debug_logs.txt)
[weston_ini.txt](https://github.com/user-attachments/files/20326774/weston.txt)

src/drivers/wayland/lv_wayland.c
anaGrad: Regarding this subject, I first want to mention that I am testing with version 3. 
The listener for format is not called even for this version, but I added the support you mentioned in the modifier listener, since one of its parameters is the color format. 
Now, according to linux dma-buf protocol documentation, for version > 4 there should be another listener used and that is `get_default_feedback`. I also added an implementation for that, but I cannot exactly test it since I only run version 3.
The question is, would you like me to also keep the implementation for version > 4? If so, should I also do the registry bind with the according version? At the moment it is forced to 3.

anaGrad:
Hi @AndreCostaaa ! I just now realized that my board does not have support for RGB565, therefor showing the wrong colors. So, maybe the buttons are there, but I wouldn't be able to see them since the whole bar is black. That being said, I am so sorry for sending you on a wild goose hunt. I did however try to run with `LV_COLOR_DEPTH` 32, and even with or without setting by hand the display color format to `LV_COLOR_FORMAT_ARGB8888`, but in this case there is no bar showing whatsoever.

AndreCostaaa:
> Hi @AndreCostaaa ! I just now realized that my board does not have support for RGB565, therefor showing the wrong colors. So, maybe the buttons are there, but I wouldn't be able to see them since the whole bar is black. That being said, I am so sorry for sending you on a wild goose hunt. I did however try to run with `LV_COLOR_DEPTH` 32, and even with or without setting by hand the display color format to `LV_COLOR_FORMAT_ARGB8888`, but in this case there is no bar showing whatsoever.

Hi, no worries, had just started looking at it. I'm still unable to reproduce any error. Did you modify your `lv_conf.h` ? `LV_WAYLAND_WINDOW_DECORATIONS` must be enabled in your for the decorations to show up

Can you send me the new logs using COLOR_DEPTH 32 ?

src/drivers/wayland/lv_wayland.c
AndreCostaaa: > Now, according to linux dma-buf protocol documentation, for version > 4 there should be another listener used and that is `get_default_feedback`. I also added an implementation for that, but I cannot exactly test it since I only run version 3.

No need. Incremental changes are good, no need to support everything. If you're using v3 we should at least support v1-v3 as they're the same. v4 support can come in later, in a new PR

> The listener for format is not called even for this version, but I added the support you mentioned in the modifier listener, since one of its parameters is the color format.

Are you sure your binding the listener ? You should see something like 
 ```
 -> wl_registry@2.bind(17, "zwp_linux_dmabuf_v1", 3, new id [unknown]@4)
 ```
 in your app logs when running with WAYLAND_DEBUG=1
 
 >  If so, should I also do the registry bind with the according version? At the moment it is forced to 3.
 
We should support v1-3 as it can be done using the same implementation and emit a warning `LV_LOG_WARN` if the version is higher than that saying that the version is not supported

anaGrad: Yes, I am binding the listener, and I do see that line in my logs, but after that I only get the modifier, and not the format. I have added a commit for the formats, with two listeners, one for v1-2 and one for v3, since for v3 the modifiers are called, and for < 3 there are no modifiers and format should be present. I will also attach here the logs I receive, maybe I am missing something.
[zwp_v3_logs.txt](https://github.com/user-attachments/files/20366921/zwp_v3_logs.txt)

anaGrad:
> I also can't seem to reproduce the error using your weston config so, i believe this has nothing to do with how LVGL is handling things. I understand that you can't implement the window decorations yourself as you can't test them so i propose pushing your latest changes and i can take it over from there ? What do you think ?

If you have the time to also look into creating the decorations using dma buf, that's fine by me. I will also keep looking into this subject, at least try and make it look the same for RGB565.

AndreCostaaa:
Just wanted to update the plans for this. I'll finish it at the end of this week and when it's ready we can merge it even during the feature freeze before 9.3. Thank you for taking the time to handle some of the concerns I had

Plan:

- [ ] Use dmabuf for window decorations
- [ ] Integrate #8241 changes
- [ ] Add v4 support as that's what i have on my machine currently

I'll keep you updated and thank you :rocket:

anaGrad:
> Just wanted to update the plans for this. I'll finish it at the end of this week and when it's ready we can merge it even during the feature freeze before 9.3. Thank you for taking the time to handle some of the concerns I had
> 
> Plan:
> 
> * [ ]  Use dmabuf for window decorations
> * [ ]  Integrate [refactor(wayland): separate driver into multiple files #8241](https://github.com/lvgl/lvgl/pull/8241) changes
> * [ ]  Add v4 support as that's what i have on my machine currently
> 
> I'll keep you updated and thank you 🚀

By adding v4 support are you refering to checking the format in case of v4 `zwp_linux_dmabuf_v1_inerface`? If so, I can add a commit for that since I already tried an implementation, if you want to test it on your machine since you have version 4.
Thank you for your help.

anaGrad:
> > If so, I can add a commit for that since I already tried an implementation, if you want to test it on your machine since you have version 4.
> 
> Sure thing, go ahead :)

I added the commit, let me know if it works on your machine. Thank you!

AndreCostaaa:
Hi. Please check and test https://github.com/nxp-upstream/lvgl/pull/1. If it's okay merge the changes here directly

tests/main.py
```diff
@@ -93,6 +93,30 @@ def gen_wayland_protocols(clean):
...
+        subprocess.check_call(['wayland-scanner',
+            'client-header',
+            os.path.join(wayland_protocols_dir, "unstable/linux-dmabuf/linux-dmabuf-unstable-v1.xml"),
+            os.path.join(wayland_dir, "linux-dmabuf-unstable-v1-client-protocol.h.original"),
+        ])
+
+        subprocess.check_call(['wayland-scanner',
+            'private-code',
+            os.path.join(wayland_protocols_dir, "unstable/linux-dmabuf/linux-dmabuf-unstable-v1.xml"),
+            os.path.join(wayland_dir, "linux-dmabuf-unstable-v1-protocol.c.original"),
```
AndreCostaaa: You don't have this in your stable folder ?

```console
$ file /usr/share/wayland-protocols/stable/linux-dmabuf/linux-dmabuf-v1.xml
/usr/share/wayland-protocols/stable/linux-dmabuf/linux-dmabuf-v1.xml: XML 1.0 document, Unicode text, UTF-8 text
```

tests/main.py
```diff
@@ -93,6 +93,30 @@ def gen_wayland_protocols(clean):
...
+        ])
+
+        subprocess.check_call(['wayland-scanner',
+            'client-header',
+            os.path.join(wayland_protocols_dir, "unstable/linux-explicit-synchronization/linux-explicit-synchronization-unstable-v1.xml"),
+            os.path.join(wayland_dir, "linux-explicit-synchronization-unstable-v1-client-protocol.h.original"),
```
AndreCostaaa: Is there any way to avoid using this unstable protocol ?

src/drivers/wayland/lv_wayland.h
```diff
@@ -127,7 +127,7 @@ lv_indev_t * lv_wayland_get_touchscreen(lv_display_t * disp);
  * @return true: if the cycle was completed, false if the application
  * went to sleep because the last frame wasn't completed
  */
-bool lv_wayland_timer_handler(void);
+int32_t lv_wayland_timer_handler(void);
```
AndreCostaaa: Is there a strong reason to change this? It's kind of an api breaking change

src/draw/lv_draw_buf.c
```diff
@@ -67,7 +67,8 @@ void lv_draw_buf_handlers_init(lv_draw_buf_handlers_t * handlers,
                                lv_draw_buf_align_cb align_pointer_cb,
                                lv_draw_buf_cache_operation_cb invalidate_cache_cb,
                                lv_draw_buf_cache_operation_cb flush_cache_cb,
-                               lv_draw_buf_width_to_stride_cb width_to_stride_cb)
+                               lv_draw_buf_width_to_stride_cb width_to_stride_cb,
+                               lv_draw_buf_get_fd_cb get_buf_fd_cb)
```
AndreCostaaa: I have some concerns about this:

This is currently used by the `driver` to obtain a file descriptor from the `draw_unit`. However, it feels like the `draw_unit` might not be the right component to provide this file descriptor. It might make more sense for another `driver` to handle this responsibility

For example, accessing another dmabuf is could be done through GBM, which involves a different `driver`. In that context, it would be awkward to establish the connection through the `draw_buf`.

We don't need a perfect solution right now, and this approach can work for the time being. However, to future-proof the design, we should make the handler more generic - perhaps by passing data using a void *. That way, we can continue to use the same handler even if the underlying implementation changes later on. Not sure what to name it tho, any ideas?

AndreCostaaa (changes requested):
Some additional things I noticed while doing the refactoring

src/drivers/wayland/lv_wayland.h
anaGrad: Yes, because the wayland loop was waiting more than it was needed. It was `LV_DEF_REFR_PERIOD * 1000`, and with this change it can wait only the time returned by `lv_wayland_timer_handler()`. Please have a look also at this [PR](https://github.com/lvgl/lv_port_linux/pull/84)

tests/main.py
anaGrad: No, I do not have `linux-dmabuf` directory in the `stable` directory. It's present only in the `unstable`.

tests/main.py
anaGrad: Unfortunately no, the entire `dmabuf` implementation is built around using the unstable protocol.

src/draw/lv_draw_buf.c
anaGrad: I am a little bit unsure about what the issue is here. I understand that the file descriptor can be accessed in another ways and not always through the `draw_unit`, but how would making the handler more generic change anything, as long as the handlers are related to draw units? We can change the name and pass the data as void*, but I don't see how that would change if we were to access the fd through GBM, we just wouldn't call the draw unit handler right?

AndreCostaaa: > but how would making the handler more generic change anything, as long as the handlers are related to draw units?

If we add this now, we’ll be committed to maintaining it, even if we come up with a better solution later. Removing it could break things for anyone who starts relying on it. However, if we design it to be more generic from the start, it opens the door for other use cases and avoids this limitation

I was imagining renaming it to something like `lv_draw_buf_get_draw_unit_data_cb`

src/drivers/wayland/lv_wayland.h
AndreCostaaa: Just thinking out loud here. If we return a `uint32_t` we can return `0` where we previously return `false` and the user can immediately recall this function again which works and is compatible with the previous version
Returning any other value means the caller can sleep for the value returned amount of time which gives more information than the previous version so this shouldn't break any user code already using this function. It would only mean that they will be using 3 bytes more than currently for the return value. Am i missing something ?

Having said that do you agree with changing this function's return type to `uint32_t` @kisvegabor ?

Note i'm saying `uint32_t` because, if we're making this change we should make the API cleaner and thus return the same type as `lv_timer_handler` and not yet another type

tests/main.py
AndreCostaaa: It's okay, probably an older version of `wayland-protocols` on your machine. This protocol is considered stable now

tests/main.py
AndreCostaaa: Actually, I believe we can as you don't actually use it in the code, see this commit in the other PR https://github.com/nxp-upstream/lvgl/pull/1/commits/9fe496409c070be84230162fa24d70ce99299430

src/drivers/wayland/lv_wayland.h
anaGrad: So this discussion was an older [comment](https://github.com/lvgl/lvgl/pull/8122#discussion_r2071863181). Previously I had it at `uint32_t` and I changed it to `int32_t` to send `-1` instead of false for cases like window is hidden or closed.

src/draw/lv_draw_buf.c
anaGrad: Now I understand what you are referring to. We can do this change.

kisvegabor:
Hey all,

After having a discussion with @andrecostaa and @liamHowatt I understood that there are still a few things to discuss here. As we are in significant delay with releasing v9.3 I have announced a feature freeze week for this week. 

Unfortunately, it means that this PR can be part of only the next release. :slightly_frowning_face: 

The good news is that, we set a hard time limit for v9.4 so you don't have to wait a year for that. Our target is having a feature freeze week at the first 2 weeks of this September, and having the release mid September. 

I'm sorry that we couldn't merge it this...

kisvegabor (dismissed):
I can't comment on the Wayland part but from the high level point of view it's all good. 
Let's add docs in a new PR later.

src/drivers/wayland/lv_wayland.c
```diff
@@ -1,28 +1,49 @@
...
+#endif
+
+#if LV_WAYLAND_USE_DMABUF && LV_WAYLAND_WINDOW_DECORATIONS
+    #error "LV_WAYLAND_USE_DMABUF doesn't support LV_WAYLAND_WINDOW_DECORATIONS"
+#endif
+
+#ifndef LV_DISPLAY_RENDER_MODE_PARTIAL
+    /* FIXME: Hacky fix else building fails with -Wundef=error*/
+    #define LV_DISPLAY_RENDER_MODE_PARTIAL 0
+#endif
```
AndreCostaaa:
```suggestion
#ifndef LV_DISPLAY_RENDER_MODE_PARTIAL
    /* FIXME: Hacky fix else building fails with -Wundef=error and the following checks don't work*/
    #define LV_DISPLAY_RENDER_MODE_PARTIAL 0
    #define LV_DISPLAY_RENDER_MODE_DIRECT 1
    #define LV_DISPLAY_RENDER_MODE_FULL 2
#endif
```

AndreCostaaa (changes requested):

src/drivers/wayland/lv_wayland.c
AndreCostaaa: Hi @anaGrad, could you apply this patch ? After it we can merge this. Thanks

nicusorcitu: @cosmindanielradu19

cosmindanielradu19: Hi @AndreCostaaa , I've pushed the requested change. Thanks

AndreCostaaa (approved):
Thank you everyone :rocket:

cosmindanielradu19:
> Thank you everyone 🚀

Thanks for the support.

src/drivers/wayland/lv_wl_shm.c
```diff
@@ -0,0 +1,465 @@
...
+    }
+}
+
+static void handle_wl_buffer_release(void * data, struct wl_buffer * wl_buffer)
+{
+    const struct smm_buffer_properties * props;
+    struct graphic_object * obj;
+    struct window * window;
+    smm_buffer_t * buf;
+
```
AndreCostaaa, line 319:
```suggestion
    /* window is unused when LV_LOG level is not set to trace */
    LV_UNUSED(window);
    LV_UNUSED(wl_buffer);

```

AndreCostaaa (changes requested):
Sorry, after doing some more testing I also noticed this was causing warnings @cosmindanielradu19 

All should be good after this one

src/drivers/wayland/lv_wl_shm.c
cosmindanielradu19: Done

AndreCostaaa (dismissed):

cosmindanielradu19:
Please don't forget that this PR https://github.com/lvgl/lv_port_linux/pull/84 also is needed for the update of the `lv_wayland_timer_handler` and the build of the dmabuf sources. It will be updated as soon as this PR is merged.

Thanks

AndreCostaaa:
Haven't forgotten about it. I was waiting for the cmake update that conflicts with yours. I merged the cmake update this morning so you should be able to rebase and update it accordingly. We can continue over there if you have any questions. Will take time to review it this evening

nicusorcitu:
Thank you @AndreCostaaa

cosmindanielradu19:
@AndreCostaaa I've fixed all the build warnings for both G2D and DMA_BUF in this PR, and also rebased and updated https://github.com/lvgl/lv_port_linux/pull/84

uLipe (approved):

AndreCostaaa (approved):
I guess we can add the width and height back to the dmabuf window resize function when it's necessary

Also nice to see that adding compilation warnings to the port linux repository is bearing fruit

Thank you once more :)

PR #8126, opened 2025-04-17
feat(cmake): disable PCPP by default

etag4048:
- Disable by default the python pre-processor (PCPP) from creating cmake variables from lv_conf_internal.h
- Add an option to create the cmake variables in the parent scope
- Make the script more robust by only matching 0|1 and if it fails in `unexpected ways` add a debug option.

### DIFF
--- a/env_support/cmake/custom.cmake
+++ b/env_support/cmake/custom.cmake
@@ -14,6 +14,7 @@ find_package(Python REQUIRED)
 
 option(LV_CONF_SKIP "Skip including lv_conf.h during configuration" OFF)
 option(LV_USE_KCONFIG "Use Kconfig to configure LVGL" OFF)
+option(LV_CMAKE_CREATE_CONF_VARS "Create variables from the options defined in lv_conf_internal.h - required for KConfig" OFF)
 
 # Option LV_CONF_PATH, which should be the path for lv_conf.h
 # If set parent path LV_CONF_DIR is added to includes
@@ -41,10 +42,15 @@ if ( LV_USE_KCONFIG )
   # and autoconf.h, which will be used by lv_conf_kconfig.h
   include(${CMAKE_CURRENT_LIST_DIR}/kconfig.cmake)
 
-  set(LV_KCONFIG_IGNORE OFF)
+  # If KConfig is used - creating CMAKE variables from lv_conf_internal.h is required,
+  # so PCPP is used to pre-process lv_conf_internal.h
+  set(LV_CMAKE_CREATE_CONF_VARS TRUE)
+
   # Set the flag to specify we are using kconfig, needed for the
   # generate_cmake_variables.py script
+  set(LV_KCONFIG_IGNORE OFF)
   set(KCONFIG_FLAG --kconfig)
+
   # If using Kconfig, we need to define additional definitions
   list(APPEND PCPP_DEFINITIONS_LIST "LV_CONF_SKIP" "LV_CONF_KCONFIG_EXTERNAL_INCLUDE=\"${LV_CONF_KCONFIG_EXTERNAL_INCLUDE}\"")
 else()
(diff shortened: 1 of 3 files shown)


kisvegabor:
Thank you! I can test it later. 
cc @Daxidz @AndreCostaaa @liamHowatt

env_support/cmake/custom.cmake
```diff
@@ -41,10 +42,15 @@ if ( LV_USE_KCONFIG )
   # and autoconf.h, which will be used by lv_conf_kconfig.h
   include(${CMAKE_CURRENT_LIST_DIR}/kconfig.cmake)
 
-  set(LV_KCONFIG_IGNORE OFF)
+  # If KConfig is used - creating CMAKE variables from lv_conf_internal.h is required,
+  # so PCPP is used to pre-process lv_conf_internal.h
+  set(LV_CMAKE_CREATE_VARIABLES TRUE)
```
AndreCostaaa:
```suggestion
  set(LV_CMAKE_CREATE_CONF_VARS ON)
```

Should be this right ?

scripts/generate_cmake_variables.py
```diff
@@ -37,7 +43,15 @@ def get_args():
 
     return args
 
-def generate_cmake_variables(path_input: str, path_output: str, kconfig: bool):
+def write_set_cmd(fout, expr, is_parent_scope):
+
+    fout.write(f'set({expr})\n')
+
+    # This makes the variable usable from the top level directory
+    if is_parent_scope == True:
```
AndreCostaaa, line 51:
```suggestion
    if is_parent_scope:
```

Small nit

scripts/generate_cmake_variables.py
```diff
@@ -100,10 +115,16 @@ def generate_cmake_variables(path_input: str, path_output: str, kconfig: bool):
             name = parts[1]
             value = parts[2].strip()
 
-            fout.write(f'set({CONFIG_PREFIX}{name} {value})\n')
+            write_set_cmd(fout, f'{CONFIG_PREFIX}{name} {value}', is_parent_scope)
+
+        else:
+            # Useful for debugging expressions that are unhandled,
+            # if the script fails in 'unexpected ways'
+            if debug == True:
```
AndreCostaaa, line 123: Same nit as before

AndreCostaaa (changes requested):
I appreciate the addition of the `parentscope` and `debug` flags. Nice work !

scripts/generate_cmake_variables.py
ghost: I don't agree, I think that ``== True`` is clearer to read, I always prefer to be explicit in languages that have a dynamic type system. Such as javascript or python.  Don't have to scroll to the top to check what was the type of ``is_parent_scope``. Not only dynamic languages, but also for languages that have a lot of type coercion rules.  This is why for example you will never see me writing ``if (!something)`` in C because it could be a NULL check or a boolean check, And I will have to lookup the type ``something``

env_support/cmake/custom.cmake
ghost: This I am not sure, *I think* ON or TRUE is the same thing - we should pick one and use it when we re-write this currently inconsistent  file soon.

AndreCostaaa: True or on is okay, as long as we keep things consistent

The variable name is wrong tho isn't it?

scripts/generate_cmake_variables.py
AndreCostaaa: Sounds reasonable enough for me :)

env_support/cmake/custom.cmake
ghost: Ah yes indeed, my bad - must have slipped by. Haven't tested Kconfig stuff.

kisvegabor (dismissed):

scripts/generate_cmake_variables.py
```diff
@@ -87,11 +101,12 @@ def generate_cmake_variables(path_input: str, path_output: str, kconfig: bool):
...
-            fout.write(f'set({CONFIG_PREFIX}{name} 1)\n')
+            write_set_cmd(fout, f'{CONFIG_PREFIX}{name} 1', is_parent_scope)
 
-        # For the rest of config, simply add CONFIG_ and write
-        # all LV_USE_* configs, as these are the one needed in cmake
-        elif line.startswith(f'{CONFIG_PATTERN}'):
+        # For the rest of the configs, simply add CONFIG_ and write the name of the define
+        # all LV_USE_* configs where the value is 0 or 1, as these are the one needed in cmake
+        # To detect the configuration of LVGL to perform conditional compilation/linking
+        elif re.search(f'{CONFIG_PATTERN}.*[01] *$', line):
```
liamHowatt: It will match
```
#define LV_COLOR_DEPTH 1
#define LV_FONT_DEFAULT &lv_font_montserrat_10
#define LV_FONT_MANAGER_NAME_MAX_LEN           100
```

```suggestion
        elif re.search(f'{CONFIG_PATTERN}.* +[01] *$', line):
```

liamHowatt: I realized it won't address the `#define LV_COLOR_DEPTH 1` case, but at least the other two cases

ghost: > It will match
> 
> #define LV_COLOR_DEPTH 1
> #define LV_FONT_DEFAULT &lv_font_montserrat_10
> #define LV_FONT_MANAGER_NAME_MAX_LEN           100

@liamHowatt I don't think it will - due to ``CONFIG_PATTERN`` being set to ``#define LV_USE`` on line 64 if Kconfig is not used. But your change of adding one or more spaces won't hurt either.

liamHowatt: Oh nice, okay

liamHowatt (dismissed):

kisvegabor:
Thank you! Please also update cmake's docs accordingly.

kisvegabor (approved):

kisvegabor:
Thank you Erik! @AndreCostaaa please take another look.

AndreCostaaa (approved):
LGTM :rocket:

Issue #7639, opened 2025-01-21
Draw filled triangle (easy way)


focussing:
### LVGL version

v9.2.2

### Platform

Arduino using RP2040

### What happened?

I need to create a process diagram on my LCD and already used a circle and rectangle using quite simple functions:

```
void draw_rect(int i, int x, int y, int w, int h, lv_color_t bg_color, lv_color_t color) {
  rect[i] = lv_obj_create(lv_scr_act());
  lv_obj_set_scrollbar_mode(rect[i], LV_SCROLLBAR_MODE_OFF);
  lv_obj_set_size(rect[i], w, h);
  lv_obj_set_pos(rect[i], x, y);
  lv_obj_set_style_radius(rect[i], 0, 0);
  lv_obj_set_style_bg_color(rect[i], bg_color, 0);
  lv_obj_set_style_border_color(rect[i], color, 0);
  lv_obj_set_style_border_width(rect[i], 2, 0);
  // rect_color(i, color);
}

void draw_circle(int i, int x, int y) {
  circle[i] = lv_obj_create(lv_scr_act());
  lv_obj_set_size(circle[i], 40, 40);
  lv_obj_set_pos(circle[i], x, y);
  lv_obj_set_style_radius(circle[i], LV_RADIUS_CIRCLE, 0);
  lv_obj_set_style_border_color(circle[i], lv_color_black(), 0);
  lv_obj_set_style_border_width(circle[i], 2, 0);
  lv_obj_set_style_bg_color(circle[i], lv_color_white(), 0);
}

```
Now I have to draw a symbol of a valve and did it with the following simple code:

```
lv_obj_t *valve[2];

void draw_valve1(int i, int x, int y) {
  const int W = 30;
  const int H = 20;
  int X_L = x;
  int X_R = x + W;
  int Y_T = y;
  int Y_B = y + H;

  // create line
  static lv_point_precise_t line_points[] = { { X_L, Y_T }, { X_R, Y_B }, { X_R, Y_T }, { X_L, Y_B }, { X_L, Y_T } };
  valve[i] = lv_line_create(lv_screen_active());
  lv_line_set_points(valve[i], line_points, 5);
  lv_obj_add_style(valve[i], &style_line, 0);
}
```

Resulting in this:
![Image](https://github.com/user-attachments/assets/5527c829-b883-47e9-bff5-3579684fd097)

Now I need to fill the triangles with color red to indicate it is closed, and green to indicate open. How can I achieve that?

[...]

PR #8129, opened 2025-04-18
perf(label): reduce the time of calling lv_text_get_size when drawing

yimoxiao:
When drawing a label,  `lv_text_get_size` may be called multiple times to get the size of the text, which may affect the frame rate. However, the size is calculated once when setting the label, the calculation here can be omitted.

### Notes

### DIFF
--- a/src/draw/lv_draw_label.c
+++ b/src/draw/lv_draw_label.c
@@ -200,6 +200,7 @@ void lv_draw_label_iterate_characters(lv_draw_task_t * t, const lv_draw_label_ds
                                       const lv_area_t * coords,
                                       lv_draw_glyph_cb_t cb)
 {
+    lv_draw_dsc_base_t * base_dsc = t->draw_dsc;
     const lv_font_t * font = dsc->font;
     int32_t w;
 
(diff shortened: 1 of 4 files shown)


src/draw/lv_draw_label.c
```diff
@@ -218,10 +218,7 @@ void lv_draw_label_iterate_characters(lv_draw_task_t * t, const lv_draw_label_ds
     }
     else {
         /*If EXPAND is enabled then not limit the text's width to the object's width*/
-        lv_point_t p;
-        lv_text_get_size(&p, dsc->text, dsc->font, dsc->letter_space, dsc->line_space, LV_COORD_MAX,
-                         dsc->flag);
-        w = p.x;
+        w = dsc->text_size.x;
```
liamHowatt: There are cases where `text_size` will be `{0, 0}` here. Is that expected?

I.e. if

```diff
diff --git a/demos/widgets/lv_demo_widgets.c b/demos/widgets/lv_demo_widgets.c
index 3851fe029..bd649324c 100644
--- a/demos/widgets/lv_demo_widgets.c
+++ b/demos/widgets/lv_demo_widgets.c
@@ -1395,6 +1395,7 @@ static void slider_event_cb(lv_event_t * e)
             label_dsc.font = font_normal;
             label_dsc.text = buf;
             label_dsc.text_local = 1;
+            label_dsc.flag |= LV_TEXT_FLAG_EXPAND;
             lv_draw_label(draw_rect_dsc->base.layer, &label_dsc, &txt_area);
         }
     }
```

yimoxiao: Oh, my bad. I overlooked this task creation method and the ADD_TASK event.

yimoxiao: I’ve modified the code so that it now only proceeds when using the draw_main event to create a draw_label task and no add_task event exists.

liamHowatt: Thank you. There are other events where users can create draw tasks, such as `LV_EVENT_DRAW_MAIN_END`, which doesn't require `LV_OBJ_FLAG_SEND_DRAW_TASK_EVENTS`. See this: https://github.com/lvgl/lvgl/issues/7639#issuecomment-2703462084

Can you think of a way to solve it?

yimoxiao: I understand that what you mean is to create a drawing task instead of obj by manually calling lv_draw_xx. If so, then in the drawing descriptor, there will be no obj pointed to by base, so text_size will be recalculated. In this way, it can be distinguished whether the task is created by obj or manually.
image

yimoxiao: @liamHowatt Hi, could you please review it again when you have time?

liamHowatt: Sorry for the delay. I tested and yes it handles all cases I can think of :slightly_smiling_face: thank you!

liamHowatt (approved):
Nicely done! I tested and yes it handles all cases I can think of :slightly_smiling_face: thank you!

FASTSHIFT (approved):

PR #8139, opened 2025-04-22
fix(list): check that LV_USE_FLEX is enabled when using LV_LIST

Wol:
Fixes include error when using `LV_LIST` but `LV_USE_FLEX` isn't set. Same way as how `lv_menu.h` checks for it.

### DIFF
--- a/src/widgets/list/lv_list.h
+++ b/src/widgets/list/lv_list.h
@@ -18,7 +18,7 @@ extern "C" {
 #if LV_USE_LIST
 
 #if LV_USE_FLEX == 0
-#error "LV_USE_FLEX needs to be enabled"
+#error "LV_LIST requires LV_USE_FLEX to be enabled"
 #endif
 
 /*********************
--- a/src/widgets/list/lv_list.h
+++ b/src/widgets/list/lv_list.h
@@ -18,7 +18,7 @@ extern "C" {
 #if LV_USE_LIST
 
 #if LV_USE_FLEX == 0
-#error "LV_LIST requires LV_USE_FLEX to be enabled"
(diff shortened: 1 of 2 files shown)


FASTSHIFT (dismissed):

uLipe (changes requested):
Nice catch, I just have a nit, to add more information about the error.

For example: "LV_LIST requires LV_FLEX to be enabled"

It will save time telling to users on error console what depends on which.

Wol:
Have updated both lv_list and lv_menu with text to clarify which widget was causing the error.

kisvegabor (commented):
Thanks!

src/widgets/list/lv_list.h
```diff
@@ -17,6 +17,10 @@ extern "C" {
 
 #if LV_USE_LIST
 
+#if LV_USE_FLEX == 0
+#error "LV_LIST requires LV_USE_FLEX to be enabled"
```
kisvegabor: Actually we have this format for [other widgets](https://github.com/lvgl/lvgl/blob/master/src/widgets/slider/lv_slider.h#L22)
```suggestion
#error "lv_list: lv_flex is required. Enable it in lv_conf.h (LV_USE_FLEX 1)"
```

Wol: Fair enough! I was copying the style from the lv_menu error rather than the lv_list one. I've updated so those three match. When I get a chance I'll have a look to see if there's other errors now which are mismatched.

liamHowatt (approved):

uLipe (approved):

PR #8145, opened 2025-04-23
fix(scripts): correct parsing of LVGL config file

W-Mai:
## Description

- Modify the parsing logic to join multiple values with a space

### Background

when I parse this configuration like this:

```cpp
#define LV_USE_SYSMON   (LV_USE_MEM_MONITOR | LV_USE_PERF_MONITOR)
```

I got an error:

### Error
image

### Reason

The script parsing like this:
```cpp
['#define', 'LV_USE_SYSMON', '(LV_USE_MEM_MONITOR', '|', 'LV_USE_PERF_MONITOR)']
```

But it only takes the first 3 element. So it causes this error.

## Fix

join the rest parts together to avoid this problem

### Before:

```bash
set(CONFIG_LV_USE_SYSMON (LV_USE_MEM_MONITOR
```

### After:

```bash
set(CONFIG_LV_USE_SYSMON (LV_USE_MEM_MONITOR | LV_USE_PERF_MONITOR))
```

### Notes

### DIFF
--- a/scripts/generate_cmake_variables.py
+++ b/scripts/generate_cmake_variables.py
@@ -98,7 +98,7 @@ def generate_cmake_variables(path_input: str, path_output: str, kconfig: bool):
                 continue
 
             name = parts[1]
-            value = parts[2].strip()
+            value = " ".join(parts[2:]).strip()
 
             fout.write(f'set({CONFIG_PREFIX}{name} {value})\n')
 


FASTSHIFT (approved):

AndreCostaaa:
I'm having a hard time grasping how this works. Please elude me if you don't mind

1. Afaik | is not a valid operator in cmake for setting variables

2. Furthermore i'm also wondering if cmake will correctly evaluate that since LV_USE_MEM_MONITOR and LV_USE_PERF_MONITOR won't actually exist as they will be expanded with the CONFIG_ prefix to CONFIG_LV_USE_XXX in cmake land

3. And lastly, I believe the variables need to be dereferenced using the `${VAR_NAME}` syntax

The way I see it, the correct way to expand that expression to cmake is:

`set(CONFIG_LV_USE_SYSMON ${CONFIG_LV_USE_PERF_MONITOR} OR ${CONFIG_LV_USE_MEM_MONITOR})`

And your change doesn't get us there. While there's work intended to make the script more robust in the near future (after 9.3 is released), in the meantime your lv_conf can simply look like this, which would work

```c
#if LV_USE_MEM_MONITOR || LV_USE_PERF_MONITOR
 #define LV_USE_SYSMON 1
#else
 #define LV_USE_SYSMON 0
#endif
```

As a last fyi, the pcpp is getting disabled by default for v9.3. See #8126

Please, feel free to correct me if I’m missing something - happy to discuss further

W-Mai:
You're right, I tested. It was interpreted as a string, which does not meet the expectation.

And `set(CONFIG_LV_USE_SYSMON ${CONFIG_LV_USE_PERF_MONITOR} OR ${CONFIG_LV_USE_MEM_MONITOR})` is not easy to handle now.

Let me try it, if I have no idea to process this, I will close this pr 😁

W-Mai:
That's not easy. Maybe need to build a DAG to manage the dependences of each variables.

Fine, I give it up. ☹️

AndreCostaaa:
It's definitely not an easy problem to solve, don't worry about it. We need to think about how we want to extend the script as well as there's no point in having a full C preprocessor to cmake converter

AndreCostaaa:
Thank you Erik, TIL. I thought the OR operation thing would work but i guess not

> This is why I would suggest keeping
> the script simple forever and just
> write LV_USE_* defines in a
> compatible manner in 
> lv_conf_template.h

Agreed, current script is simple and thus easy to maintain which is great

Just to clarify, the only way I see us getting something with a bit more depth would be by changing strategy and not by complicating the script

W-Mai:
Overall, it's time for me to close this pr. 🚀

PR #8085, opened 2025-04-11
feat(animimage): support images play in a reversed order

luozhenhua:
Add a "reverse" parameter for "lv_animimg_set_src" function to support playing the animimage in a reversed order.

PR #8164, opened 2025-04-27
feat(animimage) support set source interfaces with or without parameter of reverse play

luozhenhua:
To ensure backwards compatibility of animimage, keep original lv_animimg_set_src function and add lv_animimg_set_src_reverse to support reverse play.

### DIFF
--- a/src/widgets/animimage/lv_animimage.c
+++ b/src/widgets/animimage/lv_animimage.c
@@ -43,6 +43,7 @@
  **********************/
 static void index_change(lv_obj_t * obj, int32_t idx);
 static void lv_animimg_constructor(const lv_obj_class_t * class_p, lv_obj_t * obj);
+static void lv_animimg_set_src_inner(lv_obj_t * obj, const void * dsc[], size_t num, bool reverse);
 
 /**********************
  *  STATIC VARIABLES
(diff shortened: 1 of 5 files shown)


src/widgets/animimage/lv_animimage.c
```diff
@@ -107,18 +107,22 @@ lv_obj_t * lv_animimg_create(lv_obj_t * parent)
...
+    lv_anim_set_values(&animimg->anim, 0, (int32_t)num);
+}
+
+void lv_animimg_set_src_reverse(lv_obj_t * obj, const void * dsc[], size_t num)
+{
+    LV_ASSERT_OBJ(obj, MY_CLASS);
+    lv_animimg_t * animimg = (lv_animimg_t *)obj;
+    animimg->dsc = dsc;
+    animimg->pic_count = num;
+    lv_anim_set_values(&animimg->anim, (int32_t)num, 0);
```
AndreCostaaa: What do you think about refactoring this code out to a separate private (`static`) function with the `bool reverse` argument to avoid code duplication ?

Something like
```c
static void lv_animimg_set_src_inner(lv_obj_t * obj, const void * dsc[], size_t num, bool reverse)
```

Liubin-glb: Hi @AndreCostaaa 
   Thanks for your comment. 
   There is already a commit merged to implement this feature, but @kisvegabor believes that this is a breaking change to the API. Therefore, we submitted this PR to add a new function. For more information, see https://github.com/lvgl/lvgl/pull/8085.
Best Regards
LB

AndreCostaaa: Hi, yes, i'm aware

What i'm proposing is moving the duplicate code to a separate _private_ function and have the _two public_ functions use the private one, essentially this:

```c
//Keep the two public functions so we don't break the api
void lv_animimg_set_src(lv_obj_t * obj, const void * dsc[], size_t num)
{
   lv_animimg_set_src_inner(obj, dsc, num, false);
}
void lv_animimg_set_src_reverse(lv_obj_t * obj, const void * dsc[], size_t num)
{
    lv_animimg_set_src_inner(obj, dsc, num, true);
}

// Note the static here, this function is private and not part of the public api
static void lv_animimg_set_src_inner(lv_obj_t * obj, const void * dsc[], size_t num, bool reverse) 
{
    LV_ASSERT_OBJ(obj, MY_CLASS);
    lv_animimg_t * animimg = (lv_animimg_t *)obj;
    animimg->dsc = dsc;
    animimg->pic_count = num;
    if(reverse) {
        lv_anim_set_values(&animimg->anim, (int32_t)num, 0);
    }
    else {
        lv_anim_set_values(&animimg->anim, 0, (int32_t)num);
    }
 }
```

Essentially avoiding the code duplication as the only thing that changes between the two reverse and non reverse function is the call to `lv_anim_set_values`

What do you think ?

Liubin-glb: Yes I agree. I will resubmit the PR for change this item.

kisvegabor (approved):
Looks good, thank you  for the quick action!

AndreCostaaa (approved):
Thank you :rocket:

PR #8173, opened 2025-04-29
fix(evdev): add support for multi-touch

PeterBee97:
Add support for multi-touch on Linux evdev

Processing multitouch events according to Linux event codes (adding states for slot and mt touch coords and sending them to gesture recognizers).

### DIFF
--- a/src/drivers/evdev/lv_evdev.c
+++ b/src/drivers/evdev/lv_evdev.c
@@ -33,6 +33,7 @@
 #include "../../display/lv_display.h"
 #include "../../display/lv_display_private.h"
 #include "../../widgets/image/lv_image.h"
+#include "../../indev/lv_indev_gesture.h"
 
 /*********************
  *      DEFINES
@@ -43,6 +44,7 @@
 #define EVDEV_DISCOVERY_PATH_BUF_SIZE 32
 #define REL_XY_MASK ((1 << REL_X) | (1 << REL_Y))
 #define ABS_XY_MASK ((1 << ABS_X) | (1 << ABS_Y))
+#define MAX_TOUCH_POINTS 5
 
 /**********************
  *      TYPEDEFS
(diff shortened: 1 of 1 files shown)


liamHowatt (dismissed):
Hmm, looks great! Surprisingly elegant integration! Does it work well when you test?

kisvegabor:
@etag4048 please review it.

PeterBee97:
> Hmm, looks great! Surprisingly elegant integration! Does it work well when you test?

To be fair I'm still having trouble with the gesture recognition. The results are not stable as expected, and I'm still working on it.

niklasf:
Tested this patch *excluding the actual gesture recognition*. No regressions with and without `LV_USE_GESTURE_RECOGNITION` ✅.

When I touch with two or more fingers and `LV_USE_GESTURE_RECOGNITION` is defined, I only get correct touch events, never mixing coordinates of different fingers.

When I touch with two fingers and `LV_USE_GESTURE_RECOGNITION` is not defined, I get events mixing the coordinates of both fingers (same without this patch). But maybe we can still respect `ABS_MT_SLOT` even in this case?

PeterBee97:
Just fixed logic issues in this patch, now gesture looks good on my touchscreen. Feel free to test it! @kisvegabor @niklasf @liamHowatt @etag4048 

> 
> When I touch with two fingers and `LV_USE_GESTURE_RECOGNITION` is not defined, I get events mixing the coordinates of both fingers (same without this patch). But maybe we can still respect `ABS_MT_SLOT` even in this case?

I would love to add ABS_MT_SLOT to the original implementation, however I'm afraid it would break compatibility with old devices. Maybe leave it for next PR?

niklasf:
> Could also try the gesture recognition part with the `lv_example_gestures`?

Tested now and it generally works as far as this PR is concerned. The touch points appear to be passed on correctly.

In the gesture example itself, I find it quite difficult to trigger anything other than rotation (which works nicely). I can scale only if I *extremely* carefully avoid any rotational component in my motion. Likewise for swiping gestures.

> I would love to add ABS_MT_SLOT to the original implementation, however I'm afraid it would break compatibility with old devices. Maybe leave it for next PR?

Sure, no problem.

kisvegabor (approved):
I've reopened it, sorry for the absence. 

Thank you for the tests. We can improve the sensitivity of touch detection independently.

XuNeo (approved):

PR #8182, opened 2025-04-30
feat(image): support symbol images with inner alignment

liamHowatt:
Support the first 9 inner-aligns and x/y offsets for images whose `src` is a symbol.

### Notes


src/widgets/image/lv_image.c
```diff
@@ -884,7 +884,24 @@ static void draw_image(lv_event_t * e)
...
             lv_obj_init_draw_label_dsc(obj, LV_PART_MAIN, &label_dsc);
             label_dsc.text = img->src;
-            lv_draw_label(layer, &label_dsc, &obj->coords);
+            lv_area_t * coords;
+            lv_area_t aligned_coords;
+            if(((img->align > LV_IMAGE_ALIGN_TOP_LEFT
+                 && !(img->w == lv_area_get_width(&obj->coords)
+                      && img->h == lv_area_get_height(&obj->coords)))
+                || img->offset.x || img->offset.y)
+               && img->align < LV_IMAGE_ALIGN_AUTO_TRANSFORM) {
```
AndreCostaaa: Reading this condition is hard, can we refactor the conditions into separate variables so we understand what each part of the condition does?
Something like
```c
const bool has_offset = img->offset.x || img->offset.y;
const bool is_image_area_same_as_coords = img->w == lv_area_get_width(&obj->coords) && img->h == lv_area_get_height(&obj->coords));
...etc
```
Hardest thing is to find the correct variable names

AndreCostaaa (changes requested):
Nice ! Just a small request to make code more readable

src/widgets/image/lv_image.c
liamHowatt: Done. The reason I did it that way was so that short-circuit evaluation would occur, and the `lv_area_get_*` function calls would not actually occur if the other conditions evaluated false. It was a pseudo optimization so I concede.

AndreCostaaa: Curious to see if the compiler can short circuit that even if it's rewritten like this :thinking: 

https://godbolt.org/z/G73xWd6P8 <- will analyze this later

AndreCostaaa (approved):
Thank you :rocket:

src/widgets/image/lv_image.c
liamHowatt: Looks like the compiler inlined the functions in that example because it can see what they do (not a black-box). In order for it to call out to them, the functions must be black-boxes, i.e. declared (prototypes), not actually defined (bodies). Since they're defined in lv_area.c, that's how it'd be in real life, unless someone compiles with link-time optimization enabled.

kisvegabor (approved):
Thank you!

PR #8198, opened 2025-05-04
feat(osal): Support for ChibiOS

dynfer:
Added support for ChibiOS which is a powerfull open-source RTOS. Besides the main OS functions I've implemented the use of the ChibiOS specific drivers for LTDC and DMA2D.


src/osal/lv_chibios.c
```diff
@@ -0,0 +1,340 @@
...
+}
+
+lv_result_t lv_thread_sync_init(lv_thread_sync_t * pxCond)
+{
+    prvCheckCondInit(pxCond);
+    return LV_RESULT_OK;
+}
+
+lv_result_t lv_thread_sync_wait(lv_thread_sync_t * pxCond)
+{
```
uLipe, line 85: I know that chibios support binary semaphores, why not use them instead of condvars+mutex pair?

src/osal/lv_chibios.c
```diff
@@ -0,0 +1,340 @@
...
+}
+#endif
+#if LV_USE_SYSMON
+
+_Static_assert(CH_DBG_STATISTICS == TRUE, "CH_DBG_STATISTICS must be TRUE in chconf.h");
+_Static_assert(CH_CFG_USE_TM == TRUE, "CH_CFG_USE_TM must be TRUE in chconf.h");
+
+static volatile uint32_t last_idle_pct = 0;
+
+static THD_WORKING_AREA(waIdleStat, 128);
```
uLipe, line 206: Hmm I think the stack size should be something exposed via lv_conf/ Kconfig

uLipe (commented):
Nice work! It is good to see that Chibios is being widely used out there.

I have two small comments.

src/osal/lv_chibios.c
dynfer: Thats a valid point, however for a rather static thread that isn't being used outside of the task its designed for I dont see a point.

src/osal/lv_chibios.c
dynfer: I've more or less followed the current FreeRTOS implementation for that. I'm not that advanced with ChibiOS yet, hence im open to suggestions.
I will try to come up with a better solution with binary semaphores.

src/osal/lv_chibios.c
liamHowatt, line 1: Please remove all LTDC code from this file.

src/osal/lv_chibios.c
```diff
@@ -0,0 +1,340 @@
+/**
+ * @file lv_chibios.c
+ *
+ * Port of LVGL OS abstraction layer using ChibiOS.
+ *
+ */
+
+#if LV_USE_OS == LV_OS_CHIBIOS
```
liamHowatt, line 9: `LV_*` macros are undefined.

```suggestion
/*********************
 *      INCLUDES
 *********************/
 
 #include "lv_chibios.h"

#if LV_USE_OS == LV_OS_CHIBIOS
```

src/drivers/display/st_ltdc/lv_st_ltdc.c
```diff
@@ -13,14 +13,35 @@
...
     #if LV_USE_DRAW_DMA2D
         #error cannot use LV_ST_LTDC_USE_DMA2D_FLUSH with LV_USE_DRAW_DMA2D
     #endif /*LV_USE_DRAW_DMA2D*/
-
-    #include "dma2d.h"
+    #if !defined(LV_OS_CHIBIOS)
+        #include "dma2d.h"
+    #elif defined(LV_OS_CHIBIOS)
+        #include "ChibiOS-Contrib/os/hal/ports/STM32/LLD/DMA2Dv1/hal_stm32_dma2d.h"
+    #endif /*LV_OS_CHIBIOS*/
```
liamHowatt: It would be better if the user was responsible for creating dma2d.h and it includes this other header internally.

src/drivers/display/st_ltdc/lv_st_ltdc.c
```diff
@@ -13,14 +13,35 @@
...
+    #define DMA2D_OUTPUT_ARGB8888 DMA2D_FMT_ARGB8888
+    #define DMA2D_OUTPUT_RGB888  DMA2D_FMT_RGB888
+    #define DMA2D_OUTPUT_RGB565  DMA2D_FMT_RGB565
+    #define DMA2D_INPUT_ARGB8888 DMA2D_FMT_ARGB8888
+    #define DMA2D_INPUT_RGB888  DMA2D_FMT_RGB888
+    #define DMA2D_INPUT_RGB565  DMA2D_FMT_RGB565
+    #define DMA2D_INPUT_L8     DMA2D_FMT_L8
+    #define DMA2D_INPUT_AL88   DMA2D_FMT_AL88
+    #define DMA2D_INPUT_A8 DMA2D_FMT_A8
+#endif
```
liamHowatt: Why do these have different names? Is it because it's LTDCv1 and this file assumes v2?

src/osal/lv_chibios.c
```diff
@@ -0,0 +1,340 @@
...
+    .line_isr = NULL,  /**< Line Interrupt ISR, or @p NULL.*/
+    .rr_isr = (ltdc_isrcb_t)reload_event_callback_handler,    /**< Register Reload ISR, or @p NULL.*/
+    .fuerr_isr = NULL, /**< FIFO Underrun ISR, or @p NULL.*/
+    .terr_isr = NULL,  /**< Transfer Error ISR, or @p NULL.*/
+
+    /* Color and layer settings.*/
+    .clear_color = LTDC_COLOR_TEAL,
+    .bg_laycfg = &ltdc_view_laycfg1,
+    .fg_laycfg = NULL,
+};
```
liamHowatt: These types of custom hardware configs belong in board project port repositories.

liamHowatt (commented):
Thanks for adding support for this.

lv_conf_internal.h is generated from lv_conf_template.h with lv_conf_internal_gen.py. Add a comment here:

https://github.com/lvgl/lvgl/blob/bb243789070c491338ed72b787d5c1773b09e584/lv_conf_template.h#L101-L109

And add ChibiOS here:

https://github.com/lvgl/lvgl/blob/bb243789070c491338ed72b787d5c1773b09e584/scripts/lv_conf_internal_gen.py#L49-L58

and run lv_conf_internal_gen.py to generate lv_conf_internal.h.

Additionally update the Kconfig.

Add a chibios.rst [here](https://github.com/lvgl/lvgl/tree/bb243789070c491338ed72b787d5c1773b09e584/docs/src/details/integration/os) similar to freertos.rst and add an entry to index.rst (in alphabetical order).

kisvegabor:
Thanks for the PR! 

A lot of things here comes down to an earlier suggestion of @uLipe about we should create a better interrupt and periphery abstraction layer. 

**For interrupts**: I think it's simpler as we can add some defines or so for creating interrupts 
**For peripheries**: supporting STM HAL, ChibiOS, Zephyr, NuttX, etc with `#if`s is not a good practice, but I have no idea how to build a good abstraction over the other RTOS abstractions here.

src/drivers/display/st_ltdc/lv_st_ltdc.c
dynfer: > Why do these have different names? Is it because it's LTDCv1 and this file assumes v2?

Those names are defined by the specific driver implementations in ChibiOS, in certain cases those differ in naming to the standard names used by ST HAL.

src/osal/lv_chibios.c
dynfer: > These types of custom hardware configs belong in board project port repositories.

Would You rather preffer to leave the config in place and make the values accessible via a config file?

src/osal/lv_chibios.c
dynfer: What would be the desired place for it then? In the current from I've streamlined it for the user so its done automatically (besides the need to define specific configuration for the given screen which could be done via lv_conf for example)?

src/drivers/display/st_ltdc/lv_st_ltdc.c
dynfer: How do You suggest to notify the user about this need? A ChibiOS specific section on the Wiki?

src/osal/lv_chibios.c
liamHowatt: The place this would go is typically in LVGL docs, and copied to a user's own code file. The docs can be in a chibios.rst. See My above comment:

> Add a chibios.rst [here](https://github.com/lvgl/lvgl/tree/bb243789070c491338ed72b787d5c1773b09e584/docs/src/details/integration/os) similar to freertos.rst and add an entry to index.rst (in alphabetical order).

src/drivers/display/st_ltdc/lv_st_ltdc.c
liamHowatt: In LVGL DMA2D docs and/or LVGL ChibiOS docs

src/drivers/display/st_ltdc/lv_st_ltdc.c
liamHowatt: Okay

src/drivers/display/st_ltdc/lv_st_ltdc.c
liamHowatt: Actually, I think this one can be ok.

dynfer:
> Thanks for adding support for this.
> 
> lv_conf_internal.h is generated from lv_conf_template.h with lv_conf_internal_gen.py. Add a comment here:
> 
> https://github.com/lvgl/lvgl/blob/bb243789070c491338ed72b787d5c1773b09e584/lv_conf_template.h#L101-L109
> 
> And add ChibiOS here:
> 
> https://github.com/lvgl/lvgl/blob/bb243789070c491338ed72b787d5c1773b09e584/scripts/lv_conf_internal_gen.py#L49-L58
> 
> and run lv_conf_internal_gen.py to generate lv_conf_internal.h.
> 
> Additionally update the Kconfig.
> 
> Add a chibios.rst [here](https://github.com/lvgl/lvgl/tree/bb243789070c491338ed72b787d5c1773b09e584/docs/src/details/integration/os) similar to freertos.rst and add an entry to index.rst (in alphabetical order).

Updated

src/drivers/display/st_ltdc/lv_st_ltdc.c
```diff
@@ -143,21 +187,34 @@ static void flush_cb(lv_display_t * disp, const lv_area_t * area, uint8_t * px_m
...
         if(lv_display_is_double_buffered(disp) && lv_display_flush_is_last(disp)) {
+#if !defined(LV_OS_CHIBIOS)
             HAL_LTDC_SetAddress_NoReload(&hltdc, (uint32_t)px_map, layer_idx);
             g_data.layer_interrupt_is_owned[layer_idx] = true;
             HAL_LTDC_Reload(&hltdc, LTDC_RELOAD_VERTICAL_BLANKING);
+#elif defined(LV_OS_CHIBIOS)
+            ltdcBgSetFrameAddress(&LTDCD1, (uint32_t *)px_map);
+            g_data.layer_interrupt_is_owned[layer_idx] = true;
+            ltdcStartReload(&LTDCD1, false);
+#endif
```
liamHowatt: `LV_OS_CHIBIOS` is always defined as `8`, even if it's not enabled.

Anywhere in this PR that there is `#if defined(LV_OS_CHIBIOS)`, please use `#if LV_USE_OS == LV_OS_CHIBIOS` instead.

docs/src/details/integration/os/chibios.rst
```diff
@@ -0,0 +1,178 @@
...
+========================================
+
+1. Create a thread for the lv_timer_handler
+/*
+   This thread is responsible for calling the lv_timer_handler() function
+   at the calculated interval. The interval is determined by the LVGL library
+   based on the current state of the GUI and the timers that are running.
+   */
+static THD_WORKING_AREA(waLV_TIMER_HANDLER, 256);
+static THD_FUNCTION(LV_TIMER_HANDLER, arg)
```
liamHowatt: In rst (ReStructuredText) files, please use `.. code-block:: c` for code blocks and indent the code blocks 4 spaces.

```suggestion
1. Create a thread for the lv_timer_handler

.. code-block:: c

    /*
       This thread is responsible for calling the lv_timer_handler() function
       at the calculated interval. The interval is determined by the LVGL library
       based on the current state of the GUI and the timers that are running.
       */
    static THD_WORKING_AREA(waLV_TIMER_HANDLER, 256);
    static THD_FUNCTION(LV_TIMER_HANDLER, arg)
```

This is how it currently looks. You can optionally build the docs locally by `pip3 install -r docs/requirements.txt` and then `python3 docs/build.py html skip_api` and view the result in a browser.

image

src/osal/lv_chibios.c
```diff
@@ -0,0 +1,255 @@
...
+        uint64_t delta_idle = idle - prev_idle;
+
+        last_idle_pct = (uint32_t)((delta_idle * 100UL) / delta_total);
+
+        prev_total = total;
+        prev_idle = idle;
+    }
+}
+
+void idleStatInit(void)
```
liamHowatt:
```suggestion
static void idleStatInit(void)
```

liamHowatt (commented):
Thanks for accepting the feedback. Do you have any questions or concerns?

docs/src/details/integration/os/chibios.rst
```diff
@@ -0,0 +1,231 @@
...
+
+How to integrate LVGL with ChibiOS
+========================================
+
+1. Create a thread for the lv_timer_handler
+========================================
+
+.. line-block::
+    This thread is responsible for calling the lv_timer_handler() function at the calculated interval.
+    The interval is determined by the LVGL library based on the current state of the GUI and the timers that are running.
```
liamHowatt: No need for special directives for normal paragraphs.

```suggestion
This thread is responsible for calling the lv_timer_handler() function at the calculated interval.
The interval is determined by the LVGL library based on the current state of the GUI and the timers that are running.
```

docs/src/details/integration/os/chibios.rst
```diff
@@ -0,0 +1,231 @@
+========
+ChibiOS
+========
+
+How to integrate LVGL with ChibiOS
+========================================
+
+1. Create a thread for the lv_timer_handler
+========================================
```
liamHowatt: Please make all heading underlines the same length as the heading text. It's an effort that vwheeler63 is trying to maintain.

Also, we use `***` for first headings, `---` for nested headings, and `~~~` for further nested headings.

```suggestion
=======
ChibiOS
=======

How to integrate LVGL with ChibiOS
**********************************

1. Create a thread for the lv_timer_handler
*******************************************
```

dynfer:
I've added documentation to use delegates which are a more efficient way instead of mutexes.

src/drivers/display/st_ltdc/lv_st_ltdc.c
```diff
@@ -13,14 +13,35 @@
 #include "lv_st_ltdc.h"
 #include "../../../display/lv_display_private.h"
 #include "../../../draw/sw/lv_draw_sw.h"
-#include "ltdc.h"
+#if LV_USE_OS != LV_OS_CHIBIOS
+    #include "ltdc.h"
+#elif LV_USE_OS == LV_OS_CHIBIOS
+    #include "ChibiOS-Contrib/os/hal/ports/STM32/LLD/LTDCv1/hal_stm32_ltdc.h"
```
liamHowatt: I realized `"ltdc.h"` is not a standard header, so since users are already required to create a header anyways, I think that should still be the case for ChibiOS.

src/drivers/display/st_ltdc/lv_st_ltdc.c
```diff
@@ -120,15 +155,24 @@ static lv_display_t * create(void * buf1, void * buf2, uint32_t buf_size, uint32
...
     }
     else {
         lv_display_set_buffers(disp, buf1, buf2, buf_size, LV_DISPLAY_RENDER_MODE_PARTIAL);
 
 #if LV_ST_LTDC_USE_DMA2D_FLUSH
+#if LV_USE_OS != LV_OS_CHIBIOS
         hdma2d.XferCpltCallback = transfer_complete_callback;
+#elif LV_USE_OS == LV_OS_CHIBIOS
+        nvicEnableVector(DMA2D_IRQn, 11);
+#endif
```
liamHowatt: Is `11` portable to all STM devices with DMA2D?

liamHowatt (commented):
I still have a few issues/questions.

- Why does a thread need to be created to measure idle time? Can the cumulative thread time logic just be in the idle time getter?
- How can `is_initialized` be reliable?
- Why are initializer functions protected by `chSysLock()`? Are they really likely to be initialized inside interrupt handlers?
- There is no FreeRTOS code in LTDC. Why does there need to be ChibiOS code in it?

src/drivers/display/st_ltdc/lv_st_ltdc.c
dynfer: This actually can be removed, I've noticed the driver does it already.

src/drivers/display/st_ltdc/lv_st_ltdc.c
dynfer: > I realized `"ltdc.h"` is not a standard header, so since users are already required to create a header anyways, I think that should still be the case for ChibiOS.

Would You rather preffer it in lv_conf to specialize our ltdc header?

dynfer:
> I still have a few issues/questions.
> 
>     * Why does a thread need to be created to measure idle time? Can the cumulative thread time logic just be in the idle time getter?
> 
>     * How can `is_initialized` be reliable?
> 
>     * Why are initializer functions protected by `chSysLock()`? Are they really likely to be initialized inside interrupt handlers?
> 
>     * There is no FreeRTOS code in LTDC. Why does there need to be ChibiOS code in it?

1. I've opted for a minimal thread to return the value as fast as possible instead of having it to be processed, this way its calculated when the system is not doing anything other than running the calculation thread.
2. What exactly does concern You?
3. chSysLock is used to prevent the scheduler for performing context switches, IRQ's have a higher priority than the kernel.
4. FreeRTOS uses ST HAL based drivers, while ChibiOS uses their own drivers.

Kconfig
```diff
@@ -143,6 +143,10 @@ menu "LVGL configuration"
 				bool "5: WINDOWS"
 			config LV_OS_MQX
 				bool "6: MQX"
+			config LV_OS_SDL2
+				bool "7: SDL2"
```
uLipe: Is this change related top the scope of this PR?

docs/src/details/integration/os/index.rst
```diff
@@ -14,3 +16,5 @@
     torizon_os
     yocto/index
     zephyr
+    torizon_os
```
uLipe: Is this related to the PR scope?

src/drivers/display/st_ltdc/lv_st_ltdc.c
uLipe: Hmmm I have concerns here, the display drivers should depend only on the hardware specific drivers and the LVGL agnostic code, here we are adding an explicit dependency of the CHIBIOS, I would rather resolve this on lv_conf or create a specific st_ltdc_chibios display driver since the current one assumes the target platform is the plain STMCUBE.

uLipe (changes requested):
Please just check some minor comments in respect of changes that may not have relationship to this PR (maybe leftovers after a merge).

The implementation of the chibios abstraction layer looks good for me, on the other hand I'm concerned about the LTDC handling, as @liamHowatt pointed, the display driver should not have any specific RTOS code inside of it but just platform related code.

I see a way to resolve this by creating a chibios own display driver based on LTDC and move the code to there, it may sound like a duplication at first sight but is not since the LTDC implementation that already exists on LVGL is not compatible to the way the Chibios handle the same peripheral (based on Chibios device driver model).

uLipe (changes requested):
It is almost there, please duplicate the display driver, the st_ltdc should not depend on what OS is being used.

Instead duplicate the display driver and make it compilable only if chibios is enabled and use the chibios header files and definitions. 

To duplicate you can create a folder called st_ltdc_chibios and move the c and .h files to there, in this way we would not violate the SRP principle of the display drivers since a driver of LCD compiled by chibios is not intended to run on any other platform.

src/drivers/display/st_ltdc/lv_st_ltdc.h
```diff
@@ -15,7 +15,7 @@ extern "C" {
  *********************/
 
 #include "../../../lv_conf_internal.h"
-#if LV_USE_ST_LTDC
+#if LV_USE_ST_LTDC && LV_USE_OS != LV_OS_CHIBIOS
```
uLipe: I think this is not mandatory, LTDC for chibios is something that is not intented to run in other scenarios.

But the contrary is possible :)

src/drivers/display/st_ltdc/lv_st_ltdc.c
```diff
@@ -8,7 +8,7 @@
  *********************/
 
 #include "../../../lv_conf_internal.h"
-#if LV_USE_ST_LTDC
+#if LV_USE_ST_LTDC && LV_USE_OS != LV_OS_CHIBIOS
```
uLipe: Not necessary.

uLipe (changes requested):
Thanks for splitting the driver, I just have two small nits then I think it is good to go.

src/drivers/display/st_ltdc/lv_st_ltdc.h
dynfer: But how do we prevent the compiler from using the generic code?

uLipe: We don't need to, usage of the display driver is a user setting, so the application is responsible for it.

uLipe:
Please fix the build issues on the CI @dynfer , happy to approve after that :)

kisvegabor:
Since we can't really see a clean way forward to handle ChibiOS in LTDC I close this PR. 

In other cases
- e.g. in Zephyr the LCD handling is abstracted out by the OS, so we need just a generic Zephyr interface. 
- e.g. in FreeRTOS, it's not handled at all, so we rely on the vendor's SDK

We need to decide which direction we want to take, as mixing the two seems messy.

Anyways, thank you very much for the collaboration, I hope we can warm it it up soon.

ghost:
Nice. What about Nuttx?

AndreCostaaa:
Hi @DrBuhu 

What about it? It's already supported

PR #8201, opened 2025-05-05
fix(wayland) - support version 2 of the XDG protocol

etag4048:
fix lv_port_linux#62 By supporting an older version of the wayland XDG shell protocol, to make it function on RaspberryPi OS without patches.


AndreCostaaa (approved):
Thanks :rocket:

kisvegabor (approved):
Thanks!

PR #8204, opened 2025-05-05
Refactor VGLite and adding static bitmap font rendering

kisvegabor:
Continue https://github.com/lvgl/lvgl/pull/8156

### Notes


src/draw/lv_draw_label.c
```diff
@@ -592,22 +592,18 @@ void lv_draw_unit_draw_letter(lv_draw_task_t * t, lv_draw_glyph_dsc_t * dsc,  co
 
     if(g.resolved_font) {
         lv_draw_buf_t * draw_buf = NULL;
-        /*If LVGL uses static bitmaps, the buffer allocation and deallocation is not necessary*/
-        const bool static_bitmap = lv_font_has_static_bitmap(g.resolved_font);
-        if(!static_bitmap) {
```
cosmindanielradu19: Exactly this if avoid useless memory allocation when it's not necessary, because the static bitmap can be rendered straight from the address of the glyph.

src/font/lv_font_fmt_txt.c
```diff
@@ -98,21 +98,20 @@ const void * lv_font_get_bitmap_fmt_txt(lv_font_glyph_dsc_t * g_dsc, lv_draw_buf
 
     const lv_font_fmt_txt_glyph_dsc_t * gdsc = &fdsc->glyph_dsc[gid];
 
-    if(g_dsc->req_raw_bitmap) return &fdsc->glyph_bitmap[gdsc->bitmap_index];
```
cosmindanielradu19, line 101: The whole idea of #8156 was to use direct access to the glyph using this statement, which returns the address of the glyph's buffer without the need of copying or iterating through it.

src/draw/lv_draw_label.c
kisvegabor: I've just added [e2471ae](https://github.com/lvgl/lvgl/pull/8204/commits/e2471ae62ff0cd4df5b4ad6b27ec036d403e8e65)
This way have typically 1 allocation per label which is acceptable IMO.

src/font/lv_font_fmt_txt.c
kisvegabor: Ahhh, how could I miss the gist? :facepalm: 
Added here: [968e9c1](https://github.com/lvgl/lvgl/pull/8204/commits/968e9c16b3a12bedc9f03767376c830dfd97ef9d)

src/draw/lv_draw_label.c
cosmindanielradu19: Why is this allocation needed?
In case of a static_bitmap, the `draw_buf` is allocated and deallocated without being used at all. For efficiency, I believe that the allocation is not needed.

cosmindanielradu19:
I've added two more commits to my PR, can you rebase?

Also, do you mind if after rebasing I'll add one more change on top of your changes for calling `lv_font_get_glyph_static_bitmap` from vglite?

kisvegabor:
Hey,

I've just rebased it. There were a conflict though. [Here](https://github.com/lvgl/lvgl/pull/8156/commits/ebd2a6064d8b74227307a9d4083c2521ad27e3d7) you pass the font stride as it is to the glyph_dsc. E.g. 4 or 8. 

[In my version ](https://github.com/lvgl/lvgl/pull/8204/files#diff-9d50f3563437f6e2a4ab70489e05c3c633d9a88267036929eddf37f968ef1b4cR260-R262) is pass the actual stride of the given glyph. E.g. for `font->stride == 4`, `glyph->box_w == 9` it will be `glyph->stride == 12`. This method is more inline with other stride handling as we store the actual stride in all other cases too. 

In VGLite something like `if(g->stride & 0b111 == 0)`  can be used to quickly check for 8 byte aligned stride.  

Is it ok for you?

cosmindanielradu19:
I've synced my PR with this one, and added the last commit with the call to `lv_font_get_glyph_static_bitmap`. I think this should be fine for now. You can either get the last commit and merge this PR, or you can merge the one created by me if everything is fine from your side.

kisvegabor:
Thank you! Let's continue in yours.

PR #8064, opened 2025-04-06
test(perf): add perf test framework

AndreCostaaa:
This PR introduces performance testing support for LVGL using an emulated ARM environment, integrated into the existing testing infrastructure. 

The PR adds:

- A new script to generate required LVGL test files and mount them in a Docker container.
- Integration of this script with the main test runner to ensure a seamless testing flow.
- A new GitHub Actions workflow to automate running the performance tests in CI.
- LVGL and the performance tests are built and executed inside the container using QEMU + SO3.

Docker is used to avoid having too many dependencies locally

The requirements for running these tests locally are docker and a host linux computer.
Right now, this script requires a host linux computer as we depend on `losetup` which is used to set up and control loop devices for creating a virtual disk that is mounted inside QEMU

Note that instead of modifying the current `tests/main.py` script to accomodate for these new tests. I've created a new one as the way the tests are setup is too different and i believe it would be really messy to merge the two

Opening this as draft for now as it requires https://github.com/smartobjectoriented/so3/pull/148 to be merged but i still wanted to see if i broke any other parts of the CI

PR #8209, opened 2025-05-06
test: dockerfile with CI env

AndreCostaaa:
This adds a Dockerfile matching the CI setup to run tests

- Refactored the requirements into separate `.txt` files because to avoid having the same list in the dockerfile and in the shell script (`sudo` doesn't work inside docker)
- Updated the docs in the `tests/README.md` file


tests/README.md
```diff
@@ -4,21 +4,38 @@ The tests in the folder can be run locally and automatically by GitHub CI.
...
+
+2. Run the tests
+
+```bash
+docker run --rm -it -v $(pwd):/work lvgl_test_env bash -c "./main.py"
+```
+
+This ensures you are testing in a consistent environment with the same dependencies as the CI pipeline.
+
+### Host
```
kisvegabor: What does `Host` mean here?

AndreCostaaa: Running directly on the host machine
If you're asking, it probably means it's not the best title here. What would you suggest ?

kisvegabor: I think most of the user are happy with running the tests natively. So I'd start with the native way of running the tests, and mention Docker as an option after that. E.g.

## Running locally

### Native

### Docker

AndreCostaaa: Sounds good to me

liamHowatt: I can imagine "Local" being a good name too. The opposite of "installing in Docker" is "installing locally". Native implies that Docker will emulate something.

tests/README.md
```diff
@@ -27,6 +26,24 @@ scripts/install-prerequisites.sh
...
+1. Build it
+
+```bash
+docker build . -f tests/Dockerfile -t lvgl_test_env
+```
+
+2. Run the tests
+
+```bash
+docker run --rm -it -v $(pwd):/work lvgl_test_env bash -c "./main.py"
```
liamHowatt:
```
bash: line 1: ./main.py: No such file or directory
```

liamHowatt (commented):
Nice!

tests/README.md
AndreCostaaa: I agree

tests/README.md
AndreCostaaa: Oh yea, I was thinking if the docs should explain how to build the image inside the tests folder (where the Dockerfile is) or from the lvgl folder. I arrived at the conclusion that it makes more sense to work from the lvgl folder as that way we can easily do `-v $(pwd):/work` which I believe is cleaner

Anyway I forgot to update the command after making the switch. Thanks for catching that :rocket:

liamHowatt (approved):
Thanks! How do you usually deal with the root privilege build artifacts? I do `sudo rm -rf tests/build* tests/wayland_protocols` or `sudo git clean -Xf .`

AndreCostaaa:
Simplest way is the `sudo rm -rf tests/build*` but you can also use docker to do it for you:
`docker run --rm -it -v $(pwd):/work lvgl_test_env bash -c "rm -rf tests/build* tests/wayland_protocols"`

In this case we have to do something like this because we are mounting the whole contents of `$(pwd)`
The better approach would be to exclude the build artifacts from the mounted volume so they don't exist locally - but you still want to create a docker volume for them so you can rebuild using the previous build cache. What I did in #8064 to solve this exact problem is mount the source files inside the container but create a docker volume for the build artifacts using `docker volume create`. This of course could be done here but we would need to wrap this in another script which I believe is out of scope for this PR

It also creates a gigantic list of volumes that we need to explicitly add which is harder to wrap our mind around:
image

AndreCostaaa:
I believe the CI issue is unrelated, something got broken with vcpkg (?)

kisvegabor (approved):

kisvegabor:
I wanted to schedule the merge until for when the CI passes, but it was merged immediately. Anyway, the CI issues seems like only a temporal one.

PR #8210, opened 2025-05-06
build: restructure cmake

AndreCostaaa:
Continue from #8193


CMakeLists.txt
```diff
@@ -20,7 +20,8 @@ elseif(ZEPHYR_BASE)
 elseif(MICROPY_DIR)
   include(${CMAKE_CURRENT_LIST_DIR}/env_support/cmake/micropython.cmake)
 else()
-  include(${CMAKE_CURRENT_LIST_DIR}/env_support/cmake/custom.cmake)
+  # Build for desktop OSes (Unix-like or Windows)
+  include(${CMAKE_CURRENT_LIST_DIR}/env_support/cmake/os.cmake)
```
uLipe: Maybe renaming the os.cmake to os_desktop.cmake should make this line self explanatory avoiding to document inline of the cmake files

scripts/preprocess_lv_conf_internal.py
```diff
@@ -57,13 +60,23 @@ def preprocess_file(pcpp_exe, input_file, tmp_file, output_file, include_dirs, d
...
 
-def remove_indentation(tmp_file, output_file):
+# Read the temporary files and return an array of lines
+def read_lines(tmp_file):
 
     try:
         with open(tmp_file, "r") as f:
             lines = f.readlines()
+    except Exception as e:
+        print(f"Failed to pre-process file: {e}")
```
uLipe, line 70: is not possible to pick also the reason why the pre-process was failed? It will be a valuable resource when things go wrong.

uLipe (commented):
Nice work, just some small nits.

CMakeLists.txt
kisvegabor: My suggestion was `generic.cmake`.

docs/src/details/integration/building/cmake.rst
```diff
@@ -132,58 +131,96 @@ This snippet adds LVGL and needs an ``lv_conf.h`` file present next to the lvgl
...
 
-    set(LV_CONF_INCLUDE_SIMPLE OFF)
     add_subdirectory(lvgl)
 
 
 This snippet adds LVGL and needs an ``lv_conf.h`` file present in lvgl/src folder:
 
 .. code-block:: cmake
 
+    set(LV_BUILD_CONF_DIR lvgl/src)
```
kisvegabor: IS there scenario when adding `lv_conf.h` `lvgl/src` makes sense?

lv_conf_template.h
```diff
@@ -1294,13 +1294,19 @@
...
-*==================*/
+/*=====================
+* BUILD OPTIONS
+*======================*/
 
 /** Enable examples to be built with the library. */
 #define LV_BUILD_EXAMPLES 1
 
+/** Build the demos */
+#define LV_BUILD_DEMOS 1
```
kisvegabor, line 1305: This define is not used in LVGL. I believe it's needed only for CMake. Can we avoid adding it here.

lv_conf_template.h
```diff
@@ -1294,13 +1294,19 @@
...
+
+/** Build the tests */
+#define LV_BUILD_TEST 1
```
kisvegabor: We shouldn't build test by default. The build will even fail in managed builds this way when all C files are collected automatically. 

This define shouldn't be public.

kisvegabor (changes requested):

scripts/preprocess_lv_conf_internal.py
AndreCostaaa: Don't you think printing the exception is enough ?

AndreCostaaa:
Putting this back to draft as there are some concerns and I can't get to them right at this moment
Will come back to this at the end of the week

kisvegabor:
We have an old issue about restructuring CI:  https://github.com/lvgl/lvgl/issues/5396
We can figure out new configs in which we want to run the tests. 

E.g. since v9  supports all color formats at runtime I think we don't need Build only tests for each color format. Instead we should have ~3-4 configs in which we really run the tests. One of them can use the CMake build options that you have mentioned too.

CMakeLists.txt
ghost: I would go for os_desktop.cmake

docs/src/details/integration/building/cmake.rst
ghost: Nope can be removed, the only scenario where lv_conf is located in the lvgl directory is when the user clones lvgl and builds using CMake directly (without using the project from a parent CMakelists.txt )

lv_conf_template.h
AndreCostaaa: I think it's worth start using it in LVGL. This way, users can simply set it to 0 to disable demo builds - just like they do with examples. It would also streamline the setup process, allowing users to configure everything in lv_conf.h and run `cmake ..` without needing to pass extra options
What do you think ?

lv_conf_template.h
AndreCostaaa: Agreed. Removed all mentions of `LV_BUILD_TEST` outside the tests directory

AndreCostaaa:
> The idea was that all cmake variables that begin with `CONFIG_` are also options from lv_conf.h - When PCPP is used to get the values from `lv_conf.h` those variables are set automatically. But they are still declared to the default values at the start of `os.cmake` to ensure that if the user not willing to use PCPP - still can decide if the examples or ThorVG are to be built.

Okay yes, this way we can keep `pcpp` optional

AndreCostaaa:
I’ve spent some time testing the CMake options and I’m comfortable with the current state of the PR

The only potential improvement I can think of is to wrap all demo files with the LV_BUILD_DEMOS flag. This would prevent the demos from being compiled when the flag is not set—similar to how we currently handle the examples

To clarify: while demo compilation is already disabled via CMake by setting the define to 0 in lv_conf and using the LV_BUILD_SET_CONFIG_OPTS flag in cmake, adding LV_BUILD_DEMOS to the demo files would allow us to disable demo compilation in other build systems as well

lv_conf_template.h
kisvegabor: As the demos are larger and the linker always eliminates unused code we have [`LV_USE_DEMO_...` for each demo](https://github.com/lvgl/lvgl/blob/master/lv_conf_template.h#L1308-L1363) to enable them one by one. 

Having `LV_BUILD_DEMOS` feels like a duplication. 

If I understand correctly the purpose of `LV_BUILD_DEMOS` is to easily exclude the compilation of the `demos` folder from CMake.  If so can we keep this option on CMake level somehow? Maybe as an exception.

AndreCostaaa: The option exists in CMake but it would make life easier if everything could be set using a single config file

kisvegabor: What about wrapping all [LV_USE_DEMO_](https://github.com/lvgl/lvgl/blob/master/lv_conf_template.h#L1308-L1363)s to `LV_BUILD_DEMOS`?

To only issue is that if `LV_BUILD_DEMOS 0` `LV_USE_DEMO_WIDGETS` will be undefined so 
```c
#if LV_BUILD_DEMOS && LV_USE_DEMO_WIDGETS
```
might trigger `LV_USE_DEMO_WIDGETS` is not defined warning.

AndreCostaaa: Yes, I like it, to avoid the "not defined warning" we can add this macro:

```c
#define LV_USE_DEMO(name)                                   \
	((LV_BUILD_DEMOS && defined(LV_USE_DEMO_##name)) && \
	 (LV_USE_DEMO_##name))
```

And then use it like

```c
#if LV_USE_DEMO(WIDGETS)
#if LV_USE_DEMO(SMARTWATCH)
...
```

What do you think ?

kisvegabor: If it's needed for better CMake experience and maintainability I'm ok with it. Just call it `LV_DEMO_IS_ENABLED` and make sure that `LV_BUILD_DEMOS`is 1 for backward compatibility.

AndreCostaaa: Done :100:

AndreCostaaa: Nevermind, expanding a macro with a `defined` operator inside is not defined by the C standard. My bad on that one

Would it be okay to wrap each LV_USE_DEMO with LV_BUILD_DEMOS ? Something like

```c
#if LV_BUILD_DEMOS
/* Enable LV_BUILD_DEMOS before enabling one of the demos */
#define LV_USE_DEMO_SMARTWATCH 0 // <- User value defined here
#define LV_USE_DEMO_WIDGETS 0 // <- User value defined here
...
#else
/* Do not modify this branch */
#define LV_USE_DEMO_SMARTWATCH 0
#define LV_USE_DEMO_WIDGETS 0
...
#endif
```

kisvegabor (dismissed):
Looks good to me!

kisvegabor:
cc @etag4048 @uLipe @liamHowatt pleasae take a look

docs/src/details/integration/building/cmake.rst
```diff
@@ -190,7 +188,7 @@ Below is a list of the available options/variables
      - Allows to set a directory containing ``lv_conf.h``
    * - LV_BUILD_USE_KCONFIG
      - BOOLEAN
-     - When set KConfig is used as the configuration source
+     - When set KConfig is used as the configuration source. This options is disabled by default.
```
ghost: This option, without 's'

AndreCostaaa:
```suggestion
     - When set KConfig is used as the configuration source. This option is disabled by default.
```

etag4048 (approved):
Looks good.

kisvegabor (dismissed):

kisvegabor:
If it causes any issues we should find it in the feature freeze week.

env_support/cmake/os_desktop.cmake
liamHowatt, line 1: I want to challenge this rename. It will be what's used for all CMake projects that are not ESP or Zephyr. The RP2040 build system uses CMake and it will include this `os_dekstop.cmake`.

`custom.cmake` is what I used when I built LVGL for [baremetal Raspberry Pi](https://github.com/rsta2/circle).

Kconfig
```diff
@@ -2067,6 +2067,10 @@ menu "LVGL configuration"
 	endmenu
 
 	menu "Demos"
+		config LV_BUILD_DEMOS
+			bool "Build demos"
+			default y if !LV_CONF_MINIMAL
+
 		config LV_USE_DEMO_WIDGETS
 			bool "Show some widget"
```
liamHowatt:
```suggestion
	menu "Demos"
		config LV_BUILD_DEMOS
			bool "Build demos"
			default y if !LV_CONF_MINIMAL

		if LV_BUILD_DEMOS

		config LV_USE_DEMO_WIDGETS
			bool "Show some widget"
```

With `endif` and the end.

liamHowatt (commented):
I get these warnings with my CMakeLists.txt, but it works.

[CMakeLists.txt](https://github.com/user-attachments/files/20330130/CMakeLists.txt)

```
-- The C compiler identification is GNU 9.4.0
-- The CXX compiler identification is GNU 9.4.0
-- Found assembler: /usr/bin/cc
CMake Warning (dev) at /home/liam/lvgl/lvgl/env_support/cmake/os_desktop.cmake:33 (option):
  Policy CMP0077 is not set: option() honors normal variables.  Run "cmake
  --help-policy CMP0077" for policy details.  Use the cmake_policy command to
  set the policy and suppress this warning.

  For compatibility with older versions of CMake, option is clearing the
  normal variable 'CONFIG_LV_BUILD_DEMOS'.
  /home/liam/lvgl/lvgl/CMakeLists.txt:23 (include)
  CMakeLists.txt:11 (include)
This warning is for project developers.  Use -Wno-dev to suppress it.

CMake Warning (dev) at /home/liam/lvgl/lvgl/env_support/cmake/os_desktop.cmake:34 (option):
  Policy CMP0077 is not set: option() honors normal variables.  Run "cmake
  --help-policy CMP0077" for policy details.  Use the cmake_policy command to
  set the policy and suppress this warning.

  For compatibility with older versions of CMake, option is clearing the
  normal variable 'CONFIG_LV_BUILD_EXAMPLES'.
  /home/liam/lvgl/lvgl/CMakeLists.txt:23 (include)
  CMakeLists.txt:11 (include)
This warning is for project developers.  Use -Wno-dev to suppress it.

-- 
-- Using lv_conf.h from the top-level project directory
-- Enabling the building of ThorVG internal
-- Enabling the building of examples
-- Enabling the building of demos
-- Found Freetype: /usr/lib/x86_64-linux-gnu/libfreetype.so (found version "2.10.1") 
-- Configuring done (0.5s)
-- Generating done (0.1s)
-- Build files have been written to: /home/liam/lvgl/lvgl_workspace/build
```

CMakeLists.txt
AndreCostaaa: @liamHowatt see discussion about rename here

AndreCostaaa:
Okay yea, the way to fix it is by actually telling CMake to modify the cache of these variables in _your_ CMakeLists.txt ie:

Modify these lines:
```cmake
set(CONFIG_LV_BUILD_DEMOS 1)
set(CONFIG_LV_BUILD_EXAMPLES 1)
```
with:
```cmake
set(CONFIG_LV_BUILD_EXAMPLES ON CACHE BOOL "enable examples" FORCE)
set(CONFIG_LV_BUILD_DEMOS ON CACHE BOOL "enable examples" FORCE)
```

The error happens because these variables already exist in CMake's cache. When you try to set them the simple way, CMake doesn't actually update the cache - it just creates regular variables that mask the cache ones. This makes things confusing because different parts of your build might see different values.
Adding CACHE and FORCE tells CMake "hey, I actually want to change the cache value itself." This keeps everything consistent and gets rid of the warning. Docs [here](https://cmake.org/cmake/help/latest/policy/CMP0077.html)

Kconfig
AndreCostaaa: Agreed and done. Thanks :)

kisvegabor (approved):
Looks good, as far as I can tell.

liamHowatt (approved):
> the way to fix it is by actually telling CMake to modify the cache of these variables in _your_ CMakeLists.txt

Ah, user error. Fix confirmed.

CMakeLists.txt
liamHowatt: > My suggestion was `generic.cmake`.

+1

PR #8211, opened 2025-05-06
feat(EVE): Add EVE draw unit

liamHowatt:
### Notes


src/libs/FT800-FT813/EVE_target/EVE_target_ATSAMC21.h
```diff
@@ -0,0 +1,215 @@
+/*
```
kisvegabor: Do we need these targets?

src/libs/FT800-FT813/library.json
```diff
@@ -0,0 +1,22 @@
+{
```
kisvegabor: It can be removed.

src/libs/FT800-FT813/lvgl_version_note.txt
```diff
@@ -0,0 +1,8 @@
+https://github.com/RudolphRiedel/FT800-FT813
```
kisvegabor: Why is this file needed?

src/libs/FT800-FT813/EVE_cpp_target.cpp
```diff
@@ -0,0 +1,616 @@
+#include "../../lv_conf_internal.h"
```
kisvegabor: As we always use it in C this file might not be needed.

src/draw/eve/lv_draw_eve_fill.c
```diff
@@ -0,0 +1,156 @@
...
+    if(bg_w == bg_h && rad == LV_RADIUS_CIRCLE) {
+        lv_eve_draw_circle_simple(coords->x1 + (bg_w / 2), coords->y1 + (bg_h / 2), real_radius);
+    }
+    else {
+        lv_eve_draw_rect_simple(coords->x1, coords->y1, coords->x2, coords->y2, real_radius);
+    }
+
+    lv_eve_restore_context();
+
+    // EVE_end_cmd_burst();
```
kisvegabor: Remove comment out code

kisvegabor (commented):
What is the intended LVGL-EVE interface to send the SPI commands? 

I think we should have some callbacks similar to the other display controller drivers.

src/libs/FT800-FT813/lvgl_version_note.txt
liamHowatt: Mainly to know which exact version of the library this is.

kisvegabor:
Please add the use EVE lib here: https://github.com/lvgl/lvgl/blob/master/COPYRIGHTS.md
and add docs too.

docs/src/details/integration/renderers/eve.rst
```diff
@@ -0,0 +1,233 @@
...
+These parameters were taken from the collection of configs provided by the EVE
+library supporting this renderer.
+`See here <https://github.com/RudolphRiedel/FT800-FT813/blob/ee35fa45c37aed78dc4f1196b874ee1c8f80429f/EVE_config.h#L856-L872>`__.
+Check there for your board parameters.
+
+.. code-block:: c
+
+    #pragma once
+
+    #define EVE_HSIZE (800L)
```
kisvegabor: It's a little against how we usually handle configs. We should either use `lv_conf.h` for these, but as there are a lot of options passing them as a larger struct  to `lv_draw_eve_display_create()` might be better.

docs/src/details/integration/renderers/eve.rst
```diff
@@ -0,0 +1,233 @@
...
+------------------------------------
+
+The user is required to implement four functions which the LVGL EVE
+renderer will call to perform SPI communication with the EVE chip.
+The functions can simply be defined in your own files. They must not be
+``static``.
+
+.. code-block:: c
+
+    #include "lvgl/lvgl.h"
```
kisvegabor: Passing these as callbacks in the large struct?

docs/src/details/integration/renderers/eve.rst
```diff
@@ -0,0 +1,233 @@
...
+The ``src`` parameter of ``lv_draw_eve_target_spi_transmit`` is not a
+pointer to ``const`` since your implementation may naturally want to
+overwrite the outgoing data with (meaningless) incoming data
+(which will be ignored by LVGL).
+
+In practice, the ``len`` parameters will be only 1 or 4 for transmit, and
+always 1 for receive. A forward-compatible implementation should account
+for other values, though.
+
+Here is an example implementation for ESP32-S3.
```
kisvegabor: Can you replace these descriptions by adding dummy function calls in the skeletons above?

docs/src/details/integration/renderers/eve.rst
```diff
@@ -0,0 +1,233 @@
...
+
+.. code-block:: c
+
+    #include "lvgl/lvgl.h"
```
kisvegabor: Or keep only these instead of the skeletons with more comments.

docs/src/details/integration/renderers/eve.rst
```diff
@@ -0,0 +1,233 @@
...
+implementation. This ``init_spi`` can happen after :cpp:func:`lv_init`
+but **must** happen before :cpp:func:`lv_draw_eve_display_create`,
+which is discussed in the next section.
+
+22 MHz was the highest SPI speed that worked during testing with the Riverdi board
+and the ESP32-S3. You may not have success with this speed so it is
+recommended to validate with an ``SPI_SPEED`` value of ``10`` (10 MHz)
+and increase experimentally in your testing.
+
+.. code-block:: c
```
kisvegabor: Being a general SPI initialization it seems unrelated to LVGL and EVE.

kisvegabor (dismissed):
LGTM! :tada:

kisvegabor:
FYI, there are conflicts.

docs/src/details/integration/renderers/eve.rst
```diff
@@ -0,0 +1,127 @@
+=========================
+EVE External GPU Renderer
+=========================
+
+EVE is a type of external GPU IC which accepts high-level drawing commands over SPI
+and outputs the rendered graphics to a display over parallel RGB.
+
+The advantage of using an EVE chip that the rendering responsibility is removed
```
AndreCostaaa:
```suggestion
The advantage of using an EVE chip is that the rendering responsibility is removed
```

src/draw/eve/lv_draw_eve.c
```diff
@@ -0,0 +1,172 @@
...
+static int32_t eve_dispatch(lv_draw_unit_t * draw_unit, lv_layer_t * layer)
+{
+    lv_draw_eve_unit_t * draw_eve_unit = (lv_draw_eve_unit_t *) draw_unit;
+
+    /*Return immediately if it's busy with draw task*/
+    if(draw_eve_unit->task_act) return 0;
+
+    lv_draw_task_t * t = NULL;
+    t = lv_draw_get_next_available_task(layer, NULL, DRAW_UNIT_ID_EVE);
+    if(t == NULL) return -1;
```
AndreCostaaa:
```suggestion
    if(t == NULL) return LV_DRAW_UNIT_IDLE;
```

To avoid magic numbers

src/draw/eve/lv_draw_eve.c
```diff
@@ -0,0 +1,172 @@
...
+/**********************
+ *   STATIC FUNCTIONS
+ **********************/
+
+static int32_t eve_dispatch(lv_draw_unit_t * draw_unit, lv_layer_t * layer)
+{
+    lv_draw_eve_unit_t * draw_eve_unit = (lv_draw_eve_unit_t *) draw_unit;
+
+    /*Return immediately if it's busy with draw task*/
+    if(draw_eve_unit->task_act) return 0;
```
AndreCostaaa: The code that calls this function doesn't check for 0:
```c
    else if(remove_task || layer->draw_task_head) {
        /*Find a draw unit which is not busy and can take at least one task*/
        /*Let all draw units to pick draw tasks*/
        lv_draw_unit_t * u = _draw_info.unit_head;
        while(u) {
            LV_PROFILER_DRAW_BEGIN_TAG("dispatch_cb");
            LV_PROFILER_DRAW_BEGIN_TAG(u->name);
            LV_PROFILER_DRAW_END_TAG(u->name);
            LV_PROFILER_DRAW_END_TAG("dispatch_cb");
            if(taken_cnt != LV_DRAW_UNIT_IDLE) task_dispatched = true;
            u = u->next;
        }
    }

    LV_PROFILER_DRAW_END;
    return task_dispatched;
```
So if i'm understating correctly returning 0 here means we could lose the task if no other draw unit is able to dispatch it but we still return `true` which we shouldn't right ?

Do we need to update the code in `bool lv_draw_dispatch_layer(lv_display_t * disp, lv_layer_t * layer)` ? 

Also noticed dma2d does the same thing so maybe i'm missing something

src/draw/eve/lv_draw_eve.c
```diff
@@ -0,0 +1,172 @@
...
+    eve_execute_drawing(draw_eve_unit);
+
+    draw_eve_unit->task_act->state = LV_DRAW_TASK_STATE_READY;
+    draw_eve_unit->task_act = NULL;
+
+    /*The draw unit is free now. Request a new dispatching as it can get a new task*/
+    lv_draw_dispatch_request();
+
+
+    return 1;
```
AndreCostaaa, line 112: Probably need to add a new #define for this value, or even better, replace the return type with a enum
dma2d does the same thing so this would be nice to tackle in a different PR

src/draw/eve/lv_draw_eve_image.c
```diff
@@ -0,0 +1,249 @@
...
+            case LV_COLOR_FORMAT_RGB565A8 :
+                convert_RGB565A8_to_ARGB1555(img_src, temp_buff, stride_px_w, img_h);
+                buffer_converted = temp_buff;
+                break;
+            case LV_COLOR_FORMAT_ARGB8888 :
+                convert_ARGB8888_to_ARGB4444(img_src, temp_buff, stride_px_w, img_h);
+                buffer_converted = temp_buff;
+                break;
+            default :
+                break;
```
AndreCostaaa: Is this unreachable or should we ensure we don't get here else `buffer_converted` will be NULL for the code that follows ?

src/draw/eve/lv_draw_eve_letter.c
```diff
@@ -0,0 +1,264 @@
...
+                nibble_1 = get_nibble_2(img_in[addr_it - 1]);
+                nibble_2 = get_nibble_1(img_in[addr_it]);
+            }
+
+            /*Combine the nibbles and assign the result to the output byte*/
+            img_out[j] = (nibble_1 << 4) | nibble_2;
+
+            addr_it++;
+            j++;
+        }
```
AndreCostaaa: `key` doesn't change in this loop so we should probably not check for it each time. Compiler may not optimize this and extra branches are expensive due to branch prediction. Not critical tho i would say but maybe add a TODO so we can get to this later ?

src/draw/eve/lv_draw_eve_line.c
```diff
@@ -0,0 +1,71 @@
...
+
+
+    uint32_t line_w = dsc->width * 8;
+    lv_eve_scissor(t->clip_area.x1, t->clip_area.y1, t->clip_area.x2, t->clip_area.y2);
+    lv_eve_save_context();
+    lv_eve_color_opa(dsc->opa);
+    lv_eve_color(dsc->color);
+
+    if(dsc->dash_gap && dsc->dash_width) {
+        /*TODO*/
```
AndreCostaaa: I propose adding a LV_LOG_WARN so we can easily debug if we get here and something is not working correctly

src/draw/eve/lv_draw_eve_line.c
```diff
@@ -0,0 +1,71 @@
...
+        lv_eve_primitive(LV_EVE_PRIMITIVE_RECTS);
+        if(is_vertical) {
+            /* Draw vertical line */
+            lv_eve_vertex_2f(dsc->p1.x, dsc->p1.y);
+            lv_eve_vertex_2f(dsc->p2.x, dsc->p2.y);
+        }
+        else {
+            /* Draw horizontal line */
+            lv_eve_vertex_2f(dsc->p1.x, dsc->p1.y);
+            lv_eve_vertex_2f(dsc->p2.x, dsc->p2.y);
```
AndreCostaaa: Code duplication, so i guess the `if else` can go ?

src/draw/eve/lv_draw_eve_ram_g.c
```diff
@@ -0,0 +1,171 @@
...
+static uint32_t ramGptr = EVE_RAM_G;
+
+
+/**********************
+ *   GLOBAL FUNCTIONS
+ **********************/
+
+uint32_t lv_draw_eve_get_bitmap_addr(uint8_t id)
+{
+    return blocks[id].address;
```
AndreCostaaa: Can we add a check here to make sure id is not out of bounds

src/draw/eve/lv_draw_eve_ram_g.c
```diff
@@ -0,0 +1,171 @@
...
+    return NOT_FOUND_BLOCK;
+}
+
+
+void lv_draw_eve_set_size_ramg_block(uint8_t Id, uint32_t sz)
+{
+
+    blocks[Id].size = sz;
+
+}
```
AndreCostaaa: Same check here and as a nit, can we stick to snake case?

src/draw/eve/lv_draw_eve_ram_g.c
```diff
@@ -0,0 +1,171 @@
...
+    blocks[id].id = id;
+    lv_draw_eve_update_ramg_ptr(sz);
+
+    if(ramGptr > EVE_RAM_G_SIZE) {
+        return true;
+    }
+    else {
+        return false;
+    }
+}
```
AndreCostaaa: Same concerns here

src/draw/eve/lv_draw_eve_ram_g.h
```diff
@@ -0,0 +1,110 @@
...
+ **********************/
+
+/* Function to update the RAM_G pointer with the given size.*/
+void lv_draw_eve_update_ramg_ptr(uint32_t size);
+
+/* Function to get the current RAM_G pointer value. */
+uint32_t lv_draw_eve_get_ramg_ptr(void);
+
+/* Init Memory blocks */
+void lv_draw_eve_init_eve_ramg();
```
AndreCostaaa:
```suggestion
void lv_draw_eve_init_eve_ramg(void);
```

src/drivers/draw/lv_draw_eve_display.c
```diff
@@ -0,0 +1,100 @@
...
+
+lv_display_t * lv_draw_eve_display_create(const lv_draw_eve_parameters_t * params, lv_draw_eve_operation_cb_t op_cb,
+                                          void * user_data)
+{
+    /* The buffer is not used, so just set something. */
+    static lv_draw_buf_t draw_buf;
+    static uint8_t dummy_buf; /* It won't be used as it will send commands instead of draw pixels. */
+    lv_draw_buf_init(&draw_buf, params->hor_res, params->ver_res, LV_COLOR_FORMAT_NATIVE,
+                     params->hor_res * LV_COLOR_FORMAT_GET_SIZE(LV_COLOR_FORMAT_NATIVE),
+                     &dummy_buf, params->hor_res * params->ver_res * LV_COLOR_FORMAT_GET_SIZE(LV_COLOR_FORMAT_NATIVE));
```
AndreCostaaa: I believe we can send NULL instead of `&dummy_buf` here. Not critical tho

AndreCostaaa (changes requested):
Overall very good job. Mainly questions

src/draw/eve/lv_draw_eve.c
liamHowatt: It means the number of tasks that were taken.

https://github.com/lvgl/lvgl/blob/02d318168471224aa6b8e11fbb39c4fc8f0f8e3b/src/draw/sw/lv_draw_sw.c#L335

https://github.com/lvgl/lvgl/blob/02d318168471224aa6b8e11fbb39c4fc8f0f8e3b/src/draw/lv_draw.c#L283

src/draw/eve/lv_draw_eve.c
liamHowatt: draw_sw also returns 0. It signals that it's not idle (has an active task), and it didn't take any tasks.

https://github.com/lvgl/lvgl/blob/02d318168471224aa6b8e11fbb39c4fc8f0f8e3b/src/draw/sw/lv_draw_sw.c#L305-L309

In this EVE case, this `if` would never happen because you can see `task_act` is set to `t` and then `NULL` again before this function returns.

liamHowatt: So I will remove it.

src/drivers/draw/lv_draw_eve_display.c
liamHowatt: Can't because of this

https://github.com/lvgl/lvgl/blob/02d318168471224aa6b8e11fbb39c4fc8f0f8e3b/src/draw/lv_draw_buf.c#L275

I copied this

https://github.com/lvgl/lvgl/blob/02d318168471224aa6b8e11fbb39c4fc8f0f8e3b/src/drivers/sdl/lv_sdl_window.c#L140-L143

src/draw/eve/lv_draw_eve_letter.c
liamHowatt: Do you know how else it could be done?

kisvegabor:
Note the CI issues.

Anything is still missing?

liamHowatt:
In touch with a partner who may have a strong case to move the chip reset responsibility out of LVGL, similar to the ft8xx framebuffer driver. Awaiting their testimony.

src/draw/eve/lv_draw_eve.c
```diff
@@ -0,0 +1,169 @@
...
+#include "../../stdlib/lv_string.h"
+#include "lv_draw_eve_ram_g.h"
+#include "lv_draw_eve.h"
+#include "lv_eve.h"
+
+/*********************
+ *      DEFINES
+ *********************/
+
+#define DRAW_UNIT_ID_EVE 9
```
uLipe, line 30: General, non-blocking, question: Is there any criteria to select the ID of the draw unit?

src/draw/eve/lv_draw_eve.c
```diff
@@ -0,0 +1,169 @@
...
+
+/**********************
+ *      TYPEDEFS
+ **********************/
+
+lv_draw_eve_unit_t * lv_draw_eve_unit_g;
```
uLipe: Is it not a typedef, right? Also is this pointer intended to be global?

src/draw/eve/lv_draw_eve.c
```diff
@@ -0,0 +1,169 @@
...
+
+
+    return 1;
+}
+
+static int32_t eve_evaluate(lv_draw_unit_t * draw_unit, lv_draw_task_t * task)
+{
+    LV_UNUSED(draw_unit);
+
+    if(((lv_draw_dsc_base_t *)task->draw_dsc)->user_data == NULL) {
```
uLipe, line 119: Since this thing is not trivial as others draw unit, what about a comment explaining this section? I understand in cases where EVE exist, no other draw units will take a task, right?

src/draw/eve/lv_draw_eve_image.c
```diff
@@ -0,0 +1,225 @@
...
+ **********************/
+
+
+
+/**********************
+ *      MACROS
+ **********************/
+
+#define F16_PIVOT_SHIFT(x) ((int32_t)((((x) >> 1)) * 65536L))
+#define F16_SCALE_DIV_256(x) ((int32_t)(((x) / 256.0f) * 65536L))
```
uLipe, line 49: We don't have this kind of macros in LVGL? If so what about (in next PR) move these to a general include file, then other modules can reuse them?

src/draw/eve/lv_draw_eve_image.c
```diff
@@ -0,0 +1,225 @@
...
+
+
+
+
+
+/**********************
+ *   STATIC FUNCTIONS
+ **********************/
+
+static void convert_RGB565A8_to_ARGB1555(const uint8_t * src, uint8_t * dst, uint16_t width, uint16_t height)
```
uLipe: This function as well can be also part of LVGL public interface

src/draw/eve/lv_draw_eve_image.c
```diff
@@ -0,0 +1,225 @@
...
+
+        uint16_t argb1555 = (a1 << 15) | (r5 << 10) | ((g6 >> 1) << 5) | b5;
+
+        dst[2 * i] = argb1555 & 0xFF;
+        dst[2 * i + 1] = (argb1555 >> 8) & 0xFF;
+
+    }
+}
+
+static void convert_ARGB8888_to_ARGB4444(const uint8_t * src, uint8_t * dst, uint16_t width, uint16_t height)
```
uLipe: This one as well

src/draw/eve/lv_draw_eve_letter.c
```diff
@@ -0,0 +1,264 @@
...
+ *      TYPEDEFS
+ **********************/
+
+/**********************
+ *  STATIC PROTOTYPES
+ **********************/
+
+static void lv_draw_eve_letter_cb(lv_draw_task_t * t, lv_draw_glyph_dsc_t * glyph_draw_dsc,
+                                  lv_draw_fill_dsc_t * fill_draw_dsc, const lv_area_t * fill_area);
+static void bitmap_to_evenWidth(uint32_t addr, const uint8_t * img_in, uint8_t * img_out, uint16_t width,
```
uLipe: Nit: out from snake case naming for functions?

src/draw/eve/lv_draw_eve_ram_g.c
```diff
@@ -0,0 +1,167 @@
...
+
+uint32_t lv_draw_eve_get_bitmap_addr(uint8_t id)
+{
+    LV_ASSERT(id < MAX_BLOCKS);
+    return blocks[id].address;
+}
+
+void lv_draw_eve_update_ramg_ptr(uint32_t size)
+{
+    ramGptr += size;
```
uLipe: Nit: out from snake case

uLipe (commented):
Nice work, I have some questions :)

kisvegabor (dismissed):

uLipe (dismissed):

src/draw/eve/lv_draw_eve.c
liamHowatt: No criteria. I see we tend to choose the smallest number unused by other draw units. Some use larger numbers though.

src/draw/eve/lv_draw_eve.c
liamHowatt: Good point. I don't know why it's like this. It's done in the opengl and sdl draw units. I assume the setter of the draw_dsc's user_data would potentially be the user, but that doesn't explain why we'd not take the task if it's set. cc @kisvegabor

src/draw/eve/lv_draw_eve_image.c
liamHowatt: As far as I can tell, `F16_SCALE_DIV_256` is to satisfy an encoding requirement for an EVE-specific command. If it's used elsewhere in LVGL, let me know.

src/draw/eve/lv_draw_eve_image.c
liamHowatt: After your review it became more specialized. It just converts 1 row now. argb4444 is not used elsewhere. I'd wait until we have more general support for argb4444.

src/draw/eve/lv_draw_eve.c
liamHowatt: Thanks for mentioning it. Fixed.

liamHowatt:
I responded and addressed the remainder of [this review](https://github.com/lvgl/lvgl/pull/8211#pullrequestreview-2906057700) @uLipe

src/draw/eve/lv_draw_eve.c
kisvegabor: In case of SDL/OpenGL we set the [draw_task's user data](https://github.com/lvgl/lvgl/blob/master/src/draw/sdl/lv_draw_sdl.c#L238) to distinguish between:
1. It needs to be rendered by SDL (user_data is NULL)
2. It needs to the be rendered by SW (user_data is set so that SDL [won't take it](https://github.com/lvgl/lvgl/blob/master/src/draw/sdl/lv_draw_sdl.c#L187).)

Such magic shouldn't be required here.

kisvegabor:
Please rebase to master to fix the docs build issue.

docs/src/details/integration/renderers/eve.rst
```diff
@@ -0,0 +1,186 @@
...
+     1-0: 2-bit sampling clocks val           use 1 (the reset default)
+    */
+    lv_draw_eve_memwrite16(disp, LV_EVE_REG_TOUCH_CONFIG, 0x0151);
+
+
+Display Rotation
+----------------
+
+Efficient display rotation is fully supported through :cpp:func:`lv_display_set_rotation`.
+Touch input rotation is handled accordingly.
```
unitware, line 165: Hi, I have tested this without success on a VM810C50A-D, a resistive touch display 800x480

At what time can I call this, I have limited success if I call it after everything is initialized and up and running (the ui is drawn on screen) but I'd prefer to do it before showing the ui. 

Limited success means that the display is rotated but but not redrawn, and the touch is not working anymore. 

If I call the rotation change directly after creating the display or after display and touch then no display is available at all

It is entirely possible that I do not know how to do this properly, please advice :)

liamHowatt: Thanks for testing it!

I just tried rotating the screen on a timer interval and it worked for me.

You are using `lv_display_set_rotation` to do the rotation, right? When do you call it? Not during rendering, I assume?

It's strange that it doesn't even work immediately after creating the display and touch. Does it work if you don't init touch?

unitware: I get a very short flash from the backlight when setting it after display creation, before touch.

unitware: can you set it up in portrait mode from the start?
I get half way there using `#define EVE_ROTATE 3` but the lvgl display is still in landscape mode then

unitware: I did put my relevant files in a gist: https://gist.github.com/unitware/136616a93532c1ad93ced84082fbb84c
The target system is rp2350 and VM810C50A-D
I'm using the pr branch, v9.3.0-222-g015a983ed

unitware: This turns out to be a simple fix I guess. After enabling logging I get
```
[Error] (0.238, +238)    layer_reshape_draw_buf: Asserted at expression: ret != NULL (NULL pointer) lv_refr.c:836
```

Commenting out that assertion does the trick. I guess Eve does not have draw buffers and should return null?
```
/**
 * Reshape the draw buffer if required
 * @param layer  pointer to a layer which will be drawn
 */
static void layer_reshape_draw_buf(lv_layer_t * layer, uint32_t stride)
{
    lv_draw_buf_t * ret = lv_draw_buf_reshape(
                              layer->draw_buf,
                              layer->color_format,
                              lv_area_get_width(&layer->buf_area),
                              lv_area_get_height(&layer->buf_area),
                              stride);
    LV_UNUSED(ret);
    // LV_ASSERT_NULL(ret);
}
```

liamHowatt: Thank you so much. I have been testing on two devices. One has these asserts disabled. The other has the asserts enabled _but_ the screen is square so the reshape maybe does not fail for that reason. I will fix it this afternoon and push the changes. Your gist code looks good to me. Thanks again for trying it!

unitware: Thank you for the timely pull request as I was just looking for this lvgl driver :)

liamHowatt: I'm glad it will have users. Should be fixed now.

Eekle: Not just @unitware either! I'm thrilled this is being landed, it's going to solve a ton of problems.

Eekle:
<details><summary>Picture</summary>
<p>

image

</p>
</details>  

Working with almost no effort on my setup - Fantastic work, thank you.

kisvegabor (approved):

AndreCostaaa (approved):

PR #8215, opened 2025-05-08
feat(osal): Support for Threadx

Linjieqiang:
> Eclipse ThreadX is an advanced real-time operating system (RTOS) designed specifically for deeply embedded applications.

Due to daily use, I hope lvgl can support ThreadX OS system. If have any questions, please feel free to ask and I will make changes.

Since [PR8198 ](https://github.com/lvgl/lvgl/pull/8198) has occupied macro definition 8, so I define LV_OS_THREADX as 9. 

Thanks all.

### DIFF
--- a/src/osal/lv_threadx.c
+++ b/src/osal/lv_threadx.c
@@ -1,0 +1,114 @@
+/*********************
+ *      INCLUDES
+ *********************/
+#include "lv_os.h"
+
+#if LV_USE_OS == LV_OS_THREADX
+
+/*********************
+ *      DEFINES
+ *********************/
+
+#define THREAD_TIMESLICE 20U
+
+/**********************
+ *      TYPEDEFS
+ **********************/
+
+/**********************
+ *  STATIC PROTOTYPES
+ **********************/
+
+/**********************
+ *  STATIC VARIABLES
+ **********************/
+
+/**********************
+ *      MACROS
+ **********************/
+
+/**********************
+ *   GLOBAL FUNCTIONS
+ **********************/
+lv_thread_init(lv_thread_t * thread, lv_thread_prio_t prio, void (*callback)(void *), size_t stack_size,
+               void * user_data)
+{
+    uint8_t * data = malloc(stack_size);
+    if(data == NULL) {
+        return LV_RESULT_INVALID;
+    }
+
+    UINT result = tx_thread_create(&thread->thread,
+                                   "lv_thread",
+                                   (VOID (*)(ULONG))callback,
+                                   (ULONG)user_data,
+                                   data,
+                                   stack_size,
+                                   prio,
+                                   prio,
+                                   THREAD_TIMESLICE,
+                                   TX_AUTO_START);
+
+    if(result != TX_SUCCESS) {
+        free(data);
+        return LV_RESULT_INVALID;
+    }
+
+    return LV_RESULT_OK;
+}
+
+lv_result_t lv_thread_delete(lv_thread_t * thread)
+{
+    return (tx_thread_delete(&thread->thread) == TX_SUCCESS) ? LV_RESULT_OK : LV_RESULT_INVALID;
+}
+
+lv_result_t lv_mutex_init(lv_mutex_t * mutex)
+{
+    return (tx_mutex_create(&mutex->mutex, "mutex", 0) == TX_SUCCESS) ? LV_RESULT_OK : LV_RESULT_INVALID;
+}
+
+lv_result_t lv_mutex_lock(lv_mutex_t * mutex)
+{
+    return (tx_mutex_get(&mutex->mutex, TX_WAIT_FOREVER) == TX_SUCCESS) ? LV_RESULT_OK : LV_RESULT_INVALID;
+}
+
+lv_result_t lv_mutex_lock_isr(lv_mutex_t * mutex)
+{
+    return (tx_mutex_get(&mutex->mutex, TX_WAIT_FOREVER) == TX_SUCCESS) ? LV_RESULT_OK : LV_RESULT_INVALID;
+}
+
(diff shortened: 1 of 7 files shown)


src/osal/lv_threadx.c
```diff
@@ -0,0 +1,114 @@
...
+ *      MACROS
+ **********************/
+
+/**********************
+ *   GLOBAL FUNCTIONS
+ **********************/
+lv_thread_init(lv_thread_t * thread, lv_thread_prio_t prio, void (*callback)(void *), size_t stack_size,
+               void * user_data)
+{
+    uint8_t * data = malloc(stack_size);
```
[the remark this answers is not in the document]
Linjieqiang: Fixed.

src/osal/lv_threadx.c
```diff
@@ -0,0 +1,114 @@
...
+}
+
+lv_result_t lv_mutex_lock(lv_mutex_t * mutex)
+{
+    return (tx_mutex_get(&mutex->mutex, TX_WAIT_FOREVER) == TX_SUCCESS) ? LV_RESULT_OK : LV_RESULT_INVALID;
+}
+
+lv_result_t lv_mutex_lock_isr(lv_mutex_t * mutex)
+{
+    return (tx_mutex_get(&mutex->mutex, TX_WAIT_FOREVER) == TX_SUCCESS) ? LV_RESULT_OK : LV_RESULT_INVALID;
```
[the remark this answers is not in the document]
Linjieqiang: I looked through the codes of other OS, and they all used the method of permanent waiting to release the mutex lock.

kisvegabor:
Hey,

Thank you very much! Very nice and clean implementation! 

Could you add an idle measurement function too? Similar to [this](https://github.com/lvgl/lvgl/blob/master/src/osal/lv_freertos.c#L406-L420). (Maybe it's easier to implement in ThreadX)

src/osal/lv_threadx.c
```diff
@@ -0,0 +1,118 @@
...
+
+/**********************
+ *   GLOBAL FUNCTIONS
+ **********************/
+static uint8_t * thread_stack = NULL;
+
+lv_thread_init(lv_thread_t * thread, lv_thread_prio_t prio, void (*callback)(void *), size_t stack_size,
+               void * user_data)
+{
+    thread_stack = malloc(stack_size);
```
AndreCostaaa, line 38: After taking a look at this [example](https://github.com/eclipse-threadx/threadx/blob/master/samples/demo_threadx.c), it seems like we should probably use a _byte pool_ instead of malloc. Is there a valid reason to use malloc instead ?

src/osal/lv_threadx.h
```diff
@@ -0,0 +1,49 @@
...
+/*********************
+ *      DEFINES
+ *********************/
+
+/**********************
+ *      TYPEDEFS
+ **********************/
+typedef struct {
+    TX_THREAD thread;
+} lv_thread_t;
```
AndreCostaaa, line 25: Can you store the _thread stack_ here ?

AndreCostaaa (changes requested):
Hi, thank you for your PR. Copilot's concerns are valid, can you take a look and respond directly in the corresponding thread ?

src/osal/lv_threadx.c
```diff
@@ -0,0 +1,118 @@
...
+    if(thread_stack == NULL) {
+        return LV_RESULT_INVALID;
```
uLipe, line 40: Using LV_LOG_ERR may save some debug time if heap area is too small

src/osal/lv_threadx.c
```diff
@@ -0,0 +1,118 @@
...
+                                   (ULONG)user_data,
+                                   thread_stack,
+                                   stack_size,
+                                   prio,
+                                   prio,
+                                   THREAD_TIMESLICE,
+                                   TX_AUTO_START);
+
+    if(result != TX_SUCCESS) {
+        free(thread_stack);
```
uLipe, line 55: Here too :)

src/osal/lv_threadx.c
```diff
@@ -0,0 +1,118 @@
...
+}
+
+lv_result_t lv_mutex_lock(lv_mutex_t * mutex)
+{
+    return (tx_mutex_get(&mutex->mutex, TX_WAIT_FOREVER) == TX_SUCCESS) ? LV_RESULT_OK : LV_RESULT_INVALID;
+}
+
+lv_result_t lv_mutex_lock_isr(lv_mutex_t * mutex)
+{
+    return (tx_mutex_get(&mutex->mutex, TX_WAIT_FOREVER) == TX_SUCCESS) ? LV_RESULT_OK : LV_RESULT_INVALID;
```
uLipe, line 81: Hmm I don't know about other OS but it would not be better to set a finite timeout instead of forever? It could be a potential deadlock source

uLipe (changes requested):
Nice work!

I also vote for handling the copilot concerns.

I also have some questions.

src/osal/lv_threadx.c
liamHowatt: How can we decide for threadx users how long is "too long" to wait for a resource? Any value arbitrary. I'd rather deadlock than unknowingly proceed without the lock.

liamHowatt (commented):
Add to the top of lv_threadx.c and lv_threadx.h

```c
/**
 * @file the_file_name.h
 *
 */

```

src/osal/lv_threadx.c
uLipe: Deadlocks failures are silent, in general will take some time to find where is the problematic locking location.

Having a timeout followed by receiving an error and logging it may save some debug of wrong mutex lock-unlocking sequence. I would suggest to extend the API of lock to include timeouts but this would break the entire OSAL API.

